PR #13, opened 2021-01-19
Berlin: Centers check marks and makes them actual checks.

MicahZoltu:


MadeofTin (approved):
Cleaver with the : , Thanks for edits.

PR #15, opened 2021-02-20
Update ethereumjs consensus

jochem-brouwer:
Update from EthereumJS: we intend to participate in YoloV3. We have most EIPs ready, except EIP 2930, which should be ready for merge soon.

CC @poojaranjan

### DIFF
--- a/client-integration-testnets/YOLOv3.md
+++ b/client-integration-testnets/YOLOv3.md
@@ -29,7 +29,7 @@ YOLO-v3
 | **Client**   | Signal | Spec | Merged | Syncing |
 |--------------|--------|------|--------|---------|
 | Besu         | x      | x    |  x     |   x     |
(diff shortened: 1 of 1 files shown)


jochem-brouwer:
That was fast! 😄  Thanks!

poojaranjan:
Because YOLOv3 EIPs are most likely going into the Berlin upgrade, we're adding the implementation status in [Berlin](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md) file because there is no more Meta EIP for the network upgrade.

Could you also create a PR with updates on individual EIPs for the record for [Berlin upgrade](https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/berlin.md)?

jochem-brouwer:
Yup, see #16!

poojaranjan:
Thanks!

PR #30, opened 2021-04-15
Add Petersburg Block to aleut genesis.json

ryanschneider:
Without this change Geth gives the error `Fatal: Failed to write genesis block: unsupported fork ordering: petersburgBlock not enabled, but istanbulBlock enabled at 0`


timbeiko:
Closing, swapped `constantinopleFixBlock` for `pertersburgBlock` instead given that Besu supports both.

PR #31, opened 2021-04-16
Added a list of protocol releases

jpitts:
I created a helpful resource in order to see the full history of changes to the Ethereum protocol. This helps people see the big-picture of the evolution of the protocol, and connects implementers, developers, and others to the EIPs which went into each release / hard fork.

Additionally, it helps community members see changes to how protocol releases are managed and to find the relevant information / specification on each. This is particularly important as the new https://github.com/ethereum/eth1.0-specs becomes more central to the release process.

Here's a quick link so you can see how it displays on GItHub: https://github.com/jpitts/eth1.0-specs#ethereum-protocol-releases

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,22 @@
 ## Description
 
 This repository contains various specification related to the Ethereum 1.0 chain, specifically the specifications for [network upgrades](/network-upgrades) and (soon) the [JSON RPC API](/json-rpc). 
+
+## Ethereum Protocol Releases
+
+| Version and Code Name | Block No. | Released | Incl EIPs | Specs | Impls |
+|-----------------------|-----------|----------|-----------|-------|-------|
+| Frontier | 1 | 07/30/2015 | | | [Geth v1.0.0](https://github.com/ethereum/go-ethereum/releases/tag/v1.0.0) |
+| Frontier Thawing | 200000 | 09/07/2015 | | | [Geth v1.0.1.1](https://github.com/ethereum/go-ethereum/releases/tag/v1.0.1.1) |
+| Homestead | 1150000 | 03/14/2016  | [EIP-2](https://eips.ethereum.org/EIPS/eip-2) <br/> [EIP-7](https://eips.ethereum.org/EIPS/eip-7) <br/> [EIP-8](https://eips.ethereum.org/EIPS/eip-8) | [HFM-606](https://eips.ethereum.org/EIPS/eip-606) | [Geth v1.3.4](https://github.com/ethereum/go-ethereum/releases/tag/v1.3.4) |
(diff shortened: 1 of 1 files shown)


poojaranjan:
Thanks for this list @jpitts. Looks very useful for this repository.

timbeiko:
@jpitts thanks for this, it's great! I'm just a bit hesitant about having the client releases with just geth, given there are many more clients. I think we should either remove that column (little work) or perhaps add the various clients who were active for the various forks (lots of work!). WDYT?

timbeiko:
Oh, I just noticed that Istanbul is missing too. Will add it now !

poojaranjan:
> I think we should either remove that column (little work) or perhaps add the various clients who were active for the various forks (lots of work!). 

I can perhaps additional columns for different clients? Will be helpful for collecting upgrade ready version of different clients, 
thoughts?

poojaranjan:
> Oh, I just noticed that Istanbul is missing too. Will add it now !

Good catch @timbeiko, i was counting why is it not matching with the total number of upgrades so far 😅

timbeiko:
Added Istanbul [here](https://github.com/ethereum/eth1.0-specs/commit/1ab07631e1dddee2e8a30811745251c911e6380d). 

> I can perhaps add additional columns for different clients? Will be helpful for collecting upgrade ready version of different clients, thoughts?

I feel like that starts to be a lot, given the number of clients (and historical ones!). My preference would be to remove implementation references altogether to keep the repository focused on specs. The `upgrade.md` files have the clients in them, but IMO it's more of a "necessary evil" for project management than something we actively want to maintain.

poojaranjan:
>  to remove implementation references altogether 

Fine with it!

jpitts:
> @jpitts thanks for this, it's great! I'm just a bit hesitant about having the client releases with just geth, given there are many more clients. I think we should either remove that column (little work) or perhaps add the various clients who were active for the various forks (lots of work!). WDYT?

Listing all clients is probably is not appropriate for this particular list, and does entail some work. Going forward it would even start to clutter each release entry given how many viable clients there now are. 

Still, there is often helpful information in release notes, not to mention what is posted on the https://blog.ethereum.org. This info can be referenced in each specification file as more get created for the entire set of releases.

timbeiko:
@jpitts how about listing the EF blog entries instead of the clients, then? I think that will link back to the clients, and be a bit more neutral.

jpitts:
> @jpitts how about listing the EF blog entries instead of the clients, then? I think that will link back to the clients, and be a bit more neutral.

This sounds good, soon I'll gather a list of those posts and submit a PR.

PR #35, opened 2021-04-26
eth_call

alita-moore:


json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+4.3.4.1   | `eth_call` SHOULD throw exception `-32000` with message `err: intrinsic gas too low: have 13107, want 21000 (supplied gas 13107)`[!] if the `gas` input parameter is less than `implicit gas`[!]
+4.3.4.2   | `eth_call` SHOULD throw exception `-32000` with message `out of gas`
+4.3.5     | `eth_call` MUST NOT accept requests if the `gas` input parameter is not properly formatted
+4.3.5.1   | `eth_call` SHOULD throw exception `-32602` with message `invalid argument 0: json: cannot unmarshal hex number with leading zero digits into Go struct field CallArgs.gas of type hexutil.Uint64`[!] if the `gas` input parameter does not meet `hex encoded`[!] specifications
+4.3.6     | `eth_call` MUST NOT accept requests if the `gas` input parameter is greater than `2^64 - 1` or `0xffffffffffffffff`
+4.3.6.1   | `eth_call` SHOULD throw exception `-32602` with message `invalid argument 0: json: cannot unmarshal hex number > 64 bits into Go struct field CallArgs.gas of type hexutil.Uint64` if the `gas` input parameter is greater than the hex string encoded decimal `2^64 - 1`
```
MariusVanDerWijden, line 42: go specific

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+4.4.1     | `eth_call` MUST accept a request if `gasPrice` input parameter is equal to `null`[!]
+4.4.1.1   | `eth_call` MUST consider `GASPRICE`[!] as equal to 0 if `gasPrice` input parameter is equal to `null`[!]
+4.4.2     | `eth_call` MUST consider `GASPRICE`[!] as equal to `gasPrice` input parameter if `gasPrice` input parameter is not equal to null and meets `Quantity` requirements
+4.4.2.1   | `eth_call` SHOULD throw exception `-32000` with message `err: insufficient funds for gas * price + value: address {from} have {from account balance} want {gas * gasPrice + value} (supplied gas {gas})` if the `gasPrice` input parameter multiplied by `gas` input parameter added to `value` input paremeter are greater than the balance of the account associated with the `from` input parameter
+4.4.2.2   | `eth_call` SHOULD throw exception `-32000` with message `err: insufficient funds for gas * price + value: address 0x0000000000000000000000000000000000000000 have 0 want {gas * gasPrice + value} (supplied gas {gas})` if the `gasPrice` input parameter multiplied by `gas` input parameter added to `value` input paremeter are greater than 0 and the `from` input parameter is equal to null
```
MariusVanDerWijden, line 48:
```suggestion
4.4.2.2   | `eth_call` SHOULD throw exception `-32000` with message `err: insufficient funds for gas * price + value: address {from} have 0 want {gas * gasPrice + value} (supplied gas {gas})` if the `gasPrice` input parameter multiplied by `gas` input parameter added to `value` input paremeter are greater than 0 and the `from` input parameter is equal to null
```

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+4.2.3.1.1 | `eth_call` MUST return an empty hex-string [!] if the `to` parameter is defined and does not exist on-chain
+4.3       | `eth_call` MUST accept an input parameter `gas` of type `QUANTITY` [!]
+4.3.1     | `eth_call` MUST accept requests with `gas` equal to `null` [!]
+4.3.1.1   | `eth_call` MUST consider gas to equal 25 million if the `gas` parameter is equal to `null`[!]
+4.3.2     | `eth_call` MUST accept requests with `gas` equal to a value greater than block `GAS LIMIT`[!] 
+4.3.3     | `eth_call` MUST output `GAS`[!] to equal the `gas` input parameter minus the gas used at `GAS`[!] execution time
+4.3.3.1   | `eth_call` MUST output `GAS`[!] to equal `25 million`[!] minus the gas used at `GAS`[!] execution time
+4.3.4     | `eth_call` MUST NOT accept requests if the `gas` input parameter is less than `implicit gas`[!] or `actual gas`[!]
+4.3.4.1   | `eth_call` SHOULD throw exception `-32000` with message `err: intrinsic gas too low: have 13107, want 21000 (supplied gas 13107)`[!] if the `gas` input parameter is less than `implicit gas`[!]
```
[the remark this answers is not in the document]
MariusVanDerWijden: {gas || 25000000[!]}

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+4.5.1.2.2 | `eth_call` SHOULD throw exception `-32000` with message `err: insufficient funds for transfer: address {from} (supplied gas {gas || 25000000[!]})` if the `value` input parameter is greater than 0 AND the `from` input parameter is equal to `null`[!]
+5         | `eth_call` MUST accept a `params` parameter [JSON Array](https://tools.ietf.org/html/rfc7159#section-5) of length 2, and its second item (index = 1) containing a `Block Identifier`[!] parameter that specifies the block height of the best block ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) to assume the state of
+5.1       | `eth_call` MUST assume the on-chain state of the latest best block ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) if the `Block Identifier` parameter is equal to the [JSON String](https://tools.ietf.org/html/rfc7159#section-7) of "latest"; the latest best block height ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) is equal to the current best block height minus one.
```
MariusVanDerWijden, line 56:
```suggestion
5.1       | `eth_call` MUST assume the on-chain state of the latest best block ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) if the `Block Identifier` parameter is equal to the [JSON String](https://tools.ietf.org/html/rfc7159#section-7) of "latest"; the latest best block height ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) is equal to the current best block height.
```

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+5.2       | `eth_call` MUST assume the on-chain state of the current best block with the greatest height ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf))  if the `Block Identifier` parameter is equal to the [JSON String](https://tools.ietf.org/html/rfc7159#section-7) of "pending"; the pending best block height ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) is equal to the current best block height.
```
MariusVanDerWijden, line 57:
```suggestion
5.2       | `eth_call` MUST assume the on-chain state of the current best block with the greatest height ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf))  if the `Block Identifier` parameter is equal to the [JSON String](https://tools.ietf.org/html/rfc7159#section-7) of "pending"; the pending best block height ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) is equal to the current best block height plus a block with the best pending transactions.
```

MariusVanDerWijden (changes requested):
Mostly nitpicks,
Some thing that's missing from the spec is overrides I think

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
+
+
+# `eth_call`
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC-2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+Specification | Description 
+--|--
+1         | `eth_call` MUST NOT affect on-chain state|
+2         | `eth_call` MUST NOT accept requests that are not in valid JSON format
```
MicahZoltu, line 9: I think this is implicit, since the RPC is JSON-RPC so it is unnecessary to repeat it here.  Similar to how we don't specify how TCP or HTTP works in this specification as it is handled at the transport layer.

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+3         | `eth_call` MUST return a response in valid JSON format
```
MicahZoltu, line 10: Same as above, this is redundant with JSON-RPC specification.

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+4         | `eth_call` MUST accept a `params` parameter [JSON Array](https://tools.ietf.org/html/rfc7159#section-5) of length 2, and its first item containing a [JSON Object](https://tools.ietf.org/html/rfc7159#section-4) with keys known here as the `input parameters`
+4.1       | `eth_call` MUST accept an input parameter `from` of type `DATA` [!]
```
MicahZoltu, line 12: It appears the specification is worded with a target audience of server developers in mind.  This API is bi-directional and this document is likely to be read by both client developers and server developers.  In this case, the `from` property is optional but this line makes it sound like it is required if you are reading from the point of view of a client.  Consider trying to phrase things in this document in a way that is neutral to whether the reader is a server or client developer.

> The `from` property is optional, but if present its type **MUST** be 20 byte DATA.

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+4.1.1     | `eth_call` MUST accept requests with the from input parameter undefined
+4.1.1.1   | `eth_call` MUST consider CALLER account to be `0x0000000000000000000000000000000000000000000000000000000000000000` if `from` is not defined
+4.1.2     | `eth_call` MUST accept requests with the `from` input parameter defined
+4.1.2.1   | `eth_call` MUST only accept requests with the `from` parameter defined as a hex string with length 40
+4.1.2.1.1 | `eth_call` SHOULD throw error `-32602` [!] on improper length with message as `invalid argument 0: hex string has length {length}, want 40 for common.Address`
+4.1.2.2   | `eth_call` MUST NOT accept `from` hex strings with invalid characters
```
MicahZoltu, line 18: As mentioned above, I think parsing errors like this should be generalized at a higher layer, otherwise we'll have to re-specify them over and over again.  We should define an `ADDRESS` type for the JSON-RPC protocol and define how parsing errors should be handled in a top level spec, then in the individual specs like this one we can just say "`from` **MUST** be of type `ADDRESS` and not have to specify all of the details of what that means.

lightclient:
This is just a general comment, but I think the way this spec is laid out is really verbose. I think it would be helpful to discuss things in terms of the inputs and outputs. The JSON-RPC spec sort of does this: https://www.jsonrpc.org/specification

alita-moore:
@lightclient from the perspective of requirement definition, it's important to be very clear. While verbose, it also minimizes ambiguity. Which, is the goal of the specification. It's not necessarily meant to be readable by anyone other than someone implementing a JSON RPC. At least that's how I understand the task, but I'm happy to reconsider the approach of course.

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+4.1.2.2.1 | `eth_call` SHOULD throw error `-32602` [!] on improper hex string and return message `invalid argument 0: json: cannot unmarshal invalid hex string into Go struct field CallArgs.from of type common.Address`
```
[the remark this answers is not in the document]
davidmurdoch: I don't think the `message` itself should be explicitly specified at all. The `message` is for humans, and implementors may want to augment it with additional metadata (which is especially useful in Ganache and other blockchain simulators).

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+5.3       | `eth_call` MUST assume the on-chain state of the genesis best block ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) if the `Block Identifier` parameter is equal to the [JSON String](https://tools.ietf.org/html/rfc7159#section-7) of "earliest"; the earliest best block height ([yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf)) is equal to 0.
+5.4       | `eth_call` MUST assume the on-chain state of the best blockchain block with height equal to the `Quantity`[!] value of the `Block Identifier` parameter if the `Block Identifier` parameter is equal to a hex encoded `Quantity`[!] value. Height is equal to the number of blocks after the genesis block on the best blockchain (block number).
+5.5       | `eth_call` SHOULD throw exception `-32602` [1] with message `invalid argument 1: hex string without 0x prefix` if the provided `Block Identifier` parameter is neither a [JSON String](https://tools.ietf.org/html/rfc7159#section-7) value equal to `earliest`, `latest`, `pending` nor a hex encoded `Quantity`[!] value.
```
ryanschneider, line 60: [EIP-1898](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1898.md) allows more block identifiers such as the block hash.

json-rpc/eth_call.md
```diff
@@ -0,0 +1,113 @@
...
+
+specification | description 
+--|--
+1         | `eth_call` MUST NOT affect on-chain state|
+2         | `eth_call` MUST NOT accept requests that are not in valid JSON format
+3         | `eth_call` MUST return a response in valid JSON format
+4         | `eth_call` MUST accept a `params` parameter [JSON Array](https://tools.ietf.org/html/rfc7159#section-5) of length 2, and its first item containing a [JSON Object](https://tools.ietf.org/html/rfc7159#section-4) with keys known here as the `input parameters`
+4.1       | `eth_call` MUST accept an input parameter `from` of type `DATA` [!]
+4.1.1     | `eth_call` MUST accept requests with the from input parameter undefined
+4.1.1.1   | `eth_call` MUST consider CALLER account to be `0x0000000000000000000000000000000000000000000000000000000000000000` if `from` is not defined
```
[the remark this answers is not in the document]
fvictorio: There are 39 zeros here :innocent:

timbeiko:
Is this PR still relevant, @alita-moore ?

SamWilsn:

PR #45, opened 2021-05-08
Add information

ligi:
btw: I also think it could make sense to distinguish between "EIP not relevant" and "work not started" - currently they are handled the same - but we would get an better overview by seeing them as separate things IMHO


timbeiko:
My weak preference at this point is to not have any projects listed here who haven't proactively added their status as it may lead to unreliable information. Won't merge this now as we'll probably discuss with @tvanepps this week what is the best way to handle these :+1:

ligi:
So I need to remove the TREZOR one to get this merged?

timbeiko:
@ligi can you add the information here: https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/ecosystem-readiness.md? 

Thanks!

ligi:
replaced by #193

PR #47, opened 2021-05-11
1559 JSON RPC Diff

alita-moore:
(sorry about the bad commit name)
This..
- adds `baseFee` to the `Block` type
- removes `gasPrice` from `Transaction` type
- adds `maxPriorityFeePerGas` to `Transaction` type
- adds `maxFeePerGas` to `Transaction` type


MicahZoltu (approved):

json-rpc/spec.json
```diff
@@ -1464,6 +1464,11 @@
               "description": "Block hash of the RLP encoding of an uncle block",
               "$ref": "#/components/schemas/Keccak"
             }
+          },
+          "baseFee": {
+            "title": "baseFee",
+            "description": "The block's baseline network fee per gas; the baseFee is burned",
+            "$ref": "#/components/schemas/Integer"
           }
```
MicahZoltu, line 1472: This needs to be optional.

MicahZoltu (changes requested):
We need to make it clear that it is not valid to have both `gasPrice` and `maxPriorityFeePerGas` and other similar combinations.

json-rpc/spec.json
```diff
@@ -1464,6 +1464,11 @@
               "description": "Block hash of the RLP encoding of an uncle block",
               "$ref": "#/components/schemas/Keccak"
             }
+          },
+          "baseFee": {
+            "title": "baseFee",
+            "description": "The block's baseline network fee per gas; the baseFee is burned",
+            "$ref": "#/components/schemas/Integer"
```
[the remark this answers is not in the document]
alita-moore: good point, I can add that. Although goes a little beyond the scope of 1559 changes. As I understand it this was copy pasted from ETC so I'll be doing some conditioning like you mention soon

timbeiko:
@alita-moore FYI we agreed on ACD this AM to use the terms from the EIP, so `baseFeePerGas`, `maxFeePerGas`, `maxPriorityFeePerGas`.

json-rpc/spec.json
```diff
@@ -1464,6 +1464,11 @@
               "description": "Block hash of the RLP encoding of an uncle block",
               "$ref": "#/components/schemas/Keccak"
             }
+          },
+          "baseFee": {
+            "title": "baseFee",
```
timbeiko, line 1469:
```suggestion
            "title": "baseFeePerGas",
```

MicahZoltu (changes requested):
Would you mind splitting the spec changes from the spec rename?  Sadly, GitHub doesn't seem to recognize this is a file rename so we can't see the "diff", which is valuable from a history standpoint.

I recommend doing the file move *first*, so that change history is easily traversed.

holgerd77:
What is actually the reason for this renaming respectively is this really necessary? Will there be more specs in different formats in the future?

In case renaming is wanted short note that there is a link in the main README.md which then would also need an update.

lightclient:
I agree, I don't really see a reason to change until there are more spec formats.

alita-moore:
@holgerd77 the reason is because the [OPEN-RPC vscode plugin](https://marketplace.visualstudio.com/items?itemName=OPEN-RPC.OPEN-RPC) uses the `openrpc.json` as a way to find the files to parse. It's not required of course, but it is helpful when making changes. Either way I have reverted the change

timbeiko:
@alita-moore @lightclient @MicahZoltu is this OK to merge now?

alita-moore:
@MicahZoltu just a heads I'm going to address your comment about the integer in a separate PR

MicahZoltu:
If we are going to rename this (and I think @alita-moore's argument for renaming is quite reasonable) we should do it *before* this PR so history doesn't get confused by the rename.

alita-moore:
@MicahZoltu i will be making significant changes to the structure of the spec. I thinks it's best to just do it after. There is no practical difference in my opinion

json-rpc/spec.json
```diff
@@ -1497,76 +1523,112 @@
...
+            },
+            {
+                "oneOf": [
+                    {
+                        "title": "Block <1559> to PRESENT",
+                        "type": "object",
+                        "properties": {
+                            "maxPriorityFeePerGas": {
+                                "title": "maxPriorityFeePerGas",
+                                "description": "maximum fee per gas the sender is willing to pay to miners in wei",
```
lightclient, line 1608:
```suggestion
                                "description": "Maximum fee per gas the sender is willing to pay to miners in wei",
```

lightclient (approved):
looks correct to me

json-rpc/spec.json
```diff
@@ -1497,76 +1523,112 @@
...
+                                "$ref": "#/components/schemas/Integer"
+                            },
+                            "maxFeePerGas": {
+                                "title": "maxPriorityFeePerGas",
```
lightclient, line 1612:
```suggestion
                                "title": "maxFeePerGas",
```

lightclient (changes requested):
Just caught an issue actually.

json-rpc/spec.json
ShayneMichaelBacik123: Payment is required.donations to autodeposit email:shaynemichaelhall@gmail.com

PR #193, opened 2021-05-20
Add Solidity, WallETH, KEthereum

ligi:
replacement for #45 as the file changed

btw: I also think it could make sense to distinguish between "EIP not relevant" and "work not started" - currently they are handled the same - but we would get an better overview by seeing them as separate things IMHO
e.g. for solidity 1559 is not relevant - but 3198 is. Now this can only be represented by - | -


timbeiko:
@ligi thanks! Re: "Not relevant", I think ideally we can just make that as a green check box. This way, we can know they "support" it. WDYT?

PR #196, opened 2021-05-26
JSON-RPC: Add `eth_createAccessList` RPC call

MariusVanDerWijden:

PR #197, opened 2021-05-27
JSON-RPC: Add eth_createAccessList RPC call

MariusVanDerWijden:
WIPPPPP! @timbeiko :D, replaces https://github.com/ethereum/eth1.0-specs/pull/196

### DIFF
--- a/json-rpc/spec.json
+++ b/json-rpc/spec.json
@@ -146,6 +146,21 @@
       }
     },
     {
+      "name": "eth_createAccessList",
+      "summary": "Creates an access list for the given transaction.",
+      "params": [
+        {
+          "$ref": "#/components/contentDescriptors/Transaction"
+        },
+        {
+          "$ref": "#/components/contentDescriptors/BlockNumber"
+        }
+      ],
+      "result": {
+        "$ref": "#/components/contentDescriptors/AccessListResult"
+      }
+    },
+    {
       "name": "eth_chainId",
       "summary": "Returns the currently configured chain id",
       "description": "Returns the currently configured chain id, a value used in replay-protected transaction signing as introduced by [EIP-155](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md).",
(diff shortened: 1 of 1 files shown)


MariusVanDerWijden:
@lightclient asked
> Is this field optional? If an error is returned, is accessList and gasUsed omitted?

No, this field is not optional. If an error is returned the `accessList` and `gasUsed` are still returned. Only if the method creates a real error (state not available, block in the future, ...) then nothing is returned.
Returning the `vmErr` and the  `accessList` and `gasUsed` together is still useful in some cases.

MariusVanDerWijden:
The implementation in geth for this call can be found here:
https://github.com/ethereum/go-ethereum/pull/22550
(Interesting part: https://github.com/ethereum/go-ethereum/pull/22550/files#diff-c426ecd2f7d247753b9ea8c1cc003f21fa412661c1f967d203d4edf8163da344R1332)

Basically we implement the following algorithm:
```
al <- tx.AccessList // Take the current accesslist 
for {
   al2 <- execute(tx, al) // execute the transaction
   if al == al2 {
      return al // nothing changed between runs 
   }
   al = al2
}
```

fvictorio:
This feels inconsistent with the result of `debug_traceTransaction`, which has the `failed` and `returnValue` properties. Would it make sense to use those properties instead?

lightclient:
@MariusVanDerWijden is this something you're still working on or can it be closed?

MariusVanDerWijden:
@lightclient I'm not sure, did someone port this to the new spec already? If not I would like to keep it open as a reminder that this RPC call exists

lightclient:
@MariusVanDerWijden I see this was supposed to replace #196, but that was merged. So I'm not sure this is still a WIP or if #196 adequately describes the method?

I will port any new methods merged to this repo into the eth1-apis repo after London.

MariusVanDerWijden:
Yeah Tim merged #196 pretty quickly. but I wasn't really finished with the discussions yet, so he reverted the change. The codebase doesn't contain the `eth_createAccessList` RPC yet.

lightclient:
@MariusVanDerWijden ah right, I remember now. Should this PR be merged then (i.e. does it accurately describe the endpoint)? I can port it to the new spec once it's merged here.

SamWilsn:

PR #212, opened 2021-06-13
add feeHistory spec

zsfelfoldi:
This PR adds the feeHistory API function spec.
See also:
- original proposal: https://gist.github.com/zsfelfoldi/9607ad248707a925b701f49787904fd6
- implementation in Geth: https://github.com/ethereum/go-ethereum/pull/23033

### DIFF
--- a/json-rpc/spec.json
+++ b/json-rpc/spec.json
@@ -197,6 +197,96 @@
 			}
 		},
 		{
+			"name": "eth_feeHistory",
+			"summary": "Transaction fee history",
+			"description": "Returns base fee per gas and transaction effective priority fee per gas history for the requested block range if available. The range between headBlock-4 and headBlock is guaranteed to be available while retrieving data from the pending block and older history are optional to support. For pre-EIP-1559 blocks the gas prices are returned as rewards and zeroes are returned for the base fee per gas.",
+			"params": [
+				{
+					"name": "blockCount",
+					"description": "Number of blocks in the requested range. Between 1 and 1024 blocks can be requested in a single query. Less than requested may be returned if not all blocks are available.",
+					"required": true,
+					"schema": {
+						"$ref": "#/components/schemas/Integer"
+					}
+				},
+				{
+					"name": "newestBlock",
+					"description": "Highest number block of the requested range.",
+					"required": true,
+					"schema": {
+						"$ref": "#/components/schemas/BlockNumberOrTag"
+					}
+				},
+				{
+					"name": "rewardPercentiles",
+					"description": "A monotonically increasing list of percentile values to sample from each block's effective priority fees per gas in ascending order, weighted by gas used.",
+					"schema": {
+						"title": "rewardPercentiles",
+						"type": "array",
+						"items": {
+							"title": "rewardPercentile",
+							"description": "Floating point value between 0 and 100.",
+							"type": "number"
+						}
+					}
+				}
+			],
+			"result": {
+				"name": "feeHistoryResult",
+				"description": "Fee history for the returned block range. This can be a subsection of the requested range if not all blocks are available.",
+				"schema": {
+					"title": "feeHistoryResults",
+					"description": "Fee history results",
+					"type": "object",
+					"required": [
+						"firstBlock",
+						"baseFeePerGas",
+						"gasUsedRatio"
+					],
+					"properties": {
+						"oldestBlock": {
+							"title": "oldestBlock",
+							"description": "Lowest number block of the returned range.",
+							"$ref": "#/components/schemas/BlockNumber"
+						},
+						"baseFeePerGas": {
+							"title": "baseFeePerGasArray",
+							"description": "An array of block base fees per gas. This includes the next block after the newest of the returned range, because this value can be derived from the newest block. Zeroes are returned for pre-EIP-1559 blocks.",
+							"type": "array",
+							"items": {
+								"$ref": "#/components/schemas/Integer"
+							}
+						},
+						"gasUsedRatio": {
+							"title": "gasUsedArray",
+							"description": "An array of block gas used ratios. These are calculated as the ratio of gasUsed and gasLimit.",
+							"type": "array",
+							"items": {
+								"title": "gasUsedRatio",
+								"description": "Floating point value between 0 and 1.",
+								"type": "number"
+							}
+						},
+						"reward": {
+							"title": "rewardArray",
+							"description": "A two-dimensional array of effective priority fees per gas at the requested block percentiles.",
+							"type": "array",
+							"items": {
+								"title": "rewardPercentileArray",
+								"description": "An array of effective priority fee per gas data points from a single block. All zeroes are returned if the block is empty.",
+								"type": "array",
+								"items": {
+									"title": "rewardPercentile",
+									"description": "A given percentile sample of effective priority fees per gas from a single block in ascending order, weighted by gas used. Zeroes are returned if the block is empty.",
+									"$ref": "#/components/schemas/Integer"
+								}
+							}
+						}
+					}
+				}
+			}
+		},
+		{
 			"name": "eth_getBalance",
 			"summary": "Returns Ether balance of a given or account or contract",
 			"params": [
@@ -1135,6 +1225,17 @@
 					"pending"
 				]
 			},
+			"BlockNumberOrTag": {
+				"title": "blockNumberOrTag",
+				"oneOf": [
+					{
+						"$ref": "#/components/schemas/BlockNumber"
+					},
+					{
+						"$ref": "#/components/schemas/BlockNumberTag"
+					}
+				]
+			},
 			"BlockOrNull": {
 				"title": "blockOrNull",
 				"oneOf": [
@@ -1881,15 +1982,7 @@
 				"name": "blockNumber",
 				"required": true,
 				"schema": {
-					"title": "blockNumberOrTag",
-					"oneOf": [
-						{
-							"$ref": "#/components/schemas/BlockNumber"
-						},
-						{
-							"$ref": "#/components/schemas/BlockNumberTag"
-						}
-					]
+					"$ref": "#/components/schemas/BlockNumberOrTag"
 				}
 			},
 			"TransactionHash": {


json-rpc/spec.json
```diff
@@ -196,6 +196,101 @@
...
 			}
 		},
+		{
+			"name": "eth_feeHistory",
+			"summary": "Transaction fee history",
+			"description": "Returns transaction base fee and effective priority fee history for the requested/supported block range.",
+			"params": [
+				{
+					"name": "blockCount",
+					"description": "Number of blocks in the requested range.",
```
MicahZoltu, line 206: Consider mentioning what a value of `0` does/means.  Also, are there any constraints on this (limit to number one can request)?

json-rpc/spec.json
```diff
@@ -196,6 +196,101 @@
...
+					"required": true,
+					"schema": {
+						"$ref": "#/components/schemas/Integer"
+					}
+				},
+				{
+					"name": "lastBlock",
+					"description": "Last block number of the requested range.",
```
MicahZoltu, line 214: Consider renaming this to "newest" instead of "last".  Depending on your frame of reference, "last" may be closer to genesis than head, or it may be closer to head than genesis.

json-rpc/spec.json
```diff
@@ -196,6 +196,101 @@
...
+					"required": true,
+					"schema": {
+						"oneOf": [
+							{
+								"$ref": "#/components/schemas/BlockNumber"
+							},
+							{
+								"$ref": "#/components/schemas/BlockNumberTag"
+							}
```
MicahZoltu, line 223: We should create a new type for `BlockNumberOrTag` since it comes up a *lot*.

json-rpc/spec.json
```diff
@@ -196,6 +196,101 @@
...
+					"type": "object",
+					"required": [
+						"firstBlock",
+						"baseFee",
+						"gasUsedRatio"
+					],
+					"properties": {
+						"firstBlock": {
+							"title": "firstBlock",
+							"description": "First block number of the returned range.",
```
MicahZoltu, line 254: Consider "newest" here if you decide to change the above to "newest" so naming is consistent.

json-rpc/spec.json
```diff
@@ -196,6 +196,101 @@
...
+							"$ref": "#/components/schemas/BlockNumber"
+						},
+						"baseFee": {
+							"title": "baseFeeArray",
+							"description": "An array of block base fees (including the next block after the returned range).",
```
MicahZoltu, line 259: It seems odd to me that this would return *more* blocks than requested.  If the user wants +1, they can just ask for +1.

json-rpc/spec.json
```diff
@@ -196,6 +196,101 @@
...
+							"type": "array",
+							"items": {
+								"$ref": "#/components/schemas/Integer"
+							}
+						},
+						"gasUsedRatio": {
+							"title": "gasUsedArray",
+							"description": "An array of block gas used ratios.",
```
MicahZoltu, line 267: Is this a decimal number between 0 and 1?  Might be good to provide a bit more description here.

json-rpc/spec.json
zsfelfoldi: This info is derived from the requested blocks (the next base fee always follows from the last block). It makes sense to specify it like this because the pending base fee is basically the most important and it has nothing to do with the pending state that is not supported at all by some backends (so it either couldn't be queried or would be an ugly corner case with only the baseFee returned). On the other hand if pending state is supported and the pending block is queried then we can even predict even the after-pending block's base fee because it follows from the pending state and it is a good indicator of congestion in the mempool. There isn't even a way to specify the after-pending block as lastBlock. So I'd strongly like to keep it this way and specify that the function returns info derivable from the requested blocks (hence the extra baseFee value).

json-rpc/spec.json
zsfelfoldi: you mean oldest I guess :)

json-rpc/spec.json
zsfelfoldi: done

json-rpc/spec.json
zsfelfoldi: done

json-rpc/spec.json
zsfelfoldi: done

json-rpc/spec.json
zsfelfoldi: done

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+					],
+					"properties": {
+						"oldestBlock": {
+							"title": "oldestBlock",
+							"description": "Oldest (lowest number) block of the returned range.",
+							"$ref": "#/components/schemas/BlockNumber"
+						},
+						"baseFee": {
+							"title": "baseFeeArray",
+							"description": "An array of block base fees (including the next block after the newest of the returned range because this value can be derived from the newest block).",
```
RatanRSur, line 252: Is this null if the block has no base fee?

alcuadrado:
> @MicahZoltu @lightclient does this seem ready to merge to you?

I'm curious about this. What will be the criteria to accept modifications to this spec? Will accepting a new method imply that all the clients and developer networks (e.g. ganache, hardhat, remix) need to implement it?

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
 				"$ref": "#/components/contentDescriptors/GasPrice"
 			}
 		},
+		{
+			"name": "eth_feeHistory",
+			"summary": "Transaction fee history",
+			"description": "Returns transaction base fee and effective priority fee history for the requested/supported block range.",
```
lightclient, line 202: What would the behaviour be before EIP-1559 blocks? I think it would be good to mention that upfront in the description.

lightclient: Also, these should be listed "base fee per gas" and "effective priority fee per gas".

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+			"params": [
+				{
+					"name": "blockCount",
+					"description": "Number of blocks in the requested range (positive integer; less may be returned if requesting the entire specified range is not supported).",
```
lightclient, line 206:
```suggestion
					"description": "Number of blocks in the requested range (less may be returned if requesting the entire specified range is not supported).",
```
The schema should denote the type.

It's also not clear to me what "not supported" means?

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+					"required": true,
+					"schema": {
+						"$ref": "#/components/schemas/Integer"
+					}
+				},
+				{
+					"name": "newestBlock",
+					"description": "Newest (highest number) block of the requested range (the range between headBlock-4 and headBlock are guaranteed to be supported while the pending block and older history are optional to support).",
```
lightclient, line 214:
```suggestion
					"description": "Highest number block of the requested range. The range between headBlock-4 and headBlock are guaranteed to be supported while the pending block and older history are optional to support.",
```

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+					"required": true,
+					"schema": {
+						"$ref": "#/components/schemas/BlockNumberOrTag"
+					}
+				},
+				{
+					"name": "rewardPercentiles",
+					"description": "A list of percentile values (0..100 floating point) to sample from each block's miner rewards in ascending order, weighted by gas used.",
```
lightclient, line 222:
```suggestion
					"description": "A list of percentile values to sample from each block's miner rewards in ascending order, weighted by gas used.",
```
Please use the `schema` field to communicate the type information.

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+					"type": "object",
+					"required": [
+						"firstBlock",
+						"baseFee",
+						"gasUsedRatio"
+					],
+					"properties": {
+						"oldestBlock": {
+							"title": "oldestBlock",
+							"description": "Oldest (lowest number) block of the returned range.",
```
lightclient, line 247:
```suggestion
							"description": "Lowest number block of the returned range.",
```

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+			"result": {
+				"name": "feeHistoryResult",
+				"description": "Fee history for the returned block range (the intersection of requested and supported range).",
+				"schema": {
+					"title": "feeHistoryResults",
+					"description": "Fee history results.",
+					"type": "object",
+					"required": [
+						"firstBlock",
+						"baseFee",
```
lightclient, line 241:
```suggestion
						"baseFeePerGas",
```

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+						"type": "array",
+						"items": {
+							"type": "number"
+						}
+					}
+				}
+			],
+			"result": {
+				"name": "feeHistoryResult",
+				"description": "Fee history for the returned block range (the intersection of requested and supported range).",
```
lightclient, line 234:
```suggestion
				"description": "Fee history for the returned block range. This is the intersection of requested and supported range.",
```

lightclient: Fwiw, this idea of a "supported range" is a bit confusing to me. I think it would be better say "available block data" or something more descriptive. It would also to help to mention in the method's description.

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+						"gasUsedRatio"
+					],
+					"properties": {
+						"oldestBlock": {
+							"title": "oldestBlock",
+							"description": "Oldest (lowest number) block of the returned range.",
+							"$ref": "#/components/schemas/BlockNumber"
+						},
+						"baseFee": {
+							"title": "baseFeeArray",
```
lightclient, line 251:
```suggestion
							"title": "baseFeePerGasArray",
```

json-rpc/spec.json
lightclient, line 252:
```suggestion
							"description": "An array of block base fees per gas. This includes the next block after the newest of the returned range, because this value can be derived from the newest block.",
```

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+							"description": "An array of block base fees (including the next block after the newest of the returned range because this value can be derived from the newest block).",
+							"type": "array",
+							"items": {
+								"$ref": "#/components/schemas/Integer"
+							}
+						},
+						"gasUsedRatio": {
+							"title": "gasUsedArray",
+							"description": "An array of block gas used ratios (gasUsed/gasLimit; 0..1 floating point).",
```
lightclient, line 260:
```suggestion
							"description": "An array of block gas used ratios. This is calculated by dividing the gasUsed by the gasLimit.",
```

lightclient: Please use the schema to denote the type.

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+							"type": "array",
+							"items": {
+								"title": "gasUsedRatio",
+								"description": "The ratio of gas used and gas limit (0..1 floating point).",
```
lightclient, line 264:
```suggestion
								"description": "The ratio of gas used and gas limit.",
```

lightclient: Please use the schema to denote the type.

json-rpc/spec.json
```diff
@@ -196,6 +196,94 @@
...
+								"type": "number"
+							}
+						},
+						"reward": {
+							"title": "rewardArray",
+							"description": "A two-dimensional array of block miner rewards (effective priority fees) at the requested block percentiles.",
```
lightclient, line 270:
```suggestion
							"description": "A two-dimensional array of effective priority fees per gas at the requested block percentiles.",
```

json-rpc/spec.json
```diff
@@ -1135,6 +1223,17 @@
 					"pending"
 				]
 			},
+			"BlockNumberOrTag": {
```
lightclient, line 1226: Did you just pull this out to be more explicit?

lightclient (changes requested):
Thanks for making this PR! In general, it would be good to avoid discussing things in parentheses. It would be better to state things explicitly. I'd also like the actual type information to be encoded in the schema.

lightclient:
> > @MicahZoltu @lightclient does this seem ready to merge to you?
> 
> I'm curious about this. What will be the criteria to accept modifications to this spec? Will accepting a new method imply that all the clients and developer networks (e.g. ganache, hardhat, remix) need to implement it?

This is a great question. I _think_ the best way to go about it is have endpoints be discussed and proposed as PRs and then merge those changes into the spec repo as "pre-releases". Every month or so, we would need to have a synchronous call with the stakeholders (like the teams you've mentioned) and client devs to get a thumbs up on the release candidate, then we can release a new version of the spec. It may also be useful to target a date where all major tooling has adapted to support the latest `json-rpc` spec. If this is the desired path, I would like to see the spec broken out into its own repository. Curious to know what people think of this.

json-rpc/spec.json
MicahZoltu: Ah, all the more reason to change this wording!  I thought by "first" you meant "latest" (most recent, first block counting down from now, etc.).  😁

MicahZoltu:
I have a mild preference for having a branch that we merge PRs like this into that is named something like `spec-alpha` or `spec-beta`.  The reason for this is because in association with the call you mentioned, we can create a Pull Request of `spec-alpha` into `spec-beta` or `spec-beta` into `main` so that the changeset being discussed on that particular call is very clear and people can leave comments directly on the changeset under discussion.

We could also have a branch for `release-candidate` that we can merge into while we wait for client implementations, then once we are comfortable asserting "this is the spec" (and we have some implementations) we can merge `release-candidate` into `main`.  This would make the workflow something like `personal_fork -> alpha -> beta -> release-candidate -> main` (exact names TBD).

A workflow like this makes it very easy to find out what is the *current* specification (broadly speaking, implemented by one or more clients) as well as making it easy to see what is currently up for "official" proposal and easy to generate a diff or a PR for discussing current proposals.

timbeiko:
@alcuadrado good question! As @lightclient and @MicahZoltu have pointed out, this process is still new and we haven't figured out the kinks yet. I'm in favor of the general approach of having multiple branch and review periods. I'll try and draft up a proposal today/tomorrow. 

As for this specific PR, I'm inclined to merge it for a couple reasons:

1. We've already merged other London/1559-related endpoints (e.g. https://github.com/ethereum/eth1.0-specs/pull/206, https://github.com/ethereum/eth1.0-specs/pull/192) and have pointed people here as the canonical place to follow these changes;
2. We've discussed adding this API first on the [Gas API call](https://github.com/ethereum/pm/issues/328) a few weeks ago and confirmed it on AllCoreDevs last Friday (https://github.com/ethereum/pm/issues/334#issuecomment-859670312) 
3. Geth and Besu already have WIP PRs to support this, see https://github.com/ethereum/go-ethereum/pull/23033 & https://github.com/hyperledger/besu/pull/2432 

Ideally, tooling/infrastructure should support this, but having full 1559 support on Day 1 of the London fork isn't a strict requirement as legacy txns will still be allowed.

json-rpc/spec.json
```diff
@@ -196,6 +196,96 @@
...
+			"name": "eth_feeHistory",
+			"summary": "Transaction fee history",
+			"description": "Returns base fee and transaction effective priority fee history for the requested block range if available. The range between headBlock-4 and headBlock is guaranteed to be available while retrieving data from the pending block and older history are optional to support. For pre-EIP-1559 blocks the gas prices are returned as rewards and zeroes are returned for the base fee.",
+			"params": [
+				{
+					"name": "blockCount",
+					"description": "Number of blocks in the requested range. Between 1 and 1024 blocks can be requested in a single query. Less than requested may be returned if not all blocks are available.",
+					"required": true,
+					"schema": {
+						"$ref": "#/components/schemas/Integer"
```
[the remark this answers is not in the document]
lightclient: On second thought, I suppose the description is good enough for now. Eventually we'll should encode in the schema though.

lightclient (approved):
Looks good, I think just a couple missed `per gas`es and then good to go.

json-rpc/spec.json
```diff
@@ -196,6 +196,96 @@
 				"$ref": "#/components/contentDescriptors/GasPrice"
 			}
 		},
+		{
+			"name": "eth_feeHistory",
+			"summary": "Transaction fee history",
+			"description": "Returns base fee and transaction effective priority fee history for the requested block range if available. The range between headBlock-4 and headBlock is guaranteed to be available while retrieving data from the pending block and older history are optional to support. For pre-EIP-1559 blocks the gas prices are returned as rewards and zeroes are returned for the base fee.",
```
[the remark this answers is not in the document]
zsfelfoldi: done

zsfelfoldi:
I added the missing "per gas" where requested and squashed my commits. I think it's mergeable now unless there is a further issue :)

lightclient (approved):

fvictorio:
@zsfelfoldi is the gist of the original proposal correct? Shouldn't be [this one](https://gist.github.com/zsfelfoldi/473e29106d38525de6b4413e2ebcddf1) instead?

timbeiko:
+1. The HackMD link seems wrong.

PR #219, opened 2021-06-22
Pyspec framework

lightclient:
This is the initial framework for an eth1 pyspec.

Ethereum 1.0 lacks a comprehensive and accessible specification. This creates several pain points, but the most prevalent is that intricate core changes must spend a [significant amount of time](https://eips.ethereum.org/EIPS/eip-1559) describing the current state of Ethereum, before even defining the proposed changes.

The [Ethereum 2.0 spec](https://github.com/ethereum/eth2.0-specs) has been a great example of "spec driven development". Eventually, we'd like to merge the two specs into a uniform, all encompassing specification. Until then, we've taken the liberty to experiment with a couple slightly different approaches than the eth2 specs:

* **_generate documentation from code_** -- the eth2 spec is a collection of markdown files that can be transpiled into an executable python project. My understanding is that this was a natural evolution of the spec, which started as mark down and slowly became more explicit, until the point where if the code snippets were concatenated, they could actually run and generate tests. Since we're starting from scratch, we can make an unbiased decision. While we agree that a "narrative" spec is desirable, we are confident that [Sphinx](https://www.sphinx-doc.org/en/master/) will give us complete control over the look and flow of the spec. It will require more effort than markdown, but we believe it will drastically simplify other portions of the toolchain (e.g. type checking and linting), make the raw spec more readable and maintainable, and allow us to use more tools that already exist instead of writing our own.
* **_non-forking code_** -- the eth2 spec generally understands the concept of "forks" and handles them accordingly. We would like the eth1 spec to completely ignore the concept of forks and instead allow it to be "the latest version of ethereum". We believe this will lead to a simpler spec and simpler proposals for core changes. Forks are an unavoidable part of Ethereum though, so we'd like to support them via a release mechanism. For example, `1.x.x` would be Frontier, `2.x.x` would be Homestead, etc. We plan to write a handler program that runs specific releases up until a fork block and then shuttle the current state of the chain into the next release.

So, with that said, we are currently working on implementing the Frontier release of Ethereum. This is very much a work-in-progress still and everything is up for discussion. We feel the sooner this work is in the open, the sooner we can get feedback from the community and potentially find additional contributors.

As of this PR, we've successfully:
* generated the mainnet genesis state root
* passed a load of RLP and trie tests
* passed a simple `add` + `sstore` test

We'll generally try to avoid writing our own tests wherever possible and instead rely on the `ethereum/tests` format. We plan to support the `t8n` interface so that the spec can generate test cases. We'd also like to eventually sync mainnet, but that may require some work arounds.

We're excited to hear people's thoughts and feedback :)


timbeiko:
This is great work, huge props for taking it on :clap: 

One thought re: versioning: IMO it may make sense to keep all pre-merge versions of the spec `1.x` and use the merge as the start of the `2.0` version of the spec (so Frontier is 1.0, Homestead 1.1, etc.). This would align better with the "Eth 2.0" nomenclature. Even if we want to deprecate it in favor of execution (eth1) and consensus (eth2), it still seems coherent to have the merge be v2.0 of the execution layer. 

We can keep incrementing it long-term when we have significant changes to how the entire protocol works (e.g. state expiry) vs. "feature additions" in network upgrades such as Istanbul, Berlin, London, etc.

lightclient:
> One thought re: versioning: IMO it may make sense to keep all pre-merge versions of the spec 1.x and use the merge as the start of the 2.0 version of the spec (so Frontier is 1.0, Homestead 1.1, etc.)

The downside of this is that this would incompatible with SemVer, which requires minor (`1.x.0`) changes be backwards compatible and major (`x.0.0`) be reserved for backwards incompatible changes.

README.md
```diff
@@ -23,4 +23,72 @@ This repository contains various specification related to the Ethereum 1.0 chain
 | Frontier Thawing | 200000 | 09/07/2015 | | | [Blog](https://blog.ethereum.org/2015/08/04/the-thawing-frontier/) |
 | Frontier | 1 | 07/30/2015 | | | [Blog](https://blog.ethereum.org/2015/07/22/frontier-is-coming-what-to-expect-and-how-to-prepare/) |
 
+## Consensus Specification (work-in-progress)
 
+ * [Rendered specification](https://quilt.github.io/eth1.0-specs/)
```
pipermerriam, line 28: A small paragraph or sentence describing what this is seems appropriate here.

> The Consensus Specification is a python implementation of the .....  that prioritizes readability and simplicity.  It is accompanied by both narrative and API level documentation of the various components written in restructured text and rendered using Sphinx....

README.md
```diff
@@ -23,4 +23,72 @@ This repository contains various specification related to the Ethereum 1.0 chain
...
+```
+
+The path to the generated HTML will be printed to the console.
+
+### Development
+
+Running the tests necessary to merge into the repository requires:
+
+ * Python 3.7.x (not 3.8 or later), and
+ * [PyPy 7.3.x](https://www.pypy.org/).
```
pipermerriam, line 51: Supporting pypy is noble but I would suggest a level of pragmatism if you run into things that this makes complex.  It is probably *ok* for this spec to have a very specific version and implementation of python for which it is considered "correct".

README.md
```diff
@@ -23,4 +23,72 @@ This repository contains various specification related to the Ethereum 1.0 chain
...
+$ tox
+```
+
+The development tools can also be run outside of `tox`, and can automatically reformat the code:
+
+```bash
+$ pip install -e .[doc,lint,test]   # Installs eth1spec, and development tools.
+$ isort src                         # Organizes imports.
+$ black src                         # Formats code.
+$ flake8                            # Reports style/spelling/documentation errors.
```
pipermerriam, line 77: we have a convention in our python repositories of having a `MAKE` command that will do these things for you: [`make lint-roll`](https://github.com/ethereum/ddht/blob/142911d134ff839f3f79ff8fe9e45d3fe5a58cd0/Makefile#L29-L40)

README.md
```diff
@@ -23,4 +23,72 @@ This repository contains various specification related to the Ethereum 1.0 chain
...
+
+It is recommended to use a [virtual environment](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment) to keep your system Python installation clean.
+
+## Contribution Guidelines
+
+This specification aims to be:
+
+1. **Correct** - Describe the _intended_ behavior of the Ethereum blockchain, and any deviation from that is a bug.
+2. **Complete** - Capture the entirety of _consensus critical_ parts of Ethereum.
+3. **Usable** - Prioritize readability, clarity, and plain language over performance and brevity.
```
pipermerriam, line 90: Maybe "Accessible" instead of "Usable"?  Simple?  Low Barrier to Entry?

b2t.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh 
+
+# Translate BlockchainTests into t8n format.
+#
+# Note: t8n requires that hex values not have leading 0s, so you'll need to
+# trim the manually (for now) :(.
+
+set -euf
```
pipermerriam, line 8: *nitpick* 

expanding these flags to their longform is good self-documenting script code.

doc/_templates/autoapi/python/class.rst
```diff
@@ -0,0 +1,56 @@
...
+   {% if "show-inheritance-diagram" in autoapi_options and obj.bases != ["object"] %}
+   .. autoapi-inheritance-diagram:: {{ obj.obj["full_name"] }}
+      :parts: 1
+      {% if "private-members" in autoapi_options %}
+      :private-bases:
+      {% endif %}
+
+   {% endif %}
+   {% endif %}
+   {% if "inherited-members" in autoapi_options %}
```
pipermerriam, line 32: can this file be documented/commented in some meaningful way?  Was it copied and modified from something existing?  Maybe inlining comments with examples of what any of the complex secitons or non-standard parts are intended to render?  this may be of little value so feel free to disregard.

main.py
```diff
@@ -0,0 +1,52 @@
...
+        Address(bytes.fromhex(env["currentCoinbase"][2:])),
+        Uint(int(env["currentNumber"], 16)),
+        Uint(int(env["currentGasLimit"], 16)),
+        Uint(int(env["currentTimestamp"], 16)),
+        Uint(int(env["currentDifficulty"], 16)),
+        [],
+        [],
+    )
+
+    print(gas_used)
```
pipermerriam, line 46: We have had good success with using `logging` over `print` which has the nice trade off of a little bit more boilerplate code which yields the ability to have different granularity about what is displayed when the *thing* is run.

https://github.com/ethereum/ddht/blob/142911d134ff839f3f79ff8fe9e45d3fe5a58cd0/ddht/logging.py#L59-L77

src/eth1spec/base_types.py
```diff
@@ -0,0 +1,553 @@
...
+        """
+        bit_length = self.bit_length()
+        byte_length = (bit_length + 7) // 8
+        return self.to_bytes(byte_length, "big")
+
+
+U256.MAX_VALUE = U256(2 ** 256 - 1)
+
+
+Bytes = bytes
```
pipermerriam, line 549: Any reason this doesn't use [`typing.NewType`](https://docs.python.org/3/library/typing.html#newtype)?

src/eth1spec/eth_types.py
```diff
@@ -0,0 +1,127 @@
...
+class Header:
+    """
+    Header portion of a block on the chain.
+    """
+
+    parent: Hash32
+    ommers: Hash32
+    coinbase: Address
+    state_root: Root
+    transactions_root: Root
```
pipermerriam, line 80: `Root` type appears to allow arbitrary length values (since it is an alias for `Bytes`).  Should this and the others be `Hash32`?

src/eth1spec/evm/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    state: State
+
+
+@dataclass
+class Evm:
+    """The internal state of the virtual machine."""
+
+    pc: Uint
+    stack: List[U256]
+    memory: bytes
```
pipermerriam, line 48: Was any consideration given to using a `bytearray` here since it would give you a mutable data structure.  This would mirror the `stack` value which also looks to use a mutable data structure (but I haven't gotten far enough yet to know whether the mutable nature is actually used).

src/eth1spec/evm/instructions.py
```diff
@@ -0,0 +1,91 @@
...
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than `GAS_VERY_LOW`.
+    """
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+
+    x = pop(evm.stack)
+    y = pop(evm.stack)
+
+    val = x + y
```
pipermerriam, line 44: I'm assuming it is intentional that this leaves out the overflow wrap-around.

src/eth1spec/base_types.py
```diff
@@ -0,0 +1,553 @@
...
+        if right < 0 or right > self.MAX_VALUE:
+            raise ValueError()
+
+        return super(U256, self).__add__(right)
+
+    def wrapping_add(self, right: int) -> "U256":
+        result = self.unchecked_add(right)
+
+        if result == NotImplemented:
+            return NotImplemented
```
pipermerriam, line 316: Returning this exception type rather than raising it is confusing to me.  Any specific reason for this?  If it was simply raised at the points where things aren't implemented then you would get clean stacktraces showing you where these are coming from.  As it stands now it will be hard to determine the origin of these errors.

src/eth1spec/spec.py
```diff
@@ -0,0 +1,354 @@
...
+            )
+        )
+
+    state[coinbase].balance += BLOCK_REWARD
+
+    receipts_map = {
+        bytes(rlp.encode(Uint(k))): v for (k, v) in enumerate(receipts)
+    }
+    receipts_y = trie.map_keys(receipts_map, secured=False)
+    return (block_gas_limit - gas_available), trie.root(receipts_y), state
```
pipermerriam, line 178: *nitpick*

Slightly more verbose (and thus maybe more readable) by removing the return statement inlining.

```python
...
gas_remaining = block_gas_limit - gas_available
post_root = trie.root(receipts_y)

return (gas_remaining, post_root, state)
```

pipermerriam (commented):
3/4 of the way through and nervious that a commit is going to invalidate all the comments I've made so far.

timbeiko:
> > One thought re: versioning: IMO it may make sense to keep all pre-merge versions of the spec 1.x and use the merge as the start of the 2.0 version of the spec (so Frontier is 1.0, Homestead 1.1, etc.)
> 
> The downside of this is that this would incompatible with [SemVer](https://semver.org/), which requires minor (`1.x.0`) changes be backwards compatible and major (`x.0.0`) be reserved for backwards incompatible changes.

That's fair, OTOH most projects who start with SemVer end up not strictly adhering to it 😅 I personally prefer "Ethereum consistency" over "Standards consistency", but it's not a very strong opinion.

src/eth1spec/trie.py
```diff
@@ -0,0 +1,279 @@
...
+
+        # "secure" tries hash keys once before construction
+        key = crypto.keccak256(preimage) if secured else preimage
+
+        nibble_list = bytearray(2 * len(key))
+        for i in range(2 * len(key)):
+            if i % 2 == 0:  # even
+                nibble_list[i] = key[i // 2] // 16
+            else:
+                nibble_list[i] = key[i // 2] % 16
```
pipermerriam, line 116: Just typing it out so that we can see whether it looks more readable using shifting and masking.  (not sure if this is actual equivalent logic)

```python
if is_even:
    nibbles_list[i] = key[i // 2] << 4
else:
    nibbles_list[i] = key[i // 2] & 0x0f
```

Code comments that just say "the high four bits" and "the low four bits" might be the most effective readability improvement here.

pipermerriam (approved):
Solid beginning.  I'm :+1: on getting this merged once it's been open for a few days to give people time to find it and :eyes:

b2t.sh
lightclient: Unfortunately i) set doesn't have longform flags and ii) I didn't document _why_ I put those flags there so now I've forgotten! :(

src/eth1spec/evm/__init__.py
lightclient: You're right, this should be a `bytearray`.

src/eth1spec/evm/instructions.py
lightclient: Nope, this should've used `wrapping_add`. I didn't realize we were throwing a `ValueError` on overflow.

b2t.sh
pipermerriam: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

See the `-o`:

```
set -o errexit -o nounset -o functrace
```

Though I'm not really sure this is objectively better now that I wrote it out....

src/eth1spec/spec.py
```diff
@@ -0,0 +1,354 @@
...
+                tx.gas,
+                tx.to,
+                tx.value,
+                tx.data,
+            )
+        )
+    )
+
+
+def print_state(state: State) -> None:
```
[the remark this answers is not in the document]
lightclient: I agree with this, but I think for simplicity we'll keep it as-is until we get around to setting up a non-consensus module things like this.

README.md
SamWilsn: I think, as a mostly pure python project, it should be pretty easy to maintain compatibility, and I'd like to at least attempt to until we hit something that makes it difficult.

Ideally, we want to this project to sync mainnet, and PyPy might help speed that up. If that fails miserably, we can maybe fall back on Cython.

README.md
SamWilsn: So far I've managed to avoid using `make` for anything ;)

`tox` does the checks, but as long as the actual build doesn't rely on `make`, I guess I'd be okay with a script to simplify auto-formatting.

b2t.sh
SamWilsn: Might specifically be `bash` syntax too. Other shells might only support the short options.

doc/_templates/autoapi/python/class.rst
SamWilsn: All the templates are slightly modified versions from the upstream projects. I could at least link to the repository and commit I got them from though.

 - [ ] Add upstream links to templates.

src/eth1spec/base_types.py
SamWilsn: We haven't finished the type annotations yet :sweat_smile:

src/eth1spec/base_types.py
SamWilsn: `NotImplemented` isn't an exception (see https://docs.python.org/3/library/constants.html#NotImplemented).

This is just how these functions are supposed to work, apparently.

pipermerriam: I was reading these as: https://docs.python.org/3/library/exceptions.html#NotImplementedError

README.md
thomkeh: If the Python code is fully typed and basically only uses the standard library, then you could also compile it to C with [mypyc](https://github.com/mypyc/mypyc) for speed. It might not be as fast as pypy (though I haven't seen any benchmarks comparing them), but mypyc seems much more actively developed than pypy.

pipermerriam:
I would nominate that this get merged soon so that additional changes can be done in smaller units and directly targeting this repository rather than the quilt fork.

SamWilsn:
Agreed! None of us can actually do the merge though.

pipermerriam:
Still seeing a stream of commits coming in.  You wanna give me a poke when you're ready with something stable.

lightclient:
@pipermerriam we can keep it stable until Monday if someone wants to merge this. I'm guessing by then, we'll have some more PRs that we'll want to get merged.

timbeiko:
@lightclient happy to merge this now if you think it's ready to go!

lightclient:
It's ready! :)

PR #221, opened 2021-06-29
Omit `effectiveGasPrice` before EIP-1559

lightclient:
~~It [appears](https://github.com/ethereum/go-ethereum/blob/4fcc93d9229e99b14d3d1500f05ada6ba1f789bb/internal/ethapi/api.go#L1599-L1600) geth is currently omitting `effectiveGasPrice` from the RPC for pre-London blocks.~~ Sorry for the mistake! OE [prefers](https://github.com/openethereum/openethereum/pull/450) to do the same.

I weakly prefer keeping as-is so tooling doesn't have to potentially look in two places to see the `effectiveGasPrice`. This PR is to discuss the merits of that.

### DIFF
--- a/json-rpc/spec.json
+++ b/json-rpc/spec.json
@@ -1318,7 +1318,7 @@
 					},
 					"effectiveGasPrice": {
 						"title": "ReceiptEffectiveGasPrice",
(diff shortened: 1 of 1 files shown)


lightclient:
Rereading now, I think I misunderstood geth's handling there. It *does* appear that geth returns the `effectiveGasPrice` as depicted in the spec even for pre-london blocks -- my mistake! Not sure if we should still discuss? OE did seem to indicate that they prefer to not return `effectiveGasPrice` for pre-london blocks.

lightclient:
closed per ACD 117.

PR #222, opened 2021-06-30
Adds live preview tox command

lightclient:
Live previews can now be spun up with `tox -e doc-autobuild`.


lightclient:
wrong repo :(

PR #247, opened 2021-07-08
DUP and SWAP EVM opcodes implementation

Bhargavasomu:
### What was wrong?
DUP and SWAP EVM operations aren't implemented

### How was it fixed?
Following frontier yellowpaper

#### Cute Animal Picture


Bhargavasomu:
@SamWilsn @lightclient can you please review the code? Thanks a lot.

src/ethereum/frontier/vm/instructions/computation.py
```diff
@@ -0,0 +1,27 @@
+"""
+Ethereum Virtual Machine (EVM) Computation Instructions
```
SamWilsn, line 2: How would you feel about "control_flow" instead of "computation"?

src/ethereum/frontier/vm/instructions/computation.py
```diff
@@ -0,0 +1,27 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. contents:: Table of Contents
+    :backlinks: none
+    :local:
+
+Introduction
+------------
+
+Implementations of the EVM Arithmetic instructions.
```
SamWilsn, line 12:
```suggestion
Implementations of the EVM control flow instructions.
```

src/ethereum/frontier/vm/instructions/stack.py
```diff
@@ -0,0 +1,175 @@
...
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+
+    data_to_duplicate = evm.stack[len(evm.stack) - 1 - item_number]
+    push(evm.stack, data_to_duplicate)
+
+
+def swap_n(evm: Evm, item_number: int) -> None:
+    """
+    Swap the 1st and Nth items in the stack. All items are 0-indexed from the
+    top of the stack.
```
SamWilsn, line 83: Is `swap_n(evm, 0)` a no-op? Either way, it might be nice to show what it does in the docstring.

src/ethereum/frontier/vm/instructions/stack.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Raises
+    ------
+    OutOfGasError
+        If `evm.gas_left` is less than `GAS_VERY_LOW`.
+    """
+    assert item_number < len(evm.stack)
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+
+    data_to_duplicate = evm.stack[len(evm.stack) - 1 - item_number]
```
SamWilsn, line 76: We went with negative indexing in the previous PR, so we should use it here for consistency.

src/ethereum/utils.py
```diff
@@ -1,4 +1,14 @@
 """
+Utility Functions
+^^^^^^^^^^^^^^^^^^^^^^^
```
SamWilsn, line 3: Tiny formatting nit ;)

src/ethereum/frontier/vm/interpreter.py
```diff
@@ -86,7 +86,7 @@ def process_call(
         evm.env.state[evm.current].balance += evm.value
 
     while evm.running:
-        op = evm.code[evm.pc]
+        op = Ops(evm.code[evm.pc])
```
SamWilsn, line 89: What happens if the instruction isn't in the enum (ex. `0xfe`)?

SamWilsn (changes requested):

src/ethereum/frontier/vm/instructions/stack.py
Bhargavasomu: swap_n with n = 0 is a no-op but it isn't a valid opcode. So do we still mention about it?

SamWilsn: Yeah, I think either a note or an assert would be nice to have.

src/ethereum/frontier/vm/interpreter.py
Bhargavasomu: It raises a `ValueError: 254 is not a valid Ops`

src/ethereum/frontier/vm/instructions/stack.py
Bhargavasomu: I forgot to mention this in the previous PR. I didn't use the negative indexing, because the number was a `U256`, and doing something like `data[-1 - num: ]` was raising an error as negative values are not included in `U256`. Hence I kind of didn't do the negative indexing. But let me know if you want me to do negative indexing. I will make the change. Thanks.

src/ethereum/frontier/vm/instructions/stack.py
Bhargavasomu: Cool will mention in the docstring.

src/ethereum/frontier/vm/instructions/stack.py
SamWilsn: Ha. No it's totally fine to not use negative indexing in that case.

src/ethereum/frontier/vm/interpreter.py
SamWilsn: Should we make a new error type and raise that (for invalid instructions?)

Bhargavasomu: Sounds good. I will create a new error type for this.

tests/frontier/vm/vm_test_helpers.py
```diff
@@ -0,0 +1,123 @@
...
+        gas=test_data["gas"],
+        depth=test_data["depth"],
+        env=env,
+    )
+
+    assert gas_left == test_data["expected_gas_left"]
+    assert keccak256(rlp.encode(logs)) == test_data["expected_logs_hash"]
+    # We are checking only the storage here and not the whole state, as the
+    # balances in the testcases don't change even though some value is
+    # transferred along with code invokation. But our evm execution transfers
```
SamWilsn, line 39: `s/invokation/invocation`

SamWilsn (approved):
LGTM

tests/frontier/vm/vm_test_helpers.py
Bhargavasomu: Weird why flake8 didn't catch this. Anyhow fixed this, thanks.

PR #248, opened 2021-07-08
Comparison and bitwise EVM operations

Bhargavasomu:
### What was wrong?
Comparison and bitwise operations are not implemented.

### How was it fixed?
Following frontier yellowpaper.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/frontier/vm/instructions/bitwise.py
+++ b/src/ethereum/frontier/vm/instructions/bitwise.py
@@ -1,0 +1,147 @@
+"""
+Ethereum Virtual Machine (EVM) Bitwise Instructions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. contents:: Table of Contents
+    :backlinks: none
+    :local:
+
+Introduction
+------------
+
+Implementations of the EVM bitwise instructions.
+"""
+
+from ethereum.base_types import U256
+
+from .. import Evm
+from ..gas import GAS_VERY_LOW, subtract_gas
+from ..stack import pop, push
+
+
+def bitwise_and(evm: Evm) -> None:
+    """
+    Bitwise AND operation of the top 2 elements of the stack. Pushes the
+    result back on the stack.
+
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+
+    Raises
+    ------
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than `GAS_VERY_LOW`.
+    """
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+    x = pop(evm.stack)
+    y = pop(evm.stack)
+    push(evm.stack, x & y)
+
+
+def bitwise_or(evm: Evm) -> None:
+    """
+    Bitwise OR operation of the top 2 elements of the stack. Pushes the
+    result back on the stack.
+
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+
+    Raises
+    ------
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than `GAS_VERY_LOW`.
+    """
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+    x = pop(evm.stack)
+    y = pop(evm.stack)
+    push(evm.stack, x | y)
+
+
+def bitwise_xor(evm: Evm) -> None:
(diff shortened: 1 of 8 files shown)


Bhargavasomu:
The first 2 commits here are from #247 . This should be reviewed post merging the mentioned PR.

Bhargavasomu:
@SamWilsn @lightclient can you please review the code? Thanks!

src/ethereum/frontier/vm/instructions/bitwise.py
```diff
@@ -0,0 +1,148 @@
...
+    word = pop(evm.stack)
+
+    if byte_index >= 32:
+        result = U256(0)
+    else:
+        extra_bytes_to_left = byte_index
+        extra_bytes_to_right = 31 - byte_index
+        # Remove the extra bytes in the right
+        word = word >> (extra_bytes_to_right * 8)
+        if extra_bytes_to_left != 0:
```
lightclient, line 143: I think I would prefer to just shift right and always `&` out the upper bits, even if it's slightly less efficient.

src/ethereum/frontier/vm/instructions/comparison.py
```diff
@@ -0,0 +1,172 @@
...
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than `GAS_VERY_LOW`.
+    """
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+
+    x = pop(evm.stack)
+    y = pop(evm.stack)
+    result = U256(x < y)
```
lightclient, line 43: I wonder if something like this is a bit clearer: 

```python
result = U256(1) if x < y else U256(0)
```

lightclient: I guess not, rest of the comparisons do look nice. I just worry about things not being super explicit.

lightclient (approved):
Looks good. Only 1 main nit on the `get_byte` instruction.

src/ethereum/frontier/vm/instructions/comparison.py
Bhargavasomu: I think it's common to understand that a bool can be converted to `int` or `Uint`. Let me know your thoughts.

src/ethereum/frontier/vm/instructions/bitwise.py
Bhargavasomu: I initially did it that way, but later felt that this optimization is pretty readable and highly explicit. Can you please let me know why you prefer to always perform the `&` operation? Thanks.

Bhargavasomu:
@SamWilsn did you get a chance to review this code? Thanks.

SamWilsn:
> @SamWilsn did you get a chance to review this code? Thanks.

I didn't realize this PR was stacked. My comments are in https://github.com/ethereum/eth1.0-specs/pull/255

Bhargavasomu:
@SamWilsn @lightclient I believe I have addresed your comments. (Including Sam's comments in #255 ). Can you please re-review the code at your free time? Thanks a lot.

@lightclient I believe the `get_byte` comment you raised is still pending for discussion. @SamWilsn feel free to pitch your thoughts regarding this. Thanks.

src/ethereum/frontier/vm/instructions/bitwise.py
SamWilsn: First, I don't think we should be optimizing at all, especially without data. Second, I would be incredibly surprised if a branch is more efficient than just a shift and mask.

I did a little informal benchmarking, just because I was curious. I wouldn't consider these results as rigorous.

Here's my test harness:

```python
from ethereum.frontier.vm import Evm
from ethereum.frontier.vm.instructions.bitwise import get_byte
import timeit

def test():
    evm = Evm(pc=0, stack=[0xAABBCC, 30], memory=[], code=[], gas_left=10000000, current=[], caller=[], data=[], value=0, depth=0, env=None, refund_counter=0, running=True)
    get_byte(evm)
    return evm.stack[0]

timeit.timeit("test()", globals=locals())
```

With the branch, it took `2.8620279809983913` seconds. Without the branch, it took `2.770114379003644`.

SamWilsn (changes requested):
I think we can remove that branch, then we should be good to go.

src/ethereum/frontier/vm/instructions/bitwise.py
Bhargavasomu: Thanks for this benchmark Sam. Learned from it. I'll remove the branch.

Bhargavasomu:
@SamWilsn I have removed the branch. Can you please re-review?

SamWilsn (approved):

Issue #228, opened 2021-07-07
Implement memory expansion cost calculator


lightclient:
original issue: https://github.com/quilt/eth1.0-specs/issues/10

voith:
I am working on this. I have also implemented `MSTORE`, `MSTORE8` and `MLOAD`. However, I'm stuck on some failing test cases.

PR #250, opened 2021-07-08
Implemented opcodes MSTORE, MSTORE8, MLOAD, MSIZE

voith:
### What was wrong?

Related to Issue #228
- Memory-related operations of the EVM weren't implemented. 

### How was it fixed?
- Implemented memory expansion calculator
- Implemented opcodes `MSTORE`, `MSTORE8`, `MLOAD`, `MSIZE`

## TODO
- [x] Check if more tests can be added.  
- [x] See if some validations need to be added before writing and reading from memory.
- [x] Move the helpers used in the tests to a common file so that they can be re-used.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/frontier/vm/gas.py
+++ b/src/ethereum/frontier/vm/gas.py
@@ -71,3 +71,38 @@ def calculate_memory_gas_cost(size_in_bytes: Uint) -> U256:
     quadratic_cost = size_in_words ** 2 // 512
     total_gas_cost = linear_cost + quadratic_cost
     return U256(total_gas_cost)
+
+
+def calculate_gas_extend_memory(
+    memory: bytearray, start_position: Uint, size: U256
+) -> U256:
+    """
+    Calculates the gas amount to extend memory
+
+    Parameters
+    ----------
+    memory :
+        Memory contents of the EVM.
+    start_position :
+        Starting pointer to the memory.
+    size:
+        Amount of bytes by which the memory needs to be extended.
+
+    Returns
+    -------
+    to_be_paid : `ethereum.base_types.U256`
+        returns `0` if size=0 or if the
+        size after extending memory is less than the size before extending
+        else it returns the amount that needs to be paid for extendinng memory.
+    """
+    if size == 0:
+        return U256(0)
+    memory_size = Uint(len(memory))
+    before_size = ceil32(memory_size)
+    after_size = ceil32(start_position + size)
+    if after_size <= before_size:
+        return U256(0)
+    already_paid = calculate_memory_gas_cost(before_size)
+    total_cost = calculate_memory_gas_cost(after_size)
+    to_be_paid = total_cost - already_paid
+    return to_be_paid
--- a/src/ethereum/frontier/vm/gas.py
+++ b/src/ethereum/frontier/vm/gas.py
@@ -11,8 +11,8 @@ Introduction
 
 EVM gas constants and calculators.
 """
-
-from ethereum.base_types import U256
+from ethereum.base_types import U256, Uint
+from ethereum.utils import ceil32
 
 from .error import OutOfGasError
 
@@ -24,6 +24,8 @@ GAS_STORAGE_CLEAR_REFUND = U256(15000)
 GAS_LOW = U256(5)
 GAS_MID = U256(8)
 GAS_EXPONENTIATION = U256(10)
+GAS_BASE = U256(2)
+GAS_MEMORY = U256(3)
 
 
 def subtract_gas(gas_left: U256, amount: U256) -> U256:
@@ -46,3 +48,26 @@ def subtract_gas(gas_left: U256, amount: U256) -> U256:
         raise OutOfGasError
 
     return gas_left - amount
+
+
+def calculate_memory_gas_cost(size_in_bytes: Uint) -> U256:
+    """
+    Calculates the gas cost for allocating memory
+    to the smallest multiple of 32 bytes,
+    such that the allocated size is at least as big as the given size.
+
+    Parameters
+    ----------
+    size_in_bytes :
+        The size of the data in bytes.
+
+    Returns
+    -------
+    total_gas_cost : `ethereum.base_types.U256`
+        The gas cost for storing data in memory.
+    """
+    size_in_words = ceil32(size_in_bytes) // 32
+    linear_cost = size_in_words * GAS_MEMORY
+    quadratic_cost = size_in_words ** 2 // 512
+    total_gas_cost = linear_cost + quadratic_cost
+    return U256(total_gas_cost)
--- a/src/ethereum/frontier/vm/instructions/memory.py
+++ b/src/ethereum/frontier/vm/instructions/memory.py
@@ -14,12 +14,13 @@ Implementations of the EVM Memory instructions.
 from ethereum.base_types import U8_MAX_VALUE, U256, Uint
 
 from .. import Evm
-from ..gas import GAS_BASE, GAS_VERY_LOW, subtract_gas
-from ..memory import (
-    extend_memory_and_subtract_gas,
-    memory_read_bytes,
-    memory_write,
+from ..gas import (
+    GAS_BASE,
+    GAS_VERY_LOW,
+    calculate_gas_extend_memory,
+    subtract_gas,
 )
+from ..memory import extend_memory, memory_read_bytes, memory_write
 from ..stack import pop, push
 
 
@@ -45,8 +46,15 @@ def mstore(evm: Evm) -> None:
     # convert to Uint as start_position + size_to_extend can overflow.
     start_position = Uint(pop(evm.stack))
     value = pop(evm.stack).to_be_bytes32()
-    extend_memory_and_subtract_gas(evm, start_position, U256(32))
-    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+
+    total_gas_cost = (
+        calculate_gas_extend_memory(evm.memory, start_position, U256(32))
+        + GAS_VERY_LOW
+    )
+    evm.gas_left = subtract_gas(evm.gas_left, total_gas_cost)
+
+    # extend memory and subtract gas for allocating 32 bytes of memory
+    extend_memory(evm.memory, start_position, U256(32))
     memory_write(evm.memory, start_position, value)
 
 
@@ -74,8 +82,15 @@ def mstore8(evm: Evm) -> None:
     value = pop(evm.stack)
     # make sure that value doesn't exceed 1 byte
     normalized_bytes_value = (value & U8_MAX_VALUE).to_be_bytes()
-    extend_memory_and_subtract_gas(evm, start_position, U256(1))
-    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+
+    total_gas_cost = (
+        calculate_gas_extend_memory(evm.memory, start_position, U256(1))
+        + GAS_VERY_LOW
+    )
+    evm.gas_left = subtract_gas(evm.gas_left, total_gas_cost)
+
+    # extend memory and subtract gas for allocating 32 bytes of memory
+    extend_memory(evm.memory, start_position, U256(1))
     memory_write(evm.memory, start_position, normalized_bytes_value)
 
 
@@ -98,12 +113,18 @@ def mload(evm: Evm) -> None:
     """
     # convert to Uint as start_position + size_to_extend can overflow.
     start_position = Uint(pop(evm.stack))
+
+    total_gas_cost = (
+        calculate_gas_extend_memory(evm.memory, start_position, U256(32))
+        + GAS_VERY_LOW
+    )
+    evm.gas_left = subtract_gas(evm.gas_left, total_gas_cost)
+
     # extend memory and subtract gas for allocating 32 bytes of memory
-    extend_memory_and_subtract_gas(evm, start_position, U256(32))
+    extend_memory(evm.memory, start_position, U256(32))
     value = U256.from_be_bytes(
         memory_read_bytes(evm.memory, start_position, U256(32))
     )
-    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
     push(evm.stack, value)
 
 
--- a/src/ethereum/frontier/vm/instructions/memory.py
+++ b/src/ethereum/frontier/vm/instructions/memory.py
@@ -1,0 +1,126 @@
+"""
+Ethereum Virtual Machine (EVM) Memory Instructions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. contents:: Table of Contents
+    :backlinks: none
+    :local:
+
+Introduction
+------------
+
+Implementations of the EVM Memory instructions.
+"""
+from ethereum.base_types import U8_MAX_VALUE, U256, Uint
+
+from .. import Evm
+from ..gas import GAS_BASE, GAS_VERY_LOW, subtract_gas
+from ..memory import (
+    extend_memory_and_subtract_gas,
+    memory_read_bytes,
+    memory_write,
+)
+from ..stack import pop, push
+
+
+def mstore(evm: Evm) -> None:
+    """
+    Stores a word to memory.
+    This also expands the memory, if the memory is
+    insufficient to store the word.
+
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+
+    Raises
+    ------
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than
+        `3` + gas needed to extend memeory.
+    """
+    # convert to Uint as start_position + size_to_extend can overflow.
+    start_position = Uint(pop(evm.stack))
+    value = pop(evm.stack).to_be_bytes32()
+    extend_memory_and_subtract_gas(evm, start_position, U256(32))
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+    memory_write(evm.memory, start_position, value)
+
+
+def mstore8(evm: Evm) -> None:
+    """
+    Stores a byte to memory.
+    This also expands the memory, if the memory is
+    insufficient to store the word.
+
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+
+    Raises
+    ------
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than
+        `3` + gas needed to extend memory.
+    """
+    # convert to Uint as start_position + size_to_extend can overflow.
+    start_position = Uint(pop(evm.stack))
+    value = pop(evm.stack)
+    # make sure that value doesn't exceed 1 byte
+    normalized_bytes_value = (value & U8_MAX_VALUE).to_be_bytes()
+    extend_memory_and_subtract_gas(evm, start_position, U256(1))
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+    memory_write(evm.memory, start_position, normalized_bytes_value)
+
+
+def mload(evm: Evm) -> None:
+    """
+    Load word from memory.
+
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+
+    Raises
+    ------
+    StackUnderflowError
+        If `len(stack)` is less than `1`.
+    OutOfGasError
+        If `evm.gas_left` is less than
+        `3` + gas needed to extend memory.
+    """
+    # convert to Uint as start_position + size_to_extend can overflow.
+    start_position = Uint(pop(evm.stack))
+    # extend memory and subtract gas for allocating 32 bytes of memory
+    extend_memory_and_subtract_gas(evm, start_position, U256(32))
+    value = U256.from_be_bytes(
+        memory_read_bytes(evm.memory, start_position, U256(32))
+    )
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
+    push(evm.stack, value)
+
+
+def msize(evm: Evm) -> None:
+    """
+    Push the size of active memory in bytes onto the stack.
+
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+
+    Raises
+    ------
+    OutOfGasError
+        If `evm.gas_left` is less than `2`
+    """
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_BASE)
+    memory_size = U256(len(evm.memory))
+    push(evm.stack, memory_size)
--- a/src/ethereum/frontier/vm/memory.py
+++ b/src/ethereum/frontier/vm/memory.py
@@ -11,24 +11,20 @@ Introduction
 
 EVM memory operations.
 """
-from ethereum.frontier.vm import Evm
 from ethereum.utils import ceil32
 
 from ...base_types import U256, Bytes, Uint
-from .gas import calculate_memory_gas_cost, subtract_gas
 
 
-def extend_memory_and_subtract_gas(
-    evm: Evm, start_position: Uint, size: U256
-) -> None:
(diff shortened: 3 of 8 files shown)


whitelist.txt
```diff
@@ -216,3 +216,7 @@ staticcall
 revert
 invalid
 selfdestruct
+
+ceil32
+rjust
+memoryview
```
SamWilsn, line 222: I don't see where this is used.

SamWilsn (commented):
Looking good so far!

src/ethereum/frontier/vm/instructions.py
```diff
@@ -462,6 +463,99 @@ def push_n(evm: Evm, num_bytes: int) -> None:
...
+    ------
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than `GAS_LOW`.
+    """
+    start_position = pop(evm.stack)
+    value = pop(evm.stack).to_be_bytes()
+    padded_value = value.rjust(32, b"\x00")
+    normalized_value = padded_value[-32:]
```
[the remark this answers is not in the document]
voith: This is to make sure that the value received doesn't exceed 32 bytes. There's a test that fails if I don't normalize the value.

voith:
@SamWilsn @lightclient This is ready for review.

src/ethereum/frontier/vm/instructions/memory.py
```diff
@@ -0,0 +1,142 @@
...
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than
+        `GAS_LOW` + gas needed to extend memeory.
+    """
+    start_position = pop(evm.stack)
+    value = pop(evm.stack).to_be_bytes()
+    padded_value = value.rjust(32, b"\x00")
+    # normalize this to make sure that size doesn't exceed 32 bytes
+    normalized_value = padded_value[-32:]
```
SamWilsn, line 50: I'm still somewhat unsure how this is different from [`U256.to_be_bytes32`][0]. Could you maybe provide an example of how this snippet behaves differently? Should we change the implementation of [`U256.to_be_bytes32`][0]?

[0]: https://github.com/ethereum/eth1.0-specs/blob/3cac857d3556898adfd2ea988adfa6a794a0b367/src/ethereum/base_types.py#L548

src/ethereum/frontier/vm/gas.py
```diff
@@ -45,3 +48,56 @@ def subtract_gas(gas_left: U256, amount: U256) -> U256:
...
+        The gas cost for storing data in memory.
+    """
+    size_in_words = ceil32(size_in_bytes) // 32
+    linear_cost = size_in_words * GAS_MEMORY
+    quadratic_cost = size_in_words ** 2 // GAS_MEMORY_QUADRATIC_DENOMINATOR
+    total_gas_cost = linear_cost + quadratic_cost
+    return total_gas_cost
+
+
+def calculate_gas_for_extending_memory(
```
SamWilsn, line 75: What does this return if you try to `mstore` a bit of memory which already exists? Something like:

```python
from ethereum.frontier.vm.gas import calculate_gas_for_extending_memory 

memory = bytearray(b"\x00" * 256)
calculate_gas_for_extending_memory(memory, 0, 32)
```

I suspect it tries to return a negative number?

src/ethereum/utils.py
```diff
@@ -34,3 +35,24 @@ def get_sign(value: int) -> int:
...
+    """
+    Converts a 256-bit unsigned integer to the next closest multiple of 32.
+
+    value :
+        The value whose ceil32 is to be calculated.
+
+    Returns
+    -------
+    ceil32 : `ethereum.base_types.U256`
+        The same value if its a perfect multiple of 32
```
SamWilsn, line 50:
```suggestion
        The same value if it's a perfect multiple of 32
```

src/ethereum/utils.py
```diff
@@ -34,3 +35,24 @@ def get_sign(value: int) -> int:
...
+        else it gets the next closest multiple of 32.
```
SamWilsn, line 51: Maybe the one place a slightly more mathematical explanation is better :stuck_out_tongue_closed_eyes: 

```suggestion
        else it returns the smallest multiple of 32 that is greater than `value`.
```

SamWilsn (commented):
Besides a few small docstring suggestions, the only thing I need clarification on is `calculate_gas_for_extending_memory`.

src/ethereum/frontier/vm/gas.py
voith: @SamWilsn Nice catch Sam! I'll add a fix for this.

src/ethereum/utils.py
voith: Thanks for helping me fix the docs. Docs is the part I struggle with the most being a non-native English speaker.

src/ethereum/frontier/vm/instructions/memory.py
voith: I was confused earlier because `U256.to_be_bytes32` would raise `OverflowError` for `value > U256_MAX_VALUE`. But now that I think of it that case is not possible as the stack is limited to U256 numbers only.  
Thanks for catching this. I have applied your suggestion.

src/ethereum/frontier/vm/gas.py
voith: fixed.

voith:
Thanks for the excellent review @SamWilsn. I have applied all your suggestions. 
@lightclient It would be nice to get your review too.

src/ethereum/frontier/vm/gas.py
```diff
@@ -45,3 +48,56 @@ def subtract_gas(gas_left: U256, amount: U256) -> U256:
...
+        Size to extend memory.
+
+    Returns
+    -------
+    total_gas_cost : `ethereum.base_types.U256`
+        The gas cost for extending memory.
+    """
+    memory_size = U256(len(memory))
+    before_size = ceil32(memory_size)
+    after_size = ceil32(start_position + size)
```
SamWilsn, line 99: Not sure if you'd agree, but maybe an easier way of implementing this would be something like:

```suggestion
    after_size = ceil32(start_position + size)
    
    if after_size <= before_size:
        return U256(0)
```

src/ethereum/frontier/vm/instructions/memory.py
```diff
@@ -0,0 +1,141 @@
...
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than
+        `GAS_LOW` + gas needed to extend memeory.
+    """
+    start_position = pop(evm.stack)
+    value = pop(evm.stack).to_be_bytes32()
+    gas_extend_memory = calculate_gas_for_extending_memory(
+        evm.memory, start_position, U256(1)
```
SamWilsn, line 49: Not saying that anything is incorrect here, but `U256(1)` jumps out as odd compared to the following two `U256(32)`. If this is correct, maybe add a short comment describing what you're doing?

voith: maybe I should clarify in `calculate_gas_for_extending_memory` that the size parameter is the number of **bytes** by which we want to extend the memory. I can add comment here too.

SamWilsn: So here we check the price for extending by one byte, and if it costs anything to add one byte, we change the cost to extend by 32 bytes?

voith: Yeah, that's because the memory is a word addressable array. That logic is handled by the `ceil32` method. 
not sure If I understood your question completely and answered it.

src/ethereum/frontier/vm/gas.py
voith: Please take a look at 850f028 and see if you'd be okay with that. I decided to club all of the logic into a single method and called it `extend_memory_and_subtract_gas`. I can change it if you don't like it.

voith:
@SamWilsn I and @Bhargavasomu have mutually decided that we will merge this PR. I will him add as a co-author. Please let me know when you're happy with the code. I will squash the commits and add Somu as co-author before merging.

src/ethereum/frontier/vm/memory.py
```diff
@@ -0,0 +1,112 @@
...
+    ----------
+    memory :
+        Memory contents of the EVM.
+    start_position :
+        Starting pointer to the memory.
+    value :
+        Data to write to memory.
+    """
+    for idx, v in enumerate(value):
+        memory[start_position + idx] = v
```
Bhargavasomu, line 89:
```suggestion
    for idx, byte in enumerate(value):
        memory[start_position + idx] = byte
```

src/ethereum/frontier/vm/memory.py
```diff
@@ -0,0 +1,112 @@
...
+        Starting pointer to the memory.
+    size :
+        Amount by which the memory needs to be extended.
+    """
+    if size == 0:
+        return None
+    memory_size = U256(len(memory))
+    new_size = ceil32(start_position + size)
+    if new_size <= memory_size:
+        return None
```
Bhargavasomu, line 68: As far as the memory expansion goes in the yellow paper, I don't think the memory can ever shrink in a single transaction execution. Hence this condition can maybe be replaced with a comment mentioning that this condition isn't checked as it's assumed no entity will use it with a negative size.

\cc @SamWilsn

src/ethereum/frontier/vm/memory.py
```diff
@@ -0,0 +1,112 @@
...
+        Memory contents of the EVM.
+    start_position :
+        Starting pointer to the memory.
+    size :
+        Amount by which the memory needs to be extended.
+    """
+    if size == 0:
+        return None
+    memory_size = U256(len(memory))
+    new_size = ceil32(start_position + size)
```
Bhargavasomu, line 66:
```suggestion
    current_memory_size = Uint(len(memory))
    expanded_memory_size = ceil32(start_position + size)
```
I feel these variables add more readability. Theoretically, the memory size is not limited and hence we should be using `Uint` instead of `U256`. Same goes for the output of ceil32 in above.

src/ethereum/frontier/vm/memory.py
```diff
@@ -0,0 +1,112 @@
...
+    if new_size <= memory_size:
+        return None
+    size_to_extend = new_size - memory_size
+    memory.extend(b"\x00" * size_to_extend)
```
Bhargavasomu, line 70: image

Seems like doing the following is faster.
```
memory += b"\x00" * size_to_extend
```

src/ethereum/frontier/vm/memory.py
```diff
@@ -0,0 +1,112 @@
...
+    ----------
+    memory :
+        Memory contents of the EVM.
+    start_position :
+        Starting pointer to the memory.
+    size :
+        Amount by which the memory needs to be extended.
+    """
+    if size == 0:
+        return None
```
Bhargavasomu, line 64: I think both these functions can be merged into a function called `expand_memory` or `extend_memory`. Since it's already mentioned in the docstring, I don't think it should be part of the function name.

FWIW, the condition `if size == 0: return None` should be present in the `extend_memory_and_subtract_gas` function.

I also prefer using `current_memory_size` and `expanded_memory_size` instead of `before_size` and `after_size` respectively.

Given that I don't think the memory can shrink, the following condition seems unnecessary.
```
if total_cost < already_paid:
    return None
```

src/ethereum/frontier/vm/memory.py
```diff
@@ -0,0 +1,112 @@
...
+EVM memory operations.
+"""
+from ethereum.frontier.vm import Evm
+from ethereum.utils import ceil32
+
+from ...base_types import U256, Bytes
+from .gas import calculate_memory_gas_cost, subtract_gas
+
+
+def extend_memory_and_subtract_gas(
```
Bhargavasomu, line 21: I also feel that we can get away with only 2 parameters to this function being the following:

- evm: Evm
- memory_size: U256

But this goes against how the yellowpaper defines the expansion. @SamWilsn @lightclient let me know your thoughts on this. Thanks.

Bhargavasomu (changes requested):

src/ethereum/frontier/vm/memory.py
voith: for the sake of consistency, I prefer to keep the pop statements as is i.e inside the method that implements the Opcode.

src/ethereum/frontier/vm/memory.py
voith: I need to check this thoroughly. I'll compare the results with a working EVM.

tests/frontier/vm/test_memory_operations.py
```diff
@@ -0,0 +1,53 @@
...
+
+@pytest.mark.parametrize(
+    "test_file",
+    [
+        "mstore8_0.json",
+        "mstore8_1.json",
+    ],
+)
+def test_mstore8(test_file: str) -> None:
+    run_memory_vm_test(test_file)
```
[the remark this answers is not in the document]
voith: These tests raised an `OutOfGasError`. Upon investigation, I learned that the test is meant to raise an exception.
I also learned that fixtures that don't have a post-state are assumed to error out.

src/ethereum/frontier/vm/gas.py
```diff
@@ -45,3 +48,25 @@ def subtract_gas(gas_left: U256, amount: U256) -> U256:
         raise OutOfGasError
 
     return gas_left - amount
+
+
+def calculate_memory_gas_cost(size_in_bytes: U256) -> U256:
+    """
+    Calculates the gas cost for allocating memory up
+    and including the given address.
```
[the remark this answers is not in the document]
SamWilsn: LGTM

src/ethereum/frontier/vm/gas.py
```diff
@@ -11,18 +11,21 @@
...
 
-GAS_VERY_LOW = U256(3)
-GAS_STORAGE_SET = U256(20000)
-GAS_STORAGE_UPDATE = U256(5000)
-GAS_STORAGE_CLEAR_REFUND = U256(15000)
+GAS_BASE = U256(2)
+GAS_EXPONENTIATION = U256(10)
 GAS_LOW = U256(5)
+GAS_MEMORY = U256(3)
+GAS_MEMORY_QUADRATIC_DENOMINATOR = U256(512)
```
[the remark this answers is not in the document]
SamWilsn: I probably have an unpopular opinion here, but if a value is only used in one place and has no obvious meaningful name, putting it into a named constant just makes things harder to follow.

```python
quadratic_cost = size_in_words ** 2 // GAS_MEMORY_QUADRATIC_DENOMINATOR
```

vs.

```python
quadratic_cost = size_in_words ** 2 // 512
```

I think the second one is clearly easier to read.

That said, I know this is a contentious opinion, so I won't let it stand in the way of merging.

src/ethereum/frontier/vm/instructions/memory.py
```diff
@@ -0,0 +1,119 @@
...
+    evm :
+        The current EVM frame.
+
+    Raises
+    ------
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than
+        `GAS_LOW` + gas needed to extend memeory.
```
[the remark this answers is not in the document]
SamWilsn: `memeory` is my new favourite word.

src/ethereum/frontier/vm/memory.py
SamWilsn: > Given that I don't think the memory can shrink, the following condition seems unnecessary.

It's true that the memory can't shrink, but you can initially `mload(100)` and later `mload(0)`. For the second `mload`, if I'm not mistaken, that would make `total_cost - already_paid` negative?

tests/frontier/vm/test_memory_operations.py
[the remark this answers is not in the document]
Bhargavasomu: Yes, that's the expected end state of those test files.

src/ethereum/frontier/vm/memory.py
Bhargavasomu: Ahh, I now understand what's happening here and why those checks were enforced.

@voith I think you need to check if the memory which we are trying to access is already available or not. If it's already available the memory expansion shouldn't happen. Enforcing this check is much more clearer and readable than checking if the new gas is less than the already paid gas.

I leave it up to you whether you want to enforce this check in external code wherever this function is being called or inside this function itself. I weakly prefer having this check outside this function. You can refer to [this condition](https://github.com/ethereum/eth1.0-specs/pull/260/files#diff-5858d8f65876b560db7245474aad6c9acc689d7b6c887a008929cad952b32882R50-R53) in my old PR which actually implements this check.

src/ethereum/frontier/vm/memory.py
```diff
@@ -0,0 +1,112 @@
...
+        Starting pointer to the memory.
+    size :
+        Size of the data that needs to be read.
+
+    Returns
+    -------
+    data_bytes :
+        Data read from memory.
+    """
+    return bytes(memory[start_position : start_position + size])
```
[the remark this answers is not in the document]
voith: I just thought I'd match it with the signature of `U256.from_be_bytes`.

src/ethereum/frontier/vm/instructions/memory.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    start_position = pop(evm.stack)
+    value = pop(evm.stack).to_be_bytes()
+    padded_value = value.rjust(1, b"\x00")
+    # normalize this to make sure that size doesn't exceed 1 byte
+    normalized_value = padded_value[-1:]
+    extend_memory_and_subtract_gas(evm, start_position, U256(1))
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
```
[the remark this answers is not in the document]
voith: I prefer to keep it as is. Memory extension happens before writing to memory, so If there is an OutofGasException I'd like to follow the trace sequentially.

src/ethereum/frontier/vm/instructions/memory.py
```diff
@@ -0,0 +1,119 @@
...
+    StackUnderflowError
+        If `len(stack)` is less than `2`.
+    OutOfGasError
+        If `evm.gas_left` is less than
+        `GAS_LOW` + gas needed to extend memeory.
+    """
+    start_position = pop(evm.stack)
+    value = pop(evm.stack).to_be_bytes32()
+    extend_memory_and_subtract_gas(evm, start_position, U256(32))
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
```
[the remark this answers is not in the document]
voith: I prefer to keep it as is. Memory extension happens before writing to memory, so If there is an OutofGasException I'd like to follow the trace sequentially.

src/ethereum/frontier/vm/instructions/memory.py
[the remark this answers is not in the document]
Bhargavasomu: Well, AFAIU I don't think it's economical to continue with memory expansion, given that you don't have enough Gas to even start the operation. It's a general pattern we follow in the rest of the opcode instructions as well. We first try to subtract the gas required for kicking off the operation, and later go to the internals of the operation.

\cc @SamWilsn @lightclient

src/ethereum/frontier/vm/instructions/memory.py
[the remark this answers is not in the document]
Bhargavasomu: Well, AFAIU I don't think it's economical to continue with memory expansion, given that you don't have enough Gas to even start the operation. It's a general pattern we follow in the rest of the opcode instructions as well. We first try to subtract the gas required for kicking off the operation, and later go to the internals of the operation.

\cc @SamWilsn @lightclient

src/ethereum/frontier/vm/memory.py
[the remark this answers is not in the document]
Bhargavasomu: I think `bytes` and `bytearray` are kind of synonymous as far as mypy is concerned. Hence I think it's ok if you drop the `bytes` wrapper.

src/ethereum/frontier/vm/instructions/memory.py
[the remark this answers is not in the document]
voith: This is a spec, not a client. Readability triumphs over efficiency. For me reading the code in the order it occurs adds to better readability.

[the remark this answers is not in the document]
Bhargavasomu: I never mentioned anything about efficiency. For me it doesn't make sense to even reach the memory expansion bit if you don't even have sufficient gas to power the opcode.

My philosophy here is that you can't run a car unless it has enough power to at least start.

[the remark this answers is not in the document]
voith: My real reason for my approach is this section of the YP. 
Screenshot 2021-07-15 at 11 25 25 PM
for better readability, I decided to have the gas logic as consecutive statements so that its easy to see the actual cost function for memory operations and that is very closely related to how the YP defines it.  
Up until now all the opcodes implemented had `Cmemory = 0`. So it was easier to have the gas logic right at the start of the method.

> My philosophy here is that you can't run a car unless it has enough power to at least start.

I'm going to nitpick on some code that you wrote for the sake of this discussion(No malice intended just healthy debate):
https://github.com/ethereum/eth1.0-specs/blob/308b8ced06daeefca14d1b2c47ff0336e0d6cd5f/src/ethereum/frontier/vm/instructions/arithmetic.py#L305-L315
Going by your philosophy, why didn't you do?:
```python
evm.gas_left = subtract_gas(evm.gas_left, GAS_EXPONENTIATION)

if exponent != 0:
     exponent_bits = exponent.bit_length()
     exponent_bytes = (exponent_bits + 7) // 8
     evm.gas_left = subtract_gas(evm.gas_left, GAS_EXPONENTIATION * exponent_bytes)
```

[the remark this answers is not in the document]
Bhargavasomu: @voith I don't think you understood my philosophy correctly. As my philosophy states, in the code that you highlighted, I am predicting the gas and deducting it previous to the actual opcode internals (which is to do the exponentiation).

But in your case, you are first calculating the cost needed for memory expansion and also expanding the memory which is an internal of the opcode, which is something none of our opcode implementations currently do (AFAIR). Let me know if I'm still not clear on my point.

I kindly request @SamWilsn @lightclient to opine on this issue. It's unclear how to proceed regarding this.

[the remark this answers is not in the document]
SamWilsn: If I understand this debate correctly, the PR [as currently written](https://github.com/ethereum/eth1.0-specs/blob/ae4b619f3b5848cdbea0428beb01242283bd15b5/src/ethereum/frontier/vm/memory.py) does the following:

 1. Checks if there is enough gas for memory expansion.
 2. Expands the memory.
 3. Checks if there is enough gas for the non-variable portion of `mstore`.

The disagreement here is specifically about (2) coming before (3)?

The arguments for switching (2) and (3)—so that we do all the gas accounting up front and expand the memory after—are:
 * Sane order of operations (it's a waste to start the car if it'll run out of fuel as soon as you try to move); and
 * Consistency with how other opcodes are implemented.

That seems to make sense to me, but I might be misinterpreting @voith's position.

In the grand scheme of things, I don't see this making a huge difference either way, and even if it does, I think it's something we can fix in the future.

This discussion is blocking quite a bit of code, so whatever gets us unblocked the quickest is my personal preference (and _please_ don't consider me as an ultimate arbiter for resolving PRs; I don't want that responsibility!)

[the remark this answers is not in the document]
voith: @SamWilsn @Bhargavasomu Take a look at d5b38d6d647cc6302a7e00889f9c6a6c4bb3cdca
I split the gas calculation method into a separate method. It's a bit more verbose but it exactly follows the yellow paper. 
Let me know what you think.

src/ethereum/frontier/vm/memory.py
voith: I have used `before_size` and `after_size` which IMO are quite similar to this suggestion here. However, @SamWilsn I'll commit this if you feel strongly about it.

SamWilsn (approved):

PR #251, opened 2021-07-09
Require `gasPrice` field for both legacy and EIP-1559 transactions

lightclient:
### What was wrong?

As discussed in ACD 117, a lot of tooling relies on the `gasPrice` field of the `Transaction` object. Although this field doesn't make much sense with EIP-1559 transactions, for the sake of backwards compatibility we should continue supporting it. Please note that this field is **DEPRECATED** for EIP-1559 transactions, so developers should immediately begin transitioning to using the `effectiveGasPrice` field on the `Receipt` object.

### How was it fixed?

Return the `effectiveGasPrice` as the `gasPrice` field for EIP-1559 transactions. If the transaction has not yet been mined, clients should return the `maxFeeCapPerGas` value as `gasPrice`.

### DIFF
--- a/json-rpc/spec.json
+++ b/json-rpc/spec.json
@@ -1716,6 +1716,11 @@
 									"maxPriorityFeePerGas"
 								],
 								"properties": {
+									"gasPrice": {
+										"title": "transactionGasPrice",
+										"type": "string",
+										"description": "The effective gas price paid by the sender in Wei. For transactions not yet mined, this value should be set equal to the max fee cap per gas. This field is DEPRECATED, please transition to using effectiveGasPrice in the Receipt object going forward."
+									},
 									"maxPriorityFeePerGas": {
 										"title": "transactionMaxPriorityFeePerGas",
 										"description": "Maximum fee per gas the sender is willing to pay to miners in wei",


lightclient:
Please note, after the London hard fork we'll be coordinating JSON-RPC spec changes over here: https://github.com/ethereum/eth1.0-apis

lightclient:
@timbeiko before I had write access you consistently reviewed and merged the JSON-RPC changes. Let me know if you want to continue doing that or if you prefer I just maintain it.

timbeiko:
@lightclient happy for you to merge, but it'd be good to at least get a :+1: from another client dev IMO. Even if they can't actually "Review" on GH.

json-rpc/spec.json
```diff
@@ -1716,6 +1716,11 @@
 									"maxPriorityFeePerGas"
 								],
 								"properties": {
+									"gasPrice": {
+										"title": "transactionGasPrice",
+										"type": "string",
+										"description": "The effective gas price paid by the sender in Wei. For transactions not yet mined, this value should be set equal to the max fee cap per gas. This field is DEPRECATED, please transition to using effectiveGasPrice in the Receipt object going forward."
```
MicahZoltu, line 1722:
```suggestion
										"description": "The effective gas price paid by the sender in attoeth. For transactions not yet mined, this value should be set equal to the max fee cap per gas. This field is DEPRECATED, please transition to using effectiveGasPrice in the Receipt object going forward."
```
I figure it is worth a shot...

MicahZoltu, line 1722: We should include a clear end of life (EOL) policy here.  I suggest something like "This field will no longer be returned by clients sometime after 2022-01-01."

MicahZoltu:
I'm generally against this PR.  I suspect it is going to cause more problems than it solves.

By omitting the field entirely, tools will fail but they will fail fast and hard and cleanly and the fix becomes immediately obvious.  By including the filed but having it mean two different things depending on context, tools are likely to experience weird and hard to debug bugs that aren't easy to reproduce and whose behavior changes depending on when you look at things.  This means you'll have a long tail of bugs that will take a long time to get fixed rather than a short burst of tooling bugs that are rapidly fixed.

I think a preferable option for *temporary* backward compatibility would be to set `gasPrice` to the transaction's `maxFeePerGas`.  This is a *reasonable* value that has a well defined meaning and it doesn't change based on the lifecycle of the transaction.  If we go this route, we should still deprecate and EOL this field with a clear EOL date, but at least any bugs that come up can much more easily be fixed.

json-rpc/spec.json
lightclient: haha

alcuadrado:
Any updates on this? 

We released our (opt-in) London support in Hardhat yesterday, which includes `gasPrice` for EIP-1559 txs, as described here. We'd be fine with removing it if this PR is not accepted. We'd rather do that before the Mainnet activation, as we'll make London the default hardfork in Hardhat at the same time, and removing it later would be a larger breaking change.

lightclient:
What was the decision on ACD @timbeiko? Are we waiting for the infra call this week?

timbeiko:
@lightclient we agreed to keep the field in 1559-style transactions as well, have it return `maxFeePerGas` before being mined and the `effectiveGasPrice` after. We will deprecate this field right after the upgrade following London.

json-rpc/spec.json
danfinlay: I'd change that to _may_ no longer be returned. And in practice I would probably recommend returning it indefinitely, since this will be a breaking change.

MicahZoltu:
@timbeiko That wasn't my takeaway from the meeting... I thought the agreement was to wait and talk to dapp developers in the upcoming dapp integration call and then make a decision after that?  Some discussions in the Ethereum R&D Discord seem to imply that at least some dapp developers do want something that won't break everything instantly (so they want a `gasPrice` field) but they are concerned with the current Geth solution of including a value that changes when a block is mined and a reorg happens.

timbeiko:
@MicahZoltu now that you mention it, we did saw we'd bring this up then! My bad. It's on the agenda for this Friday.

json-rpc/spec.json
shanejonas: this could be refactored to change the `oneOf` wrapping it to be an `allOf`

lightclient:
Per the [EIP-1559 coordination call #11](https://youtu.be/r37ShBRdVDg?t=1802), tooling devs do prefer the gas price to continue being an element on the transaction object. They weren't against deprecation later on, but no clear path to that was agreed upon on the call. 

Since this has been outstanding for about a month and there has been no unresolved push back, I will merge this.

PR #253, opened 2021-07-09
[WIP] Implement an incrementally updating merkle trie

ultratwo:
**This is a very rough draft and is not ready for code review other than high level design comments**

This draft implements a incrementally updating merkle trie. This is necessary if `eth1.0-specs` is to have the ability to do basic state operations in a reasonable time.

### Use of non-dataclass class

The merkle trie is implemented as a `StateTrie` class. This violates the general principle that non-dataclass classes aren't used. I think this is justified because it has to hide functionality that will be implemented in different ways and classes are the best way to achieve that. I don't expect it to be a complicated class, it's interface is just a dictionary and a `root()` method.

### Flat state

The implementation uses the flat state model popularised by Erigon (formerly turbo-geth). The trie is implemented as key-value dictionary representing it's contents and the internal nodes are simply a cache that makes recalculating the root efficient after a small number of nodes have changed. When the `root` method is called we walk the trie depth first left-to-right, updating all the nodes that cover keys that have changed.

I used the flat state model because it if fundamentally more efficient, but also simpler. We will never have to implement garbage collection based pruning. If we need to access historical states (e.g. for block reorgs) we just need a backend that supports copy-on write checkpoints.

## Further work before merging

I plan to move the incremental update code into a separate file, probably `ethereum/frontier/optimized/trie`. I will then refactor the current trie to use my `LeafNode`, `ExtensionNode` and `BranchNode` data classes and to expose the same `StateTrie` interface. I also need to add tests to exercise all the incremental update codepaths and write all the docstrings.

### DIFF
--- a/src/ethereum/frontier/genesis.py
+++ b/src/ethereum/frontier/genesis.py
@@ -18,15 +18,15 @@ from dataclasses import dataclass
 from typing import Dict, cast
 
 from ethereum.base_types import U256, Bytes, Bytes8, Uint, slotted_freezable
-from ethereum.frontier.eth_types import Address
-
-from .utils.hexadecimal import (
-    hex_to_address,
+from ethereum.utils.hexadecimal import (
     hex_to_bytes,
     hex_to_bytes8,
     hex_to_u256,
     hex_to_uint,
 )
+
+from .eth_types import Address
+from .utils.hexadecimal import hex_to_address
 
 
 @slotted_freezable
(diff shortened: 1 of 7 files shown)


SamWilsn:
Fair warning, I haven't dived deeply into this yet.

First, I think this'll really come in handy once we get the actual "client" code setup, so thank you so much for the contribution! It might be a while before we actually get to that point; hopefully that won't be an issue.

> This violates the general principle that non-dataclass classes aren't used.

I think ideally I'd like to be able to monkey patch in the faster trie implementation, as as long as that's true, I don't mind using a more traditional Pythonic style.

ultratwo:
I've closed this due to it hogging the `master` branch of my fork.

PR #257, opened 2021-07-12
Format protocol dates using dd/mm/yyyy

axic:
### What was wrong?

The dates are following the slightly unusual (to anyone outside of the USA) date format of mm/dd/yyyy.

### How was it fixed?

Changed that to dd/mm/yyyy, but perhaps the yyyy-mm-dd ISO format would be more appropriate?

(Please double check the actual dates)


timbeiko:
Closing because https://github.com/ethereum/eth1.0-specs/pull/258 was merged instead.

Issue #249, opened 2021-07-08
Implement the Storage and Flow operations


Bhargavasomu:
- [x] SLOAD
- [x] Integrate test cases for SSTORE and Refund counter
- [x] JUMP
- [x] JUMPI
- [x] PC
- [x] MSIZE
- [x] GAS
- [x] JUMPDEST

Bhargavasomu:
I'm working on this. @lightclient please assign this to me as I don't have permissions to assign to myself.

Bhargavasomu:
Closing this as it has been implemented.

PR #273, opened 2021-07-20
Implement control flow operations

Bhargavasomu:
### What was wrong?
Few control flow operations weren't implemented.
- [x] POP
- [x] JUMP
- [x] JUMPI
- [x] PC
- [x] GAS
- [x] JUMPDEST

Related to Issue #249 

### How was it fixed?
Following frontier yellow paper.

#### Cute Animal Picture


Bhargavasomu:
@SamWilsn @lightclient can you please review this code whenever free? Thanks a lot!

src/ethereum/frontier/vm/runtime.py
```diff
@@ -0,0 +1,60 @@
...
+    Returns
+    -------
+    valid_jump_destinations: `Set[Uint]`
+        The set of valid jump destinations in the code.
+    """
+    valid_jump_destinations = set()
+    pc = Uint(0)
+
+    while pc < len(code):
+        current_opcode = Ops(code[pc])
```
SamWilsn, line 47: I think we need to handle `ValueError` and treat it as a a non-jumpdest non-push opcode?

SamWilsn (changes requested):

src/ethereum/frontier/vm/interpreter.py
```diff
@@ -93,7 +102,9 @@ def process_call(
             raise InvalidOpcode(evm.code[evm.pc])
 
         op_implementation[op](evm)
-        evm.pc += 1
+
+        if op not in PC_CHANGING_OPS:
+            evm.pc += 1
```
SamWilsn, line 107: I don't think this kind of "do something _except_ in certain cases" logic is the best for the specification.

I know it would be more tedious, but what would you think of lowering the `evm.pc += 1` into each opcode's implementation function? I think it would make it immediately obvious how each opcode behaves, at the risk of repeating ourselves.

src/ethereum/frontier/vm/instructions/control_flow.py
```diff
@@ -25,3 +36,136 @@ def stop(evm: Evm) -> None:
...
+            * The jump destination should have the `JUMPDEST` opcode (0x5B).
+            * The jump destination shouldn't be part of the data corresponding
+            to `PUSH-N` opcodes.
+    StackUnderflowError
+        If `len(stack)` is less than `1`.
+    OutOfGasError
+        If `evm.gas_left` is less than `8`.
+    """
+    evm.gas_left = subtract_gas(evm.gas_left, GAS_MID)
+    jump_dest = pop(evm.stack)
```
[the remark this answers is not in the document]
Bhargavasomu: Given the `jump_dest` here would always be a `U256` which is a subset of `Uint`. That said, we can check if a U256 is present in a set of Uints for the same reason. So I felt that this conversion is unnecessary. But let me know if converting to Uint actually adds to the readability. Thanks.

src/ethereum/frontier/vm/runtime.py
Bhargavasomu: I raised `InvalidOpcode` here instead of `InvalidJumpDestError` as the first error seemed apt here. Let me know if you want it the other way around. Thanks.

src/ethereum/frontier/vm/__init__.py
```diff
@@ -42,6 +42,16 @@ class Environment:
     state: State
 
 
+@dataclass
+class RuntimeContext:
```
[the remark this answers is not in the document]
Bhargavasomu: I just felt that going forward there might be more stuff coming in `RuntimeContext`. That's why I created a separate class. Let me know if you prefer doing that separation when needed. Thanks.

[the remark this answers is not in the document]
SamWilsn: Yeah, I think I'd prefer that for now.

src/ethereum/frontier/vm/instructions/control_flow.py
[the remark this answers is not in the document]
SamWilsn: As long as the set works as expected with `U256` instead of `Uint`, I'm happy to leave as is. I was just concerned since `U256` isn't a subclass of `Uint`.

src/ethereum/frontier/vm/runtime.py
SamWilsn: We can't raise here because the opcode wasn't actually executed. For example:

```
push1 0x04
jump
invalid
jumpdest
...
```

The invalid opcode is never executed, and doesn't mess up the jumpdest analysis.

src/ethereum/frontier/vm/instructions/control_flow.py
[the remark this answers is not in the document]
Bhargavasomu:
```
In [3]: a1 = {Uint(2), Uint(3), Uint(5), Uint(7), Uint(11), Uint(13)}

In [4]: Uint(5) in a1
Out[4]: True

In [5]: U256(5) in a1
Out[5]: True

In [6]: Uint(9) in a1
Out[6]: False

In [7]: U256(9) in a1
Out[7]: False

```

src/ethereum/frontier/vm/runtime.py
Bhargavasomu: Ok, I'll just skip invalid opcodes while performing jumpdest analysis, without raising any error.

src/ethereum/frontier/vm/interpreter.py
Bhargavasomu: While development, I felt that an argument can come in support either ways and hence went ahead with this as it seemed easy to revert.

Moving evm.pc increment to each opcodes kind of makes the workflow a bit obscure and not so straight forward. But moving it actually makes the design more streamlined. Let me know if the obscurity is worth it. Thanks.

SamWilsn: In the interest of avoiding another round trip, I'll merge as is, but this'll be a candidate for changing later on.

SamWilsn (approved):

src/ethereum/frontier/vm/interpreter.py
Bhargavasomu: Logged #278 for this.

PR #275, opened 2021-07-21
Make types immutable

ultratwo:
This pull request makes types immutable if they are stored in the trie or represent long standing unchanging things.

I am aware that `eth_types.py` is not a great home for `increase_balance()` and `decrease_balance()`. I find a better one for them when I do the state refactor I am currently working on.

I also made `Account.balance` a `U256`.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/base_types.py
+++ b/src/ethereum/base_types.py
@@ -16,7 +16,8 @@ Integer and array types which are used by—but not unique to—Ethereum.
 
 from __future__ import annotations
 
-from typing import Optional, Tuple, Type
+from dataclasses import replace
+from typing import Any, Callable, Optional, Tuple, Type, TypeVar
 
 U255_MAX_VALUE = (2 ** 255) - 1
 U255_CEIL_VALUE = 2 ** 255
@@ -636,4 +637,69 @@ Bytes8 = Bytes
 Bytes20 = Bytes
 Bytes32 = Bytes
 Bytes64 = Bytes
-BytesMutable256 = bytearray
+Bytes256 = Bytes
+
+
+def _setattr_function(self: Any, attr: str, value: Any) -> None:
+    if self._frozen:
+        raise Exception("Mutating frozen dataclasses is not allowed.")
+    else:
+        object.__setattr__(self, attr, value)
+
+
+def _delattr_function(self: Any, attr: str) -> None:
+    if self._frozen:
+        raise Exception("Mutating frozen dataclasses is not allowed.")
+    else:
+        object.__delattr__(self, attr)
+
+
+def _make_init_function(f: Callable) -> Callable:
+    def init_function(self: Any, *args: Any, **kwargs: Any) -> None:
+        will_be_frozen = kwargs.pop("_frozen", True)
+        object.__setattr__(self, "_frozen", False)
+        f(self, *args, **kwargs)
+        self._frozen = will_be_frozen
+
+    return init_function
+
+
+def slotted_freezable(cls: Any) -> Any:
+    """
+    Monkey patches a dataclass so it can be frozen by setting `_frozen` to
+    `True` and uses `__slots__` for efficiency.
+
+    Instances will be created frozen by default unless you pass `_frozen=False`
+    to `__init__`.
+    """
+    cls.__slots__ = ("_frozen",) + tuple(cls.__annotations__)
+    cls.__init__ = _make_init_function(cls.__init__)
+    cls.__setattr__ = _setattr_function
+    cls.__delattr__ = _delattr_function
+    return type(cls)(cls.__name__, cls.__bases__, dict(cls.__dict__))
+
+
+T = TypeVar("T")
+
+
+def modify(obj: T, f: Callable[[T], None]) -> T:
+    """
+    Create a mutable copy of `obj` (which must be `@slotted_freezable`) and
+    apply `f` to the copy before freezing it.
+
+    Parameters
+    ----------
+    obj : `T`
+        Object to copy.
+    f : `Callable[[T], None]`
+        Function to apply to `obj`.
+
+    Returns
+    -------
+    new_obj : `T`
+        Compact byte array.
+    """
+    new_obj = replace(obj, _frozen=False)
+    f(new_obj)
+    new_obj._frozen = True  # type: ignore
+    return new_obj
--- a/src/ethereum/frontier/eth_types.py
+++ b/src/ethereum/frontier/eth_types.py
@@ -13,7 +13,7 @@ Types re-used throughout the specification, which are specific to Ethereum.
 """
 
 from dataclasses import dataclass
-from typing import Dict, List, Optional
+from typing import Callable, Dict, Optional, Tuple
 
 from ..base_types import (
     U256,
@@ -21,8 +21,10 @@ from ..base_types import (
     Bytes8,
     Bytes20,
     Bytes32,
-    BytesMutable256,
+    Bytes256,
     Uint,
+    modify,
+    slotted_freezable,
 )
 from ..crypto import Hash32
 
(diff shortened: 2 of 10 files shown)


src/ethereum/frontier/eth_types.py
```diff
@@ -120,7 +120,24 @@ class Receipt:
...
+    )
+
+
+def reduce_balance(state: State, address: Address, amount: U256) -> None:
+    """Reduce the balance of an account in the State."""
+    account = state[address]
+    assert account.balance >= amount
+    state[address] = Account(
+        account.nonce, account.balance - amount, account.code, account.storage
+    )
```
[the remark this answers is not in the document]
SamWilsn: It'd be even better to include the actual state update in the transformation as well, so maybe this is better suited as a method on `State`, and we ignore my second suggestion?

Something like:

```python
@dataclass                               # Don't freeze this class.
class MutableAccount(Account):           # Might need to reverse the inheritance order, I'm not sure.
    pass

def state_update(state: State, address: Address, transform: Callable[[MutableAccount], None]):
    # ...
```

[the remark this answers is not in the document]
ultratwo: There are basically two ways to do this

## Freeze/thaw

There is a function `thaw()` that takes a immutable object and makes a mutable copy and a function `freeze()` which makes an object immutable (without copying). You can then write the following:

```python
account = thaw(state[address])
account.balance += BLOCK_REWARD
account.freeze()
state[address] = account
```
## Replace/update

The state object comes with two functions, `replace()` and `update()`. `replace()` works like `dataclasses.replace()`. `update()` takes mutator functions rather than new values. For example:

```python
state.replace(address, balance=state[address].balance + BLOCK_REWARD)
```

```python
state.update(address, balance=lambda x: x + BLOCK_REWARD)
```

## Thoughts

I prefer the replace/update method to freeze/thaw. It's neater and avoids the possibility of the following bug:

```python
account = thaw(state[address])
account.balance += BLOCK_REWARD

do_something_to_account(address)

account.freeze()
state[address] = account
```
The effects of `do_something_to_account()` get clobbered.

ultratwo:
I've added a `move_funds()` state modifier which is actually useful and switched to `dataclasses.replace()` where it doesn't apply.

ultratwo:
I've made the `move_funds()` to `move_ether()` change suggested by @SamWilsn and rebased up to HEAD. Due to the amount of rebase effort required I'd like to suggest we merge this even if we haven't reached agreement on how changing mutable objects should work in the long term.

src/ethereum/frontier/eth_types.py
```diff
@@ -120,7 +120,23 @@ class Receipt:
...
+    amount: U256,
+) -> None:
+    """Move funds between accounts."""
+    sender = state[sender_address]
+    assert sender.balance >= amount
+    state[sender_address] = replace(sender, balance=sender.balance - amount)
+    recipient = state[recipient_address]
+    state[recipient_address] = replace(
+        recipient, balance=recipient.balance + amount
+    )
```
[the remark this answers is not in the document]
ultratwo: I personally prefer my suggestion from earlier which was:

```python
state.update(address, balance=lambda x: x + BLOCK_REWARD, nonce=lambda x: x + 1)
```

vs your

```python
def make_changes(account: MutableAccount):
    account.nonce += 1
    account.balance += BLOCK_REWARD

state.update(address, make_changes)
```

Your version is more work to implement because it requires rewriting `dataclasses.dataclass` to produce a immutable and mutable version of each class.

src/ethereum/frontier/eth_types.py
```diff
@@ -128,7 +120,23 @@ class Receipt:
...
 State = Dict[Address, Account]
+
+
+def move_ether(
+    state: State,
+    sender_address: Address,
+    recipient_address: Address,
+    amount: U256,
+) -> None:
+    """Move funds between accounts."""
```
lightclient, line 135:
```suggestion
    """
    Move funds between accounts.
    """
```

src/ethereum/frontier/eth_types.py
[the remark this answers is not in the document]
lightclient: I find @SamWilsn's approach more readable.

lightclient (changes requested):
Overall, looks really good.

ultratwo:
I have implemented a `@slotted_freezable` decorator that takes a dataclass and makes it:
1. Use `__slots__`, which massively reduces instance size.
2. Not allow attribute changes if the `_frozen` attribute is set to `True`.

src/ethereum/base_types.py
```diff
@@ -636,4 +637,69 @@ def to_signed(self) -> int:
...
+    def init_function(self: Any, *args: Any, **kwargs: Any) -> None:
+        will_be_frozen = kwargs.pop("_frozen", True)
+        object.__setattr__(self, "_frozen", False)
+        f(self, *args, **kwargs)
+        self._frozen = will_be_frozen
+
+    return init_function
+
+
+def slotted_freezable(cls: Any) -> Any:
```
SamWilsn, line 667: I assume the mypy type checks still work with decorated classes?

SamWilsn (approved):
I wouldn't mind a few test cases for `@slotted_frozen`, but they can come in a separate PR.

Just something simple where you decorate a class, and try and mutate/update it.

src/ethereum/base_types.py
ultratwo: mypy doesn't understand `@slotted_freezable` and treats it as a no-op. The only problem with this is mypy will incorrectly complain if you try to access the `_frozen` element.

Otherwise everything works fine.

src/ethereum/frontier/eth_types.py
```diff
@@ -128,7 +136,37 @@ class Receipt:
...
 State = Dict[Address, Account]
+
+
+def modify_state(
+    state: State, address: Address, f: Callable[[Account], None]
+) -> None:
+    """
+    Modify an `Account` in the `State`.
+    """
+    state[address] = modify(state[address], f)
```
SamWilsn, line 151: Can/should we implement initializing and removing empty accounts?

src/ethereum/base_types.py
```diff
@@ -636,4 +637,69 @@ def to_signed(self) -> int:
...
+    """
+    Monkey patches a dataclass so it can be frozen by setting `_frozen` to
+    `True` and uses `__slots__` for efficiency.
+
+    Instances will be created frozen by default unless you pass `frozen=True`
```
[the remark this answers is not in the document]
ultratwo: Well spotted.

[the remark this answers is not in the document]
ultratwo: Well spotted.

src/ethereum/base_types.py
```diff
@@ -636,4 +637,69 @@ def to_signed(self) -> int:
...
+
+    return init_function
+
+
+def slotted_freezable(cls: Any) -> Any:
+    """
+    Monkey patches a dataclass so it can be frozen by setting `_frozen` to
+    `True` and uses `__slots__` for efficiency.
+
+    Instances will be created frozen by default unless you pass `frozen=False`
```
SamWilsn, line 672: It [looks](https://github.com/ethereum/eth1.0-specs/pull/275/files#diff-b8a22b1b8cce59cc5b6d01201a7f2d75394c510f127101f09a93e69a3d679a87R659) like the argument is named `_frozen`? Not 100% sure though.

src/ethereum/frontier/eth_types.py
ultratwo: I'd like to deal with both those problems in my upcoming state refactor rather than implementing a temporary fix now.

src/ethereum/base_types.py
ultratwo: Well spotted.

lightclient (approved):

PR #279, opened 2021-07-25
Fee history oldest block should only return an integer and not a block tag

lightclient:
### What was wrong?

The `oldestBlock` return value for `eth_history` should only return a integer block number, not a block tag like `latest`. 

### How was it fixed?

Use the integer type instead.


lightclient:
cc: @zsfelfoldi

SamWilsn:

PR #280, opened 2021-07-28
Implemented contract creation, create and return opcode and fixed bugs encountered while implementing

voith:
### What was been implemented?
- Create contracts when the receiver address is empty
- Implemented create Opcode
- Implemented Return Opcode

### TODO:
- [x] Add docstrings
- [x] Split different features and bugs into separate commits

#### Cute Animal Picture
Screenshot 2021-08-27 at 8 12 54 PM


src/ethereum/frontier/vm/instructions/system.py
```diff
@@ -0,0 +1,110 @@
+"""
+Ethereum Virtual Machine (EVM) System Instructions
```
SamWilsn, line 2: Why `System`?

src/ethereum/frontier/vm/instructions/system.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    evm.gas_left = subtract_gas(evm.gas_left, gas_cost)
+    extend_memory(evm.memory, memory_start_position, memory_size)
+    sender_address = evm.env.origin
+    sender = evm.env.state[sender_address]
+    if (sender.balance < endowment) or (
+        evm.message.depth + 1 > STACK_DEPTH_LIMIT
+    ):
+        push(evm.stack, U256(0))
+        return None
```
SamWilsn, line 59: Would you mind doing this as two separate checks? I think it makes the failure cases more obvious.

```suggestion
    if sender.balance < endowment:
        push(evm.stack, U256(0))
        return None

    if evm.message.depth + 1 > STACK_DEPTH_LIMIT:
        push(evm.stack, U256(0))
        return None
```

src/ethereum/frontier/vm/instructions/system.py
```diff
@@ -0,0 +1,110 @@
...
+        create_address=contract_address,
+        storage_address=contract_address,
+        depth=evm.message.depth + 1,
+    )
+    child_evm = process_create_message(child_message, evm.env)
+    push(evm.stack, U256.from_be_bytes(child_evm.message.storage_address))
+    evm.gas_left = child_evm.gas_left
+
+
+def return_opcode(evm: Evm) -> None:
```
SamWilsn, line 90: Yuck. Stupid keyword clashes.

According to [PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments), we can use `return_` which might be more consistent with the other opcode implementations?

src/ethereum/frontier/vm/interpreter.py
```diff
@@ -25,74 +27,130 @@
...
+    Parameters
+    ----------
+    message :
+        Transaction specific items.
+    env :
+        External items required for EVM execution.
+
+    Returns
+    -------
+    evm: `ethereum.vm.EVM`
```
SamWilsn, line 106:
```suggestion
    evm: `ethereum.frontier.vm.Evm`
```

src/ethereum/utils/address.py
```diff
@@ -0,0 +1,40 @@
...
+
+Address specific utility functions used in this application.
+"""
+from ethereum.base_types import Uint
+from ethereum.crypto import keccak256
+from ethereum.frontier import rlp
+from ethereum.frontier.eth_types import Address
+
+
+def compute_contract_address(address: Address, nonce: Uint) -> Address:
```
SamWilsn, line 20: This might be taking futureproofing too far, but with all the recent talk of Address Space Extension, this function might be better located in the `frontier` package?

src/ethereum/utils/numeric.py
```diff
@@ -9,34 +9,11 @@
...
-        The return value is based on math signum function.
-    """
-    if value < 0:
-        return -1
-    elif value == 0:
-        return 0
-    else:
-        return 1
-
-
```
SamWilsn, line 39: This feels like a generic numeric function, why put it in `transaction`?

whitelist.txt
```diff
@@ -224,6 +224,7 @@ callcode
 return
 delegatecall
 create2
+codedeposit
```
SamWilsn, line 227: I would lightly prefer using the underscored spelling: `GAS_CODE_DEPOSIT`. Thoughts?

tests/frontier/vm/vm_test_helpers.py
```diff
@@ -24,13 +28,15 @@ def run_test(test_dir: str, test_file: str) -> None:
...
-        depth=test_data["depth"],
+    gas_left, logs = process_message_call(
+        message=prepare_message(
+            caller=test_data["caller"],
+            target=target,
+            value=test_data["value"],
+            data=test_data["data"],
+            gas=test_data["gas"],
+            env=env,
+        ),
```
SamWilsn, line 39: Would you mind creating a temporary for `message`?

SamWilsn (changes requested):
I definitely feel like this'll need more eyes on it, if you wouldn't mind @lightclient.

src/ethereum/frontier/vm/instructions/system.py
voith: That's what the YP calls it :)

src/ethereum/frontier/vm/instructions/arithmetic.py
```diff
@@ -351,7 +351,8 @@ def signextend(evm: Evm) -> None:
 
         # Now among the obtained value bytes, consider only
         # N `least significant bytes`, where N is `byte_num + 1`.
-        value_bytes = value_bytes[len(value_bytes) - 1 - byte_num :]
+        if len(value_bytes) > byte_num:
+            value_bytes = value_bytes[len(value_bytes) - 1 - byte_num :]
```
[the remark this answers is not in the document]
voith: cc @Bhargavasomu. I didn't really go too deep to understand the logic but this was breaking a test and I realized that I could fix this with a simple if. 
Happy to open a separate PR for this.

src/ethereum/frontier/eth_types.py
```diff
@@ -46,7 +46,7 @@ class Transaction:
     nonce: U256
     gas_price: U256
     gas: U256
-    to: Optional[Address]
+    to: Address
```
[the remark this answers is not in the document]
voith: Here's what they YP says:
Screenshot 2021-07-28 at 11 35 32 PM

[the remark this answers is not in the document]
SamWilsn: Ah, so the type annotation in `Transaction` should be something like `Union[Tuple[()], Address]`?

src/ethereum/frontier/vm/instructions/arithmetic.py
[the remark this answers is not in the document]
SamWilsn: https://github.com/ethereum/eth1.0-specs/issues/283

src/ethereum/frontier/vm/instructions/system.py
SamWilsn: Good enough for me.

src/ethereum/frontier/vm/instructions/arithmetic.py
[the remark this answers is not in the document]
Bhargavasomu: I believe that `len(value_bytes) - 1 - byte_num` is overflowing U256. A simple fix for this would be `len(value_bytes) - 1 - Uint(byte_num)`.

I don't think converting to bytes32 is needed unless it adds more readability.

src/ethereum/frontier/eth_types.py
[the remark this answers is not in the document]
lightclient: If there is a way to continue to use optional, I think that is more readable than the `Union`. RLP generally understands things as "0 length" as being either `0xc0` or `0x80` depending on the type. I would _think_ that is how optional would be understood, but maybe I'm wrong.

src/ethereum/frontier/spec.py
```diff
@@ -338,6 +342,64 @@ def process_transaction(
     return (gas_used, logs)
 
 
+def prepare_message(
```
lightclient, line 345: I'm not sure if this warrants a separate function?

src/ethereum/frontier/vm/__init__.py
```diff
@@ -42,22 +42,36 @@ class Environment:
     state: State
 
 
+@dataclass
+class Message:
+    """
+    Items that are passed between two accounts
```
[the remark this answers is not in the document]
lightclient: This is a pattern that I've found relatively annoying in `go-ethereum`, that is the tx / msg objects. Is there a functional benefit of the msg abstraction vs. passing in parameters and having the EVM objects track the "msg" values for a certain frame?

lightclient (changes requested):

voith:
This needs some more work. I will ping the maintainers when its ready for merge.

Tkb68 (approved):

src/ethereum/frontier/vm/__init__.py
```diff
@@ -42,22 +42,36 @@ class Environment:
...
+    """
+
+    caller: Address
+    current: Address
+    gas: U256
+    value: U256
+    data: Bytes
+    storage_address: Address
```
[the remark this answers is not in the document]
voith: I have renamed this to `current_target`. I hope this is name makes better sense.

voith:
@SamWilsn @lightclient This is ready for review. Can you please take a look?

src/ethereum/frontier/vm/__init__.py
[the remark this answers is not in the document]
voith: @SamWilsn Let me know if docstring in the latest commit makes sense.

src/ethereum/frontier/utils/address.py
```diff
@@ -32,3 +34,26 @@ def to_address(data: Union[Uint, U256]) -> Address:
...
+        account.
+
+    Returns
+    -------
+    address: `ethereum.frontier.eth_types.Address`
+        The computed address of the new account.
+    """
+    computed_address = keccak256(rlp.encode([address, nonce]))
+    canonical_address = computed_address[-20:]
+    padded_address = canonical_address.rjust(20, b"\x00")
```
ultratwo, line 58: What does this line do? `computed_address` is always 32 bytes (because it is a 32 byte keccak), `canonical_address` always chops it down to 20 bytes, so padding to 20 bytes is always a no-op.

SamWilsn (approved):
I think I'm good with this, but if anyone wants more time to review, speak today, or forever hold your silence.

Issue #241, opened 2021-07-07
Consider using Enum for Nibbles type


lightclient:
From @pipermerriam:

Currently, the `trie` implementation uses a `bytes` type for the sequence of nibbles, with the stated validation rules that each byte may be no larger than 15.  This requirement doesn't appear to actually be enforced **and** it would be ideal if the type definitions actually constrained the inputs via the type system rather than runtime validation.

I would suggest using an `enum.IntEnum` for this:

```python
class Nibble(enum.IntEnum):
    NIBBLE_0 = 0x0
    NIBBLE_1 = 0x1
    ...
    NIBBLE_F = 0xf

Nibbles = List[Nibble]
```

Not only will this give you strong guarantees on anything that needs to take a `Nibbles` type as an input, but it also lets you have runtime validation for free

```python
def bytes_to_nibbles(value: bytes) -> Nibbles:
    return [Nibble(v) for v in value]  # will raise a TypeError if the value isn't part of the Enum)
```

abdulqgg:
Hey, I would like to tackle this problem as my first.

lightclient:
@abdulqgg I have assigned you to this, let us know if you have any questions. Thanks!

abdulqgg:
@lightclient Where can i find the `trie` implication that i will need to edit for this issue?

SamWilsn:
> @lightclient Where can i find the `trie` implication that i will need to edit for this issue?

https://github.com/ethereum/eth1.0-specs/blob/master/src/ethereum/frontier/trie.py

Unless you mean something else?

gurukamath:
@SamWilsn @lightclient I think the `Nibbles` datatype mentioned in the issue will eventually have to serve as the `key` for various nodes in a trie. I don't think it is right to have the key of a dictionary be a list.

We should however, just implement an explicit `byte <= 0xF` check on all `bytes` that are supposed to be `nibbles`

BoboTiG:
What is the status?
#281 PR was abandoned, but I could give a hand if everyone is OK with the initial request.

PR #281, opened 2021-07-28
Update trie.py

abdulqgg:
Lines 29-52
Solving issue #241 
I think ive done it right? First issue not sure

### What was wrong?
Enforcing requirement that each byte may be no larger than 15
Related to Issue #241

### How was it fixed?
Using enum.IntEnum

### DIFF
--- a/src/ethereum/frontier/trie.py
+++ b/src/ethereum/frontier/trie.py
@@ -26,6 +26,30 @@ verbose = False
 
 Node = Union[Account, Bytes, Transaction, Receipt, Uint, U256]
 
+from enum import Enum
+
+class Nibble(enum.IntEnum):
+    NIBBLE_0 = 0x0
+    NIBBLE_1 = 0x1
+    NIBBLE_2 = 0x2
+    NIBBLE_3 = 0x3
+    NIBBLE_4 = 0x4
+    NIBBLE_5 = 0x5
+    NIBBLE_6 = 0x6
+    NIBBLE_7 = 0x7
+    NIBBLE_8 = 0x8
+    NIBBLE_9 = 0x9
+    NIBBLE_A = 0xa
+    NIBBLE_B = 0xb
+    NIBBLE_C = 0xc
+    NIBBLE_D = 0xd
+    NIBBLE_E = 0xe
+    NIBBLE_F = 0xf
+
+Nibbles = List[Nibble]
+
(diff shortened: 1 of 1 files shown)


src/ethereum/frontier/trie.py
```diff
@@ -26,6 +26,30 @@
 
 Node = Union[Account, Bytes, Transaction, Receipt, Uint, U256]
 
+from enum import Enum
```
SamWilsn, line 29: Looking good so far! I think you'll want to run the [linting tools](https://github.com/ethereum/eth1.0-specs#development), because there are a [few errors](https://github.com/ethereum/eth1.0-specs/pull/281/checks?check_run_id=3184303985#step:6:13).

src/ethereum/frontier/trie.py
```diff
@@ -26,6 +26,30 @@
...
+    NIBBLE_D = 0xd
+    NIBBLE_E = 0xe
+    NIBBLE_F = 0xf
+
+Nibbles = List[Nibble]
+
+def bytes_to_nibbles(value: bytes) -> Nibbles:
+    return [Nibble(v) for v in value]  # will raise a TypeError if the value isn't part of the Enum)
 
 def nibble_list_to_compact(x: Bytes, terminal: bool) -> bytearray:
```
SamWilsn, line 54: I think the next step would be to change `x` from `Bytes` to `Nibbles` (and make the associated changes in the implementation.)

src/ethereum/frontier/trie.py
abdulqgg: Im struggling to use the linting tools to solve the few errors i have. Ive installed Python 3.7 and Pypy 7.3 and unsure where to go from there.

abdulqgg: I was following your instruction and managed to install `tox` but at step 2 with `git submodule update --init --recursive` i got the error: 

> fatal: not a git repository (or any of the parent directories): .git

I continued to the second part (was unsure whether i had to complete the first part to proceed) and reached step 4. I didnt understand which directory to change to to install the tools.
Thanks for your help and please may you link the discord group because i cant seem to find it.

SamWilsn: You can find the Discord here: https://discord.com/invite/qGpsxSA

I have the same username, feel free to @ me in the `#specifications` channel and I can help you get set up!

PR #284, opened 2021-07-29
Run commented test cases as the required opcodes have been implemented

Bhargavasomu:
### What was wrong?
Few tests were previously commented out as all the opcodes haven't been implemented. Most of those commented tests are now running. This also fixes the infinite loop in `get_valid_jump_destinations` when there is no valid opcode.

#### Cute Animal Picture


Bhargavasomu:
@SamWilsn @lightclient can you please review this? Thanks.

SamWilsn (approved):

PR #293, opened 2021-08-04
Fixes transaction parameters for different transaction types.

MicahZoltu:
Note: The `type` should be a single value literal for each transaction type, but I don't know how to do that.  I moved the type into the specific transactions in hopes that someone will be able to fix it to be the appropriate literal instead of a byte for each.

### DIFF
--- a/json-rpc/spec.json
+++ b/json-rpc/spec.json
@@ -1635,7 +1635,6 @@
 							"nonce",
 							"to",
 							"value",
-							"v",
 							"r",
 							"s"
 						],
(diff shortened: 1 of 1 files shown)


timbeiko:
@MicahZoltu who should I ping to review this as a sanity check ?

MicahZoltu:
@lightclient or @alita-moore would probably be a good candidate.

lightclient:
@MicahZoltu 1) I think this should be moved to the `eth1.0-apis` repo since this spec is deprecated now and 2) since some of these changes are breaking `v -> yParity` we'll need to discuss how to properly handle that.

lightclient:
I opened an issue on that repo to track this: https://github.com/ethereum/eth1.0-apis/issues/41

MicahZoltu:
Ah, I didn't realize that we moved JSON-RPC spec.

PR #294, opened 2021-08-04
Add genesis block

Bhargavasomu:
### What's new?
Add infrastructure and tests for the genesis block.

#### Cute Animal Picture


Bhargavasomu:
@SamWilsn @lightclient can you please review the code? Thanks.

Bhargavasomu:
FWIW, I know that the utils file is pretty clumsy at the moment. Will open a new PR to refactor some areas of the codebase.

src/ethereum/utils.py
```diff
@@ -59,3 +76,354 @@ def ceil32(value: Uint) -> Uint:
...
+    byte_stream : `bytes`
+        The byte stream obtained from the passed string.
+    """
+    if len(data) == 20:
+        return data
+
+    if len(data) < 20:
+        return data.rjust(20, b"\x00")
+
+    return data[-20:]
```
pipermerriam, line 101: Any thoughts on making sure that we aren't truncating non-zero bytes?

src/ethereum/utils.py
```diff
@@ -59,3 +76,354 @@ def ceil32(value: Uint) -> Uint:
...
+
+
+def to_bytes(data: str) -> Bytes:
```
pipermerriam, line 104: This function is probably a bad idea.  We had it in `eth-utils` a long time back and ultimately kept running up against the fact that not all `str` types can be naively decoded into a `bytes` type.

src/ethereum/utils.py
Bhargavasomu: I would actually use an assert statement here to ensure that non-zero bytes are not getting truncated, but it seems that non-zero bytes can be truncated when forced to an address as per the yellow paper. To be more specific, I am referring to `BALANCE` EVM opcode from the yellow paper, as this opcode implementation also uses the same function. Let me know what you think, thanks.

src/ethereum/utils.py
Bhargavasomu: I agree Piper. Currently it's being used only in the tests module, so maybe I'll add a note mentioning that this can't work against all input formats?

src/ethereum/genesis.py
```diff
@@ -0,0 +1,84 @@
...
+
+    difficulty: Uint
+    extra_data: Bytes
+    gas_limit: Uint
+    nonce: Bytes8
+    timestamp: U256
+    alloc: Dict[Address, Dict[str, U256]]
+
+
+def load_genesis_config(genesis_file: str) -> GenesisConfig:
```
SamWilsn, line 48: I think we can simplify this to:

```suggestion
def genesis_configuration() -> GenesisConfiguration:
```
Or maybe:

```suggestion
def genesis_configuration(genesis_path: Optional[str] = None) -> GenesisConfiguration:
```

I don't think configuring the path will be super common, though I guess we should support it so we can sync testnets?

src/ethereum/genesis.py
```diff
@@ -0,0 +1,84 @@
...
+        difficulty=hex_to_uint(genesis_data["difficulty"]),
+        extra_data=hex_to_bytes(genesis_data["extraData"]),
+        gas_limit=hex_to_uint(genesis_data["gasLimit"]),
+        nonce=hex_to_bytes8(genesis_data["nonce"]),
+        timestamp=hex_to_u256(genesis_data["timestamp"]),
+        alloc=alloc_data,
+    )
+
+
+ASSETS_DIR = Path(__file__).parent / "assets"
```
SamWilsn, line 82: There's a bit of funkiness in Python (ex. packages can be installed as zip files), so access to static assets should be done through the `importlib.resources` package.

I'd probably recommend using [`importlib.resources.open_binary`](https://docs.python.org/3.7/library/importlib.html?highlight=importlib#importlib.resources.open_binary). Maybe something like:

```python
if genesis_path is None:
    from . import assets
    with open_binary(assets, "mainnet.json") as genesis_resource:
        genesis_data = json.load(genesis_resource)
else:
    with open(genesis_path, "r") as genesis_file:
        genesis_data = json.load(genesis_file)
```

src/ethereum/genesis.py
```diff
@@ -0,0 +1,84 @@
...
+
+MAINNET_GENESIS_CONFIG = load_genesis_config(f"{ASSETS_DIR}/mainnet.json")
```
SamWilsn, line 84: Constants are evaluated at import time, and exist forever. In the interest of speeding up imports, I think we should just call `genesis_configuration()` directly in the code that needs it, instead of loading it during import.

src/ethereum/scripts/gen_genesis_config.go
```diff
@@ -0,0 +1,15 @@
+package main
```
SamWilsn, line 1: Would you mind putting this and [`b2t.sh`](https://github.com/ethereum/eth1.0-specs/blob/master/b2t.sh) into `<repository-root>/scripts/...`?

I'd like to keep `src` as pure python, if possible :P

src/ethereum/utils.py
SamWilsn: I'd actually just move it to the tests module.

src/ethereum/utils.py
```diff
@@ -59,3 +76,354 @@ def ceil32(value: Uint) -> Uint:
         return value
     else:
         return value + ceiling - remainder
+
+
+def to_valid_address(data: bytes) -> Address:
```
SamWilsn, line 81: Can we maybe go with something like:

```python
def to_address(data: Union[U256, UInt]) -> Address:
```

Or are there places where an actual `bytes` or `bytearray` is passed in?

If we can use the integer types, we can simplify this function to use `to_be_bytes32`:

```python
def to_address(data: Union[U256, UInt]) -> Address:
    return data.to_be_bytes32()[-20:]
```

src/ethereum/utils.py
```diff
@@ -59,3 +76,354 @@ def ceil32(value: Uint) -> Uint:
...
+
+    Returns
+    -------
+    modified_hex_string : `str`
+        The hexadecimal string with the 0x prefix removed if present.
+    """
+    if has_hex_prefix(hex_string):
+        return hex_string[len("0x") :]
+
+    return hex_string
```
SamWilsn, line 161: This is probably a huge preference thing, but I would much rather we _require_ a hex prefix, rather than allowing it.

Thoughts?

src/ethereum/utils.py
```diff
@@ -59,3 +76,354 @@ def ceil32(value: Uint) -> Uint:
...
+
+    Returns
+    -------
+    converted : `U256`
+        The U256 integer obtained from the given hexadecimal string.
+    """
+    return U256(int(remove_hex_prefix(hex_string), 16))
+
+
+def rlp_hash(data: Any) -> Hash32:
```
SamWilsn, line 317:
```suggestion
def rlp_hash(data: rlp.RLP) -> Hash32:
```

(I know mypy treats them as the same, but this is technically correct.)

whitelist.txt
```diff
@@ -106,6 +107,7 @@ sar
 
 keccak256
 
+alloc
```
SamWilsn, line 110: I'd like to avoid as many short forms and abbreviations as possible.

```suggestion
```

SamWilsn (changes requested):

src/ethereum/utils.py
pipermerriam: We dealt with this via the following code organization convention.

The module `library.tools` is where things go that are **not** allowed to be used by the rest of the library.  They are tools built on-top of or alongside the library but are not part of the core functionality.  This keeps them in the main library namespace, allowing them to be linted and type checked, but avoiding the problem of mixing in things like testing utilities with the rest of the library code.

Putting stuff in `./tests/...` that you cross import proved problematic in some situations since `./tests/` tends to not be a valid python module (since it doesn't and shouldn't have an `__init__.py` ).

src/ethereum/utils.py
pipermerriam: maybe this is two functions then?

- one for testing/utility that does `bytes -> bytes20`
- one for EVM operations that does `uint256 -> bytes20` with truncation/padding

The `BALANCE` opcode case should in theory be operating on a `uint256` popped off the stack and would need both padding and truncation.  The other use cases such as turning `b'\x01` into a full address for things like precompiles probably only needs to deal with padding and probably shouldn't truncate.

src/ethereum/genesis.py
Bhargavasomu: Yes the idea is to keep it generic for potentially syncing other chains, though I don't think the rest of the code is **yet** generic to support multiple chains.

src/ethereum/utils.py
Bhargavasomu: I don't know, this utility deals well with both non-prefixed and prefixed hex strings. So it might be a good to have feature right?

src/ethereum/genesis.py
Bhargavasomu: Going forward while we support multiple chains and custom genesis blocks, I think we would need a configuration manager and we can directly call the `genesis_configuration` code over there. Is it ok if I mention this as a `todo` comment for the time being? Thanks.

src/ethereum/genesis.py
Bhargavasomu: But this would involve making assets a python module, which doesn't seem technically necessary. With the presence of `Manifest` and the above code, I think `Path.parent / assets` would work as after the installation, the directory structure would look the same.

Please let me know if I am missing something or incorrect. Thanks.

src/ethereum/genesis.py
SamWilsn: That makes sense. I guess we'll need to come up with a more generic solution where we can ship testnets as part of the package too, and support loading from an external file.

src/ethereum/genesis.py
SamWilsn: Technically there doesn't have to be a file _at all_ for the resource. An example is [executable zip-files](https://docs.python.org/3/library/zipapp.html), where all the dependencies and source are packaged into a single file.

There's probably a way to import it without creating a new module, maybe `open_binary(".", "foo.json")`?

src/ethereum/genesis.py
SamWilsn: As long as we avoid loading the JSON at import time I'm pretty happy to merge.

src/ethereum/utils.py
SamWilsn: I'm not sure. If you require the `0x` prefix, it's unambiguous that the data is hex.

Like I said, not a huge deal though.

src/ethereum/utils.py
lightclient: +1

src/ethereum/utils.py
lightclient, line 101:
```suggestion
    return data[12:32]
```
seems clearer and less pythony?

src/ethereum/utils.py
lightclient: I would prefer we require prefixed hex strings, but there are a handful of places in eth tooling where naked hex strings are used. :/

src/ethereum/genesis.py
Bhargavasomu: I have removed this from import time, and please consider this discussion resolved.

Though I completely agree that the import time would be more, I would like a chance to explain why I preferred loading the JSON at import time itself. Please find the minimal reproducer below.

### constants.py
```
from time import sleep

def load(val):
    sleep(10)
    return val

c1 = load(5)
c2 = load(7)
```

### a.py
```
from .constants import c1, c2

def hello():
    print("Calling function hello")
    print(c1)
    print(c2)
```

### b.py
```
from .constants import c1, c2

def hello1():
    print("Called function hello1")
    print(c1)
    print(c2)
```

### ipython output
```
In [1]: time from test_proj.a import hello                                                                                                                                                                         
CPU times: user 660 µs, sys: 3.74 ms, total: 4.4 ms
Wall time: 20 s

In [2]: time hello()                                                                                                                                                                                               
Calling function hello
5
7
CPU times: user 70 µs, sys: 8 µs, total: 78 µs
Wall time: 84.9 µs

In [3]: time from test_proj.b import hello1                                                                                                                                                                        
CPU times: user 1.24 ms, sys: 132 µs, total: 1.37 ms
Wall time: 1.09 ms

In [4]: time hello1()                                                                                                                                                                                              
Called function hello1
5
7
CPU times: user 278 µs, sys: 40 µs, total: 318 µs
Wall time: 232 µs
```

## Doubt
Hence as per above reproducer, it doesn't seem like there is a difference between loading the genesis wherever needed and loading the genesis as a constant and importing it. Rather making the genesis as a constant and importing it wherever needed is actually memorizing it automatically. That's why I initially loaded the config as a constant. Given that at this point of time, there are no multiple usages of this config, I'm removing it. Please let me know if I am incorrect anywhere, I'm just posting this to improve my knowledge. Thanks.

Bhargavasomu:
@SamWilsn I have modified the code accordingly. I am not able to get it succeeded in tox, though `tox -e py3` is working smoothly locally.

Can you please help me by checking if it is working locally for you as well? Thanks.

src/ethereum/genesis.py
SamWilsn: On line `[1]` of your output, we see that _just importing_ `hello` took 20 seconds of wall time.

That's exactly what we want to avoid. Someone who's simply importing the software and not calling any functions shouldn't have to pay the cost of loading the genesis data.

src/ethereum/utils.py
SamWilsn: I'm too lazy to dig up my Emperor Palpatine "I am the senate" meme, but I'm sure you'll remember it ;)

lightclient: haha

lightclient (approved):

SamWilsn (approved):

src/ethereum/genesis.py
Bhargavasomu: Got it, thanks Sam.

PR #302, opened 2021-08-09
Add a proper state interface

ultratwo:

This pull request adds a proper state interface in the form of `State` and `Trie` types. Both of these types are designed so they can be easily replaced by faster implementations if needed. The trie code has also been refactored to be clearer.

Initializing and deletion of accounts happens automatically. Since in the Ethereum state datamodel values such as `EMPTY_ACCOUNT` and `U256(0)` are represented by their absence, an entry can be deleted by setting it to one of these values.

`SELFDESTRUCT` will require special handling, which is not currently implemented.

### DIFF
--- a/src/ethereum/frontier/state.py
+++ b/src/ethereum/frontier/state.py
@@ -1,0 +1,221 @@
+"""
+State
+^^^^^
+
+.. contents:: Table of Contents
+    :backlinks: none
+    :local:
+
+Introduction
+------------
+
+The state contains all information that is preserved between transactions.
+
+It consists of a main account trie and storage tries for each contract.
+"""
+from dataclasses import dataclass, field
+from typing import Callable, Dict
+
+from ethereum.base_types import U256, Bytes, modify
+
+from .eth_types import EMPTY_ACCOUNT, Account, Address, Root
+from .trie import EMPTY_TRIE_ROOT, Trie, root, trie_get, trie_set
+
+
+@dataclass
+class State:
+    """
+    Contains all information that is preserved between transactions.
+    """
+
+    _main_trie: Trie[Account] = field(
+        default_factory=lambda: Trie(secured=True, default=EMPTY_ACCOUNT)
+    )
+    _storage_tries: Dict[Address, Trie[U256]] = field(default_factory=dict)
+
+
+def get_account(state: State, address: Address) -> Account:
+    """
+    Get the `Account` object at an address. Returns `EMPTY_ACCOUNT` if there
+    is no account at the address.
+
+    Parameters
+    ----------
+    state: `State`
+        The state
+    address : `Address`
+        Address to lookup.
+
+    Returns
+    -------
+    account : `Account`
+        Account at address.
+    """
+    account = trie_get(state._main_trie, address)
+    assert isinstance(account, Account)
+    return account
+
+
+def set_account(state: State, address: Address, account: Account) -> None:
+    """
+    Set the `Account` object at an address. Setting to `EMPTY_ACCOUNT` deletes
+    the account (but not its storage, see `destroy_account()`).
+
+    Parameters
+    ----------
+    state: `State`
+        The state
+    address : `Address`
+        Address to set.
+    account : `Account`
+        Account to set at address.
+    """
+    trie_set(state._main_trie, address, account)
+
+
+def destroy_account(state: State, address: Address) -> None:
+    """
+    Completely remove the account at `address` and all of its storage.
+
+    This function is made available exclusively for the `SELFDESTRUCT`
+    opcode. It is expected that `SELFDESTRUCT` will be disabled in a future
+    hardfork and this function will be removed.
+
+    Parameters
+    ----------
+    state: `State`
+        The state
+    address : `Address`
+        Address of account to destroy.
+    """
+    del state._storage_tries[address]
+    set_account(state, address, EMPTY_ACCOUNT)
+
+
+def get_storage(state: State, address: Address, key: Bytes) -> U256:
+    """
+    Get a value at a storage key on an account. Returns `U256(0)` if the
+    storage key has not been set previously.
+
+    Parameters
+    ----------
+    state: `State`
+        The state
+    address : `Address`
+        Address of the account.
+    key : `Bytes`
+        Key to lookup.
+
+    Returns
+    -------
+    value : `U256`
+        Value at the key.
+    """
+    trie = state._storage_tries.get(address)
+    if trie is None:
+        return U256(0)
+
+    value = trie_get(trie, key)
+
+    assert isinstance(value, U256)
+    return value
+
+
+def set_storage(
+    state: State, address: Address, key: Bytes, value: U256
+) -> None:
+    """
+    Set a value at a storage key on an account. Setting to `U256(0)` deletes
+    the key.
+
+    Parameters
+    ----------
+    state: `State`
+        The state
+    address : `Address`
+        Address of the account.
+    key : `Bytes`
+        Key to set.
+    value : `U256`
+        Value to set at the key.
+    """
+    trie = state._storage_tries.get(address)
+    if trie is None:
+        trie = Trie(secured=True, default=U256(0))
+        state._storage_tries[address] = trie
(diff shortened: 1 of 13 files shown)


src/ethereum/frontier/trie.py
```diff
@@ -38,22 +238,22 @@ def nibble_list_to_compact(x: Bytes, terminal: bool) -> bytearray:
...
     used.
 
     Parameters
     ----------
     x :
         Array of nibbles.
-    terminal :
-        Flag denoting if the key points to a terminal (leaf) node.
+    is_leaf :
+        Flag denoting whether this is part of a leaf node or an extension node.
```
SamWilsn, line 256:
```suggestion
        True if this is part of a leaf node, or false if it is an extension node.
```

src/ethereum/frontier/trie.py
```diff
@@ -63,78 +263,85 @@ def nibble_list_to_compact(x: Bytes, terminal: bool) -> bytearray:
...
-    Maps all compact keys to nibble-list format. Optionally hashes the keys.
+    nibble_list = bytearray(2 * len(bytes))
+    for i in range(2 * len(bytes)):
+        byte_idx = i // 2
+        if i % 2 == 0:
+            # get upper nibble
+            nibble_list[i] = (bytes[byte_idx] & 0xF0) >> 4
+        else:
+            # get lower nibble
+            nibble_list[i] = bytes[byte_idx] & 0x0F
```
SamWilsn, line 299:
```suggestion
    for byte_index, byte in enumerate(bytes):
        nibble_index = byte_index * 2
        
        # Get upper nibble.
        nibble_list[nibble_index] = (bytes[byte_idx] & 0xF0) >> 4
        
        # Get lower nibble.
        nibble_list[nibble_index + 1] = bytes[byte_idx] & 0x0F
```

SamWilsn: At least I think that'll work.

src/ethereum/frontier/trie.py
```diff
@@ -63,78 +263,85 @@ def nibble_list_to_compact(x: Bytes, terminal: bool) -> bytearray:
...
-                nibble_list[i] = key[byte_idx] & 0x0F
-
-        mapped[Bytes(nibble_list)] = value
+        if isinstance(value, Account):
+            assert get_storage_root is not None
+            encoded_value = encode_node(value, get_storage_root(preimage))
+        else:
+            encoded_value = encode_node(value)
+        # skip empty values, these are defined to be omitted from the trie
+        if value != b"":
```
SamWilsn, line 336: Can we remove empty values in the modifier functions, and just have an `assert` here?

It would make me uncomfortable to have multiple possible Python representations for the same conceptual trie.

SamWilsn (commented):
I'll take a closer look once we're using unbound functions, but so far I'm pretty happy with this

ultratwo:
I've made all the changes suggested above.

`State` and `Trie` both have canonical forms that the modifier functions keep them in.

src/ethereum/frontier/trie.py
```diff
@@ -199,73 +375,53 @@ def patricialize(obj: Mapping[Bytes, Node], i: Uint) -> rlp.RLP:
...
+        leaf = LeafNode(key[level:], obj[key])
+        return leaf
 
     # prepare for extension node check by finding max j such that all keys in
     # obj have the same key[i:j]
-    substring = copy(key)
-    j = Uint(len(substring))
+    substring = key[level:]
+    prefix_length = len(substring)
     for key in obj:
```
SamWilsn, line 391: I believe this for loop clobbers `key`. Just wanted to make sure that's intentional.

SamWilsn (commented):
Looking good! I do want a second pair of eyes on this to double check it though. @lightclient if you please?

src/ethereum/frontier/trie.py
ultratwo: All the uses of `key` outside loops are only used to interrogate properties common to all keys, so the fact that key gets clobbered haphazardly doesn't matter.

Renamed to `arbitrary_key` for clarity.

src/ethereum/frontier/trie.py
```diff
@@ -13,21 +13,209 @@
...
+    else:
+        trie._data[key] = value
+
+
+def trie_get(trie: Trie, key: Bytes) -> Node:
+    """
+    Gets an item from the Merkle Trie.
+
+    This method returns `b""` if the key is missing, since the Merkle Trie
+    does not distinguish between missing and `b""`.
```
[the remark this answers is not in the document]
ultratwo: Your proposed code is complete nonsense. Since annotations are simply hints that allow `pypy` to do typechecking and aren't used at runtime. `type(N) == TypeVar` and `N` isn't even callable.

I've taken the straightforward approach of giving `trie_get()` a `default` parameter with the same semantics as `dict.get()`.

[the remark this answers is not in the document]
SamWilsn: > Your proposed code is complete nonsense. Since annotations are simply hints that allow `pypy` [sic] to do typechecking and aren't used at runtime. `type(N) == TypeVar` and `N` isn't even callable.

This is a little harsh, especially since I noted _in my comment_ that you can't directly call a `TypeVar`. I was quite aware the sample wouldn't run as written, but I figured the spirit of what I suggested was clear enough: pass in the type of value stored in the trie, and default construct it when the key doesn't exist. Here's a more fleshed out version:

```python
N = TypeVar("N", Account, Bytes, Transaction, Receipt, Uint, U256)

def trie_get(class_: Type[N], trie: Trie[N], key: Bytes) -> N:
    try:
        return trie._data[key]
    except KeyError:
        return class_()
```

**NB**: I haven't run this through `mypy`, and would need to check that the `Type[N]` annotation behaves as I expect.

> I've taken the straightforward approach of giving `trie_get()` a `default` parameter with the same semantics as `dict.get()`.

The upside to adding a generic argument to `Trie` is that we get type safety between the different kinds of tries. You can't accidentally put a `Bytes` in the root trie (only `Account`), and vice versa. I understand that might be a bit more work, though, for a tiny payoff.

Whether we pass in a `Callable` or a default value into `trie_get` is somewhat immaterial to me, since constructing a dataclass should be a relatively lightweight operation.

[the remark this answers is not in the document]
ultratwo: > This is a little harsh, especially since I noted in my comment that you can't directly call a TypeVar.

Sorry for being thoughtlessly rude.

> Whether we pass in a `Callable` or a default value into `trie_get` is somewhat immaterial to me, since constructing a dataclass should be a relatively lightweight operation.

I think it's simpler to store the default value in the `Trie` object. This is incompatible with Verkle Tries, when both accounts and storage will be placed in the same trie, but that is a bridge to cross when we come to it.

SamWilsn (approved):

src/ethereum/frontier/state.py
```diff
@@ -0,0 +1,218 @@
...
+    key : `Bytes`
+        Key to set.
+    value : `U256`
+        Value to set at the key.
+    """
+    trie = state._storage_tries.get(address)
+    if trie is None:
+        trie = Trie(secured=True, default=U256(0))
+        state._storage_tries[address] = trie
+    trie_set(trie, key, value)
```
Bhargavasomu, line 143: I feel it would be giving more clarity to the readers if we mentioned here and in docstring that if the value is 0, and if the key is present in the storage, then it would be deleted from the storage. Please let me know what you think. Thanks.

src/ethereum/frontier/state.py
```diff
@@ -0,0 +1,218 @@
...
+    Parameters
+    ----------
+    state: `State`
+        The state
+    address : `Address`
+        Address to set.
+    account : `Account`
+        Account to set at address.
+    """
+    trie_set(state._main_trie, address, account)
```
Bhargavasomu, line 72: I feel it would be giving more clarity to the readers if we mentioned here and in docstring that if the account is `EMPTY_ACCOUNT`, then the address and account pair would be removed from the state trie. Please let me know what you think. Thanks.

Bhargavasomu (approved):

PR #306, opened 2021-08-10
Move modules depending on frontier types into frontier hardfork

SamWilsn:
### What was wrong?

Code outside of the `frontier` package depended on `frontier` types.

### How was it fixed?

Moved fork-specific code into `frontier`

#### Cute Animal Picture

pugsly

### DIFF
--- a/src/ethereum/frontier/genesis.py
+++ b/src/ethereum/frontier/genesis.py
@@ -18,15 +18,15 @@ from dataclasses import dataclass
 from typing import Dict, cast
 
 from ethereum.base_types import U256, Bytes, Bytes8, Uint, slotted_freezable
-from ethereum.frontier.eth_types import Address
-
-from .utils.hexadecimal import (
-    hex_to_address,
+from ethereum.utils.hexadecimal import (
     hex_to_bytes,
     hex_to_bytes8,
     hex_to_u256,
     hex_to_uint,
 )
+
+from .eth_types import Address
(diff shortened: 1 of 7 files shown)


voith (approved):
LGTM

Mmrad07:
hi

src/ethereum/frontier/genesis.py
```diff
@@ -18,16 +18,16 @@
...
-from .utils.hexadecimal import (
-    hex_to_address,
+from ethereum.utils.hexadecimal import (
     hex_to_bytes,
     hex_to_bytes8,
     hex_to_u256,
     hex_to_uint,
 )
 
+from .eth_types import Address
```
Bhargavasomu, line 28: I understand the intent that modules outside the frontier related stuff shouldn't import the frontier stuff. But the genesis seems that it can be uncoupled from the frontier stuff as a standalone entity.

Further I think that some of the artefacts present in `eth_types` (like `Address`), utils should be moved out of `frontier` module as they don't seem to be tightly coupled with `frontier`. Let me know what you think about this.

SamWilsn: `Address` does seem pretty generic, and my only argument for putting it in the hardfork is because of the recent address space extension talks.

This PR is just a quick bandaid to get things "good enough" for now.

Bhargavasomu: Sounds good, thanks.

Issue #238, opened 2021-07-07
Implement the Proof of Work Algorithm


lightclient:
From @voith:

What was missing?
We need to implement EthHash, the proof of work algorithm.

How can this be fixed?
* Refer to this link for details and test cases.
* Refer to this implementation to see if some code can be borrowed.

norswap:
Relevant link with what looks like a proper Python implementation: https://eth.wiki/en/concepts/ethash/ethash

PR #315, opened 2021-08-26
Add cache and dataset generation as part of PoW

Bhargavasomu:
### What's new?
Cache and Dataset generation pertaining to PoW has been implemented.
Related to Issue #238 

#### Cute Animal Picture


tests/test_ethash.py
```diff
@@ -0,0 +1,245 @@
...
+        1. The first 100 dataset indices are same when the python
+        implementation is compared with the DAG dataset.
+        2. Randomly take 500 indices between
+        [101, `dataset size in words` - 1] and ensure that the values are
+        same between python implementation and DAG dataset.
+    """
+    download_geth(tmpdir)
+
+    epoch_number = Uint(randint(0, 100))
+    block_number = epoch_number * HASH_BYTES
```
ultratwo, line 220: I think this should be `block_number = epoch_number * EPOCH_SIZE`.

Bhargavasomu: Not really. I kind of used `HASH_BYTES` instead of `EPOCH_SIZE` as I felt it would add a bit more entropy rather than testing on only the end blocks of the epoch.

Now to think of it, it makes sense to actually make this block_number random as well within the epoch for much more entropy. I'll be pushing this change in a while.

SamWilsn (approved):

norswap:
The test from this seems to fail for me:

```
E               OSError: [Errno 8] Exec format error: '/private/var/folders/19/ty9xxj7j3418_tsbtp06tjrc0000gn/T/pytest-of-norswap/pytest-0/test_dataset_generation_random0/geth'

...

1 failed, 858 passed in 71.25s (0:01:11)
```

(when running `pytest -m "not slow"`)

Are there some special requirements to run this? Could this be a problem with my setup?

- macOS Big Sur x64 (fairly sturdy dev machine)
- python = `pypy3.7-7.3.5`
- using a virtulenv (actually **venv**) in which I additionally installed `pytest` and `requests`

PR #316, opened 2021-08-26
Implemented opcodes CALL and CALLCODE

### LINKED ISSUE #230: Implement `*CALL` family
original issue: https://github.com/quilt/eth1.0-specs/issues/15

voith:
### Tickets
closes #230
depends on #280 

### What was implemented?
Implemented opcodes `CALL` and `CALLCODE`. 

#### Cute Animal Picture

Pic credits: [Shizzi](https://www.flickr.com/photos/izzi/5676414339/in/photostream)

### DIFF
--- a/src/ethereum/frontier/state.py
+++ b/src/ethereum/frontier/state.py
@@ -177,7 +177,7 @@ def state_root(state: State) -> Bytes:
     Parameters
     ----------
     state: `State`
-        The state
+        The current state.
 
     Returns
     -------
@@ -189,6 +189,25 @@ def state_root(state: State) -> Bytes:
         return storage_root(state, address)
 
     return root(state._main_trie, get_storage_root=get_storage_root)
+
+
+def account_exists(state: State, address: Address) -> bool:
+    """
+    Checks if an account exists in the state trie
+
+    Parameters
+    ----------
+    state:
+        The state
+    address:
+        Address of the account that needs to be checked.
+
+    Returns
+    -------
+    account_exists : `bool`
+        True if account exists in the state trie, False otherwise
+    """
+    return address in state._main_trie._data
 
 
 def modify_state(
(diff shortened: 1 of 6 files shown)


voith:
I have pushed

voith:
@SamWilsn @lightclient This is ready for review.

src/ethereum/frontier/state.py
```diff
@@ -191,6 +191,24 @@ def get_storage_root(address: Address) -> Root:
     return root(state._main_trie, get_storage_root=get_storage_root)
 
 
+def account_exists(state: State, address: Address) -> bool:
+    """
+    Checks if an account exists in the state trie
+    Parameters
```
SamWilsn, line 197:
```suggestion

    Parameters
```

src/ethereum/frontier/vm/instructions/system.py
```diff
@@ -107,3 +109,154 @@ def return_(evm: Evm) -> None:
...
+    extend_memory(evm.memory, memory_output_start_position, memory_output_size)
+    call_data = memory_read_bytes(
+        evm.memory, memory_input_start_position, memory_input_size
+    )
+    sender_balance = get_account(
+        evm.env.state, evm.message.current_target
+    ).balance
+    if sender_balance < value:
+        push(evm.stack, U256(0))
+        evm.gas_left += message_call_gas_fee
```
SamWilsn, line 159: Should we have an `add_gas` (or possibly `return_gas`) for symmetry with `subtract_gas`?

Probably not, since returning gas can't raise an exception, but I'm curious what your opinion is.

src/ethereum/frontier/vm/gas.py
```diff
@@ -118,3 +124,50 @@ def calculate_gas_extend_memory(
...
+    value:
+        The amount of `ETH` that needs to be transferred.
+
+    Returns
+    -------
+    call_gas_cost: `ethereum.base_types.U256`
+        The total gas amount for executing Opcodes `CALL` and `CALLCODE`.
+    """
+    _account_exists = account_exists(state, to)
+    create_gas_cost = GAS_NEW_ACCOUNT if not _account_exists else U256(0)
```
SamWilsn, line 152: Let's avoid the `not` if we can:

```suggestion
    create_gas_cost = U256(0) if _account_exists else GAS_NEW_ACCOUNT
```

src/ethereum/frontier/vm/gas.py
```diff
@@ -118,3 +124,50 @@ def calculate_gas_extend_memory(
...
+    transfer_gas_cost = GAS_CALL_VALUE if value != 0 else U256(0)
```
SamWilsn, line 153: Same thing here:

```suggestion
    transfer_gas_cost = U256(0) if value == 0 else GAS_CALL_VALUE
```

src/ethereum/frontier/vm/gas.py
```diff
@@ -118,3 +124,50 @@ def calculate_gas_extend_memory(
...
+        The amount of gas provided to the message call.
+    value:
+        The amount of `ETH` that needs to be transferred.
+
+    Returns
+    -------
+    message_call_gas_cost: `ethereum.base_types.U256`
+        The total gas amount for making the message-call.
+    """
+    return gas + (GAS_CALL_STIPEND if value != 0 else U256(0))
```
SamWilsn, line 173: Could we do something like:

```suggestion
def calculate_message_call_gas_stipend(value: U256) -> U256:
    """
    Calculates the gas stipend for making the message call with the given value.

    Parameters
    ----------
    value:
        The amount of `ETH` that needs to be transferred.

    Returns
    -------
    message_call_gas_stipend: `ethereum.base_types.U256`
        The gas stipend for making the message-call.
    """
    return U256(0) if value == 0 else GAS_CALL_STIPEND
```

Would need to do the addition in the caller though.

src/ethereum/frontier/vm/instructions/system.py
voith: As you pointed out, I would add an `add_gas` method if there was some extra logic to be handled like raising an exception. 
```python
evm.gas_left = add_gas(evm.gas_left, gas_mount)
```
vs 
```python
evm.gas_left += gas_amount
```
I don't see any added benefit between the two examples mentioned above. 


----

I would have added a method if the `gas_left` was mutated inside the method `return_gas`
```python
  evm.return_gas(gas_amount) # this call would automatically increase evm.gas_left by gas_amount
```

PR #324, opened 2021-09-06
Optimize integer types

ultratwo:
### What was wrong?

Integer types in `base_types.py` were slow.

### How was it fixed?

I have optimized the integer code in `base_types.py`, mostly by endrunning python's classes machinery (no calls to `super()`) and eliminating redundant checks. This results in a more than 2x speedup in dataset item generation time and knocks about 25% off cache generation on my machine under `cpython`.

`U256` and `Uint32` now use mostly shared code.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/base_types.py
+++ b/src/ethereum/base_types.py
@@ -55,14 +55,12 @@ class Uint(int):
     def from_le_bytes(cls: Type, buffer: "Bytes") -> "Uint":
         return cls(int.from_bytes(buffer, "little"))
 
-    def __new__(cls: Type, value: int) -> "Uint":
+    def __init__(self, value: int) -> None:
         if not isinstance(value, int):
             raise TypeError()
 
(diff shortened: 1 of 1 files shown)


src/ethereum/base_types.py
```diff
@@ -55,15 +55,13 @@ def from_be_bytes(cls: Type, buffer: "Bytes") -> "Uint":
     def from_le_bytes(cls: Type, buffer: "Bytes") -> "Uint":
         return cls(int.from_bytes(buffer, "little"))
 
-    def __new__(cls: Type, value: int) -> "Uint":
+    def __init__(self, value: int) -> None:
```
SamWilsn, line 58: Ah. I guess since we don't modify the int, we don't need the whole `__new__` nonsense. Nice!

SamWilsn (approved):

PR #325, opened 2021-09-06
Fix bugs in the state trie

ultratwo:
### What was wrong?

The state trie incorrectly handling cases involving empty accounts and storage resulting in incorrect state roots. Most state roots involving storage will be wrong prior to this fix.

Code (mostly involving gas costs) that needs to know about the difference between `EMPTY_ACCOUNT` and a non-existent one will need to use the new function `get_account_optional()`.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/frontier/trie.py
+++ b/src/ethereum/frontier/trie.py
@@ -54,7 +54,13 @@ EMPTY_TRIE_ROOT = hex_to_bytes(
 
 Node = Union[Account, Bytes, Transaction, Receipt, Uint, U256, None]
 T = TypeVar(
-    "T", Account, Bytes, Optional[Transaction], Optional[Receipt], Uint, U256
+    "T",
+    Optional[Account],
+    Bytes,
+    Optional[Transaction],
+    Optional[Receipt],
+    Uint,
+    U256,
 )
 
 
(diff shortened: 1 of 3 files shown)


src/ethereum/frontier/trie.py
```diff
@@ -141,10 +147,12 @@ def encode_node(node: Node, storage_root: Optional[Bytes] = None) -> Bytes:
     if isinstance(node, Account):
         assert storage_root is not None
         return rlp.encode_account(node, storage_root)
-    elif isinstance(node, (Transaction, Receipt)):
+    elif isinstance(node, (Transaction, Receipt, U256)):
         return rlp.encode(cast(rlp.RLP, node))
+    elif isinstance(node, Bytes):
+        return node
     else:
-        return cast(Bytes, node)
```
voith, line 147: maybe raise an exception here: 
```python
raise NotImplementedError(f"encoding for {type(node)} is not currently implemented")
```

src/ethereum/frontier/trie.py
```diff
@@ -141,10 +147,12 @@ def encode_node(node: Node, storage_root: Optional[Bytes] = None) -> Bytes:
...
+        assert False  # Unrecognized node type
```
voith, line 155: maybe raise an exception here: 
```python
raise NotImplementedError(f"encoding for {type(node)} is not currently implemented")
```

ultratwo: Fair

voith (approved):
This fixes all the bugs that I encountered while implementing pre-compiles.

SamWilsn (approved):

Issue #326, opened 2021-09-07
Add test dependencies


norswap:
I set up the repo today, and tried to run the tests as instructed in the README, however, it seems that the `requests` library is needed for that. Shouldn't this be added to some kind of dependency tracking list? 

There doesn't seem to any in the project, and I'm definitely a Python amateur, but shouldn't a `requirements.txt`be used in this case? Looking a bit more into it, it looks like to specify some dependencies in `setup.py` (and [many other ways](https://dephell.readthedocs.io/cmd-deps-convert.html) to do it) It guess it depends where `pytest` looks?

Also we want to avoid to have `requests` as an install dependencies if it is only used for tests.

A bit unrelated, but I also ran into another issue while running the tests: https://github.com/ethereum/execution-specs/pull/315#issuecomment-914467970

SamWilsn:
I think you may have missed the step that installed the test dependencies? [`requests` is in there](https://github.com/ethereum/execution-specs/blob/c2274790e8ac2d637c7dbe092477a1b21243916c/setup.cfg#L48).

```bash
$ pip install -e .[test]
```

I'd recommend either setting up a virtual environment before installing or just using tox (which is the more convenient way.)

norswap:
Ah thanks, I just thought that `pip install -e .` would install everything!

norswap:
Oh wow the square brackets are actual syntax. I thought that meant "optional".

norswap:
Felt stupid for a minute, as both `pip install -e .[doc,lint,test]` and `pip install -e .[test]` both result in "-e option requires 1 argument".

Turns out in bash, `[]` are [wildcard syntax](https://tldp.org/LDP/GNU-Linux-Tools-Summary/html/x11655.htm).

So the proper bash-compatible command ends up being `pip install -e ".[doc,lint,test]". I'll open a pull-request to fix.

I'll setup `tox`, I (foolishly) assumed it was a documentation toolchain!

PR #327, opened 2021-09-07
Quote globbing syntax in README

norswap:
### What was wrong?

cf. https://github.com/ethereum/execution-specs/issues/326#issuecomment-914574299

### How was it fixed?

Editing the README to quote the globbing syntax.

#### Cute Animal Picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ $ tox
 The development tools can also be run outside of `tox`, and can automatically reformat the code:
 
 ```bash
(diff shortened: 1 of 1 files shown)


SamWilsn (approved):
I'm curious which shell uses `[` and `]` for globbing?

norswap:
@SamWilsn Just plain old bash: https://www.gnu.org/software/bash/manual/bash.html#Pattern-Matching
I wasn't aware either!

PR #328, opened 2021-09-07
Implemented precompiled contracts

### LINKED ISSUE #264: Add support for precompiled contracts
### What was wrong?
The current implementation of message call doesn't handle the execution of precompiled smart contracts. 

### How can it be fixed?
The following section from the `frontier` yellow paper needs to be implemented.
Screenshot 2021-07-13 at 9 26 32 PM
As part of this implement at least one of the precompile for successful testing.

voith:
### Related Issue

closes #264

### What feature was added to this PR?
TLDR: Implemented precompiles specified in the frontier version of the YP.
List of things Implemented:
1. Added support for the interpreter to process pre-compiles.
2. Implemented pre-compiles ECRECOVER, SHA256, RIPEMD160, IDENTITY.
3. Allowed reverting state back to a previous snapshot if the execution results in an error.
4. Handled exceptions encountered while executing bytecode gracefully.
5. Refactored previous tests that expected exceptions to be raised.

Bonus Bug fix:
Some stack operations started failing after this PR started handling exceptions gracefully. Upon investigating I realized that the stack operations would assert statements before subtracting. This was causing a mismatch in gas consumption. I have fixed this in a separate [commit](https://github.com/voith/eth1.0-specs/commit/3319f9c5da7fcde1d46e310b600427b04bc1dd51)

#### Cute Animal Picture

Pic credits: [https://www.animalsaustralia.org/](https://www.animalsaustralia.org/features/cute-animals-help-you-concentrate.php)


voith:
I have no idea why the doc tests are failing. `sphinx-autobuild` seems to work fine.

voith:
@SamWilsn This is ready for review.

SamWilsn:
I removed this check in [`state.py`](https://github.com/ethereum/execution-specs/blob/37a9e408fc97265ec91d48897a098a155bd35495/src/ethereum/frontier/state.py#L174):

```python
    if value == 0:
        return None
```

@voith just want to make sure that's correct.

voith:
> I removed this check-in state.py while rebasing:

    if value == 0:
        return None

Thanks, it's no longer needed. 

Also, thanks for fixing the docs. Your fix makes me think there's a bug in `sphinx-build`.
Somehow your [commit](https://github.com/ethereum/execution-specs/commit/c5ed00ef4097aa40dbc7680e3cf5e4b14d995e83) mapped the precompiles incorrectly. I have fixed it and [squashed it](https://github.com/voith/eth1.0-specs/commit/e737f5c7639542789e653fc24531a0ef027f544e) with your commit.

SamWilsn:
> Somehow your [commit](https://github.com/ethereum/execution-specs/commit/c5ed00ef4097aa40dbc7680e3cf5e4b14d995e83) mapped the precompiles incorrectly.

Eep :sweat_smile:

src/ethereum/crypto.py
```diff
@@ -20,6 +20,8 @@
 Hash32 = Bytes32
 Hash64 = Bytes64
 
+secp256k1n = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141
```
SamWilsn, line 23: Not sure exactly how to do the screaming snake case for this one, but I think I prefer uppercase constants.

```suggestion
SECP256K1N = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141
```

src/ethereum/frontier/state.py
```diff
@@ -333,3 +350,47 @@ def write_code(sender: Account) -> None:
         sender.code = code
 
     modify_state(state, address, write_code)
+
+
+def snapshot_state(state: State) -> State:
```
SamWilsn, line 355: It was my understanding that everything in the state was frozen already, so couldn't you just do a single level copy?

voith: Yeah, that's the first thing I tried. However a simple deepcopy on state results in an error:

Option 1: Use copy
```
>>> state_snapshot = copy(state)
>>> acc_2 = bytes.fromhex("095e7baea6a6c7c4c2dfeb977efac326af552d00")
>>> increment_nonce(state, acc_2)
>>> assert state_snapshot != state
AssertionError                            Traceback (most recent call last)
<ipython-input-20-698273bffc21> in <module>
----> 1 assert new_state != state

AssertionError:
```
in the above example changes made to `state` are also applied to `state_snapshot`.

Option 2: Use deepcopy
```python
>>> from copy import copy, deepcopy
>>>  from ethereum.frontier.state import State, set_account, EMPTY_ACCOUNT, increment_nonce
>>> state = State()
>>> acc_1 = bytes.fromhex("095e7baea6a6c7c4c2dfeb977efac326af552d87")
>>> set_account(state, acc_1, EMPTY_ACCOUNT)
>>> deepcopy(state)
~/Python-Envs/eth-spec/lib/python3.7/copy.py in _reconstruct(x, memo, func, args, state, listiter, dictiter, deepcopy)
    290             if slotstate is not None:
    291                 for key, value in slotstate.items():
--> 292                     setattr(y, key, value)
    293
    294     if listiter is not None:

~/Projects/eth1.0-specs/src/ethereum/base_types.py in _setattr_function(self, attr, value)
    696
    697 def _setattr_function(self: Any, attr: str, value: Any) -> None:
--> 698     if self._frozen:
    699         raise Exception("Mutating frozen dataclasses is not allowed.")
    700     else:

AttributeError: _frozen
```

Which is why I had to come up with a custom implementation of deepcopy.

voith: @ultratwo any thoughts on this?

src/ethereum/crypto.py
voith: done

src/ethereum/frontier/state.py
ultratwo: A two layer copy is needed. `copy()` is two shallow, but your `deepcopy()` unnecessarily goes a layer too far and copies all the frozen stuff.

I think the best way to proceed is to implement the following interface, with the obvious semantics.
```
start_transaction()
commit_transaction()
abort_transaction()
```
I'll go implement that now.

voith: Cool, it would be awesome to have a `commit` `rollback` type interface.  But that shouldn't block this PR from being merged.

PR #334, opened 2021-09-10
Increase Python's default recursion limit

### LINKED ISSUE #335: Move recursion limit adjustment out of frontier
I don't think setting the recursion limit makes sense in the `frontier` part of the spec at all. This is probably common to all the hardforks.

I guess we don't have a well defined entrypoint for the cross hardfork stuff yet. Once we do, we should move this stuff there.

_Originally posted by @SamWilsn in https://github.com/ethereum/execution-specs/pull/334#pullrequestreview-752070433_

voith:
### What was wrong?
closes #335

There are some tests that use `CALL` and `CALLCODE` to make recursive calls. This was exceeding python's default recursion limit and was causing some tests to fail. 

### How was it fixed?
Increased the recursion limit to `1024 * 12`. I found this number in [py-evm](https://github.com/ethereum/py-evm/blob/2da165f98cedf08d3a8d21ee92f9de82d0a1baa8/eth/__init__.py#L14-L15). I haven't calculated the limit myself. 

#### Cute Animal Picture

Picture credits: [pinclipart.com](https://www.pinclipart.com/maxpin/bhhTTR/)

### DIFF
--- a/src/ethereum/__init__.py
+++ b/src/ethereum/__init__.py
@@ -4,3 +4,10 @@ Ethereum Specification
 
 Core specifications for Ethereum clients.
 """
+import sys
+
+#
+#  Ensure we can reach 1024 frames of recursion
+#
+EVM_RECURSION_LIMIT = 1024 * 12
+sys.setrecursionlimit(max(EVM_RECURSION_LIMIT, sys.getrecursionlimit()))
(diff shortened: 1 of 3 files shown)


voith:
@SamWilsn Please review this whenever you get time.

SamWilsn (approved):

SamWilsn (commented):
Wait, actually, I don't think this makes sense in the `frontier` part of the spec at all. This is probably common to all the hardforks.

I guess we don't have a well defined entrypoint for the cross hardfork stuff yet. I'll open a ticket.

voith:
@SamWilsn Please take a look at the latest commit. That should close #335. 
As long as we have imports like `from ethereum.<some_module> import something`, this fix should work.

SamWilsn (approved):

PR #338, opened 2021-09-13
Fixed a bug in the selfdestruct refund logic

voith:
### What was wrong?
The refund logic in `selfdestruct` was implemented incorrectly. The current implementation would refund gas every time an account was self-destructed in the same transaction. In fact, the implementation is clearly described in the YP as shown below:

Screenshot 2021-09-13 at 5 34 06 PM

The conversation held in [discord](https://discord.com/channels/595666850260713488/856943705209569291/885164284903063596) helped me identify this bug.

### How was it fixed?
<strike>Made sure that gas is only refunded once for the address being deleted in case `selfdestruct` is called multiple times for the address in the same transaction.</strike>

calculated gas refund for `selfdestruct` at the end of bytecode execution by multiplying the number of unique `accounts to delete` with the refund amount for `selfdestruct`.

PS: sadly, I haven't found any tests that check this logic. 

#### Cute Animal Picture

Pic credits: [pngtree.com](https://pngtree.com/freepng/cute-rabbit_4358746.html)

### DIFF
--- a/src/ethereum/frontier/vm/instructions/system.py
+++ b/src/ethereum/frontier/vm/instructions/system.py
@@ -12,7 +12,6 @@ Introduction
 Implementations of the EVM system related instructions.
 """
 from ethereum.base_types import U256, Uint
-from ethereum.frontier.vm.gas import REFUND_SELF_DESTRUCT
 
 from ...state import get_account, increment_nonce, set_account_balance
 from ...utils.address import compute_contract_address, to_address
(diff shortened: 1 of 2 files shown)


voith:
I converted this PR to draft as I don't think the current fix will work. `CALL` and `CALLCODE` create a [new execution frame every time](https://github.com/ethereum/execution-specs/blob/d2c99d50d03cde4f167470e27f2277984532a802/src/ethereum/frontier/vm/interpreter.py#L138-L171). Hence selfdestruct made by two separate calls might not have the address in `evm.accounts_to_delete`.

I have a workaround in mind which I need to test. Please do not review this PR until then.

voith:
@SamWilsn This is ready for review.

SamWilsn (approved):

PR #345, opened 2021-09-16
[BugFix]Handle collision of addresses and OutofGasError while setting code during contract creation 

voith:
### What was wrong?

Tried to pass tests from `stCreateTest` and found the following problems:
- Collision of Address isn't currently handled
- OutOfGasError raised while setting code during contract creation isn't handled correctly.

### How was it fixed?
Fixed the two problems discussed above in separate commits as described below:
- Handled collision of address during contract creation.  b10bbf1
- Handled `OutOfGasError` raised while setting code during contract creation.  8266384

#### Cute Animal Picture

Pic credits: [pexels.com](https://www.pexels.com/search/cute%20animals/)

### DIFF
--- a/src/ethereum/frontier/state.py
+++ b/src/ethereum/frontier/state.py
@@ -19,7 +19,7 @@ There is a distinction between an account that does not exist and
 from dataclasses import dataclass, field
 from typing import Callable, Dict, List, Optional, Tuple
 
-from ethereum.base_types import U256, Bytes, modify
+from ethereum.base_types import U256, Bytes, Uint, modify
 
 from .eth_types import EMPTY_ACCOUNT, Account, Address, Root
 from .trie import EMPTY_TRIE_ROOT, Trie, copy_trie, root, trie_get, trie_set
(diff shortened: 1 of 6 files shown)


src/ethereum/frontier/state.py
```diff
@@ -293,6 +293,33 @@ def account_exists(state: State, address: Address) -> bool:
...
+        False otherwise.
+    """
+    account = get_account_optional(state, address)
+    if account is None:
+        return False
+    else:
+        if account.nonce != Uint(0) or account.code != b"":
+            return True
+        else:
+            return False
```
SamWilsn, line 320:
```suggestion
    return account is not None and (account.nonce != Uint(0) or account.code != b"")
```

Or

```suggestion
if account is None:
    return False

return account.nonce != Uint(0) or account.code != b""
```

src/ethereum/frontier/vm/instructions/system.py
```diff
@@ -69,6 +70,10 @@ def create(evm: Evm) -> None:
         evm.message.current_target,
         get_account(evm.env.state, evm.message.current_target).nonce - U256(1),
     )
+    is_collision = account_has_code_or_nonce(evm.env.state, contract_address)
+    if is_collision:
+        push(evm.stack, U256(0))
+        return None
```
SamWilsn, line 83:
```suggestion
        return
```

This should be fine, I think?

src/ethereum/frontier/state.py
voith: found a simpler way to do this:
```python
# since we are checking non empty nonce or code, it doesn't matter 
# if `get_account` returns `EMPTY_ACCCONT` when account is not present in state. 
account = get_account(state, address)
return account.nonce != Uint(0) or account.code != b""
```

src/ethereum/frontier/vm/instructions/system.py
voith: Done

SamWilsn (approved):

PR #363, opened 2021-10-05
fix OverflowError in the BLOCKHASH opcode

voith:
### What was wrong?

While Trying to passing the remaining legacy tests, I found a bug in the blockhash opcode.  U256 was overflowing while subtracting `evm.env.number - block_number`.

### How was it fixed?
converted `U256` to `int` before subtracting.

#### Cute Animal Picture

Pic credits: [imgur](https://i.imgur.com/0Dwtce3.jpg)

### DIFF
--- a/src/ethereum/frontier/vm/instructions/block.py
+++ b/src/ethereum/frontier/vm/instructions/block.py
@@ -36,23 +36,17 @@ def block_hash(evm: Evm) -> None:
     OutOfGasError
         If `evm.gas_left` is less than `20`.
     """
-    # TODO: This function hasn't been tested. Need to come up with unit tests
-    # for this function.
     evm.gas_left = subtract_gas(evm.gas_left, GAS_EXTERNAL)
 
     block_number = pop(evm.stack)
-    # The block age (1-indexed) with respect to the current block number.
-    # A block having age of `a` indicates that it is `a` blocks
-    # behind the current block number.
-    block_age = evm.env.number - block_number
 
-    if block_age < 1 or block_age > 256:
+    if evm.env.number <= block_number or evm.env.number > block_number + 256:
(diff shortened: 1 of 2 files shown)


src/ethereum/frontier/vm/instructions/block.py
```diff
@@ -36,15 +36,13 @@ def block_hash(evm: Evm) -> None:
...
 
-    block_number = pop(evm.stack)
+    block_number = int(pop(evm.stack))
     # The block age (1-indexed) with respect to the current block number.
     # A block having age of `a` indicates that it is `a` blocks
     # behind the current block number.
-    block_age = evm.env.number - block_number
+    block_age = int(evm.env.number) - block_number
 
     if block_age < 1 or block_age > 256:
```
SamWilsn, line 47: Would it be possible to write this without negative integers, maybe `if block_number + 1 > evm.env.number`?

voith: can you check the latest commit? Not sure if that complicates things.

SamWilsn (commented):
I think it's more complex, but also skirts the issue of requiring negative integer support.

SamWilsn (approved):

Bhargavasomu:
FYI, I don't think this is a bug. This was left the way it was intended because the `block age` is not supposed to be negative. If it's negative, then it should throw an error (potentially even an overflow error) or gracefully continue by logging the same error. I just left it with an overflow error because we didn't have a mechanism to gracefully log errors.

SamWilsn:
I'm not sure the yellow paper agrees, though I didn't check the frontier version:

> 0 is left on the stack if the looked for block number is greater than or equal to the current block number or more than 256 blocks behind the current block.

PR #365, opened 2021-10-06
Migrate away from assert keyword for non-bugs. (closes #320)

### LINKED ISSUE #320: Be consistent with exceptions and asserts
Currently `assert` is used to mean multiple different things in the spec. It can indicate an impossible situation that means a bug in the spec if it is reached or an invalid block is being processed. We should standardise on a clear set of rules.

I propose `assert` should be use exclusively for impossible situations that indicate a bug. There should be an exception (possibly with subclasses), `Invalid(str)` to indicate that a block, transaction, etc... is invalid. There will probably be a third set of exceptions to deal with reverts in the evm.

SamWilsn:
### What was wrong?

Related to Issue #320 

The assert keyword in python can be disabled with the `-O` flag. Since
many of the assertions in the spec are actually intended to be runtime
checks, using assert doesn't make sense.

### How was it fixed?

This commit adds an `ensure` function, which pretty much works the same
way as assert, but doesn't get optimized out.

#### Cute Animal Picture


src/ethereum_optimized/frontier/state_db.py
```diff
@@ -772,7 +773,7 @@ def make_extension_node(
     This function returns the new value of the visited node, but does not write
     it to the database.
     """
-    assert node_key != target_key
+    ensure(node_key != target_key)
```
ultratwo, line 776: This should remain an assert.

PR #369, opened 2021-10-11
Remove JSON-RPC spec from repo

### LINKED ISSUE #213: Proposal: New Repository and Process for JSON RPC Spec
The JSON RPC specification in this repository has different goals and requires different processes than the core protocol specification. I propose that we move it in a separate repository, similarly to how [eth2.0-specs](https://github.com/ethereum/eth2.0-specs/) and [eth2.0-apis](https://github.com/ethereum/eth2.0-apis) are structured. 

This repository would hence be only used to maintain the set of Core EIPs that each network upgrade consists of, the specification for devnets and potentially an executable version of the Yellowpaper/eth1 spec. 

Moving the JSON RPC spec to a separate repository would also allow us to have a better process to update the spec. As mentioned [here](https://github.com/ethereum/eth1.0-specs/pull/212#issuecomment-861128564) and [here](https://github.com/ethereum/eth1.0-specs/pull/212#issuecomment-861420062), we could move the JSON RPC spec to have branch-based releases.

lightclient:
The spec has been deprecated in this repo since mid June, with the expectation to completely remove it from the repository after the London upgrade. I believe sufficient warning has been given and we can now close #213 with this PR.


timbeiko (approved):
SGTM. How about @alita-moore ?

PR #375, opened 2021-10-15
Sync fixes

ultratwo:
This pull request uses standard JSON RPC calls to get blocks during sync rather than the nonstandard call `debug_getBlockRlp`.

It also fixes various minor bugs necessary than enable the chain to sync to block 1149999.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/frontier/spec.py
+++ b/src/ethereum/frontier/spec.py
@@ -781,9 +781,16 @@ def calculate_block_difficulty(
     if number == 0:
         return GENESIS_DIFFICULTY
     elif timestamp < parent_timestamp + 13:
-        return parent_difficulty + max_adjustment_delta
+        difficulty = parent_difficulty + max_adjustment_delta
     else:  # timestamp >= parent_timestamp + 13
-        return max(
-            GENESIS_DIFFICULTY,
-            parent_difficulty - max_adjustment_delta,
-        )
+        difficulty = parent_difficulty - max_adjustment_delta
+
+    # Historical Note: The difficulty bomb was not present in Ethereum at the
+    # start of Frontier, but was added shortly after launch. However since the
+    # bomb has no effect prior to block 200000 we pretend it existed from
+    # genesis.
+    # See https://github.com/ethereum/go-ethereum/pull/1588
+    num_bomb_periods = int(number) // 100000 - 2
+    if num_bomb_periods >= 0:
+        difficulty += 2 ** num_bomb_periods
+    return Uint(max(difficulty, GENESIS_DIFFICULTY))
--- a/src/ethereum/frontier/spec.py
+++ b/src/ethereum/frontier/spec.py
@@ -193,8 +193,8 @@ def state_transition(chain: BlockChain, block: Block) -> None:
 
     ensure(gas_used == block.header.gas_used)
     ensure(transactions_root == block.header.transactions_root)
-    ensure(receipt_root == block.header.receipt_root)
     ensure(state_root(state) == block.header.state_root)
+    ensure(receipt_root == block.header.receipt_root)
     ensure(block_logs_bloom == block.header.bloom)
 
     chain.blocks.append(block)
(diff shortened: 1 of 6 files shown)


src/ethereum_spec_tools/sync.py
```diff
@@ -321,12 +414,36 @@ def fetch_blocks(
                     raise Exception("mismatched request id")
 
                 if "error" in reply:
-                    blocks[reply_id] = RpcError(
+                    raise RpcError(
```
SamWilsn, line 417: Same concern about aborting early.

src/ethereum_spec_tools/sync.py
```diff
@@ -321,12 +414,36 @@ def fetch_blocks(
...
+                                if t["to"]
+                                else Bytes0(b""),
+                                hex_to_u256(t["value"]),
+                                hex_to_bytes(t["input"]),
+                                hex_to_u256(t["v"]),
+                                hex_to_u256(t["r"]),
+                                hex_to_u256(t["s"]),
+                            )
+                        )
+                    uncles = self.fetch_uncles(reply_id, len(res["uncles"]))
```
SamWilsn, line 443: Oof, so now we have one extra request per block? Any way we can collect all the uncle fetches into one giant request, so we only have two requests (one to get blocks, and one to get all their uncles)?

src/ethereum/frontier/spec.py
```diff
@@ -781,9 +781,18 @@ def calculate_block_difficulty(
...
+    # start of Frontier, but was added shortly after launch. However since the
+    # bomb has no effect prior to block 200000 we pretend it existed from
+    # genesis.
+    num_bomb_periods = int(number) // 100000 - 2
+    if num_bomb_periods >= 0:
+        return Uint(
+            max(difficulty + 2 ** num_bomb_periods, GENESIS_DIFFICULTY)
         )
+    else:
+        return Uint(max(difficulty, GENESIS_DIFFICULTY))
```
lightclient, line 798: What about something like this:

```python
if num_bomb_periods >= 0:
    difficulty + 2 ** num_bomb_periods

return Uint(max(difficulty, GENESIS_DIFFICULTY))
```

src/ethereum_spec_tools/sync.py
lightclient: One extra call for *every uncle* in each block :D

src/ethereum_spec_tools/sync.py
```diff
@@ -273,14 +282,98 @@ def module(self, name: str) -> Any:
...
+            Bytes256(hex_to_bytes(json["logsBloom"])),
+            hex_to_uint(json["difficulty"]),
+            hex_to_uint(json["number"]),
+            hex_to_uint(json["gasLimit"]),
+            hex_to_uint(json["gasUsed"]),
+            hex_to_u256(json["timestamp"]),
+            hex_to_bytes(json["extraData"]),
+            hex_to_bytes32(json["mixHash"]),
+            hex_to_bytes8(json["nonce"]),
+        )
```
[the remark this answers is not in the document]
ultratwo: Yes, London is the first time the header format changes. We will have to add a conditional when London is implemented. The same problem will happen with transactions in Berlin and London (due to new types being added).

ultratwo:
I have resolved the issues mentioned above and added two further fixes to get to the end of Frontier.

src/ethereum/frontier/vm/instructions/stack.py
```diff
@@ -64,7 +65,6 @@ def push_n(evm: Evm, num_bytes: int) -> None:
         If `evm.gas_left` is less than `3`.
     """
     evm.gas_left = subtract_gas(evm.gas_left, GAS_VERY_LOW)
-    ensure(evm.pc + num_bytes < len(evm.code))
```
SamWilsn, line 67: Huh, this is surprising. I even checked it on Remix. Good catch.

PR #378, opened 2021-10-20
Only store most recent 255 blocks in the Blockchain object

petertdavies:
This patch makes the spec forget blocks older than 255 blocks ago. This is all that is required in order to implement the EVM. This saves about 5GB of memory usage by the end of Frontier.

I experimented with the `deque` object for this, but the complexity it introduced wasn't worth it.

#### Cute Animal Picture


ultratwo:
> s/255/256?

255. We need the block hash for 256th most recent block, but we get it by looking at the parent of the 255th most recent block.

PR #392, opened 2021-11-13
Add semantic diffs between hardforks to the docs.

SamWilsn:
### What was wrong?

No easy way to see changes between hard forks.

### How was it fixed?

Using a modified version of `rstdiff` from the docutils package.

Won't really show much until a new hard fork is added, but I've done some limited testing with fake forks locally.

Couple known issues:

 - Links inside the diffed documents point to `.pickle` files instead of `.html`, which don't exist.
 - Very noisy diff since anywhere the module name appears, there's a change reported.
 - No quick navigation to the diffs.

Here's what an example change might look like so far: https://samwilsn.github.io/eth1.0-specs/diffs/frontier_banana/rlp/index.html

#### Cute Animal Picture

@voith would you also mind taking a look?

### DIFF
--- a/src/ethereum_spec_tools/diff.py
+++ b/src/ethereum_spec_tools/diff.py
@@ -1,0 +1,161 @@
+#!/usr/bin/env python3
+
+"""
+Generates diffs between Ethereum hardforks documentation.
+"""
+
+import os.path
+import pickle
+from typing import Iterator, List, TypeVar
+
+import rstdiff
+from docutils import SettingsSpec
+from docutils.utils import new_document, new_reporter
+
+from ethereum_spec_tools.forks import Hardfork
+
+T = TypeVar("T")
+
+
+def window(forks: List[T], window_size: int = 2) -> Iterator[List[T]]:
+    """
+    Group a list into overlapping chunks.
+    """
+    for i in range(len(forks) - window_size + 1):
+        yield forks[i : i + window_size]
+
+
+def find_pickles(path: str, fork: Hardfork) -> Iterator[str]:
+    """
+    Find files ending with :code:`.pickle`.
+    """
+    for directory, _, filenames in os.walk(path):
+        for filename in filenames:
+            if filename.endswith(".pickle"):
+                file_path = os.path.join(directory, filename)
+                file_path = file_path[len(path) + 1 :]
+                yield file_path
+
+
+def diff(
+    output_path: str, input_path: str, old: Hardfork, new: Hardfork
+) -> Iterator[str]:
+    """
+    Calculate the structured diff between two hardforks.
+    """
+    old_path = old.name.replace(".", os.sep)
+    old_path = os.path.join(input_path, old_path)
+
+    new_path = new.name.replace(".", os.sep)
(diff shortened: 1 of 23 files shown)


src/ethereum_spec_tools/diff.py
```diff
@@ -0,0 +1,161 @@
...
+    old_path = old.name.replace(".", os.sep)
+    old_path = os.path.join(input_path, old_path)
+
+    new_path = new.name.replace(".", os.sep)
+    new_path = os.path.join(input_path, new_path)
+
+    diff_path = old.name.split(".")[-1] + "_" + new.name.split(".")[-1]
+    diff_path = os.path.join(output_path, diff_path)
+
+    old_pickles = set(find_pickles(old_path, old))
```
lightclient, line 55: Just curious, why return an iterator and create a set vs. simply returning a set? More flexibility?

src/ethereum_spec_tools/diff.py
```diff
@@ -0,0 +1,161 @@
...
+    new_pickles = set(find_pickles(new_path, new))
+
+    pickles = old_pickles | new_pickles
+
+    for count, pickle_path in enumerate(pickles, start=1):
+        old_pickle_path = os.path.join(old_path, pickle_path)
+        new_pickle_path = os.path.join(new_path, pickle_path)
+        diff_pickle_path = os.path.join(diff_path, pickle_path + "64")
```
lightclient, line 63: Why `64`?

lightclient (approved):
Looks good to me.

whitelist.txt
```diff
@@ -274,3 +274,12 @@ noop
 0x0
 0x1
 0x2
+
+rstdiff
+docutils
+dirname
+Text2
```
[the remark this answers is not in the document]
SamWilsn: Yeppers. Probably worth doing some work upstream to improve it a bit.

src/ethereum_spec_tools/diff.py
SamWilsn: Didn't really think much about it. Seems a bit clunky to write:

```python
x = set()
for y in z:
    x.append(x)
return x
```

src/ethereum_spec_tools/diff.py
SamWilsn: The build process has two stages, `stage0` and `stage1`.

`stage0` writes to binary pickle files, and those files are opened by `rstdiff`, which is fine handling binary files.

Sphinx, on the other hand, only lets you define parsers in extensions and not readers. Parsers can only handle strings (not `bytes`) so the input has to be valid UTF-8.

The output from the `rstdiff` is written to `.pickle64` files (base64 encoded pickle) to satisfy that requirement.

lightclient: ahh okay gotcha!

src/ethereum_spec_tools/diff.py
lightclient: okay that makes sense, you're version is a lot more terse

PR #395, opened 2021-11-16
Remove newline from description field in setup.cfg

ultratwo:
### What was wrong?

There is a newline in the description field in setup.cfg. This is [illegal](https://github.com/pypa/setuptools/issues/1390), but was only caught by python buildtools in [after a recent patch](https://github.com/pypa/setuptools/pull/2870).

### How was it fixed?

Removed the newline.

#### Cute Animal Picture


lightclient (commented):
@ultratwo this is also addressed in #394

PR #398, opened 2021-11-17
Move RLP out of Frontier & Homestead

ultratwo:
This pull request moves the RLP logic out of Frontier. Rather than knowing about specific datatypes, the new `encode()` and `decode_to()` function takes a type and does introspection on it to work out how to do the encoding/decoding.

This is a bit hacky for various reasons. The types of annotations aren't part of the public interface in typing (but will be in Python 3.10). The type `Union[Bytes0, Address]` has to be special cased.

I haven't changed any code dealing with internal trie nodes. That code is important for performance and the `decode_to()` will be slower than how it currently works.

Please can someone test whether `ethereum-spec-sync --geth` works up to block 47000 (when the first transactions appear), before pulling this.

#### Cute Animal Picture


SamWilsn:
Added a few commits:

 - Applied to homestead
 - Fixed sync tool's `--geth` mode

@ultratwo I'm able to sync to block ~61000 before getting a bad state root. Happens with and without `--geth`.

PR #405, opened 2021-12-07
Add --start option

ultratwo:
This pull request adds a `--start` option that allows doing a "fast-sync" from a block of your choice. This requires an archive node. It has only been tested against Geth, Erigon's architecture cannot serve the required queries efficiently.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/sync.py
+++ b/src/ethereum_spec_tools/sync.py
@@ -7,26 +7,12 @@ Using an RPC provider, fetch each block and validate it with the specification.
 
 import argparse
 import base64
-import copyreg
-import dataclasses
 import json
 import logging
-import os.path
-import pickle
 import time
 from queue import Empty, Full, Queue
 from threading import Thread
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    List,
-    Optional,
-    Tuple,
-    Type,
-    TypeVar,
-    Union,
-)
+from typing import Any, Dict, List, Optional, TypeVar, Union
 from urllib import request
 
 from ethereum import rlp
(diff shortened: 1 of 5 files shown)


ultratwo:
I've added code to remove the `blocks.pickle` file and instead put the block number in the db. This updates the db version.

src/ethereum_spec_tools/sync.py
```diff
@@ -179,70 +189,54 @@ def __init__(self) -> None:
...
-            except FileNotFoundError as e:
-                raise Exception("found blocks file but no state") from e
-
-        self.chain = self.module("spec").BlockChain(
-            blocks=blocks,
-            state=state,
-        )
-
-        # TODO: Fast forward to correct hard fork.
+        if persisted_block is not None:
```
SamWilsn, line 216: Would you mind flipping the order of these conditions (so `if persisted_block is None` comes first)? I like avoiding extra negation if possible.

src/ethereum_spec_tools/sync.py
```diff
@@ -255,25 +249,9 @@ def persist(self) -> None:
 
         start = time.monotonic()
 
-        temp_path = os.path.join(self.options.persist, "blocks.pickle.temp")
-        blocks_path = os.path.join(self.options.persist, "blocks.pickle")
-
-        with open(temp_path, "wb") as f:
-            DataclassPickler(f).dump(self.chain.blocks)
```
SamWilsn, line 262: Can we remove the all the `DataclassPickler` machinery too?

SamWilsn (approved):

Issue #387, opened 2021-11-01
Parent hash isn't validated


SamWilsn:
### Metadata

* **Hardfork**: frontier

### What was wrong?

The spec doesn't ensure that the `parent_hash` of the current block matches the hash of the previous block.

### Sources

> **parentHash**: The Keccak 256-bit hash of the parent block’s header, in its entirety; formally H<sub>p</sub>.

[source](https://ethereum.github.io/yellowpaper/paper.pdf#subsection.4.3)

gurukamath:
Does this also mean additional test will have to be added, since the existing tests weren’t able to identify the issue? (I see all existing tests pass)

SamWilsn:
@gurukamath adding a python test would be great!

gurukamath:
@SamWilsn Added the relevant asserts to the state transition tests

PR #414, opened 2022-01-08
test parent hash validation

gurukamath:
### What was wrong?
Test to validate the parent hash of a block was missing

Related to Issue #387

### How was it fixed?
Added an additional assert while testing state transition (frontier and homestead)

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/frontier/spec.py
+++ b/src/ethereum/frontier/spec.py
@@ -218,6 +218,9 @@ def validate_header(header: Header, parent_header: Header) -> None:
         parent_header.difficulty,
     )
 
+    block_parent_hash = crypto.keccak256(rlp.encode(parent_header))
+
+    ensure(header.parent_hash == block_parent_hash)
     ensure(header.difficulty == block_difficulty)
     ensure(header.number == parent_header.number + 1)
     ensure(check_gas_limit(header.gas_limit, parent_header.gas_limit))
--- a/src/ethereum/homestead/spec.py
+++ b/src/ethereum/homestead/spec.py
@@ -188,6 +188,9 @@ def validate_header(header: Header, parent_header: Header) -> None:
         parent_header.difficulty,
     )
 
+    block_parent_hash = crypto.keccak256(rlp.encode(parent_header))
+
+    ensure(header.parent_hash == block_parent_hash)
     ensure(header.difficulty == block_difficulty)
     ensure(header.number == parent_header.number + 1)
     ensure(check_gas_limit(header.gas_limit, parent_header.gas_limit))


tests/frontier/blockchain_st_test_helpers.py
```diff
@@ -76,6 +76,13 @@ def run_blockchain_st_test(
 
 def add_blocks_to_chain(chain: BlockChain, test_data: Dict[str, Any]) -> None:
     for idx, block in enumerate(test_data["blocks"]):
+        if idx == 0:
```
voith, line 79: These assert statements basically check that the test data is correct. This doesn't ensure that the `spec.py` actually validates the parent hash.
What we should be doing is writing a unit test that provides an invalid parent hash and checks that test reverts with a validation error.

tests/frontier/blockchain_st_test_helpers.py
```diff
@@ -76,6 +76,13 @@ def run_blockchain_st_test(
...
+            assert test_data["genesis_header_hash"] == block.header.parent_hash
```
voith, line 80: These assert statements basically check that the test data is correct. This doesn't ensure that the `spec.py` actually validates the parent hash.
What we should be doing is writing a unit test that provides an invalid parent hash and checks that test reverts with a validation error.

gurukamath: @voith Thank you for the input. I have a few questions here.

I see that the invalid blocks test cases are already present under tests/fixtures/LegacyTests/Constantinople/BlockchainTests/InvalidBlocks/bcInvalidHeaderTest/

I don't think any of these test cases are implemented yet (at least I couldn't find the implementations). Is that right? If I've missed it, could you please point me to the implementation?

If they have not been implemented yet, please suggest the file where adding these tests would be most appropriate. (test_state_transition.py? or maybe a new file?)

voith: > I see that the invalid blocks test cases are already present under tests/fixtures/LegacyTests/Constantinople/BlockchainTests/InvalidBlocks/bcInvalidHeaderTest/


Some of those tests were added sometime back and were accidentally removed by me. Please refer to #415 for more context. You can add them back by referring to the commit I've posted in that issue. 
However, only [bcUncleTest](https://github.com/ethereum/tests/tree/82f6a5a7362e703a64dce303dd0c30f264b8195f/LegacyTests/Constantinople/BlockchainTests/InvalidBlocks/bcUncleTest)  and [bcUncleHeaderValidity](https://github.com/ethereum/tests/tree/82f6a5a7362e703a64dce303dd0c30f264b8195f/LegacyTests/Constantinople/BlockchainTests/InvalidBlocks/bcUncleHeaderValidity) were tested earlier. You probably need [bcInvalidHeaderTest](https://github.com/ethereum/tests/tree/82f6a5a7362e703a64dce303dd0c30f264b8195f/LegacyTests/Constantinople/BlockchainTests/InvalidBlocks/bcInvalidHeaderTest) in order to close #387

gurukamath: @voith Accidentally happened to remove the repo from which I had raised the PR. Have opened a new PR with updates to include the relevant unit tests. Please review

https://github.com/ethereum/execution-specs/pull/416

PR #421, opened 2022-01-12
Add a PREV_FORK variable to each fork use it in trie.py

petertdavies:
This pull request adds a `PREV_FORK` variable to each fork and uses it to refer back to the previous fork in `trie.py`. As a result the `trie.py` file is now identical in each fork.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/dao_fork/__init__.py
+++ b/src/ethereum/dao_fork/__init__.py
@@ -6,3 +6,4 @@ The third Ethereum hardfork.
 """
 
 MAINNET_FORK_BLOCK = 1920000
+PREV_FORK = "homestead"
--- a/src/ethereum/dao_fork/trie.py
+++ b/src/ethereum/dao_fork/trie.py
@@ -15,7 +15,9 @@ The state trie is the structure responsible for storing
 
 import copy
 from dataclasses import dataclass, field
+from importlib import import_module
 from typing import (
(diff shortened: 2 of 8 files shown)


src/ethereum/dao_fork/__init__.py
```diff
@@ -6,3 +6,4 @@
 """
 
 MAINNET_FORK_BLOCK = 1920000
+PREV_FORK = "homestead"
```
SamWilsn, line 9: Would something like:

```python
from ethereum import homestead

PREVIOUS_FORK = homestead
```

work?

SamWilsn (commented):
Hm, this is an interesting idea. It certainly cleans up the diffs, but adds some complexity to the implementation itself.

Before this PR, each fork has at least two important changes in different places (`MAINNET_FORK_BLOCK` and the previous fork in the trie code), and after this PR those changes will be located in the same place.

If we can simplify the implementation of this feature, I think it'll be worthwhile, but right now I think it adds too much complexity.

src/ethereum/dao_fork/__init__.py
ultratwo: No, we have to import `ethereum.FORK.trie` (as opposed to just `ethereum.FORK`) somewhere.

Issue #401, opened 2021-11-26
EIP-2681: Limit account nonce to 2^64-1 


axic:
[EIP-2681](https://eips.ethereum.org/EIPS/eip-2681) applies from genesis. It was properly merged into Geth this week and also the state tests repository has it merged.

EmreKspToken:
EmreKspcoin

EmreKspToken:
0xea3779dc2d94f40cdbc9cc953eb4f1a7a7538975

gurukamath:
Which test cases would one use to build this? I am unable to find any JSON under tests that mentions this issue.

SamWilsn:
> Which test cases would one use to build this? I am unable to find any JSON under tests that mentions this issue.

They can be found in the upstream repository: https://github.com/ethereum/tests/tree/develop/TransactionTests/ttNonce

The `fixtures` directory in [https://github.com/ethereum/execution-specs](execution-specs) is a submodule, and needs to be updated to get the latest tests.

gurukamath:
@SamWilsn Thanks.

Also for the other change mentioned in [EIP-2681](https://eips.ethereum.org/EIPS/eip-2681)
"The CREATE and CREATE2 instructions to abort with an exceptional halt, where the account nonce is 2^64-1", I assume the relevant test is [High Nonce Test](https://github.com/ethereum/tests/blob/develop/GeneralStateTests/stCreateTest/CREATE_HighNonce.json).

However, I see that the structure of the above JSON is quite different from the fixtures have been tested so far for the VM System Operations (I looked at [suicide0](https://github.com/ethereum/tests/blob/82f6a5a7362e703a64dce303dd0c30f264b8195f/LegacyTests/Constantinople/VMTests/vmSystemOperations/suicide0.json) as a reference). For example, the "exec" key from the old JSON is missing in the new one (perhaps replaced by "transaction" based on a quick glance).

Is there any reference that could help me load the new format of tests? Or at least a dictionary that translates the old keys in JSON to the new one?

lightclient:
[...]

PR #422, opened 2022-01-18
EIP-2681: Limit account nonce to 2^64-1

gurukamath:
### What was wrong?

Related to Issue #401 

### How was it fixed?
1. Updated the fixtures module to the latest commit
2. Updated GitHub test.yml to fetch submodules recursively, since LegacyTests itself is now a submodule in the latest commit of the fixtures submodule
3. Implemented the two conditions specified in [EIP-2681](https://eips.ethereum.org/EIPS/eip-2681). Updated the CREATE instruction and validate_transaction function
4. Added transaction tests for high nonce to all the hard forks where tests have been implemented

#### Cute Animal Picture

### DIFF
--- a/tests/frontier/blockchain_st_test_helpers.py
+++ b/tests/frontier/blockchain_st_test_helpers.py
@@ -207,24 +207,6 @@ def json_to_state(raw: Any) -> State:
     return state
 
 
-def load_test_transaction(
-    test_dir: str, test_file: str, network: str
-) -> Dict[str, Any]:
-    pure_test_file = os.path.basename(test_file)
-    test_name = os.path.splitext(pure_test_file)[0]
-    path = os.path.join(test_dir, test_file)
-
-    with open(path, "r") as fp:
-        json_data = json.load(fp)[f"{test_name}"]
-
-    tx_rlp = hex_to_bytes(json_data["txbytes"])
-    tx = rlp.decode_to(Transaction, tx_rlp)
-
-    test_result = json_data["result"][network]
-
-    return {"transaction": tx, "test_result": test_result}
-
-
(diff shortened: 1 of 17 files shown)


tests/frontier/blockchain_st_test_helpers.py
```diff
@@ -207,6 +207,24 @@ def json_to_state(raw: Any) -> State:
     return state
 
 
+def load_test_transaction(
+    test_dir: str, test_file: str, network: str
```
SamWilsn, line 211: Since this is in the `tests/frontier` directory, this function will only ever be called for the `Frontier` network.

Would it make sense to make a file directly under `tests/`, deduplicate these functions there, and import it in each fork?

SamWilsn:
@gurukamath you should join us for our weekly sync calls! We're in the `#specifications` channel in the Eth R&D discord.

tests/frontier/blockchain_st_test_helpers.py
gurukamath: Yes. That would make sense. I imagine there will be other such functions going forward, I will create a helpers package under `tests/` to house such functions.

gurukamath: @SamWilsn On second thought, the function does use the `Transaction` class which is specific to a particular hard-fork. We could still create generic helpers outside the hard-fork specific test packages, but we will have to implement that with dynamic module imports. What do you think?

gurukamath: @SamWilsn Have updated the PR

SamWilsn (approved):

Issue #418, opened 2022-01-12
`ethereum-spec-lint` should prevent absolute imports within forks


SamWilsn:
To help maintain clean diffs between hard fork modules, `ethereum-spec-lint` needs to detect absolute imports of things within hard fork modules.

The rules when inside a hard fork file:

 - **Deny** absolute imports from within the active fork.
 - **Deny** absolute imports from future forks.
 - **Deny** absolute imports from active-minus-two and earlier hard forks.
 - **Allow** relative imports from the active hard fork.
 - **Allow** absolute imports from the active-minus-one hard fork.
 - **Allow** absolute imports of non-fork specific things.

For example, in `src/ethereum/dao_fork/foo.py`, the linter should:

```python
import ethereum.dao_fork.spec                    # DENY
from ethereum.dao_fork.spec import Foo           # DENY

import ethereum.tangerine_whistle.spec           # DENY
from ethereum.tangerine_whistle.spec import Foo  # DENY

import ethereum.frontier.spec                    # DENY
from ethereum.frontier.spec import Foo           # DENY

import .spec                                     # ALLOW
from .spec import Foo                            # ALLOW

import ethereum.homestead.spec                   # ALLOW
from ethereum.homestead.spec import Foo          # ALLOW

import ethereum.crypto                           # ALLOW
from ethereum.crypto import Foo                  # ALLOW
```

Note that the imports can appear anywhere (eg. inside functions) and not just at the top of files.

PR #446, opened 2022-02-04
ethereum-spec-lint to prevent absolute imports within forks

gurukamath:
### What was wrong?
There were no existing checks in `ethereum-spec-lint` for controlling the hygiene of import statements

Related to Issue #418 

### How was it fixed?
Added `check_import` to enforce rules outlined in #418 

Additionally extended the existing linting such that, the linting is performed recursively. E.g:- individual modules in `vm.instructions` were not being checked earlier.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/lint.py
+++ b/src/ethereum_spec_tools/lint.py
@@ -9,11 +9,11 @@ import ast
 import importlib
 import inspect
 from abc import ABCMeta, abstractmethod
-from collections import OrderedDict
 from dataclasses import dataclass
 from typing import Generator, List, Optional, Sequence, Tuple
 
 from .forks import Hardfork
+from .visitors import ImportHygieneVisitor, PatchHygieneVisitor
 
 
 def walk_sources(fork: Hardfork) -> Generator[Tuple[str, str], None, None]:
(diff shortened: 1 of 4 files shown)


src/ethereum_spec_tools/lint.py
```diff
@@ -151,20 +177,25 @@ def lint(
...
-            (k, v, all_previous.get(k, None)) for (k, v) in all_current.items()
-        )
-
         diagnostics: List[Diagnostic] = []
-        for (name, current, previous) in items:
-            diagnostics += self.compare(name, current, previous)
+        if position == 0:
+            for (name, source) in all_current.items():
+                # No need to run compare since this is the first fork
+                diagnostics += self.check_import(forks, position, name, source)
```
SamWilsn, line 185: I think this feature might make more sense as a separate `Lint` subclass with its own visitor?

src/ethereum_spec_tools/lint.py
```diff
@@ -92,14 +94,26 @@ def visit_AsyncFunctionDef(self, function: ast.AsyncFunctionDef) -> None:
...
         """
         self._insert(function.name)
-        # Explicitly don't visit the children of functions.
+
+        # Detect imports inside function body
+        for node in function.body:
+            if isinstance(node, ast.ImportFrom):
+                self.item_imports.append(str(node.module))
+            if isinstance(node, ast.Import):
+                self.item_imports.append(node.names[0].name)
```
SamWilsn, line 116: I think you'll need to actually visit the children of functions, because I don't think the current implementation would catch:

```python
def foo():
    if True:
        import ethereum.some_fork.spec
```

src/ethereum_spec_tools/lint.py
```diff
@@ -205,15 +237,67 @@ def compare(
...
+        """
+        diagnostics: List[Diagnostic] = []
+
+        invalid_imports = {
+            "active_fork": forks[position].name,
+            "future_fork": [fork.name for fork in forks[position + 1 :]],
+            "minus2_fork": [fork.name for fork in forks[: position - 1]]
+            if position > 1
+            else [],
+        }
```
SamWilsn, line 255: Couple suggestions here:
 - I don't think the dictionary is necessary?
 - Can directly create the tuples here.
 - Python's array slicing syntax should work without the conditional (but I haven't tried, and could be very wrong.)

```suggestion
    active_fork = forks[position].name
    future_forks = tuple(fork.name for fork in forks[position + 1 :])
    ancient_forks = tuple(fork.name for fork in forks[: position - 1])
```

gurukamath: The conditional is required, since the index will be -1 for `frontier` and so, every fork until the latest one will be selected

src/ethereum_spec_tools/lint.py
gurukamath: Have implemented the change. Skipped implementing the `visit_FunctionDef` so it visits the child nodes

src/ethereum_spec_tools/lint.py
gurukamath: Have implemented the `ImportHygiene` sub-class of `Lint` with its own visitor `ImportHygieneVisitor`.
Restored `PatchHygiene` and `PatchHygieneVisitor` to their original form.

The `lint` module was starting to get unwieldy. So I just separated all the visitors out into a different module.

gurukamath:
@SamWilsn Have included the relevant updates. See the comments on the individual items above.

SamWilsn (approved):

PR #447, opened 2022-02-04
Add CFI EIPs

timbeiko:
On ACD131, we agreed to move EIPs 3540, 3670 and 3860 to CFI.

### DIFF
--- a/network-upgrades/mainnet-upgrades/shanghai.md
+++ b/network-upgrades/mainnet-upgrades/shanghai.md
@@ -8,6 +8,7 @@ Specifies changes potentially included in the Network Upgrade, pending successfu
 
 * [EIP-3540: EVM Object Format (EOF) v1](https://eips.ethereum.org/EIPS/eip-3540)
 * [EIP-3670: EOF - Code Validation](https://eips.ethereum.org/EIPS/eip-3670)
+* [EIP-3855: PUSH0 instruction](https://eips.ethereum.org/EIPS/eip-3855)
 * [EIP-3860: Limit and meter initcode](https://eips.ethereum.org/EIPS/eip-3860)
 
 ### Readiness Checklist
--- a/network-upgrades/mainnet-upgrades/shanghai.md
+++ b/network-upgrades/mainnet-upgrades/shanghai.md
@@ -3,6 +3,12 @@
 ### Included EIPs
 Specifies changes included in the Network Upgrade.
 
+### EIPs Considered for Inclusion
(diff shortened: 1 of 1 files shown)


network-upgrades/mainnet-upgrades/shanghai.md
```diff
@@ -3,6 +3,12 @@
 ### Included EIPs
 Specifies changes included in the Network Upgrade.
 
+### EIPs Considered for Inclusion
+Specifies changes potentially included in the Network Upgrade, pending successful deployment on Client Integration Testnets.
+
+* [EIP-3540: EVM Object Format (EOF) v1](https://eips.ethereum.org/EIPS/eip-3540)
+* [EIP-3670: EOF - Code Validation](https://eips.ethereum.org/EIPS/eip-3670)
+* [EIP-3860: Limit and meter initcode](https://eips.ethereum.org/EIPS/eip-3860)
```
axic, line 12: I think also 3855 was part of that, at least some client devs referred to it, but would need to listen to the recording for confirmation.

timbeiko: Right, I am actually unclear about this one.

timbeiko: @holiman @MariusVanDerWijden @tkstanczak @yperbasis @shemnon @garyschulte if we could get a 👍 / 👎  from each of your teams on including PUSH0, it'd be helpful so we don't have to wait until the next call.

shemnon: :+1: Implementing PUSH0 is easy, the harder (but still easy) part is ensuring it doesn't leak into pre-Shanghai execution (a well trodden path with ops like chainid). 

The real work is in reference tests and upstream use by solidity and friends.

That extends to all EVM work, good reference tests are more work than client implementations typically.

axic: For the record, Ipsilon has tests for all these EIPs: https://github.com/ethereum/tests/pulls/gumb0

yperbasis: 👍 for [EIP-3855](https://eips.ethereum.org/EIPS/eip-3855) from Erigon.

MariusVanDerWijden: :+1: from geth on PUSH0

timbeiko: Thanks all for the rapid responses! Will add and merge.

PR #448, opened 2022-02-04
Kintsugi Incident Report

timbeiko:
@parithosh made this part of the specs repo to archive it. Does that work with you?


parithosh:
thanks, this looks good to me!

PR #450, opened 2022-02-05
Refactor lint

SamWilsn:
### What was wrong?

The linter code was getting a little unruly.

### How was it fixed?

Refactored the lints into plugins inside a namespace package. Each lint should now be a (mostly) self contained subclass of `Lint` under the `ethereum_spec_tools.lint.lints` package.

#### Cute Animal Picture


SamWilsn:
@gurukamath since you've been in this area, I wouldn't mind a code review when you have the time!

gurukamath:
@SamWilsn Looks good to me. Just one additional point. Would there be a case for standardising the `parse` function and moving it into the parent `Lint` class so that the code becomes just a little bit DRYer?

PR #451, opened 2022-02-07
Add implementation statuses for Shanghai

axic:


axic:
@timbeiko can you take a look please?

PR #453, opened 2022-02-09
Move lint parse code into super class

### LINKED ISSUE #452: Investigate moving lint parse code into super class
Would there be a case for standardising the `parse` function and moving it into the parent `Lint` class so that the code becomes just a little bit DRYer? It's essentially going to be the same code for all sub-classes with a different visitor passed in.

_Originally posted by @gurukamath in https://github.com/ethereum/execution-specs/issues/450#issuecomment-1031128339_

gurukamath:
(closes #452 )

### What was wrong?
`parse` function was being defined by every sub-class of `Lint`. This is not needed since it is essentially the same function with different `Visitors`

Related to Issue #452 

### How was it fixed?
Move the `parse` function into the `Lint` class and pass the `Visitor` instance as an argument

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/lint/__init__.py
+++ b/src/ethereum_spec_tools/lint/__init__.py
@@ -5,6 +5,7 @@ Lints
 Checks specific to the Ethereum specification source code.
 """
 
+import ast
 import importlib
 import inspect
 import pkgutil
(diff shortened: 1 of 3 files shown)


src/ethereum_spec_tools/lint/__init__.py
```diff
@@ -57,6 +58,16 @@ def lint(
             The particular hardfork to lint.
         """
 
+    def parse(
+        self, source: str, visitor: ast.NodeVisitor, attr: str
```
SamWilsn, line 62: Should probably call it `_parse` to indicate that it's an internal method.

SamWilsn (commented):
Looks good otherwise!

PR #454, opened 2022-02-10
EIP-2681: Limit account nonce to 2^64-1 - Create Test Cases

### LINKED ISSUE #401: EIP-2681: Limit account nonce to 2^64-1

gurukamath:
(closes #401 )

### What was wrong?
The Non-Legacy tests were not implemented.
These tests include ones for exceptional halting of CREATE opcode when the nonce is high

Related to Issue #401 

### How was it fixed?
Added the High Nonce tests out of `tests/fixtures/BlockchainTests/GeneralStateTests/`

#### Cute Animal Picture

### DIFF
--- a/tests/frontier/blockchain_st_test_helpers.py
+++ b/tests/frontier/blockchain_st_test_helpers.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 from functools import partial
 from typing import Any, Dict, List, Tuple, cast
 from unittest.mock import call, patch
@@ -90,7 +91,19 @@ def load_test(test_dir: str, test_file: str, network: str) -> Dict[str, Any]:
     test_name = os.path.splitext(pure_test_file)[0]
     path = os.path.join(test_dir, test_file)
     with open(path, "r") as fp:
-        json_data = json.load(fp)[f"{test_name}_{network}"]
+        data = json.load(fp)
+
+        # Some newer test files have for example _d0g0v0_
+        # between test_name and network
+        keys_to_search = re.compile(
+            f"{test_name}_d[0-9]g[0-9]v[0-9]_{network}"
+        )
+        found_keys = list(filter(keys_to_search.match, data.keys()))
+
+        if len(found_keys) > 0:
+            json_data = data[found_keys[0]]
+        else:
+            json_data = data[f"{test_name}_{network}"]
 
     blocks, block_header_hashes, block_rlps = json_to_blocks(
         json_data["blocks"]
--- a/tests/frontier/blockchain_st_test_helpers.py
+++ b/tests/frontier/blockchain_st_test_helpers.py
@@ -93,17 +93,17 @@ def load_test(test_dir: str, test_file: str, network: str) -> Dict[str, Any]:
     with open(path, "r") as fp:
         data = json.load(fp)
 
-        # Some newer test files have for example _d0g0v0_
+        # Some newer test files have patterns like _d0g0v0_
         # between test_name and network
         keys_to_search = re.compile(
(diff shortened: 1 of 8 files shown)


tests/frontier/blockchain_st_test_helpers.py
```diff
@@ -90,7 +91,19 @@ def load_test(test_dir: str, test_file: str, network: str) -> Dict[str, Any]:
     test_name = os.path.splitext(pure_test_file)[0]
     path = os.path.join(test_dir, test_file)
     with open(path, "r") as fp:
-        json_data = json.load(fp)[f"{test_name}_{network}"]
+        data = json.load(fp)
+
+        # Some newer test files have for example _d0g0v0_
+        # between test_name and network
+        keys_to_search = re.compile(
+            f"{test_name}_d[0-9]g[0-9]v[0-9]_{network}"
```
SamWilsn, line 99: Would it be a good idea to [`re.escape`](https://docs.python.org/3/library/re.html#re.escape) these?

tests/frontier/blockchain_st_test_helpers.py
```diff
@@ -90,7 +91,19 @@ def load_test(test_dir: str, test_file: str, network: str) -> Dict[str, Any]:
     test_name = os.path.splitext(pure_test_file)[0]
     path = os.path.join(test_dir, test_file)
     with open(path, "r") as fp:
-        json_data = json.load(fp)[f"{test_name}_{network}"]
+        data = json.load(fp)
+
+        # Some newer test files have for example _d0g0v0_
```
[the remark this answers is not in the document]
gurukamath: I am not entirely sure. But they do appear in both, the legacy as well as non-legacy tests. In the non-legacy tests, there is a slight discrepancy that this pattern does not appear in the file name but appears in the keys within, which is why I had to implement this pattern matching

tests/frontier/blockchain_st_test_helpers.py
gurukamath: All the regex meta-characters that we have passed in will be appropriately handled by `re.compile`. And no meta-characters appear in the `test_name` as far as I have seen. So, we should be ok.

voith: Instead of using the same function for legacy tests and the new tests, would it make sense to have a separate function for handling the new tests?

voith (commented):
LGTM, Besides the one comment that I left.

gurukamath:
@voith @SamWilsn Thanks for the feedback.

After considering the points that you raised, I did a detailed analysis of the `GeneralStateTests` cases within the legacy and non-legacy tests. I also looked at the keys that they use.

Here are the important points

1. A small number of test case names do in fact have regex meta-charactes in their names Eg:- `mem64kb_singleByte+32`. So we will have to do an `re.escape` after all :)
2. Occasionally, the keys in the new test case have larger numbers between the letters `d` `g` `v` and not just a single digit as I previously thought. Example:- `sstore_combinations_initial21_2_d452g0v0_Berlin` inside `sstore_combinations_initial21_2.json`

I say we just keep it simple and handle it all with the same regex `f"{re.escape(test_name)}.*{re.escape(network)}"` i.e starts with `test_name` and ends with `network` without worrying about what occurs in the middle. This handles old and new test cases in one go.

I have already tested this and it picks up exactly all the tests that we've been running thus far. Let me know what you think and I'll update the PR accordingly.

gurukamath:
Updated as described in the comment above

SamWilsn:
I think, for the time being, I'm fine with using a single function for both styles of test.

PR #465, opened 2022-03-03
Standardize calculate_block_difficulty across forks and implement difficulty tests

### LINKED ISSUE #464: Standardize `calculate_block_difficulty` across forks and implement difficulty tests
Currently, the `calculate_block_difficulty` has slightly different function signature across forks. 
For example, in Frontier, the function accepts the `block number` as one of the inputs whereas in the later forks, the `parent block number` is taken as an input and the `block number` is derived from it.

I couldn't quite see a reason why this has to be so. We can probably standardise this function across the forks and also implement the `difficulty tests` that are available in the fixtures.

@SamWilsn Do let me know if this makes sense. I can take up the implementation part of it.

gurukamath:
(closes #464 )

### What was wrong?
Currently, the calculate_block_difficulty has slightly different function signature across forks.

Related to Issue #464 

### How was it fixed?
Created a common format for the `calculate_block_difficulty` function before `Byzantium`
Implemented difficulty test which makes it much easier to implement and test any EIPs related to the difficulty calculation and difficulty bomb

#### Cute Animal Picture


tests/homestead/test_difficulty.py
```diff
@@ -0,0 +1,22 @@
...
+
+import pytest
+
+from ..helpers.load_difficulty_tests import DifficultyTestLoader
+
+test_loader = DifficultyTestLoader("Homestead", "homestead")
+
+
+@pytest.mark.parametrize("test_file", test_loader.test_files)
+def test_difficulty(test_file: str) -> None:
```
petertdavies, line 11: I think you should move all the logic in this function into the `DifficultyTestLoader` class and save a bunch of code duplication. Like this:

```python
@pytest.mark.parametrize("test_file", test_loader.test_files)
def test_difficulty(test_file: str) -> None:
    test_loader.run_tests(test_file)
```

petertdavies (changes requested):
I have a couple of comments, but otherwise this looks really good.

tests/homestead/test_difficulty.py
gurukamath: The implementation of this function differs for the different forks. All pre-byzantium forks have slightly different outputs when compared to Byzantium and later. I wanted to be able to handle it on a case by case basis, rather than in the helper class.

tests/helpers/load_difficulty_tests.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def load_test(self, test_file: str) -> List[Dict]:
+        test_name = os.path.splitext(test_file)[0]
+        path = os.path.join(self.test_dir, test_file)
+
+        with open(path, "r") as fp:
+            json_data = json.load(fp)[test_name][self.network]
+
+        test_list = []
+        for _test in json_data:
```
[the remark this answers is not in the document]
gurukamath: I guess I could have named this variable a bit better but essentially, I am using the variable name `test` to refer to the dictionary containing the test parameters in line 53

tests/homestead/test_difficulty.py
petertdavies: Ah, didn't notice that.

tests/helpers/load_difficulty_tests.py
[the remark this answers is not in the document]
petertdavies: It's more idiomatic to use an underscore suffix for this, but even better to use a different name. `test_name` would make it really obvious what is going on here.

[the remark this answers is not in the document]
gurukamath: Updated. Thanks for the feedback.

petertdavies (approved):
LGTM!

src/ethereum/tangerine_whistle/spec.py
```diff
@@ -735,9 +735,9 @@ def calculate_block_difficulty(
 
     Parameters
     ----------
-    parent_block_number :
-        Block number of the parent block.
```
voith, line 739: Why did you remove "parent" from the description? We are passing the parent's block number here.
https://github.com/ethereum/execution-specs/blob/8d24c57914258947466b47a842228d3bf9b648a0/src/ethereum/tangerine_whistle/spec.py#L184-L190

gurukamath: Essentially, in frontier, we were passing the `block_number`. For the later forks, we were passing the `parent block number` and then adding 1 to it inside the function. I just wanted to standardise this function to whatever extent possible. I have also updated the corresponding line where this function is eventually called.

petertdavies: He changed the semantics as well as the name, but seems to have forgotten to change the callers. It only matters for blocks where `block_number % 100000 == 0` so the test suite hasn't picked it up.

voith: Ok, so as peter mentioned you need to fix the callers.

gurukamath: Hmm. This is strange cause I distinctly remember updating these callers. Perhaps, I failed to commit some changes.

@voith @petertdavies Thanks for catching this error. I'll update

gurukamath: Updated!

SamWilsn (approved):
LGTM, wouldn't mind a second opinion though.

voith (approved):
LGTM! would be great if you could clean up the commit history.

gurukamath:
Done.

PR #470, opened 2022-03-10
Exception reform

### LINKED ISSUE #437: Comprehensive exception reform
@ultratwo suggests that we do more comprehensive exception reform. There should be a superclasses called things like `FailWithAllGasError`, `InvalidTransactionError`, etc.. from which most other exceptions derive. That way we can avoid long `except:` lists.

_Originally posted by @ultratwo in https://github.com/ethereum/execution-specs/pull/423#discussion_r789669025_

petertdavies:
(closes #437)

### What was wrong?

There was no organised scheme for exceptions in the specifications. Instead the specifications raised exceptions haphazardly (usually `EnsureError`) causing a number of unintentional exception swallowing issues. 

### How was it fixed?

`EnsureError` is removed. Callers to `ensure()` must specify an exception.

There is a new hierarchy of exceptions:
* `EthereumException`
    * `InvalidBlock` 
    *  `VMError`
        * `ConsumeAllGasException`
        * `Revert`
    * `RLPDecodingError`
    * `RLPEncodingError`

When `ensure()` is actually an `assert`, I have gone with `AssertionError`.

I removed `get_block_header_by_hash()` because it was unused and it wasn't clear which exceptions it should `raise`. I also made the test `bcInvalidHeaderTest/GasLimitHigherThan2p63m1.json` an expected failure, because I wasn't clear whether we should add the check being tested (it previously passed accidentally due to exception swallowing).

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/byzantium/spec.py
+++ b/src/ethereum/byzantium/spec.py
@@ -19,6 +19,7 @@ from ethereum.base_types import Bytes0
 from ethereum.crypto.elliptic_curve import SECP256K1N, secp256k1_recover
 from ethereum.crypto.hash import keccak256
 from ethereum.ethash import dataset_size, generate_cache, hashimoto_light
+from ethereum.exceptions import InvalidBlock
 from ethereum.utils.ensure import ensure
 
 from .. import rlp
@@ -160,11 +161,11 @@ def state_transition(chain: BlockChain, block: Block) -> None:
         block.transactions,
         block.ommers,
     )
-    ensure(gas_used == block.header.gas_used)
-    ensure(transactions_root == block.header.transactions_root)
-    ensure(state_root(state) == block.header.state_root)
-    ensure(receipt_root == block.header.receipt_root)
-    ensure(block_logs_bloom == block.header.bloom)
+    ensure(gas_used == block.header.gas_used, InvalidBlock)
+    ensure(transactions_root == block.header.transactions_root, InvalidBlock)
+    ensure(state_root(state) == block.header.state_root, InvalidBlock)
+    ensure(receipt_root == block.header.receipt_root, InvalidBlock)
+    ensure(block_logs_bloom == block.header.bloom, InvalidBlock)
 
     chain.blocks.append(block)
     if len(chain.blocks) > 255:
@@ -185,6 +186,8 @@ def validate_header(header: Header, parent_header: Header) -> None:
         Parent Header of the header to check for correctness
     """
     parent_has_ommers = parent_header.ommers_hash != EMPTY_OMMER_HASH
+    ensure(header.timestamp > parent_header.timestamp, InvalidBlock)
+
     block_difficulty = calculate_block_difficulty(
         header.number,
         header.timestamp,
@@ -195,12 +198,14 @@ def validate_header(header: Header, parent_header: Header) -> None:
 
     block_parent_hash = keccak256(rlp.encode(parent_header))
 
-    ensure(header.parent_hash == block_parent_hash)
-    ensure(header.difficulty == block_difficulty)
-    ensure(header.number == parent_header.number + 1)
-    ensure(check_gas_limit(header.gas_limit, parent_header.gas_limit))
-    ensure(header.timestamp > parent_header.timestamp)
-    ensure(len(header.extra_data) <= 32)
+    ensure(header.parent_hash == block_parent_hash, InvalidBlock)
+    ensure(header.difficulty == block_difficulty, InvalidBlock)
+    ensure(header.number == parent_header.number + 1, InvalidBlock)
+    ensure(
+        check_gas_limit(header.gas_limit, parent_header.gas_limit),
+        InvalidBlock,
+    )
+    ensure(len(header.extra_data) <= 32, InvalidBlock)
 
     validate_proof_of_work(header)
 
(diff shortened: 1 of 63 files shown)


src/ethereum/byzantium/spec.py
```diff
@@ -185,6 +186,8 @@ def validate_header(header: Header, parent_header: Header) -> None:
     parent_header :
         Parent Header of the header to check for correctness
     """
+    ensure(header.timestamp > parent_header.timestamp, InvalidBlock)
```
SamWilsn, line 189: Why'd this move up here?

src/ethereum/byzantium/vm/error.py
```diff
@@ -12,24 +12,43 @@
 Errors which cause the EVM to halt exceptionally.
 """
 
+from ethereum.exceptions import EthereumException
 
-class StackUnderflowError(Exception):
+
+class ConsumeAllGasException(EthereumException):
+    """
+    Indicates that EVM execution has fails with all gas being consumed.
```
SamWilsn, line 20:
```suggestion
    Indicates that EVM execution has failed with all gas being consumed.
```

SamWilsn, line 20:
```suggestion
    Indicates that EVM execution has failed with all gas being consumed.
```

src/ethereum/byzantium/vm/error.py
```diff
@@ -12,24 +12,43 @@
 Errors which cause the EVM to halt exceptionally.
 """
 
+from ethereum.exceptions import EthereumException
 
-class StackUnderflowError(Exception):
+
+class ConsumeAllGasException(EthereumException):
```
SamWilsn, line 18: Would `ExceptionalHalt` make sense with the Yellow Paper's terminology?

src/ethereum/utils/ensure.py
```diff
@@ -12,17 +12,11 @@
...
-class EnsureError(Exception):
-    """
-    Represents a failed check.
-    """
+from typing import Callable, Union
 
 
 def ensure(
-    value: bool, exception_class: Callable[[], Exception] = EnsureError
+    value: bool, exception: Union[Callable[[], BaseException], BaseException]
```
SamWilsn, line 19: This is to avoid `ensure(..., lambda: Foo("msg..."))`?

SamWilsn, line 19: This is to avoid `ensure(..., lambda: Foo("msg..."))`?

src/ethereum/byzantium/vm/error.py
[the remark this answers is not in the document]
petertdavies: I don't believe there is standard terminology for this. The Yellow Paper calls this "Exceptional Halt" while most EIPs use "Out Of Gas Error". "Consumes all gas" is also common in prose.

I went with `ConsumeAllGasException`, because I want to maximize clarity for the uninformed reader who will be confused by `OutOfGasError` and have to look up the meaning of `ExceptionalHalt`.

src/ethereum/utils/ensure.py
petertdavies: This is actually wrong, it should be `Union[Type[BaseException], BaseException]`.

The point is to make the semantics of `ensure()` identical to [`raise`](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement), which takes either a subclass or instance of `BaseException`.

src/ethereum/byzantium/spec.py
SamWilsn: Interesting. Should this check be moved into `calculate_block_difficulty` then?

Alternatively, would it make sense to move all the other checks (besides difficulty) up as well?

I worry that if we split the block of checks into several, it'll be less obvious.

src/ethereum/utils/ensure.py
SamWilsn: I believe the original type hint was more correct:

```python
>>> class Foo(Exception):
...     def __init__(self, arg1):
...         print(arg1)
... 
>>> raise Foo
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __init__() missing 1 required positional argument: 'arg1'
```

src/ethereum/byzantium/vm/error.py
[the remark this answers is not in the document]
SamWilsn: I think the name of the class should represent the cause of the exception (aka what it _is_), and the docstring can describe its effects. This is more consistent with `InvalidBlock` and the other VM errors.

src/ethereum/exceptions.py
```diff
@@ -0,0 +1,38 @@
+"""
```
SamWilsn, line 1: Perhaps we should rename this file to `error.py` for consistency with the analogous file in `vm/`?

src/ethereum/utils/ensure.py
petertdavies: No, `raise` takes a class derived from `BaseException` or an instance of such a class. If given a class it instantiates it with no arguments. I don't think the `typing` module can express "class that can be validly instantiated with no arguments".

```python
>>> def f():
...     return Exception()
... 
>>> raise f
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: exceptions must derive from BaseException
```

SamWilsn: Well that was a fruitless rabbithole to explore today. Python really needs an `Intersection` type hint...

With that pointless waste of time out of the way, I still think it would be better to use `Callable`:

```python
def ensure(
    value: bool, exception: Union[Callable[[], BaseException], BaseException]
) -> None:
    # ...
    if isinstance(exception, BaseException):
        raise exception
    else:
        raise exception()
```

It doesn't _exactly_ capture what `raise` requires, but I think it accepts a strict superset of `raise`'s behaviour, and catches the case where the exception requires an argument.

petertdavies:
I have a couple of things to add to this before you merge it.

petertdavies:
I have now added the two things I wanted to add.

On the subject of "errors" vs "exceptions", I think it makes more sense to name `vm/errors.py` to `vm/exceptions.py` since I'm not convinced that things like `Revert` are actually errors, but that is not worth doing in this PR. If you feel strongly I am OK with you renaming `exceptions.py` to `errors.py`.

SamWilsn (approved):

PR #471, opened 2022-03-10
Add patch tool

petertdavies:
This PR adds the simple patching tool that I used for the exception reform patch. This should make changes that affect multiple forks much easier.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/patch_tool.py
+++ b/src/ethereum_spec_tools/patch_tool.py
@@ -1,0 +1,78 @@
+"""
+Simple patching tool
+"""
+
+import argparse
+import subprocess
+from sys import stderr, stdout
+
+DESCRIPTION = """
+Apply the unstaged changes in SOURCE_FORK to each TARGET_FORK. If some of the
+change didn't apply, '.rej' files listing the unapplied changes will be left in
+TARGET_FORK.
+"""
+
+parser = argparse.ArgumentParser(description=DESCRIPTION)
+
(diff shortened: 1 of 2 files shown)


src/ethereum_spec_tools/patch_tool.py
```diff
@@ -0,0 +1,78 @@
...
+        elif line.startswith("--- a/"):
+            patch += line.replace(
+                "--- a/" + options.prefix + options.source_fork[0],
+                "--- " + options.prefix + target,
+            )
+        elif line.startswith("+++ b/"):
+            patch += line.replace(
+                "+++ b/" + options.prefix + options.source_fork[0],
+                "+++ " + options.prefix + target,
+            )
```
SamWilsn, line 64: I think you can use `-p1` on `patch` to achieve a similar goal?

whitelist.txt
```diff
@@ -297,3 +297,7 @@ dao
 precompile
 recurse
 pre155
+
+metavar
+const
+keepends
```
SamWilsn, line 303: Holy hell, we really need to help this plugin out or something...

src/ethereum_spec_tools/patch_tool.py
petertdavies: It ends up being just as complicated. You have to pass `-p3` for `src/ethereum` and `src/ethereum_optimized`, but `-p2` for `tests/`.

I'm happy to switch if you would prefer.

SamWilsn (approved):

PR #468, opened 2022-03-08
Implement modular exponentiation and alt_bn128 precompiles

petertdavies:
(closes #426 closes #427 closes #428)

This PR implements the remaining precompiles for Byzantium. It adds a mini elliptic curve library in the `crypto/` directory. This library is sufficiently general it should be usable for the BLS12-381 precompiles and the curve used by verkle trees, both of which are proposed extensions to the execution layer.

The performance is decent, but it may be necessary to add an optimized version of the curve precompiles using a high performance library written in a compiled language to obtain better performance in future.

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #473, opened 2022-03-15
Byzantium fixes

petertdavies:
*Marked draft until #468 gets merged.*

This PR contains fixes for issues found when I started syncing Byzantium. I have successfully synced from block 4370001 to block 4412454 (as of time of writing) with this branch.

I have had to change the way logs work in order to get the ordering correct. I cannot see a way to get this to work with the `collect_logs()` approach. The problem is illustrated by the following execution, which should result in logs in order `1, 2, 3` or the receipt root will be wrong:
```
A emits log 1
A calls B
    B emits log 2
A emits log 3
```

@voith Since this is your code, do you have any opinions?

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/byzantium/vm/instructions/system.py
+++ b/src/ethereum/byzantium/vm/instructions/system.py
@@ -126,6 +126,7 @@ def create(evm: Evm) -> None:
         push(evm.stack, U256(0))
         evm.return_data = child_evm.output
     else:
+        evm.logs += child_evm.logs
         push(evm.stack, U256.from_be_bytes(child_evm.message.current_target))
         evm.return_data = b""
     evm.gas_left += child_evm.gas_left
@@ -250,6 +251,7 @@ def call(evm: Evm) -> None:
         else:
             evm.return_data = b""
     else:
+        evm.logs += child_evm.logs
         push(evm.stack, U256(1))
         evm.return_data = child_evm.output
 
@@ -349,6 +351,7 @@ def callcode(evm: Evm) -> None:
         else:
             evm.return_data = b""
     else:
+        evm.logs += child_evm.logs
         push(evm.stack, U256(1))
         evm.return_data = child_evm.output
     actual_output_size = min(memory_output_size, U256(len(child_evm.output)))
@@ -480,6 +483,7 @@ def delegatecall(evm: Evm) -> None:
         else:
             evm.return_data = b""
     else:
+        evm.logs += child_evm.logs
         push(evm.stack, U256(1))
         evm.return_data = child_evm.output
     actual_output_size = min(memory_output_size, U256(len(child_evm.output)))
--- a/src/ethereum/byzantium/vm/interpreter.py
+++ b/src/ethereum/byzantium/vm/interpreter.py
@@ -117,7 +117,7 @@ def process_message_call(
     return MessageCallOutput(
         gas_left=evm.gas_left,
         refund_counter=refund_counter,
-        logs=collect_logs(evm),
+        logs=evm.logs if not evm.has_erred else (),
         accounts_to_delete=accounts_to_delete,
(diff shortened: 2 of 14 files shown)


voith:
Thanks for pinging me @petertdavies. I can review this PR on saturday.

src/ethereum/byzantium/vm/instructions/system.py
```diff
@@ -126,6 +126,7 @@ def create(evm: Evm) -> None:
         push(evm.stack, U256(0))
         evm.return_data = child_evm.output
     else:
+        evm.logs += child_evm.logs
```
voith, line 129: The last time I cleaned up this code,  I didn't want the `child_evm` logs to be a part of `evm.logs` so that there is no confusion while inspecting `evm` objects. 

Here's an alternate suggestion:
- Add a new attribute called `log_counter` to [Environment](https://github.com/ethereum/execution-specs/blob/a8ff48220f350decbba543015b4a2ea2b930aaf3/src/ethereum/frontier/vm/__init__.py#L29)
- Add a new attribute called `position` to [Log](https://github.com/ethereum/execution-specs/blob/a8ff48220f350decbba543015b4a2ea2b930aaf3/src/ethereum/frontier/eth_types.py#L134)
- Increment the `env.log_counter` every time a log is added and store this value in `Log`: `position = env.log_counter`
- sort the logs by `position` in [collect_logs](https://github.com/ethereum/execution-specs/blob/a8ff48220f350decbba543015b4a2ea2b930aaf3/src/ethereum/frontier/vm/interpreter.py#L289)

I can implement this myself if you'd like.

petertdavies: I think the easiest thing is if you fix this in a separate PR.

voith: sure, I'll take care of this.

SamWilsn: I've been thinking about this a bit, and I'm not sure @voith's solution is the most spec appropriate.

The position/sort approach is what came to my mind as well when I was considering the bug, but it's less "obvious" than taking the logs from the child evm after exiting a call frame.

Would adding a `clear()` call after taking the logs sort out the confusion you mentioned @voith?

SamWilsn: Ah, interesting. So instead of it being tracked per call frame, you get one list per transaction.

How would that work if a frame is reverted?

voith: > How would that work if a frame is reverted?

Good question: logs should be cleared on Excepteptional halting:
https://github.com/ethereum/execution-specs/blob/2f35c4cdd8400990439ef53844dd487fbf7f20e0/src/ethereum/byzantium/vm/interpreter.py#L266-L269

voith: On second thought this won't work as I won't know which logs to clear when the frame reverts.

gurukamath: Would it be possible to capture the message depth for each log within the log itself and then, during the final stage, remove the logs where depth is the same as an erred child evm?

voith: Well, that solution is not too different from what I suggested [here](https://github.com/ethereum/execution-specs/pull/473#discussion_r831425827). That's added complexity.

gurukamath: Yeah. Plus I realised that there might be multiple EVMs at the same depth, so it won't work anyway cause we Weill end up deleting all the logs at a certain depth even if only one fails

petertdavies:
I think we should pull this in it's current state. The logs issue seems to be a subject of ongoing debate, but having a working version in `master` would be nice.

PR #483, opened 2022-03-28
Re-use the same temp folder for pytest

### LINKED ISSUE #444: `tox test` keeps filling up my `/tmp` partition
### What was wrong?

If I run `tox test` too many times in a row, it eventually fills up the partition and further tests fail.

### How can it be fixed?

Either reusing the ethash directory, or cleaning it up more frequently.

gurukamath:
(closes #444 )

### What was wrong?
By default, `pytest` does not remove the temporary files created during each run. The current temp files are only deleted after another four runs. This can lead to issues with space since each run can accumulate up to 7-8 GB of files.

Related to Issue #444 

### How was it fixed?
Define a fixed temp folder within specs, rather than the `pytest` default of `pytest-NUM` where `NUM` is the number of the current run. This way, the folder is reused on successive runs. It can also be deleted manually as it is available within the specs itself.

#### Cute Animal Picture


tox.ini
```diff
@@ -11,7 +11,7 @@ commands =
     black src tests setup.py --check --diff --exclude "tests/fixtures/*"
     flake8 src tests setup.py
     mypy src tests setup.py --exclude "tests/fixtures/*" --namespace-packages
-    pytest --cov=ethereum --cov-report=term --cov-report "xml:{toxworkdir}/coverage.xml" --ignore-glob='tests/fixtures/*' -n 2
+    pytest --cov=ethereum --cov-report=term --cov-report "xml:{toxworkdir}/coverage.xml" --ignore-glob='tests/fixtures/*' -n 2 --basetemp=specs-temp
```
SamWilsn, line 14: Would it make sense to put the directory under [`{temp_dir}`](https://tox.wiki/en/3.24.5/config.html#globally-available-substitutions)?

```suggestion
    pytest --cov=ethereum --cov-report=term --cov-report "xml:{toxworkdir}/coverage.xml" --ignore-glob='tests/fixtures/*' -n 2 --basetemp="{temp_dir}/pytest"
```

gurukamath: @SamWilsn Sure. As long as we don't leave it up to `pytest` to decide the temp folder, we could place it wherever we want. Have updated the command and reverted my changes to `.gitignore` since `.tox` is already in it.

SamWilsn (approved):

PR #485, opened 2022-03-29
Current version is incompatible with click 8.1.0

gurukamath:
### What was wrong?
The version of `black` used in specs is incompatible with the latest version of click (see [issue](https://github.com/psf/black/issues/2964))

Since the version of click is not locked-in within the spec `setup.cfg`, CI on `Github Actions` fails.

### How was it fixed?
We have two options here.
1. Start using the newer version of `black`. However, there seem to be some rule changes in black which will lead to errors on many spec files (about 40 of them)
2. Lock-in the version of click in `setup.cfg`. This has been implemented in the PR

#### Cute Animal Picture

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -92,6 +92,7 @@ test =
 lint =
     isort>=5.8,<6
     mypy==0.812; implementation_name == "cpython"
+    click==8.0.4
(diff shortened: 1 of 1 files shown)


SamWilsn:
Closing in favour of #486

PR #486, opened 2022-03-29
Update black version

petertdavies:
### What was wrong?

The current version of black is [incompatible](https://github.com/psf/black/issues/2964) with the latest version of click.

### How was it fixed?

I updated to the latest black version. The only change is the whitespace around `**`.

Alternatively you can solve the problem by pinning click with PR #485.

#### Cute Animal Picture


voith (approved):
LGTM. 
I prefer being on the latest version of black rather than fixing this issue by pinning click.

SamWilsn (approved):

PR #487, opened 2022-03-29
Prevent exception swallowing

petertdavies:
### What was wrong?

The code:
```python
try:
    ...
finally:
    return evm
```

swallows all exceptions (even things like `SystemExit`), because `return` causes `finally` to discard the exception. 

### How was it fixed?

Remove inappropriate `finally:`.

#### Cute Animal Picture


SamWilsn:
Looks like there's a flake8 extension to catch this (among other things): [`flake8-bugbear`](https://github.com/PyCQA/flake8-bugbear).

Any reason not to add it?

PR #492, opened 2022-04-04
Gas and memory refactoring

petertdavies:
This PR does a couple of useful refactorings.

The first moves the gas deduction logic into `subtract_gas()` making invocation simpler. `vm.gas_left = subtract_gas(evm.gas_left, GAS_LOW)` becomes `subtract_gas(evm, GAS_LOW)`.

The second makes memory extension an implicit part the memory access functions. This removes quite a lot of boilerplate from the instruction code. `touch_memory()` is required because `CALL` and friends extend memory based on region assigned for return data rather than the region actually written to.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/byzantium/vm/gas.py
+++ b/src/ethereum/byzantium/vm/gas.py
@@ -70,7 +70,7 @@ def subtract_gas(evm: Evm, amount: U256) -> None:
     evm :
         Current `evm` instance.
     amount :
-        The amount of gas the current operation requires.
+        The amount of gas to deduct.
 
     Raises
     ------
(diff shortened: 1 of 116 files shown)


petertdavies:
Based on discussions on the spec sync call I will update this PR so the only way to extend memory is the `touch_memory()` function.

voith:
>Based on discussions on the spec sync call I will update this PR so the only way to extend memory is the touch_memory() function.

Mind giving me a TLDR of the discussion?

petertdavies:
> Mind giving me a TLDR of the discussion?

Basically there are two things wrong with my approach:
1. It doesn't match the way the EVM works (see #493).
2. It makes it difficult for a reader to work out the potential costs of an opcode.

(point 2 wasn't actually discussed on the call)

petertdavies:
Superseded by #581

PR #502, opened 2022-04-11
Upgrade state testing framework for upcoming forks

gurukamath:
### What was wrong?
The current state testing framework assumes that there is only a single test case for every json test fixture. While this may be true for a large majority of the tests for the forks that have thus far been implemented, it certainly does not hold for the upcoming forks (non-legacy tests)

### How was it fixed?
Updates state tests to be able to run multiple tests per json. i.e. looking at a key within a json file rather than the file itself, as an individual test case.

Additionally, the small optimisation has been implemented where `fetch_state_test_files` is only responsible for fetching the test cases and `run_<fork>_blockchain_st_tests` is only responsible for running them. This separation of duties was not entirely clear in the current set-up.

#### Cute Animal Picture

### DIFF
--- a/tests/helpers/load_state_tests.py
+++ b/tests/helpers/load_state_tests.py
@@ -334,7 +334,7 @@ def load_json_fixture(test_file: str, network: str) -> Generator:
 def fetch_state_test_files(
     test_dir: str,
     network: str,
-    custom_file_list: Tuple[str, ...] = (),
+    only_in: Tuple[str, ...] = (),
     slow_list: Tuple[str, ...] = (),
     ignore_list: Tuple[str, ...] = (),
 ) -> Generator[Union[Dict, ParameterSet], None, None]:
@@ -345,10 +345,10 @@ def fetch_state_test_files(
     # Get all the files to iterate over
     # Maybe from the custom file list or entire test_dir
     files_to_iterate = []
-    if any(custom_file_list):
+    if len(only_in):
         # Get file list from custom list, if one is specified
-        for _file in custom_file_list:
-            files_to_iterate.append(os.path.join(test_dir, _file))
+        for test_path in only_in:
+            files_to_iterate.append(os.path.join(test_dir, test_path))
     else:
         # If there isnt a custom list, iterate over the test_dir
         all_jsons = [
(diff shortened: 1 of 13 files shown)


tests/helpers/load_state_tests.py
```diff
@@ -325,24 +302,86 @@ def add_blocks_to_chain(
...
+    ignore_list: Tuple[str, ...] = (),
+) -> Generator[Union[Dict, ParameterSet], None, None]:
+
+    all_slow = [re.compile(x) for x in slow_list]
+    all_ignore = [re.compile(x) for x in ignore_list]
+
+    # Get all the files to iterate over
+    # Maybe from the custom file list or entire test_dir
+    files_to_iterate = []
+    if any(custom_file_list):
```
SamWilsn, line 348:
```suggestion
    if len(custom_file_list):
```

tests/helpers/load_state_tests.py
```diff
@@ -325,24 +302,86 @@ def add_blocks_to_chain(
...
-                yield pytest.param(_test_file, marks=pytest.mark.slow)
-            else:
-                try:
-                    load_json_fixture(test_dir, _test_file, load)
-                    yield _test_file
-                except KeyError:
-                    # file doesn't contain tests for the given fork
-                    pass
+    network: str,
+    custom_file_list: Tuple[str, ...] = (),
```
SamWilsn, line 337:
```suggestion
    only_in: Tuple[str, ...] = (),
```

I think this is more clear than `custom_file_list`, especially since these files are not appended to the discovered tests.

tests/helpers/load_state_tests.py
```diff
@@ -325,24 +302,86 @@ def add_blocks_to_chain(
...
+            files_to_iterate.append(os.path.join(test_dir, _file))
+    else:
+        # If there isnt a custom list, iterate over the test_dir
+        all_jsons = [
+            y
+            for x in os.walk(test_dir)
+            for y in glob(os.path.join(x[0], "*.json"))
+        ]
+
+        for _full_path in all_jsons:
```
SamWilsn, line 360:
```suggestion
        for full_path in all_jsons:
```

Only prefix with an underscore if the variable is unused.

tests/helpers/load_state_tests.py
```diff
@@ -325,24 +302,86 @@ def add_blocks_to_chain(
...
+    all_slow = [re.compile(x) for x in slow_list]
+    all_ignore = [re.compile(x) for x in ignore_list]
+
+    # Get all the files to iterate over
+    # Maybe from the custom file list or entire test_dir
+    files_to_iterate = []
+    if any(custom_file_list):
+        # Get file list from custom list, if one is specified
+        for _file in custom_file_list:
+            files_to_iterate.append(os.path.join(test_dir, _file))
```
SamWilsn, line 351:
```suggestion
        for test_path in custom_file_list:
            files_to_iterate.append(os.path.join(test_dir, test_path))
```

tests/helpers/load_state_tests.py
```diff
@@ -325,24 +302,86 @@ def add_blocks_to_chain(
...
+                # If a file or folder is marked for ignore,
+                # it can already be dropped at this stage
+                files_to_iterate.append(_full_path)
+
+    # Start yielding individual test cases from the file list
+    for _test_file in files_to_iterate:
+        try:
+            for _test_case in load_json_fixture(_test_file, network):
+                # identifier could identifiy files, folders or individual cases
+                _identifier = _test_case["test_file"] + _test_case["test_key"]
```
SamWilsn, line 371: Hm, concatenating these together seems like you might get some weird effects?

tests/helpers/load_state_tests.py
gurukamath: Updated

tests/helpers/load_state_tests.py
gurukamath: Updated

tests/helpers/load_state_tests.py
gurukamath: Updated

tests/helpers/load_state_tests.py
gurukamath: Updated

tests/helpers/load_state_tests.py
gurukamath: Since the file name always ends with `.json`, I do not really expect any side-effects. 
But you are right. This is not the smartest way to handle this. I have updated the code to handle this using the regex `OR`

`_identifier = "(" + _test_case["test_file"] + "|" + _test_case["test_key"] + ")"`

SamWilsn (approved):

PR #504, opened 2022-04-19
Opcode refactoring proposal

petertdavies:
This PR is a demo of my proposal to completely restructure the way opcodes work. I have only modified the frontier fork. This is intended mainly for comment rather than review.

There a few significant issues:
* `evm.pc` should be renamed to `evm.next_pc`
* Docstrings have not been done properly
* We should consider whether the order of arguments to the opcode functions should be reversed.

Is there general support this substantial change?

#### Cute Animal Picture


SamWilsn:
I'd suggest renaming `stack` to `rest`.

I'm still not _entirely_ sold on this format, though I do agree adding the argument is an improvement. Do you have any thoughts @lightclient and/or @gurukamath?

petertdavies:
I've made a couple of changes:
* I've renamed `rest` to `stack` as suggested above
* I've switched the order of arguments to correspond to be in same order as the python list that represents the stack.

The argument reorder results in arguments appearing in weird orders, but this is a consequence of the choice of argument ordering made by the EVM. For example:

```python
def do_less_than(evm: Evm, stack: List[U256], right: U256, left: U256) -> U256:
    """
    Checks if the top element is less than the next top element. Pushes the
    result back on the stack.
    """
    return U256(left < right)
```

petertdavies:
Superseded by #581

Issue #510, opened 2022-05-04
Some `InvalidBlocks` tests under Istanbul don't fail


gurukamath:
The tests under `InvalidBlocks` are supposed to fail. However, [some of them](https://github.com/gurukamath/execution-specs/blob/db3ff58b41fc127bb8c351ab2c212f7d0b26862d/tests/istanbul/test_state_transition.py#L105) do not fail in Istanbul.

This is very likely an issue with the test cases, since all other tests run absolutely fine. However, a further investigation into this might be necessary.

====================
Reference from #503 
We probably also want an issue to investigate why some of the [tests](https://github.com/gurukamath/execution-specs/blob/db3ff58b41fc127bb8c351ab2c212f7d0b26862d/tests/istanbul/test_state_transition.py#L105) under `InvalidBlocks` don't fail.

_Originally posted by @gurukamath in https://github.com/ethereum/execution-specs/issues/503#issuecomment-1099875903_

gurukamath:
Have created a PR for adding the folder name checks to `xfail` since I found a couple of tests that had the exact same `test_key`.
The main issue mentioned above can only be handled once the ethereum tests repository closes [#1037](https://github.com/ethereum/tests/issues/1037) which I have raised.

gurukamath:
As per the comments on the tests repository issue, the team maintains a test in InvalidBlock tests, if it is invalid for any single fork. It might be valid for earlier or later forks. So we will have to continue to add maintain  a list of tests that won't fail for each for test.

PR #529, opened 2022-05-28
Include folder names for xfail tests check

gurukamath:
### What was wrong?
The tests under InvalidBlocks are supposed to fail. However, [some of them](https://github.com/gurukamath/execution-specs/blob/db3ff58b41fc127bb8c351ab2c212f7d0b26862d/tests/istanbul/test_state_transition.py#L105) do not fail in Istanbul.

Related to Issue #510 

### How was it fixed?
The issue seems to be with the ethereum tests repository since the tests pass without raising any exceptions when executed with `retesteth` either. I have created an [issue](https://github.com/ethereum/tests/issues/1037) in the tests repository.

I noticed however, that the tests `bcUncleHeaderValidity/wrongStateRoot.json` and `bcInvalidHeaderTest/wrongStateRoot.json` have exactly the same name which should ideally not have been the case.

This PR adds an additional check of folder name to determine if it is a case of `xfail`

#### Cute Animal Picture

)

### DIFF
--- a/tests/istanbul/test_state_transition.py
+++ b/tests/istanbul/test_state_transition.py
@@ -96,17 +96,29 @@ def test_uncles_correctness(test_case: Dict) -> None:
 # Run legacy invalid block tests
 test_dir = "tests/fixtures/BlockchainTests/InvalidBlocks"
 
-# TODO: Investigate why some of the below tests pass
+# TODO: Handle once https://github.com/ethereum/tests/issues/1037
+# is resolved
 # All except GasLimitHigherThan2p63m1_Istanbul
 xfail_candidates = (
-    "timestampTooLow_Istanbul",
-    "timestampTooHigh_Istanbul",
-    "wrongStateRoot_Istanbul",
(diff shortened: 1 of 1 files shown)


tests/istanbul/test_state_transition.py
```diff
@@ -96,19 +96,31 @@ def test_uncles_correctness(test_case: Dict) -> None:
...
+    ("bcUncleHeaderValidity", "wrongStateRoot_Istanbul"),
+    ("bcUncleHeaderValidity", "incorrectUncleTimestamp4_Istanbul"),
+    ("bcUncleHeaderValidity", "incorrectUncleTimestamp5_Istanbul"),
+    ("bcUncleSpecialTests", "futureUncleTimestamp3_Istanbul"),
+    ("bcInvalidHeaderTest", "GasLimitHigherThan2p63m1_Istanbul"),
 )
 
 
+def is_in_xfail(test_case: Dict) -> bool:
+    for _dir, _test_key in xfail_candidates:
```
SamWilsn, line 114: Why the underscores?

PR #547, opened 2022-06-23
Smoothing out irrelevant diffs (closes #519)

### LINKED ISSUE #519: Investigate and smoothen out irrelevant diffs
Once #515 gets merged, analyse the diffs between forks and smoothen out any irrelevant/incorrect diffs.

As an e.g - the Dao Fork [ecrecover code](https://github.com/ethereum/execution-specs/blob/master/src/ethereum/dao_fork/vm/precompiled_contracts/ecrecover.py) seems to be slightly different from the other forks which should not be the case.

nam0mi:


SamWilsn (approved):

gurukamath:
@nam0mi There might be a few more irrelevant diffs in addition to the ecrecover one that I mentioned in the issue. That was one example. 

It will involve looking through the various diff links on the rendered docs and reducing unnecessary diffs

As an additional example, if I look at the `Frontier -> Homestead` [diff ](https://ethereum.github.io/execution-specs/diffs/frontier_homestead/bloom/index.html) for `bloom.py`. I see that differences are mainly how the sentences in the comment are framed, which I believe is not relevant.

`Calculate the Bloom filter for a set of logs.` vs `Obtain the logs bloom from a list of log entries.`

PR #551, opened 2022-06-28
Updated introduction documentation to the __init__.py file

nam0mi:
The issue was that I went over the line character limit in the last PR, so I fixed that issue and made sure each line in the introduction was under 75 characters long.

### DIFF
--- a/src/ethereum/__init__.py
+++ b/src/ethereum/__init__.py
@@ -4,16 +4,17 @@ Ethereum Specification
 Seeing as internet connections have been vastly expanding across the
 world, spreading information has become as cheap as ever. Bitcoin, for
 example, has demonstrated the possibility of creating a decentralized,
-and practically free, trade system that is accessible around the world.
-Namecoin is another system that built off of Bitcoin's currency structure
-to create other simple technological applications.
+trade system that is accessible around the world. Namecoin is another
+system that built off of Bitcoin's currency structure to create other
+simple technological applications.
 
 Ethereum's goal is to create a cryptographically secure system in which
 any and all types of transaction-based concepts can be built. It provides
 an exceptionally accessible and decentralized system to build software
 and execute transactions.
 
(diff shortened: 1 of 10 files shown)


src/ethereum/__init__.py
```diff
@@ -1,6 +1,17 @@
 """
 Ethereum Specification
 ^^^^^^^^^^^^^^^^^^^^^^
+Seeing as internet connections have been vastly expanding across the
+world, spreading information has become as cheap as ever. Bitcoin, for
+example, has demonstrated the possibility of creating a decentralized,
+and practically free, trade system that is accessible around the world.
```
SamWilsn, line 7: > and practically free

Is this still the case?

src/ethereum/__init__.py
```diff
@@ -1,6 +1,17 @@
...
+Namecoin is another system that built off of Bitcoin's currency structure
+to create other simple technological applications.
+
+Ethereum's goal is to create a cryptographically secure system in which
+any and all types of transaction-based concepts can be built. It provides
+an exceptionally accessible and decentralized system to build software
+and execute transactions.
 
 Core specifications for Ethereum clients.
```
SamWilsn, line 16: I think you can either remove this line, or expand on it and describe what this repository is for. Something like:

```suggestion
This package contains a reference implementation, written as simply as possible, to aid in defining the behavior of Ethereum clients.
```

src/ethereum/__init__.py
nam0mi: It said that on the Yellow Paper but I may be wrong.

src/ethereum/__init__.py
nam0mi: I'll add that!

src/ethereum/__init__.py
SamWilsn: good enough for me!

PR #552, opened 2022-07-04
re-factor `evm_trace`

gurukamath:
### What was wrong?

The `evm_trace` function is currently defined in the `__init__.py` file of the `ethereum` package. This seems to be a bit out of place especially from the perspective of the rendered docs.

### How was it fixed?
Moved `evm_trace` to it's own module within the `ethereum` package. Additionally, there is no need to keep an initial placeholder function and then replace it with the trace during testing. Updated to make the `evm_trace` depend solely on the `logging` level. This way, a trace can be triggered even outside of `pytest`.

#### Cute Animal Picture


src/ethereum/__init__.py
```diff
@@ -14,11 +13,3 @@
...
 EVM_RECURSION_LIMIT = 1024 * 12
 sys.setrecursionlimit(max(EVM_RECURSION_LIMIT, sys.getrecursionlimit()))
-
-
-def evm_trace(evm: Any, op: Any) -> None:
-    """
-    Placeholder for an evm trace function. The spec does not trace evm by
-    default. EVM tracing will be injected if the user requests it.
-    """
-    pass
```
SamWilsn, line 24: Hm... interesting idea. The current `evm_trace` implementation is certainly simple enough that it isn't a problem to include.

The reason we had the placeholder and put the implementation in `ethereum_spec_tools` was to clearly differentiate between what is specification and what is not. Moving the implementation into `ethereum/` would imply that it is somehow spec related.

Why do you want to move it?

gurukamath: @SamWilsn updated as discussed in the sync i.e. marking the doctoring with a tag.

SamWilsn (approved):

Issue #374, opened 2021-10-14
Write tests for bugs caught while syncing the chain


SamWilsn:
_Originally posted by @lightclient in https://github.com/ethereum/execution-specs/pull/373#pullrequestreview-779880591_

**Should re-check to see if these have been added to EEST since this issue was written**

This is a bit of a meta-issue for tracking bugs that weren't caught by the tests that were triggered while syncing the chain.

These tests should theoretically get upstreamed to https://github.com/ethereum/tests and checked off here once they're merged.

 - [ ] `sstore` persists between transactions (#368)
 - [ ] charge gas for `ECRECOVER`
 - [ ] `blockhash` should hash only the header, not the block (#373)
 - [ ] `signextend` (#375)
 - [ ] Push-data extends past end of contract code (#375)
 - [ ] Deploying an empty contract (#375)
 - [ ] `ECRECOVER` with truncated call data (#445)
 - [ ] Logs emitted during `CREATE2` [#466](https://github.com/ethereum/execution-specs/pull/466#discussion_r845540487)
 - [ ] Blockhash at 255/256/257

ultratwo:
I have two more:

- [ ] Doing a `CALL` with non-zero bits in the 8 high bytes
- [ ] Check compliance with condition d) in EIP 161

gurukamath:
- [ ] Check compliance with [EIP-3607](https://eips.ethereum.org/EIPS/eip-3607) #583

SamWilsn:
https://github.com/ethereum/execution-spec-tests/issues/866 would probably be useful for this

gurukamath:
Also needd to upstream the `insufficient_balance_for_value` test case that we wrote some time ago, if not done already. See [this](https://github.com/ethereum/execution-specs/blob/78fb726158c69d8fa164e28f195fabf6ab59b915/tests/paris/test_state_transition.py#L96)

PR #553, opened 2022-07-04
Always clear returndata

petertdavies:
### What was wrong?

Returndata was not cleared in some failure returns of `CREATE`.

### How was it fixed?

All `CREATE` and `CALL` variants clear returndata at the start of the opcode, as specified by EIP-211.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/byzantium/vm/instructions/system.py
+++ b/src/ethereum/byzantium/vm/instructions/system.py
@@ -77,6 +77,8 @@ def create(evm: Evm) -> None:
 
     evm.pc += 1
 
+    evm.return_data = b""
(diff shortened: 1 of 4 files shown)


SamWilsn (approved):
This looks good. Would it be possible to add a python test (or even better, a filled test) for this as well?

petertdavies:
> This looks good. Would it be possible to add a python test (or even better, a filled test) for this as well?

I have added this bug to the list in #374. I wasn't planning to actually write tests until we implement test generation/filling with specs. The bugs in #374 will make a nice set of test testcases.

PR #560, opened 2022-07-12
Add incantations to make ripemd work on OpenSSL 3

### LINKED ISSUE #506: Openssl has disabled RIPEMD160 by default
OpenSSL 3 has disabled various legacy hash functions by moving them to a legacy provider that isn't enabled by default. This includes RIPEMD160 which we use.

If you have OpenSSL 3 and try to use RIPEMD160 you will get `ValueError: [digital envelope routines] unsupported` or `ValueError: unsupported hash type ripemd160` under CPython and `Segmentation fault` under PyPy.

The issue can be resolved by adding this lines to `/etc/ssl/openssl.cnf`:

```
 [openssl_init]
 providers = provider_sect
 
 # List of providers to load
 [provider_sect]
 default = default_sect
+legacy = legacy_sect

 [default_sect]
 activate = 1

+[legacy_sect]
+activate = 1
```

petertdavies:
(fixes #506)

### What was wrong?

OpenSSL 3 disabled legacy crypto algorithms by default, including RIPEMD (which we use).

### How was it fixed?

Add some incantations to `__init__.py` to load the "legacy" provider, which will reenable RIPEMD. I have decided to go with an approach of silently ignoring all failures in the incantations, since the code is inherently quite hacky (It uses the C ABI, not the API) and in some failure cases (e.g. not having OpenSSL 3) things are likely to work anyway.

#### Cute Animal Picture

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -83,6 +83,7 @@ install_requires =
     pysha3>=1,<2
     coincurve>=17,<18
     typing_extensions>=4
+    cffi>=1.15
 
 [options.package_data]
 ethereum =
--- a/src/ethereum/__init__.py
+++ b/src/ethereum/__init__.py
@@ -19,7 +19,31 @@ possible, to aid in defining the behavior of Ethereum clients.
 import sys
 from typing import Any
 
+import cffi
+
 __version__ = "0.1.0"
+
+try:
+    # These incantations are necessary to make ripemd160 work on OpenSSL 3
+    ffi = cffi.FFI()
+    ffi.cdef("void *OSSL_PROVIDER_load(void *libctx, const char *name);")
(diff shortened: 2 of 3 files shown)


SamWilsn:
Yikes... at this point should we just switch to a pure python ripemd160, or find a different library?

SamWilsn:
This doesn't seem too difficult to implement from scratch: https://github.com/karpathy/cryptos/blob/main/cryptos/ripemd160.py

petertdavies:
OpenSSL have [reversed](https://www.openssl.org/blog/blog/2022/10/18/rmd160-and-the-legacy-provider/) the upstream deprecation of RIPEMD160. This resolves the issues this PR was working around.

PR #566, opened 2022-07-22
Button to display only elements of diffs that changed

gurukamath:
This is an additional feature. It allows users to narrow down to only those parts of the diffs that actually changed. The diffs pages will have a button that allows one to toggle the two views.

Also updated the rendered spec to display meaningful names for diff files instead of just the module names.

The final result of this PR will look something like [this](https://gurukamath.github.io/execution-specs/index.html).

#### Cute Animal Picture

### DIFF
--- a/doc/_static/js/toggle_diffs.js
+++ b/doc/_static/js/toggle_diffs.js
@@ -1,5 +1,15 @@
+// Click Event Listner for toggle button
+window.addEventListener("load", (e) => {
+    const btn = document.getElementById("diff-toggle")
+    if (btn){
+        btn.addEventListener("click", toggleElements);
+    }
+})
+
 // All diffs will have one of the below classes
-const changeClasses = ["change-replaced", "change-replacement", "change-added", "change-removed"];
+const changeClassList = ["change-replaced", "change-replacement", "change-added", "change-removed"];
+const changeClassString = "." + changeClassList.join(", .");
+
 // The list of components to be scanned for diffs - selected by ID
 const componentsList = [
     '#module-details > *',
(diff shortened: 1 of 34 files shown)


doc/_templates/layout.html
```diff
@@ -23,6 +23,10 @@
         </li>
     </ol>
 </nav>
+{% elif "diffs/" in pagename %}
+<nav class="menu">
+    <button class="diff-toggle" onclick="toggleElements()">Only Show Diffs</button>
```
SamWilsn, line 28:
```suggestion
    <button class="diff-toggle">Only Show Diffs</button>
```

Then use an `addEventListener` in the javascript.

doc/_static/js/toggle_diffs.js
```diff
@@ -0,0 +1,68 @@
...
+
+        for (j = 0; j < children.length; j++){
+            if (changeClasses.some(r => children[j].classList.value.includes(r))){
+                hasDiffs = true;
+                break;
+            } 
+        }
+    }
+
+    return hasDiffs;
```
SamWilsn, line 67: Perhaps:

```suggestion
    if (changeClasses.some(r => element.classList.value.includes(r))){
        return true;
    } else {
        const children = element.querySelectorAll('*');

        for (j = 0; j < children.length; j++){
            if (changeClasses.some(r => children[j].classList.value.includes(r))){
                return true;
            } 
        }
    }

    return false;
```

Also, would it be better to join the classes with `, ` like you do for `componentsList` and let the browser handle it?

gurukamath: Will update the `childrenHaveDiffs` function. 
Sorry. I didn't quite fully get the comment about joining the classes with `, `.

doc/_templates/layout.html
gurukamath: Will update

doc/_static/js/toggle_diffs.js
SamWilsn: Instead of:

```js
const children = element.querySelectorAll('*');
```

How about:

```js
const children = element.querySelectorAll("." + changeClasses.join(", ."));
```

I totally could be way off the mark about what this does though ^^;;

gurukamath: Good idea. That way I avoid the for loop. Updated

gurukamath:
Closed all the review comments.

SamWilsn (approved):

PR #571, opened 2022-07-29
Improve sync tool and add Berlin support

petertdavies:
### What was wrong?

The sync tool would deserialize blocks with the wrong hardfork for blocks near hardfork boundaries and did not support the new transaction types added in Berlin.

### How was it fixed?

Split the `Sync` class into a `Sync` class and a `Downloader` class with independent tracking of the current fork. Add support for Berlin blocks.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/sync.py
+++ b/src/ethereum_spec_tools/sync.py
@@ -638,7 +638,7 @@ class Sync(ForkTracking):
         if persisted_block is None:
             self.set_block(0)
             self.downloader = BlockDownloader(
-                self.log, self.options.rpc_url, self.options.geth, 0
+                self.log, self.options.rpc_url, self.options.geth, 1
             )
             self.chain = self.module("spec").BlockChain(
                 blocks=[],
--- a/src/ethereum_spec_tools/sync.py
+++ b/src/ethereum_spec_tools/sync.py
@@ -236,6 +236,73 @@ class BlockDownloader(ForkTracking):
 
             return [v for (_, v) in sorted(blocks.items())]
 
+    def load_transaction(self, t: Any) -> Any:
+        """
+        Turn a json transaction into a `Transaction`.
+        """
+        if hasattr(self.module("eth_types"), "LegacyTransaction"):
+            if t["type"] == "0x1":
+                access_list = []
+                for sublist in t.get("accessList", []):
+                    access_list.append(
+                        (
+                            self.module("utils.hexadecimal").hex_to_address(
+                                sublist.get("address")
+                            ),
+                            [
+                                hex_to_bytes32(key)
+                                for key in sublist.get("storageKeys")
+                            ],
+                        )
+                    )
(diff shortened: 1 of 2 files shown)


src/ethereum_spec_tools/sync.py
```diff
@@ -499,23 +296,87 @@ def fetch_blocks_eth(
...
-                                if t["to"]
-                                else Bytes0(b""),
-                                hex_to_u256(t["value"]),
-                                hex_to_bytes(t["input"]),
-                                hex_to_u256(t["v"]),
-                                hex_to_u256(t["r"]),
-                                hex_to_u256(t["s"]),
+                        if not hasattr(
+                            self.module("eth_types"), "LegacyTransaction"
+                        ):
```
SamWilsn, line 301: Unnecessary `not`, I think. Can just flip the `if`/`else`.

src/ethereum_spec_tools/sync.py
```diff
@@ -128,239 +71,93 @@ def next_fork(self) -> Optional[Hardfork]:
...
 
-        with request.urlopen(post) as response:
-            replies = json.load(response)
-            ommers: Dict[int, Dict[int, Any]] = {}
+        while running:
+            count = max(1, self.queue.maxsize // 2)
+            if self.next_fork:
+                # Don't fetch a batch across a fork boundary
+                count = min(count, self.next_fork.block - self.block_number)
+            replies = self.fetch_blocks(self.block_number, count)
```
SamWilsn, line 138: Am I correct in assuming that we can't batch across a fork boundary because something in the fetch code needs to understand the header? Something feels weird about having to split batches because of that.

src/ethereum_spec_tools/sync.py
```diff
@@ -499,23 +296,87 @@ def fetch_blocks_eth(
...
+                                            else Bytes0(b""),
+                                            hex_to_u256(t["value"]),
+                                            hex_to_bytes(t["input"]),
+                                            access_list,
+                                            hex_to_u256(t["v"]),
+                                            hex_to_u256(t["r"]),
+                                            hex_to_u256(t["s"]),
+                                        )
+                                    )
+                                )
```
SamWilsn, line 361: Can we make a few functions out of these? Not necessarily to reuse code, but to reduce the indentation level a bit :rofl:

src/ethereum_spec_tools/sync.py
petertdavies: Fork A blocks need to processed by the code in `src/ethereum/A` and fork B blocks need to be processed by the code in `src/ethereum/B`. The current code tracks forks globally rather than on a per block basis. This means we must process all fork A blocks, switch to fork B and then start processing fork B blocks.

Up until now we had been getting away with this due to python duck typing.

We could track forks on a per block basis instead, but I can't see any benefit to doing that.

SamWilsn (approved):

PR #579, opened 2022-07-31
Update/optimise `meaningful_diffs` calculation in diff tool

### LINKED ISSUE #572: Some pages without diffs are rendered
There are about 9 pages that do not have any diffs but are rendered nonetheless. See [example](https://ethereum.github.io/execution-specs/diffs/tangerine_whistle_spurious_dragon/utils/message/index.html)

The reason is that the text `tangerine whistle` is not viewed by the diff tool to be trivial (although `tangerine_whistle` and `Tangerine Whistle` are viewed as trivial)

gurukamath:
(closes #572 )

### What was wrong?
Under the current logic, minor differences such as a new line character in the doctoring are treated as non-trivial. See [dao](https://github.com/ethereum/execution-specs/blob/10dcc75108c84f396e61bd7e1b720648f36cd1eb/src/ethereum/dao_fork/utils/address.py#L12) vs [tangerine whistle](https://github.com/ethereum/execution-specs/blob/10dcc75108c84f396e61bd7e1b720648f36cd1eb/src/ethereum/tangerine_whistle/utils/address.py#L12)

Related to Issue #572 

### How was it fixed?
Updated the `meaningful_diffs` calculation as follows

- Treat the old and new documents as a string instead of traversing each of its nodes individually. (This also has a nice side-effect of  calculating the diffs faster - ~2 mins on GitHub Actions)
- Strip `\n`, `\t` and `\s` from both strings since these differences are trivial
- Strip the fork and package names from both strings using regex
- Compare if the two strings are the same

Additionally, I have updated some docstrings of Berlin to port the changes from #566 

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/diff.py
+++ b/src/ethereum_spec_tools/diff.py
@@ -6,7 +6,7 @@ Generates diffs between Ethereum hardforks documentation.
 
 import os.path
 import pickle
-from copy import deepcopy
+import re
 from multiprocessing import Pool
 from typing import Any, Iterator, List, Tuple, TypeVar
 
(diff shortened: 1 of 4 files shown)


src/ethereum_spec_tools/diff.py
```diff
@@ -40,30 +40,23 @@ def find_pickles(path: str, fork: Hardfork) -> Iterator[str]:
...
-    opcode_counter = rstdiff.OpcodeCounter(opcodes)
-    opcode_counter.count()
-
-    return bool(
-        opcode_counter.replace
-        or opcode_counter.delete
-        or opcode_counter.insert
-    )
+    old = re.sub(r"[\n\t\s]", "", str(old))
+    new = re.sub(r"[\n\t\s]", "", str(new))
```
SamWilsn, line 49: Two minor suggestions here: 
  - Compile the regex
  - Convert whitespace to a space, instead of the empty string. `Foo Bar` to `FooBar` is a pretty meaningful change IMO.

```suggestion
    pattern = re.compile(r"[\n\t\s]")  # Can probably move to a package-level constant.
    old = pattern.sub(" ", str(old))
    new = pattern.sub(" ", str(new))
```

src/ethereum_spec_tools/diff.py
```diff
@@ -40,30 +40,23 @@ def find_pickles(path: str, fork: Hardfork) -> Iterator[str]:
...
+
+    for old_word, new_word, flag in trivial_changes:
+        old = re.sub(old_word, "", old, flags=flag)
+        new = re.sub(new_word, "", new, flags=flag)
```
SamWilsn, line 53: `old_word` and `new_word` are escaped?

I assume we're using `re.sub` instead of `str.replace` for the case-insensitivity?

gurukamath: For case insensitivity as well as to identify other corner cases in class names and other places. For example - `dao fork` becomes `dao-fork`, `dao_fork` etc.

src/ethereum_spec_tools/diff.py
gurukamath: Updated.

SamWilsn (approved):

PR #581, opened 2022-08-02
Refactor EVM opcodes

petertdavies:
This PR refactors EVM opcodes, they are now all divided into 4 clearly demarcated sections (`STACK`, `GAS`, `OPERATION`, `PROGRAM COUNTER`). `subtract_gas()` has been replaced by `charge_gas()`, which is less verbose. `extend_memory()` now handles gas charges for memory expansion. Finally common logic between `CALL` and `CALLCODE` has been merged into a single `do_call()` function.

Overall these changes should make the EVM code less verbose and clearer.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/base_types.py
+++ b/src/ethereum/base_types.py
@@ -649,6 +649,8 @@ class U256(FixedUInt):
     inclusive.
     """
 
+    MAX_VALUE: "U256"
+
     __slots__ = ()
 
     @classmethod
--- a/src/ethereum/base_types.py
+++ b/src/ethereum/base_types.py
@@ -15,7 +15,7 @@ Integer and array types which are used by—but not unique to—Ethereum.
 from __future__ import annotations
 
 from dataclasses import replace
-from typing import Any, Callable, Optional, Tuple, Type, TypeVar
+from typing import Any, Callable, ClassVar, Optional, Tuple, Type, TypeVar
 
 U8_MAX_VALUE = (2**8) - 1
 UINT32_MAX_VALUE = (2**32) - 1
@@ -299,7 +299,7 @@ class FixedUInt(int):
     directly, but rather to be subclassed.
     """
 
-    MAX_VALUE: "FixedUInt"
+    MAX_VALUE: ClassVar["FixedUInt"]
 
     __slots__ = ()
 
@@ -649,7 +649,7 @@ class U256(FixedUInt):
     inclusive.
     """
 
-    MAX_VALUE: "U256"
+    MAX_VALUE: ClassVar["U256"]
 
     __slots__ = ()
 
@@ -740,7 +740,7 @@ class Uint32(FixedUInt):
     inclusive.
     """
 
-    MAX_VALUE: "Uint32"
+    MAX_VALUE: ClassVar["Uint32"]
 
     __slots__ = ()
 
@@ -792,7 +792,7 @@ class Uint64(FixedUInt):
     inclusive.
     """
 
-    MAX_VALUE: "Uint64"
+    MAX_VALUE: ClassVar["Uint64"]
 
     __slots__ = ()
 
--- a/src/ethereum/berlin/vm/gas.py
+++ b/src/ethereum/berlin/vm/gas.py
@@ -73,7 +73,7 @@ def charge_gas(evm: Evm, amount: Uint) -> None:
 
     Raises
     ------
-    :py:class:`~ethereum.homestead.vm.exceptions.OutOfGasError`
+    :py:class:`~ethereum.berlin.vm.exceptions.OutOfGasError`
         If `gas_left` is less than `amount`.
     """
     if evm.gas_left < amount:
--- a/src/ethereum/berlin/vm/gas.py
+++ b/src/ethereum/berlin/vm/gas.py
@@ -13,76 +13,76 @@ EVM gas constants and calculators.
 """
 from ethereum.base_types import U256, Uint
 from ethereum.utils.numeric import ceil32
-from ethereum.utils.safe_arithmetic import u256_safe_add
 
+from . import Evm
 from .exceptions import OutOfGasError
 
-GAS_JUMPDEST = U256(1)
-GAS_BASE = U256(2)
-GAS_VERY_LOW = U256(3)
-GAS_STORAGE_SET = U256(20000)
-GAS_STORAGE_UPDATE = U256(5000)
-GAS_STORAGE_CLEAR_REFUND = U256(15000)
-GAS_LOW = U256(5)
-GAS_MID = U256(8)
-GAS_HIGH = U256(10)
-GAS_EXPONENTIATION = U256(10)
-GAS_EXPONENTIATION_PER_BYTE = U256(50)
-GAS_MEMORY = U256(3)
-GAS_KECCAK256 = U256(30)
-GAS_KECCAK256_WORD = U256(6)
-GAS_COPY = U256(3)
-GAS_BLOCK_HASH = U256(20)
-GAS_LOG = U256(375)
-GAS_LOG_DATA = U256(8)
-GAS_LOG_TOPIC = U256(375)
-GAS_CREATE = U256(32000)
-GAS_CODE_DEPOSIT = U256(200)
-GAS_ZERO = U256(0)
-GAS_NEW_ACCOUNT = U256(25000)
-GAS_CALL_VALUE = U256(9000)
-GAS_CALL_STIPEND = U256(2300)
-GAS_SELF_DESTRUCT = U256(5000)
-GAS_SELF_DESTRUCT_NEW_ACCOUNT = U256(25000)
-REFUND_SELF_DESTRUCT = U256(24000)
-GAS_ECRECOVER = U256(3000)
-GAS_SHA256 = U256(60)
-GAS_SHA256_WORD = U256(12)
-GAS_RIPEMD160 = U256(600)
-GAS_RIPEMD160_WORD = U256(120)
-GAS_IDENTITY = U256(15)
-GAS_IDENTITY_WORD = U256(3)
-GAS_RETURN_DATA_COPY = U256(3)
-GAS_FAST_STEP = U256(5)
-GAS_BLAKE2_PER_ROUND = U256(1)
-GAS_COLD_SLOAD = U256(2100)
-GAS_COLD_ACCOUNT_ACCESS = U256(2600)
-GAS_WARM_ACCESS = U256(100)
+GAS_JUMPDEST = Uint(1)
+GAS_BASE = Uint(2)
+GAS_VERY_LOW = Uint(3)
+GAS_STORAGE_SET = Uint(20000)
+GAS_STORAGE_UPDATE = Uint(5000)
+GAS_STORAGE_CLEAR_REFUND = Uint(15000)
+GAS_LOW = Uint(5)
+GAS_MID = Uint(8)
+GAS_HIGH = Uint(10)
+GAS_EXPONENTIATION = Uint(10)
+GAS_EXPONENTIATION_PER_BYTE = Uint(50)
+GAS_MEMORY = Uint(3)
+GAS_KECCAK256 = Uint(30)
+GAS_KECCAK256_WORD = Uint(6)
+GAS_COPY = Uint(3)
+GAS_BLOCK_HASH = Uint(20)
+GAS_LOG = Uint(375)
+GAS_LOG_DATA = Uint(8)
+GAS_LOG_TOPIC = Uint(375)
+GAS_CREATE = Uint(32000)
+GAS_CODE_DEPOSIT = Uint(200)
+GAS_ZERO = Uint(0)
+GAS_NEW_ACCOUNT = Uint(25000)
+GAS_CALL_VALUE = Uint(9000)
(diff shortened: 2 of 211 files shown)


gurukamath:
I am all for these changes. Think it improves readability significantly.

petertdavies:
This is now ready for review. I realise this is an unusually large PR, but these changes inevitably affect most of the VM for every hardfork.

src/ethereum/berlin/vm/instructions/log.py
```diff
@@ -50,37 +43,21 @@ def log_n(evm: Evm, num_topics: U256) -> None:
...
     for _ in range(num_topics):
         topic = pop(evm.stack).to_be_bytes32()
         topics.append(topic)
 
+    # GAS
+    extend_memory(evm, memory_start_index, size)
+    charge_gas(evm, GAS_LOG + GAS_LOG_DATA * size + GAS_LOG_TOPIC * num_topics)
+
+    # OPERATION
+    ensure(not evm.message.is_static, WriteInStaticContext)
```
SamWilsn, line 60: Changing the order of this is fine, because there's no observable difference between an `OutOfGasError` and a `WriteInStaticContext`?

src/ethereum/berlin/vm/instructions/memory.py
```diff
@@ -45,24 +38,18 @@ def mstore(evm: Evm) -> None:
         If `evm.gas_left` is less than
         `3` + gas needed to extend memeory.
     """
-    # convert to Uint as start_position + size_to_extend can overflow.
-    start_position = Uint(pop(evm.stack))
+    # STACK
+    start_position = pop(evm.stack)
     value = pop(evm.stack).to_be_bytes32()
```
SamWilsn, line 43: Should we be doing type conversions in the `STACK` section, or should it be purely `pop`s?

src/ethereum/berlin/vm/instructions/memory.py
```diff
@@ -85,26 +72,19 @@ def mstore8(evm: Evm) -> None:
...
-    )
-    evm.gas_left = subtract_gas(evm.gas_left, total_gas_cost)
+    # GAS
+    extend_memory(evm, start_position, U256(1))
+    charge_gas(evm, GAS_VERY_LOW)
 
-    # extend memory and subtract gas for allocating 32 bytes of memory
-    extend_memory(evm.memory, start_position, U256(1))
+    # OPERATION
+    normalized_bytes_value = (value & U8_MAX_VALUE).to_bytes(1, "big")
```
SamWilsn, line 84: Using the `to_bytes` function seems off in some way. Should we create a `Uint8` type instead?

SamWilsn (commented):
Note to self: you still need to review system, interpreter, memory, and all the precompiles.

Can we remove the safe arithmetic functions after this PR?

src/ethereum/berlin/vm/instructions/environment.py
```diff
@@ -223,43 +257,24 @@ def calldatacopy(evm: Evm) -> None:
...
-        return
-
-    extend_memory(evm.memory, memory_start_index, size)
-
-    value = evm.message.data[data_start_index : data_start_index + size]
-    # But it is possible that data_start_index + size won't exist in evm.data
-    # in which case we need to right pad the above obtained bytes with 0.
-    value = right_pad_zero_bytes(value, size)
+    copy_gas_cost = GAS_COPY * words
+    extend_memory(evm, memory_start_index, size)
```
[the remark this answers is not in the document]
petertdavies: My attitude is that memory extension is strictly a gas accounting phenomenon. From the EVM's point of view memory is an infinitely long buffer initialised with zeros. The only observable effect of memory expansion (to the EVM) is gas consumed.

src/ethereum/berlin/vm/memory.py
```diff
@@ -80,4 +85,27 @@ def memory_read_bytes(
     data_bytes :
         Data read from memory.
     """
-    return memory[start_position : start_position + size]
+    return memory[start_position : Uint(start_position) + Uint(size)]
+
+
+def buffer_read(buffer: Bytes, start_position: U256, size: U256) -> Bytes:
```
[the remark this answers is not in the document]
petertdavies: Naming the `buffer_read()` function is hard. Geth has the impressively vague:

```
// getData returns a slice from the data based on the start and size and pads
// up to size with zero's. This function is overflow safe.
func getData(data []byte, start uint64, size uint64) []byte {
```

We could use `buffer_read()` for memory. The padding is never used for memory because memory extension guarantees that memory reads will be in bounds unless the length is zero.

src/ethereum/berlin/vm/instructions/log.py
petertdavies: Correct. We never distinguish between subclasses of `ExceptionalHalt`.

src/ethereum/berlin/vm/instructions/memory.py
petertdavies: I like it. It's basically a type annotation. `x = pop(evm.stack).to_signed()` tells you that the opcode treats `x` as a signed value.

src/ethereum/berlin/vm/instructions/memory.py
petertdavies: We should probably send the write through the normal `memory_write()` call for consistency so we need to cast to a 1 byte long `Bytes`.

I guess we could use `Bytes([value & U8_MAX_VALUE])`.

src/ethereum/berlin/vm/gas.py
```diff
@@ -95,22 +95,22 @@ def calculate_memory_gas_cost(size_in_bytes: Uint) -> U256:
...
+    total_gas_cost : `ethereum.base_types.Uint`
         The gas cost for storing data in memory.
     """
     size_in_words = ceil32(size_in_bytes) // 32
     linear_cost = size_in_words * GAS_MEMORY
     quadratic_cost = size_in_words**2 // 512
     total_gas_cost = linear_cost + quadratic_cost
     try:
-        return U256(total_gas_cost)
+        return Uint(total_gas_cost)
```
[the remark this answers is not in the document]
petertdavies: Fixed

src/ethereum/berlin/vm/instructions/control_flow.py
```diff
@@ -88,19 +105,24 @@ def jumpi(evm: Evm) -> None:
...
+    # PROGRAM COUNTER
     if conditional_value == 0:
         evm.pc += 1
-        return
-
-    if jump_dest not in evm.valid_jump_destinations:
-        raise InvalidJumpDestError
+    else:
+        if jump_dest not in evm.valid_jump_destinations:
+            raise InvalidJumpDestError
```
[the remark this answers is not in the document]
petertdavies: Agreed

src/ethereum/berlin/vm/instructions/control_flow.py
```diff
@@ -55,9 +65,16 @@ def jump(evm: Evm) -> None:
...
 
+    # GAS
+    charge_gas(evm, GAS_MID)
+
+    # OPERATION
+    pass
+
+    # PROGRAM COUNTER
     if jump_dest not in evm.valid_jump_destinations:
         raise InvalidJumpDestError
```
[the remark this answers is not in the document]
petertdavies: Agreed, fixed

src/ethereum/berlin/vm/instructions/environment.py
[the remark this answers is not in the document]
SamWilsn: I would argue that since two invocations of the same opcode with the same parameters can consume different amounts of gas based on the size of memory that it is an observable effect (and that's discounting [`msize`](https://ethereum.github.io/execution-specs/autoapi/ethereum/berlin/vm/instructions/memory/index.html?highlight=msize#ethereum.berlin.vm.instructions.memory.msize), which makes it directly observable.)

I'm okay leaving as written in this PR though, if you think that's the right call.

src/ethereum/berlin/vm/instructions/environment.py
```diff
@@ -553,15 +540,22 @@ def self_balance(evm: Evm) -> None:
 
     Raises
     ------
+    :py:class:`~ethereum.berlin.vm.exceptions.StackUnderflowError`
+        If `len(stack)` is less than `1`.
     :py:class:`~ethereum.berlin.vm.exceptions.OutOfGasError`
-        If `evm.gas_left` is less than GAS_FAST_STEP.
+        If `evm.gas_left` is less than `20`.
```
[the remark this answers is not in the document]
SamWilsn: Huh. I'll open an issue.

src/ethereum/berlin/vm/memory.py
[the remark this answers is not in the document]
SamWilsn: > Naming the `buffer_read()` function is hard.

Agreed. This seems like a decent PR to fix it in though.

src/ethereum/berlin/vm/instructions/system.py
```diff
@@ -42,110 +42,117 @@
...
+    charge_gas,
     max_message_call_gas,
-    subtract_gas,
 )
 from ..memory import extend_memory, memory_read_bytes, memory_write
 from ..stack import pop, push
 
 
-def create(evm: Evm) -> None:
+def do_create(
```
SamWilsn, line 53: `do_create` isn't particularly descriptive. Perhaps something like `deploy_code`, or `construct`?

src/ethereum/berlin/vm/instructions/system.py
```diff
@@ -290,111 +300,72 @@ def call(evm: Evm) -> None:
...
-        access_gas_cost = GAS_COLD_ACCOUNT_ACCESS
         evm.accessed_addresses.add(to)
+        access_gas_cost = GAS_COLD_ACCOUNT_ACCESS
 
-    is_account_alive = account_exists(
-        evm.env.state, to
-    ) and not is_account_empty(evm.env.state, to)
     create_gas_cost = (
-        U256(0) if is_account_alive or value == 0 else GAS_NEW_ACCOUNT
+        Uint(0) if is_account_alive or value == 0 else GAS_NEW_ACCOUNT
```
SamWilsn, line 326: I'm really not happy with how `black` is formatting `is_account_alive = ...` :rofl: 

Could we do something like:

```suggestion
        Uint(0) if value == 0 or account_alive(evm.env.state, to) else GAS_NEW_ACCOUNT
```

The `value == 0` check is pretty cheap to do, so we should do it first to short circuit when able.

src/ethereum/berlin/vm/instructions/system.py
```diff
@@ -406,104 +377,69 @@ def callcode(evm: Evm) -> None:
...
-        evm.memory, memory_output_start_position, memory_output_size
-    )
-    evm.gas_left = subtract_gas(evm.gas_left, gas_output_memory)
-    extend_memory(evm.memory, memory_output_start_position, memory_output_size)
-    call_data = memory_read_bytes(
-        evm.memory, memory_input_start_position, memory_input_size
-    )
+    extend_memory(evm, memory_input_start_position, memory_input_size)
+    extend_memory(evm, memory_output_start_position, memory_output_size)
+    _account_exists = account_exists(evm.env.state, to)
```
SamWilsn, line 394: I think leading underscores are generally used for unused variables/private fields:

```suggestion
    account_exists_ = account_exists(evm.env.state, to)
```

or maybe more specifically:

```suggestion
    to_exists = account_exists(evm.env.state, to)
```

src/ethereum/berlin/vm/instructions/system.py
petertdavies: I went with `generic_create()`/`generic_call()`.

src/ethereum/berlin/vm/instructions/bitwise.py
```diff
@@ -154,75 +180,91 @@ def get_byte(evm: Evm) -> None:
...
+    # OPERATION
     if shift < 256:
-        shifted_value = signed_value >> shift
+        result = U256.from_signed(signed_value >> shift)
     elif signed_value >= 0:
-        shifted_value = 0
+        result = U256(0)
     else:
-        shifted_value = U256.MAX_VALUE
+        result = U256(U256.MAX_VALUE)
```
SamWilsn, line 265: Ugh, does mypy not pick this up correctly? I thought `MAX_VALUE` was already a `U256`.

src/ethereum/berlin/vm/instructions/system.py
```diff
@@ -42,87 +41,63 @@
...
+    memory_size: U256,
+) -> None:
     """
-    Creates a new account with associated code.
-
-    Parameters
-    ----------
-    evm :
-        The current EVM frame.
+    Core logic used by the `CREATE* family of opcodes.
```
SamWilsn, line 60: Missing a close backtick?

src/ethereum/base_types.py
```diff
@@ -649,6 +649,8 @@ class U256(FixedUInt):
     inclusive.
     """
 
+    MAX_VALUE: "U256"
```
SamWilsn, line 652: \*winces\* Should this be a `ClassVar`?

SamWilsn (approved):

PR #597, opened 2022-09-03
Run all available `ValidBlock` tests

### LINKED ISSUE #594: Support all `ValidBlock` tests
We do support [specific ones](https://github.com/ethereum/execution-specs/blob/cc697f256d281476b9c9f810542e84fb78be2505/tests/berlin/test_state_transition.py#L77). But these are the ones that are not expected to fail/except.

_Originally posted by @gurukamath in https://github.com/ethereum/execution-specs/issues/587#issuecomment-1219037640_

gurukamath:
(closes #594 )

### What was wrong?
Only a small sub-set of the `ValidBlock` test is currently being run

Related to Issue #594 

### How was it fixed?
Some of the `ValidBlock` have Invalid blocks in them. For each hard fork, the following steps were done

1. Independently identify which tests have Invalid Blocks in them. This was done by using `geth`. A test for which `geth` was unable to generate a valid evm trace is assumed to have an invalid block

2. Run separate category of tests - `test_valid_block_incorrect` for tests with invalid blocks and `test_valid_block_correct` for ones with only valid blocks.

3. Identify and mark tests that are slow or use large memory. 

#### Cute Animal Picture

### DIFF
--- a/tests/berlin/test_state_transition.py
+++ b/tests/berlin/test_state_transition.py
@@ -74,42 +74,12 @@ def test_general_state_tests(test_case: Dict) -> None:
 # Run legacy valid block tests
 test_dir = "tests/fixtures/BlockchainTests/ValidBlocks/"
 
-only_in_incorrect = (
+IGNORE_LIST = (
     "bcForkStressTest/ForkStressTest.json",
     "bcGasPricerTest/RPC_API_Test.json",
-    "bcMultiChainTest/CallContractFromNotBestBlock.json",
-    "bcMultiChainTest/ChainAtoChainB_BlockHash.json",
-    "bcMultiChainTest/ChainAtoChainB_difficultyB.json",
-    "bcMultiChainTest/ChainAtoChainB.json",
-    "bcMultiChainTest/ChainAtoChainBCallContractFormA.json",
-    "bcMultiChainTest/ChainAtoChainBtoChainA.json",
-    "bcMultiChainTest/ChainAtoChainBtoChainAtoChainB.json",
-    "bcTotalDifficultyTest/lotsOfBranchesOverrideAtTheEnd.json",
-    "bcTotalDifficultyTest/lotsOfBranchesOverrideAtTheMiddle.json",
-    "bcTotalDifficultyTest/lotsOfLeafs.json",
-    "bcTotalDifficultyTest/newChainFrom4Block.json",
-    "bcTotalDifficultyTest/newChainFrom5Block.json",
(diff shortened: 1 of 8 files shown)


petertdavies:
Is there a way of distinguishing between valid and invalid blocks, other that running them with Geth and seeing if they fail?

If not, we should consider opening a bug upstream, since the tests should provide an expected result.

gurukamath:
> If not, we should consider opening a bug upstream, since the tests should provide an expected result.

Tracking [upstream issue](https://github.com/ethereum/tests/issues/1078)

gurukamath:
@SamWilsn @petertdavies Based on the response on the issue that I created on the ethereum tests repo, looks like these aren't necessarily tests with Invalid Blocks. Rather these include multi chain tests/re-orgs. Please see [details](https://github.com/ethereum/tests/issues/1078#issuecomment-1240802291). Do you think we should simply ignore these tests for specs since they might not be relevant?

petertdavies:
Tests involving reorgs are out of scope for us.

Based on the names, I doubt the tests in `bcMultiChainTest` and `bcTotalDifficultyTest` are applicable to us.

gurukamath:
> Tests involving reorgs are out of scope for us.
> 
> Based on the names, I doubt the tests in `bcMultiChainTest` and `bcTotalDifficultyTest` are applicable to us.

Updated the valid block tests to ignore unrelated tests including `bcMultiChainTest` and `bcTotalDifficultyTest`

tests/berlin/test_state_transition.py
```diff
@@ -74,27 +74,33 @@ def test_general_state_tests(test_case: Dict) -> None:
...
+        slow_list=SLOW_TESTS,
+    ),
     ids=idfn,
 )
-def test_uncles_correctness(test_case: Dict) -> None:
-    run_berlin_blockchain_st_tests(test_case)
+def test_valid_block_tests(test_case: Dict) -> None:
+    try:
+        run_berlin_blockchain_st_tests(test_case)
+    except KeyError:
```
SamWilsn, line 101: Is there some way we can make these `except` clauses more specific? Maybe `xfail` closer to the source, or rethrowing with a more specific error type?

I'm worried that this will hide all `KeyError` exceptions as xfailed, when they might actually be real failures.

gurukamath: This will have to be done not only for `ValidBlock` tests but also for other tests as well. I have opened #617 for this.

SamWilsn (approved):

SamWilsn:
tests passed locally.

PR #599, opened 2022-09-06
Fix handling of minimum difficulty

petertdavies:
### What was wrong?

The `GENESIS_DIFFICULTY` was misnamed as it is not the difficulty at genesis, but instead the minimum difficulty allowed. Additionally there is inconsistently about whether the difficulty is raised to the minimum first or the difficulty bomb is applied first. The Yellow Paper, Nethermind and Akula apply the difficulty bomb first, whereas Geth, Erigon, Besu raise the difficulty to the minimum first. The situation is entirely moot because the minimum difficulty is not reached outside the testsuite.

### How was it fixed?

* Renamed `GENESIS_DIFFICULTY` to `MINIMUM_DIFFICULTY`.
* Arbitrarily implement the Yellow Paper/Nethermind/Akula approach, with an explanatory comment.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/berlin/spec.py
+++ b/src/ethereum/berlin/spec.py
@@ -65,7 +65,7 @@ from .vm.interpreter import process_message_call
 BLOCK_REWARD = U256(2 * 10**18)
 GAS_LIMIT_ADJUSTMENT_FACTOR = 1024
 GAS_LIMIT_MINIMUM = 5000
-GENESIS_DIFFICULTY = Uint(131072)
+MINIMUM_DIFFICULTY = Uint(131072)
 MAX_OMMER_DEPTH = 6
 BOMB_DELAY_BLOCKS = 9000000
 EMPTY_OMMER_HASH = keccak256(rlp.encode([]))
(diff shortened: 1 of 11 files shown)


src/ethereum/frontier/spec.py
```diff
@@ -860,12 +860,10 @@ def calculate_block_difficulty(
     Computes difficulty of a block using its header and
     parent header.
 
-    Since the genesis block is the first on the chain and has no parent,
-    its difficulty is a predefined value stored as ``GENESIS_DIFFICULTY``.
-    For all other blocks the difficulty is determined by the time the
-    block was created after its parent. If a block's timestamp is more
-    than 13 seconds after its parent block then its difficulty is set
-    as the difference between the parent's difficulty and the
+    The difficulty of a block is determined by the time the block was
```
SamWilsn, line 863: I think you should still note that the genesis block is a special case somewhere, and that its difficulty is `MINIMUM_DIFFICULTY`.

whitelist.txt
```diff
@@ -355,4 +355,6 @@ Transaction2930
 
 downloader
 Downloader
-sublist
\ No newline at end of file
+sublist
+
+testnets
```
SamWilsn, line 360:
```suggestion
```

We're the specification for mainnet. If we also happen to be useful to define testnets, so be it, but we should explicitly focus on mainnet.

src/ethereum/frontier/spec.py
petertdavies: The difficulty of the genesis block is bigger than `MINIMUM_DIFFICULTY`. The fact that the constant was previously called `GENESIS_DIFFICULTY` was an outright bug.

src/ethereum/berlin/spec.py
```diff
@@ -1002,10 +1002,11 @@ def calculate_block_difficulty(
     # bomb has no effect prior to block 200000 we pretend it existed from
     # genesis.
     # See https://github.com/ethereum/go-ethereum/pull/1588
-    num_bomb_periods = ((int(block_number) - BOMB_DELAY_BLOCKS) // 100000) - 2
+    num_bomb_periods = (int(block_number) - BOMB_DELAY_BLOCKS) // 100000 - 2
```
SamWilsn, line 1005: I know there's no ambiguity with the new version, but I'd prefer to keep the extra parentheses.

SamWilsn (approved):

PR #600, opened 2022-09-06
Refactor balance check (closes #587)

### LINKED ISSUE #587: Sender balance may not be checked when sending a value
### What was wrong?

> Hi, I've been reading the berlin execution specs in conjunction with the berlin yellow paper and noticed a discrepancy. The execution-specs only invalidate the block if the sender balance can cover the up-front gas_fee:
https://github.com/ethereum/execution-specs/blob/cc697f256d281476b9c9f810542e84fb78be2505/src/ethereum/berlin/spec.py#L599-L601 
However, in the YP (and client implementations), the tx.value is also added to the gas_fee for this check: https://github.com/ethereum/yellowpaper/blob/7251f10f25fc2a340c51c57e4a71970b525c0691/Paper.tex#L708-L711
>
> A check is done later as part of the EVM https://github.com/ethereum/execution-specs/blob/cc697f256d281476b9c9f810542e84fb78be2505/src/ethereum/berlin/vm/interpreter.py#L239-L243 which can raise an InsufficientFunds exception.
> However I believe this results in an exception which sets the gas to 0, which seems like a significant difference compared to the transaction being recognised as invalid earlier.
>
> Have I missed something here?

### Additional Context

[Reported by siladu on Discord](https://discord.com/channels/595666850260713488/856943705209569291/1008983249445474304).

SamWilsn:
Consider this a proof-of-concept.

### What was wrong?

Related to Issue #587 

### How was it fixed?

Moved balance check to `process_transaction`, removed `InsufficientFunds` from the vm.

#### Cute Animal Picture

photo_2022-09-06_18-17-31

### DIFF
--- a/src/ethereum/berlin/spec.py
+++ b/src/ethereum/berlin/spec.py
@@ -598,7 +598,7 @@ def process_transaction(
     sender_account = get_account(env.state, sender)
     gas_fee = tx.gas * tx.gas_price
     ensure(sender_account.nonce == tx.nonce, InvalidBlock)
-    ensure(sender_account.balance >= gas_fee, InvalidBlock)
+    ensure(sender_account.balance >= gas_fee + tx.value, InvalidBlock)
     ensure(sender_account.code == bytearray(), InvalidBlock)
 
     gas = tx.gas - calculate_intrinsic_cost(tx)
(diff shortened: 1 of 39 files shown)


petertdavies (commented):
LGTM. Approve subject to backporting.

petertdavies (approved):
LGTM!

Adding manual tests like this isn't sustainable in the long term, but as a stop gap measure it is fine.

PR #606, opened 2022-09-18
Add Paris final details

axic:


network-upgrades/mainnet-upgrades/paris.md
```diff
@@ -39,7 +39,7 @@ Once the proof-of-work to proof-of-stake transition has completed, an additional
 | Ropsten | N/A | N/A | N/A | 
 | Sepolia | 1735371 | August 17, 2022 | `0xb96cbd13` | 
 | Goerli  | TBD | TBD | TBD | 
-| Mainnet | TBD | TBD | TBD |  
+| Mainnet | 15537394 | September 15, 2022 | TBD |
```
petertdavies, line 42: This is a table of `FORK_NEXT` blocks, but the number you have put in it is block number of the first POS block. The entry should remain "TBD" since we haven't decided when mainnet `FORK_NEXT` will happen.

petertdavies (changes requested):

README.md
```diff
@@ -11,7 +11,7 @@ This repository contains the consensus specifications related to the Ethereum ex
 | Version and Code Name | Block No. | Released | Incl EIPs | Specs | Blog |
 |-----------------------|-----------|----------|-----------|-------|-------|
 | Shanghai | TBD | TBD | TBD| [Specification](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md) | |
-| Paris | TBD | TBD | TBD| [Specification](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/paris.md) | |
+| Paris | 15537394 | 2022-09-15 | [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) <br/> [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399) | [Specification](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/paris.md) | [Blog](https://blog.ethereum.org/2022/08/24/mainnet-merge-announcement) |
```
timbeiko, line 14: Hmm - I wonder if we should add the block here or use the TTD instead? It might be confusing to have the block number if we then add the `mergeNetSplit` block as we did on Sepolia?

network-upgrades/mainnet-upgrades/paris.md
timbeiko: +1

README.md
djrtwo: Once the merge happens and is finalized, a node really shouldn't use pow TTD or block number to *sync* to the head given that alternative histories at that TTD or block number could be crafted.

assuming you have direction from the beacon chain, you can premise the logic transition upon TTD or EL block number.

I don't see much issue putting block number here, as long as it is deeply known that to safely sync you need a beacon chain.

timbeiko: I guess the easy fix is to just put both here? Something like this, even if it's really long?
```suggestion
| Paris | 15537394 (TTD: 58750000000000000000000) | 2022-09-15 | [EIP-3675](https://eips.ethereum.org/EIPS/eip-3675) <br/> [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399) | [Specification](https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/paris.md) | [Blog](https://blog.ethereum.org/2022/08/24/mainnet-merge-announcement) |
```

axic: > I don't see much issue putting block number here, as long as it is deeply known that to safely sync you need a beacon chain.

I don't think this table here ever was used for syncing, given it doesn't record block hashes or anything of bigger certainty.

axic:
@timbeiko @djrtwo how about this updated version?

timbeiko:
LGTM, let's merge this for now (given how long it's already been since The Merge) and if we want to bikeshed over having the TTD there, we can do it in another PR :-)

axic:
> LGTM, let's merge this for now (given how long it's already been since The Merge) and if we want to bikeshed over having the TTD there, we can do it in another PR :-)

I think a bigger revision to how all this data is presented is needed once the new rollout strategy is agreed on, as there may be 3 different ways by then. Adding that many columns will look weird :)

Issue #609, opened 2022-09-21
Arbitrary File Write via Archive Extraction (Tar Slip)


bhaskarvilles:
Unsanitized input from a command line argument flows into extractall, where it is used to extract files from a tar archive. This may result files outside destination directory to be overwritten, resulting in an arbitrary file write.

https://github.com/ethereum/execution-specs/blob/5e8a9c33b4fa607cac58e1a92e186c1ef4f0230a/scripts/download_geth_linux.py#L31

### 10 steps in 1 file
`scripts/download_geth_linux.py`

16:12 `parser.parse_argsparser.parse_args` (SOURCE)
1 - 2 `parser.parse_args` (SOURCE)
1
`parser.parse_args`
2
`args`
3
`argsargs.dir`
4 - 5
`args`
4
`args.dir`
5
`dir: str`
6
`dir`
7
`dir`
8
`dirgeth_tar.extractall`
9 - 10
`dir`
9
`geth_tar.extractall`

petertdavies:
The `geth_tar` file is trusted (indeed we execute the program it contains). The fact that `extract_all()` can misbehave on malformed tar files is irrelevant.

PR #615, opened 2022-09-27
Add protection against arbitrary file write via archive extraction

fredriksvantes:
### What was wrong?
Script was in theory vulnerable to arbitrary file write via archive extraction, but unlikely in practice given that this script is downloading a fixed tar and using github actions.

#### Cute Animal Picture

### DIFF
--- a/scripts/download_geth_linux.py
+++ b/scripts/download_geth_linux.py
@@ -3,8 +3,54 @@ import argparse
 import os
 import shutil
 import tarfile
-
 import requests
+from os.path import abspath
+from os.path import dirname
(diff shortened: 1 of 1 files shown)


SamWilsn (changes requested):
I think I'd just prefer checking a hard coded SHA of the tarball before extracting. Should be less code, and protect against more attacks.

As mentioned in #609, we run a binary from the extracted tarball so even if we clean all the paths, a malicious binary would still have arbitrary writes.

petertdavies:
This PR appears to be abandoned.

PR #620, opened 2022-10-05
Paris (aka The Merge)

petertdavies:
This PR implements the Paris hardfork. I have few minor comments:

* The test `correct_Merge` in `InvalidBlocks` has become invalid because it has uncles contrary to the specification. Meanwhile the test `DifficultyIsZero_Merge` has become valid, because it is not longer invalid for difficulty to be zero.
* I have renamed `mix_digest` to `prev_randao` which is the name that field has in the consensus specs. The field continues to be called `mixHash` in JSON serialisations of blocks. 
* The test helpers need to distinguish between POW and POS hardforks in order to disable mocking seal validation functions that don't exist post merge. Currently, I have put `== "paris"` as a test, but that won't be sustainable long term. 

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/paris/__init__.py
+++ b/src/ethereum/paris/__init__.py
@@ -1,0 +1,8 @@
+"""
+Ethereum Paris Hardfork
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The Eleventh Ethereum hardfork.
+"""
+
+MAINNET_FORK_BLOCK = 15537394
--- a/src/ethereum/paris/spec.py
+++ b/src/ethereum/paris/spec.py
@@ -13,18 +13,17 @@ Entry point for the Ethereum specification.
 """
 
 from dataclasses import dataclass
-from typing import List, Optional, Set, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 from ethereum.base_types import Bytes0
 from ethereum.crypto.elliptic_curve import SECP256K1N, secp256k1_recover
 from ethereum.crypto.hash import keccak256
-from ethereum.ethash import dataset_size, generate_cache, hashimoto_light
 from ethereum.exceptions import InvalidBlock
 from ethereum.utils.ensure import ensure
 
 from .. import rlp
-from ..base_types import U256, U256_CEIL_VALUE, Bytes, Uint, Uint64
-from . import MAINNET_FORK_BLOCK, vm
+from ..base_types import U256, Bytes, Uint, Uint64
+from . import vm
 from .bloom import logs_bloom
 from .eth_types import (
     TX_ACCESS_LIST_ADDRESS_COST,
@@ -51,7 +50,6 @@ from .eth_types import (
 from .state import (
     State,
     account_exists,
-    create_ether,
     destroy_account,
     get_account,
     increment_nonce,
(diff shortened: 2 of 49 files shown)


timbeiko:
giphy (1)

tests/helpers/load_state_tests.py
```diff
@@ -378,6 +378,7 @@ def run_blockchain_st_test(test_case: Dict, load: BaseLoad) -> None:
     )
 
     if not test_data["ignore_pow_validation"] or load.fork_module == "paris":
+        # FIXME: Test for a POS hardfork properly, rather than `== "paris"`
```
gurukamath, line 381: What does the `sealEngine` field represent for post merge fork tests?

src/ethereum/paris/spec.py
```diff
@@ -0,0 +1,848 @@
...
+    # transfer miner fees
+    coinbase_balance_after_mining_fee = (
+        get_account(env.state, env.coinbase).balance + transaction_fee
+    )
+    if coinbase_balance_after_mining_fee != 0:
+        set_account_balance(
+            env.state, env.coinbase, coinbase_balance_after_mining_fee
+        )
+    elif is_account_empty(env.state, env.coinbase):
+        destroy_account(env.state, env.coinbase)
```
SamWilsn, line 520: It is not obvious to me why we're changing this behaviour, especially when EIP-3675 says:

> Note: Transaction fee mechanics affecting the block’s beneficiary account MUST remain unchanged.

src/ethereum/paris/trie.py
```diff
@@ -0,0 +1,469 @@
...
+    MutableMapping,
+    Optional,
+    Sequence,
+    TypeVar,
+    Union,
+    cast,
+)
+
+from ethereum.crypto.hash import keccak256
+from ethereum.london import trie as previous_trie
```
SamWilsn, line 33: Shouldn't this be `gray_glacier`? I guess the linter didn't get to run yet :rofl:

SamWilsn (changes requested):
- `create_ether` looks unused.

src/ethereum/paris/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
+Ethereum Paris Hardfork
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The Eleventh Ethereum hardfork.
+"""
+
+MAINNET_FORK_BLOCK = 15537394
```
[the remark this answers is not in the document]
SamWilsn: As long as there's a hefty docstring describing this, I'm okay leaving this as a block number.

```python
MAINNET_FORK_BLOCK = ...
"""
I think this works?
"""
```

src/ethereum/paris/spec.py
SamWilsn: Should we open a bug for this then?

tests/helpers/load_state_tests.py
petertdavies: `sealEngine` is set to `NoProof` everywhere, except in `LegacyTests`. The problem is that our response to seeing `NoProof` is to try and mock a bunch of POW validation functions that don't exist in `paris` and later forks.

petertdavies:
Even though the `create_ether` function is unused, I have decided to keep it since it will become used again by withdrawals in Shanghai.

SamWilsn (approved):

src/ethereum/paris/spec.py
petertdavies: I've backported the change all the way to the introduction of account clearing in `spurious_dragon`.

Issue #624, opened 2022-10-10
Add support for EVM trace specification 


petertdavies:
In order support EVM debugging tools we need to emit traces in the format specified by [EIP-3155: EVM trace specification](https://eips.ethereum.org/EIPS/eip-3155).

PR #627, opened 2022-10-15
Memory `extend_memory` re-factor

gurukamath:
The changes have only been implemented in Frontier. Will be applied to the other forks post feedback.

### What was wrong?
1.  The `extend_memory` function and the `calculate_gas_extend_memory` function perform mostly the same calculations.
2. The `extend_memory` gas is charged separately from the other gas charges involved. It is better to calculate the overall gas costs and then deduct it in one go (which also makes it easier to implement an EVM Trace see #624 )

### How was it fixed?
Remove the `extend_memory` function and allow the `calculate_gas_extend_memory` function to return the gas_cost as well as the size to be extended

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/frontier/vm/instructions/memory.py
+++ b/src/ethereum/frontier/vm/instructions/memory.py
@@ -14,8 +14,13 @@ Implementations of the EVM Memory instructions.
 from ethereum.base_types import U8_MAX_VALUE, U256, Bytes
 
 from .. import Evm
-from ..gas import GAS_BASE, GAS_VERY_LOW, charge_gas
-from ..memory import extend_memory, memory_read_bytes, memory_write
+from ..gas import (
+    GAS_BASE,
+    GAS_VERY_LOW,
+    calculate_gas_extend_memory,
+    charge_gas,
+)
+from ..memory import memory_read_bytes, memory_write
 from ..stack import pop, push
 
 
@@ -36,10 +41,14 @@ def mstore(evm: Evm) -> None:
     value = pop(evm.stack).to_be_bytes32()
 
     # GAS
-    extend_memory(evm, start_position, U256(len(value)))
-    charge_gas(evm, GAS_VERY_LOW)
+    extend_memory = calculate_gas_extend_memory(
+        evm.memory, [(start_position, U256(len(value)))]
+    )
+
+    charge_gas(evm, GAS_VERY_LOW + extend_memory.cost)
 
     # OPERATION
+    evm.memory += b"\x00" * extend_memory.expand_by
     memory_write(evm.memory, start_position, value)
 
     # PROGRAM COUNTER
@@ -63,10 +72,14 @@ def mstore8(evm: Evm) -> None:
     value = pop(evm.stack)
 
     # GAS
-    extend_memory(evm, start_position, U256(1))
-    charge_gas(evm, GAS_VERY_LOW)
+    extend_memory = calculate_gas_extend_memory(
+        evm.memory, [(start_position, U256(1))]
+    )
+
+    charge_gas(evm, GAS_VERY_LOW + extend_memory.cost)
 
     # OPERATION
+    evm.memory += b"\x00" * extend_memory.expand_by
     normalized_bytes_value = Bytes([value & U8_MAX_VALUE])
     memory_write(evm.memory, start_position, normalized_bytes_value)
 
@@ -88,10 +101,13 @@ def mload(evm: Evm) -> None:
     start_position = pop(evm.stack)
 
     # GAS
-    extend_memory(evm, start_position, U256(32))
-    charge_gas(evm, GAS_VERY_LOW)
+    extend_memory = calculate_gas_extend_memory(
+        evm.memory, [(start_position, U256(32))]
+    )
+    charge_gas(evm, GAS_VERY_LOW + extend_memory.cost)
 
     # OPERATION
+    evm.memory += b"\x00" * extend_memory.expand_by
     value = U256.from_be_bytes(
         memory_read_bytes(evm.memory, start_position, U256(32))
     )
--- a/src/ethereum/london/vm/gas.py
+++ b/src/ethereum/london/vm/gas.py
@@ -77,6 +77,21 @@ class ExtendMemory:
     size: Uint
 
 
+@dataclass
+class MessageCallGas:
+    """
+    Define the gas cost and stipend for executing the call opcodes.
+
+    `cost`: `ethereum.base_types.Uint`
+        The gas amount for executing the Opcodes.
+    `stipend`: `ethereum.base_types.Uint`
+        The gas stipend for executing the Opcodes.
+    """
+
+    cost: Uint
+    stipend: Uint
+
+
 def charge_gas(evm: Evm, amount: Uint) -> None:
     """
     Subtracts `amount` from `evm.gas_left`.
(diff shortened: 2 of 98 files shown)


src/ethereum/frontier/vm/instructions/memory.py
```diff
@@ -43,8 +48,12 @@ def mstore(evm: Evm) -> None:
...
 
     # GAS
-    extend_memory(evm, start_position, U256(len(value)))
-    charge_gas(evm, GAS_VERY_LOW)
+    memory_cost, size_to_extend = calculate_gas_extend_memory(
+        evm.memory, [(start_position, U256(len(value)))]
+    )
+
+    charge_gas(evm, GAS_VERY_LOW + memory_cost)
+    evm.memory += b"\x00" * size_to_extend
```
petertdavies, line 56: I think `evm.memory += b"\x00" * size_to_extend` belongs in the `# OPERATION` section.

petertdavies (changes requested):
This is a sensible approach. I support these changes.

gurukamath:
I have updated the commit as discussed in the call. Applied to London hard fork. There are a couple of differences with the earlier implementation for Frontier. Specifically `calculate_call_gas_cost` and `calculate_message_call_gas_stipend`

gurukamath:
Also, perhaps the call gas and stipend calculations should be part of the same function. Your thoughts?

SamWilsn:
> Also, perhaps the call gas and stipend calculations should be part of the same function. Your thoughts?

And returned in a tuple? I think I prefer avoiding complex return types where possible, so maybe we keep the separate functions.

src/ethereum/london/vm/gas.py
```diff
@@ -159,10 +174,10 @@ def calculate_call_gas_cost(
     call_gas_cost: `ethereum.base_types.Uint`
         The total gas amount for executing Opcodes `CALL` and `CALLCODE`.
     """
-    if gas_left < extra_gas:
-        raise OutOfGasError
+    if gas_left < extra_gas + memory_cost:
+        return gas + extra_gas
 
-    gas = min(gas, max_message_call_gas(gas_left - extra_gas))
+    gas = min(gas, max_message_call_gas(gas_left - extra_gas - memory_cost))
```
SamWilsn, line 180: Looking at this again, I think the new version is less understandable. Could you walk me through what's going on here and why we're making this particular change?

gurukamath: The main idea was that we want to to be able to deduct gas in one single go (inside `charge_gas`) in order to be able to implement the proper `evm_trace`. For that, we need to calculate the `memory_cost` but not deduct it from the `evm.gas_left` right away (which is what we currently). This step is to account for that fact i.e. the `gas_left` is effectively `gas_left - memory_cost`

SamWilsn: Ok... that makes sense. So what about the line that used to `raise OutOfGasError`?

gurukamath: `OutOgGasError` will now be called in the `charge_gas` function

petertdavies:
As discussed on the weekly standup call, I think that a rule that `charge_gas()` should be called only once, at the end of the `# GAS` section should be observed. This lines up with how real clients deal with it and could lead to subtle semantic issues if we call `charge_gas()` twice. Also we need to record the gas usage with a tracing hook and `charge_gas()` is a convenient place to do that.

I also think that `calculate_call_gas_cost()` and `calculate_message_call_gas_stipend()` should be merged into a single function. They currently implicitly duplicate each others logic and it is quite hard to understand what is going on with them separate.

src/ethereum/london/vm/gas.py
gurukamath: Also, I will change `if gas_left < extra_gas + memory_cost:` to `if gas_left - memory_cost < extra_gas:` which I think makes it a bit clearer

gurukamath:
@SamWilsn @petertdavies I have made the following two changes
1. Introduced the `ExtendMemory` class to hold the cost and size of memory extensions. Makes the code more readable without having to return tuples
2. Introduces the `MessageCallGas` class to hold the call gas cost and stipend. This allows us to combine the `calculate_call_gas_cost` and `calculate_message_call_gas_stipend` functions which perform mostly the same computations. The code for the call opcodes also more readable.

Open to any suggestions on names for these classes/attributes :)

src/ethereum/london/vm/gas.py
```diff
@@ -104,79 +137,55 @@ def calculate_memory_gas_cost(size_in_bytes: Uint) -> Uint:
...
+        size_to_extend += after_size - before_size
+        already_paid = calculate_memory_gas_cost(before_size)
+        total_cost = calculate_memory_gas_cost(after_size)
+        to_be_paid += total_cost - already_paid
 
-    gas = min(gas, max_message_call_gas(gas_left - extra_gas))
+        current_size = after_size
 
-    return gas + extra_gas
+    return ExtendMemory(to_be_paid, size_to_extend)
```
SamWilsn, line 175: Probably a dumb question, but this isn't equal to the cost of just expanding for the largest `after_size` out of all the `extensions`, is it?

gurukamath: In most cases, yeah. But not always. If the `extension` with the largest `after_size` has a `size` of 0 (i.e. it has a very large `start_position`), the memory is not to be extended.

src/ethereum/london/vm/gas.py
```diff
@@ -59,6 +62,36 @@
...
 
+@dataclass
+class ExtendMemory:
+    """
+    Define the parameters for memory extension in opcodes
+
+    `cost`: `ethereum.base_types.Uint`
+        The gas required to perform the extension
+    `size`: `ethereum.base_types.Uint`
+        The size by which the memory will be extended
```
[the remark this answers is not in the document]
gurukamath: The idea was that this attribute will always be called on an instance of this class. Like `extend_memory.cost` and `extend_memory.size` which I thought made it verbose enough.

That said, I see the point in making it explicit here. Do you think the `cost` attribute also needs a rename?

src/ethereum/london/vm/gas.py
```diff
@@ -59,6 +62,36 @@
...
+    """
+
+    cost: Uint
+    size: Uint
```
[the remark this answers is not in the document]
gurukamath: I see that the other docstrings use the former syntax. If we have to change the syntax, I guess it will be a good idea to create a new issue and change the syntax for all the docstrings.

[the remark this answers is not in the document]
gurukamath: Opened a new issue #671

src/ethereum/london/vm/gas.py
```diff
@@ -59,6 +62,36 @@
...
+
+@dataclass
+class MessageCallGas:
+    """
+    Define the gas cost and stipend for executing the call opcodes.
+
+    `cost`: `ethereum.base_types.Uint`
+        The gas amount for executing the Opcodes.
+    `stipend`: `ethereum.base_types.Uint`
+        The gas stipend for executing the Opcodes.
```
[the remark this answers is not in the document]
gurukamath: Updated

src/ethereum/london/vm/gas.py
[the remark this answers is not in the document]
gurukamath: Updated to `expand_by`

gurukamath:
@SamWilsn Have updated the code to align with the review feedback (more detailed docstring for `MessageCallGas` and change the `ExtendMemory` attribute to `expand_by`)

SamWilsn (approved):

gurukamath:
@SamWilsn Have re-based the London changes on top of the latest master and ported the changes to the other forks

SamWilsn:
I'm getting this from the linter:

```
ethereum.arrow_glacier - GlacierForksHygiene:
	 the item `MessageCallOutput` in `.vm.interpreter` has changed.Glacier forks may only differ in difficulty block.
	 collect_accounts_to_delete in .vm.interpreter has been added.Glacier forks may only differ in difficulty block.
	 the item `process_message_call` in `.vm.interpreter` has changed.Glacier forks may only differ in difficulty block.
	 the item `execute_code` in `.vm.interpreter` has changed.Glacier forks may only differ in difficulty block.
	 collect_touched_accounts in .vm.interpreter has been added.Glacier forks may only differ in difficulty block.
	 calculate_gas_refund in .vm.interpreter has been added.Glacier forks may only differ in difficulty block.
	 the item `generic_call` in `.vm.instructions.system` has changed.Glacier forks may only differ in difficulty block.
	 the item `selfdestruct` in `.vm.instructions.system` has changed.Glacier forks may only differ in difficulty block.
	 the item `generic_create` in `.vm.instructions.system` has changed.Glacier forks may only differ in difficulty block.
	 account_exists_and_is_empty in .state has been deleted.Glacier forks may only differ in difficulty block.
	 the item `process_transaction` in `.spec` has changed.Glacier forks may only differ in difficulty block.
	 the item `Evm` in `.vm` has changed.Glacier forks may only differ in difficulty block.
	 RIPEMD160_ADDRESS in .vm has been deleted.Glacier forks may only differ in difficulty block.
	 incorporate_child_on_success in .vm has been deleted.Glacier forks may only differ in difficulty block.
	 incorporate_child_on_error in .vm has been deleted.Glacier forks may only differ in difficulty block.
Total diagnostics: 15
```

gurukamath:
> I'm getting this from the linter:
> 
> ```
> ethereum.arrow_glacier - GlacierForksHygiene:
> ```

Yes. This is because Peter's re-factor is already in `London` but not ported over to `Arrow Glacier` yet. Should be fixed once he ports his changes to the other forks.

PR #629, opened 2022-10-21
Update tests - `NoPostState` and `bigmem`

### LINKED ISSUE #617: Handle `xfail` tests with a more specific error
Is there some way we can make these `except` clauses more specific? Maybe `xfail` closer to the source, or rethrowing with a more specific error type?

I'm worried that this will hide all `KeyError` exceptions as xfailed, when they might actually be real failures.

_Originally posted by @SamWilsn in https://github.com/ethereum/execution-specs/pull/597#discussion_r983894651_

### LINKED ISSUE #619: Homestead tests cause the Github worker to crash
See [action](https://github.com/ethereum/execution-specs/actions/runs/3154105570/jobs/5131332236)

Investigate the cause. Eventually the tests can be marked as `bigmem` to prevent them from running in parallel.

gurukamath:
(closes #619 )
(closes #617 )

### What was wrong?
This PR resolves the following issues related to tests
1. Some tests cause  the Github CI worker to crash. 
2. Currently, the tests are configured to `xfail` upon `KeyError` which could potentially hide actual `KeyError`

Related to Issue #617 #619

### How was it fixed?
1. Marked the responsible tests as `bigmem`
2. Raised a more specific error when the post state is missing

#### Cute Animal Picture

### DIFF
--- a/tests/helpers/load_state_tests.py
+++ b/tests/helpers/load_state_tests.py
@@ -24,6 +24,19 @@ from ethereum.utils.hexadecimal import (
 )
 
 
+class NoTestsFound(Exception):
+    """
+    An exception thrown when the test for a particular fork isn't
+    available in the json fixture
+    """
+
+
+class NoPostState(Exception):
+    """
+    An exception thrown when the test does not have a postState defined.
+    """
+
+
 class BaseLoad(ABC):
     @property
     @abstractmethod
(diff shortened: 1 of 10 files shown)


tests/helpers/load_state_tests.py
```diff
@@ -335,6 +348,11 @@ def load_test(test_case: Dict, load: BaseLoad) -> Dict:
         json_data["blocks"]
     )
 
+    try:
+        post_state = load.json_to_state(json_data["postState"])
+    except KeyError:
+        raise NoPostState
```
SamWilsn, line 354: Not a huge blocker or anything, but this'll catch `KeyError`s from inside `json_to_state` as well. If the JSON format changes, we'll get xfails which might not be noticed.

I'd do something like:

```suggestion
    try:
        raw_post_state = json_data["postState"]
    except KeyError:
        raise NoPostState
    post_state = load.json_to_state(raw_post_state)
```
```

SamWilsn (approved):

PR #635, opened 2022-10-29
Create diff tests for minor hardforks

### LINKED ISSUE #605: Create diff tests for glacier hardforks
The Dao Fork and the three difficulty bomb hardforks are extremely minor and don't come with tests. Rather than implementing tests for them we should add diff tests that make sure that the diffs between those forks and their predecessors are limited to the expected changes.

gurukamath:
(closes #605 )

### What was wrong?
The minor hard forks do not currently have any tests.

Related to Issue #605 

### How was it fixed?
Create diff tests control for diffs in the hard fork source code, with the previous fork as the reference.

#### Cute Animal Picture


tests/helpers/diff_test_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+    if not os.path.isfile(path):
+        open(name, "a").close()
+        return
+
+    with open(path) as f:
+        code = f.read()
+
+    code = re.sub(re.compile('""".*?"""', re.DOTALL), "", code)
+    code = re.sub(re.compile("#.*?\n"), "\n", code)
+    code = re.sub(re.compile("\n\s*\n"), "\n", code)
```
SamWilsn, line 58: Haven't tested, but I think you can replace the `compile` with just a plain string:

```suggestion
    code = re.sub('""".*?"""', "", code, flags=re.DOTALL)
    code = re.sub("#.*?\n", "\n", code)
    code = re.sub("\n\s*\n", "\n", code)
```

tests/helpers/diff_test_helpers.py
```diff
@@ -0,0 +1,147 @@
...
+            new_processed_path = os.path.join(self.found_diffs_path, "new.py")
+
+            process_source_code(old_source_path, old_processed_path)
+            process_source_code(new_source_path, new_processed_path)
+
+            diffs = subprocess.run(
+                ["diff", old_processed_path, new_processed_path],
+                capture_output=True,
+                text=True,
+            )
```
SamWilsn, line 109: Hm... do you think it would be possible to implement a solution using Python's [`ast`](https://docs.python.org/3/library/ast.html)? You could walk the tree and remove docstrings and comments, plus whitespace would already be taken care of. Wouldn't need to shell out to `diff` that way either, or write files.

Apparently [there's no built-in comparison function](https://github.com/python/cpython/issues/60191) but [there's a post with some ideas on how to do it](https://stackoverflow.com/questions/3312989/elegant-way-to-test-python-asts-for-equality-not-reference-or-object-identity).

I guess the hardest part of this approach is specifying what changes are allowed. Could just hard code the test for `*_glacier` forks for now I guess.

SamWilsn (commented):
I wonder if these checks would be better suited to `ethereum-spec-lint`? I believe the `Lint` base class already gives you the previous hard fork, and the `walk_sources` function will be useful.

Could filter for forks ending with `_glacier` and make sure the only changed node is the fork block. Wouldn't work for the DAO fork, but maybe that could be its own special lint.

tests/helpers/diff_test_helpers.py
gurukamath: Sure. I will look into it.

gurukamath:
Closing this in favour of the linting approach #639

PR #637, opened 2022-11-01
Hide submodules and subpackages in diffs

### LINKED ISSUE #636: Links to `Subpackages`  in the diffs of rendered specs don't work
Diffs should technically have no `Subpackages` links displayed in them since not all of the submodules are likely to have diffs.

Rendered Specs

gurukamath:
(closes #636 )

### What was wrong?
Currently, the rendered docs have links to sub-packages and sub-modules in the diffs which are linked to the pickle files instead of html. It is not sensible to have these links in the first place because not all sub-packages might have diffs.

Related to Issue #636 

### How was it fixed?
Hide the relevant sections upon rendering the page.

[This](https://gurukamath.github.io/execution-specs/diffs/frontier_homestead/index.html) is what the rendered docs will look like after this PR.

#### Cute Animal Picture

### DIFF
--- a/doc/_templates/autoapi/python/module.rst
+++ b/doc/_templates/autoapi/python/module.rst
@@ -12,38 +12,40 @@
 {{ obj.docstring|prepare_docstring }}
 {% endif %}
 
-{% block subpackages %}
-{% set visible_subpackages = obj.subpackages|selectattr("display")|list %}
-{% if visible_subpackages %}
-Subpackages
------------
-.. toctree::
-   :titlesonly:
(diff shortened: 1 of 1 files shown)


SamWilsn:
Can we do this with the `stage0` and `stage` tags in the templates? If so, would that be a better approach?

gurukamath:
This is indeed handled much more elegantly using the `stage1` tag inside the templates. Thanks for the suggestion @SamWilsn. I have updated the PR.

SamWilsn (approved):

PR #639, opened 2022-11-05
add `_glacier` forks lints

gurukamath:
### What was wrong?
There are no controls on the changes in the minor forks

Related to Issue #605 

### How was it fixed?
Add lints for `_glacier` forks

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/lint/__init__.py
+++ b/src/ethereum_spec_tools/lint/__init__.py
@@ -24,7 +24,13 @@ def compare_ast(old: ast.AST, new: ast.AST) -> bool:
     if type(old) is not type(new):
         return False
 
-    if isinstance(old, ast.Expr):
+    # This is a way to identify comments in the current specs code
+    # Returns true irrespective of what the comment is
+    if (
+        isinstance(old, ast.Expr)
+        and isinstance(old.value, ast.Constant)
+        and isinstance(old.value.value, str)
+    ):
         return True
 
     if isinstance(old, ast.AST):
@@ -91,15 +97,6 @@ class Lint(metaclass=ABCMeta):
         position :
             The particular hardfork to lint.
         """
-
-    def add_diagnostic(
-        self, diagnostics: List[Diagnostic], message: str
-    ) -> None:
-        """
-        Adds a new diagnostic message.
-        """
-        diagnostic = Diagnostic(message=message)
-        diagnostics.append(diagnostic)
 
     def _parse(
         self, source: str, visitor: ast.NodeVisitor, attr: str
--- a/src/ethereum_spec_tools/lint/__init__.py
+++ b/src/ethereum_spec_tools/lint/__init__.py
@@ -11,9 +11,43 @@ import inspect
 import pkgutil
 from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass
+from itertools import zip_longest
 from typing import Generator, List, Optional, Sequence, Tuple
 
 from ..forks import Hardfork
+
+
+def compare_ast(old: ast.AST, new: ast.AST) -> bool:
+    """
+    Check if two nodes are the equal.
+    """
+    if type(old) is not type(new):
+        return False
+
+    if isinstance(old, ast.Expr):
+        return True
+
+    if isinstance(old, ast.AST):
+        for k, v in vars(old).items():
+            if k in {
+                "lineno",
+                "end_lineno",
+                "col_offset",
+                "end_col_offset",
+                "ctx",
+            }:
+                continue
+            if not compare_ast(v, getattr(new, k)):
+                return False
+        return True
+
+    elif isinstance(old, list) and isinstance(new, list):
+        return all(
+            compare_ast(old_item, new_item)
+            for old_item, new_item in zip_longest(old, new)
+        )
+    else:
+        return old == new
 
 
 def walk_sources(fork: Hardfork) -> Generator[Tuple[str, str], None, None]:
@@ -57,6 +91,15 @@ class Lint(metaclass=ABCMeta):
         position :
             The particular hardfork to lint.
         """
+
+    def add_diagnostic(
+        self, diagnostics: List[Diagnostic], message: str
+    ) -> None:
+        """
+        Adds a new diagnostic message.
+        """
+        diagnostic = Diagnostic(message=message)
+        diagnostics.append(diagnostic)
 
     def _parse(
         self, source: str, visitor: ast.NodeVisitor, attr: str
--- a/src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
+++ b/src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
@@ -16,6 +16,14 @@ from ethereum_spec_tools.lint import (
 )
 
 
+def add_diagnostic(diagnostics: List[Diagnostic], message: str) -> None:
+    """
+    Adds a new diagnostic message.
+    """
+    diagnostic = Diagnostic(message=message)
+    diagnostics.append(diagnostic)
+
+
 class GlacierForksHygiene(Lint):
     """
     Ensures that the glacier forks have changes only in BOMB_DELAY_BLOCKS.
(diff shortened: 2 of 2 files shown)


src/ethereum_spec_tools/lint/__init__.py
```diff
@@ -11,11 +11,45 @@
...
 
+def compare_ast(old: ast.AST, new: ast.AST) -> bool:
+    """
+    Check if two nodes are the equal.
+    """
+    if type(old) is not type(new):
+        return False
+
+    if isinstance(old, ast.Expr):
+        return True
```
SamWilsn, line 28: As discussed on call, we should probably limit this check to expressions that contain only a single string and document that we're ignoring docstrings.

src/ethereum_spec_tools/lint/__init__.py
```diff
@@ -11,11 +11,45 @@
 import pkgutil
 from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass
+from itertools import zip_longest
 from typing import Generator, List, Optional, Sequence, Tuple
 
 from ..forks import Hardfork
 
 
+def compare_ast(old: ast.AST, new: ast.AST) -> bool:
```
SamWilsn, line 20: If we want to put `compare_ast` in the superclass (and not in a specific lint), we want it to be generic. Maybe something like this, where `ignore` is called whenever two nodes differ and returns `True` if that difference should be ignored:

```python
def compare_ast(old: ast.AST, new: ast.AST, ignore: Callable[[ast.AST, ast.AST], bool]):
    pass
```

That said, I think I'd prefer the quick and dirty option of putting `compare_ast` in `glacier_forks_hygiene.py` file for now. It's simpler and we don't need to worry about making it generic until we actually need another lint that compares syntax trees.

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+    Ensures that the glacier forks have changes only in BOMB_DELAY_BLOCKS.
+    """
+
+    def __init__(self) -> None:
+        self.delay_blocks = {
+            "muir_glacier": 9000000,
+            "arrow_glacier": 10700000,
+            "gray_glacier": 11400000,
+        }
```
SamWilsn, line 29: Is it useful to compare a constant in file A with a constant in file B? If someone is going to mess up the fork number, they'll just add the same incorrect number here and under `src/` :rofl:

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Visit a class.
+        """
+        self._insert(klass.name, klass)
+
+    def visit_Assign(self, assign: ast.Assign) -> None:
+        """
+        Visit an assignment.
+        """
+        self._insert(assign.targets[0].id, assign)  # type: ignore
```
SamWilsn, line 177: Why do we ignore types here, and is there a way to implement this without ignoring?

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
```diff
@@ -0,0 +1,183 @@
...
+                )
+                continue
+
+            current_node = self._parse(all_current[file], _Visitor(), "items")
+            previous_node = self._parse(
+                all_previous[file], _Visitor(), "items"
+            )
+
+            diagnostics += self.compare(
+                fork_name, file, current_node, previous_node  # type: ignore
```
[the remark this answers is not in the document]
gurukamath: I guess this will be fixed once `_parse` is updated

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
```diff
@@ -0,0 +1,183 @@
...
+                self.add_diagnostic(
+                    diagnostics,
+                    (f"the file `{file}` is added to `{fork_name}`."),
+                )
+                continue
+
+            if file not in all_current:
+                self.add_diagnostic(
+                    diagnostics,
+                    (f"the file `{file}` is deleted from `{fork_name}`."),
```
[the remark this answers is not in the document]
gurukamath: done

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
```diff
@@ -0,0 +1,183 @@
...
+
+        all_previous = dict(walk_sources(forks[position - 1]))
+        all_current = dict(walk_sources(forks[position]))
+
+        all_files = set(all_previous.keys()) | set(all_current.keys())
+        for file in all_files:
+            if file not in all_previous:
+                self.add_diagnostic(
+                    diagnostics,
+                    (f"the file `{file}` is added to `{fork_name}`."),
```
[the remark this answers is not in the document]
gurukamath: done

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+    def lint(
+        self, forks: List[Hardfork], position: int
+    ) -> Sequence[Diagnostic]:
+        """
+        Walks the sources for each hardfork and emits Diagnostic messages.
+        """
+        fork_name = forks[position].short_name
+        if not fork_name.endswith("_glacier"):
```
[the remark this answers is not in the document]
gurukamath: done

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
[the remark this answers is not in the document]
gurukamath: Seems to work just as well without. Not sure how the parentheses were added. Updated

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
```diff
@@ -0,0 +1,183 @@
...
+
+        diagnostics: List[Diagnostic] = []
+
+        all_previous = dict(walk_sources(forks[position - 1]))
+        all_current = dict(walk_sources(forks[position]))
+
+        all_files = set(all_previous.keys()) | set(all_current.keys())
+        for file in all_files:
+            if file not in all_previous:
+                self.add_diagnostic(
```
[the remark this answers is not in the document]
gurukamath: done

src/ethereum_spec_tools/lint/__init__.py
```diff
@@ -58,6 +92,15 @@ def lint(
...
         """
 
+    def add_diagnostic(
+        self, diagnostics: List[Diagnostic], message: str
+    ) -> None:
+        """
+        Adds a new diagnostic message.
+        """
+        diagnostic = Diagnostic(message=message)
+        diagnostics.append(diagnostic)
```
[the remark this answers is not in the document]
gurukamath: done

src/ethereum_spec_tools/lint/__init__.py
gurukamath: done

src/ethereum_spec_tools/lint/__init__.py
gurukamath: ok.

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
gurukamath: The only other thing that I could think of was to just ignore any comparisons if the node is `BOMB_DELAY_BLOCKS` but then concluded that some comparison is better than nothing :D

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
gurukamath: Done. Explicitly handle when target is name. Notify when it is something else (should ideally not happen in the specs code).

SamWilsn (approved):

PR #640, opened 2022-11-08
Refactor account deletion and clearing

petertdavies:
This removes the preservation of stack frames of the course of execution through the `children` element of `EVM` and instead switches to a model where `child_evm` s are incorporated into their parents. This substantially reduces the memory usage of the test suite and gives the specs semantics that better match real clients.

This only implements the changes for `london`, backporting to other hardforks will be required before merging.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/london/vm/__init__.py
+++ b/src/ethereum/london/vm/__init__.py
@@ -14,15 +14,18 @@ The abstract computer which runs the code stored in an
 """
 
 from dataclasses import dataclass
-from typing import Dict, List, Optional, Set, Tuple, Union
+from typing import List, Optional, Set, Tuple, Union
 
 from ethereum.base_types import U256, Bytes, Bytes0, Bytes32, Uint, Uint64
 from ethereum.crypto.hash import Hash32
 
 from ..eth_types import Address, Log
-from ..state import State
+from ..state import State, account_exists_and_is_empty
+from ..utils.address import to_address
 
 __all__ = ("Environment", "Evm", "Message")
+
+RIPEMD160_ADDRESS = to_address(Uint(3))
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


src/ethereum/london/vm/__init__.py
```diff
@@ -82,10 +85,64 @@ class Evm:
...
+        The parent `EVM`.
+    child_evm :
+        The child evm to incorporate.
+    """
+    evm.gas_left += child_evm.gas_left
+    evm.logs += child_evm.logs
+    evm.refund_counter += child_evm.refund_counter
+    evm.accounts_to_delete.update(child_evm.accounts_to_delete)
+    evm.touched_accounts.update(child_evm.touched_accounts)
+    if account_exists_and_is_empty(
```
petertdavies, line 113: I have wrapped every touch with this `account_exists_and_is_empty()` check. This is not necessary, because the check has to be repeated at the end of the transaction anyway. I've included it to demonstrate that only touches of empty accounts need to be tracked, but this is arguably an optimization that shouldn't be included in the specs.

src/ethereum/london/vm/__init__.py
```diff
@@ -82,10 +85,64 @@ class Evm:
...
+    touched_accounts: Set[Address]
     has_erred: bool
-    children: List["Evm"]
     return_data: Bytes
     error: Optional[Exception]
     accessed_addresses: Set[Address]
     accessed_storage_keys: Set[Tuple[Address, Bytes32]]
+
+
+def incorporate_child_evm_successful(evm: Evm, child_evm: Evm) -> None:
```
SamWilsn, line 97: Minor nit, but I think either `incorporate_successful_child` or `incorporate_child_on_success` would read better.

src/ethereum/london/vm/__init__.py
```diff
@@ -82,10 +85,64 @@ class Evm:
...
+    evm.touched_accounts.update(child_evm.touched_accounts)
+    if account_exists_and_is_empty(
+        evm.env.state, child_evm.message.current_target
+    ):
+        evm.touched_accounts.add(child_evm.message.current_target)
+    evm.accessed_addresses.update(child_evm.accessed_addresses)
+    evm.accessed_storage_keys.update(child_evm.accessed_storage_keys)
+
+
+def incorporate_child_evm_error(evm: Evm, child_evm: Evm) -> None:
```
SamWilsn, line 121: Minor nit, but I think either `incorporate_reverted_child` or `incorporate_child_on_error` would read better.

src/ethereum/london/vm/__init__.py
```diff
@@ -82,10 +85,64 @@ class Evm:
...
+    # reverted in order to preserve this historical behaviour.
+    if RIPEMD160_ADDRESS in child_evm.touched_accounts:
+        evm.touched_accounts.add(RIPEMD160_ADDRESS)
+    if (
+        child_evm.message.current_target == RIPEMD160_ADDRESS
+        and account_exists_and_is_empty(
+            evm.env.state, child_evm.message.current_target
+        )
+    ):
+        evm.touched_accounts.add(RIPEMD160_ADDRESS)
```
SamWilsn, line 147: To avoid the parentheses hell:

```suggestion
    if child_evm.message.current_target == RIPEMD160_ADDRESS:
        if account_exists_and_is_empty(evm.env.state, child_evm.message.current_target):
            evm.touched_accounts.add(RIPEMD160_ADDRESS)
```

src/ethereum/london/vm/instructions/system.py
```diff
@@ -110,17 +116,15 @@ def generic_create(
         accessed_storage_keys=evm.accessed_storage_keys.copy(),
     )
     child_evm = process_create_message(child_message, evm.env)
-    evm.children.append(child_evm)
+
     if child_evm.has_erred:
-        push(evm.stack, U256(0))
+        incorporate_child_evm_error(evm, child_evm)
         evm.return_data = child_evm.output
```
SamWilsn, line 122: Should the `incorporate_child_*` functions handle the `return_data`?

src/ethereum/london/vm/interpreter.py
```diff
@@ -107,16 +105,19 @@ def process_message_call(
...
+        refund_counter=U256(evm.refund_counter)
+        if not evm.has_erred
+        else U256(0),
         logs=evm.logs if not evm.has_erred else (),
-        accounts_to_delete=accounts_to_delete,
-        touched_accounts=collect_touched_accounts(evm),
+        accounts_to_delete=evm.accounts_to_delete
+        if not evm.has_erred
+        else set(),
+        touched_accounts=evm.touched_accounts if not evm.has_erred else set(),
```
SamWilsn, line 120: These conditionals are maybe too Pythonic? Plus they're definitely too hard to read :rofl:

src/ethereum/london/vm/instructions/system.py
petertdavies: The behaviour for `return_data` for `CREATE*` vs `CALL*` opcodes, so has to be handled separately.

SamWilsn (approved):

PR #654, opened 2022-11-21
Refactor apply_body()

petertdavies:
This PR refactors the `apply_body()` method, by moving the most of the fork specific rules into other functions. Many of the spec tools `t8n`, block building, basically boil down to re-implementing `apply_body()` and it would be easier to do this if `apply_body()` is as a generic as reasonably possible.

Paris only. Will require backporting prior to merge.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/paris/spec.py
+++ b/src/ethereum/paris/spec.py
@@ -271,6 +271,97 @@ def validate_header(header: Header, parent_header: Header) -> None:
     ensure(header.parent_hash == block_parent_hash, InvalidBlock)
 
 
+def check_transaction(
+    tx: Transaction,
+    base_fee_per_gas: Uint,
+    gas_available: Uint,
+    chain_id: Uint64,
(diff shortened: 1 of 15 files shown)


src/ethereum/paris/spec.py
```diff
@@ -272,6 +272,97 @@ def validate_header(header: Header, parent_header: Header) -> None:
...
+    -------
+    sender_address :
+        The sender of the transaction.
+    effective_gas_price :
+        The price to charge for gas when the transaction is executed.
+
+    Raises
+    ------
+    InvalidBlock :
+        If the transaction is not includable.
```
SamWilsn, line 305: I think we've dropped this section of the docstrings.

SamWilsn (approved):

SamWilsn:
Is it safe to assume the errors from the optimized build aren't related to these changes? If they are not related, merge away!

PR #656, opened 2022-11-24
Add EIPs to CFI, based on ACD150

timbeiko:
Decided on https://github.com/ethereum/pm/issues/662

### DIFF
--- a/network-upgrades/mainnet-upgrades/shanghai.md
+++ b/network-upgrades/mainnet-upgrades/shanghai.md
@@ -12,8 +12,13 @@ Specifies changes included in the Network Upgrade.
 Specifies changes potentially included in the Network Upgrade, pending successful deployment on Client Integration Testnets.
 
 * [EIP-1153: Transient storage opcodes](https://eips.ethereum.org/EIPS/eip-1153)
+* [EIP-2537: Precompile for BLS12-381 curve operations](https://eips.ethereum.org/EIPS/eip-2537)
 * [EIP-3540: EVM Object Format (EOF) v1](https://eips.ethereum.org/EIPS/eip-3540)
 * [EIP-3670: EOF - Code Validation](https://eips.ethereum.org/EIPS/eip-3670)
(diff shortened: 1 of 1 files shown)


network-upgrades/mainnet-upgrades/shanghai.md
```diff
@@ -1,19 +1,24 @@
...
-Specifies changes included in the Network Upgrade.
+Changes included in the Network Upgrade.
 
 * [EIP-3651: Warm COINBASE](https://eips.ethereum.org/EIPS/eip-3651)
 * [EIP-3855: PUSH0 instruction](https://eips.ethereum.org/EIPS/eip-3855)
 * [EIP-3860: Limit and meter initcode](https://eips.ethereum.org/EIPS/eip-3860)
 * [EIP-4895: Beacon chain push withdrawals as operations](https://eips.ethereum.org/EIPS/eip-4895)
 
 ### EIPs Considered for Inclusion
-Specifies changes potentially included in the Network Upgrade, pending successful deployment on Client Integration Testnets.
```
moodysalem, line 12: i think the change in definition warrants a reclassification of those EIPs already in this category, for example i think transient storage was CFI for shanghai meaning that assuming it did not delay shanghai or present issues during testing it would be included, but this change makes that far less clear

moodysalem (changes requested):

network-upgrades/mainnet-upgrades/shanghai.md
timbeiko: @moodysalem in the interest of providing an accurate view of where things are at sooner rather than later, I think I'd merge it as is. On the next ACD, I expect we'll have a clear list of what's included in Shanghai and a rough idea of what to do about other EIPs. It makes sense to revamp CFI for the next fork (as we do a bunch of other process changes as well). For 1153 specifically, I don't think it should change what we do: next step is to get it on devnets along with everything else!

PR #659, opened 2022-11-30
Further update Shanghai PR table, add new columns

haltman-at:
I noticed the PR table hadn't been updated with new columns, so I went ahead and did that (including any PRs I could find for these columns).  I also added new PRs in the existing columns that weren't listed.  Note that Besu appeared to have two PRs open for 3860?  So I listed them both.  Again, I have no affiliation with any of these projects and have no clue which of these PRs might be preferred.

This table's getting pretty big, maybe it would be a good idea to reorder the columns so as to put the included ones before the ones that are merely CFI?  Or split it into two tables, or something?  Well, feel free to make edits to this before merging.


timbeiko:
@haltman-at thanks! will merge as is, but agreed it probably makes sense to split between CFI and Included tables. Might wait until after the next ACD to do so, though, given the set of Included/CFI'd EIPs will probably change.

PR #660, opened 2022-12-02
Update table for recent merges

haltman-at:
Here's one more of these since I noticed two more merges!  Some notes:

1. With Besu and 3860, it looks like the merged PR really does implement it; the second PR is redundant as to implementing 3860, but has been left open because it apparently contains some useful tests and they want to merge those.  But I think we can just call this merged.  (Edit: The second one has now been closed!  So yeah we can just focus on the first one.)

2. With Erigon and 4895, the linked PR as now been merged, but since it was only part 1 of implementing 4895, I just changed the label to "partly merged".  Hope that's OK!  (I mean arguably the link should be removed entirely, but, well, if you think that's what should happen... feel free to edit this.)


haltman-at:
Oh, Erigon has added their second 4895 PR, so I've updated the table to show both.

haltman-at:
Oh also it looks like the Geth 3670 PR that is linked has been supplanted by a different one, so I updated that link too.

haltman-at:
...oh, and I just found a few more PRs that can be added to the table!  Think that should be it for now though. :)

timbeiko:
Thank you!

PR #663, opened 2022-12-08
Update Shanghai and add Cancun

timbeiko:
This reflects the decisions made on [ACD151](https://github.com/ethereum/pm/issues/675), namely: 

- Including the EOF EIPs into Shanghai (may be removed if not ready by January)
- Including EIP-4844 in a newly specified Cancun upgrade, set to follow Shanghai (naming follows [devcon locations chronologically](https://archive.devcon.org/archive/))
- Moving previously CFI'd EIPs to the Cancun spec to discuss them further in future calls. 
   - Note: CFI'd should probably be moved outside of specific upgrade specs, but wanted to minimize changes in this specific PR, so just stuck to existing pattern.


djrtwo (approved):

holgerd77:
LGTM

holgerd77:
Do we want an entry in the table from the root README as well? 🤔

holgerd77:
(so: for Cancun)

timbeiko:
Good catch, @holgerd77 ! Quickly added it in a follow up commit: https://github.com/ethereum/execution-specs/commit/ccf47a3a61035ce68bdd77733b12e5704a25f40d

PR #666, opened 2022-12-12
Add two PRs to Shanghai progress table

haltman-at:
Spotted two more PRs that can be added to the table, here's a PR for them.


haltman-at:
Whoops, two cells got swapped, fixed that.

PR #676, opened 2023-01-05
Remove EOF from Shanghai Spec

timbeiko:
Decided on ACD#152


yperbasis:
Shall we move them to Cancun CFI instead?

timbeiko:
@shemnon asked the same thing in the R&D discord. Happy to make that change if Geth + Nethermind also +1 :-)

holgerd77:
Would also support moving to Cancun CFI, I guess this will help that the topic stays on the table in a sufficient manner and also helps to serve as a reminder to continuously evolve thoughts on the topic and have necessary in-between discussions and decisions so that we hopefully get to a consistent and holistic plan on this incuding some long-term thinking throughout the months without the need for some last-minute rush where things run the risk to fall under the table.

timbeiko:
Merged as-is because other changes are blocked on it. Added an agenda item about Cancun CFI to the next ACD (https://github.com/ethereum/pm/issues/704) given we couldn't get everyone to agree async here.

Issue #643, opened 2022-11-11
Single-source precompile addresses


SamWilsn:
RIPEMD160's address is duplicated in two places in each fork. For example:

https://github.com/ethereum/execution-specs/blob/c004ccdc04a28259d3ce3de1c19620ba8a215238/src/ethereum/constantinople/vm/interpreter.py#L51

https://github.com/ethereum/execution-specs/blob/761b08f5ced813fa8cf6799d7a53170a79d7d526/src/ethereum/constantinople/vm/precompiled_contracts/mapping.py#L28

The address of each precompile should be defined in a single location.

PR #680, opened 2023-01-10
Backport stack frame changes

petertdavies:
### What was wrong?

The change to the stack frames had only been applied to the London hardfork.

### How was it fixed?

Port it to all the other forks.

#### Cute Animal Picture


src/ethereum/berlin/vm/__init__.py
```diff
@@ -14,16 +14,19 @@
...
 from ethereum.crypto.hash import Hash32
 
 from ..eth_types import Address, Log
-from ..state import State
+from ..state import State, account_exists_and_is_empty
+from ..utils.address import to_address
 
 __all__ = ("Environment", "Evm", "Message")
 
+RIPEMD160_ADDRESS = to_address(Uint(3))
```
SamWilsn, line 28: I think you can do this now:

```python
from .precompiled_contracts.mapping import RIPEMD160_ADDRESS
```

src/ethereum/berlin/vm/__init__.py
```diff
@@ -81,10 +84,62 @@ class Evm:
...
+    Incorporate the state of an unsuccessful `child_evm` into the parent `evm`.
+
+    Parameters
+    ----------
+    evm :
+        The parent `EVM`.
+    child_evm :
+        The child evm to incorporate.
+    """
+    # In block 2675119, the empty account at 0x2 (the RIPEMD160 precompile) was
```
SamWilsn, line 131: Isn't the RIPEMD160 precompile at 0x3?

src/ethereum/berlin/vm/__init__.py
petertdavies: Unfortunately this fails due to a circular import.

SamWilsn: Hm, that's a problem then. I'd rather not regress on #643. Could we move the `*_ADDRESS` constants to `__init__.py` maybe?

gurukamath (approved):

PR #673, opened 2023-01-03
Add final Erigon PR for EIP 4895 to table

haltman-at:
Erigon has put up (what I believe is) the last of their 3 PRs for implementing EIP 4895, so I've added it to the table!

PR #681, opened 2023-01-10
More Shanghai table updates

haltman-at:
Redo of #673 now that EOF is removed from Shanghai.

Updates:
1. Erigon put up the third of their three 4895 PRs, so that PR is added to the table and that cell is marked merged.
2. Besu closed their 4895 PR and put up a new one, so that cell is changed over.
3. It turns out that even though Erigon was marked merged for 3860, they actually need a second PR for it, so that PR is added to the table and I've marked the cell as just partly merged.


timbeiko:
Thank you 🙏 !

PR #682, opened 2023-01-11
Mark three PRs as merged, update another PR to its replacement

haltman-at:
Two Shanghai PRs have been merged, marking them as such!

Edit: Also, Geth closed their 4895 PR and replaced it with a new one, so I've also updated that cell.

Edit: Also, Erigon's second (and final?) PR for 3860 has been merged, so I've updated that cell to merged as well.


timbeiko (approved):
thank you!

PR #688, opened 2023-01-19
Add EOF EIPs to Cancun CFI List

timbeiko:
Decided on ACDE153

### DIFF
--- a/network-upgrades/mainnet-upgrades/cancun.md
+++ b/network-upgrades/mainnet-upgrades/cancun.md
@@ -10,6 +10,12 @@ Changes [Considered for Inclusion](https://github.com/ethereum/execution-specs/t
 
 * [EIP-1153: Transient storage opcodes](https://eips.ethereum.org/EIPS/eip-1153)
 * [EIP-2537: Precompile for BLS12-381 curve operations](https://eips.ethereum.org/EIPS/eip-2537)
+* [EIP-3540: EVM Object Format (EOF) v1](https://eips.ethereum.org/EIPS/eip-3540)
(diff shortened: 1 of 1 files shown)


axic (approved):

holgerd77:
Would it be worth to add the client readiness table from https://github.com/ethereum/execution-specs/commit/ad2bd9333a7958a8443c4591e4cd15c553fc6323 so that it won't get forgotten (this was some substantial work to put this together I guess)? 🤔 I would assume this will stay relatively static over time.

timbeiko:
@holgerd77 I'm personally not a huge fan of the tables for CFI lists (vs. Included EIPs), so I think we should hold off for now.

holgerd77:
> @holgerd77 I'm not a huge fan of the CFI table & it's something we sort of started to do only for Shanghai. IMO they make more sense for Included EIPs. Given we're still pretty early in the Cancun planning, an we may see non-trivial EOF spec changes, I'd hold off for now. At the very least, I think it should be a separate PR.

Sure, makes sense. 👍

jakerockland:
@timbeiko which EOF EIPs is this? Is it all of: [EIP-3540](https://eips.ethereum.org/EIPS/eip-3540), [EIP-3670](https://eips.ethereum.org/EIPS/eip-3670), [EIP-4200](https://eips.ethereum.org/EIPS/eip-4200), [EIP-4750](https://eips.ethereum.org/EIPS/eip-4750), and [EIP-5450](https://eips.ethereum.org/EIPS/eip-5450)?

PR #691, opened 2023-01-21
decode blocks from `rlp`

### LINKED ISSUE #690: Retrieve blocks in tests from `rlp` instead of the individual fields
In tests, the individual block fields for a test are meant to be primarily for human readability. The testing framework should look to retrieve the blocks from `rlp` wherever applicable. Although the json fields are almost always going to be consistent with the `rlp`, some cases have been identified where this may not be the case. See [test](https://github.com/ethereum/tests/blob/7c1ffe6e35962e52124ef6dab627f10308155a29/BlockchainTests/GeneralStateTests/EIPTests/stEIP3860/create2InitCodeSizeLimit.json#L167) where the `withdrawals` are missing from the block.

Update tests to use the `rlp` wherever available and use the json fields only where there is no `rlp`

gurukamath:
(closes #690 )

### What was wrong?
Blocks for tests are read from json instead of rlp.

Related to Issue #690 

### How was it fixed?
decode from rlp whenever it is available.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/rlp.py
+++ b/src/ethereum/rlp.py
@@ -282,7 +282,7 @@ def _decode_to(cls: Type[T], raw_rlp: RLP) -> T:
         ensure(type(raw_rlp) == list, RLPDecodingError)
         assert isinstance(raw_rlp, list)
         args = []
-        # FIXME: Add length check
+        ensure(len(fields(cls)) == len(raw_rlp), RLPDecodingError)
         for (field, rlp_item) in zip(fields(cls), raw_rlp):
             args.append(_decode_to(field.type, rlp_item))
         return cls(*args)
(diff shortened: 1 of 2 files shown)


petertdavies (approved):
LGTM

src/ethereum/rlp.py
```diff
@@ -282,7 +282,7 @@ def _decode_to(cls: Type[T], raw_rlp: RLP) -> T:
         ensure(type(raw_rlp) == list, RLPDecodingError)
         assert isinstance(raw_rlp, list)
         args = []
-        # FIXME: Add length check
+        ensure(len(fields(cls)) == len(raw_rlp), RLPDecodingError)
```
SamWilsn, line 285: :tada:

PR #692, opened 2023-01-21
Implement Shanghai

gurukamath:
### What was wrong?
Shanghai Hardfork not implemented

### How was it fixed?
Implemented EIP-4895, EIP-3651, EIP-3860 and EIP-3855

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/shanghai/spec.py
+++ b/src/ethereum/shanghai/spec.py
@@ -490,6 +490,7 @@ def process_transaction(
     preaccessed_addresses = set()
     preaccessed_storage_keys = set()
     if isinstance(tx, (AccessListTransaction, FeeMarketTransaction)):
+        preaccessed_addresses.add(env.coinbase)
         for (address, keys) in tx.access_list:
             preaccessed_addresses.add(address)
             for key in keys:
--- a/src/ethereum/shanghai/spec.py
+++ b/src/ethereum/shanghai/spec.py
@@ -60,7 +60,8 @@ from .state import (
 )
 from .trie import Trie, root, trie_set
 from .utils.message import prepare_message
-from .vm.interpreter import process_message_call
+from .vm.gas import init_code_cost
+from .vm.interpreter import MAX_CODE_SIZE, process_message_call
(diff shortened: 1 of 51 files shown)


gurukamath:
Currently in draft because the tests for some EIPs are in `Merge+EIP` phase rather than `Shanghai`

src/ethereum/shanghai/spec.py
```diff
@@ -573,7 +574,15 @@ def validate_transaction(tx: Transaction) -> bool:
     verified : `bool`
         True if the transaction can be executed, or False otherwise.
     """
-    return calculate_intrinsic_cost(tx) <= tx.gas and tx.nonce < 2**64 - 1
+    valid_transaction = True
```
petertdavies, line 577:
```python
if ...:
    return False
elif ...:
    return False

return True
```
would be clearer and more idiomatic.

petertdavies (changes requested):
I found a couple of nits.

src/ethereum/shanghai/spec.py
gurukamath: Updated

src/ethereum/shanghai/state.py
```diff
@@ -453,6 +453,20 @@ def increase_recipient_balance(recipient: Account) -> None:
...
+    wd: Withdrawal,
+) -> None:
+    """
+    Increase the balance of the withdrawing account.
+    """
+
+    def increase_recipient_balance(recipient: Account) -> None:
+        recipient.balance += wd.amount * 10**9
+
+    modify_state(state, wd.address, increase_recipient_balance)
```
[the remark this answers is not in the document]
gurukamath: The `create_ether` function was previously used for issuing mining rewards and is no longer relevant post merge. Having something more closely related to the context of withdrawals makes more sense to me in this case. I am removing the `create_ether` function from Shanghai.

Will also create a new issue to remove the function from Paris

petertdavies (approved):

PR #695, opened 2023-01-24
single source precompile address

gurukamath:
### What was wrong?

The pre-compile addresses for some of the pre-compiles were moved to `precompiles/__init__.py`  as a part of #680 but some were missed

### How was it fixed?

Moved addresses to the relevant module
#### Cute Animal Picture


petertdavies (approved):
LGTM

PR #697, opened 2023-01-25
update test collection

gurukamath:
This commit updates the test collection logic to collect the tests for a particular fork by looking at the  attribute instead of the test key. This collects some additional tests that werent being collected before. Some of these tests, specifically the  currently fail but have been added to the ignore list. This will need to be explored in depth in the future

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
     path = tests/fixtures
     url = https://github.com/ethereum/tests
     branch = develop
+[submodule "tests/execution-spec-generated-tests"]
+	path = tests/execution-spec-generated-tests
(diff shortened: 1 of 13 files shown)


petertdavies (approved):
LGTM

.gitmodules
```diff
@@ -2,3 +2,6 @@
     path = tests/fixtures
     url = https://github.com/ethereum/tests
     branch = develop
+[submodule "tests/execution-spec-generated-tests"]
+	path = tests/execution-spec-generated-tests
+	url = https://github.com/petertdavies/execution-spec-generated-tests.git
```
SamWilsn, line 7: Should we move this directly into this repository?

petertdavies: See #699

PR #702, opened 2023-01-25
Mark Geth withdrawals as merged

haltman-at:
Geth merged withdrawals, marking it as such.

No clue what's up with Besu, I'm leaving that alone.


petertdavies (approved):
LGTM

PR #705, opened 2023-01-26
Explicitly list precompile addresses in `__all__`; fix sorting of forks

SamWilsn:
### What was wrong?

Documentation rendering was broken.

### How was it fixed?

Limit what is re-exported in `__init__.py`.

#### Cute Animal Picture


gurukamath (approved):
LGTM

PR #706, opened 2023-01-31
Gunzip pickle files

SamWilsn:
### What was wrong?

Running out of disk space

### How was it fixed?

gzipping and ungzipping pickle files.

#### Cute Animal Picture


gurukamath (approved):
LGTM! The only point that needs to be re-considered in my opinion is the rolling back of the test fixtures to a previous commit. This will remove test fixtures relevant to Shanghai

PR #715, opened 2023-02-08
Decode blocks from RLP

gurukamath:
### What was wrong?
Currently the testing framework creates the Block and Genesis Block datatypes  from the various fields available in the fixture json files. There are a couple of problems with this approach

1. The ethereum tests repository looks upon a lot of these json fields as an aid for human readability and looks at the rlp fields as the primary source of the data.
2. If new fields are added to the Block, Header, Transaction and Withdrawal datatypes,  patches will have to be added to fix the testing logic

### How was it fixed?
The specs need to derive the data structures for testing by decoding the rlp instead of using the json fields. This will also serve as an additional layer of testing for the rlp decode logic within the specs. The current commit makes that change and also re-factors some of the code in the state test helper.

#### Cute Animal Picture

### DIFF
--- a/tests/helpers/load_state_tests.py
+++ b/tests/helpers/load_state_tests.py
@@ -12,14 +12,12 @@ import pytest
 from _pytest.mark.structures import ParameterSet
 
 from ethereum import rlp
-from ethereum.base_types import U256, Bytes0, Uint64
-from ethereum.crypto.hash import Hash32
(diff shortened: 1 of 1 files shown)


gurukamath:
Converting this to a draft until the upstream [issue](https://github.com/ethereum/execution-spec-tests/issues/64) with the Shanghai Withdrawal tests is fixed

PR #718, opened 2023-02-16
Improve bad RLP handling

petertdavies:
### What was wrong?

The spec was incorrectly handling block with bad RLP in them.

### How was it fixed?

* The RLP decoder now raises `RLPDecodingError` rather than attempting to create an invalid value for a type (e.g. out of bounds integers).
* `RLPDecodingError` is a subclass of `InvalidBlock`.

This should fix the current CI failure with `--optimized`. The tests no longer need to disguish between RLP issues and other types of invalid blocks, but that is not in this PR.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/rlp.py
+++ b/src/ethereum/rlp.py
@@ -51,7 +51,7 @@ def encode(raw_data: RLP) -> Bytes:
     if isinstance(raw_data, (bytearray, bytes)):
         return encode_bytes(raw_data)
     elif isinstance(raw_data, (Uint, FixedUInt)):
-        return encode(raw_data.to_be_bytes())  # type: ignore
+        return encode(raw_data.to_be_bytes())
     elif isinstance(raw_data, str):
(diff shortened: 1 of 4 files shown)


src/ethereum/rlp.py
```diff
@@ -50,8 +50,8 @@ def encode(raw_data: RLP) -> Bytes:
     """
     if isinstance(raw_data, (bytearray, bytes)):
         return encode_bytes(raw_data)
-    elif isinstance(raw_data, (Uint, U256, U64)):
-        return encode(raw_data.to_be_bytes())
+    elif isinstance(raw_data, (Uint, FixedUInt)):
+        return encode(raw_data.to_be_bytes())  # type: ignore
```
SamWilsn, line 54: Would making `FixedUInt` abstract with `to_be_bytes` fix this ignore?

SamWilsn (approved):
We have _way_ too many ignore comments in this module :rofl:

Issue #625, opened 2022-10-10
Implement `t8n` and `b11r` command line tools


petertdavies:
This would allow us to add support for EELS to to [goevmlab](https://github.com/holiman/goevmlab) and access the fuzzers there.

SamWilsn:
Would this look like a CLI tool that takes a JSON filled test file as input?

gurukamath:
SOme inconsistency in `t8n` needs to be sorted out. See [issue](https://github.com/ethereum/go-ethereum/issues/26861)

PR #737, opened 2023-03-25
Support `Dao Fork` and other `state.fork` options for t8n

gurukamath:
### What was wrong?
The go-ethereum version of t8n supports various options for `--state.fork` which involve transition for one fork to the next at a certain block number. E.g - `FrontierToHomesteadAt5`. The t8n tool also doesn't support the DAO Fork.

Related to Issue #625 

### How was it fixed?
This commit adds support for the transition options. It also adds support for the Dao Fork.

#### Cute Animal Picture


src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -205,7 +203,11 @@ def __init__(self, t8n: Any, stdin: Optional[Dict] = None):
                 elif not value.startswith("0x"):
                     data[address][key] = "0x" + hex(int(value))
 
-        self.state = t8n.json_to_state(data)
+        state = t8n.json_to_state(data)
+        if t8n.fork_module == "dao_fork":
+            t8n.fork.apply_dao(state)
```
SamWilsn, line 208: Shouldn't we run `apply_fork` for every fork, and not just the DAO fork?

src/ethereum_spec_tools/evm_tools/utils.py
```diff
@@ -12,6 +14,39 @@
 
 W = TypeVar("W", Uint, U64, U256)
 
+exception_maps = {
```
SamWilsn, line 17: Probably should have an upper case name if it's a constant:

```suggestion
EXCEPTION_MAPS = {
```

gurukamath: updated

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
gurukamath: The `t8n` inputs do not have enough details to re-create the entire chain which is why we decided to use the `apply_body` function as the point of entry for the tool. That is why I have also separated out the state changes in `dao_fork` into a different function `apply_dao` and used that instead.

gurukamath: I think forks that modify the state directly (like the DAO Fork) should be looked upon as more of an exception than a rule. But we can discuss this.

gurukamath: I have created an issue for this for now (see #740 ). Would like to implement the next few steps in the evm tools and re-visit this at a later point

SamWilsn (approved):

src/ethereum_spec_tools/evm_tools/utils.py
```diff
@@ -12,6 +14,39 @@
 
 W = TypeVar("W", Uint, U64, U256)
 
+EXCEPTION_MAPS = {
+    "FrontierToHomesteadAt5": {
```
winsvega, line 18: this logic can be done also on testers side. if tool is transition only it does not keep blockchain data.
so the testing tool can run it with Frontier or Homestead to build a transition chain

PR #738, opened 2023-03-26
Consolidate test fixtures in a single directory

gurukamath:
### What was wrong?
Currently, the test fixtures are downloaded or cloned into their own individual folders. The paths to these fixtures are also hard coded in the test source files.

### How was it fixed?
This commit consolidates all the fixtures into a single directory. Also, the paths to the fixtures are set as environment variables in `conftest.py` and used in the tesing source files

#### Cute Animal Picture

Cute Animal

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -109,8 +109,10 @@ def pytest_sessionstart(session: Session) -> None:
 
     fixtures_location = "tests/fixtures"
 
-    if not os.path.exists(fixtures_location):
+    try:
         os.mkdir(fixtures_location)
+    except FileExistsError:
+        pass
 
     for idx, props in test_fixtures.items():
         fixture_path = f"{fixtures_location}/{idx}"
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,14 +16,14 @@ from ethereum_spec_tools.evm_trace import evm_trace
 # Update the links and commit has in order to consume
 # newer/other tests
 test_fixtures = {
-    "execution-spec-generated-tests": {
+    "execution_spec_generated_tests": {
         "url": "https://github.com/ethereum/execution-spec-tests/releases/download/v0.2.3/fixtures.tar.gz",
     },
     "t8n_testdata": {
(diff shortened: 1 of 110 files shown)


tests/conftest.py
```diff
@@ -107,11 +107,18 @@ def git_clone_fixtures(url: str, commit_hash: str, location: str) -> None:
 
 def pytest_sessionstart(session: Session) -> None:
 
-    fixtures_location = "tests"
+    fixtures_location = "tests/fixtures"
+
+    if not os.path.exists(fixtures_location):
+        os.mkdir(fixtures_location)
```
SamWilsn, line 113: Better to just try and make the directory than to check if it exists (avoids [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) attacks, not that it's relevant here):

```suggestion
    try:
        os.mkdir(fixtures_location)
    except FileExistsError:
        pass
```

tests/conftest.py
```diff
@@ -107,11 +107,18 @@ def git_clone_fixtures(url: str, commit_hash: str, location: str) -> None:
...
 
     for idx, props in test_fixtures.items():
         fixture_path = f"{fixtures_location}/{idx}"
 
+        # Set the path of the various downloaded fixtures
+        # as environment variables
+        os.environ[idx.upper()] = fixture_path
```
SamWilsn, line 120: Hm, is there no pytest native way to get this information into the tests? Using an environment variable seems a bit strange.

gurukamath: Pytest generally discourages using global variables but the way our tests are set up, we need to use them if we want to avoid hardcoding the paths imo. I think in our scenario there aren't many downsides to using these custom global variables. 

As far as using a pytest native way for this is concerned, [this](https://github.com/pytest-dev/pytest/issues/3118#issuecomment-357702092) is the closest solution I could find but I am not sure it is much better than the `os.environ` variant :)

tests/conftest.py
gurukamath: Thanks for this context. I will update.

PR #739, opened 2023-03-29
Consolidate test fixtures in a single directory

gurukamath:
### What was wrong?
Currently, the test fixtures are downloaded or cloned into their own individual folders. The paths to these fixtures are also hard coded in the test source files.

### How was it fixed?
This commit consolidates all the fixtures into a single directory. Also, the paths to the fixtures are set as environment variables in `conftest.py` and used in the tesing source files

#### Cute Animal Picture

Cute Animal


gurukamath:
Just moved the `TEST_FIXTURES` setting from `conftest.py` to `test.helpers` removing the need to set `env` variables

SamWilsn (approved):

PR #743, opened 2023-04-01
Refactor base fee calculation

gurukamath:
### What was wrong?

1. The base fee calculation should be separated into its own function so that the t8n tool can call that function in order to derive the `base_fee_per_gas` from the parent block parameters.
2. Currently, the fork block in `Arrow Glacier` and `Gray Glacier` forks has `INITIAL_BASE_FEE` as the base fee. This was originally meant to be true only in case of London.

Related to Issue #625 

### How was it fixed?
This commit performs refactors and simplifies the base fee calculation and also corrects the bug in `Arrow Glacier` and `Gray Glacier`.

#### Cute Animal Picture


gurukamath:
Added the feature to derive `base_fee_per_gas` from parent in t8n

SamWilsn (approved):

PR #744, opened 2023-04-06
Make sync tool support testnets and timestamp forking

petertdavies:
### What was wrong?

* Timestamp forking wasn't supported in the sync tool
* Testnets weren't supported by the sync tool
* There was a bug with `chain_id` being hardcoded as 1 in some places.

### How was it fixed?

Fixed all the above.

#### Cute Animal Picture


SamWilsn:
Looks like the fork spec tool needs to be updated?

src/ethereum/shanghai/fork.py
```diff
@@ -559,8 +559,8 @@ def process_transaction(
 
     preaccessed_addresses = set()
     preaccessed_storage_keys = set()
+    preaccessed_addresses.add(env.coinbase)
     if isinstance(tx, (AccessListTransaction, FeeMarketTransaction)):
-        preaccessed_addresses.add(env.coinbase)
```
SamWilsn, line 563: This seems unrelated to the title of this pull request. Is it supposed to be here, and if so, should there be a test that differentiates these behaviours? Maybe a `LegacyTransaction` that tries to do a cold read from the coinbase?

src/ethereum_spec_tools/forks.py
```diff
@@ -40,13 +40,14 @@ def discover(cls: Type[H]) -> List[H]:
         for pkg in modules:
             mod = importlib.import_module(pkg.name)
             block = getattr(mod, "MAINNET_FORK_BLOCK", -1)
+            timestamp = getattr(mod, "MAINNET_FORK_TIMESTAMP", -1)
```
SamWilsn, line 43: This might be a better way to differentiate between not having a fork start and it being `None`:

```suggestion
            try:
                block = getattr(mod, "MAINNET_FORK_BLOCK")
                timestamp = getattr(mod, "MAINNET_FORK_TIMESTAMP")
            except AttributeError:
                continue
```

src/ethereum_spec_tools/forks.py
```diff
@@ -57,12 +58,69 @@ def discover(cls: Type[H]) -> List[H]:
...
+        """
+        config = sorted(config_dict.items(), key=lambda x: x[0])
+
+        forks = []
+
+        for (block_or_time, name) in config:
+            mod = importlib.import_module("ethereum." + name)
+
+            if block_or_time < 1_000_000_000:
+                mod.MAINNET_FORK_BLOCK = block_or_time  # type: ignore
```
SamWilsn, line 86: Is the `type: ignore` comment because `block_or_time` is the wrong type? You should be able to do an `assert isinstance(block_or_time, U64)` or similar.

src/ethereum/genesis.py
```diff
@@ -143,28 +148,52 @@ def add_genesis_block(
...
-        number=Uint(0),
-        gas_limit=genesis.gas_limit,
-        gas_used=Uint(0),
-        timestamp=genesis.timestamp,
-        extra_data=genesis.extra_data,
-        mix_digest=hardfork.eth_types.Hash32(b"\0" * 32),
-        nonce=genesis.nonce,
-    )
+    for address, account in genesis.initial_accounts.items():
+        address = hardfork.utils.hexadecimal.hex_to_address(address)
```
SamWilsn, line 152: What's the advantage of storing a string in `initial_accounts` over doing the conversion in the constructor?

src/ethereum/genesis.py
```diff
@@ -80,22 +79,28 @@ def get_genesis_configuration(genesis_file: str) -> GenesisConfiguration:
...
         gas_limit=hex_to_uint(genesis_data["gasLimit"]),
         nonce=hex_to_bytes8(genesis_data["nonce"]),
-        timestamp=hex_to_u256(genesis_data["timestamp"]),
-        initial_balances=initial_balances,
+        timestamp=balance_str_to_u256(genesis_data["timestamp"]),
+        initial_accounts=genesis_data["alloc"],
     )
 
 
+def balance_str_to_u256(balance: str) -> U256:
```
SamWilsn, line 93: This function name is a misnomer since you're using it for the timestamp as well.

src/ethereum_spec_tools/sync.py
```diff
@@ -77,19 +81,39 @@ def module(self, name: str) -> Any:
...
-    def set_block(self, block_number: int) -> None:
+    def set_block(self, block_number: int, block_timestamp: int) -> None:
         """Set the block number and switch to the correct fork."""
         self.block_number = block_number
         self.active_fork_index = 0
-        while self.next_fork and block_number >= self.next_fork.block:
+        while (
+            self.next_fork
+            and hasattr(self.next_fork, "block")
+            and block_number >= self.next_fork.block
```
SamWilsn, line 91: I'm really leaning towards the OOP approach we discussed on the call a while back, where we have `Timestamp` and `BlockNumber` classes that each know how to test a block to see if it's time.

src/ethereum/shanghai/fork.py
SamWilsn: See https://github.com/ethereum/execution-specs/pull/761

minimayki6232 (approved):

src/ethereum_spec_tools/forks.py
petertdavies: The issue is that mypy has no way to know whether `mod` has a `MAINNET_FORK_BLOCK` attribute. The typecheck failures being suppressed are:
```
src/ethereum_spec_tools/forks.py:148: error: Module has no attribute "MAINNET_FORK_BLOCK"
src/ethereum_spec_tools/forks.py:152: error: Module has no attribute "MAINNET_FORK_TIMESTAMP"
```

src/ethereum_spec_tools/forks.py
```diff
@@ -5,23 +5,115 @@
...
-from typing import Any, Iterator, List, Optional, Type, TypeVar
+from typing import Any, Dict, Iterator, List, Optional, Type, TypeVar
 
 import ethereum
 
 H = TypeVar("H", bound="Hardfork")
 
 
+@functools.total_ordering
+class ForkCriteria:
```
SamWilsn, line 21: Ah. I missed the mark on explaining this I think. Or maybe my idea was terrible :rofl:

I was thinking we'd have:

```python
# src/ethereum/base_types.py (maybe)

class HeaderThing(Protocol):
    number: Uint
    timestamp: U256

class ForkActivation(ABC):
    @abstractmethod
    def activate_for(self, header: HeaderThing) -> bool:
        pass

class ByNumber(ForkActivation):
    def __init__(self, number: Uint):
        self.number = number

    def activate_for(self, header: HeaderThing) -> bool:
        return header.number >= self.number

class ByTimestamp(ForkActivation):
    def __init__(self, timestamp: U256):
        self.timestamp = timestamp

    def activate_for(self, header: HeaderThing) -> bool:
        return header.timestamp >= self.timestamp
```

```python
# src/ethereum/paris/__init__.py

MAINNET_ACTIVATION = ByTimestamp(1681338455)
```

SamWilsn: Now that I've written it up, I am not so certain it's a good fit. It's very object oriented. What are your thoughts?

petertdavies: I disagree with changing the `MAINNET_FORK_BLOCK` and `MAINNET_FORK_TIMESTAMP` constants. Forks exist in almost total ignorance of anything that happens outside of them. We consistently adopt hacky behaviour in `ethereum_spec_tools` to work around this and I don't see why we should change that policy here.

I think we need a `ForkCriteria` class with the same semantics as the following Rust type:
```rust
#[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
enum ForkCriteria {
    BlockNumber(u64),
    Timestamp(u64),
    Unscheduled,
}
```

`Ord` is required so that we can easily sort forks by activation order. Implementing that type clearly and succinctly in Python is surprisingly hard. I think your approach would work if we didn't need the ordering.

SamWilsn: > I disagree with changing the `MAINNET_FORK_BLOCK` and `MAINNET_FORK_TIMESTAMP` constants. [...] We consistently adopt hacky behaviour in `ethereum_spec_tools` to work around this and I don't see why we should change that policy here.

Hm, so you find `MAINNET_ACTIVATION = ByTimestamp(1681338455)` less readable than `MAINNET_FORK_TIMESTAMP = 1681338455`? Or maybe that it's adding too much complexity to the spec?

Personally I think we can change the spec to make it more compatible with the tooling, but _only as long as it doesn't make understanding the spec more difficult_. If we can avoid a hacky solution in `ethereum_spec_tools` **and** keep the spec understandable, I'd be the happiest.

> I think we need a `ForkCriteria` class with the same semantics as the following Rust type [...]

That's exactly what I was trying to replicate in Python, but with classes instead of an enum.

> I think your approach would work if we didn't need the ordering.

Could we add `@abstractmethod` magic methods (eg. `__lt__`) to `ForkActivation` to implement ordering?

SamWilsn (approved):

PR #751, opened 2023-04-21
re-run failed tests

gurukamath:
### What was wrong?
The test workflow on GitHub actions have been failing every now and then due to crashed workers.

### How was it fixed?
Re-run the failed test using `pytest-rerunfailures`

#### Cute Animal Picture

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -146,6 +146,7 @@ test =
     pytest>=6.2,<7
     pytest-cov>=2.12,<3
     pytest-xdist>=2.3.0,<3
+    pytest-rerunfailures>=11.1,<12
     GitPython>=3.1.0,<3.2
     filelock>=3.9.0,<3.10
     requests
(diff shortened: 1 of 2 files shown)


SamWilsn:
I am very leery of rerunning tests. I'd much rather they pass consistently on their own.

Do we still think they're crashing because of memory issues, or is this something else?

gurukamath:
> I am very leery of rerunning tests. I'd much rather they pass consistently on their own.
> 
> Do we still think they're crashing because of memory issues, or is this something else?

I do believe it has to do with memory issues on the Github worker based on the following observations
1. The issue is always in the `pypy3` environment which tends to be more momory intensive than `py3`
2. The issue never occurs when tax is run on a local system with more memory
3. The issue only occurs sporadically with `-n auto`, which probably means that the combined memory used by the pytest threads cannot be handled at certain points.

As far as the way we handle the issue is concerned, if we don't want to re-run the failed test, we can also handle it by running pytest in a single thread for `pypy3` env while running on Github actions. I'll propose an alternative PR.

PR #757, opened 2023-05-01
Run single thread pytest in github actions

gurukamath:
Alternative to #751 

Instead of re-running the tests as proposed in #751, we simply run a single thread of `pytest` if we are on GitHub actions (for the `pypy` env). When run locally, `pytest` automatically uses `-n auto`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,6 +27,8 @@ jobs:
       - name: Run Tox (PyPy)
         if: ${{ matrix.python == 'pypy-3.8' }}
         run: tox -e pypy3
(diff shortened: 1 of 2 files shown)


petertdavies:
Why is this only for PyPy and not CPython?

gurukamath:
> Why is this only for PyPy and not CPython?

PyPy is the only environment that is facing memory issues currently. We don't want to run single thread CI jobs unless we absolutely have to in my opinion. If CPython has similar issues in the future, we can re-visit this at that point.

SamWilsn (approved):

PR #764, opened 2023-05-25
Add opcodes to "Lists" documentation

shemnon:
Add a summary of used, pending, and proposed opcodes for the EVM. 
Also adds a list describing each 16 opcode block,

### DIFF
--- a/lists/evm/README.md
+++ b/lists/evm/README.md
@@ -1,0 +1,35 @@
+EVM Summary Information
+=======================
+
+This is a set of tables summarizing current and pending EVM information.
+
+Where information between the executable spec and this file conflict,
+the executable spec prevails.
+
+[Opcodes](opcodes.md)
+---------------------
+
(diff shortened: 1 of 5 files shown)


shemnon:
@SamWilsn @lightclient can I get a review on this?

petertdavies (approved):
I think this is a perfectly reasonable. In the long term it would be nice to auto generate from the specs, but we are nowhere near that at the moment.

shemnon:
`opcodes.md` and `opcode-blocks.md` could be auto generated from the spec and other pages, but I think the more uniquely useful page is the `pending-opcodes.md`, where opcodes that are targeting the next release are laid out, and it could serve as an aid in final selection of opcode numbers.

chfast (approved):
We have applied changes in https://github.com/ethereum/evmone/pull/665 and we are happy with it.

I'm only slightly concerned that there is some duplication of information in these 3 tables.

PR #767, opened 2023-05-29
Automatically detect expected exceptions

### LINKED ISSUE #760: Automatically detect tests that throw exceptions
In the new testing fixtures, a test that is supposed to throw an exception has an `expectedException` field within the fixture. This can be used to detect where an exception is to be expected. Currently, the tests where exceptions are expected are hard coded in the tests in a large number of modules.

gurukamath:
(closes #760 )

### What was wrong?
The tests that throw exceptions were currently hardcoded in the testing framework.

Related to Issue #760 

### How was it fixed?
Automatically detect which test fixtures have the `expectException` fields. This PR also does some refactoring of the testing framework, making it more streamlined and a bit more efficient.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/fixture_loader.py
+++ b/src/ethereum_spec_tools/evm_tools/fixture_loader.py
@@ -6,7 +6,7 @@ testing framework.
 
 import importlib
 from abc import ABC, abstractmethod
-from typing import Any, List, Tuple
+from typing import Any, Optional, Tuple
 
 from ethereum import rlp
 from ethereum.base_types import U64, U256, Bytes0
@@ -22,6 +22,14 @@ from ethereum.utils.hexadecimal import (
 )
 from ethereum_spec_tools.forks import Hardfork
 
+RLP_DECODING_EXCEPTIONS = (
+    "RLP_VALUESIZE_MORE_AVAILABLEINPUTLENGTH",
+    "RLP_ExpectedAsList",
+    "INPUT_UNMARSHAL_SIZE_ERROR",
+    "RLP_BODY_UNMARSHAL_ERROR",
+    "INPUT_UNMARSHAL_ERROR",
+)
+
 
 class UnsupportedTx(Exception):
     """Exception for unsupported transactions"""
@@ -30,6 +38,12 @@ class UnsupportedTx(Exception):
         super().__init__(error_message)
         self.encoded_params = encoded_params
         self.error_message = error_message
+
+
+class ExpectedRLPException(Exception):
+    """Expected RLP exceptions"""
+
+    pass
 
 
 class BaseLoad(ABC):
@@ -124,7 +138,7 @@ class BaseLoad(ABC):
         pass
 
     @abstractmethod
-    def json_to_blocks(self, json_data: Any) -> Any:
+    def json_to_block(self, json_data: Any) -> Any:
         """Converts json block data to a list of blocks"""
         pass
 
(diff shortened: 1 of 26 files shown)


src/ethereum_spec_tools/evm_tools/fixture_loader.py
```diff
@@ -346,55 +360,57 @@ def json_to_withdrawals(self, raw: Any) -> Any:
...
-                withdrawals = tuple(
-                    self.json_to_withdrawals(wd)
-                    for wd in json_block["withdrawals"]
-                )
-                parameters.append(withdrawals)
+        if "rlp" in json_block:
+            # Always decode from rlp
+            block_rlp = hex_to_bytes(json_block["rlp"])
+            try:
+                block = rlp.decode_to(self.Block, block_rlp)
```
SamWilsn, line 378: Do we fail the test if an expected exception doesn't occur?

src/ethereum_spec_tools/evm_tools/fixture_loader.py
```diff
@@ -346,55 +360,57 @@ def json_to_withdrawals(self, raw: Any) -> Any:
...
-            parameters = [
-                header,
-                transactions,
-                uncles,
-            ]
+        block_exception = None
+        for key, value in json_block.items():
+            if key.startswith("expectException"):
+                block_exception = value
+                break
```
SamWilsn, line 372: Might it be better to make a `Mapping[str, List[Type[Exception]]` and handle it all at the top level, or are our exception types too vague?

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -360,11 +360,11 @@ def parse_json_tx(self) -> Iterator[Tuple[int, Any]]:
...
             # for v, r, s
-            if not json_tx["v"]:
+            if not json_tx.get("v"):
                 json_tx["v"] = "0x00"
-            if not json_tx["r"]:
+            if not json_tx.get("r"):
                 json_tx["r"] = "0x00"
-            if not json_tx["s"]:
+            if not json_tx.get("s"):
                 json_tx["s"] = "0x00"
```
SamWilsn, line 368: What about:

```python
            json_tx["v"] = json_tx.get("v") or "0x00"
            json_tx["r"] = json_tx.get("r") or "0x00"
            json_tx["s"] = json_tx.get("s") or "0x00"
```

src/ethereum_spec_tools/evm_tools/fixture_loader.py
gurukamath: There are 2 cases here

1. Some other exception is raised which is not an `ExpectedRLPException`. In this case, the same exception will be raised and the test will fail. We only catch the `ExpectedRLPException` [here](https://github.com/gurukamath/execution-specs/blob/2f2da0c601e9870a52e41e25aacfe751de699207/tests/helpers/load_state_tests.py#L76)
2. No exception is raised. In this case, the further processing will continue. However, one of the later checks will have to fail. Assertion of state_root, block_hash etc. for example.

src/ethereum_spec_tools/evm_tools/fixture_loader.py
gurukamath: This is only for parsing if the fixture expects any exception. All exception handling is done at the topmost level in the `run_blockchain_st_test` function

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
gurukamath: Updated

src/ethereum_spec_tools/evm_tools/fixture_loader.py
SamWilsn: And it's guaranteed that one will fail?

src/ethereum_spec_tools/evm_tools/fixture_loader.py
SamWilsn: There is the bit here: https://github.com/ethereum/execution-specs/pull/767/files#diff-22ed9d4c129a5cb0772b573f1d75a7468e431a3f75131680d9a5e12c5e3e456fR381

gurukamath: I have moved this up directly into the `run_blockchain_st_test` function

src/ethereum_spec_tools/evm_tools/fixture_loader.py
gurukamath: Yes. When we successfully process an invalid block, the last_block_hash will not match. The post state will also not match in most cases.

However, in the most recent upsate I pushed, I check explicitly for the this with `with pytest.raises(RLPDecodingError):`. So we don't have to rely on the above checks failing.

SamWilsn (approved):

PR #769, opened 2023-05-30
(Partially) update Cancun PR table

haltman-at:
I figured I'd update the Cancun PR table.  I removed the column for EIP-6475 and updated the remaining three columns.  Well, partly -- I actually largely left the column for 4844 alone, because that one's complicated and I didn't want to take the time right now to figure out everything going on there.  But I updated the other two columns, and updated one entry in the 4844 column.

I haven't added columns for the ones that are merely CFI for now.  (One of which still hasn't even been merged as a proper EIP yet!)


timbeiko:
Sorry, I missed this! Will close given it's now superseded by https://github.com/ethereum/execution-specs/pull/792

PR #772, opened 2023-06-01
t8n:: move env to a separate file

winsvega:
To get better with python I start with refactoring the existing code.
I first try to make it better without breaking existing functionality.


gurukamath:
The code needs to be formatted but otherwise, it looks good to me. To identify the issues with the code formatting, you could run the `isort`, `black`, `flake8` and `mypy` commands from [tox.ini](https://github.com/ethereum/execution-specs/blob/635f4bedbd65157767b1a98e43f25e665ea3eaa0/tox.ini#L11)

petertdavies:
> I run ` tox -e py3` but how can I make it change write the corrections instead of reporting?

You need to run `isort src` and `black src` manually. `tox` provides the `--check` flag which doesn't rewrite the file. You may need to `pip install black` and `pip install isort`, if they aren't already installed on your system or in you venv.

gurukamath (approved):
LGTM!

PR #791, opened 2023-06-06
Fetch from Github only when required

gurukamath:
### What was wrong?
Currently, the testing framework fetches updates from Github every time the `pytest` command is run. This is not essential.

Related to Issue #

### How was it fixed?
First check if the required fixtures are available locally (perhaps they have been checkout out in previous runs, which is overwhelmingly going to be the case). If some details are missing, only then try to fetch them from GitHub.

#### Cute Animal Picture


petertdavies (approved):
LGTM

PR #794, opened 2023-06-13
Handle create collisions

petertdavies:
### What was wrong?

Some tests involving an obscure edgecase were not supported.

### How was it fixed?

Add support for the edgecase to the specs and optimized state.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/arrow_glacier/state.py
+++ b/src/ethereum/arrow_glacier/state.py
@@ -593,7 +593,8 @@ def get_storage_original(state: State, address: Address, key: Bytes) -> U256:
     key:
         Key of the storage slot.
     """
-    # When an account is created, preexisting storage is ignored.
+    # In the transaction where an account is created, its preexisting storage
+    # is ignored.
     if address in state._created_accounts:
         return U256(0)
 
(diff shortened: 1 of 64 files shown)


src/ethereum/arrow_glacier/state.py
```diff
@@ -85,6 +87,8 @@ def commit_transaction(state: State) -> None:
         The state.
     """
     state._snapshots.pop()
+    if state._snapshots == []:
```
SamWilsn, line 90:
```suggestion
    if not state._snapshots:
```

src/ethereum/london/state.py
```diff
@@ -567,6 +593,10 @@ def get_storage_original(state: State, address: Address, key: Bytes) -> U256:
     key:
         Key of the storage slot.
     """
+    # When an account is created, preexisting storage is ignored.
```
SamWilsn, line 596:
```suggestion
    # In the transaction where an account is created, its preexisting storage is ignored.
```

SamWilsn (commented):
Thoughts?

SamWilsn (approved):

Issue #773, opened 2023-06-03
t8n tool errors


winsvega:
Here is the results of t8n execution on the state tests (skipping ConstantinopleFix as it is not declared in tool)

Fixed
 - [x] info: (GeneralStateTests/stCreate2/RevertInCreateInInitCreate2, fork: Berlin, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x305060..`) (empty contract collision)
- [x] info: (GeneralStateTests/stCreateTest/CreateTransactionHighNonce, fork: London, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x600160..`) (tr has nonce overflow the allowed value) 
- [x] info: (GeneralStateTests/stTransactionTest/HighGasPrice, fork: Berlin, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x..`) (tr has gaslimit x gasprice overflow)
- [x] info: (GeneralStateTests/stSStoreTest/InitCollision, fork: Berlin, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x600060..`) (empty account collision)
- [x] info: (GeneralStateTests/stRevertTest/RevertInCreateInInit, fork: Berlin, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x305060..`) (empty accounts?)
- [x] info: (GeneralStateTests/stExample/mergeTest, fork: Merge, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x00..`) (merge support?) 
- [x] info: (GeneralStateTests/stExample/invalidTr, fork: Berlin, TrInfo: d: 0, g: 0, v: 0, TrData: ` 0x..`) (intrinsic gas exception)
- [x] info: (GeneralStateTests/stEIP2930/addressOpcodes, fork: Berlin, TrInfo: d: 0, g: 0, v: 0, TrData: `:label valid 0x1a8451..`)
- [x] info: (GeneralStateTests/stEIP2930/coinbaseT01, fork: Berlin, TrInfo: d: 1, g: 0, v: 0, TrData: `:label T1baseInList 0x693c61..`)
- [x]  info: (GeneralStateTests/stEIP2930/coinbaseT2, fork: London, TrInfo: d: 0, g: 0, v: 0, TrData: `:label T2baseInList 0x693c61..`)
[...]

PR #797, opened 2023-06-20
fix t8n tool errors

gurukamath:
### What was wrong?
Some edge cases were not handled properly by t8n

Related to Issue #773 

### How was it fixed?
1. Fix bug to accept state fork `ConstantinopleFix` instead of `Constantinople`
2. Update evm tools tests to use `GeneralStateTests` fixtures
3. Handle invalid transactions without crashing the `t8n` tool.
4. Handle a couple of testing edge cases.
 
#### Cute Animal Picture

### DIFF
--- a/src/ethereum/arrow_glacier/fork.py
+++ b/src/ethereum/arrow_glacier/fork.py
@@ -394,7 +394,7 @@ def check_transaction(
     base_fee_per_gas: Uint,
     gas_available: Uint,
     chain_id: U64,
-) -> Tuple[Address, U256]:
+) -> Tuple[Address, Uint]:
     """
     Check if the transaction is includable in the block.
 
@@ -731,7 +731,7 @@ def pay_rewards(
 
 def process_transaction(
     env: vm.Environment, tx: Transaction
-) -> Tuple[U256, Tuple[Log, ...], bool]:
+) -> Tuple[Uint, Tuple[Log, ...], bool]:
     """
     Execute a transaction against the provided environment.
 
@@ -763,13 +763,10 @@ def process_transaction(
     sender = env.origin
     sender_account = get_account(env.state, sender)
 
-    try:
-        if isinstance(tx, FeeMarketTransaction):
-            gas_fee = tx.gas * tx.max_fee_per_gas
-        else:
-            gas_fee = tx.gas * tx.gas_price
-    except ValueError as e:
-        raise InvalidBlock from e
+    if isinstance(tx, FeeMarketTransaction):
+        gas_fee = tx.gas * tx.max_fee_per_gas
+    else:
+        gas_fee = tx.gas * tx.gas_price
 
     ensure(sender_account.nonce == tx.nonce, InvalidBlock)
     ensure(sender_account.balance >= gas_fee + tx.value, InvalidBlock)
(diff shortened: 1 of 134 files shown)


src/ethereum/arrow_glacier/fork.py
```diff
@@ -762,10 +763,13 @@ def process_transaction(
...
-    if isinstance(tx, FeeMarketTransaction):
-        gas_fee = tx.gas * tx.max_fee_per_gas
-    else:
-        gas_fee = tx.gas * tx.gas_price
+    try:
+        if isinstance(tx, FeeMarketTransaction):
+            gas_fee = tx.gas * tx.max_fee_per_gas
+        else:
+            gas_fee = tx.gas * tx.gas_price
+    except ValueError as e:
```
SamWilsn, line 771: What raises `ValueError` here? Integer overflow?

If it is an integer overflow, we should handle this _for all_ integer overflows, and not just in these specific cases. Should we change `base_types` so that it throws a specific `OverflowError` that subclasses `ValueError`?

src/ethereum_spec_tools/evm_tools/utils.py
```diff
@@ -80,6 +85,8 @@ def get_module_name(forks: Any, options: Any, stdin: Any) -> Tuple[str, int]:
     """
     Get the module name and the fork block for the given state fork.
     """
+    if options.state_fork.upper() in UNSUPPORTED_FORKS:
```
SamWilsn, line 88: Should use [`str.casefold`](https://docs.python.org/3/library/stdtypes.html#str.casefold) for case-insensitive comparisons.

src/ethereum/arrow_glacier/fork.py
gurukamath: > What raises `ValueError` here? Integer overflow?

Yes. This is due to overflows. This change has been made because the testing framework expects an `InvalidBlock` exception (as it should) but receives `ValueError`.


> If it is an integer overflow, we should handle this _for all_ integer overflows, and not just in these specific cases. Should we change `base_types` so that it throws a specific `OverflowError` that subclasses `ValueError`?

I did consider modifying the `base_types`, but raising an `InvalidBlock` exception (or a subclass of it) directly in the `base_types` does not seem right to me. (Although one could argue that the spec base types are used only in the context of processing blocks and so, an `InvalidBlock` exceptions is not out of place there). What do you think?

Alternatively, I also considered modifying the testing framework to handle the `ValueError` but that seems too generic of an exception to expect in the tests. Would be better to keep the expected exceptions as specific as possible.

src/ethereum_spec_tools/evm_tools/utils.py
gurukamath: Will update.

src/ethereum/arrow_glacier/fork.py
petertdavies: Another approach, which we take with gas tracking in the EVM, is to make everything a `Uint` instead of a `U256`. If we do that the overflow won't happen at all and instead `InvalidBlock` will get raised a few lines later.

gurukamath: Updated to use `Uint` for all gas.

SamWilsn (approved):

PR #811, opened 2023-07-11
Unify optimized implementations

petertdavies:
### What was wrong?

Having to duplicate the code in `ethereum_optimized` on a per-fork basis was unwieldy and annoying.

### How was it fixed?

The optimized implementations now consist of a function that takes a fork and returns a dictionary of optimized patches.

#### Cute Animal Picture


src/ethereum_optimized/__init__.py
```diff
@@ -13,46 +13,77 @@
...
+        "spurious_dragon",
+        "byzantium",
+        "constantinople",
+        "istanbul",
+        "muir_glacier",
+        "berlin",
+        "london",
+        "arrow_glacier",
+        "gray_glacier",
+    ):
```
SamWilsn, line 81: Is there some way we can put this information into the fork itself and retrieve it using functions from [`ethereum_spec_tools/forks.py`](https://github.com/ethereum/execution-specs/blob/master/src/ethereum_spec_tools/forks.py)?

I _really_ want to keep metadata about forks (like these pre-/post-merge lists) in as few places as possible.

src/ethereum_optimized/fork.py
```diff
@@ -0,0 +1,60 @@
...
+            int(epoch_number),
+            header_hash,
+            header.mix_digest,
+            int.from_bytes(header.nonce, "big"),
+            (U256_CEIL_VALUE // header.difficulty).to_be_bytes32(),
+        )
+
+        ensure(result, InvalidBlock)
+
+    return {k: v for (k, v) in locals().items() if not k.startswith("_")}
```
SamWilsn, line 60: This seems like a footgun. If someone adds a non-prefixed local to the function down the road, it'll get added to the dict, and it's very easy to imagine myself using a temporary variable without thinking.

src/ethereum_optimized/state_db.py
```diff
@@ -0,0 +1,322 @@
...
+        state.destroyed_accounts.add(address)
+        set_account(state, address, get_account_optional(state, address))
+
+    def mark_account_created(state: State, address: Address) -> None:
+        """
+        See `state`.
+        """
+        state.created_accounts.add(address)
+
+    return locals()
```
SamWilsn, line 322: Quickly thinking about `locals`, could you maybe use a class or module to hold all the replacement functions/variables and iterate through that?

Something like:

```python
class Patch:
    @dataclass
    class State:
        ...

    def close_state(...):
        ...

return {k: v for k, v in vars(Foo).items() if not k.startswith('_')}
```

petertdavies: I changed to an approach where everything is manually added to a patches dictionary with a decorator.

src/ethereum_optimized/__init__.py
petertdavies: Having thought about this it is possible and I have done it.

src/ethereum_optimized/__init__.py
```diff
@@ -13,46 +13,68 @@
...
+
+    for fork in forks:
+        monkey_patch_optimized_state_db(fork.short_name, state_path)
+
+        # Only patch the POW code on POW forks
+        if (
+            isinstance(fork.criteria, ByBlockNumber)
+            and fork.short_name != "paris"
+        ):
+            monkey_patch_optimized_spec(fork.short_name)
```
SamWilsn, line 80: Would the `Hardfork.consensus` property from #820 work here?

SamWilsn (approved):

PR #819, opened 2023-07-28
add EIP 6780: only SELFDESTRUCT in same transaction

gballet:
This is my first shot at implementing EIP 6780 in EELS. I'm no Python expert, so this will probably require an extra cautious look-over.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/cancun/vm/instructions/system.py
+++ b/src/ethereum/cancun/vm/instructions/system.py
@@ -76,6 +76,7 @@ def generic_create(
     )
 
     evm.accessed_addresses.add(contract_address)
+    evm.env.created_contracts.add(contract_address)
 
     create_message_gas = max_message_call_gas(Uint(evm.gas_left))
     evm.gas_left -= create_message_gas
@@ -524,15 +525,18 @@ def selfdestruct(evm: Evm) -> None:
     # beneficiary).
     set_account_balance(evm.env.state, originator, U256(0))
 
-    # register account for deletion
-    evm.accounts_to_delete.add(originator)
+    # Only continue if the contract has been created in the same tx
+    if originator in evm.env.created_contracts:
 
-    # mark beneficiary as touched
-    if account_exists_and_is_empty(evm.env.state, beneficiary):
-        evm.touched_accounts.add(beneficiary)
+        # register account for deletion
+        evm.accounts_to_delete.add(originator)
 
-    # HALT the execution
-    evm.running = False
+        # mark beneficiary as touched
+        if account_exists_and_is_empty(evm.env.state, beneficiary):
+            evm.touched_accounts.add(beneficiary)
+
+        # HALT the execution
+        evm.running = False
 
     # PROGRAM COUNTER
     pass
--- a/src/ethereum/cancun/vm/instructions/system.py
+++ b/src/ethereum/cancun/vm/instructions/system.py
@@ -76,7 +76,6 @@ def generic_create(
     )
 
     evm.accessed_addresses.add(contract_address)
-    evm.env.created_contracts.add(contract_address)
 
     create_message_gas = max_message_call_gas(Uint(evm.gas_left))
     evm.gas_left -= create_message_gas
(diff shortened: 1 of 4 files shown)


src/ethereum/cancun/vm/instructions/system.py
```diff
@@ -519,15 +520,18 @@ def selfdestruct(evm: Evm) -> None:
...
+        evm.accounts_to_delete.add(originator)
 
-    # HALT the execution
-    evm.running = False
+        # mark beneficiary as touched
+        if account_exists_and_is_empty(evm.env.state, beneficiary):
+            evm.touched_accounts.add(beneficiary)
+
+        # HALT the execution
+        evm.running = False
```
SamWilsn, line 534: Hm, I think this'll also lead to an issue. If `evm.running` is not set to `False`, then you need to increment the program counter.

src/ethereum/cancun/vm/instructions/system.py
```diff
@@ -519,15 +520,18 @@ def selfdestruct(evm: Evm) -> None:
...
-    if account_exists_and_is_empty(evm.env.state, beneficiary):
-        evm.touched_accounts.add(beneficiary)
+        # register account for deletion
+        evm.accounts_to_delete.add(originator)
 
-    # HALT the execution
-    evm.running = False
+        # mark beneficiary as touched
+        if account_exists_and_is_empty(evm.env.state, beneficiary):
+            evm.touched_accounts.add(beneficiary)
```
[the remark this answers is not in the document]
petertdavies: Yes, this should be outside the `if` block.

src/ethereum/cancun/vm/instructions/system.py
petertdavies: Execution of the stack frame should end and return regardless of whether the contract was marked for deletion. So this should also be outside the `if` block.

src/ethereum/cancun/vm/instructions/system.py
```diff
@@ -522,15 +523,18 @@ def selfdestruct(evm: Evm) -> None:
     # beneficiary).
     set_account_balance(evm.env.state, originator, U256(0))
```
gurukamath, line 524: Balance transfer also has to only happen under certain conditions i.e. if the contract was created in the same transaction or if the beneficiary and originator are different.

gurukamath:
@SamWilsn @petertdavies Have updated the implementation

petertdavies (approved):

src/ethereum/cancun/vm/instructions/system.py
```diff
@@ -515,17 +515,24 @@ def selfdestruct(evm: Evm) -> None:
...
-    # First Transfer to beneficiary
-    set_account_balance(
-        evm.env.state, beneficiary, beneficiary_balance + originator_balance
-    )
-    # Next, Zero the balance of the address being deleted (must come after
-    # sending to beneficiary in case the contract named itself as the
-    # beneficiary).
-    set_account_balance(evm.env.state, originator, U256(0))
+    if (
+        originator in evm.env.state._created_accounts
```
SamWilsn, line 519: If `_created_accounts` is going to be accessed outside the module in which it was defined, we should drop the leading underscore.

src/ethereum/cancun/vm/instructions/system.py
```diff
@@ -515,17 +515,24 @@ def selfdestruct(evm: Evm) -> None:
...
+        # First Transfer to beneficiary
+        set_account_balance(
+            evm.env.state,
+            beneficiary,
+            beneficiary_balance + originator_balance,
+        )
+        # Next, Zero the balance of the address being deleted (must come after
+        # sending to beneficiary in case the contract named itself as the
+        # beneficiary).
+        set_account_balance(evm.env.state, originator, U256(0))
```
SamWilsn, line 531: Something about these `if` conditions is bothering me, but I can't quite put my finger on it.

I haven't thought to deeply about the logic, but would something like this be easier to understand?

```python
# Increase beneficiary's balance
move_ether(
    evm.env.state,
    originator,
    beneficiary,
    originator_balance,
)

if originator in evm.env.state._created_accounts:
    # Zero the balance of the address being deleted (must come after
    # sending to beneficiary in case the contract named itself as the
    # beneficiary).
    set_account_balance(evm.env.state, originator, U256(0))

    # Register account for deletion
    evm.accounts_to_delete.add(originator)
````

Comment (gurukamath) [reply src/ethereum/cancun/vm/instructions/system.py]: You're right! This is in fact simpler to understand. Updated

File: src/ethereum/cancun/vm/instructions/system.py (hunk 4 above)
Comment (gurukamath) [reply src/ethereum/cancun/vm/instructions/system.py]: Updated

File: src/ethereum/cancun/vm/instructions/system.py [hunk 6]
```diff
@@ -512,20 +512,19 @@ def selfdestruct(evm: Evm) -> None:
...
-    # sending to beneficiary in case the contract named itself as the
-    # beneficiary).
-    set_account_balance(evm.env.state, originator, U256(0))
 
-    # register account for deletion
-    evm.accounts_to_delete.add(originator)
+    # register account for deletion only if it was created
+    # in the same transaction
+    if originator in evm.env.state.created_accounts:
+        evm.accounts_to_delete.add(originator)
```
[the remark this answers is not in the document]
gurukamath: Updated

SamWilsn (approved):

PR #824, opened 2023-08-24
Rename `v` to `y_parity`

### LINKED ISSUE #814: 2930 and 1559 transactions do not have a `v`.
https://github.com/ethereum/execution-specs/blob/2da1ad34bb0f17bf61d41bdbad6fa8c814286dcf/src/ethereum/berlin/fork_types.py#L81
https://github.com/ethereum/execution-specs/blob/2da1ad34bb0f17bf61d41bdbad6fa8c814286dcf/src/ethereum/london/fork_types.py#L101

These should both be named `yParity`.  `v` has a very well defined meaning in legacy transactions and EIP-155 transactions, but it is not mentioned anywhere in the EIP-2930 or EIP-1559 specification.  What *is* mentioned in both of those specifications is a `yParity` field that sits in a similar place in the RLP structure, but it's definition doesn't match that of the `v` value defined in either legacy transactions or EIP-155 transactions.

Glancing at the code here, it appears you are just putting `yParity` into a property called `v`, but this will lead to a great amount of confusion since it doesn't actually match the definition of `v` anywhere (neither in the yellow paper nor in any EIP).

Hopefully the fix for this is as simple as a rename, but I'm not confident enough in Python being able to do rename refactoring to want to attempt a PR myself.

petertdavies:
### What was wrong?

Fixes #814 

### How was it fixed?

`v` has now been renamed to `y_parity` in the appropriate places. For compatibility reasons, we accept either `v` or `y_parity` in JSON and emit both.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
@@ -178,7 +178,7 @@ class Txs:
 
         # tf tool might provide None instead of 0
         # for v, r, s
-        raw_tx["v"] = raw_tx.get("v") or "0x00"
+        raw_tx["v"] = raw_tx.get("v") or "0x00" or raw_tx.get("y_parity")
         raw_tx["r"] = raw_tx.get("r") or "0x00"
         raw_tx["s"] = raw_tx.get("s") or "0x00"
 
@@ -276,6 +276,9 @@ class Txs:
         json_tx["r"] = hex(r)
         json_tx["s"] = hex(s)
         json_tx["v"] = hex(y + v_addend)
+
+        if v_addend == 0:
+            json_tx["y_parity"] = json_tx["v"]
 
 
 @dataclass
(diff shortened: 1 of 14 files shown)


src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -178,7 +178,7 @@ def parse_json_tx(self, raw_tx: Any) -> Any:
 
         # tf tool might provide None instead of 0
         # for v, r, s
-        raw_tx["v"] = raw_tx.get("v") or "0x00"
+        raw_tx["v"] = raw_tx.get("v") or "0x00" or raw_tx.get("y_parity")
```
SamWilsn, line 181: `"0x00"` is always truthy, so you probably want:

```suggestion
        raw_tx["v"] = raw_tx.get("v") or raw_tx.get("y_parity") or "0x00"
```

SamWilsn (changes requested):

SamWilsn (approved):

Issue #795, opened 2023-06-15
t8n better exception reporting


winsvega:
Currently the transaction rejected message looks like this:

```
Failed transaction:
```

But we can check the precise exception in tests to why it has failed. (field overflow, signature values, gaslimit etc)

PR #828, opened 2023-08-31
Implement Evm Trace

### LINKED ISSUE #624: Add support for EVM trace specification

gurukamath:
(closes #624 )

### What was wrong?
The specs don't currently implement the full evm trace.

Related to Issue #624 

### How was it fixed?
Implemented full evm trace and integrated it fully with `t8n`.

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->/cloudfront-us-east-1.images.arcpublishing.com/gray/T6KDDQFEVVBGTPZTFVQFCZOOZU.jpg)


gurukamath:
Since the evm tests fail for the forks before Shanghai, I have added a temporary [if](https://github.com/gurukamath/execution-specs/blob/f588b28a77b3c62451181681ef7bb4d7f7358fa6/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L259) condition. This is just so we see all the tests passing in the CI. Once the Shanghai changes have been reviewed and are ported over to the other forks, this if condition goes away before merging all the changes with master.

src/ethereum/trace.py
```diff
@@ -0,0 +1,86 @@
...
+Introduction
+------------
+
+Defines the functions required for creating evm traces during execution.
+"""
+
+from typing import Any
+
+
+def capture_tx_start(env: Any) -> None:
```
SamWilsn, line 20: I'd recommend using `object` instead of `Any` for the arguments to these.

src/ethereum/trace.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    Capture the state at the beginning of a transaction.
+    """
+    pass
+
+
+def capture_tx_end(
```
SamWilsn, line 27: Pretty insignificant, but what do you think of `trace_tx_end`, `trace_precompile_start`, etc instead of `capture_*`?

src/ethereum/trace.py
```diff
@@ -0,0 +1,86 @@
...
+    env: Any, gas_used: int, output: bytes, has_erred: bool
+) -> None:
+    """
+    Capture the state at the end of a transaction.
+    """
+    pass
+
+
+def capture_precompile_start(evm: Any, address: Any) -> None:
```
SamWilsn, line 36: My only architectural concern is that with multiple functions like you have here, a tracing implementation cannot listen for _all_ events, only specific ones. So lets say we create a logging tool that prints to stdout each trace event. The logging tool doesn't understand the events, just prints out the arguments.

If we have to add a new trace event much later on, now we have to go and update every tool built on top of the specs because those tools have to inject a new function to get notified about those events.

If we were in Rust, I'd suggest something like:

```rust
enum TraceEvent {
    PrecompileStart(address: Address),
    PrecompileEnd,
    // ...
}

fn evm_trace(evm: Evm, event: TraceEvent) {
    // ...
}
```

In Python... it doesn't feel as nice, but maybe:

```python
@dataclass
class PrecompileStart:
    address: object

@dataclass
class PrecompileEnd:
    pass

TraceEvent = Union[PrecompileStart, PrecompileEnd, ...]

def evm_trace(evm: object, event: TraceEvent) -> None:
    pass
```

Is this too much overhead for a theoretical concern? Too many new types cluttering up the spec?

tests/conftest.py
```diff
@@ -43,9 +54,18 @@ def pytest_configure(config: Config) -> None:
...
+        trace.capture_op_exception = capture_op_exception
+        trace.capture_gas_and_refund = capture_gas_and_refund
+        trace.capture_tx_end = capture_tx_end
+        trace.output_traces = output_traces
+        trace.capture_evm_stop = capture_evm_stop
+        trace.capture_precompile_start = capture_precompile_start
+        trace.capture_precompile_end = capture_precompile_end
+        # config.option.__dict__["log_cli_level"] = "10"
+        # config.option.__dict__["log_format"] = "%(message)s"
+        # setattr(ethereum, "evm_trace", evm_trace)
```
SamWilsn, line 68: Are these debugging comments?

src/ethereum_spec_tools/evm_trace.py
```diff
@@ -1,66 +1,212 @@
...
+        return
+    else:
+        capture_op_start(evm, op)
+
+
+def capture_gas_and_refund(evm: Any, gas_cost: Any) -> None:
+    """
+    Capture the gas cost and refund during opcode execution.
+    """
+    if not any(evm.env.traces):
```
SamWilsn, line 203: Do you mean to iterate through `evm.env.traces` for truthy values, or is this simply an empty check?

If it's the latter, I'd suggest:

```suggestion
    if not evm.env.traces:
```

src/ethereum_spec_tools/evm_trace.py
```diff
@@ -1,66 +1,212 @@
...
+        # In contract creation transactions, there may not be any traces
+        return
+    else:
+        last_trace = evm.env.traces[-1]
```
SamWilsn, line 207: This might be a bit easier to read:

```suggestion

    last_trace = evm.env.traces[-1]
```

src/ethereum_spec_tools/evm_trace.py
```diff
@@ -1,66 +1,212 @@
...
+
+def capture_op_exception(evm: Any) -> None:
+    """Capture the state in case of exceptions."""
+    if any(evm.env.traces):
+        last_trace = evm.env.traces[-1]
+    if (
+        not any(evm.env.traces)
+        or last_trace.errorTraced
+        or last_trace.depth == evm.message.depth
+    ):
```
SamWilsn, line 130: This might be a bit more readable:

```suggestion
    last_trace = None
    if evm.env.traces:
        last_trace = evm.env.traces[-1]

    if (
        not last_trace
        or last_trace.errorTraced
        or last_trace.depth == evm.message.depth
    ):
```

SamWilsn (commented):
Probably still here because you've only done Shanghai so far, but just in case, you'll want to remove this as well:

https://github.com/ethereum/execution-specs/blob/fade6e95726c9c0b47b779fd8873b5737fbea3b3/src/ethereum/__init__.py#L31-L37

gurukamath:
> Probably still here because you've only done Shanghai so far, but just in case, you'll want to remove this as well:
> 
> https://github.com/ethereum/execution-specs/blob/fade6e95726c9c0b47b779fd8873b5737fbea3b3/src/ethereum/__init__.py#L31-L37

Yes. This will also go after porting the changes to the other forks.

src/ethereum_spec_tools/evm_trace.py
```diff
@@ -1,66 +1,212 @@
...
+        if field.name not in EXCLUDE_FROM_OUTPUT
+        and getattr(trace, field.name) is not None
+    }
+
+    json.dump(dict_trace, json_file, separators=(",", ":"))
+    json_file.write("\n")
+
+
+def output_traces(
+    traces: Any, tx_hash: bytes, output_basedir: str = "."
```
danceratopz, line 168:
```suggestion
    traces: Any, tx_index: int, tx_hash: bytes, output_basedir: str = "."
```

danceratopz (changes requested):
Really looking forward to using this!

I quickly tested the `--trace` flag (combined with the changes required from #823) against https://github.com/ethereum/execution-spec-tests/pull/289; also slightly modified due to differences in filename (more on that below):

```diff
diff --git a/src/evm_transition_tool/transition_tool.py b/src/evm_transition_tool/transition_tool.py
index ae24b319..14c1ee2b 100644
--- a/src/evm_transition_tool/transition_tool.py
+++ b/src/evm_transition_tool/transition_tool.py
@@ -242,6 +242,7 @@ class TransitionTool:
         traces: List[List[Dict]] = []
         for i, r in enumerate(receipts):
             trace_file_name = f"trace-{i}-{r['transactionHash']}.jsonl"
+            # trace_file_name = f"spec-trace-{r['transactionHash']}.json"
             if debug_output_path:
                 shutil.copy(
                     os.path.join(temp_dir.name, trace_file_name),
```
Here's the execution-spec-tests command:
```
fill --evm-bin=/path/to/ethereum/execution-specs/venv/bin/ethereum-spec-evm --t8n-dump-dir=/tmp/ethereum-spec-evm/ --fork=Shanghai -v --traces
```

Some background: The interface used in execution-spec-tests to execute the `ethereum-spec-evm t8n` tool is shared with geth's `evm t8n` and `evmone-t8n`, so any differences are quickly highlighted. Would it be possible to change the filename as below, to make them consistent across these tools? The main effort will be adding the index of the transaction processed by the t8n tool.

Further help on using execution-spec-tests with execution-spec-evm [here](https://danceratopz.github.io/execution-spec-tests/main/library/evm_transition_tool/#evm_transition_tool.ExecutionSpecsTransitionTool).

src/ethereum/trace.py
gurukamath: Using `object` created issues because mypy keeps complaining about the existence of certain attributes in `object` type.

src/ethereum/trace.py
gurukamath: Since we are moving everything under a single function `evm_trace`, this change is no longer needed imo.

src/ethereum/trace.py
gurukamath: Updated

tests/conftest.py
gurukamath: I have removed the tracing capability from pytest and maintained it exclusively within the `t8n` tool. That gives us finer control over the tracing within t8n, for example control the output directories and file names without unwieldy workarounds. This is relevant to Dan's comments below regarding the file names.

src/ethereum_spec_tools/evm_trace.py
gurukamath: Done.

src/ethereum_spec_tools/evm_trace.py
gurukamath: Done

src/ethereum_spec_tools/evm_trace.py
gurukamath: Done.

src/ethereum_spec_tools/evm_trace.py
```diff
@@ -1,66 +1,212 @@
...
+
+    last_trace.gasCostTraced = True
+    last_trace.errorTraced = True
+
+
+def capture_op_exception(evm: Any) -> None:
+    """Capture the state in case of exceptions."""
+    if any(evm.env.traces):
+        last_trace = evm.env.traces[-1]
+    if (
```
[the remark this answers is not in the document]
gurukamath: Done.

src/ethereum_spec_tools/evm_trace.py
gurukamath: Updated

src/ethereum_spec_tools/evm_trace.py
```diff
@@ -1,66 +1,212 @@
...
+
+def output_traces(
+    traces: Any, tx_hash: bytes, output_basedir: str = "."
+) -> None:
+    """
+    Output the traces to a json file.
+    """
+    tx_hash_str = "0x" + tx_hash.hex()
+    output_path = os.path.join(
+        output_basedir, f"spec-trace-{tx_hash_str}.json"
```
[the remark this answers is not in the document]
gurukamath: Updated.

gurukamath:
@SamWilsn @petertdavies Have updated the relevant bits based on your comments.

@danceratopz Have updated the tool to reflect specific output file names. Please verify.

src/ethereum/trace.py
SamWilsn: That is expected, because `object` is the most generic type. You'll want to do some runtime checking on the object in `ethereum_spec_tools` to get at the properties.

Something like:

```python
from typing import runtime_checkable, Protocol

@runtime_checkable
class Foo(Protocol):
    property_a: int
    property_b: int

def bar(arg: object) -> int:
    assert isinstance(arg, Foo)
    return arg.property_a
```

Note that using structural typing at runtime (with `Protocol` and `isinstance`) is much slower than normal Python inheritance, so use it judiciously.

SamWilsn (approved):

danceratopz:
@gurukamath Yes, if I apply #823 to this branch, it now works as expected with https://github.com/ethereum/execution-spec-tests/pull/289 as-is, thanks!

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -62,6 +63,7 @@ def t8n_arguments(subparsers: argparse._SubParsersAction) -> None:
         "--state.reward", dest="state_reward", type=int, default=0
     )
     # TODO: Add support for the following trace options
+    t8n_parser.add_argument("--trace", action="store_true")
     t8n_parser.add_argument(
         "--trace.memory", dest="trace_memory", type=bool, default=False
```
danceratopz, line 68: These arguments would then also have `action=store_true`.

danceratopz (commented):
Ah, one last niggle/discrepancy, sorry! I just saw that the following flags are also availalbe:
```
  --trace.memory TRACE_MEMORY
  --trace.nomemory TRACE_NOMEMORY
  --trace.noreturndata TRACE_NORETURNDATA
  --trace.nostack  TRACE_NOSTACK 
  --trace.returndata TRACE_RETURNDATA
```
which require an argument. To make the behavior exactly as geth's `evm t8n` command, these would be switches which would take an argument.

Of course, we can abstract this away in our interface, but I hope you appreciate it's nice to keep these tools as consistent as possible.

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
gurukamath: I can change the way these flags are used. However, I need to point out that this current PR does not implement these functionalities. So I would create a separate issue to track the implementation of these.

danceratopz: Sounds good, they're not currently used in execution-exec-specs, but we'd like to add them at some point. I was playing around with memory tracing (from geth) and it was very helpful.

src/ethereum/trace.py
gurukamath: Updated

gurukamath:
@SamWilsn @petertdavies Have ported the changes over to the older forks.

The main update from the earlier commits in this PR is that the refund calculation for SELFDESTRUCT is moved from the `interpreter` module into the opcode itself (mainly relevant to Berlin and before). This requires making the `accounts_to_delete` from the parent evm available to the child evm.

danceratopz:
Hi @gurukamath, I took a deeper look at the traces generated from the current execution-spec-tests and diff'd the traces generated from geth `1.13.1-stable-3f40e65c` and the `ethereum-spec-evm` from this branch (753416e7a70c43bc00c032e3e58ebd9570b7192c). 

The traces are almost identical :partying_face: and only differ upon an error in execution. They fall into 2 categories.

1. Missing exception message upon error. Here are examples of the 4 missing message types (here they are in [ethereum/go-ethereum/core/vm/errors.go#L24@90d5bd85](https://github.com/ethereum/go-ethereum/blob/90d5bd85bcf2919ac2735a47fde675213348a0a6/core/vm/errors.go#L24), I have no idea if they are standardized):
      - `test_push0_stack_overflow_fork_Shanghai/1/trace-0-0xf8616cc40f214a4729758189c77720a6b672f29dcb62cd206a49ebacaf00e96f.jsonl`
          ```diff
          1028,1029c1028,1029
          < {"pc":1028,"op":95,"gas":"0xd63f","gasCost":"0x2","memSize":0,"stack":["0x0", <full stack ommitted>, "0x0"],"depth":1,"refund":0,"opName":"PUSH0","error":"stack limit reached 1024 (1023)"}
          < {"output":"","gasUsed":"0x13498","error":"stack limit reached 1024 (1023)"}
          ---
          > {"pc":1028,"op":95,"gas":"0xd63f","gasCost":"0x2","memSize":0,"stack":["0x0", <full stack ommitted>, "0x0"],"depth":1,"refund":0,"opName":"PUSH0","error":""}
          > {"output":"","gasUsed":"0x13498","error":""}
          ```
      - `test_create_opcode_initcode_fork_Shanghai_create2_over_limit_ones/1/trace-0-0xff8b41168839e22964df27b1810063b71235dd344b1473025c2c9473f380cd02.jsonl`
          ```diff
          23c23
          < {"pc":17,"op":245,"gas":"0x4c151c","gasCost":"0x7d00","memSize":49184,"stack":["0x4c1527","0xdeadbeef","0xc001","0x0","0x0"],"depth":2,"refund":0,"opName":"CREATE2","error":"out of gas"}
          ---
          > {"pc":17,"op":245,"gas":"0x4c151c","gasCost":"0xad08","memSize":49184,"stack":["0x4c1527","0xdeadbeef","0xc001","0x0","0x0"],"depth":2,"refund":0,"opName":"CREATE2","error":""}
          ```
      - `test_gas_usage_fork_Shanghai_too_little_execution_gas_49121_bytes/1/trace-0-0x1e4296b38ebdc740cd34f5b63602660f115815a352821261416a379a03054832.jsonl`
          ```diff
          8c8
          < {"output":"00","gasUsed":"0xdf","error":"contract creation code storage out of gas"}
          ---
          > {"output":"","gasUsed":"0xdf","error":""}
          ```
      - `test_withdrawing_to_precompiles_fork_Shanghai_precompile_9_amount_1/2/trace-0-0x3e9f556a7efc77edc7e00178f4c3d226633320d06a8289cce10d363f713f50a9.jsonl`
          ```diff
          1c1
          < {"output":"","gasUsed":"0x13498","error":"invalid input length"}
          ---
          > {"output":"","gasUsed":"0x13498","error":""}
          ```
2. Different reported `gasCost` upon an exception caused by initcode that is over the allowed limit when calling `CREATE` or `CREATE2`, the gas cost for the reverted transaction is, however, identical:
    - `test_create_opcode_initcode_fork_Shanghai_create2_over_limit_ones/1/trace-0-0xff8b41168839e22964df27b1810063b71235dd344b1473025c2c9473f380cd02.jsonl`
        ```diff
        23c23
        < {"pc":17,"op":245,"gas":"0x4c151c","gasCost":"0x7d00","memSize":49184,"stack":["0x4c1527","0xdeadbeef","0xc001","0x0","0x0"],"depth":2,"refund":0,"opName":"CREATE2","error":"out of gas"}
        ---
        > {"pc":17,"op":245,"gas":"0x4c151c","gasCost":"0xad08","memSize":49184,"stack":["0x4c1527","0xdeadbeef","0xc001","0x0","0x0"],"depth":2,"refund":0,"opName":"CREATE2","error":""}
        ```

gurukamath:
> 1. Missing exception message upon error. Here are examples of the 4 missing message types (here they are in [ethereum/go-ethereum/core/vm/errors.go#L24@90d5bd85](https://github.com/ethereum/go-ethereum/blob/90d5bd85bcf2919ac2735a47fde675213348a0a6/core/vm/errors.go#L24), I have no idea if they are standardized):

The specs do not currently emit comprehensive error messages. We are working on it and have an active issue ( see #795 )

> 2. Different reported `gasCost` upon an exception caused by initcode that is over the allowed limit when calling `CREATE` or `CREATE2`, the gas cost for the reverted transaction is, however, identical:
>    
>    * `test_create_opcode_initcode_fork_Shanghai_create2_over_limit_ones/1/trace-0-0xff8b41168839e22964df27b1810063b71235dd344b1473025c2c9473f380cd02.jsonl`
>      ```diff
>      23c23
>      < {"pc":17,"op":245,"gas":"0x4c151c","gasCost":"0x7d00","memSize":49184,"stack":["0x4c1527","0xdeadbeef","0xc001","0x0","0x0"],"depth":2,"refund":0,"opName":"CREATE2","error":"out of gas"}
>      ---
>      > {"pc":17,"op":245,"gas":"0x4c151c","gasCost":"0xad08","memSize":49184,"stack":["0x4c1527","0xdeadbeef","0xc001","0x0","0x0"],"depth":2,"refund":0,"opName":"CREATE2","error":""}
>      ```

We are aware that in some cases where the opcode runs out of gas, geth and specs might emit different gas costs. In almost all such cases, I have found that geth emits only the static part of the gas cost. See [this geth issue](https://github.com/ethereum/go-ethereum/issues/28128#issuecomment-1725038209) for more details.

However, I would also like to take a deeper look at this particular case just to make sure that it is a similar case. Would it be possible for you to provide the input json files for running this? Or alternatively, point me to the fixture in the tests repo that has this case?

danceratopz:
Thanks for info and links.

> However, I would also like to take a deeper look at this particular case just to make sure that it is a similar case. Would it be possible for you to provide the input json files for running this? Or alternatively, point me to the fixture in the tests repo that has this

Github wouldn't let me upload json or tgz - I'll send you the files another way. 

In general, you could run this test case from [ethereum/execution-spec-tests](https://github.com/ethereum/execution-spec-tests) in isolation with:
```
fill tests/shanghai/eip3860_initcode/test_initcode.py::TestCreateInitcode::test_create_opcode_initcode[fork=Shanghai-create2-over_limit_ones] --evm-bin=/path/to/execution-specs/venv/bin/ethereum-spec-evm--output=output/
```

gurukamath:
> In general, you could run this test case from [ethereum/execution-spec-tests](https://github.com/ethereum/execution-spec-tests) in isolation with:
> 
> ```
> fill tests/shanghai/eip3860_initcode/test_initcode.py::TestCreateInitcode::test_create_opcode_initcode[fork=Shanghai-create2-over_limit_ones] --evm-bin=/path/to/execution-specs/venv/bin/ethereum-spec-evm--output=output/
> ```

@danceratopz I did look into this case and is similar to the scenario that I pointed out earlier. Geth figures out that the opcode is going to run out of gas due to large init code. For efficiency reasons, it then does not even bother to calculate the dynamic gas because it makes no difference. So it just emits the static part in the trace.

The specs on the other hand calculate all the components of the gas (static + dynamic) and emits that in the trace. I am not sure if there is a standard behaviour that is agreed upon in this case. At least, I couldn't find anything in [EIP-3155](https://eips.ethereum.org/EIPS/eip-3155). For greater context on such issues you can follow the issue on the geth repo that linked earlier. See [here](https://github.com/ethereum/go-ethereum/issues/28128)

petertdavies (approved):
LGTM

PR #843, opened 2023-09-27
fix `python3 -m evm_tools t8n -v` version flag output

winsvega:
### What was wrong?

python3 -m evm_tools t8n -v 
python3 -m evm_tools t8n --version 

commands didn't work

### How was it fixed?

introducing git subcall and message to return when asked -v/--version from python3 -m evm_tools t8n command


winsvega:
@SamWilsn up

SamWilsn (approved):

PR #846, opened 2023-10-06
add error messages to traces

### LINKED ISSUE #795: t8n better exception reporting

gurukamath:
(closes #795 )

### What was wrong?
The error messages in the evm traces are not clear.

Related to Issue #795 

### How was it fixed?
Add error messages based on the exception types.

### DIFF
--- a/src/ethereum/dao_fork/vm/__init__.py
+++ b/src/ethereum/dao_fork/vm/__init__.py
@@ -81,6 +81,7 @@ class Evm:
     output: Bytes
     accounts_to_delete: Set[Address]
     has_erred: bool
+    error: Optional[Exception]
 
 
 def incorporate_child_on_success(evm: Evm, child_evm: Evm) -> None:
--- a/src/ethereum/dao_fork/vm/__init__.py
+++ b/src/ethereum/dao_fork/vm/__init__.py
@@ -80,7 +80,6 @@ class Evm:
     message: Message
     output: Bytes
     accounts_to_delete: Set[Address]
-    has_erred: bool
     error: Optional[Exception]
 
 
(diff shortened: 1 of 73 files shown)


src/ethereum_spec_tools/evm_tools/t8n/evm_trace.py
```diff
@@ -125,7 +147,11 @@ def evm_trace(
     len_memory = len(evm.memory)
 
     return_data = None
-    if trace_return_data and evm.return_data:
+    if (
+        isinstance(evm, EvmWithReturnData)
+        and trace_return_data
+        and evm.return_data
```
petertdavies, line 153: Should empty return data on forks with return data be `"0x"` or `None`? This implements `None`, `"0x"` seems more logical to me.

petertdavies (changes requested):

src/ethereum_spec_tools/evm_tools/t8n/evm_trace.py
gurukamath: Updated

src/ethereum/dao_fork/vm/__init__.py
```diff
@@ -81,6 +81,7 @@ class Evm:
     output: Bytes
     accounts_to_delete: Set[Address]
     has_erred: bool
+    error: Optional[Exception]
```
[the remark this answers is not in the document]
gurukamath: Good point. Have updated.

SamWilsn (approved):

PR #849, opened 2023-10-27
Implement EIP-1153 for Cancun

gurukamath:
### What was wrong?
EIP-1153, which is slated for Cancun, is not currently implemented in the specs.

### How was it fixed?
Implemented EIP-1153


gurukamath:
Closed this since we probably don't want this to be merged directly into `master`. Will open a new PR as per the right naming conventions for the branches.

PR #850, opened 2023-10-27
Implement EIP-1153 for Cancun

gurukamath:
### What was wrong?
EIP-1153, which is slated for Cancun, is not currently implemented in the specs.

### How was it fixed?
Implemented EIP-1153

### DIFF
--- a/src/ethereum/cancun/state.py
+++ b/src/ethereum/cancun/state.py
@@ -46,6 +46,19 @@ class State:
     _created_accounts: Set[Address] = field(default_factory=set)
 
 
+@dataclass
+class TransientStorage:
+    """
+    Contains all information that is preserved between message calls
+    within a transaction.
+    """
+
+    _tries: Dict[Address, Trie[Bytes, U256]] = field(default_factory=dict)
+    _snapshots: List[Dict[Address, Trie[Bytes, U256]]] = field(
+        default_factory=list
+    )
+
+
 def close_state(state: State) -> None:
     """
     Free resources held by the state. Used by optimized implementations to
@@ -57,7 +70,9 @@ def close_state(state: State) -> None:
     del state._created_accounts
 
 
-def begin_transaction(state: State) -> None:
+def begin_transaction(
+    state: State, transient_storage: Optional[TransientStorage] = None
+) -> None:
     """
     Start a state transaction.
 
@@ -68,6 +83,8 @@ def begin_transaction(state: State) -> None:
     ----------
     state : State
         The state.
+    transient_storage : TransientStorage
+        The transient storage of the transaction.
     """
     state._snapshots.append(
         (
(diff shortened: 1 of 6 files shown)


gurukamath:
Rebased the EIP-1153 implementation on the `forks/cancun` branch

petertdavies (changes requested):
I think the basic architecture of this PR is wrong.

Firstly `transient_state` is not a `State`. `State` comes with a lot of extra baggage (disk backing, state root calculating, etc) which doesn't apply to transient storage.

Secondly, transient storage is strictly a transaction scoped rather that block scoped. It is "discarded" at the end of the transaction. It belongs in the `Evm` dataclass with all of the other transaction scoped ephemeral data.

gurukamath:
> I think the basic architecture of this PR is wrong.
> 
> Firstly `transient_state` is not a `State`. `State` comes with a lot of extra baggage (disk backing, state root calculating, etc) which doesn't apply to transient storage.
> 

You are right. Not sure how I missed  that. Will update

> Secondly, transient storage is strictly a transaction scoped rather that block scoped. It is "discarded" at the end of the transaction. It belongs in the `Evm` dataclass with all of the other transaction scoped ephemeral data.

The `Environment` data class is actually not entirely block scoped in the sense that a new `env` is created for each transaction within a block (`env` has to have `caller` and `origin` which vary by transaction). The `Evm` data class on the other hand has the scope of a message and might not be usable for multi-level inter-message call communication as is required by this EIP. I made this choice keeping those points in mind.

gurukamath:
@petertdavies Updated the PR to include the `TransientStorage` class and include an attribute in the transaction scoped `EVM` class

src/ethereum/cancun/state.py
```diff
@@ -603,3 +616,90 @@ def get_storage_original(state: State, address: Address, key: Bytes) -> U256:
     assert isinstance(original_value, U256)
 
     return original_value
+
+
+def backup_transient_storage(transient_storage: TransientStorage) -> None:
```
SamWilsn, line 621: Does this function mark where a rollback will rollback to? Like this takes a snapshot? If so, I think this function's name should be something to do with "begin". It's pretty recognized as how you begin a transaction. We do this ourselves up in `begin_transaction`.

gurukamath: Updated

src/ethereum/cancun/vm/interpreter.py
```diff
@@ -257,6 +261,11 @@ def execute_code(message: Message, env: Environment) -> Evm:
     code = message.code
     valid_jump_destinations = get_valid_jump_destinations(code)
 
+    if message.parent_evm and message.parent_evm.transient_storage:
```
SamWilsn, line 264: When would the parent not have a transient storage? Are there cases where an `Evm` would have a different `TransientStorage` than all of its parents?

Why not put the `TransientStorage` on the `Environment` instead?

gurukamath: Updated the if condition. The `TransientStorage` was moved from the `Environment` to `Evm` data class after [this](https://github.com/ethereum/execution-specs/pull/850#pullrequestreview-1706392390) discussion.

petertdavies (approved):
Looks perfectly reasonable to me. We need to add support to the optimized state so the test suite passes.

PR #856, opened 2023-12-08
Implement EIP-4788 for Cancun

gurukamath:
What was wrong?

EIP-4788, which is slated for Cancun, is not currently implemented in the specs.

How was it fixed?

Implemented EIP-4788


gurukamath:
Rebased the PR on the latest `forks/cancun`

SamWilsn (approved):

PR #858, opened 2023-12-11
Implement EIP-6780 for Cancun

gurukamath:
What was wrong?

EIP-6780, which is slated for Cancun, is not currently implemented in the specs.

How was it fixed?

Implemented EIP-6780


gurukamath:
Already implemented in another PR

Issue #863, opened 2024-01-12
EIP-145 listed in `petersburg` hard fork


leovct:
### Metadata

* **Hardfork**: `petersburg`

### What was wrong?

I think [EIP-145](https://eips.ethereum.org/EIPS/eip-145) was introduced in the `constantinople` hard fork. It should be removed from the list of EIPs of the `petersburg` hard fork.

<img width="731" alt="Screenshot 2024-01-12 at 09 55 57" src="https://github.com/ethereum/execution-specs/assets/28714795/a8a716e0-0e21-45b4-96e5-9651a627181d">

leovct:
My bad, was a mistake :)

PR #864, opened 2024-01-12
doc: remove EIP-145 from Petersburg hard fork

leovct:
### What was wrong?

Related to Issue #863 

### How was it fixed?

Simply remove EIP-145 from the list of Petersburg EIPs.


leovct:
My bad, was a mistake :)

PR #871, opened 2024-01-30
Delete empty account at `BEACON_ROOTS_ADDRESS`

gurukamath:
### What was wrong?
The current implementation of EIP-4788 did not delete the account at `BEACON_ROOTS_ADDRESS` if it is empty after being touched by the system message call. For example, in a case where the contract has not been deployed yet. EIP-4788 does not explicitly mention this deletion but it is implied post EIP-161

### How was it fixed?
Delete the account at `BEACON_ROOTS_ADDRESS` if it is empty.

### DIFF
--- a/src/ethereum/cancun/fork.py
+++ b/src/ethereum/cancun/fork.py
@@ -51,6 +51,7 @@ from .state import (
     State,
     account_exists_and_is_empty,
     destroy_account,
+    destroy_touched_empty_accounts,
     get_account,
     increment_nonce,
     process_withdrawal,
@@ -525,7 +526,14 @@ def apply_body(
         traces=[],
     )
 
-    process_message_call(system_tx_message, system_tx_env)
+    system_tx_output = process_message_call(system_tx_message, system_tx_env)
+
+    destroy_touched_empty_accounts(
+        system_tx_env.state, system_tx_output.touched_accounts
+    )
+
+    if account_exists_and_is_empty(state, BEACON_ROOTS_ADDRESS):
+        destroy_account(state, BEACON_ROOTS_ADDRESS)
 
     for i, tx in enumerate(map(decode_transaction, transactions)):
         trie_set(
@@ -694,9 +702,7 @@ def process_transaction(
     for address in output.accounts_to_delete:
         destroy_account(env.state, address)
 
-    for address in output.touched_accounts:
-        if account_exists_and_is_empty(env.state, address):
-            destroy_account(env.state, address)
+    destroy_touched_empty_accounts(env.state, output.touched_accounts)
 
     return total_gas_used, output.logs, output.error
 
--- a/src/ethereum/cancun/state.py
+++ b/src/ethereum/cancun/state.py
@@ -17,7 +17,7 @@ There is a distinction between an account that does not exist and
 `EMPTY_ACCOUNT`.
 """
 from dataclasses import dataclass, field
-from typing import Callable, Dict, List, Optional, Set, Tuple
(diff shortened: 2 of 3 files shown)


src/ethereum/cancun/fork.py
```diff
@@ -527,6 +527,9 @@ def apply_body(
 
     process_message_call(system_tx_message, system_tx_env)
 
+    if account_exists_and_is_empty(state, BEACON_ROOTS_ADDRESS):
+        destroy_account(state, BEACON_ROOTS_ADDRESS)
+
```
SamWilsn, line 532: Doing this here means we now have two distinct places where we destroy touched accounts. I feel that makes it harder to understand. Does `BEACON_ROOTS_ADDRESS` appear in `process_message_call`'s `touched_accounts`? We could factor out account removal into its own function (shared by system transactions here and by `process_transaction`.)

gurukamath: Updated

src/ethereum/cancun/fork.py
```diff
@@ -525,7 +526,14 @@ def apply_body(
...
 
-    process_message_call(system_tx_message, system_tx_env)
+    system_tx_output = process_message_call(system_tx_message, system_tx_env)
+
+    destroy_touched_empty_accounts(
+        system_tx_env.state, system_tx_output.touched_accounts
+    )
+
+    if account_exists_and_is_empty(state, BEACON_ROOTS_ADDRESS):
+        destroy_account(state, BEACON_ROOTS_ADDRESS)
```
SamWilsn, line 536: Hm, is it necessary to check again after running `destroy_touched_empty_accounts`?

petertdavies:
The question of whether `BEACON_ROOTS_ADDRESS` gets cleared is verifiably moot, because the account is not empty and cannot become empty. My recollection is that it was decided to remove the test from `execution-spec-tests`, rather than force everyone to implement this. If that's the case I don't see why we should be added code to the specs.

src/ethereum/cancun/fork.py
gurukamath: Oh. This is an error. Will update.

gurukamath:
> The question of whether `BEACON_ROOTS_ADDRESS` gets cleared is verifiably moot, because the account is not empty and cannot become empty. My recollection is that it was decided to remove the test from `execution-spec-tests`, rather than force everyone to implement this. If that's the case I don't see why we should be added code to the specs.

The way the `ethereum/tests` for cancun are currently filled, this account can end up being touched and empty. Will check with the team. @marioevz Any comments on this?

marioevz:
Execution spec tests fix: https://github.com/ethereum/execution-spec-tests/pull/425

Still looking into ethereum/tests.

chfast (commented):
This is by design unspecified because this cannot happen in practice.

gurukamath:
Opened [issue](https://github.com/ethereum/tests/issues/1341) for this to be fixed in the `ethereum/tests` repo

chfast:
[EIP-7523](https://eips.ethereum.org/EIPS/eip-7523) fixes this because then the `BEACON_ROOTS_ADDRESS` cannot be empty in the pre-state.

winsvega:
but can it be non existent?

winsvega:
but then there is this option which is also legal - to just write storages without a silent call. it creates double reading. 
because writing storages directly is not a call so it does not fail. 

I think it would be better 

if there is no contract in the state. deploy contract to the state. make silent call / write storages
if contrac alsready in the state verify  it has nonce = 1, code = CODE

PR #872, opened 2024-01-31
Replace gas_available with gas_used in the description of apply_body

fradamt:
### What was wrong?

The first return value in the description of `apply_body`is `gas_available` instead of `gas_used`.

### How was it fixed?

Replacing `gas_available` with `gas_used`.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/arrow_glacier/fork.py
+++ b/src/ethereum/arrow_glacier/fork.py
@@ -533,7 +533,7 @@ def apply_body(
 
     Returns
     -------
-    gas_used : `ethereum.base_types.Uint`
+    block_gas_used : `ethereum.base_types.Uint`
         Gas used for executing all transactions.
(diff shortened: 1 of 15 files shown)


src/ethereum/arrow_glacier/fork.py
```diff
@@ -533,8 +533,8 @@ def apply_body(
 
     Returns
     -------
-    gas_available : `ethereum.base_types.Uint`
-        Remaining gas after all transactions have been executed.
+    gas_used : `ethereum.base_types.Uint`
```
SamWilsn, line 536: Maybe

```suggestion
    block_gas_used : `ethereum.base_types.Uint`
```

to match the actual variable name?

---

If you don't want to change it, I'm fine merging as-is too. We're switching to markdown shortly anyway.

SamWilsn (approved):

Issue #874, opened 2024-02-01
In `fork.py`'s `make_receipt`, multiple `if`/`return` blocks are followed by an `else`


SamWilsn:
For example:

https://github.com/ethereum/execution-specs/blob/cd9b7d6a9af2f5e07cad02a4971744dd6a553b10/src/ethereum/london/fork.py#L484-L489

This is a really weird construct. Either they should all be `elif` or we shouldn't have the `else:`. 

_Originally posted by @SamWilsn in https://github.com/ethereum/execution-specs/pull/867#discussion_r1475032326_

richardgreg:
Could you assign me to this task?

SamWilsn:
@richardgreg feel free to take a stab at it (don't forget to make a separate PR targeting `forks/cancun`). Thanks!

richardgreg:
~(don't forget to make a separate PR targeting `forks/cancun`)~. Thanks!

Because of the above comment, I opened a PR targeting the `cancun` branch. lol. I hope I did the right thing. If not, let me know and I'll open it against the `master` branch 🙂

PR #882, opened 2024-02-09
Refactor make receipt control flow

richardgreg:
### What was wrong?
In `fork.py`'s `make_receipt`, multiple `if/return` blocks are followed by an `else`
Related to Issue #
#874
### How was it fixed?
Simple refactoring. Use `if/elif/else` flow

#### Cute Animal Picture


gurukamath:
@richardgreg This needs to eventually be merged into `master`. So you could point the PR to `master` instead of `cancun`

PR #889, opened 2024-02-21
Fix cancun `t8n`

petertdavies:
### What was wrong?

`t8n` had not been updated for Cancun

### How was it fixed?

Update `t8n`

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #891, opened 2024-02-22
Evm daemon

SamWilsn:


src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -391,6 +397,15 @@ def apply_body(self) -> None:
                 env = self.environment(tx, gas_available)
 
                 process_transaction_return = self.process_transaction(env, tx)
+
+                if self.is_after_fork("ethereum.cancun"):
+                    blob_gas_used += self._module(
+                        "vm.gas"
+                    ).calculate_total_blob_gas(tx)
+                    ensure(
```
SamWilsn, line 405: By @gurukamath in [#889](https://github.com/ethereum/execution-specs/pull/889/files#r1500618491):

> Based on the EIP, this check has to happen once outside the for loop, after all the transactions have been processed. But tbh, this makes more sense imo unless I am missing something.

PR #893, opened 2024-02-22
Cancun fixes

gurukamath:
### What was wrong?

1. Cancun spec does not handle undeployed beacon roots contract properly.
2. `blob_versioned_hashes` are constant and need to be part of the `Environment` class instead of the `Message` class.

### How was it fixed?
1. Delete touched beacon roots contract at the end of the system message call
2. Move `blob_versioned_hashes`

#### Cute Animal Picture

cute dog

### DIFF
--- a/src/ethereum/cancun/fork.py
+++ b/src/ethereum/cancun/fork.py
@@ -53,6 +53,7 @@ from .state import (
     State,
     account_exists_and_is_empty,
     destroy_account,
+    destroy_touched_empty_accounts,
     get_account,
     increment_nonce,
     process_withdrawal,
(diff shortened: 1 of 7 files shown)


src/ethereum/cancun/fork.py
```diff
@@ -547,18 +554,25 @@ def apply_body(
...
         excess_blob_gas=excess_blob_gas,
+        blob_versioned_hashes=(),
     )
 
-    process_message_call(system_tx_message, system_tx_env)
+    system_tx_output = process_message_call(system_tx_message, system_tx_env)
+
+    destroy_touched_empty_accounts(
+        system_tx_env.state, system_tx_output.touched_accounts
+    )
```
SamWilsn, line 564: I think we're going to try and rip out empty accounts entirely today, so hopefully we won't need this at all.

SamWilsn (approved):

src/ethereum/cancun/fork.py
```diff
@@ -749,9 +759,7 @@ def process_transaction(
     for address in output.accounts_to_delete:
         destroy_account(env.state, address)
 
-    for address in output.touched_accounts:
-        if account_exists_and_is_empty(env.state, address):
-            destroy_account(env.state, address)
+    destroy_touched_empty_accounts(env.state, output.touched_accounts)
```
gurukamath, line 762: @SamWilsn Had missed using the re-factored function to destroy the empty accounts. This is the only update I've just made.

PR #894, opened 2024-02-23
Moved Transaction types into their own files (eg. FeeMarketTransaction, LegacyTransaction) as mentioned in the issue assigned to me 

Redidacove:
### The fork_types modules are gonna increase in size in the future so I refactored transactions-related functions and variables in a separate module for all the forks 

Related to Issue #

### I created a separate file named transactions.py  in the same directory as fork_types.py for different forks from arrow-glacier to tangerine-whistle. I moved the transaction-related content into transactions.py with the necessary imports and variables used in it .
###Also, I created different commits for each fork for my accountability for which forks are done so far by me in another branch transaction_types_refactoring. 

#### Cute Animal Picture
Thank you for allowing me to contribute to my dream org.😌

229ad628cbdf49c5a0ae302d2e59bdf4

### DIFF
--- a/src/ethereum/arrow_glacier/fork_types.py
+++ b/src/ethereum/arrow_glacier/fork_types.py
@@ -14,7 +14,7 @@ Types re-used throughout the specification, which are specific to Ethereum.
 
 from dataclasses import dataclass
 from typing import Tuple, Union
-
+from .transactions import LegacyTransaction
 from .. import rlp
 from ..base_types import (
     U64,
@@ -35,106 +35,6 @@ Address = Bytes20
 Root = Hash32
 
 Bloom = Bytes256
-
-TX_BASE_COST = 21000
-TX_DATA_COST_PER_NON_ZERO = 16
-TX_DATA_COST_PER_ZERO = 4
-TX_CREATE_COST = 32000
-TX_ACCESS_LIST_ADDRESS_COST = 2400
-TX_ACCESS_LIST_STORAGE_KEY_COST = 1900
-
-
-@slotted_freezable
-@dataclass
-class LegacyTransaction:
-    """
-    Atomic operation performed on the block chain.
-    """
-
-    nonce: U256
-    gas_price: Uint
-    gas: Uint
-    to: Union[Bytes0, Address]
-    value: U256
-    data: Bytes
-    v: U256
-    r: U256
-    s: U256
-
-
-@slotted_freezable
-@dataclass
-class AccessListTransaction:
-    """
-    The transaction type added in EIP-2930 to support access lists.
-    """
-
-    chain_id: U64
-    nonce: U256
-    gas_price: Uint
-    gas: Uint
-    to: Union[Bytes0, Address]
-    value: U256
-    data: Bytes
-    access_list: Tuple[Tuple[Address, Tuple[Bytes32, ...]], ...]
-    y_parity: U256
-    r: U256
-    s: U256
-
-
-@slotted_freezable
-@dataclass
-class FeeMarketTransaction:
-    """
-    The transaction type added in EIP-1559.
-    """
-
-    chain_id: U64
-    nonce: U256
-    max_priority_fee_per_gas: Uint
-    max_fee_per_gas: Uint
-    gas: Uint
-    to: Union[Bytes0, Address]
-    value: U256
-    data: Bytes
-    access_list: Tuple[Tuple[Address, Tuple[Bytes32, ...]], ...]
-    y_parity: U256
-    r: U256
-    s: U256
-
-
-Transaction = Union[
-    LegacyTransaction, AccessListTransaction, FeeMarketTransaction
-]
-
-
-def encode_transaction(tx: Transaction) -> Union[LegacyTransaction, Bytes]:
-    """
-    Encode a transaction. Needed because non-legacy transactions aren't RLP.
-    """
-    if isinstance(tx, LegacyTransaction):
-        return tx
-    elif isinstance(tx, AccessListTransaction):
-        return b"\x01" + rlp.encode(tx)
-    elif isinstance(tx, FeeMarketTransaction):
-        return b"\x02" + rlp.encode(tx)
-    else:
-        raise Exception(f"Unable to encode transaction of type {type(tx)}")
-
-
-def decode_transaction(tx: Union[LegacyTransaction, Bytes]) -> Transaction:
-    """
-    Decode a transaction. Needed because non-legacy transactions aren't RLP.
-    """
-    if isinstance(tx, Bytes):
-        if tx[0] == 1:
-            return rlp.decode_to(AccessListTransaction, tx[1:])
-        elif tx[0] == 2:
-            return rlp.decode_to(FeeMarketTransaction, tx[1:])
-        else:
-            raise InvalidBlock
-    else:
-        return tx
 
 
 @slotted_freezable
--- a/src/ethereum/arrow_glacier/fork_types.py
+++ b/src/ethereum/arrow_glacier/fork_types.py
@@ -13,8 +13,9 @@ Types re-used throughout the specification, which are specific to Ethereum.
 """
 
 from dataclasses import dataclass
-from typing import Tuple, Union
-from .transactions import LegacyTransaction
+from typing import Tuple, Union, TYPE_CHECKING
+if TYPE_CHECKING:
+    from .transactions import LegacyTransaction
 from .. import rlp
 from ..base_types import (
     U64,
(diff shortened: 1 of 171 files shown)


SamWilsn:
Not sure why GitHub isn't running the checks, but you've got a few problems with your typing/formatting/etc.

[`CONTRIBUTING.md`](https://github.com/ethereum/execution-specs/blob/master/CONTRIBUTING.md) has the full instructions, but to run the checks locally:

```bash
$ python3 -m venv venv
$ . venv/bin/activate
(venv) $ pip install tox
(venv) $ tox -e static
```

And to automatically fix a lot of errors:

```bash
$ python3 -m venv venv
$ . venv/bin/activate
(venv) $ pip install -e .[test,doc,lint]
(venv) $ isort src/
(venv) $ black src/
```

src/ethereum/arrow_glacier/transactions.py
```diff
@@ -0,0 +1,121 @@
...
+    Bytes32,
+    Bytes256,
+    Uint,
+    slotted_freezable,
+)
+from ..crypto.hash import Hash32
+from ..exceptions import InvalidBlock
+
+Address = Bytes20
+Root = Hash32
```
SamWilsn, line 21: These definitions don't really make sense in the `transactions.py` module. I assume you moved them because of a circular dependency problem? My next few comments will be a proposed solution.

SamWilsn, line 21:
```suggestion
```

src/ethereum/arrow_glacier/transactions.py
```diff
@@ -0,0 +1,121 @@
...
+    Bytes0,
+    Bytes8,
+    Bytes20,
+    Bytes32,
+    Bytes256,
+    Uint,
+    slotted_freezable,
+)
+from ..crypto.hash import Hash32
+from ..exceptions import InvalidBlock
```
SamWilsn, line 18:
```suggestion
from ..exceptions import InvalidBlock
from .fork_types import Address, Root
```

src/ethereum/arrow_glacier/fork_types.py
```diff
@@ -14,7 +14,7 @@
 
 from dataclasses import dataclass
 from typing import Tuple, Union
```
SamWilsn, line 16:
```suggestion
from typing import Tuple, Union, TYPE_CHECKING
```

src/ethereum/arrow_glacier/fork_types.py
```diff
@@ -14,7 +14,7 @@
...
-
+from .transactions import LegacyTransaction
```
SamWilsn, line 17:
```suggestion
if TYPE_CHECKING:
    from .transactions import LegacyTransaction
```

SamWilsn (commented):
And finally, I can't comment directly on the line, but change:

```python
transactions: Tuple[Union[Bytes, LegacyTransaction], ...]
```

To:

```python
transactions: Tuple[Union[Bytes, "LegacyTransaction"], ...]
```

src/ethereum/arrow_glacier/fork_types.py
Redidacove: ok will do the above changes and then commit them and say thnks for the help @SamWilsn

Redidacove:
Actually I did changes after forking the repo and then made changes in separate branch ad then raised the PR where i am just confused about the fact that it is mentioned in the contribution.md that we need to clone original repo  with recursive tag.Did I do anything wrong it I do it my way

Redidacove:
Now I have updated as per you said do check it out

SamWilsn:
>  i am just confused about the fact that it is mentioned in the contribution.md that we need to clone original repo with recursive tag.Did I do anything wrong it I do it my way

No your way is fine!

SamWilsn:
I've fixed some of the easier errors and added docstrings. Please actually run the tooling, and fix the problems it reports. For example:

```
tests/berlin/test_rlp.py:5: error: Module "ethereum.berlin.fork_types" has no attribute "encode_transaction"  [attr-defined]
src/ethereum/homestead/fork.py:29: error: Module "ethereum.homestead.fork_types" has no attribute "TX_BASE_COST"  [attr-defined]
```

Redidacove:
How can I run the tooling like you did like what command

Redidacove:
Fixed the errors coming from my change plz review now @SamWilsn and thanks for fixing issues by your commits

Redidacove:
Any updates on this @SamWilsn

SamWilsn:
looking now.

src/ethereum/shanghai/fork_types.py
```diff
@@ -173,18 +74,6 @@ def encode_account(raw_account_data: Account, storage_root: Bytes) -> Bytes:
...
-class Withdrawal:
-    """
-    Withdrawals that have been validated on the consensus layer.
-    """
-
-    index: U64
-    validator_index: U64
-    address: Address
-    amount: U256
-
```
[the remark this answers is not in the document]
Redidacove: I think withdrawal is also a type of transaction and since I moved all the transactiontypes like LegacyTransaction and Transaction into separate module so it made more sense to keep withdrawal also within transaction.
In order to maintain a similar fashion .Thats it :)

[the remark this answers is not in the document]
Redidacove: @SamWilsn  What's your suggestion on this because I am not sure about it. Its just that withdrawal here is different from normal withdrawal transaction that we think of in traditional finance so it would be better in fork_types ig.
if you say I will revert back the change that I did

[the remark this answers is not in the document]
SamWilsn: `Withdrawal` doesn't _seem_ like a transaction, and if there are going to be different types, yeah I agree that it'll probably deserve its own module in the future.

I'll move it back and start the tests again.

SamWilsn:
This is actually a bit more complicated than I thought. Using `"Transaction"` and `"LegacyTransaction"` (note the quotes) annotations breaks our RLP encoder. I have a couple ideas on how to fix it, but I'm not sure what the best option is. Let me get back to you.

SamWilsn:
It's a result of your changes, but not your fault. Python doesn't convert the annotation from a string to a class, so we break somewhere around here:

https://github.com/ethereum/execution-specs/blob/0f9e4345b60d36c23fffaa69f70cf9cdb975f4ba/src/ethereum/rlp.py#L269

I haven't had a chance to think about it yet. Feel free to take a stab at it, but let me know your plan before you write a ton of code.

Redidacove:
> And finally, I can't comment directly on the line, but change:
> 
> ```python
> transactions: Tuple[Union[Bytes, LegacyTransaction], ...]
> ```
> 
> To:
> 
> ```python
> transactions: Tuple[Union[Bytes, "LegacyTransaction"], ...]
> ```

You asked for this change so I did it but even i didnt knew that it would be an issue coz I dont know all the places it is been used and its complications . Nevertheless if you have any solutions do let me know I am also trying to find some myself

Redidacove:
> And finally, I can't comment directly on the line, but change:
> 
> ```python
> transactions: Tuple[Union[Bytes, LegacyTransaction], ...]
> ```
> 
> To:
> 
> ```python
> transactions: Tuple[Union[Bytes, "LegacyTransaction"], ...]
> ```

@SamWilsn any solution for this you have ,Do share I am not able to figure out how to fix this I can only think if we do it as it was previously without TYPE_CHECKING normal transaction and LegacyTransaction without " " (quotes) then it would work i believe.

SamWilsn:
> we do it as it was previously without TYPE_CHECKING normal transaction and LegacyTransaction without " " (quotes) then it would work i believe. What do you think.

Unfortunately that would mean duplicating the definition of a few types like `Address`, right?

I still haven't sat down and tried any solutions yet, but I was thinking of maybe moving the block types into their own file (eg. `src/ethereum/frontier/block.py`.) I think that would resolve the circular dependency while not having to duplicate types.

I can't promise that'll work. I'll see if I can try it Monday.

Redidacove:
okay sir 🫡

Redidacove:
@SamWilsn any updates on this any solutions to the issue

Redidacove:
@SamWilsn Did you got the time to look into the issue yet Ik you are super busy but can you just give some hint to me what solution you think to this can be so that I can do something I did a lot of hardship in this want this to complete hope you understand 😊😇.

SamWilsn:
If you want to investigate it yourself, please feel free!

Best hint I can give right now is try keeping `Address` in `fork_types.py`
and moving the block types to something like `blocks.py`.

The other approach would be fixing the RLP stuff to work with quoted types.

You'll have to try and see which one works.

SamWilsn:
Finally got the time to look at this. I've run the tests locally, so this _should_ pass CI.

Redidacove:
Awesome actually I was also busy studying for my sem exams due to which I am not getting much time for coding and stuffs I am glad you fixed it I was really clueless on this what was the issue that you tweaked would you like to share just curious

gurukamath (approved):
LGTM!

Issue #752, opened 2023-04-24
Use Bytes instead of bytes in logs.


andy-thomason:
### Metadata

* **Hardfork**: all

### What was wrong?

As a tiny pedantic issue, the use of "bytes" instead of "Bytes" rather breaks the spell.

### Source

https://github.com/ethereum/execution-specs/blob/5e8aeb3e5b4ba6d2ff2308b1ab9625ef0c38a4f6/src/ethereum/shanghai/blocks.py#L84

jeevansiddu:
Hii team , If this is still open , would love to contribute and close this one , looks like a great first issue , can I pick this one and close it ?

andy-thomason:
I do love this manifestation of the spec which has been very useful in generating experimental Ethereum implementations.

I was being a little pedantic, but it would be good to close the set.

PR #898, opened 2024-03-03
Consistent type hints in `fork_types.py`

DistributedDoge:
### What was wrong?

Some type annotations in `fork_types.py` for each fork use `bytes` type, and some use `Bytes` alias imported from `..base_types`. 

[fork_types.py](https://github.com/ethereum/execution-specs/blob/master/src/ethereum/spurious_dragon/fork_types.py#L142)

```
class Transaction:
    ...
    data: Bytes

class Account:
    ...
    code: bytes

class Log:
    ...
    data: bytes

```

Issue #752 suggests that spec would be more consistent if `Bytes` alias was applied everywhere to match surrounding code.

This change should not result in any functional difference as `Bytes` is [alias](https://github.com/ethereum/execution-specs/blob/0f9e4345b60d36c23fffaa69f70cf9cdb975f4ba/src/ethereum/base_types.py#L975) for `bytes`.

### How was it fixed?

For every `Log.data` and `Account.code` in `src/ethereum/$fork_name/fork_types.py` replaced type annotation like so: `bytes` => `Bytes`.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/arrow_glacier/blocks.py
+++ b/src/ethereum/arrow_glacier/blocks.py
@@ -72,7 +72,7 @@ class Log:
 
     address: Address
     topics: Tuple[Hash32, ...]
-    data: bytes
+    data: Bytes
 
 
 @slotted_freezable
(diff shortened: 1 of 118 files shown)


SamWilsn (commented):
This looks good! Would you mind rebasing it on the current master?

gurukamath:
@DistributedDoge Would you be able to rebase this on current master.

SamWilsn (approved):

PR #904, opened 2024-03-11
Improve `MessageCallGas` documentation

trocher:
### What was wrong?

- `MessageCallGas.cost`'s documentation stated that it was _"The non-refundable portion of gas reserved for executing the call opcode."_. However, `MessageCallGas.cost` includes the gas to be passed to the sub-call (excluding the `2300` gas stipend) and hence, it is not true that it is non-refundable since some of the gas passed might be refunded upon returning from the sub-call.

- The name `MessageCallGas.stipend` can be misleading as "stipend" is also often used to refer to the `2300` gas stipend appended to calls with a non-null value.

### How was it fixed?

- Updated the documentation of `MessageCallGas.cost` to remove the "non-refundable" mention.
- Renamed `MessageCallGas.stipend` to `MessageCallGas.sub_call`.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/arrow_glacier/vm/gas.py
+++ b/src/ethereum/arrow_glacier/vm/gas.py
@@ -89,7 +89,7 @@ class MessageCallGas:
         The gas required to execute the call opcode.
     `sub_call`: `ethereum.base_types.Uint`
         The portion of gas available to sub-calls that is refundable
-        if not consumed
+        if not consumed.
     """
 
     cost: Uint
--- a/src/ethereum/arrow_glacier/vm/gas.py
+++ b/src/ethereum/arrow_glacier/vm/gas.py
@@ -82,18 +82,18 @@ class ExtendMemory:
 @dataclass
 class MessageCallGas:
     """
-    Define the gas cost and stipend for executing the call opcodes.
+    Define the gas cost and gas given to the sub-call for
+    executing the call opcodes.
 
     `cost`: `ethereum.base_types.Uint`
-        The non-refundable portion of gas reserved for executing the
-        call opcode.
-    `stipend`: `ethereum.base_types.Uint`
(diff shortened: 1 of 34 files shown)


src/ethereum/arrow_glacier/vm/gas.py
```diff
@@ -81,18 +81,18 @@ class ExtendMemory:
...
-        call opcode.
-    `stipend`: `ethereum.base_types.Uint`
+        The gas required to execute the call opcode.
+    `sub_call`: `ethereum.base_types.Uint`
         The portion of gas available to sub-calls that is refundable
-        if not consumed
+        if not consumed.
     """
 
     cost: Uint
```
SamWilsn, line 94: Would it make sense to also rename `cost` to `total` or something similar? Looks like it excludes the memory expansion cost though, and that might be a tad misleading.

SamWilsn (commented):
Your analysis seems sound; `cost` must contain the total cost of the instruction. See:

https://github.com/ethereum/execution-specs/blob/32992adebf6999ab5a3d06745d9fae9f25f64d9d/src/ethereum/arrow_glacier/vm/instructions/system.py#L354

src/ethereum/arrow_glacier/vm/gas.py
trocher: - Added in a comment that it excludes memory expansion to be explicit.
- Keeping `cost` in the variable name is good imo since it is added to some other `cost` afterwards, maybe `total_cost` would work? 

https://github.com/ethereum/execution-specs/blob/6c337f7c579e38170813d10cf4685afe0f04c869/src/ethereum/arrow_glacier/vm/instructions/system.py#L353

SamWilsn (approved):

PR #905, opened 2024-03-12
Update README for Cancun blog and Prague fork

ensi321:
- Add Cancun block number and date
- Add blog url for Cancun announcement
- Add TBDs for Prague fork

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ This repository contains the specifications related to the Ethereum execution cl
 
 | Version and Code Name | Block No. | Released | Incl EIPs | Specs | Blog |
 |-----------------------|-----------|----------|-----------|-------|-------|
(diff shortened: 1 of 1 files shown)


cryptohazard:
I was going to make that PR because I need it as reference for work but saw it was already made. If this could be merge quickly that would be awesome.

@ensi321 you are missing 3 EIP in the list for Cancun:

- [EIP-7044](https://eips.ethereum.org/EIPS/eip-7044): Perpetually Valid Signed Voluntary Exits
- [EIP-7045](https://eips.ethereum.org/EIPS/eip-7045): Increase Max Attestation Inclusion Slot
- [EIP-7514](https://eips.ethereum.org/EIPS/eip-7514): Add Max Epoch Churn Limit

PR #906, opened 2024-03-12
Evm tools tests

SamWilsn:


tests/evm_tools/test_t8n.py
```diff
@@ -54,7 +54,7 @@ def t8n_tool_test(test_case: Dict) -> None:
         t8n_tool = T8N(options, sys.stdout, sys.stdin)
         t8n_tool.apply_body()
     except Exception as e:
-        raise FatalException(e)
+        raise FatalException(e) from e
```
gurukamath, line 57: In my opinion, we could even do away with the `tests/evm_tools/test_t8n.py` cause there is now a more comprehensive set of tests for the `t8n` tool. We now run the state tests using the `t8n` for each fork.
The `tests/evm_tools/test_t8n.py` was created as a stop-gap arrangement back when we were first building the evm tools, since we didn't yet have the state test pipeline up and running.

gurukamath (approved):

PR #907, opened 2024-03-18
Support `ethereum-spec-evm statetest`

SamWilsn:
### What was wrong?

We don't have fuzzy eels.

### How was it fixed?

Implemented the `statetest` command for `ethereum-spec-evm`.

#### Cute Animal Picture

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,6 +16,7 @@ packages =
     ethereum_spec_tools/evm_tools
     ethereum_spec_tools/evm_tools/t8n
     ethereum_spec_tools/evm_tools/b11r
+    ethereum_spec_tools/evm_tools/statetest
     ethereum_spec_tools/evm_tools/loaders
     ethereum_spec_tools/lint
     ethereum_spec_tools/lint/lints
--- a/src/ethereum/arrow_glacier/vm/exceptions.py
+++ b/src/ethereum/arrow_glacier/vm/exceptions.py
@@ -63,7 +63,11 @@ class InvalidOpcode(ExceptionalHalt):
     Raised when an invalid opcode is encountered.
     """
 
-    pass
+    code: int
+
+    def __init__(self, code: int) -> None:
+        super().__init__(code)
+        self.code = code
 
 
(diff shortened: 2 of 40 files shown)


SamWilsn:
Note to self: comment on https://github.com/holiman/goevmlab/pull/131 when this is merged.

holiman:
In https://github.com/ethereum/execution-specs/pull/907/commits/88d195e7392119ce54e174740617c83edc62fc8e, you made it output the result to stderr even if we're not tracing. Was that in response to my comment here: https://github.com/holiman/goevmlab/pull/131#issuecomment-2009148751 ? 

> When in "skip trace" mode, (a.k.a speedMode), we still read from `stderr`. But goevmlab does not supply the `--json` switch, so the `{ "stateroot": ...` is not output on `stderr`. There is json output on `stdout`, but we're not looking at it, though. 
> 
> So the `copyUntilEnd` just doesn't get anything. The easiest fix is if you make it always output the stateroot on stderr, which is how I fixed geth here https://github.com/ethereum/go-ethereum/pull/29290 

If so, while it does improve things, it's still a bit wonky. Because you output the full `json` struct on stderr. So, in order to read it, goevmlab needs to know not to expect `jsonl`, but to instead expect multiline output. Which means either parsing json or just trying to parse out the stateroot from the blob. This is what I do for nethermind: https://github.com/holiman/goevmlab/blob/master/evms/nethermind.go#L143 

A much (from my PoV) simpler solution is if eels instead just emits `jsonl`, that is: 
```
{ "stateRoot": "0x..." } 
```
So, my wishlist would be: same output as `--json`, _but_ just do not generate the per-opcode lines of output.

SamWilsn:
@holiman I totally misunderstood what you meant :sweat_smile: This makes a lot more sense.

holiman:
If you rebase this  on top of https://github.com/ethereum/execution-specs/pull/912, that'd be great

holiman (approved):
LGTM :partying_face:

gurukamath (approved):

Issue #899, opened 2024-03-04
Update `optimized` implementation for Cancun


gurukamath:
The `optimized` implementation has not been updated to account for Cancun.

SamWilsn:
Is this good to close now that #919 is merged?

gurukamath:
Yes

PR #908, opened 2024-03-20
Fix memory bloat in CI

gurukamath:
### What was wrong?
The DeprecationWarnings in the Hardfork.discover method were causing significant memory bloat resulting in a crash in the CI.

### How was it fixed?
Migrating to the latest APIs for loading the modules removes these warnings.

#### Cute Animal Picture

Library - 1 of 1


SamWilsn (approved):

gurukamath:
Temporarily disabled the optimised tests in CI as we wait for the updated optimised implementation for cancun. See [issue](https://github.com/ethereum/execution-specs/issues/899)

PR #918, opened 2024-04-01
create transient storage snapshots for `process_create_message`

gurukamath:
(closes #917 )

### What was wrong?
`process_create_message` did not create the right snapshots for rollback in case something went wrong.

Related to Issue #917 

### How was it fixed?
Add `transient_storage` to `Environment` and create appropriate snapshots.

#### Cute Animal Picture

Cute Animals - 1 of 1

PR #919, opened 2024-04-02
Support transient storage in optimized state

petertdavies:
### What was wrong?

The optimized state didn't work with Cancun.

### How was it fixed?

Add Cancun support to optimized state.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_optimized/state_db.py
+++ b/src/ethereum_optimized/state_db.py
@@ -241,8 +241,7 @@ def get_optimized_state_patches(fork: str) -> Dict[str, Any]:
             """
             _begin_transaction(state)
 
-    @add_item(patches)
-    def commit_transaction(state: State) -> None:
+    def _commit_transaction(state: State) -> None:
         """
         See `state`.
         """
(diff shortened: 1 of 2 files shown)


gurukamath:
Pushed a commit to re-enable the optimised tests

src/ethereum/cancun/state.py
```diff
@@ -98,19 +98,26 @@ def begin_transaction(
         )
 
 
-def commit_transaction(state: State) -> None:
+def commit_transaction(
+    state: State, transient_storage: Optional[TransientStorage] = None
```
SamWilsn, line 102: Does this need to be reworked after #918?

src/ethereum_optimized/state_db.py
```diff
@@ -250,6 +301,30 @@ def rollback_transaction(state: State) -> None:
         if not state.tx_restore_points:
             state.created_accounts.clear()
 
+    if has_transient_storage:
+
+        @add_item(patches)
+        def rollback_transaction(
+            state: State,
+            transient_storage: Optional[Any] = None,
+        ) -> None:
```
SamWilsn, line 310: The original function has a signature of:

https://github.com/ethereum/execution-specs/blob/11bb742f063335d1c94878356d02bf49cafb08d4/src/ethereum/cancun/state.py#L121-L123

Do we still need the `Optional` in the patch?

petertdavies: No, but it's harmless.

SamWilsn (approved):

Issue #938, opened 2024-04-23
Implement EIP-2935


gurukamath:
https://eips.ethereum.org/EIPS/eip-2935

gurukamath:
Closed in #941

PR #941, opened 2024-04-29
Implement EIP-2935

gurukamath:
### What was wrong?
EIP-2935 is not currently implemented.

Related to Issue #938 

### How was it fixed?
Implemented the EIP for Prague.

#### Cute Animal Picture

Cute Animals - 1 of 1


Bbjj88h (approved):

gurukamath:
Re-based on `forks/prague`

src/ethereum/prague/fork.py
```diff
@@ -113,44 +118,56 @@ def apply_fork(old: BlockChain) -> BlockChain:
...
-
-    # We are computing the hash only for the most recent block and not for
-    # the rest of the blocks as they have successors which have the hash of
-    # the current block as parent hash.
-    most_recent_block_hash = keccak256(rlp.encode(recent_blocks[-1].header))
-    recent_block_hashes.append(most_recent_block_hash)
+    # This function should not be called on pre-fork block
+    # since it determines whether the current block is the fork block
+    # by looking at the parent block's timestamp.
+    # The following ensures that the block is not a pre-fork block.
```
SamWilsn, line 136:
```suggestion
    # Calling this function on blocks from previous forks is a bug
    # since it determines whether `block` is the fork block
    # by looking at the parent block's timestamp.
    #
    # The following ensures that `block` belongs to this fork.
```

Played around with the wording a bit. Take it or leave it :P

src/ethereum/prague/fork.py
```diff
@@ -113,44 +118,56 @@ def apply_fork(old: BlockChain) -> BlockChain:
...
+    assert FORK_CRITERIA.check(
+        block.header.number,
+        block.header.timestamp,
+    )
 
-    return recent_block_hashes
+    if not FORK_CRITERIA.check(
```
SamWilsn, line 142: Can you remove the `not` and flip the `if`/`else` bodies?

gurukamath: Done

src/ethereum/prague/fork.py
gurukamath: Done

src/ethereum/prague/fork.py
```diff
@@ -113,44 +118,57 @@ def apply_fork(old: BlockChain) -> BlockChain:
...
+            set_storage(
+                chain.state,
+                HISTORY_STORAGE_ADDRESS,
+                (
+                    (current_block.header.number - 1) % HISTORY_SERVE_WINDOW
+                ).to_be_bytes32(),
+                U256.from_be_bytes(current_block.header.parent_hash),
+            )
+            # Set the earlier block as the current block
+            current_block = chain.blocks[-1 - i]
```
SamWilsn, line 171: Wait, shouldn't this just be in the `apply_fork` function? Sorry I didn't catch that earlier!

If it can't be in the `apply_fork` function, we should add a comment explaining why.

gurukamath: Updated

src/ethereum/prague/fork.py
```diff
@@ -104,53 +109,32 @@ def apply_fork(old: BlockChain) -> BlockChain:
...
-        recent_block_hashes.append(prev_block_hash)
-
-    # We are computing the hash only for the most recent block and not for
-    # the rest of the blocks as they have successors which have the hash of
-    # the current block as parent hash.
-    most_recent_block_hash = keccak256(rlp.encode(recent_blocks[-1].header))
-    recent_block_hashes.append(most_recent_block_hash)
+    # At fork block, store the hashes for earlier blocks.
+    current_block = block
+    for i in range(HISTORY_SERVE_WINDOW - 1):
```
petertdavies, line 122: Currently `old.blocks` is only 256 blocks long. So this function will fail if we run the transition on a real chain.

I don't think this is a blocker for this PR, but we should open an issue.

petertdavies (approved):

PR #953, opened 2024-05-16
Make daemon timeout configurable

petertdavies:
### What was wrong?

The t8n daemon timeout can't be disabled.

### How was it fixed?

Add a `--timeout` option to the daemon.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/daemon.py
+++ b/src/ethereum_spec_tools/evm_tools/daemon.py
@@ -64,12 +64,14 @@ class _EvmToolHandler(BaseHTTPRequestHandler):
 
 
 class _UnixSocketHttpServer(socketserver.UnixStreamServer):
-    last_response: Optional[float] = None
+    last_response: float
     shutdown_timeout: int
 
     def __init__(self, *args: Any, shutdown_timeout: int, **kwargs: Any) -> None:
         self.shutdown_timeout = shutdown_timeout
-        self.last_response = time.monotonic()
+        # Add a 60-second allowance to prevent server from timing out during
+        # startup
+        self.last_response = time.monotonic() + 60.0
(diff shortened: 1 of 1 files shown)


src/ethereum_spec_tools/evm_tools/daemon.py
```diff
@@ -73,15 +85,16 @@ def finish_request(
...
+        if self.shutdown_timeout != 0:
+            while True:
+                time.sleep(11.0)
+                now = time.monotonic()
+                last_response = self.last_response
+                if last_response is None:
+                    self.last_response = now
+                elif now - last_response > float(self.shutdown_timeout):
+                    self.shutdown()
+                    break
```
SamWilsn, line 97:
```suggestion
        while self.shutdown_timeout != 0:
            time.sleep(11.0)
            now = time.monotonic()
            last_response = self.last_response
            if last_response is None:
                self.last_response = now
            elif now - last_response > float(self.shutdown_timeout):
                self.shutdown()
                break
```

src/ethereum_spec_tools/evm_tools/daemon.py
```diff
@@ -59,6 +65,12 @@ def do_POST(self) -> None:
 
 class _UnixSocketHttpServer(socketserver.UnixStreamServer):
     last_response: Optional[float] = None
+    shutdown_timeout: int
+
+    def __init__(self, *args: Any, shutdown_timeout: int, **kwargs: Any) -> None:
+        self.shutdown_timeout = shutdown_timeout
+        self.last_response = time.monotonic()
```
SamWilsn, line 72: Won't this make the first check timeout if there is too much of a delay between creating the instance and starting the thread?

SamWilsn (approved):

PR #960, opened 2024-06-06
enhancement: factor out a type from access lists

### LINKED ISSUE #947: Factor out a type from access lists
For example:

https://github.com/ethereum/execution-specs/blob/0f116b58173d3871c98dabec0163b216b463288e/src/ethereum/cancun/transactions.py#L62

Because it's repeated in several different transaction types, the `Tuple[Address, Tuple[Bytes32, ...]]` portion should be promoted to its own type. Something like:

```python
@slotted_freezable
@dataclass
class Access:
    account: Address
    slots: Tuple[Bytes32, ...]
```

richardgreg:
### What was wrong?
Because it's repeated in several different transaction types, the `Tuple[Address, Tuple[Bytes32, ...]]` portion should be promoted to its own type.

Related to Issue #
Fixes #947

### How was it fixed?
This PR refactors the access list type used in various transaction classes to reduce redundancy and improve maintainability. Specifically, it introduces a new `Access` class to represent the `access_list` structure, which is used in multiple transaction types.

1. Created a new Access class to encapsulate the account and slots structure.
```python
@slotted_freezable
@dataclass
class Access:
    account: Address
    slots: Tuple[Bytes32, ...]
```

2. Updated Transaction Classes

3. RLP Encoding and Decoding Adjustments:
* Introduced helper functions encode_access_list and decode_access_list to handle the conversion between the new Access type and the previous tuple structure for RLP encoding and decoding.
* Updated encode_transaction and decode_transaction functions to use these helpers.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/cancun/transactions.py
+++ b/src/ethereum/cancun/transactions.py
@@ -19,6 +19,13 @@ from ..base_types import (
 from ..exceptions import InvalidBlock
 from .fork_types import Address, VersionedHash
 
+TX_BASE_COST = 21000
+TX_DATA_COST_PER_NON_ZERO = 16
+TX_DATA_COST_PER_ZERO = 4
+TX_CREATE_COST = 32000
+TX_ACCESS_LIST_ADDRESS_COST = 2400
+TX_ACCESS_LIST_STORAGE_KEY_COST = 1900
+
 
 @slotted_freezable
 @dataclass


SamWilsn:
This looks like pretty much exactly what I had in mind!

SamWilsn (commented):
I've been playing around with this PR a bit, and it's looking great!

I think it should be rebased on top of #968. That branch can handle nested dataclasses better than `master`, and you shouldn't need the `encode_access_list` and `decode_access_list` helpers there.

That said, there's going to be a bit more work involved in this refactor. You'll need to use the `Access` type in tests near, for example, here:

https://github.com/ethereum/execution-specs/blob/c854868f4abf2ab0c3e8790d4c40607e0d251147/tests/shanghai/test_rlp.py#L62

There are a few other required changes, like dealing with unpacking here:

https://github.com/ethereum/execution-specs/blob/c854868f4abf2ab0c3e8790d4c40607e0d251147/src/ethereum/shanghai/fork.py#L620-L624

mypy is your best friend when doing refactoring. The easiest way to run it in this project is:

```bash
$ tox -e static
```

src/ethereum/cancun/transactions.py
```diff
@@ -1,8 +1,3 @@
-"""
-Transactions are atomic units of work created externally to Ethereum and
-submitted to be executed. If Ethereum is viewed as a state machine,
-transactions are the events that move between states.
-"""
```
SamWilsn, line 5: Why remove the docstrings?

SamWilsn:
Took the liberty of rebasing and running `isort`/`black`. You'll need to force pull/reset your local branch.

src/ethereum/cancun/transactions.py
```diff
@@ -19,12 +19,12 @@
 from ..exceptions import InvalidBlock
 from .fork_types import Address, VersionedHash
 
-TX_BASE_COST = 21000
-TX_DATA_COST_PER_NON_ZERO = 16
-TX_DATA_COST_PER_ZERO = 4
-TX_CREATE_COST = 32000
-TX_ACCESS_LIST_ADDRESS_COST = 2400
-TX_ACCESS_LIST_STORAGE_KEY_COST = 1900
```
SamWilsn, line 27: Was removing these intentional?

SamWilsn:
> Running `tox -e static` breaks due to some issue with rust on my system that I'm still trying to fix

What seems to be the problem?



> I also noticed running `mypy src tests setup.py --exclude "tests/fixtures/*" --namespace-packages` returns many errors. Should that be an issue on its own?

Does it return errors on an unmodified `master` checkout, or just on your branch? `master` passes mypy for me without any errors. If both branches are erroring, it's likely that you're missing some packages in your install. Inside your virtual environment, try `pip install -e .[lint,test]`.

PR #985, opened 2024-08-13
Fix platformdirs dependency

petertdavies:
### What was wrong?

The `platformdirs` dependency is required by `ethereum-spec-evm daemon`, but is only depended on by the `test` extra. 

### How was it fixed?

Move `platformdirs` dependency to `install-requires`.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_optimized/fork.py
+++ b/src/ethereum_optimized/fork.py
@@ -25,10 +25,10 @@ try:
     import ethash
 except ImportError as e:
     # Add a message, but keep it an ImportError.
-    raise e from Exception(
+    raise Exception(
         "Install with `pip install 'ethereum[optimized]'` to enable this "
         "package"
-    )
+    ) from e
 
 Header_ = Any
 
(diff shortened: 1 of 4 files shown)


SamWilsn:
`platformdirs` is not required to use the specs as a library, so I think it should remain optional. Perhaps we do something like this instead:

https://github.com/ethereum/execution-specs/blob/fcd12750edd4443a91f138728689a1d0a503a7c1/src/ethereum_optimized/state_db.py#L20-L27

src/ethereum_optimized/fork.py
```diff
@@ -25,10 +25,10 @@
     import ethash
 except ImportError as e:
     # Add a message, but keep it an ImportError.
-    raise e from Exception(
+    raise Exception(
         "Install with `pip install 'ethereum[optimized]'` to enable this "
         "package"
-    )
+    ) from e
```
SamWilsn, line 31: Now this doesn't match the comment. The weird `raise e from Exception(...)` pattern was to keep the thrown exception an `ImportError`, while adding a message. With your change, it is no longer an `ImportError`.

petertdavies: I've reverted the `raise .. from` commit.

SamWilsn (approved):

Issue #948, opened 2024-05-13
Implement optimised for Prague and re-enable testing


gurukamath:
The optimised testing for Prague has been disabled in #946 

Optimised will have to be implemented and the testing has to be re-enabled post Prague
```[tasklist]
### Tasks
- [ ] Implement `has_storage` introduced in EIP-7610 - #999
```

PR #999, opened 2024-09-07
Implement EIP-7610

gurukamath:
### What was wrong?
EIP-7610 is not implemented currently.

### How was it fixed?
Implement EIP-7610

#### Cute Animal Picture

Cute Animals - 1 of 1

### DIFF
--- a/src/ethereum/berlin/state.py
+++ b/src/ethereum/berlin/state.py
@@ -362,11 +362,30 @@ def account_has_code_or_nonce(state: State, address: Address) -> bool:
     Returns
     -------
     has_code_or_nonce : `bool`
-        True if if an account has non zero nonce or non empty code,
+        True if the account has non zero nonce or non empty code,
         False otherwise.
     """
     account = get_account(state, address)
     return account.nonce != Uint(0) or account.code != b""
+
+
+def account_has_storage(state: State, address: Address) -> bool:
+    """
+    Checks if an account has storage.
+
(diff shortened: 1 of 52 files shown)


src/ethereum/berlin/state.py
```diff
@@ -361,10 +364,12 @@ def account_has_code_or_nonce(state: State, address: Address) -> bool:
...
     -------
-    has_code_or_nonce : `bool`
-        True if if an account has non zero nonce or non empty code,
-        False otherwise.
+    has_code_or_nonce_or_storage : `bool`
+        True if if an account has non zero nonce or non empty code or
+        non empty storage, False otherwise.
     """
+    if address in state._storage_tries:
+        return True
```
SamWilsn, line 372: Would it make sense to introduce a `has_storage(state: State, address: Address)` function?

Either way, I think we need to modify the `ethereum_optimized` monkey patches, since that implementation doesn't have `_storage_tries`.

Adding and patching a `has_storage` function (instead of patching `account_has_code_or_nonce_or_storage`) would probably mean less code in `ethereum_optimized`, which is usually good.

gurukamath: Have updated to introduced the `has_storage` function. Haven't implemented any updates to the optimized implementation but made a note in the post-prague re-factoring issue #948

SamWilsn (approved):

PR #1000, opened 2024-09-09
Implement EIP-7702

### LINKED ISSUE #987: Implement EIP-7702
https://eips.ethereum.org/EIPS/eip-7702

gurukamath:
(closes #987 )

### What was wrong?

The proposal for setting code to EOA's which i currently CFI'd for Prague is not currently supported

Related to Issue #987 

### How was it fixed?
Implement EIP-7702

#### Cute Animal Picture

Cute Animals - 1 of 1 (2)

### DIFF
--- a/src/ethereum/prague/vm/eoa_delegation.py
+++ b/src/ethereum/prague/vm/eoa_delegation.py
@@ -1,0 +1,182 @@
+"""
+Set EOA account code.
+"""
+
+
+from typing import Tuple
+
+from ethereum import rlp
+from ethereum.base_types import U64, U256, Bytes, Uint
+from ethereum.crypto.elliptic_curve import SECP256K1N, secp256k1_recover
+from ethereum.crypto.hash import keccak256
+from ethereum.exceptions import InvalidBlock
+
+from ..fork_types import Address, Authorization
+from ..state import account_exists, get_account, increment_nonce, set_code
+from ..vm.gas import GAS_COLD_ACCOUNT_ACCESS, GAS_WARM_ACCESS
+from . import Environment, Evm, Message
+from .exceptions import InvalidAuthorization
+
+SET_CODE_TX_MAGIC = b"\x05"
+EOA_DELEGATION_MARKER = b"\xEF\x01\x00"
+EOA_DELEGATION_MARKER_LENGTH = len(EOA_DELEGATION_MARKER)
+EOA_DELEGATED_CODE_LENGTH = 23
+PER_EMPTY_ACCOUNT_COST = 25000
+PER_AUTH_BASE_COST = 2500
+
+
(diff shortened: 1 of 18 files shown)


src/ethereum/prague/vm/eoa_delegation.py
```diff
@@ -0,0 +1,182 @@
...
+    Returns
+    -------
+    authority : `Address`
+        The recovered authority address.
+    """
+    y_parity, r, s = authorization.y_parity, authorization.r, authorization.s
+    if 0 >= r or r >= SECP256K1N:
+        raise InvalidAuthorization
+    if 0 >= s or s > SECP256K1N // 2:
+        raise InvalidAuthorization
```
petertdavies, line 69: As it currently stands failing the check on `s` invalidates the entire transaction. It should be in `check_transaction()`. Surprisingly, the same doesn't apply to the check on `r` (this is probably an oversight by the EIP author).

There is [a PR](https://github.com/ethereum/EIPs/pull/8845) in the EIPS repo to change the behaviour to what you have implemented, but it is not merged.

src/ethereum/prague/vm/eoa_delegation.py
```diff
@@ -0,0 +1,182 @@
...
+                authorization.chain_id,
+                authorization.address,
+                authorization.nonce,
+            )
+        )
+    )
+
+    try:
+        public_key = secp256k1_recover(r, s, y_parity, signing_hash)
+    except Exception as e:
```
petertdavies, line 84: Swallowing all exceptions like this is bad practice. You should validate the inputs to `secp256k1_recover` and then remove the `try .. except` block.

I think the only check you are missing is that `y_parity` must be `0` or `1`.

petertdavies (changes requested):

src/ethereum/prague/vm/eoa_delegation.py
gurukamath: I was thinking about that but there seemed to be instances where the checks on `y_parity`, `r`, `s` pass but the `secp256k1_recover` still fails. See [test vector](https://github.com/ethereum/execution-spec-tests/blob/061e41cce851e5d4d7de1d0612c84349f78b492e/tests/prague/eip7702_set_code_tx/test_set_code_txs.py#L2123)

src/ethereum/prague/vm/eoa_delegation.py
petertdavies: ACDE decided to adopt the aforementioned PR so this code is now correct!

src/ethereum/prague/vm/eoa_delegation.py
gurukamath: Added additional checks to authority recovery.

petertdavies (approved):

PR #1003, opened 2024-09-09
Merge improvements in `master` branch into `forks/prague`

petertdavies:
### What was wrong?

Fixes in `master` would be useful in `forks/prague`

### How was it fixed?

Merge them in.

#### Cute Animal Picture


SamWilsn (approved):
I usually prefer linear history, but we can do this too.

PR #1011, opened 2024-09-19
Add checks to signature recovery

gurukamath:
### What was wrong?
The current sender recovery does not check if the `r` value in the signature is a valid one.

### How was it fixed?
Add a additional check to the `secp256k1_recover` function.

### DIFF
--- a/src/ethereum/crypto/elliptic_curve.py
+++ b/src/ethereum/crypto/elliptic_curve.py
@@ -8,9 +8,12 @@ from typing import Generic, Type, TypeVar
 import coincurve
 
 from ..base_types import U256, Bytes
+from . import InvalidSignature
 from .finite_field import Field
 from .hash import Hash32
 
+SECP256K1B = 7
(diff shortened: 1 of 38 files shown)


src/ethereum/crypto/elliptic_curve.py
```diff
@@ -37,16 +40,31 @@ def secp256k1_recover(r: U256, s: U256, v: U256, msg_hash: Hash32) -> Bytes:
     public_key : `ethereum.base_types.Bytes`
         Recovered public key.
     """
+    is_square = pow(
+        pow(r, 3, SECP256K1P) + SECP256K1B, (SECP256K1P - 1) // 2, SECP256K1P
+    )
+
+    if is_square != 1:
+        raise InvalidSignature("r value is not a square")
```
petertdavies, line 48: This is wrong, it should say "r is not the x-coordinate of point on the secp256k1 curve" or similar.

gurukamath: Updated

petertdavies (approved):

PR #1015, opened 2024-09-30
Use correct field names for validator requests

etan-status:
The request fields should be amed `pubkey` instead of `public_key` to be compatible with `execution-spec-tests` and latest EIP specs.

### DIFF
--- a/src/ethereum/prague/requests.py
+++ b/src/ethereum/prague/requests.py
@@ -43,7 +43,7 @@ class DepositRequest:
     [EIP-6110]: https://eips.ethereum.org/EIPS/eip-6110
     """
 
-    public_key: Bytes48
+    pubkey: Bytes48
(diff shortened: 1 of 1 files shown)


gurukamath:
@etan-status As per latest updates under discussion for EIP-7685, I don't think the field name updates will be required since we just make the requests opaque. See [EIP-7002 Update](https://github.com/ethereum/EIPs/pull/8855), [EIP-7251 Update](https://github.com/ethereum/EIPs/pull/8857) and [EIP-6110 Update](https://github.com/ethereum/EIPs/pull/8856)

etan-status:
JSON-RPC still has the structure, no?

etan-status:
But sure, that's more of a thing for the tests repo which already uses correct names for engine API (hive tests).

PR #1018, opened 2024-10-02
Fix T8N in London

marioevz:
### What was wrong?

The T8N daemon was failing when spawned on London fork due to an extra parameter that was appended (`t8n.fork_block == self.block_number`), which seems to have been removed in https://github.com/ethereum/execution-specs/commit/8d617fbb424ded118fd29e3fb13d31242d8c38c4.

### How was it fixed?
Removing the lines that added the parameter seems to have done the trick.

#### Cute Animal Picture


SamWilsn (approved):
Yep, this makes sense. Sorry for breaking it!

Issue #1021, opened 2024-10-08
EELS t8n cut exception string in server mod


winsvega:
```

{'body': '0xf84df84b800a82cf0780808025a00372ba5e45d0a0e636f7e32ffca16f97a2f3e4e930b6584b7b114ab3aa191b4fa04afdfcd2bc179e25c7aa74cb991a08f0ae6c01541a36f700be25c6f4c39bc098', 'alloc': {'0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b': {'balance': '0x3635c9adc5dea00000'}}, 'result': {'stateRoot': '0x70c42824108fafccadbfce71e6e22660c4fad89be18be324cd15ef351969a8c8', 'txRoot': '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421', 'receiptsRoot': '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421', 'withdrawalsRoot': '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421', 'logsHash': '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347', 'logsBloom': '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', 'gasUsed': '0x0', 'currentDifficulty': None, 'currentBaseFee': '0x7', 'rejected': [{'index': 0, 'error': 'Failed transaction: '}], 'receipts': []}}

```

message Failed transaction:   is not sufficient enough to match exception code

the exception string is empty in line
[src/ethereum_spec_tools/evm_tools/t8n/__init__.py](https://github.com/ethereum/execution-specs/blob/8d617fbb424ded118fd29e3fb13d31242d8c38c4/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L388)

this probably goes to EELS repo
https://github.com/ethereum/execution-spec-tests/issues/881

SamWilsn:
[...]

PR #1024, opened 2024-10-17
Exception strings: `TYPE_NOT_SUPPORTED`

SamWilsn:
### What was wrong?

Related to Issue #1021

### How was it (partially) fixed?

Adding a new exception type to match `TransactionException.TYPE_NOT_SUPPORTED`.

#### Cute Animal Picture

### DIFF
--- a/tests/helpers/load_state_tests.py
+++ b/tests/helpers/load_state_tests.py
@@ -11,7 +11,7 @@ from _pytest.mark.structures import ParameterSet
 from ethereum_types.numeric import U64
 
 from ethereum import rlp
-from ethereum.exceptions import InvalidBlock, InvalidTransaction
+from ethereum.exceptions import EthereumException
 from ethereum.utils.hexadecimal import hex_to_bytes
 from ethereum_spec_tools.evm_tools.loaders.fixture_loader import Load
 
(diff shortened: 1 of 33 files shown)


tests/helpers/load_state_tests.py
```diff
@@ -69,7 +69,10 @@ def run_blockchain_st_test(test_case: Dict, load: Load) -> None:
                 break
 
         if block_exception:
-            with pytest.raises(InvalidBlock):
+            # TODO: Once all the specific exception types are thrown,
+            #       only `pytest.raises` the correct exception type instead of
+            #       all of them.
+            with pytest.raises((InvalidBlock, InvalidTransaction)):
                 add_block_to_chain(chain, json_block, load, mock_pow)
```
gurukamath, line 76: Wondering if we can make it more specific on our test consumption side as well. Treat `InvalidBlock` and `InvalidTransaction` separately based on `block_exception`. Not urgent right now but some thing to consider going forward.

gurukamath (approved):

tests/helpers/load_state_tests.py
SamWilsn: That's what I was thinking, and intended to convey with the above TODO. We can't do it yet because I haven't converted all the transaction exceptions yet (just one specific case.)

PR #1029, opened 2024-11-05
evm_tools: Parse `state_test` in daemon as t8n option

marioevz:
### What was wrong?

T8N's option `--state-test` could not be passed to the t8n handler if we were running in daemon mode.

### How was it fixed?

Added parsing for a `state_test` field in the daemon and, if present, the argument is added to the T8N call.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/daemon.py
+++ b/src/ethereum_spec_tools/evm_tools/daemon.py
@@ -57,6 +57,8 @@ class _EvmToolHandler(BaseHTTPRequestHandler):
             f"--state.chainid={content['state']['chainid']}",
             f"--state.reward={content['state']['reward']}",
         ]
+        if 'state_test' in content and content['state_test']:
+            args.append("--state-test")
 
         self.send_response(200)
         self.send_header("Content-type", "application/octet-stream")


gurukamath (approved):
LGTM!

SamWilsn:
Does this fix the same issue as https://github.com/ethereum/execution-specs/pull/1006 ?

marioevz:
> Does this fix the same issue as #1006 ?

It seems so, it's an easy fix for EEST to inform t8n that we are filling a state test, I'm preparing a PR right now.

src/ethereum_spec_tools/evm_tools/daemon.py
```diff
@@ -57,6 +57,8 @@ def do_POST(self) -> None:
             f"--state.chainid={content['state']['chainid']}",
             f"--state.reward={content['state']['reward']}",
         ]
+        if 'state_test' in content and content['state_test']:
+            args.append("--state-test")
```
SamWilsn, line 61: My concern here is the same as in #1006: detecting flags based on the content of the JSON makes the command line interface inconsistent. If the t8n CLI requires `--state-test` but t8n as a daemon does not, that feels weird.

Originally this file parsed the command line arguments from the query string. Should we bring that back?

https://github.com/ethereum/execution-specs/blob/457529110356df26284c38b449b672a0f2e802d8/src/ethereum_spec_tools/evm_tools/daemon.py#L36-L37

marioevz: I understand, and it makes sense that we need to standardize this, and also if that makes the code on your end cleaner it would be ok for us too.

PR #1030, opened 2024-11-06
Extract t8n arguments from daemon query string

SamWilsn:
### What was wrong?

Closes #1029
Closes #1006 

### How was it fixed?

This commit allows daemon clients to pass command line arguments in the query string (under the key `arg`.) To add multiple arguments, use multiple key/value pairs. For example:

```bash
curl --unix-socket /run/user/1000/ethereum-spec-evm/daemon.sock 'http://localhost/?arg=--help'
```

```bash
curl --unix-socket /run/user/1000/ethereum-spec-evm/daemon.sock 'http://localhost/?arg=--help&arg=--state-test'
```

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/daemon.py
+++ b/src/ethereum_spec_tools/evm_tools/daemon.py
@@ -12,6 +12,7 @@ from io import StringIO, TextIOWrapper
 from socket import socket
 from threading import Thread
 from typing import Any, Tuple, Union
+from urllib.parse import parse_qs, urlparse
 
 
 def daemon_arguments(subparsers: argparse._SubParsersAction) -> None:
@@ -57,6 +58,16 @@ class _EvmToolHandler(BaseHTTPRequestHandler):
             f"--state.chainid={content['state']['chainid']}",
             f"--state.reward={content['state']['reward']}",
         ]
+
+        query_string = urlparse(self.path).query
+        if query_string:
+            query = parse_qs(
+                query_string,
+                keep_blank_values=True,
+                strict_parsing=True,
+                errors="strict",
+            )
+            args += query.get("arg", [])
 
         self.send_response(200)
         self.send_header("Content-type", "application/octet-stream")


marioevz:
It seems to be working when passing a single parameter as "?key=--state-test" for example, but when passing none is complaining with the following error:
```
Exception occurred during processing of request from ['local', 0]                                                                    
Traceback (most recent call last):                                                                                                   
  File "/usr/lib/python3.10/socketserver.py", line 316, in _handle_request_noblock                                                   
    self.process_request(request, client_address)                                                                                    
  File "/usr/lib/python3.10/socketserver.py", line 347, in process_request                                                           
    self.finish_request(request, client_address)                                                                                     
  File "/home/marioevz/.cache/ethereum-spec-evm-resolver/Cancun/src/ethereum_spec_tools/evm_tools/daemon.py", line 100, in finish_req
uest                                                                                                                                 
    super().finish_request(request, client_address)                                                                                  
  File "/usr/lib/python3.10/socketserver.py", line 360, in finish_request                                                            
    self.RequestHandlerClass(request, client_address, self)                                                                          
  File "/usr/lib/python3.10/socketserver.py", line 747, in __init__                                                                  
    self.handle()                                                                                                                      File "/usr/lib/python3.10/http/server.py", line 433, in handle                                                                     
    self.handle_one_request()                                                                                                        
  File "/usr/lib/python3.10/http/server.py", line 421, in handle_one_request                                                         
    method()                                                                                                                         
  File "/home/marioevz/.cache/ethereum-spec-evm-resolver/Cancun/src/ethereum_spec_tools/evm_tools/daemon.py", line 62, in do_POST    
    query = parse_qs(                                                                                                                
  File "/usr/lib/python3.10/urllib/parse.py", line 714, in parse_qs                                                                  
    pairs = parse_qsl(qs, keep_blank_values, strict_parsing,                                                                         
  File "/usr/lib/python3.10/urllib/parse.py", line 775, in parse_qsl                                                                 
    raise ValueError("bad query field: %r" % (name_value,))                                                                          
ValueError: bad query field: ''
```
I'm debugging to see if I can find the issue, I tried passing empty string and `"?"` but same behavior.

SamWilsn:
What's the full URL you're using?

marioevz:
> What's the full URL you're using?

This is an example I obtained from the resolver right before sending the query:
With `--state-test`:
```
http+unix://%2Ftmp%2Ftmp5975ll06%2FCancun.3885864.sock/?arg=--state-test
```
Without `--state-test`:
```
http+unix://%2Ftmp%2Ftmpjshog8l0%2FCancun.3886038.sock/
```

marioevz:
I reproduced locally and it seems like `parse_qs` complains if the result from `urlparse(self.path).query` is an empty string, which is the case if no arguments are passed.

It's a weird behavior from `parse_qs` IMO, but the easiest solution seems to be to skip the query parsing altogether in the case that `urlparse(self.path).query` is an empty string.

I tested locally and it seems to work, but let me know what you think.

SamWilsn:
That's really odd. Only thing I can think of is that I'm in Python 3.12, and maybe they changed the behaviour? Pushed a fix; let me know if it works for you.

marioevz (approved):
LGTM.

Last change seems to have fixed my issue, thanks!

gurukamath (approved):

PR #1034, opened 2024-11-12
Switch to ethereum-rlp

SamWilsn:
### What was wrong?

Circular dependency between tests and EELS.

### How was it fixed?

Splitting out the RLP code into its own library.

#### Cute Animal Picture


tests/helpers/load_state_tests.py
```diff
@@ -72,14 +74,14 @@ def run_blockchain_st_test(test_case: Dict, load: Load) -> None:
             # TODO: Once all the specific exception types are thrown,
             #       only `pytest.raises` the correct exception type instead of
             #       all of them.
-            with pytest.raises(EthereumException):
+            with pytest.raises((EthereumException, RLPException)):
```
gurukamath, line 77: Might be worth it to investigate in a separate issue/PR if we can separate out these two exception and also see if we can start using `InvalidBlock` instead of `EthereumException`

gurukamath (approved):

tests/helpers/load_state_tests.py
SamWilsn: This used to be `InvalidBlock`, but after the changes I've done for #1021, not everything is a subclass.

PR #1038, opened 2024-11-20
Transition types

SamWilsn:
### What was wrong?

Header/block types near fork transitions weren't strict enough.

### How was it fixed?

Made them stricter.

#### Cute Animal Picture


tests/helpers/load_state_tests.py
```diff
@@ -24,6 +27,7 @@ class NoTestsFound(Exception):
 
 
 def run_blockchain_st_test(test_case: Dict, load: Load) -> None:
+    load.fork.fork_criteria = ByBlockNumber(0)
```
SamWilsn, line 30: This line feels like a hack, but I'm not sure what the correct solution would look like. Open to suggestions!

PR #1042, opened 2024-12-02
Make EXTCODE* opcodes not follow delegations

petertdavies:
### What was wrong?

EIP-7702 was changed so that EXTCODE* opcodes don't follow delegations. Instead they now behave as if the code was `0xef01` when they see a delegation.

### How was it fixed?

Update the EELS behaviour. Note other than linting I have not tested these changes in any way.

#### Cute Animal Picture


src/ethereum/prague/vm/eoa_delegation.py
```diff
@@ -144,6 +145,32 @@ def access_delegation(
     return True, address, code, access_gas_cost
 
 
+def access_delegation_read(
```
SamWilsn, line 148: The name of this function doesn't convey that it's reading code. I think it's also in the wrong place.

What do you think of adding this function to every fork, probably in their `environment.py` files. In preceding forks, it'll just return `get_account(evm.env.state, address).code` and in prague it'll have the delegation check.

gurukamath: Or maybe add a `get_code` function to `state.py`. It already has the `set_code` function

petertdavies: The function is so simple that I just removed it entirely

gurukamath (approved):

SamWilsn (approved):

PR #1045, opened 2024-12-03
Implement EIP 7742 in Prague

petertdavies:
This PR implements EIP 7742 on top of Prague. It have done some brief tests with the [execution-spec-tests PR](https://github.com/ethereum/execution-spec-tests/pull/986) and these changes seem to work.

~This PR should not be merged since the status of EIP 7742 is very uncertain.~

Please check that EIP 7742 is actually going into `prague-devnet-5` before merging.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/prague/vm/gas.py
+++ b/src/ethereum/prague/vm/gas.py
@@ -75,7 +75,7 @@ GAS_BLS_G2_MAP = Uint(23800)
 
 GAS_PER_BLOB = Uint(2**17)
 MIN_BLOB_GASPRICE = Uint(1)
-BLOB_GASPRICE_UPDATE_FRACTION = Uint(5007716)
+BLOB_BASE_FEE_UPDATE_FRACTION = Uint(5007716)
 
 
 @dataclass
@@ -338,7 +338,7 @@ def calculate_blob_gas_price(excess_blob_gas: U64) -> Uint:
     return taylor_exponential(
         MIN_BLOB_GASPRICE,
         Uint(excess_blob_gas),
-        BLOB_GASPRICE_UPDATE_FRACTION,
+        BLOB_BASE_FEE_UPDATE_FRACTION,
     )
 
 
--- a/src/ethereum/prague/vm/gas.py
+++ b/src/ethereum/prague/vm/gas.py
@@ -73,7 +73,6 @@ GAS_BLS_G2_ADD = Uint(600)
 GAS_BLS_G2_MUL = Uint(22500)
 GAS_BLS_G2_MAP = Uint(23800)
 
-TARGET_BLOB_GAS_PER_BLOCK = U64(786432)
 GAS_PER_BLOB = Uint(2**17)
 MIN_BLOB_GASPRICE = Uint(1)
 BLOB_GASPRICE_UPDATE_FRACTION = Uint(5007716)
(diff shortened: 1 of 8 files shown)


spencer-tb:
Thanks again for adding this :D

We were wondering the following should be added too:
https://github.com/spencer-tb/execution-specs/commit/07f4774873cf25e062d925f43c740f6fdf8ada24#diff-1ead0e19bcf5b72f6e0ff1f07a9d44a25a738020803d5ea42839da1e3d58a8c5R433-R434

Checking for contract creating blob txs!

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -357,7 +357,9 @@ def apply_body(self) -> None:
                     env, tx
                 )
 
-                if self.fork.is_after_fork("ethereum.cancun"):
+                if self.fork.is_after_fork(
+                    "ethereum.cancun"
+                ) and not self.fork.is_after_fork("ethereum.prague"):
                     blob_gas_used += self.fork.calculate_total_blob_gas(tx)
```
gurukamath, line 363: Incrementing the `blob_gas_used` has to happen post Prague as well I believe. It's only the `MAX_BLOB_GAS_PER_BLOCK` check that is removed

src/ethereum/prague/vm/gas.py
```diff
@@ -286,10 +288,11 @@ def calculate_excess_blob_gas(parent_header: Header) -> U64:
     parent_blob_gas = (
         parent_header.excess_blob_gas + parent_header.blob_gas_used
     )
-    if parent_blob_gas < TARGET_BLOB_GAS_PER_BLOCK:
+    blob_gas = GAS_PER_BLOB * header.target_blobs_per_block
```
gurukamath, line 291: Does naming this variable `blob_gas_target` make sense?

src/ethereum_spec_tools/evm_tools/t8n/env.py
```diff
@@ -114,12 +116,25 @@ def read_excess_blob_gas(self, data: Any, t8n: "T8N") -> None:
             self.parent_excess_blob_gas + self.parent_blob_gas_used
         )
 
-        target_blob_gas_per_block = t8n.fork.TARGET_BLOB_GAS_PER_BLOCK
+        if "currentTargetBlobsPerBlock" in data:
```
gurukamath, line 119: For post-Prague forks, `currentTargetBlobsPerBlock` seems to be a mandatory field. So, if that is not available in the data, I believe the tools should throw an exception

marioevz:
Created https://github.com/petertdavies/execution-specs/pull/4 to implement changes in https://github.com/ethereum/EIPs/pull/9122.

On a side note, I'm seeing zero value in the result's field `blobGasUsed`, regardless of whether a blob is included in the transactions or not, I'm not sure if this is derived from the changes in this PR though @gurukamath

gurukamath:
> On a side note, I'm seeing zero value in the result's field `blobGasUsed`, regardless of whether a blob is included in the transactions or not, I'm not sure if this is derived from the changes in this PR though @gurukamath

Could [this](https://github.com/ethereum/execution-specs/pull/1045#discussion_r1876499398) be the issue?

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
petertdavies: Fixed

src/ethereum_spec_tools/evm_tools/t8n/env.py
petertdavies: As currently written, we don't take into account what the current fork is at this point. This is an issue for the t8n refactor/rewrite to deal with.

gurukamath:
LGTM!

PR #1046, opened 2024-12-04
EIP-7685: exclude empty requests in requests list

marioevz:
### What was wrong?

The following PRs include breaking changes in the behavior of requests for devnet-5:

https://github.com/ethereum/execution-apis/pull/599
https://github.com/ethereum/EIPs/pull/8989

### How was it fixed?

Remove requests elements from the requests list if the byte array is empty.

Also do not remove the `request_type` byte in the return list of the T8N result since it's now necessary to calculate the requests commitment.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/prague/fork.py
+++ b/src/ethereum/prague/fork.py
@@ -890,7 +890,8 @@ def process_general_purpose_requests(
     """
     # Requests are to be in ascending order of request type
     requests_from_execution: List[Bytes] = []
-    requests_from_execution.append(DEPOSIT_REQUEST_TYPE + deposit_requests)
+    if len(deposit_requests) > 0:
+        requests_from_execution.append(DEPOSIT_REQUEST_TYPE + deposit_requests)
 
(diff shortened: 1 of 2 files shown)


marioevz:
I see that the tox issue is from running our latest generated blockchain tests, which makes sense since we are altering the expected hash in the header of any Prague test, but to generate a new set of tests we need this PR.

It's an interesting dependency problem.

SamWilsn:
Can you generate and release new tests before this is merged, or should I force merge this?

marioevz:
> Can you generate and release new tests before this is merged, or should I force merge this?

I think our option could be to combine all required branches into a temporary `forks/prague` branch in some of our forks, make the release by pointing to that, then re-run CI in this PR.

petertdavies:
This has now been pointed at the `devnets/prague/5` branch. Tests will be disabled in that branch until EEST generates `prague-devnet-5` fixtures.

petertdavies (approved):
LGTM

Issue #1049, opened 2024-12-06
bug: not able to call `get_type_hints` on *Node types


ClementWalter:
See this MWE

```
from typing import get_type_hints
from ethereum.cancun.trie import BranchNode

get_type_hints(BranchNode)
```

This raises because the nodes use `rlp.Extended` as a type annotation instead of `Extended`.

Doing in `rlp.py`

```
from ..rlp import Extended

# ...

@slotted_freezable
@dataclass
class LeafNode:
    """Leaf node in the Merkle Trie"""

    rest_of_key: Bytes
    value: Extended


@slotted_freezable
@dataclass
class ExtensionNode:
    """Extension node in the Merkle Trie"""

    key_segment: Bytes
    subnode: Extended


@slotted_freezable
@dataclass
class BranchNode:
    """Branch node in the Merkle Trie"""

    subnodes: List[Extended]
    value: Extended
```

solves the issue.

SamWilsn:
@ClementWalter is this still an issue after merging #1050?

ClementWalter:
tried now, problem solved indeed

<img width="484" height="117" alt="Image" src="https://github.com/user-attachments/assets/131f1581-e17c-4f42-becb-7a3c7fbc4507" />

PR #1050, opened 2024-12-06
Import Extended type directly instead of using rlp.Extended

ClementWalter:
### What was wrong?

Solves #1049

### How was it fixed?

As proposed in the issue, I added the `Extended` type to the `Node` union in `trie.py`.

### DIFF
--- a/src/ethereum/arrow_glacier/trie.py
+++ b/src/ethereum/arrow_glacier/trie.py
@@ -30,7 +30,7 @@ from typing import (
     cast,
 )
 
-from ethereum_rlp import rlp
+from ethereum_rlp import Extended, rlp
 from ethereum_types.bytes import Bytes
 from ethereum_types.frozen import slotted_freezable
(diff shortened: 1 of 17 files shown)


SamWilsn:
Seems reasonable. Bit weird that `get_type_hints` can't handle what was there before though.

Assuming the tests pass, can you port this to the rest of the forks?

ClementWalter:
sure, you mean replace all `rlp.Encoded` in the whole cancun fork?

SamWilsn:
Unless it's causing problems elsewhere in Cancun, I think it's fine to leave as you have it now. I mean make the same change in all the other `trie.py` files, like:

https://github.com/ethereum/execution-specs/blob/1adcc1bfe774798bcacc685aebc17bd9935078c3/src/ethereum/frontier/trie.py#L82

SamWilsn (approved):

PR #1053, opened 2024-12-12
Refactor the Environment dataclass

gurukamath:
### What was wrong?

The `Environment` data class has some transaction scope attributes like `sender`, `origin` and a few others. Ideally, this data class should only have block scope attributes and the transaction scope attributes should be moved to the `Message` data class.

This refactor has a few really nice benefits including

1. Clear and logical separation between the scopes
2. Simplifying the signature of a lot of functions including the `apply_body` function (We can now pass just a single `env` parameter instead of a large number of individual attributes that might change with hard forks)
3. Simplification of the code in `t8n`
4. Simplification of system transaction processing code

### How was it fixed?
Move transaction scope attributes to `Message`

#### Cute Animal Picture
Cute Animals - 1 of 1


src/ethereum/cancun/fork.py
```diff
@@ -180,28 +180,29 @@ def state_transition(chain: BlockChain, block: Block) -> None:
...
-    excess_blob_gas = calculate_excess_blob_gas(parent_header)
-    if block.header.excess_blob_gas != excess_blob_gas:
-        raise InvalidBlock
-
-    validate_header(block.header, parent_header)
+    validate_header(chain, block.header)
     if block.ommers != ():
         raise InvalidBlock
+
+    env = vm.Environment(
```
SamWilsn, line 187: I _really_ like this change.

src/ethereum/cancun/vm/__init__.py
```diff
@@ -58,6 +54,8 @@ class Message:
     """
 
     caller: Address
+    origin: Address
```
SamWilsn, line 57: The rule I've been loosely following (not that I've modified either class recently 😅) is that things remaining constant during a transaction belong in `Environment` and things that can change per-call belong in `Message`.

What is your proposed new rule for placing new members?

gurukamath: The rules I have followed are

- Block scoped items go into `Environment`. These are constants over a block. Examples would be anything chain specific or block header specific (chain_id, timestamp etc.). This way, we don't have to re-compute `env` for each transaction.
- Transaction scoped items go into `Message`. These could be constant over calls (e.g. origin, gas_price etc.) or change (e.g. depth, target etc.)

However, I do see value in separating constants over a transaction from those things that change per call.
So may be good idea to have the following
- `BlockEnvironment` - For block scope items
- `TransactionEnvironment` - For constants over a tx
- `Message` - For items that change per message call. This can have an instance of `block_env` and `tx_env`

I feel like this will give us an even cleaner and more logical separation of concerns. What do you think of that idea?

gurukamath: PS: The dataclass names are just a suggestion :). Totally happy to have some other names

SamWilsn: > The rules I have followed are [...]

That makes a lot of sense! I'm happy with that.

> However, I do see value in separating constants over a transaction from those things that change per call.

This sounds like a big refactor. Do you think it's worth it?

gurukamath: I do believe its worth it. I just pushed a few more commits on the Cancun fork to demonstrate the benefits.

1. This allows us to neatly separate out the logic for processing an entire block vs processing individual transactions. So if we want to process a new block, we just run the `apply_body` function and if we want to process an individual transaction, we just run the `process_transaction` function.
2. Point 1 above makes it really clear which components need to be updated when someone is implementing a new EIP. Right now, I feel like its still a bit vague.
3. This change greatly simplifies our `t8n` tool, removing the need for a majority of the fork specific conditionals. See [here](https://github.com/gurukamath/execution-specs/blob/10f3a2766a076afb51580709165cba4fca4efd7f/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L325). So for state tests we essentially just call `process_transaction` and for blockchain tests, we just call the `apply_body` function. And the signatures of these functions are unlikely to change greatly from fork to fork.
4. Simplifies some of our tracing logic by providing us a vehicle to store traces over an entire transaction. This also makes it easier to move to writing traces immediately to file (as opposed to status quo where we store them all and output them in the end).

src/ethereum/cancun/vm/__init__.py
```diff
@@ -57,6 +70,8 @@ class Message:
     Items that are used by contract creation or message call.
     """
 
+    block_env: BlockEnvironment
+    tx_env: TransactionEnvironment
```
SamWilsn, line 74: Would it make sense to introduce a third "environment" for the frame?

```python
class BlockEnvironment:
    ...

class TransactionEnvironment:
    ...

class FrameEnvironment:
    caller: Address
    target: Union[Bytes0, Address]
    current_target: Address
    gas: Uint
    value: U256
    data: Bytes
    code_address: Optional[Address]
    code: Bytes
    depth: Uint
    should_transfer_value: bool
    is_static: bool
    parent_evm: Optional["Evm"]

class Message:
    block: BlockEnvironment
    transaction: TransactionEnvironment
    frame: FrameEnvironment
```

src/ethereum/cancun/vm/__init__.py
```diff
@@ -28,27 +28,40 @@
...
 
 @dataclass
-class Environment:
+class BlockEnvironment:
     """
     Items external to the virtual machine itself, provided by the environment.
     """
 
-    caller: Address
+    chain_id: U64
```
SamWilsn, line 36: I almost wish there was a `ChainEnvironment` or a reference to the genesis here instead of `chain_id`.

src/ethereum/cancun/vm/__init__.py
gurukamath: This would line up the nomenclature pretty well but I have been looking at the `Message` dataclass serving the same purpose as the `FrameEnvironment` dataclass in your suggestion. I don't have any strong opinions on either choice tbh.

src/ethereum/cancun/vm/__init__.py
gurukamath: Do you see any further chain specific attributes that could be added to this in the future? If not, just maintaining the `chain_id` as a special case attribute in `BlockEnvironment` for now makes sense to me.

SamWilsn: I guess the alternative would be including a `GenesisConfiguration`, which seems overkill for just the chain id.

src/ethereum/cancun/vm/__init__.py
gurukamath: The symmetry argument sounds good to me but the question largely boils down to complexity since introducing `FrameEnvironment` could be a large change. Also, I think we can introduce `FrameEnvironment` in the next iteration to avoid a huge PR

SamWilsn: I think that makes sense.

gurukamath:
Closing in favor of #1131

PR #1055, opened 2024-12-16
Add EIP-7691

marioevz:
### What was wrong?

EIP-7691 is missing from Prague

### How was it fixed?

Implemented the changes in the constants for EIP-7691.

#### Cute Animal Picture


src/ethereum/prague/fork.py
```diff
@@ -99,7 +100,7 @@
     "0x01aBEa29659e5e97C95107F20bb753cD3e09bBBb"
 )
 SYSTEM_TRANSACTION_GAS = Uint(30000000)
-MAX_BLOB_GAS_PER_BLOCK = 786432
+MAX_BLOB_GAS_PER_BLOCK = 9 * GAS_PER_BLOB
```
gurukamath, line 103: Although the numbers add up, I don't see this logic explicitly mentioned in either EIP-7691 or in EIP-4844. Is this an upcoming change expected in EIP-7691?

gurukamath: I think it is implied that this is the relationship but would be good to have it explicitly mentioned in EIP-7691 since EIP-4844 never explicitly defined `MAX_BLOBS_PER_BLOCK`

marioevz: Yes I think you're right, I borrowed this from EIP-7742 since that one does drop the blob-gas-per-block in favor of blobs-per-block.

I think this 7691 still needs an update to be uncoupled from 7742.

marioevz: Created a PR in the EIPs repo: https://github.com/ethereum/EIPs/pull/9149

I'll update this PR accordingly. Thanks for catching this!

gurukamath (approved):

PR #1057, opened 2024-12-17
prague: update `chain_id` in 7702 set code txs to u256

danceratopz:
Updates the EIP-7702 set code transaction type to have a nonce of type u256 instead of u64, see:
- ethereum/EIPs#9143.

Please check that no other instances were missed.

Corresponding change in EEST:
- ethereum/execution-spec-tests#1026

#### Cute Animal Picture

image

### DIFF
--- a/src/ethereum/prague/transactions.py
+++ b/src/ethereum/prague/transactions.py
@@ -117,7 +117,7 @@ class SetCodeTransaction:
     The transaction type added in EIP-7702.
     """
 
-    chain_id: U256
+    chain_id: U64
     nonce: U64
     max_priority_fee_per_gas: Uint
     max_fee_per_gas: Uint
--- a/src/ethereum/prague/transactions.py
+++ b/src/ethereum/prague/transactions.py
@@ -117,7 +117,7 @@ class SetCodeTransaction:
     The transaction type added in EIP-7702.
     """
 
-    chain_id: U64
+    chain_id: U256
     nonce: U64
     max_priority_fee_per_gas: Uint
     max_fee_per_gas: Uint
--- a/src/ethereum/prague/vm/eoa_delegation.py
+++ b/src/ethereum/prague/vm/eoa_delegation.py
@@ -163,7 +163,7 @@ def set_delegation(message: Message, env: Environment) -> U256:
     """
     refund_counter = U256(0)
     for auth in message.authorizations:
-        if auth.chain_id not in (env.chain_id, U64(0)):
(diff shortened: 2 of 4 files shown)


src/ethereum/prague/transactions.py
```diff
@@ -117,7 +117,7 @@ class SetCodeTransaction:
     The transaction type added in EIP-7702.
     """
 
-    chain_id: U64
+    chain_id: U256
```
gurukamath, line 120: The EIP still says

```
The fields chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, value, data, and access_list of the outer transaction follow the same semantics as [EIP-4844](https://github.com/lightclient/EIPs/blob/7702-chainid-u256/EIPS/eip-4844.md).
```

I looked back at 4844, 1559 and a couple of other ones in the series and none of them clearly define whether `chain_id` is `U64` or `U256`. In fact, EIP-1559 has some pseudo-code that looks at it as `int`

So I suggest explicitly changing the 7702 EIP text to specify `tx.chain_id` as `U256`. Right now it only defines `auth.chain_id` to be `U256`

src/ethereum/prague/vm/eoa_delegation.py
```diff
@@ -163,7 +163,7 @@ def set_delegation(message: Message, env: Environment) -> U256:
     """
     refund_counter = U256(0)
     for auth in message.authorizations:
-        if auth.chain_id not in (env.chain_id, U64(0)):
+        if auth.chain_id not in (env.chain_id, U256(0)):
```
gurukamath, line 166: The dataclass definition of the `Authorization` dataclass also has to be updated

danceratopz: Done in febb39764.

src/ethereum/prague/transactions.py
danceratopz: febb39764 should now reflect the ethereum/EIPs#9143.

gurukamath (approved):
LGTM!

gurukamath:
Keeping this on hold and unmerged until ACD since there are other proposals. See [Update EIP-7702: remove chain_id in authorization tuple](https://github.com/ethereum/EIPs/pull/9152) and [this](https://ethereum-magicians.org/t/eip-7702-set-eoa-account-code/19923/326)

gurukamath:
This change was agreed upon in ACD 202 and is merged.

PR #1065, opened 2024-12-20
Run all tests

### LINKED ISSUE #1028: Address collision when storage exists: spec vs impl mismatch
### Metadata

* **Hardfork**: Cancun

### What was wrong?

The Python execution specs consider a create message to fail due to address collision if the account in question has nonce or code:
https://github.com/ethereum/execution-specs/blob/98d6ddaaa709a2b7d0cd642f4cfcdadc8c0808e1/src/ethereum/cancun/vm/interpreter.py#L110

But the tests and client implementation seems to consider it a collision also if the account has non-empty storage.

### Sources

EIP specifying the geth behaviour: https://eips.ethereum.org/EIPS/eip-7610

Geth check: https://github.com/ethereum/go-ethereum/blob/a1093d98eb3260f2abf340903c2d968b2b891c11/core/vm/evm.go#L460

Test relying on the geth behaviour: https://github.com/ethereum/tests/blob/4c87ebbf024a3e9ec842bcce90564f629a3bcd82/BlockchainTests/GeneralStateTests/stRevertTest/RevertInCreateInInit_Paris.json

### Additional Context

If the python spec is corrected, there are a few simplifications also possible, e.g. no need to delete the storage when creating an account nor to keep track of created accounts.

### LINKED ISSUE #1064: Run all tests for pre-Cancun forks
The `ethereum/tests` repository recently moved tests for forks earlier than `Cancun` into the `LegacyTests`. The testing code should be update to reflect this.

gurukamath:
( closes #1064 )
( closes #1028 )

### What was wrong?
The ethereum/tests repository recently moved tests for forks earlier than Cancun into the LegacyTests. The testing code should be update to reflect this.

Related to Issue #1064 

### How was it fixed?
Update the test fixture path for pre-Cancun forks

#### Cute Animal Picture

Cute Animals - 1 of 1

### DIFF
--- a/tests/berlin/test_state_transition.py
+++ b/tests/berlin/test_state_transition.py
@@ -27,7 +27,7 @@ run_berlin_blockchain_st_tests = partial(
 ETHEREUM_TESTS_PATH = TEST_FIXTURES["ethereum_tests"]["fixture_path"]
 
 # Run state tests
-test_dir = f"{ETHEREUM_TESTS_PATH}/BlockchainTests/"
+test_dir = f"{ETHEREUM_TESTS_PATH}/LegacyTests/Cancun/BlockchainTests/"
 
 # Every test below takes more than  60s to run and
 # hence they've been marked as slow
--- a/setup.cfg
+++ b/setup.cfg
@@ -177,7 +177,7 @@ doc =
     fladrif>=0.2.0,<0.3.0
 
 optimized =
-    rust-pyspec-glue>=0.0.7,<0.1.0
+    rust-pyspec-glue>=0.0.9,<0.1.0
     ethash @ git+https://github.com/chfast/ethash.git@e08bd0fadb8785f7ccf1e2fb07b75f54fe47f92e
 
 [flake8]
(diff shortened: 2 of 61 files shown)


tests/berlin/test_state_transition.py
```diff
@@ -27,7 +27,7 @@
 ETHEREUM_TESTS_PATH = TEST_FIXTURES["ethereum_tests"]["fixture_path"]
 
 # Run state tests
-test_dir = f"{ETHEREUM_TESTS_PATH}/BlockchainTests/"
+test_dir = f"{ETHEREUM_TESTS_PATH}/LegacyTests/Cancun/BlockchainTests/"
```
SamWilsn, line 30: Just confirming this should be "Cancun" in the berlin fork.

gurukamath: Yes. `Frontier` to `Constantinople` are under `/LegacyTests/Constantinople` and then `Istanbul` to `Shanghai` are under `/LegacyTests/Cancun`

PR #1066, opened 2024-12-20
dev: change type of branch node subnodes

enitrat:
### What was wrong?
The type for the BranchNode's subnodes was `List[rlp.Extended]`. While this is not inherently wrong, a more suitable type would be a tuple with 16 rlp.Extended elements, each representing a BranchNode of the trie, as there's no intent to mutate the subnodes after instantiation.

Related to Issue #

### How was it fixed?

Changed to `Annotated[Tuple[rlp.Extended, ...], 16]`. Also open to just repeating 16 times the rlp.Extended argument, but using an annotation sounds more convenient rather than repeating. The annotation conveys at type hint that the tuple is of size 16. 

#### Cute Animal Picture
Our beloved moodeng friend

### DIFF
--- a/src/ethereum/arrow_glacier/trie.py
+++ b/src/ethereum/arrow_glacier/trie.py
@@ -16,7 +16,6 @@ The state trie is the structure responsible for storing
 import copy
 from dataclasses import dataclass, field
 from typing import (
-    Annotated,
     Callable,
     Dict,
     Generic,
--- a/src/ethereum/arrow_glacier/trie.py
+++ b/src/ethereum/arrow_glacier/trie.py
@@ -33,6 +33,7 @@ from typing import (
 from ethereum_types.bytes import Bytes
 from ethereum_types.frozen import slotted_freezable
 from ethereum_types.numeric import U256, Uint
+from typing_extensions import assert_type
 
 from ethereum.crypto.hash import keccak256
 from ethereum.london import trie as previous_trie
@@ -485,6 +486,6 @@ def patricialize(
         for k in range(16)
     )
     return BranchNode(
-        cast(BranchSubnodes, subnodes),
+        cast(BranchSubnodes, assert_type(subnodes, Tuple[rlp.Extended, ...])),
         value,
     )
--- a/src/ethereum/arrow_glacier/trie.py
+++ b/src/ethereum/arrow_glacier/trie.py
@@ -461,7 +461,7 @@ def patricialize(
             branches[key[level]][key] = obj[key]
 
     return BranchNode(
-        (
+        tuple(
             encode_internal_node(patricialize(branches[k], level + Uint(1)))
             for k in range(16)
         ),
--- a/src/ethereum/arrow_glacier/trie.py
+++ b/src/ethereum/arrow_glacier/trie.py
@@ -16,6 +16,7 @@ The state trie is the structure responsible for storing
 import copy
 from dataclasses import dataclass, field
 from typing import (
+    Annotated,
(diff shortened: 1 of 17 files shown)


SamWilsn:
Does the annotation get enforced? If not, I'd prefer the tuple with 16 elements, probably as a proper type or type alias.

enitrat:
The Annotation is not enforced. It's only a "hint" for the reader that it expects 16 elements. We could create a proper type that

enitrat:
done :)

SamWilsn (approved):

enitrat:
Hi @SamWilsn, any chance you can help me understand why CI is failing ?

src/ethereum/arrow_glacier/trie.py
```diff
@@ -458,10 +480,11 @@ def patricialize(
...
+    subnodes = tuple(
+        encode_internal_node(patricialize(branches[k], level + Uint(1)))
+        for k in range(16)
+    )
     return BranchNode(
-        [
-            encode_internal_node(patricialize(branches[k], level + Uint(1)))
-            for k in range(16)
-        ],
+        cast(BranchSubnodes, subnodes),
```
SamWilsn, line 488: Why do you need a `cast` here?

enitrat: I need to create an explicit tuple (class tuple), then assign it the type explicitly. Was getting a bunch of errors trying other things, but if you have a better idea I can do it

gurukamath: While verbose, would maybe explicitly unpacking the elements of the `subnodes` worth the type checking gains?

enitrat: you mean like:

```
    subnodes = (
        encode_internal_node(patricialize(branches[0], level + Uint(1))),
        encode_internal_node(patricialize(branches[1], level + Uint(1))),
        encode_internal_node(patricialize(branches[2], level + Uint(1))),
        encode_internal_node(patricialize(branches[3], level + Uint(1))),
        encode_internal_node(patricialize(branches[4], level + Uint(1))),
        encode_internal_node(patricialize(branches[5], level + Uint(1))),
        encode_internal_node(patricialize(branches[6], level + Uint(1))),
        encode_internal_node(patricialize(branches[7], level + Uint(1))),
        encode_internal_node(patricialize(branches[8], level + Uint(1))),
        encode_internal_node(patricialize(branches[9], level + Uint(1))),
        encode_internal_node(patricialize(branches[10], level + Uint(1))),
        encode_internal_node(patricialize(branches[11], level + Uint(1))),
        encode_internal_node(patricialize(branches[12], level + Uint(1))),
        encode_internal_node(patricialize(branches[13], level + Uint(1))),
        encode_internal_node(patricialize(branches[14], level + Uint(1))),
        encode_internal_node(patricialize(branches[15], level + Uint(1))),
    )
    return BranchNode(subnodes, value)

```
?

gurukamath: Yeah. Something like that if we really want to avoid `cast`.

SamWilsn:
I think avoiding the `cast` here hurts readability, so here's my proposal: https://github.com/kkrt-labs/execution-specs/pull/1

SamWilsn:
This passed for me. I'll dig into this a bit. I bet it's something to do with my local Python version.

SamWilsn:
Fixed: https://github.com/kkrt-labs/execution-specs/pull/2

Issue #1067, opened 2024-12-31
dev: `refund_counter` should be `Uint`, not `int`


enitrat:
In https://github.com/ethereum/execution-specs/blob/068b9a786152ad069043308d584e0442bd65a285/src/ethereum/cancun/vm/__init__.py#L77-L101 the correct data type for `refund_counter` would be `Uint`, not `int`

PR #1068, opened 2024-12-31
dev: make refund counter a `Uint`

enitrat:
### What was wrong?

Related to Issue #1067

### How was it fixed?
I modified refund_counter to be a Uint. It cannot go negative, nor can it go beyond the gas_used, which is itself a Uint

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/cancun/fork.py
+++ b/src/ethereum/cancun/fork.py
@@ -779,7 +779,7 @@ def process_transaction(
     output = process_message_call(message, env)
 
     gas_used = tx.gas - output.gas_left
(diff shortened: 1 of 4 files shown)


SamWilsn:
I'm pretty sure our refund counter can go negative per call frame. We switched to `int` pretty recently, in b3b8e1433a01735a42fcb83ede0b4e8e084ac968.

enitrat:
you're right, that can happen if a frame n clears a value and frame n+1  sets it back... sorry

SamWilsn:
No worries! We should probably add a comment explaining why it's an `int`.

PR #1069, opened 2025-01-02
chore: format markdown table

NotNotARobot:
This PR makes the markdown table easier to understand in code-view via it's formatting


SamWilsn:
This table is ugly with or without extra spacing.

Issue #1070, opened 2025-01-06
storage trie type should by Trie[Bytes32, U256] and not Trie[Bytes, U256]


enitrat:
### Metadata

* **Hardfork**: hardfork-name

### What was wrong?

The type used for the storage trie is Trie[Bytes, U256]. However the yellow paper specifies that

> storageRoot: A 256-bit hash of the root node of a
Merkle Patricia tree that encodes the storage contents of the account (a mapping between 256-bit
integer values), encoded into the trie as a mapping
from the Keccak 256-bit hash of the 256-bit integer
keys to the RLP-encoded 256-bit integer values.
The hash is formally denoted σ[a]s.

The storage content is implemented as "(a mapping between 256-bit integer values)" - which means the proper key type should be Bytes32, and not Bytes, as the bit-length of the keys is always 256.

When popping the key from the stack in `sload`, we convert it to a byte32. https://github.com/ethereum/execution-specs/blob/4c7eaa840c421a1db2c01617532f31d08dc3dc6e/src/ethereum/cancun/vm/instructions/storage.py#L49

### Sources

[Yellow Paper, page 4](https://ethereum.github.io/yellowpaper/paper.pdf)

gurukamath:
Closed with #1071 and #1096

PR #1071, opened 2025-01-06
dev: update storage trie type

enitrat:
### What was wrong?

Related to Issue # https://github.com/ethereum/execution-specs/issues/1070

### How was it fixed?

Change Bytes to Bytes32

#### Cute Animal Picture

)

### DIFF
--- a/src/ethereum_optimized/state_db.py
+++ b/src/ethereum_optimized/state_db.py
@@ -345,7 +345,7 @@ def get_optimized_state_patches(fork: str) -> Dict[str, Any]:
 
     @add_item(patches)
     def get_storage_original(
-        state: State, address: Address, key: Bytes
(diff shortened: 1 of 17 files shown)


src/ethereum_optimized/state_db.py
```diff
@@ -328,7 +328,7 @@ def rollback_transaction(state: State) -> None:
             _rollback_transaction(state)
 
     @add_item(patches)
-    def get_storage(state: State, address: Address, key: Bytes) -> U256:
+    def get_storage(state: State, address: Address, key: Bytes32) -> U256:
```
gurukamath, line 331: The `key` in the `get_storage_original` function should also be `Bytes32`

gurukamath (changes requested):

SamWilsn (approved):

PR #1095, opened 2025-01-25
change type of branch node subnodes (Prague port)

SamWilsn:
### What was wrong?

Related to Issue #1066


gurukamath (approved):
depencency in `setup.cfg` should be updated to `typing_extensions>=4.2`

PR #1096, opened 2025-01-25
update storage trie type (#1070 #1071)

SamWilsn:
### What was wrong?

Related to Issue #1070 #1071


gurukamath (changes requested):
The changes to `src/ethereum_optimized/state_db.py` will also need to be ported over I think

gurukamath (approved):

Issue #817, opened 2023-07-26
Idea: "latest" directory to always hold latest fork


lightclient:
Just an idea. I always have to search for "shanghai" when coming to the spec, it might be nice to have a directory that is update with each fork called "latest". Downside is if people link to things in it, they could break big time.

SamWilsn:
We might be able to do this with a custom 404 page and some JavaScript: https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-custom-404-page-for-your-github-pages-site

PR #1104, opened 2025-01-31
Add latest symlinks

frangio:
For discussion. I'd personally find this very useful. 

### What was wrong?

Navigating to the latest specs requires knowing the fork name. Even though this information is available in the repo, it can be slow to remember and find in the list.

### How was it fixed?

A symlink called `latest` was added.

#### Cute Animal Picture

image


SamWilsn:
Looks like this closes #817.

Apparently you can `from ethereum import latest` in Python, which is interesting. Means that this would probably interact with our [versioning scheme](https://github.com/ethereum/execution-specs/blob/master/RELEASING.md#about-versions). Simplest fix would be to say that imports from `latest` are not guaranteed to remain stable across HARDFORK version bumps?

SamWilsn:
Note that if we do decide to go forward with this, we'll need to:

- [ ] Add the file to [`setup.cfg`](https://github.com/ethereum/execution-specs/blob/master/setup.cfg).
- [ ] Add [a lint](https://github.com/ethereum/execution-specs/tree/master/src/ethereum_spec_tools/lint) that ensures the symlink is always pointing to the latest fork.
- [ ] Document the symlink in [`RELEASING.md`](https://github.com/ethereum/execution-specs/blob/master/RELEASING.md).

frangio:
> * Add [a lint](https://github.com/ethereum/execution-specs/tree/master/src/ethereum_spec_tools/lint) that ensures the symlink is always pointing to the latest fork.

What should I use as the source of truth for what is the latest fork?

SamWilsn:
[`Hardfork.discover`](https://github.com/ethereum/execution-specs/blob/78fb726158c69d8fa164e28f195fabf6ab59b915/src/ethereum_spec_tools/forks.py#L65) returns an ordered list.

Let me get a vibe check from @gurukamath and @petertdavies before you put in too much work. Personally, I don't have any objections offhand.

gurukamath:
> Let me get a vibe check from @gurukamath and @petertdavies before you put in too much work. Personally, I don't have any objections offhand.

With the understanding that the latest here means the most recent fork to be live on mainnet, I'm in favour of this with a lint in place.

SamWilsn:
Let's go ahead with this. If you're willing to work on it @frangio, by all means go ahead. Otherwise, we'll get around to it.

SamWilsn:
Unfortunately it seems like pip/setuptools don't support symlinks. It might be possible to emulate this with a post-deploy hook, but that wouldn't show up in the repository itself, so less useful.

If we ever come back to this, here's a lint:

```python
"""
Latest Symlink
^^^^^^^^^^^^^^

Ensures that the `latest` symlink points to the latest hardfork.
"""
from pathlib import Path
from typing import List, Sequence

from ethereum_spec_tools.forks import Hardfork
from ethereum_spec_tools.lint import Diagnostic, Lint


class LatestSymlink(Lint):
    """
    Ensures that the `latest` symlink poins to the latest hardfork.
    """

    def lint(
        self, forks: List[Hardfork], position: int
    ) -> Sequence[Diagnostic]:
        """
        Reads the path of the `latest` symlink and compares it to the latest
        hardfork.
        """
        if position != len(forks) - 1:
            return []

        fork = forks[position]
        assert fork.path is not None

        fork_path = Path(fork.path)
        symlink_path = fork_path.parent.joinpath("latest")

        # Use `readlink` instead of `resolve` because there should only ever be
        # one symlink in the chain.
        try:
            print(symlink_path)
            target_path = symlink_path.readlink()
        except FileNotFoundError:
            return [Diagnostic("The latest symlink is missing.")]

        if target_path.is_absolute():
            return [Diagnostic("The latest symlink is not relative.")]

        if len(target_path.parts) != 1:
            return [Diagnostic("The latest symlink is not relative to forks.")]

        target_path = fork_path.parent.joinpath(target_path)

        if not symlink_path.is_symlink():
            return [Diagnostic("The latest pointer is not a symlink.")]

        if target_path.is_symlink():
            return [Diagnostic("The latest symlink target is also a symlink.")]

        target_path = target_path.resolve()
        fork_path = fork_path.resolve()

        if target_path == fork_path:
            return []

        fork_path = fork_path.relative_to(fork_path.parent)
        target_path = target_path.relative_to(target_path.parent)
        return [
            Diagnostic(
                f"The latest symlink should point to `{fork_path}` "
                f"(but points to `{target_path}` instead)."
            )
        ]
```

frangio:
FWIW, my motivation for this PR was to have a quick way to navigate to the latest spec in the repo, on GitHub. I still think that would be valuable in some form. It could be via symlinks (even if not supported in import paths in Python), or it could be by adding direct links [in the readme](https://github.com/ethereum/execution-specs?tab=readme-ov-file#ethereum-protocol-releases).

Issue #1106, opened 2025-02-03
y_parity should be checked to be equal to 0 or 1


obatirou:
### Metadata

* **Hardfork**: All concerned fork with tx types different from legacy

### What was wrong?

When recovering sender, for tx type differrent from legacy, there is no check on the y_parity value which is a U256.
https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/cancun/transactions.py#L282-L293

The yellow specify that tx signature are valid if v is 0 or 1 but the usage of `secp256k1_recover` do not make the check either.

https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/crypto/elliptic_curve.py#L23

It will fail in coincurve but NOT when v is equal to 2 or 3
https://github.com/ofek/coincurve/blob/b66f4b7129d93582f758697df0eb9be7ba6981d2/src/coincurve/ecdsa.py#L51-L59

![Image](https://github.com/user-attachments/assets/ce475784-af72-417c-83e8-dd3317e8d3f8)

PR #1107, opened 2025-02-03
fix: check y_parity value

obatirou:
### What was wrong?
Related to Issue #1106 

### How was it fixed?
Add a check to ensure y_parity is 0 or 1

### DIFF
--- a/src/ethereum/arrow_glacier/transactions.py
+++ b/src/ethereum/arrow_glacier/transactions.py
@@ -241,10 +241,14 @@ def recover_sender(chain_id: U64, tx: Transaction) -> Address:
                 signing_hash_155(tx, chain_id),
             )
     elif isinstance(tx, AccessListTransaction):
+        if tx.y_parity not in (U256(0), U256(1)):
+            raise InvalidSignatureError("bad y_parity")
         public_key = secp256k1_recover(
             r, s, tx.y_parity, signing_hash_2930(tx)
         )
     elif isinstance(tx, FeeMarketTransaction):
+        if tx.y_parity not in (U256(0), U256(1)):
+            raise InvalidSignatureError("bad y_parity")
         public_key = secp256k1_recover(
             r, s, tx.y_parity, signing_hash_1559(tx)
         )
--- a/src/ethereum/berlin/transactions.py
+++ b/src/ethereum/berlin/transactions.py
@@ -213,6 +213,8 @@ def recover_sender(chain_id: U64, tx: Transaction) -> Address:
                 signing_hash_155(tx, chain_id),
             )
     elif isinstance(tx, AccessListTransaction):
+        if tx.y_parity not in (U256(0), U256(1)):
+            raise InvalidSignatureError("bad y_parity")
         public_key = secp256k1_recover(
             r, s, tx.y_parity, signing_hash_2930(tx)
         )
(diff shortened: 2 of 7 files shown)


src/ethereum/arrow_glacier/transactions.py
```diff
@@ -241,10 +241,14 @@ def recover_sender(chain_id: U64, tx: Transaction) -> Address:
                 signing_hash_155(tx, chain_id),
             )
     elif isinstance(tx, AccessListTransaction):
+        if tx.y_parity not in (U256(0), U256(1)):
+            raise InvalidSignatureError("bad y_parity")
```
SamWilsn, line 245: Are there cases where we'd want `secp256k1_recover` to accept `y_parity not in (0, 1)`?

petertdavies: No, it's a boolean. It gets encoded in strange ways in `LegacyTransaction`, but that is weirdness inherited from Bitcoin.

obatirou: From my understanding, no
Valid recovery id values are in (0, 1, 2, 3) (see [coincurve](https://github.com/ofek/coincurve/blob/b66f4b7129d93582f758697df0eb9be7ba6981d2/src/coincurve/ecdsa.py#L51-L59)) but ECDSA signatures are considered valid by the [yellow paper](https://ethereum.github.io/yellowpaper/paper.pdf) only when v is in (0, 1).
image

This is confirmed by the fact that even for erecover, the input of `secp256k1_recover` will be 0 or 1 see [ecrecover in exec spec](https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/cancun/vm/precompiled_contracts/ecrecover.py#L47-L55) and yellow paper

Capture d’écran 2025-02-04 à 16 46 40

Example with go-ethereum `ValidateSignatureValues`
https://github.com/ethereum/go-ethereum/blob/e6f3ce7b168b8f346de621a8f60d2fa57c2ebfb0/crypto/crypto.go#L274

and `Sign` function
https://github.com/ethereum/go-ethereum/blob/e6f3ce7b168b8f346de621a8f60d2fa57c2ebfb0/crypto/secp256k1/secp256.go#L68

petertdavies (approved):
LGTM

src/ethereum/arrow_glacier/transactions.py
SamWilsn: So should we move the check to `secp256k1_recover`?

obatirou: It is possible in this context but I think it would introduce a business need in `secp256k1_recover` which goal should only be to recover the public key from a given _valid_ signature imo. 
Shouldn't the validation of the signature happen before ?
The checks for r and s range are already outside the function https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/cancun/transactions.py#L260-L265 and https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/cancun/vm/precompiled_contracts/ecrecover.py#L47-L52

Let me know what you think. Open to make the changes if you prefer of course.

SamWilsn: I can certainly see the argument for keeping `secp256k1_recover` pure. That said, the `s` validation is not uniform across forks, so it cannot be included in `secp256k1_recover` directly. See

https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/frontier/transactions.py#L134-L135

vs.

https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/cancun/transactions.py#L263-L264

I suppose I'm happy either way!

PR #1109, opened 2025-02-04
Add tracing support for daemon

fselmo:
### What was wrong?

Related to ethereum/execution-spec-tests#1174

Tracing is not currently working via daemon when ``--traces`` is passed via EEST when filling tests.

### How was it fixed?

- If ``trace`` is passed in the POST, turn on full trace in the T8N tool.
- Validate that ``output-basedir`` is also specified as EEST relies on a known location for the trace files. Raise a clarifying 
error if it is not specified.

#### Cute Animal Picture

Screenshot 2025-02-04 at 08 43 22

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/daemon.py
+++ b/src/ethereum_spec_tools/evm_tools/daemon.py
@@ -59,6 +59,24 @@ class _EvmToolHandler(BaseHTTPRequestHandler):
             f"--state.reward={content['state']['reward']}",
         ]
 
+        trace = content.get("trace", False)
+        output_basedir = content.get("output-basedir")
+        if trace:
+            if not output_basedir:
+                raise ValueError(
+                    "`output-basedir` should be provided when `--trace` "
+                    "is enabled."
+                )
+            # send full trace output if ``trace`` is ``True``
+            args.extend(
+                [
+                    "--trace",
+                    "--trace.memory",
+                    "--trace.returndata",
+                    f"--output.basedir={output_basedir}",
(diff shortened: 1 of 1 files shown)


fselmo:
Ah, it looks like there's an `ethereum.rlp` issue if merging into `master` here. Which branch does EELS expect / which should I PR against?

danceratopz:
> Ah, it looks like there's an `ethereum.rlp` issue with EEST if merging into `master` here. Which branch does EEST expect / which should I PR against?
> 
> edit: I added a comment on the related EEST PR as well. Might be worth PRing the EEST changes to the branch with `ethereum.rlp` support.

This problem is on EEST's side, we'll take care of it there. More background here: https://github.com/ethereum/execution-spec-tests/pull/1174#issuecomment-2636848367

marioevz (commented):
Just a tiny comment, but overall works great for EEST.

src/ethereum_spec_tools/evm_tools/daemon.py
```diff
@@ -59,6 +59,24 @@ def do_POST(self) -> None:
             f"--state.reward={content['state']['reward']}",
         ]
 
+        trace = content.get("trace", False)
+        output_basedir = content.get("output-basedir")
+        if trace:
+            if output_basedir is None:
```
marioevz, line 65:
```suggestion
            if not output_basedir:
```
Maybe a bit better in case the value is empty string.

petertdavies (approved):

PR #1129, opened 2025-02-17
added address after checks

Shashwat-Nautiyal:
### What was wrong?

contract_address is being added to the accessed_addresses before the checks validating that the sender nonce is small enough, the sender has enough balance, and the call-stack-depth has not overflowed.

### How was it fixed?

Simply added contract_address to accessed_addresses after the checks.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/arrow_glacier/vm/instructions/system.py
+++ b/src/ethereum/arrow_glacier/vm/instructions/system.py
@@ -94,6 +94,8 @@ def generic_create(
         push(evm.stack, U256(0))
         return
 
+    evm.accessed_addresses.add(contract_address)
+
     if account_has_code_or_nonce(
         evm.message.block_env.state, contract_address
     ) or account_has_storage(evm.message.block_env.state, contract_address):
@@ -104,8 +106,6 @@ def generic_create(
         return
 
     increment_nonce(evm.message.block_env.state, evm.message.current_target)
-
-    evm.accessed_addresses.add(contract_address)
 
     child_message = Message(
         block_env=evm.message.block_env,
--- a/src/ethereum/arrow_glacier/vm/instructions/system.py
+++ b/src/ethereum/arrow_glacier/vm/instructions/system.py
@@ -75,7 +75,6 @@ def generic_create(
         evm.memory, memory_start_position, memory_size
     )
 
-    evm.accessed_addresses.add(contract_address)
 
     create_message_gas = max_message_call_gas(Uint(evm.gas_left))
     evm.gas_left -= create_message_gas
@@ -94,6 +93,7 @@ def generic_create(
         evm.gas_left += create_message_gas
         push(evm.stack, U256(0))
         return
+    
 
     if account_has_code_or_nonce(
         evm.message.block_env.state, contract_address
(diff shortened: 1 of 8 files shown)


gurukamath:
@marioevz Perhaps adding a test case each for `nonce`, `balance` and `stack_depth` checks?

gurukamath:
Although this makes intuitive sense to me, I am not sure if the EIPs specify this behaviour precisely. Another thing that we need to investigate is if the `CALL*` opcodes need to be looked into as well

Shashwat-Nautiyal:
@gurukamath , Am i supposed to add those test cases ?

gurukamath (changes requested):
Couple of points on this

- Similar to #1161, could you rebase this on `forks/prague` and create the request against that branch? You don't need to create a new PR. Just force push the updates to `Shashwat-Nautiyal:address_seq` and change the target in this PR from `ethereum:master` to `ethereum:forks/prague`
- Also make the relevant changes to `prague`

Shashwat-Nautiyal:
@gurukamath, rebased onto forks/prague, force-pushed the updates, and changed the PR target. Let me know if it's ok.

EIP_AUTHORS_MANUAL.md
```diff
@@ -65,4 +65,4 @@ Please refer the following tutorial for writing new EIP. It takes you through a
 
 In addition to having a reference implementation, it is also very useful for the community and core development if the EIP author concieves and writes test vectors for the EIP. There is a very user friendly framework for writing ethereum tests in the `execution-spec-tests` repository. Please refer to the following guide for writing tests to your EIP.
 
-[Writing Tests with `execution-spec-tests`](https://ethereum.github.io/execution-spec-tests/main/getting_started/quick_start/)
\ No newline at end of file
+[Writing Tests with `execution-spec-tests`](https://ethereum.github.io/execution-spec-tests/getting_started/quick_start/)
```
gurukamath, line 68: There are these changes that are not directly related to the issue at hand. These may be a result of the rebase not happening correctly. One simple way to get a clean PR is to create a feature branch from the latest `forks/prague` and `cherry-pick` the feature commit on top of it. That should take care of all these spurious changes.

Also, the changes to Prague need to be in there as well.

gurukamath (changes requested):

EIP_AUTHORS_MANUAL.md
Shashwat-Nautiyal: @gurukamath, I created a new feature branch and made all the changes in it. To avoid a new PR, I reset the commit history of the current branch and then force pushed the new commit of feature branch on top of it.

gurukamath (changes requested):
This already looks quite good! However, there seem to be some minor linting issues with static code checks. These should be easily fixable. You can find out the issues by running the command `tox -e static`. 

I would also suggest you to run `tox -e py3` locally to make sure all the tests are passing.

Shashwat-Nautiyal:
I ran tox -e py3 after the last commit and these are the test results. 
Screenshot from 2025-03-30 09-37-59

I am not sure about the 1301 xfails.

Shashwat-Nautiyal:
@gurukamath some of the checks were cancelled. Do i need to make any changes?

gurukamath (approved):

PR #1156, opened 2025-03-13
Refactored AccessListTransaction by hoisting class member into own class for readability

Carsons-Eels:
### What was wrong?

AccessListTransaction had a complex class member which was reducing the acceptable readability of related code.

Related to Issue #947 

### How was it fixed?

The class member was hoisted into its own class called `Access` which simplifies use in the rest of the code 

#### Cute Animal Picture

white baby fox

### DIFF
--- a/src/ethereum/arrow_glacier/transactions.py
+++ b/src/ethereum/arrow_glacier/transactions.py
@@ -46,6 +46,18 @@ class LegacyTransaction:
 
 @slotted_freezable
 @dataclass
+class Access:
+    """
+    A mapping from account address to storage slots that are pre-warmed as part
+    of a transaction.
+    """
+
+    account: Address
+    slots: Tuple[Bytes32, ...]
+
+
+@slotted_freezable
+@dataclass
 class AccessListTransaction:
     """
     The transaction type added in EIP-2930 to support access lists.
@@ -58,7 +70,7 @@ class AccessListTransaction:
     to: Union[Bytes0, Address]
     value: U256
     data: Bytes
-    access_list: Tuple[Tuple[Address, Tuple[Bytes32, ...]], ...]
+    access_list: Tuple[Access, ...]
     y_parity: U256
     r: U256
     s: U256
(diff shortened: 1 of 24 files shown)


src/ethereum/arrow_glacier/transactions.py
```diff
@@ -43,6 +43,16 @@ class LegacyTransaction:
     r: U256
     s: U256
 
+@slotted_freezable
+@dataclass
+class Access:
+    """
+    Hoisted from AccessListTransaction for readability
```
SamWilsn, line 51: Instead of documenting where this type came from, document what it is. Maybe:

```suggestion
    A mapping from account address to storage slots that are pre-warmed as part of a transaction.
```

src/ethereum/arrow_glacier/transactions.py
```diff
@@ -189,9 +200,10 @@ def calculate_intrinsic_cost(tx: Transaction) -> Uint:
 
     access_list_cost = 0
     if isinstance(tx, (AccessListTransaction, FeeMarketTransaction)):
-        for _address, keys in tx.access_list:
+        for access in tx.access_list:
             access_list_cost += TX_ACCESS_LIST_ADDRESS_COST
-            access_list_cost += len(keys) * TX_ACCESS_LIST_STORAGE_KEY_COST
+            for slot in access.slots:
```
SamWilsn, line 205: I think this'll add way too much gas cost. Instead of the for loop, what you want is just:

```python
access_list_cost += len(access.slots) * TX_ACCESS_LIST_STORAGE_KEY_COST
```

Carsons-Eels: Good point :star: 
Will amend :hammer_and_wrench:

SamWilsn (commented):
Looks good! Would you mind rebasing this on top of `forks/prague` and retargeting the pull request there as well?

SamWilsn (approved):

SamWilsn:
Ran the tests locally, they pass.

PR #1161, opened 2025-03-17
Update the type for `blob_gas_used`

### LINKED ISSUE #1125: `blob_gas_used` is a `U64` in some places and `Uint` in others
Reported by @fselmo:

This is one such case that I remembered along the way and was easy enough to track down:

 blob_gas_used is a [U64](https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/cancun/vm/gas.py#L272-L289) and also a [Uint](https://github.com/ethereum/execution-specs/blob/9923823367b5586228e590537d47aa9cc4c6a206/src/ethereum/cancun/fork.py#L565-L565) in other parts of the code.

It caused issues when trying to compare values or use them in addition or subtraction in some places. I'm not sure why only I saw these but math (e.g. <, +, - ) doesn't seem to work across different types for me. Led to me having to force a Uint to U64 somewhere in eth-tester (this part I don't remember now but could track down if needed).

---

We should make them the same type, or at least document why they aren't.

Shashwat-Nautiyal:
### What was wrong?
The blob_gas_used variable had inconsistent types across the repo. Somewhere U64 and somewhere Uint.

Related to Issue #1125 

### How was it fixed?
I updated the type of blob_gas_used to U64 across all files. Ensured to also update the return types of functions that populated the blob_gas_used variable.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/cancun/fork.py
+++ b/src/ethereum/cancun/fork.py
@@ -423,7 +423,7 @@ def check_transaction(
         if Uint(tx.max_fee_per_blob_gas) < blob_gas_price:
             raise InvalidBlock
 
-        max_gas_fee += calculate_total_blob_gas(tx) * Uint(
+        max_gas_fee += Uint(calculate_total_blob_gas(tx)) * Uint(
             tx.max_fee_per_blob_gas
         )
         blob_versioned_hashes = tx.blob_versioned_hashes
--- a/src/ethereum/cancun/fork.py
+++ b/src/ethereum/cancun/fork.py
@@ -76,7 +76,7 @@ BEACON_ROOTS_ADDRESS = hex_to_address(
     "0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02"
 )
 SYSTEM_TRANSACTION_GAS = Uint(30000000)
-MAX_BLOB_GAS_PER_BLOCK = Uint(786432)
+MAX_BLOB_GAS_PER_BLOCK = U64(786432)
 VERSIONED_HASH_VERSION_KZG = b"\x01"
 
 
@@ -350,7 +350,7 @@ def check_transaction(
     block_env: vm.BlockEnvironment,
     block_output: vm.BlockOutput,
     tx: Transaction,
-) -> Tuple[Address, Uint, Tuple[VersionedHash, ...], Uint]:
+) -> Tuple[Address, Uint, Tuple[VersionedHash, ...], U64]:
     """
     Check if the transaction is includable in the block.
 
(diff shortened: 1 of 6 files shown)


src/ethereum/arrow_glacier/fork.py
```diff
@@ -671,11 +671,8 @@ def pay_rewards(
 
 
 def process_transaction(
-    block_env: vm.BlockEnvironment,
-    block_output: vm.BlockOutput,
-    tx: Transaction,
-    index: Uint,
-) -> None:
+    env: vm.Environment, tx: Transaction
```
gurukamath, line 674: The `blob_gas_used` was introduced in Cancun and as such should only affect the `Cancun` and the `Prague` fork folders. The code in the earlier forks folders should not change unless we are back-porting something.

Shashwat-Nautiyal: umm, actually these changes are not made by me. Maybe they got included while I was rebasing the branch. Can you suggest what should i do ?

Shashwat-Nautiyal: @gurukamath Sorry but I am not exactly sure how to handle this. Should i create a new branch and cherry-pick only the feature commit or is there any other alternative?

gurukamath: That should work.

Shashwat-Nautiyal: @gurukamath, I cherry-picked the commits into a new feature branch and then force pushed it to the current branch,

src/ethereum/cancun/fork.py
```diff
@@ -587,23 +557,143 @@ def apply_body(
 
     Returns
     -------
-    block_output :
-        The block output for the current block.
+    apply_body_output : `ApplyBodyOutput`
```
gurukamath, line 560: I assume some of these changes are unintended and a result of the rebase not happening properly.

gurukamath:
There are some linting issues with static code checks. You can find out the issues by running the command tox -e static.

I would also suggest you to run tox -e py3 locally to make sure all the tests are passing.

gurukamath (changes requested):

src/ethereum/cancun/fork.py
Shashwat-Nautiyal: Made changes in a clean_feature_branch and pushed it to the current branch. 
Also cleared issues notified by `tox -e static`

Issue #1133, opened 2025-02-27
Simplify `validate_header` in post-merge forks


SamWilsn:
The `validate_header` function in pre-merge forks has to account for validating ommer headers. Since post-merge forks cannot have ommers, this function can be simplified to only check the last header in the chain.

> This is to ensure we have the same function to validate header for blocks as well a `ommers` in pre-merge forks. However, I am not against simplifying this logic for post-merge forks and simply assuming that the parent header is the last header in the chain.

_Originally posted by @gurukamath in https://github.com/ethereum/execution-specs/pull/1131#discussion_r1971348910_

anushkasomani:
Can I be assigned to this issue?

gurukamath:
Closed after #1163

PR #1163, opened 2025-03-19
Simplified validate_header func in post-merge forks

Antrikshgwal:
Related to Issue #1133

### How was it fixed?
Simplified the validate_header function in Cancun, shanghai and Paris forks in the forks.py file since they are the post-merge forks.
Added documentation in the function docstring clarifying that for post-merge forks, the parent header is always the last header in the chain.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/cancun/fork.py
+++ b/src/ethereum/cancun/fork.py
@@ -310,7 +310,7 @@ def validate_header(chain: BlockChain, header: Header) -> None:
     """
     if header.number < Uint(1):
         raise InvalidBlock
-    # Post-merge forks do not allow ommers, so parent_header is always the last header in the chain
+    
     parent_header = chain.blocks[-1].header
 
     excess_blob_gas = calculate_excess_blob_gas(parent_header)
@@ -337,6 +337,8 @@ def validate_header(chain: BlockChain, header: Header) -> None:
     if header.difficulty != 0:
         raise InvalidBlock
     if header.nonce != b"\x00\x00\x00\x00\x00\x00\x00\x00":
+        raise InvalidBlock
+    if header.ommers_hash != EMPTY_OMMER_HASH:
         raise InvalidBlock
 
     block_parent_hash = keccak256(rlp.encode(parent_header))
--- a/src/ethereum/cancun/fork.py
+++ b/src/ethereum/cancun/fork.py
@@ -349,6 +349,7 @@ def validate_header(chain: BlockChain, header: Header) -> None:
         raise InvalidBlock
     if header.nonce != b"\x00\x00\x00\x00\x00\x00\x00\x00":
         raise InvalidBlock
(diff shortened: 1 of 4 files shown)


src/ethereum/cancun/fork.py
```diff
@@ -349,6 +349,7 @@ def validate_header(chain: BlockChain, header: Header) -> None:
         raise InvalidBlock
     if header.nonce != b"\x00\x00\x00\x00\x00\x00\x00\x00":
         raise InvalidBlock
+     # Post-merge forks do not allow ommers, so parent_header is always the last header in the chain.
```
gurukamath, line 352: You need to simplify the logic at the beginning of the function (the one for determining the `parent_header`). Since there are no ommers, the only block whose header needs to be validated is the current block. Which implies that the `parent_header` is just the header of the last block of the chain.

gurukamath (changes requested):

src/ethereum/cancun/fork.py
```diff
@@ -310,19 +310,8 @@ def validate_header(chain: BlockChain, header: Header) -> None:
...
-    if (
-        parent_header_number < first_block_number
-        or parent_header_number > last_block_number
-    ):
-        raise InvalidBlock
-
-    parent_header = chain.blocks[
-        parent_header_number - first_block_number
-    ].header
+    # Post-merge forks do not allow ommers, so parent_header is always the last header in the chain
```
gurukamath, line 313:
```suggestion

```

src/ethereum/cancun/fork.py
```diff
@@ -349,8 +338,6 @@ def validate_header(chain: BlockChain, header: Header) -> None:
         raise InvalidBlock
     if header.nonce != b"\x00\x00\x00\x00\x00\x00\x00\x00":
         raise InvalidBlock
```
gurukamath, line 340: The ommers hash check should continue to be performed on post-merge forks

src/ethereum/cancun/fork.py
```diff
@@ -310,19 +310,8 @@ def validate_header(chain: BlockChain, header: Header) -> None:
...
-        parent_header_number < first_block_number
-        or parent_header_number > last_block_number
-    ):
-        raise InvalidBlock
-
-    parent_header = chain.blocks[
-        parent_header_number - first_block_number
-    ].header
+    # Post-merge forks do not allow ommers, so parent_header is always the last header in the chain
+    parent_header = chain.blocks[-1].header
```
gurukamath, line 315: You also need to implement the update to `Paris` and `Prague` which are also post-merge

gurukamath (changes requested):

src/ethereum/cancun/fork.py
Antrikshgwal: Done.

gurukamath (approved):

PR #1165, opened 2025-03-20
Type fixes for pyright

SamWilsn:
### What was wrong?

We're hitting a performance issue in mypy in #1038. This PR is the first of many fixing type errors caught by pyright that aren't caught by mypy.

### How was it fixed?

Slowly fixing type errors.

#### Cute Animal Picture


SamWilsn:
> Bit of an unfortunate discovery re: pyright. Turns out it _is_ smart enough to figure out those assertions, which means either: (a) there's an actual typing bug, or (b) pyright can only do that narrowing when all other errors are fixed. I think it means I need to do the whole switch in one PR...

Carsons-Eels (approved):

SamWilsn:
This is too out of date.

Issue #728, opened 2023-03-17
`download_fixtures` function does not check for the fixtures version


gurukamath:
See [download_fixtures](https://github.com/ethereum/execution-specs/blob/e7ac16f31b32798bb6b8e96e6fdaf89c1faae347/tests/conftest.py#L65)

Currently, the testing framework downloads the json fixtures as and when needed. There are two methods for this - 1. using git clone 2. download raw fixtures (using the `download_fixtures` function)

While the former checks out the appropriate version of the fixtures, the latter does not. The `download_fixtures` only checks if the fixtures folder is available. This could lead to issues if the fixture version has been changed or the downloaded fixtures have been corrupted in some way. 

At the same time, it might not be optimal to download the fixtures everytime, even when the correct fixtures are already present from a previous run. Some middle ground has to be worked out here.

SamWilsn:
Would it make sense to use something like [`requests-cache`](https://pypi.org/project/requests-cache/) for this? Assuming the HTTP server supports cache headers, that is.

SamWilsn:
Ah, I see that the version is in the URL. Probably makes sense to save the URL in a file alongside the extracted fixtures and only download if the requested URL doesn't match the saved one.

gurukamath:
> Ah, I see that the version is in the URL. Probably makes sense to save the URL in a file alongside the extracted fixtures and only download if the requested URL doesn't match the saved one.

Yeah. This is one way to do this.

Antrikshgwal:
> Ah, I see that the version is in the URL. Probably makes sense to save the URL in a file alongside the extracted fixtures and only download if the requested URL doesn't match the saved one.

Can I implement this? I'll modify the function to create a file that stores the initially downloaded URL and compare it with the requested URL on subsequent runs. If they match, the download will be skipped; otherwise, the fixtures will be redownloaded.

Antrikshgwal:
Hey @gurukamath. Could you please check the pr?

PR #1166, opened 2025-03-20
Implement check for fixtures version

Antrikshgwal:
### What was wrong?
download_fixtures method doesn't check for fixture version and re-downloads fixtures even when the same version already exists on the location. 

Related to Issue #728 

### How was it fixed?
Implemented a mechanism that stores the URL while downloading fixtures and compares it with the URL used on subsequent function calls. Since the URL changes on a version change, we can know if the version has been changed and if we need to re-download the fixtures. Otherwise, we can skip the download.

#### Cute Animal Picture


gurukamath:
The issue has been fixed in #1183

Issue #1139, opened 2025-03-06
Exclude glacier forks from code coverage


SamWilsn:
The `*_glacier` and `dao_fork` forks aren't tested specifically. Instead they are verified with the [`glacier_fork_hygiene`](https://github.com/ethereum/execution-specs/blob/master/src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py) lint.

They bring down our code coverage percentage unfairly, and so should be excluded from the report.

Carsons-Eels:
Closed by https://github.com/ethereum/execution-specs/pull/1168

Issue #1141, opened 2025-03-06
Enable branch code coverage tracking


SamWilsn:
https://coverage.readthedocs.io/en/latest/branch.html

PR #1168, opened 2025-03-21
Add codecov config file, excluded glacier & dao forks, add codecov badge to README

Carsons-Eels:
### What was wrong?

Code coverage unfairly included glacier forks and the dao fork, for which we will not create tests

Closes issues #1139 and #1141

### How was it fixed?

Excluded glacier forks in config file

#### Cute Animal Picture


README.md
```diff
@@ -1,6 +1,7 @@
 # Ethereum Execution Client Specifications
 
 [![GitPOAP Badge](https://public-api.gitpoap.io/v1/repo/ethereum/execution-specs/badge)](https://www.gitpoap.io/gh/ethereum/execution-specs)
+[![codecov](https://codecov.io/gh/ethereum/execution-specs/graph/badge.svg?token=${{ secrets.CODECOV_TOKEN }})](https://codecov.io/gh/ethereum/execution-specs)
```
SamWilsn, line 4: I don't think GitHub will substitute a secret here. Are you sure this is the token we're supposed to use?

codecov.yml
```diff
@@ -0,0 +1,2 @@
+ignore:
+  - "src/ethereum/*_glacier"  # ignore folders and all its contents
```
SamWilsn, line 2: Could you also include the dao fork? It's essentially a glacier fork and is covered by the same linting tool.

SamWilsn:
I realize this is still a draft and that you're probably aware, but just in case you haven't seen it yet, `arrow_glacier` is still included:

SamWilsn:
I hope you don't mind, but I took a quick look at [pytest-cov's documentation](https://pytest-cov.readthedocs.io/en/latest/config.html), and it looks like we can exclude directories earlier in the process. We could maybe add a section like this to `pyproject.toml` (and add the `--cov-config=pyproject.toml` option in `tox.ini`).

```toml
[tool.coverage.run]
omit = [
    "src/ethereum/*_glacier/*",
    "src/ethereum/dao_fork/*",
]
```

Carsons-Eels:
> I hope you don't mind, but I took a quick look at [pytest-cov's documentation](https://pytest-cov.readthedocs.io/en/latest/config.html), and it looks like we can exclude directories earlier in the process. We could maybe add a section like this to `pyproject.toml` (and add the `--cov-config=pyproject.toml` option in `tox.ini`).
> 
> ```toml
> [tool.coverage.run]
> omit = [
>     "src/ethereum/*_glacier/*",
>     "src/ethereum/dao_fork/*",
> ]
> ```

Not at all, this is good! I'll update the PR

SamWilsn:
For future PRs, prefer rebase over merge commits please!

SamWilsn (approved):

Issue #1176, opened 2025-04-02
Update Union syntax, move assert_type import,


Carsons-Eels:
# Update these

- `Union` is now a core type with a new syntax style
  - `var: Union[type, type2]` => `var: type | type2`
- `Optional` core type also uses now uses a similar syntax
  - `var: Optional[type, type2]` =>`var: Optional[type | type2]`
- `assert_type` has moved from `typing_extensions` lib to `typing`

Issue #1144, opened 2025-03-06
Upgrade minimum supported Python version to 3.11


SamWilsn:
New pypy landed, we should update to that.

Additionally, we should look at updating our code style (eg. `Union[Foo, Bar]` -> `Foo | Bar`).

Carsons-Eels:
These are some changes I will investigate from Python 3.11 that might affect us or are potentially interesting for improving code quality:

- [PEP 673 – Self Type](https://peps.python.org/pep-0673/)
- [PEP 654 – Exception Groups and except*](https://peps.python.org/pep-0654/)
- [PEP 678 – Enriching Exceptions with Notes](https://peps.python.org/pep-0678/)
- [`dataclasses`](https://docs.python.org/3/whatsnew/3.11.html#dataclasses)
- [`enum`](https://docs.python.org/3/whatsnew/3.11.html#enum)
- [`hashlib.file_digest`](https://docs.python.org/3/library/hashlib.html#hashlib.file_digest)
- [`logging.getLevelNamesMapping`](https://docs.python.org/3/library/logging.html#logging.getLevelNamesMapping)
- [`re` Atomic grouping and possessive quantifiers](https://docs.python.org/3/whatsnew/3.11.html#re)
- [`string`](https://docs.python.org/3/whatsnew/3.11.html#string)
- [`shutil.rmtree`](https://docs.python.org/3/library/shutil.html#shutil.rmtree)
- [`time.sleep`](https://docs.python.org/3/whatsnew/3.11.html#time)
- [`typing`](https://docs.python.org/3/whatsnew/3.11.html#typing)
- [Unpacking without parenthesis is now allowed: see `bpo-46725`](https://github.com/python/cpython/issues/90881https://github.com/python/cpython/issues/90881)
- [`-P` command line option and `PYTHONSAFEENV` environment variable](https://github.com/python/cpython/issues/57684)

This should maybe be moved to a separate issue, as this could greatly impact delivery time for this item if we consider it in scope. I'd also like feedback on the list to narrow it down if any of these are not considered worth pursuing.

SamWilsn:
Let's keep this issue for just updating the Python version and fixing any regressions. You can make issues for any other improvements you think we should do.

Carsons-Eels:
Version has been bumped, refactors need a review and to be merged.

Carsons-Eels:
Merged

PR #1177, opened 2025-04-02
Upgrade pypy/python version to 3.11

Carsons-Eels:
### What was wrong?

Pypy was out of date. It now handles python 3.11 so we should too.

Related to Issue #1144

### How was it fixed?

Updated a bunch of config files, the README, etc

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #1179, opened 2025-04-03
1176 inital 311 refactors

Carsons-Eels:
### What was wrong?

New pypy version was released supporting python 3.11. Basic upgrades to code were needed.

Related to Issue #1176 #1144

### How was it fixed?

Refactored
- Use primitive `Union` type introduced by python `3.10` rather than `typing.Union`
- Import `assert_type` from typing rather than `typing_extensions`

#### Cute Animal Picture


Carsons-Eels:
I included the commit from PR #1177 because otherwise you can't test it locally as it requires the upgrade to py/py 3.11 to function correctly. That commit can either be dropped pre-approval, or this commit can just be merged **_immediately_** after 1177.

src/ethereum/crypto/finite_field.py
SamWilsn, line 1: Where is this coming from?

PR #1180, opened 2025-04-04
Implement EIP-7873 and EIP-5920

### LINKED ISSUE #1173: Implement EIP-7873
[EIP-7873](https://eips.ethereum.org/EIPS/eip-7873)

### LINKED ISSUE #1181: Implement EIP-5920
[EIP-5920](https://eips.ethereum.org/EIPS/eip-5920)

gurukamath:
(closes #1173 )
(closes #1181 )

### What was wrong?
The EOF implementation lacks the EIP-7673 (TX Create) and EIP-5920 (PAY Opcode)

Related to Issue #1173 
Related to Issue #1181 

### How was it fixed?
Implement the relevant EIPs.

### DIFF
--- a/src/ethereum/osaka/transactions.py
+++ b/src/ethereum/osaka/transactions.py
@@ -19,11 +19,13 @@ from .exceptions import TransactionTypeError
 from .fork_types import Address, Authorization, VersionedHash
 
 TX_BASE_COST = Uint(21000)
-FLOOR_CALLDATA_COST = Uint(10)
-STANDARD_CALLDATA_TOKEN_COST = Uint(4)
+TOTAL_COST_FLOOR_PER_TOKEN = Uint(10)
+STANDARD_TOKEN_COST = Uint(4)
 TX_CREATE_COST = Uint(32000)
 TX_ACCESS_LIST_ADDRESS_COST = Uint(2400)
 TX_ACCESS_LIST_STORAGE_KEY_COST = Uint(1900)
+
+MAX_INIT_CODE_COUNT = 256
 
 
 @slotted_freezable
@@ -130,12 +132,35 @@ class SetCodeTransaction:
     s: U256
 
 
+@slotted_freezable
+@dataclass
+class EofInitCodeTransaction:
+    """
+    The transaction type added in EIP-7873.
+    """
+
+    chain_id: U64
+    nonce: U256
+    max_priority_fee_per_gas: Uint
+    max_fee_per_gas: Uint
+    gas: Uint
+    to: Union[Bytes0, Address]
+    value: U256
+    data: Bytes
+    access_list: Tuple[Tuple[Address, Tuple[Bytes32, ...]], ...]
+    init_codes: Tuple[Bytes, ...]
+    y_parity: U256
+    r: U256
+    s: U256
+
+
 Transaction = Union[
     LegacyTransaction,
     AccessListTransaction,
     FeeMarketTransaction,
     BlobTransaction,
     SetCodeTransaction,
+    EofInitCodeTransaction,
 ]
 
 
@@ -153,6 +178,8 @@ def encode_transaction(tx: Transaction) -> Union[LegacyTransaction, Bytes]:
         return b"\x03" + rlp.encode(tx)
     elif isinstance(tx, SetCodeTransaction):
         return b"\x04" + rlp.encode(tx)
+    elif isinstance(tx, EofInitCodeTransaction):
+        return b"\x05" + rlp.encode(tx)
     else:
         raise Exception(f"Unable to encode transaction of type {type(tx)}")
 
@@ -170,6 +197,8 @@ def decode_transaction(tx: Union[LegacyTransaction, Bytes]) -> Transaction:
             return rlp.decode_to(BlobTransaction, tx[1:])
         elif tx[0] == 4:
             return rlp.decode_to(SetCodeTransaction, tx[1:])
+        elif tx[0] == 5:
+            return rlp.decode_to(EofInitCodeTransaction, tx[1:])
         else:
             raise TransactionTypeError(tx[0])
     else:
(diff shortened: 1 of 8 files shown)


Carsons-Eels (commented):
Will review 7873 tomorrow after I familiarize myself with the EIP and better understand it.

src/ethereum/osaka/transactions.py
```diff
@@ -130,12 +132,35 @@ class SetCodeTransaction:
...
+
+    chain_id: U64
+    nonce: U256
+    max_priority_fee_per_gas: Uint
+    max_fee_per_gas: Uint
+    gas: Uint
+    to: Union[Bytes0, Address]
+    value: U256
+    data: Bytes
+    access_list: Tuple[Tuple[Address, Tuple[Bytes32, ...]], ...]
```
Carsons-Eels, line 150: Just a heads up, this will end up needing to be refactored from `Tuple[Address, Tuple[Bytes32, ...]]` to `Access` after Osaka merges the upstream changes. Just noting this, since this branch is behind upstream changes to master/prague that include this.

gurukamath: Thanks. Yeah. There will be a larger re-basing effort on all our EOF related branches soon

src/ethereum/osaka/transactions.py
```diff
@@ -250,19 +307,27 @@ def calculate_intrinsic_cost(tx: Transaction) -> Tuple[Uint, Uint]:
...
+
+    tokens_in_tx += calculate_tokens_in_data(tx.data)
+
+    if isinstance(tx, EofInitCodeTransaction):
+        for init_code in tx.init_codes:
+            if len(init_code) == 0:
+                raise InvalidTransaction(
+                    "Type 5 tx with zero-length init code"
+                )
+            if len(init_code) > 2 * MAX_CODE_SIZE:
```
Carsons-Eels, line 322: I think there should either be a comment referencing the EIP definition for `MAX_INITCODE_SIZE`, or `MAX_INITCODE_SIZE` should just be defined in the same place as MAX_CODE_SIZE and referenced from there. IMO this would be more readable that way.

Carsons-Eels (approved):

src/ethereum/osaka/vm/instructions/system.py
```diff
@@ -1271,3 +1379,58 @@ def return_contract(evm: Evm) -> None:
...
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+    """
+    # STACK
+    try:
+        to = to_address_without_mask(pop(evm.stack))
+    except ValueError as e:
+        raise ExceptionalHalt from e
```
jochem-brouwer, line 1397: This error cannot happen in EVM. The `to_address_without_mask` will only throw if the value is larger than an uint256 and pushing such value is not possible in EVM. I think what was intended to do here is to check for the `bytes20`/`uint160` case, so it should be verified that the stack value fits in an `uint160` or `bytes20`. If this is not the case then throw (EVM should internally run like the INVALID opcode)

src/ethereum/osaka/vm/instructions/system.py
```diff
@@ -1271,3 +1379,58 @@ def return_contract(evm: Evm) -> None:
...
+    # OPERATION
+    if value > U256(0):
+        try:
+            move_ether(
+                evm.message.block_env.state,
+                evm.message.current_target,
+                to,
+                value,
+            )
+            push(evm.stack, U256(1))
```
jochem-brouwer, line 1426: Items being pushed to stack is not part of current spec (April 15) https://github.com/ethereum/EIPs/blob/5fae72b226be42a043e75c682d13d3eb889da17b/EIPS/eip-5920.md

An item being pushed is part of this proposed PR for the EIP: https://github.com/ethereum/EIPs/pull/9590

Note: it pushes 0 to stack if it is successful, and nonzero (1) in case it fails (so the other way around as of here). Motivation is in the EIP but this allows to add extra failure error codes in the future.

src/ethereum/osaka/vm/instructions/system.py
```diff
@@ -1271,3 +1379,58 @@ def return_contract(evm: Evm) -> None:
 
     # PROGRAM COUNTER
     evm.running = False
+
+
+def pay(evm: Evm) -> None:
```
jochem-brouwer, line 1384: This is not part of current spec (April 15) https://github.com/ethereum/EIPs/blob/5fae72b226be42a043e75c682d13d3eb889da17b/EIPS/eip-5920.md but this should be rejected in the STATIC frame (when previously STATICCALL has been used in the call stack). Part of https://github.com/ethereum/EIPs/pull/9590

jochem-brouwer (changes requested):
Hi @gurukamath, I want to SFI the PAY EIP for Fusaka, but by the process https://eips.ethereum.org/EIPS/eip-7723#scheduled-for-inclusion we need it both implemented here and we need tests in `execution-spec-tests` (done here: https://github.com/ethereum/execution-spec-tests/issues/1509). Great to see you already started on this! I have only checked the PAY opcode (and would also suggest to split this in an individual PR for PAY, and keep the TXCREATE stuff also in an individual PR)

I also see you added stack items here, which are not specced yet, but I have them in this PR for the EIP spec: https://github.com/ethereum/EIPs/pull/9590. Would also love if you could take a look there and leave some feedback :smile: :+1:

src/ethereum/osaka/vm/instructions/system.py
gurukamath: I think the name given to the variable is slightly misleading `MAX_ADDRESS_U256` but it is in fact checking for address that have more than 20 bytes. `b"\xff" * 20`. I will change the variable name to something more appropriate. Thanks for pointing this out.

gurukamath:
@jochem-brouwer Thanks for taking the time to look at this. We don't yet have an "official" Osaka branch since we are working on a couple of small re-factors before freezing the Pectra code. Once the osaka branch is up, I will rebase this commit on there and make the necessary adjustments as per the latest spec. Will keep an eye on ethereum/EIPs#9590

src/ethereum/osaka/vm/instructions/system.py
jochem-brouwer: The PR got changed and it now pushes 1 to stack on success and 0 on failure! So, we can say that EELS was ahead of the spec :smile: :+1: The PR got merged.

gurukamath:
Closing since EOF is not a part of Osaka. Raised a separate [PR](#1225 ) for EIP-5920

PR #1182, opened 2025-04-04
t8n: Return `InvalidBlock` exception in `result.blockException`

marioevz:
### What was wrong?

Block-level exceptions, such the ones raised when the system contract calls failed, resulted in t8n returning an empty response instead of parseable JSON.

### How was it fixed?

Added a new field in the T8N's response's field `result` that is called `blockException`, which is populated in case of an error.

In this PR only an exception in `process_general_purpose_requests` will be caught and populate the extra field, so every other exception is not caught and will still result in the same behavior, but I think each case should be analyzed separately.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
@@ -199,58 +199,60 @@ class T8N(Load):
         self.result.update(self, block_env, block_output)
         self.result.rejected = self.txs.rejected_txs
 
+    def _run_blockchain_test(self, block_env: Any, block_output: Any) -> None:
+        if self.fork.is_after_fork("ethereum.prague"):
+            self.fork.process_system_transaction(
+                block_env=block_env,
+                target_address=self.fork.HISTORY_STORAGE_ADDRESS,
+                data=block_env.block_hashes[-1],  # The parent hash
+            )
+
+        if self.fork.is_after_fork("ethereum.cancun"):
+            self.fork.process_system_transaction(
+                block_env=block_env,
+                target_address=self.fork.BEACON_ROOTS_ADDRESS,
+                data=block_env.parent_beacon_block_root,
+            )
+
+        for i, tx in zip(self.txs.successfully_parsed, self.txs.transactions):
+            self.backup_state()
+            try:
+                self.fork.process_transaction(
+                    block_env, block_output, tx, Uint(i)
+                )
+            except EthereumException as e:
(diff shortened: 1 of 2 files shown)


src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -442,21 +442,25 @@ def apply_body(self) -> None:
...
+                        self.env.block_gas_limit,
+                        self.env.block_timestamp,
+                        self.env.prev_randao,
+                        self.chain_id,
+                        self.env.excess_blob_gas,
+                    )
                 )
-            )
+            except InvalidBlock as e:
+                self.result.block_exception = f"{e}"
```
SamWilsn, line 463: Should we hoist this up into:

https://github.com/ethereum/execution-specs/blob/8707c283cf21154bd7b3fba2f42f186270d97927/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L202

?

We can wrap the whole function body in a `try`/`except`, since per-transaction exceptions are caught [here](https://github.com/ethereum/execution-specs/blob/8707c283cf21154bd7b3fba2f42f186270d97927/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L229).

marioevz: Yes that sounds better. Is `run_blockchain_test` part of `forks/prague`? I cannot see it in this version of the file.

SamWilsn: I believe so, yeah.

marioevz:
@SamWilsn I've rebased to `forks/prague` and applied your suggestion.

Let me know if it's ok.

marioevz:
@SamWilsn I've rebased to forks/prague, applied your suggestion, fixed my code and now it's working finally.

Let me know if it's ok.

marioevz:
Just as a sanity check, I did a local check of the fixtures generated with:
- This fix + #1191 + https://github.com/ethereum/execution-spec-tests/pull/1454
- Only #1191 + https://github.com/ethereum/execution-spec-tests/pull/1454

And the resulting generated test fixtures were identical.

gurukamath (approved):
LGTM! The only nit would perhaps be to avoid the huge try-except block in `run_blockchain_test`. Defining the logic in the try block as a separate function.

PS: Never mind. Perhaps that was the original implementation.

PR #1183, opened 2025-04-04
Invalid block if system contract is empty on call or call fails

marioevz:
### What was wrong?

EIPs 7002 and 7251 have been updated to define the expected outcome in case of execution failure or code absence:

https://github.com/ethereum/EIPs/pull/9508
https://github.com/ethereum/EIPs/pull/9582

### How was it fixed?

Conditions for zero-length code and execution failure have been added to the system contract calls.

I'm not sure if this change affects EIP-2935, although I think we should not cherry-pick which system-contracts invalidate a block and which don't.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/cancun/fork.py
+++ b/src/ethereum/cancun/fork.py
@@ -482,10 +482,14 @@ def make_receipt(
 def process_system_transaction(
     block_env: vm.BlockEnvironment,
     target_address: Address,
+    system_contract_code: Bytes,
     data: Bytes,
 ) -> MessageCallOutput:
     """
-    Process a system transaction.
+    Process a system transaction with the given code.
+
+    Prefer calling `process_unchecked_system_transaction` unless the contract
+    code has already been read from the state.
 
     Parameters
     ----------
@@ -493,6 +497,8 @@ def process_system_transaction(
         The block scoped environment.
     target_address :
         Address of the contract to call.
+    system_contract_code :
+        Code of the contract to call.
     data :
         Data to pass to the contract.
 
@@ -501,8 +507,6 @@ def process_system_transaction(
     system_tx_output : `MessageCallOutput`
         Output of processing the system transaction.
     """
-    system_contract_code = get_account(block_env.state, target_address).code
-
     tx_env = vm.TransactionEnvironment(
         origin=SYSTEM_ADDRESS,
         gas_price=block_env.base_fee_per_gas,
@@ -540,6 +544,38 @@ def process_system_transaction(
     return system_tx_output
 
 
+def process_unchecked_system_transaction(
+    block_env: vm.BlockEnvironment,
+    target_address: Address,
+    data: Bytes,
+) -> MessageCallOutput:
+    """
+    Process a system transaction without checking if the contract contains code
+    or if the transaction fails.
+
+    Parameters
+    ----------
+    block_env :
+        The block scoped environment.
+    target_address :
+        Address of the contract to call.
+    data :
+        Data to pass to the contract.
+
+    Returns
+    -------
+    system_tx_output : `MessageCallOutput`
+        Output of processing the system transaction.
+    """
+    system_contract_code = get_account(block_env.state, target_address).code
+    return process_system_transaction(
+        block_env,
+        target_address,
+        system_contract_code,
+        data,
+    )
+
+
 def apply_body(
     block_env: vm.BlockEnvironment,
     transactions: Tuple[Union[LegacyTransaction, Bytes], ...],
@@ -571,7 +607,7 @@ def apply_body(
     """
     block_output = vm.BlockOutput()
 
-    process_system_transaction(
+    process_unchecked_system_transaction(
         block_env=block_env,
         target_address=BEACON_ROOTS_ADDRESS,
         data=block_env.parent_beacon_block_root,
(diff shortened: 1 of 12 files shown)


gurukamath:
Consider re-directing this PR at the `forks/prague` branch instead. That has the latest spec updates including the latest re-factors from Prague.

marioevz:
Might need to take another look at this since it might modify behavior of prior system contracts, but according to this message https://discord.com/channels/595666850260713488/1351309992837251195/1361391543595962669, that should not be the case.

src/ethereum/cancun/fork.py
```diff
@@ -484,6 +484,8 @@ def process_system_transaction(
     block_env: vm.BlockEnvironment,
     target_address: Address,
     data: Bytes,
+    raise_on_empty_code: bool,
+    raise_on_error: bool,
```
SamWilsn, line 488: Something feels really weird about this, but I can't quite put my finger on it.

If we were going heavily into traditional exceptions, I'd expect to see a pattern like:

```python
try:
    process_system_transaction(block_env, target_address, data)
except SystemContractEmpty:
    pass
except SystemContractRevert:
    pass
```

But we've taken the approach of minimizing our use of exceptions as much as possible to keep the code flow extremely readable.

Conditionally throwing exceptions seems to violate traditional exception patterns and our own minimal exception style.

Instead, what if we did something like:

```python
def process_checked_system_transaction(
    block_env: vm.BlockEnvironment,
    target_address: Address,
    data: Bytes,
) -> MessageCallOutput:
    system_contract_code = get_account(block_env.state, target_address).code

    if len(system_contract_code) == 0:
        raise InvalidBlock(
            f"System contract address {target_address.hex()} does not contain code"
        )

    system_tx_output = process_system_transaction(
        block_env,
        target_address,
        system_contract_code,
        data,
    )

    if system_tx_output.error:
        raise InvalidBlock(
            f"System contract ({target_address.hex()}) call failed: "
            f"{system_tx_output.error}"
        )

def process_unchecked_system_transaction(
    block_env: vm.BlockEnvironment,
    target_address: Address,
    data: Bytes,
) -> MessageCallOutput:
    system_contract_code = get_account(block_env.state, target_address).code
    return process_system_transaction(
        block_env,
        target_address,
        system_contract_code,
        data,
    )
````

Comment (marioevz) [reply src/ethereum/cancun/fork.py:488]: I like it! thanks for the suggestion, I'll work on the modifications 👍

**Comment by marioevz** [comment]
@SamWilsn applied the changes, let me know what you think.

**Comment by SamWilsn** [comment]
It's possible I broke something while rebasing, but I'm getting this error in the tests:

```
ethereum.exceptions.InvalidBlock: System contract address 00000961ef480eb55e80d19ad83579a64c007002 does not contain code
```

From test:

```
tests/prague/test_state_transition.py::test_ethereum_tests[stEIP1153-transientStorage - 10_revertUndoesStoreAfterReturn_d0g0v0_Prague]
```

**Comment by marioevz** [comment]
> It's possible I broke something while rebasing, but I'm getting this error in the tests:
> 
> ```
> ethereum.exceptions.InvalidBlock: System contract address 00000961ef480eb55e80d19ad83579a64c007002 does not contain code
> ```
> 
> From test:
> 
> ```
> tests/prague/test_state_transition.py::test_ethereum_tests[stEIP1153-transientStorage - 10_revertUndoesStoreAfterReturn_d0g0v0_Prague]
> ```

Interesting, perhaps we broke state tests parsing/running with this change? Because I think this check should be skipped on state tests.

**Comment by marioevz** [comment]
@SamWilsn wait, I'm confused, is this failure happening on the execution of a state test or blockchain test?

If it's on a blockchain test then it's expected because ethereum/tests I believe don't add the Prague system contracts to the blockchain tests.

**Comment by SamWilsn** [comment]
I would guess this is a blockchain test:

https://github.com/ethereum/execution-specs/blob/b1edd6b65f88ddf10bb39209ea204bfa46f4055e/tests/prague/test_state_transition.py#L87-L94

File: tests/helpers/__init__.py [hunk 2]
```diff
@@ -1,19 +1,30 @@
...
+TEST_FIXTURES: Dict[str, _FixtureSource] = {
     "evm_tools_testdata": {
         "url": "https://github.com/gurukamath/evm-tools-testdata.git",
         "commit_hash": "792422d",
         "fixture_path": "tests/fixtures/evm_tools_testdata",
     },
     "ethereum_tests": {
         "url": "https://github.com/ethereum/tests.git",
-        "commit_hash": "afed83b",
+        "commit_hash": "07f85aafbf2966b4db2b7d9cdf55248f91661ef9",
```
gurukamath, line 23: `tox -e py3` ends up throwing `stderr: 'fatal: reference is not a tree: 07f85aafbf2966b4db2b7d9cdf55248f91661ef9'` on my local

src/ethereum/prague/fork.py
```diff
@@ -573,6 +573,10 @@ def process_system_transaction(
         Output of processing the system transaction.
     """
     system_contract_code = get_account(state, target_address).code
+    if len(system_contract_code) == 0:
```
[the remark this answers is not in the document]
SamWilsn: This is taken care of with `process_unchecked_system_transaction`

gurukamath (approved):

PR #1185, opened 2025-04-07
Clear specific json files

### LINKED ISSUE #1178: T8N daemon clears *.json files in current working directory
I was trying to implement https://github.com/ethereum/EIPs/pull/9508, https://github.com/ethereum/EIPs/pull/9582 and my [eels_resolutions.json](https://github.com/ethereum/execution-spec-tests/blob/main/eels_resolutions.json) kept getting deleted when filling Prague tests and pointing to branch `forks/prague`.

Turns out that when running `uv run fill` the default is to not print traces, and therefore the `output-basedir` field is not passed in the request, so in this line:
https://github.com/ethereum/execution-specs/blob/75635cdfb18ef2d1294ecdf1bfd452be2a687571/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L255
`output_basedir` resolves to `'.'` and clears out the current working directory out of json files, which when running `uv run fill` clears `./eels_resolutions.json` in the EEST folder.

One possible alternative could be that `output_basedir` is `None` by default, and not clear or write any files in that case, but I have not explored the ramifications of this change so there could be some side effects.

Also following commit is necessary in `forks/prague`:
https://github.com/ethereum/execution-specs/commit/7f0e7514329cb8d604ee5ef2c1acce1521935c7a
otherwise `output-basedir` field from the request is never processed.

gurukamath:
(closes #1178 )

### What was wrong?
The `t8n` tool would indiscriminately clear all the json files in the `output.basedir` folder during its run in order to avoid any confusion with output files that my have been created in a previous run with the same parameters. This leads in some specific cases to some config files being deleted.

Related to Issue #1178 

### How was it fixed?

1. Port the commit that updates the daemon from `master` to `forks/prague`
2. Clear files selectively

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/daemon.py
+++ b/src/ethereum_spec_tools/evm_tools/daemon.py
@@ -59,6 +59,24 @@ class _EvmToolHandler(BaseHTTPRequestHandler):
             f"--state.reward={content['state']['reward']}",
         ]
 
+        trace = content.get("trace", False)
+        output_basedir = content.get("output-basedir")
+        if trace:
+            if not output_basedir:
+                raise ValueError(
+                    "`output-basedir` should be provided when `--trace` "
+                    "is enabled."
+                )
+            # send full trace output if ``trace`` is ``True``
+            args.extend(
+                [
+                    "--trace",
+                    "--trace.memory",
+                    "--trace.returndata",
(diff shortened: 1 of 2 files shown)


gurukamath:
@marioevz Do let me know if this takes care of the issue.

PR #1187, opened 2025-04-08
Use py_ecc for alt_bn128

### LINKED ISSUE #1169: bug: multiplicative inverse
### Metadata

* **Hardfork**: cancun

### What was wrong?

The multiplicative inverse algorithm seems to be flawed.

For some non-zero elements of `BNF2`, `multiplicative_inverse()` returns `(0, 0)` (i.e. `BNF2.zero()`) which shouldn't be the case as `BNF2` is a finite field. Note that the arkworks implementation yields the correct inverse.

### Sources

Arkworks `ark_curve` [BNF2 implementation](https://github.com/arkworks-rs/algebra/blob/cae69c5ad206d69f0dc5abef0c93865bd21f2b6f/curves/bn254/src/fields/fq2.rs)
Arkworks `ark_ff` implementation of the [field inversion in the quadratric extension field](https://github.com/arkworks-rs/algebra/blob/cae69c5ad206d69f0dc5abef0c93865bd21f2b6f/ff/src/fields/models/quadratic_extension.rs#L325)
Algorithm source: [Guide to pairing-based cryptography](https://www.math.u-bordeaux.fr/~damienrobert/csi/book/book.pdf) 5.19 p.135

### Additional Context

**Minimal Working Example**

```python
from ethereum.crypto.alt_bn128 import BNF2

if __name__ == "__main__":
    val = BNF2.from_int(16)
    val_inv = val.multiplicative_inverse()
    print(f"Multiplicative inverse of {val}: {val_inv}")
    assert val * val_inv == BNF2.from_int(1)
```

It also fails for `val = BNF2.from_int(242)`.

### LINKED ISSUE #1170: Switch to `py_ecc` for the alt_bn128 precompiles
The alt_bn128 precompiles are implemented using custom EC crypto code `ethereum.crypto`. There is no good reason to keep maintaining this code instead of using the implementation in `py_ecc`.

Given that a bug (#1169) has been reported in `finite_field.py`, which may cause the pairing precompile to fail on specially chosen inputs, we should switch to `py_ecc` now.

gurukamath:
(closes #1170 )
(closes #1169 )

### What was wrong?

The alt_bn128 precompile uses custom EC code from `ethereum.crypto`. We can just use `py_ecc` instead.

Related to Issue #1170 

### How was it fixed?
Use `py_ecc` for alt_bn128.


petertdavies (approved):
Looks good! I pushed a commit to remove some more of the unused code.

PR #1190, opened 2025-04-11
dev: simplify condition in is_account_alive

enitrat:
### What was wrong?

The condition for `is_account_alive` is strictly equivalent to `== EMPTY_ACCOUNT`. Thus, for simplicity, I modified it (and ensuring that if the Account model came to change, this would be properly reflected in this function.

### DIFF
--- a/src/ethereum/arrow_glacier/state.py
+++ b/src/ethereum/arrow_glacier/state.py
@@ -463,11 +463,7 @@ def is_account_alive(state: State, address: Address) -> bool:
     if account is None:
         return False
     else:
-        return not (
-            account.nonce == Uint(0)
-            and account.code == b""
-            and account.balance == 0
-        )
+        return not account == EMPTY_ACCOUNT
 
 
 def modify_state(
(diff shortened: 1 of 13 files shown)


enitrat:
hi @SamWilsn, @gurukamath, I made a small change for clarity, let me know how that sounds!

src/ethereum/arrow_glacier/state.py
```diff
@@ -463,11 +463,7 @@ def is_account_alive(state: State, address: Address) -> bool:
     if account is None:
         return False
     else:
-        return not (
-            account.nonce == Uint(0)
-            and account.code == b""
-            and account.balance == 0
-        )
+        return not account == EMPTY_ACCOUNT
```
SamWilsn, line 466: Why not:

```suggestion
        return account != EMPTY_ACCOUNT
```

src/ethereum/arrow_glacier/state.py
```diff
@@ -463,11 +463,7 @@ def is_account_alive(state: State, address: Address) -> bool:
     if account is None:
```
SamWilsn, line 463: Or even:

```python
    return account is not None and account != EMPTY_ACCOUNT
```

enitrat: yeah definitely better this way :)

SamWilsn (approved):

PR #1191, opened 2025-04-14
EOAs that delegate to a precompile

gurukamath:
### What was wrong?
Currently, the EOA account that delegate to a precompile in 7702 are not handled properly. This PR also updates to 4.2.0 version of the EEST tests

### How was it fixed?
Execute empty code when an EOA delegates to a pre-compile

### DIFF
--- a/src/ethereum/prague/vm/__init__.py
+++ b/src/ethereum/prague/vm/__init__.py
@@ -131,7 +131,7 @@ class Message:
     is_static: bool
     accessed_addresses: Set[Address]
     accessed_storage_keys: Set[Tuple[Address, Bytes32]]
-    is_delegated: bool
+    disable_precompiles: bool
     parent_evm: Optional["Evm"]
 
 
--- a/src/ethereum/prague/vm/__init__.py
+++ b/src/ethereum/prague/vm/__init__.py
@@ -131,6 +131,7 @@ class Message:
     is_static: bool
     accessed_addresses: Set[Address]
     accessed_storage_keys: Set[Tuple[Address, Bytes32]]
+    is_delegated: bool
     parent_evm: Optional["Evm"]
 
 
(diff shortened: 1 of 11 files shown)


danceratopz:
There seems to be a problematic commit in `forks/prague`: 8f89be57c9f1c764d1fa5b95eb32db0eac7fd3d6

If I revert this commit, `7702-to-precompile` fills all tests on EEST main a1c4eeff34.

gurukamath:
> There seems to be a problematic commit in `forks/prague`: [8f89be5](https://github.com/ethereum/execution-specs/commit/8f89be57c9f1c764d1fa5b95eb32db0eac7fd3d6)
> 
> If I revert this commit, `7702-to-precompile` fills all tests on EEST main a1c4eeff34.

The problem here seems to be the fact that all other `t8n` tools expect the variable `accessList` and eels-t8n expects `accessLists`. The state test fixtures also, provide `accessLists`.

So in the long run it might be good to align on one variable name. But in the short term, in the interest of avoiding changes to 5 different client t8n tools, eels-t8n can also start expecting `accessList`. I will push a commit with the hot fix in this PR and create an issue to track this in the long term.

gurukamath:
Created #1194

danceratopz:
Thanks @gurukamath, this fills with EEST main a1c4eeff34!

src/ethereum/prague/vm/__init__.py
```diff
@@ -131,6 +131,7 @@ class Message:
     is_static: bool
     accessed_addresses: Set[Address]
     accessed_storage_keys: Set[Tuple[Address, Bytes32]]
+    is_delegated: bool
```
SamWilsn, line 134: The name of this variable is ambiguous. Is it for `DELEGATECALL` or EIP-7702 or both? Perhaps something like `precompile_behavior: PrecompileBehavior` where:

```python
class PrecompileBehavior(Enum):
    STOP_IMMEDIATELY
    EXECUTE
```

(names to be bikeshedded)

gurukamath: Updated. Changed the variable name to `disable_precompiles`

SamWilsn (approved):
Confirmed tests pass, merge when ready.

Issue #909, opened 2024-03-20
Emit trace events as they happen instead of storing them


SamWilsn:
As recommended by @holiman on https://github.com/holiman/goevmlab/pull/131#issuecomment-2009562191 and @MariusVanDerWijden on discord, we shouldn't be storing EVM trace events longer than necessary.

For example, we'd need to change this:

https://github.com/ethereum/execution-specs/blob/8fad0e6b6d89313a639bea55f6942cd666b1afd9/src/ethereum_spec_tools/evm_tools/t8n/evm_trace.py#L207

And remove this:

https://github.com/ethereum/execution-specs/blob/119208cf1a13d5002074bcee3b8ea4ef096eeb0d/src/ethereum/shanghai/vm/__init__.py#L47

gurukamath:
This is not a huge prio but is not stale

PR #1195, opened 2025-04-16
Fix EVM Tracing

### LINKED ISSUE #994: Changes to `ethereum.trace.evm_trace` after import don't get applied
Pretty simple. Change the value of `evm_trace` after launching, and traces don't show up.

gurukamath:
(closes #994 )

### What was wrong?
The assigning of the `output_basedir` via `functools.partial` in t8n was not working as intended. As a result EEST is not able to fill tests with traces.

Related to Issue #994 

### How was it fixed?
Instead of patching the trace via `functools.partial`, this PR creates tracer classes which can be incorporated into the `TransactionEnvironment` and be switched as desired.

PS: This PR has many files changed simply because this feature needs to be backported to earlier forks.

#### Cute Animal Picture

Cute Animals - 1 of 1

### DIFF
--- a/src/ethereum/arrow_glacier/fork.py
+++ b/src/ethereum/arrow_glacier/fork.py
@@ -26,6 +26,7 @@ from ethereum.exceptions import (
     InvalidBlock,
     InvalidSenderError,
 )
+from ethereum.trace import BaseEvmTracer
 
 from . import vm
 from .blocks import Block, Header, Log, Receipt, encode_receipt
@@ -546,7 +547,9 @@ def apply_body(
     block_output = vm.BlockOutput()
 
     for i, tx in enumerate(map(decode_transaction, transactions)):
-        process_transaction(block_env, block_output, tx, Uint(i))
+        process_transaction(
+            block_env, block_output, tx, Uint(i), BaseEvmTracer()
+        )
 
     pay_rewards(block_env.state, block_env.number, block_env.coinbase, ommers)
 
@@ -675,6 +678,7 @@ def process_transaction(
     block_output: vm.BlockOutput,
     tx: Transaction,
     index: Uint,
+    tracer: BaseEvmTracer,
 ) -> None:
     """
     Execute a transaction against the provided environment.
@@ -698,6 +702,8 @@ def process_transaction(
         Transaction to execute.
     index:
         Index of the transaction in the block.
+    tracer:
+        Evm tracer for the transaction.
     """
     trie_set(
         block_output.transactions_trie,
@@ -746,7 +752,7 @@ def process_transaction(
         access_list_storage_keys=access_list_storage_keys,
         index_in_block=index,
         tx_hash=get_transaction_hash(encode_transaction(tx)),
-        traces=[],
+        tracer=tracer,
     )
 
     message = prepare_message(block_env, tx_env, tx)
--- a/src/ethereum/trace.py
+++ b/src/ethereum/trace.py
@@ -17,8 +17,8 @@ See [EIP-3155] for more details on EVM traces.
 """
 
 import enum
-from dataclasses import dataclass
-from typing import Optional, Protocol, Union
+from dataclasses import dataclass, field
+from typing import Optional, Union
 
 from ethereum.exceptions import EthereumException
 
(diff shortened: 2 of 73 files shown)


danceratopz:
There seems to be one very minor bug, which could even be solved in a follow-up PR. I would not have found it, had I not been interested in this specific test case.

When i compare the results between:
1. The last version of tracing that was working in `forks/prague`: https://github.com/ethereum/execution-specs/commit/5f2d661dd30a5813aa972f27feb332735c6caaaa
2. The current state of this PR: https://github.com/ethereum/execution-specs/pull/1195/commits/4a3100216029b014ddee1898e833852f457127e6

For the case where blocks include multiple transactions, it seems that the trace file accumulates the traces from previous transactions, instead of only including the traces from the current transaction. It's perhaps easiest to diff the two directories in the attached tarball to see the problem and inspect the files; in the screenshot below it's visible in the sizes of the trace files.

```
meld execution-specs-1195/traces_5f2d661dd/ execution-specs-1195/traces_4a3100216/
```
image

In particular, here you can see that the second trace file contains the contents of the first tx trace file:
```
meld execution-specs-1195/traces_4a3100216/prague__eip7002_el_triggerable_withdrawals__withdrawal_requests__test_withdrawal_requests/fork_Prague_blockchain_test_single_block_single_withdrawal_request_delegatecall_staticcall_callcode/0/trace-0-0x21edd7269e6a7d4083f3d0b9977938dc4d8b406ebc38102bf2de618a7d39f735.jsonl execution-specs-1195/traces_4a3100216/prague__eip7002_el_triggerable_withdrawals__withdrawal_requests__test_withdrawal_requests/fork_Prague_blockchain_test_single_block_single_withdrawal_request_delegatecall_staticcall_callcode/0/trace-1-0x9d7862f8f237c89adbb32537d6f4335da1e3f7371d4596a00cdd86e83d7bb39b.jsonl
```

Unfortunately, I didn't find a simpler test that demonstrates the problem yet (even though I thought I'd found the conditions to produce it).

This can be reproduced on EEST main https://github.com/ethereum/execution-spec-tests/commit/84d28dbc27fb36e040c8779d822be99f80b16547 with:

```
rm -rf fixtures traces_5f2d661dd/; fill tests/prague/eip7002_el_triggerable_withdrawals/test_withdrawal_requests.py::test_withdrawal_requests[fork_Prague-blockchain_test-single_block_single_withdrawal_request_delegatecall_staticcall_callcode] --until=Prague --traces --evm-dump-dir traces_5f2d661dd/ -v
```

Here's the test case:
https://github.com/ethereum/execution-spec-tests/blob/84d28dbc27fb36e040c8779d822be99f80b16547/tests/prague/eip7002_el_triggerable_withdrawals/test_withdrawal_requests.py#L468-L510

and here's how the transactions in the blocks are created:
https://github.com/ethereum/execution-spec-tests/blob/84d28dbc27fb36e040c8779d822be99f80b16547/tests/prague/eip7002_el_triggerable_withdrawals/conftest.py#L75-L117

danceratopz:
Nice! Traces from [4c98117](https://github.com/ethereum/execution-specs/pull/1195/commits/4c9811723a0e65ac9de5d67824b571c917ed1a81) match those of https://github.com/ethereum/execution-specs/commit/5f2d661dd30a5813aa972f27feb332735c6caaaa!

src/ethereum/arrow_glacier/fork.py
```diff
@@ -675,6 +678,7 @@ def process_transaction(
     block_output: vm.BlockOutput,
     tx: Transaction,
     index: Uint,
+    tracer: BaseEvmTracer,
```
SamWilsn, line 681: We should tend towards minimizing the impact of tracing on the spec as much as possible, and I think this is going in the wrong direction. Someone calling `process_transaction` directly now has to consider tracing, where they didn't before.

src/ethereum/trace.py
```diff
@@ -167,40 +167,29 @@ class GasAndRefund:
...
     """
-    An [`EvmTracer`] that discards all events.
-
-    [`EvmTracer`]: ref:ethereum.trace.EvmTracer
+    A basic trace.
     """
 
 
-class EvmTracer(Protocol):
+@dataclass
```
SamWilsn, line 177: I'm not sure that we should require subclasses to be dataclasses.

src/ethereum/trace.py
```diff
@@ -167,40 +167,29 @@ class GasAndRefund:
...
-    [`__call__`]: ref:ethereum.trace.EvmTracer.__call__
+    Default tracer for an EVM.
     """
 
-    def __call__(
+    traces: list[BaseTrace] = field(default_factory=list)
+    trace_memory: bool = False
+    trace_stack: bool = True
+    trace_return_data: bool = False
+    output_basedir: str = "."
```
SamWilsn, line 187: This class doesn't use these fields, so why does it define them? For `traces` in particular, we should consider a future implementation that fixes #909 (and so `traces` in the base class makes no sense).

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -105,13 +104,15 @@ def __init__(
             trace_memory = getattr(self.options, "trace.memory", False)
             trace_stack = not getattr(self.options, "trace.nostack", False)
             trace_return_data = getattr(self.options, "trace.returndata")
-            trace.evm_trace = partial(
-                evm_trace,
+            self.tracer: Union[BaseEvmTracer, FullEvmTracer] = FullEvmTracer(
```
SamWilsn, line 107: Since `FullEvmTracer` inherits from `BaseEvmTracer`, would this work:

```suggestion
            self.tracer: BaseEvmTracer = FullEvmTracer(
```

tests/conftest.py
```diff
@@ -48,12 +48,10 @@ def pytest_configure(config: Config) -> None:
...
     if config.getoption("evm_trace"):
         import ethereum.trace
-        from ethereum_spec_tools.evm_tools.t8n.evm_trace import (
-            evm_trace as new_trace_function,
-        )
+        from ethereum_spec_tools.evm_tools.t8n.evm_trace import FullEvmTracer
 
         # Replace the function in the module
-        ethereum.trace.evm_trace = new_trace_function
+        ethereum.trace.BaseEvmTracer = FullEvmTracer  # type: ignore[misc]
```
SamWilsn, line 54: This'll have the same problem (I think) as the current implementation. See https://stackoverflow.com/a/3536638

gurukamath:
Closing this for #1199

Issue #1167, opened 2025-03-20
Investigate code upgrades/streamlining for python 3.10 => 3.11 upgrade


Carsons-Eels:
Changes from Python 3.10 => 3.11 that might affect us or are potentially interesting for improving code quality:

- [PEP 673 – Self Type](https://peps.python.org/pep-0673/)
- [PEP 654 – Exception Groups and except*](https://peps.python.org/pep-0654/)
- [PEP 678 – Enriching Exceptions with Notes](https://peps.python.org/pep-0678/)
- [`dataclasses`](https://docs.python.org/3/whatsnew/3.11.html#dataclasses)
- [`enum`](https://docs.python.org/3/whatsnew/3.11.html#enum)
- [`hashlib.file_digest`](https://docs.python.org/3/library/hashlib.html#hashlib.file_digest)
- [`logging.getLevelNamesMapping`](https://docs.python.org/3/library/logging.html#logging.getLevelNamesMapping)
- [`re` Atomic grouping and possessive quantifiers](https://docs.python.org/3/whatsnew/3.11.html#re)
- [`string`](https://docs.python.org/3/whatsnew/3.11.html#string)
- [`shutil.rmtree`](https://docs.python.org/3/library/shutil.html#shutil.rmtree)
- [`time.sleep`](https://docs.python.org/3/whatsnew/3.11.html#time)
- [`typing`](https://docs.python.org/3/whatsnew/3.11.html#typing)
- [Unpacking without parenthesis is now allowed: see `bpo-46725`](https://github.com/python/cpython/issues/90881https://github.com/python/cpython/issues/90881)
- [`-P` command line option and `PYTHONSAFEENV` environment variable](https://github.com/python/cpython/issues/57684)

Anything turning out to be worth the investment in effort should be made it's own issue, or if there are few enough of them, a single issue grouping the desired changes together.

Carsons-Eels:
Old.

PR #1198, opened 2025-04-21
feat: Implement PEP 673 Self Type for improved type annotations

richardgreg:
- Add `Self` import from `typing` module to affected files
- Replace `TypeVar` used for self-references with `Self` type
- Update method return type annotations in:
  - Field protocol and implementations in finite_field.py
  - EllipticCurve class in elliptic_curve.py
  - State class in state_db.py
  - Blake2 class in blake2.py

Related to #1167

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/crypto/blake2.py
+++ b/src/ethereum/crypto/blake2.py
@@ -4,7 +4,7 @@ The Blake2 Implementation
 """
 import struct
 from dataclasses import dataclass
-from typing import Final, List, Tuple
+from typing import Final, List, Tuple, Self
 
 from ethereum_types.numeric import Uint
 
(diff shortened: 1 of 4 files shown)


src/ethereum/crypto/blake2.py
```diff
@@ -151,7 +151,7 @@ def get_blake2_parameters(self, data: bytes) -> Tuple:
 
     def G(
         self, v: List, a: Uint, b: Uint, c: Uint, d: Uint, x: Uint, y: Uint
-    ) -> List:
+    ) -> Self:
```
SamWilsn, line 154: This change seems like it should have some corresponding changes elsewhere, because the return type is changing. Is that not the case?

SamWilsn (changes requested):

PR #1214, opened 2025-05-08
feat: add initial EIP-7805

jacobkaufmann:
### What was wrong?

N/A

### How was it fixed?

add initial implementation of EIP-7805.

#### Cute Animal Picture

`TODO`

### DIFF
--- a/src/ethereum/osaka/fork.py
+++ b/src/ethereum/osaka/fork.py
@@ -12,6 +12,7 @@ Introduction
 Entry point for the Ethereum specification.
 """
 
+from copy import deepcopy
 from dataclasses import dataclass
 from typing import List, Optional, Tuple, Union
 
@@ -217,6 +218,7 @@ def state_transition(chain: BlockChain, block: Block) -> None:
         block_env=block_env,
         transactions=block.transactions,
         withdrawals=block.withdrawals,
+        inclusion_list=block.inclusion_list,
     )
     block_state_root = state_root(block_env.state)
     transactions_root = root(block_output.transactions_trie)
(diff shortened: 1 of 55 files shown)


src/ethereum/osaka/fork.py
```diff
@@ -0,0 +1,1003 @@
...
+        # If the transaction is already present in the block, then skip.
+        if tx in transactions:
+            continue
+
+        block_env = deepcopy(block_env)
+        block_output = deepcopy(block_output)
+        try:
+            tx = decode_transaction(tx)
+            index = Uint(len(transactions))
+            process_transaction(block_env, block_output, tx, index)
```
jihoonsong, line 923: We don't execute IL transactions. We just need to check its nonce and balance, and see if there is sufficient gas left to include any missing IL transactions.

Current version of EIP is misleading and we have a pending PR that fixes it: https://github.com/ethereum/EIPs/pull/9381/files

jacobkaufmann: ah I see. I forgot about this PR, and so I was just going off what I saw in the EIP.

gurukamath: I would say in addition to checking for nonce and balance, would we need to perform some basic checks to make sure that the transaction is well formed? Including perhaps the checks that are performed in the [validate_transaction](https://github.com/ethereum/execution-specs/blob/68f3a1f1ccebc048daf5dfec58bf0325e5fcc1b8/src/ethereum/osaka/transactions.py#L191) function?

jihoonsong: The EL builds an IL with transactions from the mempool and each transaction is validated when added to the mempool. That being said, it makes sense to specify that IL transactions are intrinsically valid.

src/ethereum/osaka/fork.py
```diff
@@ -705,6 +709,13 @@ def apply_body(
     for i, tx in enumerate(map(decode_transaction, transactions)):
         process_transaction(block_env, block_output, tx, Uint(i))
 
+    validate_inclusion_list(
```
jihoonsong, line 712: I want to share an idea about an edge case testing. It would be nice if we could test such situations that some withdrawal address become available to cover some IL transaction that they couldn't. That missing IL transaction should not reorg the block.

fselmo:
closing in favor of #1928. Happy to re-open if this was in error.

PR #1225, opened 2025-05-16
Implement EIP-5920

### LINKED ISSUE #1181: Implement EIP-5920

gurukamath:
(closes #1181 )

### What was wrong?
The current Osaka branch does not have the `PAY` opcode implementation which has been CFI'd for Osaka

Related to Issue #1181 

### How was it fixed?
Implement EIP-5920

#### Cute Animal Picture

Cute Animals - 1 of 1

### DIFF
--- a/src/ethereum/osaka/utils/address.py
+++ b/src/ethereum/osaka/utils/address.py
@@ -44,9 +44,10 @@ def to_address(data: Union[Uint, U256]) -> Address:
     return Address(data.to_be_bytes32()[-20:])
 
 
-def to_address_without_mask(data: U256) -> Address:
+def to_address_unmasked(data: U256) -> Address:
     """
-    Convert a U256 value to a valid address (20 bytes).
+    Convert a U256 value to a valid address (20 bytes),
+    raising an error if the input is too large to fit.
 
     Parameters
     ----------
@@ -65,7 +66,7 @@ def to_address_without_mask(data: U256) -> Address:
     """
     if data > U256.from_be_bytes(MAX_ADDRESS):
         raise InvalidParameter("Address is too large")
-    return Address(data.to_be_bytes32()[-20:])
+    return to_address(data)
 
 
 def compute_contract_address(address: Address, nonce: Uint) -> Address:
--- a/src/ethereum/osaka/vm/instructions/system.py
+++ b/src/ethereum/osaka/vm/instructions/system.py
@@ -31,7 +31,7 @@ from ...utils.address import (
     compute_contract_address,
     compute_create2_contract_address,
     to_address,
-    to_address_without_mask,
+    to_address_unmasked,
 )
 from ...vm.eoa_delegation import access_delegation
 from .. import (
@@ -747,7 +747,7 @@ def revert(evm: Evm) -> None:
 
 def pay(evm: Evm) -> None:
     """
-    Transfer ether to an account.
+    Transfer ether to an account without executing its code.
 
     Parameters
     ----------
@@ -755,7 +755,7 @@ def pay(evm: Evm) -> None:
         The current EVM frame.
     """
     # STACK
-    to = to_address_without_mask(pop(evm.stack))
+    to = to_address_unmasked(pop(evm.stack))
     value = pop(evm.stack)
 
     # GAS
@@ -765,11 +765,9 @@ def pay(evm: Evm) -> None:
         evm.accessed_addresses.add(to)
         access_gas_cost = GAS_COLD_ACCOUNT_ACCESS
 
-    create_gas_cost = (
-        Uint(0)
-        if is_account_alive(evm.message.block_env.state, to) or value == 0
-        else GAS_NEW_ACCOUNT
-    )
+    create_gas_cost = GAS_NEW_ACCOUNT
+    if value == 0 or is_account_alive(evm.message.block_env.state, to):
+        create_gas_cost = Uint(0)
 
     transfer_gas_cost = Uint(0) if value == U256(0) else GAS_CALL_VALUE
 


src/ethereum/osaka/utils/address.py
```diff
@@ -22,6 +22,9 @@
 from ethereum.utils.byte import left_pad_zero_bytes
 
 from ..fork_types import Address
+from ..vm.exceptions import InvalidParameter
+
+MAX_ADDRESS = Address(b"\xff" * 20)
```
jochem-brouwer, line 27: As optimization you could add `MAX_ADDRESS_UINT256` here as a constant

src/ethereum/osaka/utils/address.py
```diff
@@ -41,6 +44,30 @@ def to_address(data: Union[Uint, U256]) -> Address:
...
+    ------
+    InvalidParameter
+        If `data` is larger than `MAX_ADDRESS`.
+
+    Returns
+    -------
+    address : `Address`
+        The obtained address.
+    """
+    if data > U256.from_be_bytes(MAX_ADDRESS):
```
jochem-brouwer, line 66: And then query the `MAX_ADDRESS_UINT256` here :smile: :+1:

src/ethereum/osaka/vm/instructions/system.py
```diff
@@ -742,3 +743,52 @@ def revert(evm: Evm) -> None:
...
+        else GAS_NEW_ACCOUNT
+    )
+
+    transfer_gas_cost = Uint(0) if value == U256(0) else GAS_CALL_VALUE
+
+    charge_gas(evm, access_gas_cost + create_gas_cost + transfer_gas_cost)
+
+    # OPERATION
+    if evm.message.is_static:
+        raise WriteInStaticContext("Cannot PAY in static context")
```
jochem-brouwer, line 780: In the spec, this check is done before any other operation (so also before popping any value from stack). I don't think moving this around in this method would yield a different outcome, but it would match the spec better to do this check prior to anything else. So I would suggest moving this check up.

src/ethereum/osaka/vm/instructions/system.py
```diff
@@ -742,3 +743,52 @@ def revert(evm: Evm) -> None:
...
+    """
+    # STACK
+    to = to_address_without_mask(pop(evm.stack))
+    value = pop(evm.stack)
+
+    # GAS
+    if to in evm.accessed_addresses:
+        access_gas_cost = GAS_WARM_ACCESS
+    else:
+        evm.accessed_addresses.add(to)
```
jochem-brouwer, line 765: If the method encounters an error later, is the `to` address removed from warm addresses if it got added here?

jochem-brouwer (commented):
I think this matches the spec! I got some minor suggestions though :smile: :+1:

src/ethereum/osaka/vm/instructions/system.py
gurukamath: It will continue to be a part of the warm addresses in the current execution frame but will not be added to the warm addresses of the parent frame (contrast `incorporate_child_on_success` vs `incorporate_child_on_error` [here](https://github.com/gurukamath/execution-specs/blob/aa0bb3be926211cce791e79e312962159d935c25/src/ethereum/osaka/vm/__init__.py#L176)). This means the warmed up address play no further role in the execution.

If we are in the initial frame, the warmed addresses are not explicitly removed. However, since an error is basically the end of the execution in that case, the warmed addresses play not further role here either.

src/ethereum/osaka/vm/instructions/system.py
gurukamath: Yes. This does not make a difference to the outcome since all the gas in consumed and warmed addresses are discarded upon error. However, I implemented it in this order since that makes it consistent with the implementation of `CALL*` and other methods where you check if the frame is static.

src/ethereum/osaka/utils/address.py
```diff
@@ -22,6 +22,9 @@
 from ethereum.utils.byte import left_pad_zero_bytes
 
 from ..fork_types import Address
+from ..vm.exceptions import InvalidParameter
+
+MAX_ADDRESS = Address(b"\xff" * 20)
 
 
 def to_address(data: Union[Uint, U256]) -> Address:
```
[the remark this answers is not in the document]
gurukamath: Created #1245

src/ethereum/osaka/vm/instructions/system.py
```diff
@@ -742,3 +743,52 @@ def revert(evm: Evm) -> None:
...
+        access_gas_cost = GAS_COLD_ACCOUNT_ACCESS
```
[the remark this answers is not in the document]
gurukamath: This is just the way we define `access_gas_cost` in other opcodes. I do not see a big difference one way or another, so am ok with either representation

src/ethereum/osaka/vm/instructions/system.py
```diff
@@ -742,3 +743,52 @@ def revert(evm: Evm) -> None:
...
+    try:
+        move_ether(
+            evm.message.block_env.state,
+            evm.message.current_target,
+            to,
+            value,
+        )
+        push(evm.stack, U256(1))
+    except AssertionError:
+        push(evm.stack, U256(0))
```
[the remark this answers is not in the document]
gurukamath: I would have also prefered to not catch a general `AssertionError` but since the check is already performed in `move_ether`, I think we should just take advantage of that. Perhaps, raising a more specific exception in `move_ether` like `InsufficientBalance` and trying to catch that is a better way to go.

PR #1233, opened 2025-05-19
fix(t8n): return the receipts list

danceratopz:
### What was wrong?

v1.17.0 doesn't fill EEST Prague tests; `t8n` returns the incorrect type. It returns the first transaction receipt instead of a list of receipts.
```
tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g1add.py:207: in test_valid
    state_test(
src/pytest_plugins/filler/filler.py:750: in __init__
    fixture = self.generate(
src/ethereum_test_specs/state.py:252: in generate
    return self.make_state_test_fixture(t8n, fork, eips)
src/ethereum_test_specs/state.py:185: in make_state_test_fixture
    transition_tool_output = t8n.evaluate(
src/ethereum_clis/transition_tool.py:536: in evaluate
    return self._evaluate_server(
src/ethereum_clis/transition_tool.py:357: in _evaluate_server
    output: TransitionToolOutput = TransitionToolOutput.model_validate(
E   pydantic_core._pydantic_core.ValidationError: 1 validation error for TransitionToolOutput
E   result.receipts
E     Input should be a valid list [type=list_type, input_value={'transactionHash': '0xa5...0000000000', 'logs': []}, input_type=dict]
E       For further information visit https://errors.pydantic.dev/2.10/v/list_type
```

### How was it fixed?

Return the list.

#### Cute Animal Picture

image

Required for https://github.com/ethereum/execution-spec-tests/pull/1454

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
@@ -339,14 +339,6 @@ class Result:
 
             receipt_dict["gasUsed"] = hex(receipt.cumulative_gas_used)
             receipt_dict["bloom"] = "0x" + receipt.bloom.hex()
-            receipt_dict["logs"] = [
-                {
-                    "address": "0x" + log.address.hex(),
-                    "topics": ["0x" + topic.hex() for topic in log.topics],
-                    "data": "0x" + log.data.hex(),
-                }
-                for log in receipt.logs
-            ]
 
             receipts_json.append(receipt_dict)
 
--- a/src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
@@ -350,7 +350,7 @@ class Result:
 
             receipts_json.append(receipt_dict)
 
-            return receipt_dict
+        return receipts_json
 
     def to_json(self) -> Any:
         """Encode the result to JSON"""


gurukamath:
This is quite likely related to [this](https://github.com/ethereum/execution-specs/pull/1197) change.

In order to align EELS with the spec, it was agreed that we should wait for all the transactions to complete and only then process all the deposit logs in one go. This is as opposed to processing the deposit logs from a particular transaction within that transaction.

The former means the transaction processes successfully while the error pops up when the deposit request from that transaction is being parsed.

The latter means the error pops up when the transaction is being processed.

Both result in error, the difference being what type of error is raised. I think the EEST fixtures need to be updated to expect the correct error.

gurukamath:
> Thanks @gurukamath! The receipts, respectively receipt logs, issues have been fixed.
> 
> There is now an unrelated issue with `tests/prague/eip6110_deposits/test_modified_contract.py::test_invalid_layout` which should probably be addressed in a separate PR:

This is quite likely related to [this](https://github.com/ethereum/execution-specs/pull/1197) change.

In order to align EELS with the spec, it was agreed that we should wait for all the transactions to complete and only then process all the deposit logs in one go. This is as opposed to processing the deposit logs from a particular transaction within that transaction.

The former means the transaction processes successfully while the error pops up when the deposit request from that transaction is being parsed.

The latter means the error pops up when the transaction is being processed.

Both result in error, the difference being what type of error is raised. I think the EEST fixtures need to be updated to expect the correct error.

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -339,18 +339,10 @@ def json_encode_receipts(self) -> Any:
...
             receipt_dict["gasUsed"] = hex(receipt.cumulative_gas_used)
             receipt_dict["bloom"] = "0x" + receipt.bloom.hex()
-            receipt_dict["logs"] = [
-                {
-                    "address": "0x" + log.address.hex(),
-                    "topics": ["0x" + topic.hex() for topic in log.topics],
-                    "data": "0x" + log.data.hex(),
-                }
-                for log in receipt.logs
-            ]
```
marioevz, line 349: Why were these removed?

gurukamath: Because `fill` expects fields in the log aren't directly part of the log (e.g. blockhash). It is also unclear how these fields are being used by `fill`. It doesn't complain if the log is not supplied by the `t8n` result.

Removing the logs from the result is more of a quick fix. In the long term, we should analyse which fields are actually essential and work towards including them.

danceratopz: I think you mentioned that they weren't there previously? So we'd never noticed that fields were missing? I.e., removing them now is not really taking anything away that was there previously?

SamWilsn (approved):

PR #1241, opened 2025-05-22
lower vulture min confidence from 100 to 60

### LINKED ISSUE #1239: Lower the vulture min confidence
Continuing on from #1122, we should drop the confidence value as low as possible while still staying productive:

https://github.com/ethereum/execution-specs/blob/6b0f19f72cbc8458620ad313b6863ed0ea941412/tox.ini#L17

naevern:
This PR lowers the minimum confidence threshold for Vulture from 100 to 60, allowing it to detect more potentially unused code. This change will help identify dead code that might otherwise be missed with the stricter setting.

Fixes #1239

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,7 +14,7 @@ commands =
     flake8 src tests setup.py
     mypy src tests setup.py --exclude "tests/fixtures/*" --namespace-packages
     ethereum-spec-lint
-    vulture src tests setup.py --exclude "*/tests/fixtures/*" --min-confidence 100
+    vulture src tests setup.py --exclude "*/tests/fixtures/*" --min-confidence 60
 
 [testenv:py3]
 extras =


SamWilsn:
Presumably this'll pick up new errors when you run `tox`? Unfortunately our CI is broken at the moment, but you can run it locally. Part of closing this issue would be fixing/ignoring any diagnostics that vulture reports.

Unless you have run it locally and it's clean?

naevern:
thank for the preferred value @Carsons-Eels . so, I'm settings the value to `61` and doing commit on this PR again.

naevern:
@SamWilsn @gurukamath 

according to @Carsons-Eels preferrence I've updated to threshold which gives us the lowest possible threshold while maintaining a clean build - the best of both worlds

naevern:
hi @gurukamath @Carsons-Eels can I please have a review on the change if it's good to go?

Issue #1244, opened 2025-05-23
Implement 7212


LouisTsai-Csie:
Implement EIP-7212 (Currently RIP-7212)

PR #1249, opened 2025-05-23
feat: implement eip-7212

LouisTsai-Csie:
### What was wrong?

Related to Issue #1244 

### How was it fixed?

Implement EIP-7212

#### Cute Animal Picture

Cute Animal


src/ethereum/osaka/vm/precompiled_contracts/ecrecover.py
```diff
@@ -54,7 +54,7 @@ def ecrecover(evm: Evm) -> None:
 
     try:
         public_key = secp256k1_recover(r, s, v - U256(27), message_hash)
-    except InvalidSignatureError:
+    except Exception:
```
souradeep-das, line 57: I feel like we could use specific exceptions here, given `secp256k1_recover` method raises a kind of InvalidSignatureError?

Issue #1258, opened 2025-05-28
PyPy3 runs tests way slower than CPython


SamWilsn:
PyPy3 is slower than CPython for running our test suite. That probably shouldn't be the case.

PR #1268, opened 2025-06-05
Optimized bls12 381

fselmo:
### What was wrong?

While investigating #1258 a bit, I realized that the optimized implementations in `py_ecc` avoid the `FQ.__init__()` overhead both in the `FQP.__init__()`, as well as in `FQP.__mul__() `.

Since specifically the `FQ.__init__()` seems to be where most of the time was hanging, this should help.

### How was it fixed?

I'd implemented the optimized versions of bls12_381 in `py-evm` (heavily basing it off of the EELS implementation 🙂 ) and I figured this can only help with the load. Porting it over to here and hope this is useful!

I wasn't sure how to test here, so I filled the tests successfully with `execution-spec-tests` pointing to my local branch - all tests passing.

---

Note: fwiw it looks like there's still quite a bit of unoptimized `py_ecc` code for bn128 across many forks. That may be where most of the hangup is since bls12_381 is just prague.

### DIFF
--- a/src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
+++ b/src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
@@ -11,27 +11,29 @@ Introduction
 
 Precompile for BLS12-381 curve operations.
 """
+
 from typing import Tuple, Union
 
 from ethereum_types.bytes import Bytes
 from ethereum_types.numeric import U256, Uint
-from py_ecc.bls12_381.bls12_381_curve import (
-    FQ,
-    FQ2,
+from py_ecc.optimized_bls12_381.optimized_curve import FQ as OPTIMIZED_FQ
+from py_ecc.optimized_bls12_381.optimized_curve import FQ2 as OPTIMIZED_FQ2
+from py_ecc.optimized_bls12_381.optimized_curve import FQP as OPTIMIZED_FQP
+from py_ecc.optimized_bls12_381.optimized_curve import (
     b,
     b2,
     curve_order,
+    is_inf,
     is_on_curve,
-    multiply,
 )
-from py_ecc.optimized_bls12_381.optimized_curve import FQ as OPTIMIZED_FQ
-from py_ecc.optimized_bls12_381.optimized_curve import FQ2 as OPTIMIZED_FQ2
-from py_ecc.typing import Point2D
+from py_ecc.optimized_bls12_381.optimized_curve import (
+    multiply as bls12_multiply_optimized,
+)
+from py_ecc.optimized_bls12_381.optimized_curve import normalize
+from py_ecc.typing import Optimized_Point3D
 
 from ....vm.memory import buffer_read
 from ...exceptions import InvalidParameter
-
-P = FQ.field_modulus
 
 G1_K_DISCOUNT = [
     1000,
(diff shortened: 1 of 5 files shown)


src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -426,31 +423,24 @@ def bytes_to_FQ(
...
-        return OPTIMIZED_FQ(c)
-    else:
-        return FQ(c)
+    return FQ(c)
 
 
-def bytes_to_FQ2(
-    data: Bytes, optimized: bool = False
-) -> Union[FQ2, OPTIMIZED_FQ2]:
+def bytes_to_fq2(data: Bytes) -> Union[FQ2, FQ2]:
```
gurukamath, line 432:
```suggestion
def bytes_to_fq2(data: Bytes) -> FQ2:
```

src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -426,31 +423,24 @@ def bytes_to_FQ(
...
     ----------
     data :
         The bytes data to decode.
-    optimized :
-        Whether to use the optimized FQ2 implementation.
 
     Returns
     -------
-    fq2 : Union[FQ2, OPTIMIZED_FQ2]
+    fq2 : Union[FQ2, FQP]
```
gurukamath, line 443:
```suggestion
    fq2 : FQ2
```

src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -463,18 +453,17 @@ def bytes_to_FQ2(
...
-    if optimized:
-        return OPTIMIZED_FQ2((c_0, c_1))
-    else:
-        return FQ2((c_0, c_1))
+    return FQ2((c_0, c_1))
 
 
-def bytes_to_G2(data: Bytes) -> Point2D:
+def bytes_to_g2(
+    data: bytes,
```
gurukamath, line 465:
```suggestion
    data: Bytes,
```

src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -463,18 +453,17 @@ def bytes_to_FQ2(
...
 
-    if optimized:
-        return OPTIMIZED_FQ2((c_0, c_1))
-    else:
-        return FQ2((c_0, c_1))
+    return FQ2((c_0, c_1))
 
 
-def bytes_to_G2(data: Bytes) -> Point2D:
+def bytes_to_g2(
```
gurukamath, line 464: A nit here. We should have a consistent naming convention for the functions `bytes_to_G2` instead of `bytes_to_g2` (all small letters). Elsewhere in the module, there is `FQ2_to_bytes`. Irrespective of which way we name, we should be consistent

gurukamath (changes requested):

src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
gurukamath: Yeah. All lowercase sounds good to me.

gurukamath (approved):

PR #1275, opened 2025-06-10
feat(osaka): add eip-7939 count leading zeros opcode

spencer-tb:
#### Description

Adds the CLZ (count leading zeros opcode) for [EIP-7939](https://github.com/Vectorized/EIPs/blob/a770f43b7c15e3e217c840253e6bb8025f65c9e3/EIPS/eip-7939.md).

#### Cute Animal Picture

1269071-cute-pig-wallpapers-1920x1200-for-mac-3717821457

### DIFF
--- a/src/ethereum/osaka/vm/instructions/bitwise.py
+++ b/src/ethereum/osaka/vm/instructions/bitwise.py
@@ -238,3 +238,34 @@ def bitwise_sar(evm: Evm) -> None:
 
     # PROGRAM COUNTER
     evm.pc += Uint(1)
+
+
+def count_leading_zeros(evm: Evm) -> None:
+    """
+    Count the number of leading zero bits in a 256-bit word.
+    
+    Pops one value from the stack and pushes the number of leading zero bits.
+    If the input is zero, pushes 256.
+    
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+    """
+    # STACK
+    x = pop(evm.stack)
+    
+    # GAS
+    charge_gas(evm, GAS_VERY_LOW)
+    
+    # OPERATION
+    if x == 0:
+        result = U256(256)
+    else:
+        bit_length = x.bit_length()
+        result = U256(256 - bit_length)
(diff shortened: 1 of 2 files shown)


src/ethereum/osaka/vm/instructions/bitwise.py
```diff
@@ -238,3 +238,34 @@ def bitwise_sar(evm: Evm) -> None:
...
+        The current EVM frame.
+    """
+    # STACK
+    x = pop(evm.stack)
+    
+    # GAS
+    charge_gas(evm, GAS_VERY_LOW)
+    
+    # OPERATION
+    if x == 0:
```
gurukamath, line 262: iirc, if `x == 0`, `x.bit_length()` is also zero.

src/ethereum/osaka/vm/instructions/bitwise.py
```diff
@@ -238,3 +238,34 @@ def bitwise_sar(evm: Evm) -> None:
...
+        result = U256(256)
+    else:
+        bit_length = x.bit_length()
+        result = U256(256 - bit_length)
```
gurukamath, line 266:
```suggestion
    bit_length = x.bit_length()
    result = U256(256 - bit_length)
```

src/ethereum/osaka/vm/instructions/bitwise.py
```diff
@@ -238,3 +238,31 @@ def bitwise_sar(evm: Evm) -> None:
...
+    Parameters
+    ----------
+    evm :
+        The current EVM frame.
+    """
+    # STACK
+    x = pop(evm.stack)
+
+    # GAS
+    charge_gas(evm, GAS_VERY_LOW)
```
jochem-brouwer, line 259: Small nit: we want to charge gas before doing any operations. So the gas should be charged before the `pop`. If there is not enough gas available, then we thus also do not perform the `pop` logic where there are insufficient funds for :smile: :+1:

PR #1280, opened 2025-06-18
Draft: STEEL Merge

gurukamath:
### What was wrong?
Currently, the test cases live in a different repository (`execution-spec-tests`).

### How was it fixed?
Move the test cases to `execution-specs` and start using `execution-spec-tests` as a library.

#### Cute Animal Picture

Cute Animals - 1 of 1


gurukamath:
The PR is ready for review. The CI fails after around 30 mins because it looks like the runner is terminating the job (there aren't any actual test failures). Perhaps contacting devops is required here but I'm gonna take a quick look first.

setup.cfg
```diff
@@ -136,6 +136,8 @@ install_requires =
     ethereum-types>=0.2.1,<0.3
     ethereum-rlp>=0.1.4,<0.2
     cryptography>=45.0.1,<46
+    ethereum-execution-spec-tests @ git+https://github.com/ethereum/execution-spec-tests@dd83fba10492e6f63453b1a35608900b3aee1a87
+    ethereum-spec-evm-resolver @ git+https://github.com/petertdavies/ethereum-spec-evm-resolver
```
SamWilsn, line 140: We still need the resolver?

pytest-eest.ini
```diff
@@ -0,0 +1,23 @@
+[pytest]
+console_output_style = count
+minversion = 7.0
```
SamWilsn, line 3: Should this be 8 to match `setup.cfg`?

src/ethereum_spec_tools/evm_tools/t8n/transition_tool.py
```diff
@@ -0,0 +1,128 @@
...
+        t8n_options = parser.parse_args(t8n_args)
+
+        out_stream = StringIO()
+
+        in_stream = StringIO(json.dumps(request_data_json["input"]))
+
+        t8n = T8N(t8n_options, out_stream, in_stream)
+        t8n.run()
+
+        output_dict = json.loads(out_stream.getvalue())
```
SamWilsn, line 96: We really should make this interface less... insane :rofl:

src/ethereum_spec_tools/evm_tools/t8n/transition_tool.py
```diff
@@ -0,0 +1,128 @@
...
+            "--output.alloc=stdout",
+            f"--state.fork={request_data_json['state']['fork']}",
+            f"--state.chainid={request_data_json['state']['chainid']}",
+            f"--state.reward={request_data_json['state']['reward']}",
+        ]
+
+        if transition_tool_data.state_test:
+            t8n_args.append("--state-test")
+
+        temp_dir = tempfile.TemporaryDirectory()
```
SamWilsn, line 75: Is there a context manager (`with ...:`) version of this?

tox.ini
```diff
@@ -30,7 +30,23 @@ commands =
...
+    test
+setenv =
+    COVERAGE_FILE = {toxworkdir}/.coverage
+commands =
+    pytest \
+        -c pytest-eest.ini \
+        -m "not slow and not zkevm" \
+        -n auto --maxprocesses 10 \
+        --basetemp="{temp_dir}/pytest" \
+        --clean \
```
SamWilsn, line 47: What does this argument do?

SamWilsn (approved):
Seems fine to me. We can refactor our t8n tool to avoid the json round trip later.

tox.ini
gurukamath: cleans up any fixtures that may have been generated from previous runs. Otherwise pytest/fill will throw an error that there are already some fixtures. Not strictly required in the context of CI but is handy when someone is testing stuff locally

src/ethereum_spec_tools/evm_tools/t8n/transition_tool.py
gurukamath: The context manager returns the string path instead of the object. However, the `collect_traces` function expects the object. I will create an issue in EEST to see if we can use the context manager in this case

pytest-eest.ini
gurukamath: Updated!

danceratopz (commented):
One comment due to an EEST packaging improvement.

pytest-eest.ini
```diff
@@ -0,0 +1,23 @@
+[pytest]
```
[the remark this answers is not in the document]
gurukamath: Updated

src/ethereum_spec_tools/evm_tools/t8n/transition_tool.py
gurukamath: Ah. I see. Yes. That is definitely the long term plan

gurukamath:
Summary of a discussion with @SamWilsn 

A second command has been added in this PR for the fork under active development. See [this](https://github.com/gurukamath/execution-specs/blob/b71ea2efb1fe094f34ce184a4b50014cabc2d684/tox.ini#L46). This is because this command needs to only focus on the active development fork folder. Otherwise some of the historical tests will fail since they are not updated in real time. For example, Osaka tests need to only focus on `eest_tests/execution-spec-tests/tests/osaka`.

Since this is going to be a general thing, we should make an ethereum-spec-print-dev-fork command so we can single source hardfork information. We don't want to end up skipping amsterdam's tests for example because we forgot to update `tox.ini`

tox.ini
```diff
@@ -59,8 +89,10 @@ commands =
         -n auto --maxprocesses 5 \
         --ignore-glob='tests/fixtures/*' \
         --ignore-glob='tests/test_t8n.py' \
+        --ignore-glob='eest_tests/*' \
```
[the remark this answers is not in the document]
SamWilsn: Ah, okay. So `py3_eest` and `pypy3` test filling, while `py3` tests the JSON fixtures? Shouldn't `py3` exclude `eest_tests/*` as well then?

Or wait, if we're specifying `tests` specifically as an argument to `pytest`, do we need to exclude `eels_tests/*` at all?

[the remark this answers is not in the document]
gurukamath: Right. We don't. Updated!

SamWilsn:
Looks good to merge!

PR #1286, opened 2025-06-19
tests: add specific exceptions on tests

souradeep-das:
### What was wrong?

Related to Issue https://github.com/ethereum/execution-specs/issues/1021

### How was it fixed?

- While loading state tests - replaced `InvalidTransaction` on pytest.raises with specific exception type
- Added specific exception (`InsufficientTransactionGasError`, `NonceTooHighError` and `InitCodeTooLargeError`) for transaction validation, extending `InvalidTransaction`

#### Cute Animal Picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Some clarifications were enabled without protocol releases:
 | [EIP-2681](https://eips.ethereum.org/EIPS/eip-2681) | 0 |
 | [EIP-3607](https://eips.ethereum.org/EIPS/eip-3607) | 0 |
 | [EIP-7523](https://eips.ethereum.org/EIPS/eip-7523) | 15537394 |
-| [EIP-7610](https://github.com/ethereum/EIPs/pull/8161) | 0 |
+| [EIP-7610](https://eips.ethereum.org/EIPS/eip-7610) | 0 |
 
 
 ## Execution Specification (work-in-progress)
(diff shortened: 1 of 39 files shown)


tests/helpers/load_state_tests.py
```diff
@@ -79,10 +246,28 @@ def run_blockchain_st_test(test_case: Dict, load: Load) -> None:
...
+                InvalidSignatureError,
+                RLPException,
+            )
+            fork_specific_exceptions = FORK_SPECIFIC_EXCEPTIONS.get(
+                fork_name,
+                (),
+            )
+            all_exceptions = common_exceptions + fork_specific_exceptions
+
+            with pytest.raises(all_exceptions):
```
SamWilsn, line 270: I believe the JSON fixture contains a string representing the expected exception. Is it possible to map that to exactly one exception type?

tests/helpers/load_state_tests.py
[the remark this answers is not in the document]
souradeep-das: thanks, updated it!

tests/helpers/load_state_tests.py
souradeep-das: I found that the block_exceptions from legacy tests/blockchain test fixtures are very different from the state transition exceptions. (and the state_transition method from the fork is used to raise the error caught here).
 
This specific function also seems to work with blockchaintests fixtures so couldn't account for the expected exception from state_tests. If i'm thinking wrong about this - do lmk!

SamWilsn (approved):

souradeep-das:
updating this for `osaka`

PR #1312, opened 2025-07-08
feat: Process `state.reward` in T8N

marioevz:
### What was wrong?

T8N was not processing the `--state.reward` modifier during the state transition.

### How was it fixed?

I copied the logic from one of the pre-merge forks, and applied it using the custom value (only if `!= -1`).

I had to expose `create_ether` to properly follow the fork logic, let me know if this is not ok.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
@@ -10,7 +10,7 @@ from functools import partial
 from typing import Any, TextIO
 
 from ethereum_rlp import rlp
-from ethereum_types.numeric import U256, U64, Uint
+from ethereum_types.numeric import U64, U256, Uint
 
 from ethereum import trace
 from ethereum.exceptions import EthereumException, InvalidBlock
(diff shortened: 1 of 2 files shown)


src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -228,12 +228,25 @@ def _run_blockchain_test(self, block_env: Any, block_output: Any) -> None:
                 self.logger.warning(f"Transaction {i} failed: {e!r}")
 
         if not self.fork.is_after_fork("ethereum.paris"):
-            self.fork.pay_rewards(
-                block_env.state,
-                block_env.number,
-                block_env.coinbase,
-                self.env.ommers,
-            )
+            if self.options.state_reward is not None:
```
gurukamath, line 231: Another way to do this might be to also expose the forks `BLOCK_REWARD` constant and define a method in the `T8N` class with this whole logic. Say `def pay_block_rewards(...)` and then do

```
if not self.fork.is_after_fork("ethereum.paris"):
    if self.options.state_reward is None:
        pay_block_rewards(self.fork.BLOCK_REWARD, ...)
    elif self.options.state_reward != -1
        pay_block_rewards(self.options.state_reward, ...)
```

Or something along these lines

gurukamath (approved):

gurukamath:
There are some code formatting issues which should be simple to fix. I don't seem to have the permissions to push a fix. You can run `tox -e static` to see the issues

PR #1321, opened 2025-07-11
bls12_381: Cache point generation from raw bytes

fselmo:
### What was wrong?

We have some new tests in EEST that seem to hang on the subgroup check. I realized a lot of the times we are calling this costly calculation on the same point.

Related to Issues:
- #1268
- ethereum/execution-spec-tests#1894

### How was it fixed?

This PR uses a simple `lru_cache` to cache raw bytes values that we convert to G1 or G2 points, with or without the subgroup check. This should prevent us from having to re-build the same points from raw bytes over and over.

I had to do a bit of dancing around unhashable `bytearray` (`Bytes` after being sliced e.g. `var[:256]`) to use `bytes` instead for the lru cache argument but I don't think it's too bad. This runs all the g2msm tests in 9s as opposed to timing out on a single one of those 4 timeout cases after 20s. I think the biggest bottleneck was those tests are running through the entire g2msm discount table and there were a lot of inefficiencies re-calculating the same values.

### DIFF
--- a/src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
+++ b/src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
@@ -12,6 +12,7 @@ Introduction
 Precompile for BLS12-381 curve operations.
 """
 
+from functools import lru_cache
 from typing import Tuple
 
 from ethereum_types.bytes import Bytes
@@ -301,26 +302,13 @@ G2_MAX_DISCOUNT = 524
 MULTIPLIER = Uint(1000)
 
 
-def bytes_to_g1(
-    data: Bytes,
+@lru_cache(maxsize=128)
+def _bytes_to_g1_cached(
+    data: bytes,
+    subgroup_check: bool = False,
 ) -> Point3D[FQ]:
     """
-    Decode 128 bytes to a G1 point. Does not perform sub-group check.
-
-    Parameters
-    ----------
-    data :
-        The bytes data to decode.
-
-    Returns
-    -------
-    point : Point3D[FQ]
-        The G1 point.
-
-    Raises
-    ------
-    InvalidParameter
-        Either a field element is invalid or the point is not on the curve.
+    Internal cached version of bytes_to_g1 that works with hashable ``bytes``.
     """
     if len(data) != 128:
         raise InvalidParameter("Input should be 128 bytes long")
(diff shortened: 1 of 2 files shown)


fselmo:
> > I had to do a bit of dancing around unhashable `Bytes` -> `bytes` for the lru cache argument but I don't think it's too bad.
> 
> This is fishy, since:
> 
> https://github.com/ethereum/ethereum-types/blob/42a47dfa8635d9de92bbe356bb99a9ba9e60e55e/src/ethereum_types/bytes.py#L142-L145

Yeah I saw this but for some reason it's still unhashable 🤔. There may be a better way than this abstraction though. Happy to use something else here.

src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -301,26 +302,13 @@
 MULTIPLIER = Uint(1000)
 
 
-def bytes_to_g1(
-    data: Bytes,
+@lru_cache(maxsize=128)
```
SamWilsn, line 305: Would this pose a DoS vector if it were implemented this way in a production client? Like could an attacker choose values so that the fast path is never/rarely executed?

If so, it isn't a deal-breaker, but we should add a comment to that effect here.

src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -301,26 +302,13 @@
...
-    Returns
-    -------
-    point : Point3D[FQ]
-        The G1 point.
-
-    Raises
-    ------
-    InvalidParameter
-        Either a field element is invalid or the point is not on the curve.
+    Internal cached version of bytes_to_g1 that works with hashable ``bytes``.
```
SamWilsn, line 311:
```suggestion
    Internal cached version of bytes_to_g1 that works with hashable `bytes`.
```

src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -460,26 +478,13 @@ def bytes_to_fq2(data: Bytes) -> FQ2:
...
-    Returns
-    -------
-    point : Point3D[FQ2]
-        The G2 point.
-
-    Raises
-    ------
-    InvalidParameter
-        Either a field element is invalid or the point is not on the curve.
+    Internal cached version of bytes_to_g2 that works with hashable ``bytes``.
```
SamWilsn, line 487:
```suggestion
    Internal cached version of bytes_to_g2 that works with hashable `bytes`.
```

fselmo:
> Yeah I saw this but for some reason it's still unhashable 🤔. There may be a better way than this abstraction though. Happy to use something else here.

Ah I actually think it's when we slice it `var: Bytes` -> `var[:256]`. This makes it a `bytearray` which is unhashable. It's hard to find EEST <-> EELS errors sometimes but I'm getting used to it.

src/ethereum/prague/vm/precompiled_contracts/bls12_381/__init__.py
fselmo: Yes and I think that's exactly what `benchmark` tests are meant to test against, trending cache hits to zero while remaining inside constraints. I think a comment is well warranted 👍🏼

fselmo:
@SamWilsn updated based on comments. I added a clarifying comment on why the `bytes` dance was needed (slicing `Bytes`) since that wasn't very obvious.

SamWilsn:
Can this go into `forks/osaka` or do you need it in `master` as well?

fselmo:
> I can do it, no worries!

I had a sneaky edit to my comment. I think `master` is also struggling with this on the EEST side. I think @spencer-tb has a better idea but I assume we want it in both `master` and `forks/osaka`.

SamWilsn:
We'll keep it in `master` and it'll get picked up when we rebase `forks/osaka` next time.

fselmo:
If we prefer to leave caching out of it, [this PR in py-ecc](https://github.com/ethereum/py_ecc/pull/156) resolves the issues without caching. Or we could implement both.

SamWilsn:
Yep! I don't mind adding it.

Issue #1319, opened 2025-07-09
Implement EIP-7594


gurukamath:
The proposed cap on blobs in each transaction in EIP-7594 has to be enforced in the EL as well and will require an EELS implementation

- ethereum/EIPs#9981

PR #1325, opened 2025-07-16
feat(Osaka): Implement EIP-7594

gurukamath:
### What was wrong?
The EIP-7594 although largly relevant to the CL, imposes a limit on the number of blobs allowed in a type 3 tx. This needs to be implemented in the EL as well.

Related to Issue #1319

### How was it fixed?
Impose the blob limit.

### DIFF
--- a/src/ethereum/osaka/fork.py
+++ b/src/ethereum/osaka/fork.py
@@ -33,6 +33,7 @@ from . import vm
 from .blocks import Block, Header, Log, Receipt, Withdrawal, encode_receipt
 from .bloom import logs_bloom
 from .exceptions import (
+    BlobCountExceededError,
     BlobGasLimitExceededError,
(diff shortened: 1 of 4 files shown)


src/ethereum/osaka/fork.py
```diff
@@ -440,6 +442,8 @@ def check_transaction(
         version.
     NoBlobDataError :
         If the transaction is a type 3 but has no blobs.
+    BlobCountExceededError :
+        If the transaction is a type 3 and has nore blobs than the limit.
```
Carsons-Eels, line 446: We should list the current limit of 6 here so it's clear without having to look up the EIP

gurukamath: We can do that but I haven't cause we might forget to update this if it changes in the future.

PR #1326, opened 2025-07-16
chore: WELD tooling (switch over to uv+ruff)

Carsons-Eels:
### What was wrong?

In order to WELD we're going to need to align on tooling, and this is an attempt to see how painful it will be to migrate to uv and ruff, leaving behind flake8, black, and isort (the latter two are covered by rules in ruff).

Related to Issue #

### How was it fixed?

Updated the project config

#### Cute Animal Picture


pyproject.toml
```diff
@@ -152,25 +154,25 @@ packages = [
...
-    "flake8-spellcheck==0.28.0",
-    "flake8-unused-arguments==0.0.13",
+    "ruff==0.11.8",
     "vulture==2.14.0",
+    # Black and isort are taken care of by ruff
+    #    "black==23.12.0",
+    #    "isort==5.13.2",
+    # ruff does not cover spellcheck currently. It's an open issue.
+    # https://github.com/astral-sh/ruff/issues/5014
+    #    "flake8-spellcheck==0.28.0",
```
SamWilsn, line 175: Comments should describe the current state, while the commit message should be used to describe the reasoning behind changes.

```suggestion
```

pyproject.toml
```diff
@@ -207,13 +210,13 @@ ethereum-spec-evm = "ethereum_spec_tools.evm_tools:main"
...
-
-[tool.black]
-line-length = 79
+#[tool.isort]
+#profile = "black"
+#multi_line_output = 3
+#line_length = 79
+#
+#[tool.black]
+#line-length = 79
```
SamWilsn, line 219:
```suggestion
```

pyproject.toml
```diff
@@ -292,3 +295,50 @@ files = [
...
+    #    "D203",     # 1 blank line required before class docstring
+    #    "D212",     # Multi-line docstring summary should start at the first line
+    #    "D415",     # First line should end with a ".", "?", or "!"
+    #    "C420",     # Unnecessary dict comprehension
+    #    "C901",     # Function is too complex
+    #    "A005",     # stdlib module shadowing
+]
+
+[tool.uv]
+required-version = ">=0.7.0"
```
SamWilsn, line 344: This should match the version we run CI with. If there's a mismatch, there's no guarantee that we'll still work with the older version.

Carsons-Eels: I'll double check that they align tonight

network-upgrades/mainnet-upgrades/dao-fork.md
```diff
@@ -156,17 +156,17 @@ At the beginning of block 1920000, all ether throughout all accounts in `L` will
 // Deployed on mainnet at 0xbf4ed7b27f1d666546e30d74d50d173d20bca754
 
 contract DAO {
-    function balanceOf(address addr) returns (uint);
-    function transferFrom(address from, address to, uint balance) returns (bool);
-    uint public totalSupply;
+    function balanceOf(address addr) returns (Uint);
```
SamWilsn: Solidity uses lowercase `uint`, not uppercase.

pyproject.toml
```diff
@@ -295,3 +275,117 @@ files = [
...
+]
+ignore = [
+# Common to STEEL
+    "D205",     # Missing blank line after summary
+    "D203",     # 1 blank line required before class docstring
+    "D212",     # Multi-line docstring summary should start at the first line
+    "D415",     # First line should end with a ".", "?", or "!"
+
+# Specific to EELS
+    "B905",     # Using `zip` without an explicit `strict` parameter
```
SamWilsn: I think we should enable this one

pyproject.toml
```diff
@@ -295,3 +275,117 @@ files = [
...
+    "E501",         # Line too long
+]
+"src/ethereum_spec_tools/evm_tools/loaders/fork_loader.py" = [
+    "N802"          # Property names do not need to be lowercase
+]
+"src/ethereum_spec_tools/lint/*" = [
+    "N802"          # Special linting code absolved of function naming reqs
+]
+"src/ethereum/crypto/*" = [
+    "N806",         # Special crypto code absolved of variable naming reqs
```
SamWilsn, line 359: What triggers this one? I don't remember there being anything special in `crypto`.

src/ethereum/arrow_glacier/bloom.py
```diff
@@ -42,14 +42,14 @@ def add_to_bloom(bloom: bytearray, bloom_entry: Bytes) -> None:
     bloom_entry :
         An entry which is to be added to bloom filter.
     """
-    hash = keccak256(bloom_entry)
+    b_hash = keccak256(bloom_entry)
```
SamWilsn: "b" is not a commonly known abbreviation. We should spell it out in its entirety.

src/ethereum/arrow_glacier/vm/gas.py
```diff
@@ -140,7 +140,7 @@ def calculate_memory_gas_cost(size_in_bytes: Uint) -> Uint:
     try:
         return total_gas_cost
     except ValueError:
-        raise OutOfGasError
+        raise OutOfGasError from ValueError
```
SamWilsn: Huh, this does the same thing as the following?

```python
except ValueError as e:
    raise OutOfGasError from e
```

I just want to make sure that this isn't throwing away the context and replacing it with a blank parent exception.

src/ethereum/arrow_glacier/vm/instructions/keccak.py
```diff
@@ -56,9 +56,9 @@ def keccak(evm: Evm) -> None:
     # OPERATION
     evm.memory += b"\x00" * extend_memory.expand_by
     data = memory_read_bytes(evm.memory, memory_start_index, size)
-    hash = keccak256(data)
+    k_hash = keccak256(data)
```
SamWilsn: I'd just go with `hashed` maybe?

src/ethereum/byzantium/__init__.py
```diff
@@ -1,6 +1,6 @@
 """
 The Byzantium fork reduces the mining rewards, delays the difficulty bomb,
-lets contracts make non-state-changing calls to other contracts, and adds
+enables contracts to make non-state-changing calls to other contracts, and adds
```
SamWilsn, line 3: This is an odd one. Did something complain about informal grammar? 🤣

src/ethereum/ethash.py
```diff
@@ -240,7 +240,13 @@ def generate_cache(block_number: Uint) -> Tuple[Tuple[U32, ...], ...]:
                 U32.from_le_bytes(cache[index][0:4]) % U32(cache_size_words)
             ]
             result = Bytes(
-                [a ^ b for a, b in zip(first_cache_item, second_cache_item)]
+                [
```
SamWilsn: If we put the `zip(...)` into a temporary variable, can we clean this up a bit?

tox.ini
```diff
@@ -36,6 +41,8 @@ commands =
 [testenv:py3]
 extras =
     test
+setenv =
+    COVERAGE_FILE = {toxworkdir}/.coverage
```
SamWilsn: Does this require any changes in the GitHub workflows?

src/ethereum/byzantium/__init__.py
Carsons-Eels: No, I think there was no apostrophe in "lets" and I was like, "well, there's probably a more elegant way to say this. It doesn't so much allow the contract to make non-state-changing calls in the permissive sense, as in the power or functional sense.

src/ethereum/arrow_glacier/vm/instructions/keccak.py
Carsons-Eels: The linter does not like the shadowing, as hash is a builtin.

src/ethereum/arrow_glacier/vm/gas.py
Carsons-Eels: Hmmmm that might be better, I'll take a look at it. Yes, that's exactly what I was trying to prevent. We want a nice chain of responsibility

src/ethereum/arrow_glacier/bloom.py
Carsons-Eels: I can make that change to both `b_hash` and `k_hash` since the linter complains about shadowing for both

pyproject.toml
Carsons-Eels: Okay, that will lead some code changes. I thought this was more of a style thing since it defaults when unspecified

network-upgrades/mainnet-upgrades/dao-fork.md
Carsons-Eels: Oh! .... I have questions. I'll ask them in our team chat tomorrow

src/ethereum/arrow_glacier/vm/instructions/keccak.py
SamWilsn: Yep. `k_hash` had `k` in it, which is not a commonly used abbreviation for keccak256. One of our loose rules is to avoid abbreviations unless they're super common.

So I'm suggesting `hashed` to avoid shadowing `hash`, while avoiding the abbreviation.

network-upgrades/mainnet-upgrades/dao-fork.md
SamWilsn: These aren't Python snippets, they're a different language: Solidity. It uses lowercase integer types.

pyproject.toml
SamWilsn: You are indeed correct, but the default behaviour (`false`) is unexpected (stop on shortest iterable). It's too easy to accidentally assume `zip` continues until the end of the longest iterator, so this lint makes you explicitly ask for that behaviour.

pyproject.toml
Carsons-Eels: camelCase variable names. I was going to ask if we should just switch those variables over to underscore, but I figured it was simpler to ignore for now and ask if this was some kind of special case later.

src/ethereum/arrow_glacier/vm/gas.py
Carsons-Eels: Yep, good catch. It would foul up the traceback the way I have it since we would lose the instance context. I'll go through and fix these.

network-upgrades/mainnet-upgrades/dao-fork.md
```diff
@@ -156,17 +156,17 @@ At the beginning of block 1920000, all ether throughout all accounts in `L` will
 // Deployed on mainnet at 0xbf4ed7b27f1d666546e30d74d50d173d20bca754
 
 contract DAO {
-    function balanceOf(address addr) returns (uint);
-    function transferFrom(address from, address to, uint balance) returns (bool);
-    uint public totalSupply;
+    function balanceOf(address addr) returns (uInt);
```
SamWilsn: The original type was correct here (`uint`).

src/ethereum/arrow_glacier/vm/instructions/block.py
```diff
@@ -46,13 +46,13 @@ def block_hash(evm: Evm) -> None:
         # Default hash to 0, if the block of interest is not yet on the chain
         # (including the block which has the current executing transaction),
         # or if the block's age is more than 256.
-        hash = b"\x00"
+        b_hash = b"\x00"
```
SamWilsn: The "b" should be expanded, but not to `block_hash` or else it'll overwrite the function.

src/ethereum/ethash.py
```diff
@@ -239,9 +239,8 @@ def generate_cache(block_number: Uint) -> Tuple[Tuple[U32, ...], ...]:
             second_cache_item = cache[
                 U32.from_le_bytes(cache[index][0:4]) % U32(cache_size_words)
             ]
-            result = Bytes(
-                [a ^ b for a, b in zip(first_cache_item, second_cache_item)]
-            )
+            items = zip(first_cache_item, second_cache_item, strict=False)
```
SamWilsn: Hm, does `strict=False` make sense here? Are these ever supposed to be different lengths?

network-upgrades/mainnet-upgrades/dao-fork.md
Carsons-Eels: Whoops, one sec... fixed

src/ethereum/arrow_glacier/vm/instructions/block.py
Carsons-Eels: changed to `current_block_hash` to match the other vars

src/ethereum/ethash.py
Carsons-Eels: Hmmm good point. I was keeping to the default because I intended it to be a minimal change, but you can't xor a None type, so the comprehension it's tied to requires both operands anyway. I'll update it to True.

SamWilsn (dismissed):

src/ethereum/ethash.py
SamWilsn: I'm going to go out on a limb and guess that we don't ever want `strict=False`, but don't take my word for it.

SamWilsn:
Just these two left:
 - https://github.com/ethereum/execution-specs/pull/1326#discussion_r2256696185
 - https://github.com/ethereum/execution-specs/pull/1326#discussion_r2289166629

tox.ini
Carsons-Eels: I do not believe so, it refers to `.tox/` by default, but it occurs to me that we may want to have separate coverage files for each environment. If so I can modify the `tox.ini` to do that if we want... but I think I'm going to remove this for now though. No need to reconfigure it in this PR.

pyproject.toml
```diff
@@ -299,3 +279,119 @@ files = [
...
+    '^\.git/',
+    '^\.pytest_cache/',
+    '^\.ruff_cache/',
+    '^\.tox/',
+    '^\.venv/',
+    '^\.vscode/',
+    'tests/fixtures/*',
+    'eest_tests/*',
+    'vulture_whitelist.py',
+]
```
SamWilsn: Apparently these are supposed to be [globset](https://docs.rs/globset/latest/globset/#syntax)s. So this should be something like:

```suggestion
exclude = [
    ".cache/",
    ".git/",
    ".pytest_cache/",
    ".ruff_cache/",
    ".tox/",
    ".venv/",
    ".vscode/",
    "tests/fixtures/*",
    "eest_tests/*",
    "vulture_whitelist.py",
]
```

pyproject.toml
```diff
@@ -299,3 +279,119 @@ files = [
 [tool.docc.output]
 path = "docs"
 extension = ".html"
+
+[tool.vulture]
+exclude = ["tests/fixtures/", "eest_tests/"]
+ignore_names = ["pytest_*"]
+
+[tool.ruff]
+exclude = [
```
SamWilsn: Should this be `extend-exclude` instead? [The docs](https://docs.astral.sh/ruff/settings/#exclude) seem to recommend it.

pyproject.toml
```diff
@@ -299,3 +279,119 @@ files = [
...
+    '^\.cache/',
+    '^\.git/',
+    '^\.pytest_cache/',
+    '^\.ruff_cache/',
+    '^\.tox/',
+    '^\.venv/',
+    '^\.vscode/',
+    'tests/fixtures/*',
```
SamWilsn: Should we add `tests/json_infra/fixtures/*`?

pyproject.toml
```diff
@@ -299,3 +279,119 @@ files = [
...
+    "D101",         # Missing docstring in public class
+    "D103",         # Missing docstring in public function
+    "D104",         # Missing docstring in public package
+    "E501",         # Line too long
+]
+"src/ethereum_spec_tools/evm_tools/loaders/fork_loader.py" = [
+    "N802"          # Property names do not need to be lowercase
+]
+"src/ethereum_spec_tools/lint/*" = [
+    "N802"          # Special linting code absolved of function naming reqs
```
SamWilsn, line 356: Does ruff respect the [`@override`](https://typing.python.org/en/latest/spec/class-compat.html#override) decorator? I _think_ all of those methods are overrides.

pyproject.toml
```diff
@@ -299,3 +279,119 @@ files = [
...
+"src/ethereum_spec_tools/evm_tools/t8n/evm_trace/eip3155.py" = [
+    "N815"          # The traces must use camel case in JSON property names
+]
+"src/ethereum_spec_tools/evm_tools/t8n/evm_trace.py" = [
+    "N815"          # The traces must use camel case in JSON property names
+]
+
+[tool.ruff.lint.mccabe]
+# Set the maximum allowed cyclomatic complexity. C901 default is 10.
+max-complexity = 7
```
SamWilsn, line 371: Out of curiosity, is this the lowest we can go?

pyproject.toml
```diff
@@ -299,3 +279,119 @@ files = [
...
+    ".venv/*",
+    "build/*",
+    "dist/*",
+    "*.pyc",
+    "ethereum_execution.egg-info/*",
+    "*.coverage*",
+    "__pycache__",
+    ".ruff_cache",
+    "fixtures",
+    "tests/fixtures/*",
```
SamWilsn: json_infra as well?

pyproject.toml
SamWilsn, line 1: Does the `build-system` table need to be changed to support uv?

src/ethereum/trace.py
```diff
@@ -170,14 +170,18 @@ class GasAndRefund:
 
 
 def discard_evm_trace(
-    evm: object,  # noqa: U100
-    event: TraceEvent,  # noqa: U100
+    evm: object,
+    event: TraceEvent,
+    trace_memory: bool = False,
+    trace_stack: bool = True,
+    trace_return_data: bool = False,
```
SamWilsn: I think these are getting incorrectly re-added?

src/ethereum/trace.py
```diff
@@ -212,6 +216,7 @@ def __call__(
         [evm]: ref:ethereum.frontier.vm.Evm
         [`TraceEvent`]: ref:ethereum.trace.TraceEvent
         """
+        del evm, event
```
SamWilsn: Does `raise NotImplementedError` also mute the warning? I think raising that error would be more semantically correct.

src/ethereum_spec_tools/evm_tools/t8n/evm_trace.py
SamWilsn, line 1: Another rebase error?

tests/json_infra/test_trace.py
SamWilsn, line 1: Rebase

tox.ini
```diff
@@ -38,8 +43,9 @@ extras =
     test
 commands =
     fill \
+        -x \
         -m "not slow and not zkevm and not benchmark" \
-        -n auto --maxprocesses 6 \
+        -n auto --maxprocesses 10 \
```
SamWilsn: Is this intentional?

Carsons-Eels: Yeah, I thought we had updated this to 10, but then it switched back lower to 6. Was that because of impacts to the performance of the CI machine? If so, then I'll revert it

pyproject.toml
Carsons-Eels: No, it can be set lower. The example in the docs is 5. I just didn't want to set it too low and trip off a bunch of linting errors, especially with the EEST code incoming.

pyproject.toml
Carsons-Eels: It's weird. The docs say "_Note that you'll typically want to use `extend-exclude` to modify the excluded paths._" for `exclude`, but they also say "_A list of file patterns to omit from formatting and linting, **in addition** to those specified by `exclude`._" for `extend-exclude` which seems to imply that there is already an `exclude`. It's not... super clear to me.

pyproject.toml
Carsons-Eels: Oh! Yes, that makes sense. Updated.

pyproject.toml
Carsons-Eels: It's a mix. Some are, some (`Glacier_Forks_Hygiene.visit_Module` ) are not. Though, when I remove that line and run static it does not complain about any of the overridden functions. Should I just remove this ignore and fix the method names instead?

src/ethereum/trace.py
Carsons-Eels: Fixed. Good catch. Thought I properly removed those.

src/ethereum/trace.py
Carsons-Eels: Just checked, yep it does. I'll switch it over

src/ethereum_spec_tools/evm_tools/t8n/evm_trace.py
Carsons-Eels: Yeah, this file shouldn't even be in here.

tox.ini
SamWilsn: Yeah exactly. We were OOMing

pyproject.toml
SamWilsn: Yeah, seven is fine.

pyproject.toml
SamWilsn: If you change the method names, they won't get called.

pyproject.toml
SamWilsn: There's a default `exclude`. If you use `extend-exclude`, you append new entries to that list; but if you use `exclude` you replace the defaults.

Carsons-Eels: OIC. Okay, let me change that and test it... all good, updating

pyproject.toml
Carsons-Eels: Just going to leave this ignored for now.

pyproject.toml
Carsons-Eels: I don't believe so. UV should be able to hand off to setuptools to build the wheel without any additional configuration.

src/ethereum/trace.py
```diff
@@ -170,15 +170,17 @@ class GasAndRefund:
...
+    evm: object,
+    event: TraceEvent,
 ) -> None:
     """
     An [`EvmTracer`] that discards all events.
 
     [`EvmTracer`]: ref:ethereum.trace.EvmTracer
     """
 
+    raise NotImplementedError
```
SamWilsn: This one needs to be `del` because it actually gets called. If I commented on this one, my bad :sweat_smile:

tox.ini
SamWilsn: These are flipped I think? `json_infra` used to have 10 while `py3` had 6.

SamWilsn (approved):

PR #1338, opened 2025-07-21
Add Specs: EIP-7928: Block-level Access Lists

nerolation:
Add [EIP-7928: Block-level Access Lists](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7928.md) to the execution-specs.


src/ethereum/osaka/bal_builder.py
```diff
@@ -0,0 +1,152 @@
+"""
+Block Access List Builder for EIP-7928
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This module implements the BAL builder that tracks all account and storage
+accesses during block execution and constructs the final BlockAccessList.
```
SamWilsn, line 6: New contributions should use markdown-formatted docstrings. See [`blocks.py`](https://github.com/ethereum/execution-specs/blob/forks/osaka/src/ethereum/osaka/blocks.py) for a recent example.

src/ethereum/osaka/bal_builder.py
```diff
@@ -0,0 +1,152 @@
...
+    BalanceChange,
+    BlockAccessList,
+    CodeChange,
+    NonceChange,
+    SlotChanges,
+    StorageChange,
+)
+
+
+class BALBuilder:
```
SamWilsn, line 27: Probably want to avoid the abbreviation here too.

src/ethereum/osaka/bal_builder.py
```diff
@@ -0,0 +1,152 @@
...
+    Builder for constructing BlockAccessList efficiently during transaction execution.
+    
+    Follows the pattern: address -> field -> tx_index -> change
+    """
+
+    def __init__(self) -> None:
+        # address -> field_type -> changes
+        self.accounts: Dict[Address, Dict[str, any]] = {}
+
+    def _ensure_account(self, address: Address) -> None:
```
SamWilsn, line 38: We try to keep our classes as dumb as possible (basically structs) and use unbound methods instead.

src/ethereum/osaka/bal_builder.py
```diff
@@ -0,0 +1,152 @@
...
+class BALBuilder:
+    """
+    Builder for constructing BlockAccessList efficiently during transaction execution.
+    
+    Follows the pattern: address -> field -> tx_index -> change
+    """
+
+    def __init__(self) -> None:
+        # address -> field_type -> changes
+        self.accounts: Dict[Address, Dict[str, any]] = {}
```
SamWilsn, line 36: Should either use a `TypedDict` or a dataclass for the type of values in this dictionary.

src/ethereum/osaka/bal_builder.py
```diff
@@ -0,0 +1,152 @@
...
+                'balance_changes': [],  # [BalanceChange]
+                'nonce_changes': [],    # [NonceChange]
+                'code_changes': [],     # [CodeChange]
+            }
+
+    def add_storage_write(
+        self, 
+        address: Address, 
+        slot: Bytes, 
+        tx_index: int,
```
SamWilsn, line 53: Why accept an `int` here instead of a `U32`?

src/ethereum/osaka/bal_tracker.py
```diff
@@ -0,0 +1,118 @@
...
+    def track_balance_change(
+        self, 
+        address: Address, 
+        new_balance: U256, 
+        state: State
+    ) -> None:
+        """Track a balance change."""
+        self.track_address_access(address)
+        
+        # Convert U256 to 16-byte balance (uint128 - sufficient for total ETH supply)
```
SamWilsn, line 87:
```suggestion
        # Convert U256 to 16-byte balance (U128 is sufficient for total ETH supply)
```

src/ethereum/osaka/bal_tracker.py
```diff
@@ -0,0 +1,118 @@
...
+        balance_bytes = new_balance.to_be_bytes32()[-16:]  # Take last 16 bytes for uint128
```
SamWilsn, line 88: This should probably be:

```suggestion
        balance_bytes = U128(new_balance).to_be_bytes16()
```

In the unlikely event that `new_balance` is ever greater than `U128.MAX_VALUE`, we'd want to throw an exception instead of silently truncating.

This might require a change and new release of ethereum-types.

src/ethereum/osaka/bal_utils.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+def ssz_encode_list(items: tuple, encode_item_fn, max_length: int = None) -> bytes:
+    """Encode a list/tuple as SSZ with optional max length."""
+    # For variable-length lists, we need offset encoding
+    # First, encode the list length
+    result = bytearray()
+    
+    # If max_length is specified, this is a variable-length list
+    if max_length is not None:
```
SamWilsn, line 69: Instead of negating here, flip the order of the branches (so the condition becomes `max_length is None` and the else block is moved up).

src/ethereum/osaka/bal_utils.py
```diff
@@ -0,0 +1,338 @@
...
+    Returns
+    -------
+    hash :
+        The keccak256 hash of the SSZ-encoded BAL.
+    """
+    bal_bytes = ssz_encode_block_access_list(bal)
+    return keccak256(bal_bytes)
+
+
+def ssz_encode_uint(value: Union[int, Uint], size: int) -> bytes:
```
SamWilsn, line 50: The ssz functions should probably live in an `ssz` module, at the very least so we can have a docstring explaining what SSZ stands for 🤣

src/ethereum/osaka/ssz_types.py
```diff
@@ -0,0 +1,97 @@
...
+# Type aliases for clarity
+Address = Bytes20
+StorageKey = Bytes32
+StorageValue = Bytes32
+TxIndex = Uint
+Balance = Bytes  # uint128 - Post-transaction balance in wei (16 bytes, sufficient for total ETH supply)
+Nonce = Uint
+
+# Constants chosen to support a 630m block gas limit
+MAX_TXS = 30_000
```
SamWilsn, line 26:
```suggestion
MAX_TRANSACTIONS = 30_000
```

src/ethereum/osaka/bal_utils.py
```diff
@@ -0,0 +1,338 @@
...
+    
+    return bytes(result)
+
+
+def ssz_encode_storage_change(change: StorageChange) -> bytes:
+    """Encode a StorageChange as SSZ."""
+    result = bytearray()
+    result.extend(ssz_encode_uint(change.tx_index, 2))  # TxIndex as uint16
+    result.extend(ssz_encode_bytes(change.new_value))   # StorageValue as Bytes32
+    return bytes(result)
```
SamWilsn, line 112: Definitely not a fully thought out idea, but would this be more readable:

```suggestion
    return (
        ssz_encode_uint(change.tx_index, 2))  # TxIndex as uint16
        + ssz_encode_bytes(change.new_value))   # StorageValue as Bytes32
    )
```

src/ethereum/osaka/state.py
```diff
@@ -527,9 +533,23 @@ def increase_recipient_balance(recipient: Account) -> None:
...
+        
+        bal_tracker.track_balance_change(sender_address, U256(sender_new_balance), state)
+        bal_tracker.track_balance_change(recipient_address, U256(recipient_new_balance), state)
+
+
+def set_account_balance(
+    state: State, 
+    address: Address, 
+    amount: U256,
+    bal_tracker: Optional["StateChangeTracker"] = None,
```
SamWilsn, line 551: Is there somewhere where this is passed `None`? If not, we should simplify it.

src/ethereum/osaka/state.py
```diff
@@ -543,15 +563,22 @@ def set_account_balance(state: State, address: Address, amount: U256) -> None:
...
 
     modify_state(state, address, set_balance)
+    
+    # Track balance change for BAL
+    if bal_tracker is not None:
+        bal_tracker.track_balance_change(address, amount, state)
 
 
-def increment_nonce(state: State, address: Address) -> None:
+def increment_nonce(state: State, address: Address, bal_tracker: Optional["StateChangeTracker"] = None) -> None:
```
SamWilsn, line 581: Maybe `change_tracker` or `state_tracker` instead of `bal_tracker`?

src/ethereum/osaka/vm/__init__.py
```diff
@@ -22,12 +22,18 @@
...
 from ethereum.exceptions import EthereumException
 
+from ..bal_builder import BALBuilder
 from ..blocks import Log, Receipt, Withdrawal
 from ..fork_types import Address, Authorization, VersionedHash
 from ..state import State, TransientStorage
 from ..transactions import LegacyTransaction
 from ..trie import Trie
 
+# Forward declaration for type hints
```
SamWilsn, line 32:
```suggestion
```

src/ethereum/osaka/vm/instructions/environment.py
```diff
@@ -353,6 +357,10 @@ def extcodesize(evm: Evm) -> None:
 
     # OPERATION
     code = get_account(evm.message.block_env.state, address).code
+    
+    # BAL tracking for address access
```
SamWilsn, line 361: These comments don't add any information that isn't obvious from the code below them.

SamWilsn (commented):
Note to self, I haven't reviewed the tests yet.

SamWilsn:
I've rebased `forks/osaka` to pick up the last straggling commits from `master`. If you do something like this, should be easy enough to resolve the conflicts:

```bash
git checkout forks/osaka
git pull
git checkout bals
git reset --hard forks/osaka
git cherry-pick a24d3837cde502cb3133cbe37a5c8871ca516525^..3d986436ca1d9c406a154e3549d3b415ddf88b3b
```

SamWilsn:
Actually, there weren't any conflicts. I took the liberty of rebasing your branch. Please confirm everything looks the way it should!

nerolation:
yeah looks good, what I can tell.

src/ethereum/osaka/state.py
```diff
@@ -503,9 +509,21 @@ def increase_recipient_balance(recipient: Account) -> None:
...
     modify_state(state, recipient_address, increase_recipient_balance)
-
-
-def set_account_balance(state: State, address: Address, amount: U256) -> None:
+    
+    if change_tracker is not None:
+        sender_new_balance = get_account(state, sender_address).balance
+        recipient_new_balance = get_account(state, recipient_address).balance
+        
+        change_tracker.track_balance_change(sender_address, U256(sender_new_balance), state)
```
fselmo: Here we are expecting methods on the `change_tracker` itself, but the way it's currently implemented, the methods are [standalone methods](https://github.com/ethereum/execution-specs/pull/1338/files#diff-bf1fcde011d1f672973f27cd5a023f0921d07b186d5742e76e8d8e77c04b639eR217) with the tracker as the first argument.

I lean toward this syntax here where the methods are called on the class itself but we should reconcile this one way or the other. I wouldn't take my suggestion though, let's see what EELS would prefer cc: @SamWilsn

SamWilsn: We avoid methods everywhere, instead preferring a more C-like struct/function style.

kevaundray: I think this PR might overlap with https://github.com/ethereum/execution-specs/pull/1371 re state tracking

nerolation: Moved to the more C-like style. LMK what you think

gurukamath:
@nerolation Could we close this and continue the BAL work on #1381 ?

PR #1343, opened 2025-07-23
chore(git-history) split commit: add specific exceptions on tests

souradeep-das:
### What was wrong?

The existence of a single commit for "changes to custom exceptions across all forks" did not allow it to be usable for `master`.

Associated PR: https://github.com/ethereum/execution-specs/pull/1286

### How was it fixed?
Split commit https://github.com/ethereum/execution-specs/commit/7b1694e9a0b7647c6d53d1bddbf96454a875e7fa
into two commits, such that [2d29ba7](https://github.com/ethereum/execution-specs/pull/1343/commits/2d29ba7c50e055236ef3a5775fa2aecbf2908e41) can be applied to master

#### Cute Animal Picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![GitPOAP Badge](https://public-api.gitpoap.io/v1/repo/ethereum/execution-specs/badge)](https://www.gitpoap.io/gh/ethereum/execution-specs)
 [![codecov](https://codecov.io/gh/ethereum/execution-specs/graph/badge.svg?token=0LQZO56RTM)](https://codecov.io/gh/ethereum/execution-specs)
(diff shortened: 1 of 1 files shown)


SamWilsn:
Just to double check, the intent is to force push this to `forks/osaka`?

souradeep-das:
> Just to double check, the intent is to force push this to forks/osaka?

Yes, if we decided to move forward with this, this will need a force push to `forks/osaka` 
I can close the PR then

Carsons-Eels (approved):
Looks good!

I think 
```
chore(git-history) split commit: add specific exceptions on tests
```
would be more in line with the EEST conventions that we should probably adopt for commit messages. Though, we could also look at adding other labels for the commits rather than putting it in the commit message... 

Anyway, just make sure to pull in the latest changes before force pushing. I just merged a small PR.

PR #1353, opened 2025-07-29
Count opcodes

### LINKED ISSUE #1351: Add a trace option that counts executed opcodes
See https://github.com/ipsilon/evmone/pull/1274 for an implementation.

The general shape of this interface is:

- Add a t8n command line argument `--opcode.count <file-name>`.
- Save a JSON blob to the given file containing the number of times each instruction is executed.

Sample JSON:

```json
{
  "STOP": 1,
  "ADD": 1,
  "MOD": 1,
  "EQ": 1,
  "CALLER": 1,
  "CALLDATALOAD": 1,
  "TIMESTAMP": 2,
  "SSTORE": 2,
  "JUMPI": 1,
  "JUMPDEST": 1,
  "PUSH0": 2,
  "PUSH1": 2,
  "PUSH3": 2,
  "PUSH20": 1,
  "DUP2": 1,
  "SWAP1": 1,
  "RETURN": 1
}
```

SamWilsn:
~~Draft because there are some open questions about how to expose the feature.~~

Ready to merge, probably after #1356 

### What was wrong?

closes #1351

### How was it fixed?

Added a counting trace implementation.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/daemon.py
+++ b/src/ethereum_spec_tools/evm_tools/daemon.py
@@ -42,56 +42,68 @@ class _EvmToolHandler(BaseHTTPRequestHandler):
     def do_POST(self) -> None:
         from . import main
 
-        content_length = int(self.headers["Content-Length"])
-        content_bytes = self.rfile.read(content_length)
-        content = json.loads(content_bytes)
+        try:
+            content_length = int(self.headers["Content-Length"])
+            content_bytes = self.rfile.read(content_length)
+            content = json.loads(content_bytes)
 
-        input_string = json.dumps(content["input"])
-        input = StringIO(input_string)
+            input_string = json.dumps(content["input"])
+            input = StringIO(input_string)
 
-        args = [
-            "t8n",
-            "--input.env=stdin",
-            "--input.alloc=stdin",
-            "--input.txs=stdin",
-            "--output.result=stdout",
-            "--output.body=stdout",
-            "--output.alloc=stdout",
-            f"--state.fork={content['state']['fork']}",
-            f"--state.chainid={content['state']['chainid']}",
-            f"--state.reward={content['state']['reward']}",
-        ]
+            args = [
+                "t8n",
+                "--input.env=stdin",
+                "--input.alloc=stdin",
+                "--input.txs=stdin",
+                "--output.result=stdout",
+                "--output.body=stdout",
+                "--output.alloc=stdout",
+                f"--state.fork={content['state']['fork']}",
+                f"--state.chainid={content['state']['chainid']}",
+                f"--state.reward={content['state']['reward']}",
+            ]
 
-        trace = content.get("trace", False)
-        output_basedir = content.get("output-basedir")
-        if trace:
-            if not output_basedir:
-                raise ValueError(
-                    "`output-basedir` should be provided when `--trace` "
-                    "is enabled."
+            trace = content.get("trace", False)
+            output_basedir = content.get("output-basedir")
+            if trace:
+                if not output_basedir:
+                    raise ValueError(
+                        "`output-basedir` should be provided when `--trace` "
+                        "is enabled."
+                    )
+                # send full trace output if ``trace`` is ``True``
+                args.extend(
+                    [
+                        "--trace",
+                        "--trace.memory",
+                        "--trace.returndata",
+                        f"--output.basedir={output_basedir}",
+                    ]
                 )
-            # send full trace output if ``trace`` is ``True``
-            args.extend(
-                [
-                    "--trace",
-                    "--trace.memory",
-                    "--trace.returndata",
-                    f"--output.basedir={output_basedir}",
-                ]
-            )
 
-        query_string = urlparse(self.path).query
-        if query_string:
-            query = parse_qs(
-                query_string,
-                keep_blank_values=True,
-                strict_parsing=True,
-                errors="strict",
-            )
-            args += query.get("arg", [])
+            count_opcodes = content.get("count-opcodes", False)
+            if count_opcodes:
+                # send opcode counts if ``count-opcodes`` is ``True``
+                args.extend(["--opcodes.count", "stdout"])
+
+            query_string = urlparse(self.path).query
+            if query_string:
+                query = parse_qs(
+                    query_string,
+                    keep_blank_values=True,
+                    strict_parsing=True,
+                    errors="strict",
+                )
+                args += query.get("arg", [])
+        except Exception as e:
+            self.send_response(500)
+            self.send_header("Content-Type", "text/plain")
+            self.end_headers()
+            self.wfile.write(str(e).encode("utf-8"))
+            raise
 
         self.send_response(200)
-        self.send_header("Content-type", "application/octet-stream")
+        self.send_header("Content-Type", "application/octet-stream")
         self.end_headers()
 
         # `self.wfile` is missing the `name` attribute so it doesn't strictly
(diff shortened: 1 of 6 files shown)


src/ethereum_spec_tools/evm_tools/daemon.py
```diff
@@ -80,6 +80,11 @@ def do_POST(self) -> None:
                 ]
             )
 
+        count_opcodes = content.get("count-opcodes", False)
+        if count_opcodes:
+            # send full opcode counts if ``count-opcodes`` is ``True``
```
Carsons-Eels: The "full" is a bit misleading in that it implies that some smaller portion of opcodes will be counted if ``False``

```python
# Send opcode counts only if ``count-opcodes`` is ``True``
```

src/ethereum_spec_tools/evm_tools/daemon.py
```diff
@@ -80,6 +80,11 @@ def do_POST(self) -> None:
                 ]
             )
 
+        count_opcodes = content.get("count-opcodes", False)
```
Carsons-Eels: `content` is derived from an unsafe `json.loads` call that could raise a `json.JSONDecodeError` (`ValueError`). TIt might be better/safer to have line 47 wrapped in a try-except block.

tests/conftest.py
```diff
@@ -26,6 +26,14 @@ def get_xdist_worker_id(request_or_session: object) -> str:  # noqa: U100
         return "master"
 
 
+@fixture()
+def root_relative() -> Callable[[str | Path], Path]:
```
Carsons-Eels: Should probably have a small doc here for the fixture.

src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class CountTracer(EvmTracer):
+    """
+    EVM trace implementation that counts how many times each opcode is
+    executed.
+    """
+
+    transaction_environment: object | None
+    active_traces: dict[str, int]
```
Carsons-Eels: If we're going to use defaultdict later, shouldn't we specify that same type here for consistency and initialize it as such in the constructor? Is there a reason to mix the types?

src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py
```diff
@@ -0,0 +1,46 @@
...
+
+        if self.transaction_environment is not evm.message.tx_env:
+            self.active_traces = defaultdict(lambda: 0)
+            self.transaction_environment = evm.message.tx_env
+
+        self.active_traces[event.op.name] += 1
+
+    def results(self) -> dict[str, int]:
+        """
+        Take and clear the current opcode counts.
```
Carsons-Eels:
```suggestion
        Clear and return the current opcode counts.
```

src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py
```diff
@@ -0,0 +1,46 @@
...
+from .protocols import Evm
+
+
+class CountTracer(EvmTracer):
+    """
+    EVM trace implementation that counts how many times each opcode is
+    executed.
+    """
+
+    transaction_environment: object | None
```
Carsons-Eels, line 17: Could this not be narrowed to `TransactionEnvironment`?

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -127,6 +150,15 @@ def __init__(
             self.env.block_difficulty, self.env.base_fee_per_gas
         )
 
+    def _tracer(self, type_: Type[T]) -> T:
+        group = self.tracers
+        if group is None:
+            raise Exception("no tracer configured")
+        found = next(x for x in group.tracers if isinstance(x, type_))
+        if found is None:
+            raise Exception("no tracer found")
```
Carsons-Eels:
```suggestion
            raise Exception("no tracer found of type {type_}")
```

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -127,6 +150,15 @@ def __init__(
             self.env.block_difficulty, self.env.base_fee_per_gas
         )
 
+    def _tracer(self, type_: Type[T]) -> T:
+        group = self.tracers
+        if group is None:
+            raise Exception("no tracer configured")
+        found = next(x for x in group.tracers if isinstance(x, type_))
```
Carsons-Eels:
```suggestion
        found = next((x for x in group.tracers if isinstance(x, type_)), None)
```

Carsons-Eels (commented):
Just a few minor things

src/ethereum_spec_tools/evm_tools/daemon.py
SamWilsn: What behavior would you recommend in the exception handler?

src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py
SamWilsn: `defaultdict` is a subclass of `dict`, and so presents the same interface, so this assignment is sound. That said, `defaultdict` would be more specific 🤷

src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py
SamWilsn: Could be, but why? The function can operate with any object because it only cares about object identity.

src/ethereum_spec_tools/evm_tools/daemon.py
Carsons-Eels: Something like this? My instinct would be to provide a meaningful failure response before returning since if serialization fails, the POST cannot proceed meaningfully, then return. 

```python
try:
   content = json.loads(content_bytes)
except json.JSONDecodeError as jde:
    self.send_response(400)
    self.send_header("Content-Type", "application/json")
    self.end_headers()
    self.wfile.write(b'{"error": "Invalid request body: cannot serialize to JSON"}')
    return
    ```

File: src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py (hunk 4 above)
Comment (Carsons-Eels) [reply src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py]: Yeah, that was my thinking. That and type consistency. Basically IMO it makes the code slightly more readable.

File: src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py (hunk 6 above)
Comment (Carsons-Eels) [reply src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py:17]: _Principle of Least Responsibility_ was what I was thinking.

True, it's only ever used for a test, and that test can only really fail the first time `__call__` happens I think. Like I said, little stuff.

Comment (SamWilsn) [reply src/ethereum_spec_tools/evm_tools/t8n/evm_trace/count.py:17]: I'm not sure that's applicable here. Maybe I'm not seeing it?

But I would argue that a type annotation should be as general as the corresponding implementation allows. `CountTracer` doesn't require any of the fields or methods of `TransactionEnvironment`, so making the annotation more specific serves no benefit. It would just make the code less readable.

Unless you're talking about removing the `| None`. We need the `| None` to handle the path where `results` is called, clearing the set environment.

File: src/ethereum_spec_tools/evm_tools/t8n/__init__.py [hunk 9]
```diff
@@ -72,12 +76,16 @@ def t8n_arguments(subparsers: argparse._SubParsersAction) -> None:
     t8n_parser.add_argument("--trace.nostack", action="store_true")
     t8n_parser.add_argument("--trace.returndata", action="store_true")
 
+    t8n_parser.add_argument("--opcode.count", dest="opcode_count", type=str)
```
Carsons-Eels, line 79: Just noticed this. It's called `--opcode.count` here and `--opcodes.count` on line 87 of daemon.py

SamWilsn: Good catch!

Carsons-Eels (approved):

Issue #1354, opened 2025-07-29
Redundant account destruction and existence check after state modification


EperezOk:
### What was wrong?

The [`modify_state`](https://github.com/ethereum/execution-specs/blob/4198b9c5996713b268aed602739d5aa40e277694/src/ethereum/prague/state.py#L499) function in `state.py` updates an account in the chain state. Additionally, if the account exists after the modification and is empty, it is [destroyed](https://github.com/ethereum/execution-specs/blob/4198b9c5996713b268aed602739d5aa40e277694/src/ethereum/prague/state.py#L506-L507).

However, there are two cases where this destruction step is redundantly applied:

* In [`process_transaction`](https://github.com/ethereum/execution-specs/blob/4198b9c5996713b268aed602739d5aa40e277694/src/ethereum/prague/fork.py#L812), [when](https://github.com/ethereum/execution-specs/blob/4198b9c5996713b268aed602739d5aa40e277694/src/ethereum/prague/fork.py#L949-L956) the coinbase balance plus the mining fee is zero. Calling [`set_account_balance`](https://github.com/ethereum/execution-specs/blob/4198b9c5996713b268aed602739d5aa40e277694/src/ethereum/prague/state.py#L532)— even with a zero amount — [triggers](https://github.com/ethereum/execution-specs/blob/4198b9c5996713b268aed602739d5aa40e277694/src/ethereum/prague/state.py#L551) `modify_state`, which already handles the potential destruction.

* In [`process_withdrawals`](https://github.com/ethereum/execution-specs/blob/4198b9c5996713b268aed602739d5aa40e277694/src/ethereum/prague/fork.py#L980), where the operation is [performed](https://github.com/ethereum/execution-specs/blob/4198b9c5996713b268aed602739d5aa40e277694/src/ethereum/prague/fork.py#L999-L1002) immediately after a direct call to `modify_state`.

In both cases, removing the redundant explicit existence check and destruction would improve code readability and efficiency.

EperezOk:
The issue has been fixed in PR [#1355](https://github.com/ethereum/execution-specs/pull/1355)

PR #1355, opened 2025-07-29
Fix: Redundant account destruction after state modification

EperezOk:
### What was wrong?

Related to Issue #1354 

### How was it fixed?

Removed redundant checks and account destruction operations.

#### Cute Animal Picture

Jiraffe


EperezOk:
@SamWilsn please let me know if anything else is needed from my side or if I should mark the PR as "ready for review"

SamWilsn:
Hey! Thanks for the fix.

If you're feeling lazy, you can leave this and we'll do the rest of the work ourselves eventually. Otherwise, it'll need to be ported to the other post-merge forks before we can merge.

If you're feeling super enthusiastic, might be worth investigating whether we can remove `account_exists_and_is_empty` entirely from post-merge forks.

EperezOk:
Hi @SamWilsn!

I ported the change to the other post-merge forks. I also confirmed that `account_exists_and_is_empty` is no longer needed in those forks, so I removed the function and inlined it within `modify_state`.

SamWilsn (approved):
This makes sense as far as I can tell :3

PR #1358, opened 2025-08-01
Fix: Documentation Clarity

EperezOk:
### What was wrong?

There were some inaccuracies and typos in the docstrings.

### How was it fixed?

I've corrected the identified instances.

#### Cute Animal Picture

Penguins

### DIFF
--- a/src/ethereum/prague/state.py
+++ b/src/ethereum/prague/state.py
@@ -125,7 +125,7 @@ def rollback_transaction(
 ) -> None:
     """
     Rollback a state transaction, resetting the state to the point when the
-    corresponding `start_transaction()` call was made.
+    corresponding `begin_transaction()` call was made.
 
     Parameters
     ----------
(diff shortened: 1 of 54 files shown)


src/ethereum/prague/state.py
```diff
@@ -478,7 +478,7 @@ def account_exists_and_is_empty(state: State, address: Address) -> bool:
 
 def is_account_alive(state: State, address: Address) -> bool:
     """
-    Check whether is an account is both in the state and non empty.
+    Check whether an account is both in the state and non-empty.
```
SamWilsn, line 457: I put back the "both" here. I think it's good to emphasize that both conditions must be true.

SamWilsn (commented):
This looks good. Just needs to be applied to all the forks.

EperezOk:
@SamWilsn all forks have been updated!

SamWilsn (approved):

SamWilsn (approved):

Issue #1124, opened 2025-02-13
Refactor the tests to use the Pydantic models from EEST


SamWilsn:

SamWilsn:
Depends on https://github.com/ethereum/execution-spec-tests/issues/1234

PR #1359, opened 2025-08-01
utilize pydantic models from eest on t8n

souradeep-das:
### What was wrong?

Related to Issue #https://github.com/ethereum/execution-specs/issues/1124

### How was it fixed?
- Pydantic models from eest are input/output for t8n
- utilize these models to run t8n

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
@@ -15,7 +15,7 @@ from ethereum import trace
 from ethereum.exceptions import EthereumException, InvalidBlock
 from ethereum_clis.types import TransitionToolRequest
 from ethereum_spec_tools.forks import Hardfork
-from ethereum_clis.types import Result as TransitionToolOutput
+from ethereum_clis.types import TransitionToolOutput
 from ethereum_clis.clis.execution_specs import ExecutionSpecsExceptionMapper
 
 from ..loaders.fixture_loader import Load
@@ -93,6 +93,7 @@ class T8N(Load):
             self.forks, self.options, in_file.input.env
         )
         self.fork = ForkLoad(fork_module)
+        self.exception_mapper = ExecutionSpecsExceptionMapper()
 
         if self.options.trace:
             trace_memory = getattr(self.options, "trace.memory", False)
@@ -108,7 +109,6 @@ class T8N(Load):
                 )
             )
         self.logger = get_stream_logger("T8N")
-        self.exception_mapper = ExecutionSpecsExceptionMapper()
 
         super().__init__(
             self.options.state_fork,
--- a/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
@@ -4,9 +4,10 @@ Create a transition tool for the given fork.
 
 import argparse
 import fnmatch
+import json
 import os
 from functools import partial
-from typing import Any, TextIO
+from typing import Any, Optional, TextIO, Union
 
 from ethereum_rlp import rlp
 from ethereum_types.numeric import U64, U256, Uint
@@ -23,6 +24,7 @@ from ..loaders.fork_loader import ForkLoad
 from ..utils import (
     FatalException,
     get_module_name,
+    get_module_name_json_input,
     get_stream_logger,
     parse_hex_or_int,
 )
@@ -82,18 +84,72 @@ class T8N(Load):
     """The class that carries out the transition"""
 
     def __init__(
-        self, options: Any, out_file: TextIO, in_file: TransitionToolRequest
+        self,
+        options: Any,
+        in_file: Union[TransitionToolRequest, TextIO],
+        out_file: Optional[TextIO] = None
     ) -> None:
         self.out_file = out_file
         self.in_file = in_file
         self.options = options
         self.forks = Hardfork.discover()
 
+        if isinstance(in_file, TransitionToolRequest):
+            self._init_from_pydantic(options, in_file)
+        else:
+            if out_file is None:
+                raise ValueError("out_file is required for JSON file input")
+            self._init_from_json(options, in_file, out_file)
+
+    def _init_from_pydantic(self, options: Any, in_file: TransitionToolRequest) -> None:
         fork_module, self.fork_block = get_module_name(
-            self.forks, self.options, in_file.input.env
+            self.forks, options, in_file.input.env
         )
         self.fork = ForkLoad(fork_module)
         self.exception_mapper = ExecutionSpecsExceptionMapper()
+
+        if options.trace:
+            trace_memory = getattr(options, "trace.memory", False)
+            trace_stack = not getattr(options, "trace.nostack", False)
+            trace_return_data = getattr(options, "trace.returndata")
+            trace.set_evm_trace(
+                partial(
+                    evm_trace,
+                    trace_memory=trace_memory,
+                    trace_stack=trace_stack,
+                    trace_return_data=trace_return_data,
+                    output_basedir=options.output_basedir,
+                )
+            )
+        self.logger = get_stream_logger("T8N")
+
+        super().__init__(
+            options.state_fork,
+            fork_module,
+        )
+
+        self.chain_id = parse_hex_or_int(options.state_chainid, U64)
+        self.alloc = Alloc(self, in_file.input.alloc)
+        self.env = Env(self, in_file.input.env)
+        self.txs = Txs(self, in_file.input.txs)
+        self.result = Result(
+            self.env.block_difficulty, self.env.base_fee_per_gas
+        )
+
+    def _init_from_json(self, options: Any, in_file: TextIO, out_file: TextIO) -> None:
+        if "stdin" in (
+            options.input_env,
+            options.input_alloc,
+            options.input_txs,
+        ):
+            stdin = json.load(in_file)
+        else:
+            stdin = None
+
+        fork_module, self.fork_block = get_module_name_json_input(
+            self.forks, self.options, stdin
+        )
+        self.fork = ForkLoad(fork_module)
 
         if self.options.trace:
             trace_memory = getattr(self.options, "trace.memory", False)
@@ -116,9 +172,9 @@ class T8N(Load):
         )
 
         self.chain_id = parse_hex_or_int(self.options.state_chainid, U64)
-        self.alloc = Alloc(self, in_file.input.alloc)
-        self.env = Env(self, in_file.input.env)
-        self.txs = Txs(self, in_file.input.txs)
+        self.alloc = Alloc(self, stdin)
+        self.env = Env(self, stdin)
+        self.txs = Txs(self, stdin)
         self.result = Result(
             self.env.block_difficulty, self.env.base_fee_per_gas
         )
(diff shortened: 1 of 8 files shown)


src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -29,23 +29,20 @@ class Alloc:
...
+            for addr, acc in stdin.root.items()
+        }
+        for acc in alloc_json.values():
+            if acc is not None and "storage" in acc and acc["storage"] is not None:
+                # Ensure all storage values are hex strings with 0x prefix
+                acc["storage"] = {
+                    k: v if (isinstance(v, str) and v.startswith("0x")) else hex(v)
+                    for k, v in acc["storage"].items()
+                }
+        state = t8n.json_to_state(alloc_json)
```
souradeep-das: This could be made more efficient without having to convert from pydantic to json and then json to state

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -311,45 +305,13 @@ def run(self) -> int:
...
-        if json_output:
-            json.dump(json_output, self.out_file, indent=4)
-
-        return 0
+        txs_rlp = "0x" + rlp.encode(self.txs.all_txs).hex()
+        json_output["body"] = txs_rlp
+        json_output["alloc"] = json_state
+        json_output["result"] = json_result
+        output: TransitionToolOutput = TransitionToolOutput.model_validate(
+            json_output, context={"exception_mapper": self.exception_mapper}
```
souradeep-das: This could potentially also be simplified without having to convert json and then json to pydantic

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -80,26 +82,18 @@ class T8N(Load):
     """The class that carries out the transition"""
 
     def __init__(
-        self, options: Any, out_file: TextIO, in_file: TextIO
+        self, options: Any, out_file: TextIO, in_file: TransitionToolRequest
```
SamWilsn: Does this require [changes in the daemon](https://github.com/ethereum/execution-specs/blob/1537b0192fbb3cfb3aa4677659e425519a781431/src/ethereum_spec_tools/evm_tools/daemon.py#L103)?

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -311,45 +305,13 @@ def run(self) -> int:
...
-
-        return 0
+        txs_rlp = "0x" + rlp.encode(self.txs.all_txs).hex()
+        json_output["body"] = txs_rlp
+        json_output["alloc"] = json_state
+        json_output["result"] = json_result
+        output: TransitionToolOutput = TransitionToolOutput.model_validate(
+            json_output, context={"exception_mapper": self.exception_mapper}
+        )
+        return output
```
SamWilsn: Hm, so this change would break the command-line output? Is that what we want?

src/ethereum_spec_tools/evm_tools/t8n/env.py
```diff
@@ -212,30 +188,26 @@ def read_randao(self, data: Any, t8n: "T8N") -> None:
...
-            )
+            raw_withdrawals = getattr(data, "withdrawals", None)
+            if raw_withdrawals:
+                def to_canonical_withdrawal(raw):
+                    return t8n.fork.Withdrawal(
+                        index=U64(raw.index),
+                        validator_index=U64(raw.validator_index),
+                        address=raw.address,
+                        amount=U256(raw.amount),
+                    )
```
SamWilsn, line 398: Should probably hoist this out of `read_withdrawals` to avoid creating a bunch of closures that aren't used.

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -248,6 +239,154 @@ def sign_transaction(self, json_tx: Any) -> None:
...
+            )
+            for entry in access_list
+        ]
+
+    def convert_authorizations(auth_list):
+        if not auth_list:
+            return []
+        AuthorizationCls = getattr(fork, "Authorization", None)
+        if AuthorizationCls is None:
+            from ethereum.osaka.fork_types import Authorization as AuthorizationCls
```
SamWilsn: If the fork doesn't have an `Authorization` type, but the test transaction does, we should probably error, no? Shoving an osaka type into a frontier block is going to lead to at least an exception somewhere.

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -248,6 +239,154 @@ def sign_transaction(self, json_tx: Any) -> None:
...
+        if val is None:
+            return Bytes0()
+        if isinstance(val, Bytes20):
+            return val
+        if isinstance(val, (bytes, bytearray)) and len(val) == 20:
+            return Bytes20(val)
+        if isinstance(val, (bytes, bytearray)) and len(val) == 0:
+            return Bytes20(b"\x00" * 20)
+        if isinstance(val, (bytes, bytearray)):
+            return Bytes20(val.rjust(20, b"\x00"))
```
SamWilsn: Are these special cases necessary or does the `rjust` handle them all properly?

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -248,6 +239,154 @@ def sign_transaction(self, json_tx: Any) -> None:
...
+        return Bytes20(bytes(val).rjust(20, b"\x00"))
+
+    # build the canonical transaction
+    if tx_cls.__name__ == "FeeMarketTransaction":
```
SamWilsn: Can we use `issubclass` here instead?

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -248,6 +239,154 @@ def sign_transaction(self, json_tx: Any) -> None:
...
+        return tx_cls(
+            chain_id=U64(tx.chain_id),
```
SamWilsn, line 379: Instead of repeating most of the arguments in every branch, would it make sense to do something like:

```python
arguments = {
    'nonce': U256(tx.nonce),
    # The rest of the common items...
}

if issubclass(tx_cls, AccessListTransaction):
    arguments["access_list"] = convert_access_list(tx.access_list)
elif issubclass(tx_cls, BlobTransaction):
    ...

return tx_cls(**arguments)
```

src/ethereum_spec_tools/evm_tools/t8n/transition_tool.py
```diff
@@ -88,15 +87,15 @@ def evaluate(
...
+        output = t8n.run()
 
-        output_dict = json.loads(out_stream.getvalue())
-        output: TransitionToolOutput = TransitionToolOutput.model_validate(
-            output_dict, context={"exception_mapper": self.exception_mapper}
-        )
+        # output_dict = json.loads(out_stream.getvalue())
+        # output: TransitionToolOutput = TransitionToolOutput.model_validate(
+        #     output_dict, context={"exception_mapper": self.exception_mapper}
+        # )
```
SamWilsn: Remove code instead of commenting it out.

SamWilsn (commented):
So my review amounts basically to just a general Python review.

@gurukamath would you mind giving this a review with an eye specifically to design/architecture, since you've been working in this area lately?

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -80,26 +82,18 @@ class T8N(Load):
...
     def __init__(
-        self, options: Any, out_file: TextIO, in_file: TextIO
+        self, options: Any, out_file: TextIO, in_file: TransitionToolRequest
     ) -> None:
         self.out_file = out_file
         self.in_file = in_file
         self.options = options
         self.forks = Hardfork.discover()
 
-        if "stdin" in (
```
gurukamath: Would we still be able to consume json files if needed? We will need to continue supporting that use case.

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
souradeep-das: I think now a direct conversion or individually validating each part of `TransitionToolOutput` wouldn't really add any efficiency gains here

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -80,13 +84,59 @@ class T8N(Load):
...
+            )
+        self.logger = get_stream_logger("T8N")
+
+        super().__init__(
+            options.state_fork,
+            fork_module,
+        )
+
+        self.chain_id = parse_hex_or_int(options.state_chainid, U64)
+        self.alloc = Alloc(self, in_file.input.alloc)
```
gurukamath, line 132: Perhaps we should re-use the Alloc, Env and Txs types defined by EEST. And if the json inputs are provided, we could just serialize/de-serialize using pydantic. That way, we can get of a lot of redundant code in `env.py` and `t8n_types.py`.

gurukamath (changes requested):

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
souradeep-das: yes! with this commit https://github.com/ethereum/execution-specs/pull/1359/commits/dc31f962bfa8cc35d11b8615a52dbac70cead535 I added the dual input option where we preserve the older json input route

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
souradeep-das: right! :) The first thing i tried was to re-use the types (pydantic models) from EEST (to not have these intermediary classes at all) But seems like our execution-spec code isn't directly able to work with those (without some translations/conversions)? To be able to utilize the EEST types directly would definitely be the most efficient - any suggestions?

gurukamath: Would something like this work?

1. If EEST types are the input: EEST types -> EELS BlockEnvironment -> run the t8n tool
2. If json input: json input -> EEST types -> EELS BlockEnvironment -> run the t8n tool

You might have to transform the EEST `Alloc` to the relevant fork's `State` dataclass but I think this is still much cleaner than what we currently have.

SamWilsn:
I think we still need to figure out what we're doing for this PR, so don't worry about rebasing onto `forks/amsterdam` for now.

gurukamath:
Just mentioning #1883  here since it looks to tackle a similar issue

felix314159:
@souradeep-das the issue guru linked is still relevant, are you willing to rebase and move forward with this PR or should it be closed?

gurukamath:
> @souradeep-das the issue guru linked is still relevant, are you willing to rebase and move forward with this PR or should it be closed?

We are working through some re-factors in #2381 post which, it should be a little easier to use the pydantic models. I am tracking this PR and will re-factor/re-base if needed

gurukamath:
Post merging #2750, we now have all the pieces in `forks/amsterdam` to be able to share the `Alloc` type between the testing framework and `t8n`. That re-factor will be targeted in an upcoming PR

PR #1360, opened 2025-08-01
re-factor(weld): move current tests to json_infra 

### LINKED ISSUE #1352: Move json_infra tests into their own folder

gurukamath:
(closes #1352)

### What was wrong?
Currently, every fork has its own tests folder. So does the EEST repository. In a post-weld world, this could get really messy. Hence, `execution-specs` need to have a simpler test folder structure to enable the EEST test folders to seamlessly move into the repository.

Related to Issue #1352

### How was it fixed?

This PR performs the following changes
1. Re-factors and moves all the fork specific tests (primarily the blockchain tests and the state tests) in the `json_infra` folder
2. Make the test running less verbose and more maintainable.
3. Removes the difficulty tests which were not running until now anyway in the CI
4. Removes the fork specific rlp tests. There is enough rlp encoding and decoding happening in the json fixture tests that these tests are largely redundant.

#### Cute Animal Picture

Cute Animals - 1 of 1


SamWilsn:
> Removes the fork specific rlp tests. There is enough rlp encoding and decoding happening in the json fixture tests that these tests are largely redundant.

Is there anything we should move into [`ethereum-rlp`](https://github.com/ethereum/ethereum-rlp)?

tests/json_infra/__init__.py
SamWilsn, line 1: Something about how these constants are structured rubs me the wrong way.

tests/json_infra/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+
+OSAKA_DIRS = (
+    [BC_ETHEREUM_TESTS, EEST_OSAKA_BC_TESTS],
+    [STATE_ETHEREUM_TESTS, EEST_OSAKA_STATE_TESTS],
+)
+
+FORKS = {
+    **{
+        fork: _create_fork_config(package, *PRE_CONSTANTINOPLE_DIRS)
+        for fork, package in [
```
SamWilsn: `package` here is unhelpful as a variable name. What do these tuples represent?

tests/json_infra/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+            ("Frontier", "frontier"),
+            ("Homestead", "homestead"),
+            ("EIP150", "tangerine_whistle"),
+            ("EIP158", "spurious_dragon"),
+            ("Byzantium", "byzantium"),
+            ("ConstantinopleFix", "constantinople"),
```
SamWilsn, line 134: Should we move this information into the fork itself, maybe in the same file as `FORK_CRITERIA`? Then we could expose it through the `Hardfork` class.

tests/json_infra/conftest.py
```diff
@@ -15,7 +15,10 @@
 from requests_cache.backends.sqlite import SQLiteCache
 from typing_extensions import Self
 
-from tests.helpers import TEST_FIXTURES
+from . import TEST_FIXTURES
+
+# Global variable to store pytest config for access during test collection
+pytest_config = None
```
SamWilsn: I assume you can't use [`pytestconfig`](https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-pytestconfig) during collection?

tests/json_infra/test_blockchain_tests.py
```diff
@@ -0,0 +1,50 @@
...
+        # If specified fork is not valid, generate no tests
+        forks_to_test = []
+else:
+    # If no --fork option, generate tests for all forks
+    forks_to_test = list(FORKS.keys())
+
+for fork_name in forks_to_test:
+    locals()[
+        f"test_blockchain_tests_{fork_name.lower()}"
+    ] = _generate_test_function(fork_name)
```
SamWilsn: Could we use pytest marks for this instead, or is it important that we only create the functions for the given fork(s)?

tests/json_infra/test_ethash.py
```diff
@@ -25,14 +21,30 @@
...
+POW_FORKS = [
+    "Frontier",
+    "Homestead",
+    "EIP150",
+    "EIP158",
+    "Byzantium",
+    "ConstantinopleFix",
+    "Istanbul",
+    "Berlin",
+]
```
SamWilsn, line 39: This we can definitely get from `Hardfork`, unless there's an argument not to.

tests/json_infra/helpers/exceptional_test_patterns.py
```diff
@@ -0,0 +1,104 @@
+from typing import Tuple
+
+
+def get_exceptional_blockchain_test_patterns(
+    network: str, package: str
+) -> Tuple[tuple[str, ...], tuple[str, ...], tuple[str, ...]]:
```
SamWilsn: We should use a dataclass for complex return types:

```python
@dataclass
class BlockchainTestPatterns:
    slow: Tuple[str, ...]
    expected_fail: Tuple[str, ...]
    big_memory: Tuple[str, ...]
```

tests/json_infra/helpers/exceptional_test_patterns.py
```diff
@@ -0,0 +1,104 @@
+from typing import Tuple
+
+
+def get_exceptional_blockchain_test_patterns(
```
SamWilsn:
```suggestion
def exceptional_blockchain_test_patterns(
```

tests/json_infra/helpers/exceptional_test_patterns.py
```diff
@@ -0,0 +1,104 @@
...
+        "/stSpecialTest/",
+        "stTimeConsuming/",
+        "stBadOpcode/",
+        "stStaticCall/",
+    )
+
+    return slow_tests, ignore_tests, big_memory_tests
+
+
+def get_exceptional_state_test_patterns(
```
SamWilsn:
```suggestion
def exceptional_state_test_patterns(
```

tests/json_infra/helpers/exceptional_test_patterns.py
```diff
@@ -0,0 +1,104 @@
...
+    network: str, package: str
+) -> tuple[str, ...]:
```
SamWilsn: Would it make sense to keep this return type consistent with `get_exceptional_blockchain_test_patterns`, even if we don't yet have any ignore or big memory tests? I feel like that might let us merge some code paths elsewhere (but I could be wrong).

tests/json_infra/helpers/exceptional_test_patterns.py
```diff
@@ -0,0 +1,104 @@
...
+    slow_tests :
+        Patterns to match test files or test ids that should be marked as slow.
+    ignore_tests :
+        Patterns to match test files or test ids that should be ignored.
+    big_memory_tests :
+        Patterns to match test files or test ids that require large memory.
+    """
+    slow_tests = (
+        # GeneralStateTests
+        "stTimeConsuming/CALLBlake2f_MaxRounds.json",
```
SamWilsn: By "patterns" do you mean regexes, and if so, do we need to escape the periods?

tests/json_infra/helpers/load_blockchain_tests.py
```diff
@@ -152,37 +155,39 @@ def load_json_fixture(test_file: str, network: str) -> Generator:
...
+def fetch_blockchain_tests(
     network: str,
-    only_in: Tuple[str, ...] = (),
-    slow_list: Tuple[str, ...] = (),
-    big_memory_list: Tuple[str, ...] = (),
-    ignore_list: Tuple[str, ...] = (),
 ) -> Generator[Dict | ParameterSet, None, None]:
+    # Filter FORKS based on fork_option parameter
+    package = cast(str, FORKS[network]["package"])
+    test_dirs = cast(list[str], FORKS[network]["blockchain_test_dirs"])
```
SamWilsn: Instead of casting here, can we add a type annotation to the `FORKS` definition?

tests/json_infra/helpers/exceptional_test_patterns.py
SamWilsn: Ah, yeah, these are regexes. I'd do the compile inside the function instead of later on, and you might want to [`escape`](https://docs.python.org/3/library/re.html#re.escape) `network` and `package` to be on the safe side.

tests/json_infra/test_transaction.py
```diff
@@ -0,0 +1,96 @@
...
+test_dir = f"{ETHEREUM_TESTS_PATH}/TransactionTests"
+
+
+def _generate_high_nonce_tests_function(fork_name: str) -> Callable:
+    @pytest.mark.parametrize(
+        "test_file_high_nonce",
+        [
+            "ttNonce/TransactionWithHighNonce64Minus1.json",
+            "ttNonce/TransactionWithHighNonce64.json",
+            "ttNonce/TransactionWithHighNonce64Plus1.json",
```
SamWilsn, line 28: If EEST adds a new file in `ttNonce`, will we receive any warning here? If not, it's fairly likely we'll never update these arrays.

tests/json_infra/test_trie.py
```diff
@@ -0,0 +1,114 @@
...
+)
+
+from . import FORKS, TEST_FIXTURES
+
+FIXTURE_PATH = TEST_FIXTURES["ethereum_tests"]["fixture_path"]
+
+forks = FORKS.keys()
+
+
+def to_bytes(data: str) -> Bytes:
```
SamWilsn:
```suggestion
def to_bytes(data: str | None) -> Bytes:
```

based on the function contents

tests/json_infra/test_trie.py
```diff
@@ -0,0 +1,114 @@
...
+    package = FORKS[fork]["package"]
+    trie_module = importlib.import_module(f"ethereum.{package}.trie")
+
+    for name, test in tests.items():
+        st = trie_module.Trie(secured=True, default=b"")
+        for k, v in test.get("in").items():
+            trie_module.trie_set(st, to_bytes(k), to_bytes(v))
+        result = trie_module.root(st)
+        expected = remove_hex_prefix(test.get("root"))
+        assert result.hex() == expected, f"test {name} failed"
```
SamWilsn: Is it important that we do the comparison in hex? Feels more brittle than decoding the hex instead.

tests/json_infra/test_trie.py
```diff
@@ -0,0 +1,114 @@
...
+        for t in test.get("in"):
+            trie_module.trie_set(st, to_bytes(t[0]), to_bytes(t[1]))
+        result = trie_module.root(st)
+        expected = remove_hex_prefix(test.get("root"))
+        assert result.hex() == expected, f"test {name} failed"
+
+
+@pytest.mark.parametrize("fork", forks)
+def test_trie_any_order(fork: str) -> None:
+    tests = load_tests("trieanyorder.json")
```
SamWilsn, line 92: Same kind of concern here. Any way we can avoid hardcoding JSON file names?

tests/json_infra/vm/__init__.py
```diff
@@ -0,0 +1,27 @@
+from ..conftest import pytest_config
+
+FORKS = [
+    ("ConstantinopleFix", "constantinople"),
+    ("Byzantium", "byzantium"),
+    ("EIP158", "spurious_dragon"),
+    ("EIP150", "tangerine_whistle"),
+    ("Homestead", "homestead"),
+    ("Frontier", "frontier"),
+]
```
SamWilsn, line 8: Why are these forks listed here specifically? Is this `FORKS` related to the earlier `FORKS`? I think this deserves, at the very least, a comment.

tests/json_infra/vm/test_keccak.py
```diff
@@ -1,23 +1,19 @@
...
-    run_test,
-    f"{ETHEREUM_TESTS_PATH}/LegacyTests/Constantinople/VMTests/vmSha3Test",
-)
-run_special_sha3_vm_test = partial(
-    run_test,
-    f"{ETHEREUM_TESTS_PATH}/LegacyTests/Constantinople/VMTests/vmIOandFlowOperations",
+TEST_DIR = (
+    f"{ETHEREUM_TESTS_PATH}/LegacyTests/Constantinople/VMTests/vmSha3Test"
 )
+SPECIAL_TEST_DIR = f"{ETHEREUM_TESTS_PATH}/LegacyTests/Constantinople/VMTests/vmIOandFlowOperations"
```
SamWilsn, line 13: What's special?

tox.ini
```diff
@@ -9,12 +9,12 @@ extras =
     test
     doc
 commands =
-    isort src tests --check --diff --skip-glob "tests/fixtures/*"
-    black src tests --check --diff --exclude "tests/fixtures/*"
-    flake8 src tests
-    mypy src tests --exclude "tests/fixtures/*" --namespace-packages
+    isort src tests --check --diff --skip-glob "tests/json_infra/fixtures/*"
+    black src tests --check --diff --exclude "tests/json_infra/fixtures/*"
+    flake8 src tests --exclude "tests/json_infra/fixtures/*"
```
SamWilsn, line 14: Dumb question, but if we move `fixtures` out of `tests/`, can we drop all the exclusion stuff?

tests/json_infra/__init__.py
gurukamath: `package` would be the the folder in `src`. So the tuple represents the `(fork name as found in tests, folder in src)`.

An example would be something like ("EIP158", "spurious_dragon")

tests/json_infra/__init__.py
gurukamath: I chose not to do that because fork name as mentioned here has no relevance outside of the test suite (json fixtures), so this is really just test specific info. Like `EIP158` is just how the test suite refers to `Spurious Dragon`

tests/json_infra/conftest.py
gurukamath: Yeah. I had do do a bit of workaround because we want to be able to collect the fork based on the inputs provided

tests/json_infra/__init__.py
gurukamath: We could create straight-up dict of dicts but I just wanted to avoid repetition / duplication

tests/json_infra/helpers/exceptional_test_patterns.py
gurukamath: Yes. The logic where we process ignore some tests / mark some as slow is common to the two pathways although that is very limited overlap. I will update this to keep the return type consistent and create an issue to do the actual re-factor of the logic at a later stage

tests/json_infra/test_ethash.py
gurukamath: This is again a question of whether we want to add information like `EIP150`, `EIP158` and `ConstantinopleFix` to `Hardfork`, which is nothing more that a naming convention in test jsons. Plus this is propably never going to change.

tests/json_infra/test_blockchain_tests.py
gurukamath: IIRC, if we want to pytest.mark.parametrize over the forks, all the forks should be known at import time because that is when the decorator gets processed. Since we want to be able to control the forks using the `--fork` option, the forks are only known during the collection phase. So, I just went with dynamically generating test functions

tests/json_infra/vm/__init__.py
gurukamath: We stopped having VM tests post constantinople. Other than that, there is nothing specific about this set of forks. Also this list is never going to change

tests/json_infra/__init__.py
SamWilsn: How about `json_fork, eels_fork` then?

tests/json_infra/test_blockchain_tests.py
SamWilsn: I'm not sure I follow. Here's a sketch of what I was thinking:

```patch
diff --git a/tests/json_infra/conftest.py b/tests/json_infra/conftest.py
index 390fd1ee8..4faccfcbc 100644
--- a/tests/json_infra/conftest.py
+++ b/tests/json_infra/conftest.py
@@ -10,16 +10,13 @@ from _pytest.config import Config
 from _pytest.config.argparsing import Parser
 from filelock import FileLock
 from git.exc import GitCommandError, InvalidGitRepositoryError
-from pytest import Session, StashKey
+from pytest import Session, StashKey, skip
 from requests_cache import CachedSession
 from requests_cache.backends.sqlite import SQLiteCache
 from typing_extensions import Self
 
 from . import TEST_FIXTURES
 
-# Global variable to store pytest config for access during test collection
-pytest_config = None
-
 try:
     from xdist import get_xdist_worker_id  # type: ignore[import-untyped]
 except ImportError:
@@ -63,8 +60,6 @@ def pytest_configure(config: Config) -> None:
     """
     Configure the ethereum module and log levels to output evm trace.
     """
-    global pytest_config
-    pytest_config = config
 
     if config.getoption("optimized"):
         import ethereum_optimized
@@ -81,6 +76,16 @@ def pytest_configure(config: Config) -> None:
         ethereum.trace.set_evm_trace(new_trace_function)
 
 
+def pytest_runtest_setup(item):
+    desired_fork = item.config.getoption("fork", None)
+    if not desired_fork:
+        return
+
+    forks_of_test = [mark.args[0] for mark in item.iter_markers(name="fork")]
+    if forks_of_test and desired_fork not in forks_of_test:
+        skip(f"test not for fork `{desired_fork}`")
+
+
 class _FixturesDownloader:
     cache: Final[SQLiteCache]
     session: Final[CachedSession]
diff --git a/tests/json_infra/test_blockchain_tests.py b/tests/json_infra/test_blockchain_tests.py
index dad33eb5f..661666af6 100644
--- a/tests/json_infra/test_blockchain_tests.py
+++ b/tests/json_infra/test_blockchain_tests.py
@@ -3,7 +3,6 @@ from typing import Callable, Dict
 import pytest
 
 from . import FORKS
-from .conftest import pytest_config
 from .helpers.load_blockchain_tests import (
     Load,
     fetch_blockchain_tests,
@@ -13,6 +12,7 @@ from .helpers.load_blockchain_tests import (
 
 
 def _generate_test_function(fork_name: str) -> Callable:
+    @pytest.mark.fork(fork_name)
     @pytest.mark.json_blockchain_tests
     @pytest.mark.parametrize(
         "blockchain_test_case",
@@ -29,22 +29,7 @@ def _generate_test_function(fork_name: str) -> Callable:
     return test_func
 
 
-# Get the fork option from pytest config if available
-
-# Determine which forks to generate tests for
-if pytest_config and pytest_config.getoption("fork", None):
-    # If --fork option is specified, only generate test for that fork
-    fork_option = pytest_config.getoption("fork")
-    if fork_option in FORKS:
-        forks_to_test = [fork_option]
-    else:
-        # If specified fork is not valid, generate no tests
-        forks_to_test = []
-else:
-    # If no --fork option, generate tests for all forks
-    forks_to_test = list(FORKS.keys())
-
-for fork_name in forks_to_test:
+for fork_name in FORKS.keys():
     locals()[
         f"test_blockchain_tests_{fork_name.lower()}"
     ] = _generate_test_function(fork_name)
```

tests/json_infra/test_ethash.py
SamWilsn: True... I just hate seeing the same information in multiple places :rofl:

tests/json_infra/test_blockchain_tests.py
SamWilsn: Actually, instead of `pytest_runtest_setup` & `skip`, I should have used [`pytest_collection_modifyitems`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems) and [`pytest_deselected`](https://docs.pytest.org/en/stable/reference/reference.html#std-hook-pytest_deselected). The way I have it in the patch spams the log with skipped tests.

gurukamath:
> Is there anything we should move into [`ethereum-rlp`](https://github.com/ethereum/ethereum-rlp)?

Actually most of these tests use fork related types like `Header`, `Transaction` etc. And I don't think these contribute significantly to testing the spec. So, I'd still be in favor of just getting rid of them

tests/json_infra/test_blockchain_tests.py
gurukamath: Thanks for this suggestion. Was unaware of `pytest_collection_modifyitems`. Will update

tests/json_infra/test_ethash.py
gurukamath: Also, this is not clearly POW vs POS. We have to skip forks like `Dao Fork`, `Muir Glacier` etc. since we don't have any tests for them and `London` since we don't have these tests. I would be inclined to leave this as is

tests/json_infra/test_transaction.py
gurukamath: The `json_infra` will have the role of making sure that our json consumption is up-to-date. Given this, running every single json test available is no longer a huge priority imo. Having said that, we could define test fetchers for transaction tests as well but I'm not sure how much worth the effort that is.

tests/json_infra/test_trie.py
gurukamath: Same as transaction tests.

tests/json_infra/__init__.py
```diff
@@ -0,0 +1,143 @@
...
+from typing_extensions import NotRequired
+
+
+class _FixtureSource(TypedDict):
+    url: str
+    fixture_path: str
+    commit_hash: NotRequired[str]
+
+
+# Update the links and commit has in order to consume
```
Carsons-Eels: > Update the links and commit has in order to consume

_Update the links and commit has`h` in order to consume_

tests/json_infra/__init__.py
Carsons-Eels: What about using a dataclass to store the fixture information?

Something like:
```
@dataclass(frozen=True)
class FixtureSource:
    url: str
    fixture_path: str
    commit_hash: str = field(default=None)
    
TEST_FIXTURES: Dict[str, FixtureSource] = {
    "evm_tools_testdata": FixtureSource(
        url="https://github.com/gurukamath/evm-tools-testdata.git",
        commit_hash="792422d",
        fixture_path="tests/json_infra/fixtures/evm_tools_testdata",
    ),
    ...
}
```

tox.ini
Carsons-Eels: If we want to have our tooling validate just the code that stays in here... yes. I believe so, because we specify the directories to check explicitly, and are only excluding the fixtures directory because it is nested in a folder we otherwise want to validate.

gurukamath: We should sort this out post full weld because eest also uses the `fixtures` folder in root as the default for fill

tests/json_infra/vm/test_keccak.py
gurukamath: I think these are just keccak tests that are in other older folders (not in sha3 folder). Another carry over from the VM tests I think

tests/json_infra/__init__.py
gurukamath: Updated

tests/json_infra/test_blockchain_tests.py
gurukamath: Updated

tests/json_infra/helpers/exceptional_test_patterns.py
gurukamath: Updated

tests/json_infra/helpers/exceptional_test_patterns.py
gurukamath: Updated

tests/json_infra/helpers/exceptional_test_patterns.py
gurukamath: Updated

tests/json_infra/helpers/exceptional_test_patterns.py
gurukamath: Updated

tests/json_infra/helpers/load_blockchain_tests.py
gurukamath: Updated

tests/json_infra/test_trie.py
gurukamath: Updated

tests/json_infra/__init__.py
gurukamath: Updated

tests/json_infra/helpers/exceptional_test_patterns.py
gurukamath: Updated

tests/json_infra/test_trie.py
gurukamath: Updated

gurukamath:
@SamWilsn Have updated the PR based on the feedback

tests/json_infra/__init__.py
SamWilsn: For me, at least, it's probably the number of constants that's bothering me, but this pull request isn't the place to solve it.

tests/json_infra/helpers/exceptional_test_patterns.py
SamWilsn: The `.` in `.json` should probably be `\\.` right? Else it will match any character, not just a dot.

SamWilsn (approved):

tests/json_infra/__init__.py
gurukamath: Created an issue #1367

tests/json_infra/helpers/exceptional_test_patterns.py
gurukamath: Updated

Issue #1145, opened 2025-03-07
Investigate source code mutation testing


SamWilsn:

SamWilsn:
We should include [the patches against EELS](https://github.com/souradeep-das/mutate-execution-specs).

- The forced fail test is fine to just include.
- Instead of skipping problematic mutant tests, can we tag them instead and exclude them with a `-m 'not angry-mutant'` or some such? Same idea as the slow tests.

SamWilsn:
@souradeep-das has most (if not all) of this working now 🎉 

I'm going to leave this open until we get it running somewhere.

PR #1361, opened 2025-08-06
Update tests to align with mutation testing runs

souradeep-das:
### What was wrong?

associated with https://github.com/ethereum/execution-specs/issues/1145

### How was it fixed?

- add code `angry_mutant` for tests that cannot be run in the mutation testing suite
- update conftest for mutmut parallel run setting
- add explicit forced fail condition since mutmut fails attaching programmatic fail if the mutation paths do not have possible mutations

Corresponding update on mutmut run - https://github.com/souradeep-das/mutmut/commit/cc99e509092ed232600712d9d77fb38be31f74ec

mutmut has to be run with a `--pytest-extra-args` option.
For instance-
```
mutmut --paths-to-mutate src/ethereum/frontier/vm/precompiled_contracts/sha256.py \
       --tests-dir tests/json_infra/ \
       --debug \
       --pytest-extra-args '-m "not slow and not angry_mutant" --ignore-glob=tests/json_infra/fixtures/* --fork Frontier' \
       run
```

#### Cute Animal Picture

### DIFF
--- a/tests/json_infra/conftest.py
+++ b/tests/json_infra/conftest.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Final, Optional, Set
 
 import git
+import pytest
 import requests_cache
 from _pytest.config import Config
 from _pytest.config.argparsing import Parser
@@ -215,6 +216,14 @@ fixture_lock = StashKey[Optional[FileLock]]()
 
 
 def pytest_sessionstart(session: Session) -> None:  # noqa: U100
+    # used for mutmut mutation testing
+    all_fixtures_ready = all(
+        os.path.exists(props["fixture_path"])
+        for props in TEST_FIXTURES.values()
+    )
+    if os.environ.get("MUTANT_UNDER_TEST") and all_fixtures_ready:
+        return
+
     if get_xdist_worker_id(session) != "master":
         return
 
@@ -250,8 +259,17 @@ def pytest_sessionfinish(
     if get_xdist_worker_id(session) != "master":
         return
 
-    lock_file = session.stash[fixture_lock]
-    session.stash[fixture_lock] = None
+    if fixture_lock in session.stash:
+        lock_file = session.stash[fixture_lock]
+        session.stash[fixture_lock] = None
 
-    assert lock_file is not None
-    lock_file.release()
+        assert lock_file is not None
+        lock_file.release()
+
+
+# This is required explicitly becuase when the source does not have any
+# mutable code, mutmut does not run the forced fail condition.
+@pytest.fixture(autouse=True)
+def mutmut_forced_fail() -> None:
+    if os.environ.get("MUTANT_UNDER_TEST") == "fail":
+        pytest.fail("Forced fail for mutmut sanity check")
--- a/tests/json_infra/conftest.py
+++ b/tests/json_infra/conftest.py
@@ -239,24 +239,26 @@ def pytest_sessionstart(session: Session) -> None:  # noqa: U100
 
     # use portalocker for mutmut runs
     if os.environ.get("MUTANT_UNDER_TEST"):
-        while True:
-            with portalocker.Lock(lock_path, flags=portalocker.LOCK_SH):
-                all_fixtures_ready = all(
-                    os.path.exists(props["fixture_path"])
-                    for props in TEST_FIXTURES.values()
-                )
-                if all_fixtures_ready:
-                    return
+        shared_lock = portalocker.Lock(lock_path, flags=portalocker.LOCK_SH)
+        shared_lock.acquire()
+        session.stash['mutmut_shared_lock'] = shared_lock
 
+        all_fixtures_ready = all(
+            os.path.exists(props["fixture_path"])
+            for props in TEST_FIXTURES.values()
+        )
+        if not all_fixtures_ready:
+            shared_lock.release()
             with portalocker.Lock(lock_path, flags=portalocker.LOCK_EX):
                 all_fixtures_ready = all(
                     os.path.exists(props["fixture_path"])
                     for props in TEST_FIXTURES.values()
                 )
-                if all_fixtures_ready:
-                    continue
-
-                download_fixtures(session.config.rootpath)
+                if not all_fixtures_ready:
+                    download_fixtures(session.config.rootpath)
+            shared_lock.acquire()
+            session.stash['mutmut_shared_lock'] = shared_lock
+        return
 
     if get_xdist_worker_id(session) != "master":
         return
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -210,8 +219,17 @@ def pytest_sessionfinish(
     if get_xdist_worker_id(session) != "master":
         return
 
-    lock_file = session.stash[fixture_lock]
-    session.stash[fixture_lock] = None
```
SamWilsn: Do we want to avoid locking when running under mutmut?

SamWilsn (commented):
Looks good aside from the locking question. Might want to wait for #1360 to be merged though, since this one looks a bit easier to rebase.

tests/conftest.py
souradeep-das: yes, but only when fixtures already exist downloaded - we could skip locking then.

Since mutmut has multiple phases, the fixtures exist after the first `stat collection` run (this phase happens before the mutation tests start parallely)

SamWilsn: Couldn't another test run delete the fixtures folder while it's running if we don't lock?

souradeep-das: yeah makes sense - is there a concern given usually the fixtures persist?
I was thinking since mutmut runs parallely - acquiring the lock for each of these process might lead to sequential startup for each of the mutation run? (which should cost some time) 
But its true that if somehow fixtures are deleted, then the current workers at that time would fail

SamWilsn: I suppose it depends how mutmut runs in parallel. If it's entirely separate processes, without this lock, each one would blow away the fixtures directory when it starts; but with the lock it'll be serialized.

If mutmut works like xdist, then it'll still run in parallel.

souradeep-das: In mutmut a separate pytest process is launched for each mutation, so these run independently in parallel (i think unlike xdist?) Also, from all my runs i see the fixtures persist, should we take a tradeoff then over a rare chance of deletion? Here's a CI test run - https://github.com/souradeep-das/execution-specs/actions/runs/17051785519

SamWilsn: See [here](https://github.com/ethereum/execution-specs/blob/1626a8c15fbff66a0b08befac9b33d6b25766a7c/tests/json_infra/conftest.py#L160). While the tarball is being extracted, the state on disk is uncertain.

I think we have three options: (1) have each process use its own fixtures directory; (2) use an exclusive lock during extraction, a shared lock during test runs, and write the currently extracted tests to disk somewhere; or (3) live with serialized tests.

I'd pretty strongly prefer not ignoring this issue, because (unless I'm misunderstanding) we might end up in the situation where a mutmut process only sees 10% of the tests and then reports a false positive.

tests/json_infra/conftest.py
```diff
@@ -243,15 +234,61 @@ def pytest_sessionstart(session: Session) -> None:  # noqa: U100
...
+    # use portalocker for mutmut runs
+    if os.environ.get("MUTANT_UNDER_TEST"):
+        while True:
+            with portalocker.Lock(lock_path, flags=portalocker.LOCK_SH):
+                all_fixtures_ready = all(
+                    os.path.exists(props["fixture_path"])
+                    for props in TEST_FIXTURES.values()
+                )
+                if all_fixtures_ready:
+                    return
```
SamWilsn: This exits the context manager and, as I far understand portalocker, releases the lock. We need to hold the shared lock for the duration of the test run to prevent another process coming in and taking the exclusive lock.

souradeep-das: oh yess, updated to hold the shared lock entirely now. used a while true loop before, that was confusing so removed it!
also tested with deleting fixtures when running mutmut

tests/json_infra/conftest.py
```diff
@@ -243,15 +234,70 @@ def pytest_sessionstart(session: Session) -> None:  # noqa: U100
...
+        if not all_fixtures_ready:
+            shared_lock.release()
+            with portalocker.Lock(lock_path, flags=portalocker.LOCK_EX):
+                all_fixtures_ready = all(
+                    os.path.exists(props["fixture_path"])
+                    for props in TEST_FIXTURES.values()
+                )
+                if not all_fixtures_ready:
+                    download_fixtures(session.config.rootpath)
+            shared_lock.acquire()
```
SamWilsn: Is it possible that between when the exclusive lock is released and when the shared lock is taken that another process acquires the exclusive lock and updates the fixtures to another version?

I think your `while True` approach where you checked every iteration for the conditions was a good approach.

tests/json_infra/conftest.py
```diff
@@ -243,15 +234,70 @@ def pytest_sessionstart(session: Session) -> None:  # noqa: U100
                 )
 
 
+def pytest_sessionstart(session: Session) -> None:  # noqa: U100
+    lock_path = session.config.rootpath.joinpath("tests/fixtures/.lock")
+
+    # use portalocker for mutmut runs
+    if os.environ.get("MUTANT_UNDER_TEST"):
+        shared_lock = portalocker.Lock(lock_path, flags=portalocker.LOCK_SH)
+        shared_lock.acquire()
```
SamWilsn: What happens if an exception is thrown while this lock is held? I assume the pytest process eventually exits?

If we aren't sure, we should probably use a try-except that unlocks and unsets the `mutmut_shared_lock`.

tests/json_infra/conftest.py
```diff
@@ -243,15 +234,70 @@ def pytest_sessionstart(session: Session) -> None:  # noqa: U100
                 )
 
 
+def pytest_sessionstart(session: Session) -> None:  # noqa: U100
+    lock_path = session.config.rootpath.joinpath("tests/fixtures/.lock")
+
+    # use portalocker for mutmut runs
+    if os.environ.get("MUTANT_UNDER_TEST"):
```
SamWilsn, line 241: I think we can use the same locking code for both mutmut and non-mutmut runs. The more fine-grained locking isn't going to hurt our normal tests at all.

tests/json_infra/conftest.py
souradeep-das: got it! added that back. Also added a try-except.

tests/json_infra/conftest.py
souradeep-das: tried doing a unified locking - but somehow mutmut tests unusually exits from the point the fixtures are deleted 🤔 
can investigate into this more if we want to unify them!

SamWilsn:
I think we still need to figure out what we're doing for this PR, so don't worry about rebasing onto `forks/amsterdam` for now.

felix314159:
This PR never moved forward, and I would prefer if we kept using `filelock.FileLock` instead of introducing `portalocker`

Issue #1362, opened 2025-08-08
Integration Mutation Testing on CI


souradeep-das:
— find the file changes in commit on PR
— install mutmut (apply patch from repo and set it up)
— configure mutmut to apply mutations only on that fork specific file changes (mutations cannot be selectively run for a specific function using mutmut, mutmut could be patched however)
— find the respective tests for that specific fork  
— run mutmut using this configuration

Later,

— compare results with baseline
(Baseline is basically a statistic of what file mutations are expected to fail
— If theres any new additional failure, report that

PR #1364, opened 2025-08-08
Run mutation tests in CI for changed files

souradeep-das:
### What was wrong?

Related to Issue https://github.com/ethereum/execution-specs/issues/1362 and https://github.com/ethereum/execution-specs/issues/1145

### How was it fixed?
- Run mutation tests on files affected in the PR (files are the atomic unit for running tests, so if a file was changed all of its mutations will be tested, mutmut currently does not allow function specific tests - but a patch can be created)
- Segregate by fork and run the test suite (tests/{fork_name}) for the files changed in that fork

Test run on fork: https://github.com/souradeep-das/execution-specs/actions/runs/17051785519

**Note** - needs https://github.com/ethereum/execution-specs/pull/1361 to be merged

#### Cute Animal Picture

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -80,7 +80,7 @@ jobs:
           files: .tox/coverage.xml
           flags: unittests
           token: ${{ secrets.CODECOV_TOKEN }}
-  
+
   optimized:
     runs-on: [self-hosted-ghr, size-xl-x64]
     needs: static
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -96,13 +96,47 @@ jobs:
       - name: Run optimized tests
         run: tox -e optimized
 
-  mutation-test:
+  changed-forks:
     runs-on: ubuntu-latest
     needs: static
+    outputs:
+      changed-forks: ${{ steps.filter-tests.outputs.forks }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install project for fork detection
+        run: pip install .
+      - name: Find changed valid forks
+        id: set-forks
+        run: |
+          git fetch origin ${{ github.event.pull_request.base.ref }}
+          VALID_FORKS=$(python -c "import json; from ethereum_spec_tools.forks import Hardfork; print(json.dumps([f.short_name for f in Hardfork.discover()]))")
+          CHANGED_FORKS=$(git diff --name-only origin/${{ github.event.pull_request.base.ref }} ${{ github.sha }} -- src/ethereum/ | awk -F'/' '{print $3}' | sort -u)
+          readarray -t VALID_FORKS_ARR < <(echo "$VALID_FORKS" | jq -r '.[]')
+          FILTERED=$(for f in $CHANGED_FORKS; do for v in "${VALID_FORKS_ARR[@]}"; do [ "$f" = "$v" ] && echo $f; done; done | jq -R . | jq -s -c .)
+          echo "FILTERED (final forks for matrix): $FILTERED"
+          echo "forks=$FILTERED" >> $GITHUB_OUTPUT
+      - name: Filter out forks without test directories
+        id: filter-tests
+        run: |
+          FILTERED_WITH_TESTS=$(for f in $(echo '${{ steps.set-forks.outputs.forks }}' | jq -r '.[]'); do
+            if [ -d "tests/$f" ]; then
+              echo $f
+            fi
+          done | jq -R . | jq -s -c .)
+          echo "FILTERED_WITH_TESTS: $FILTERED_WITH_TESTS"
+          echo "forks=$FILTERED_WITH_TESTS" >> $GITHUB_OUTPUT
+
+  mutation-test:
+    runs-on: [self-hosted-ghr, size-xl-x64]
+    needs: [static, changed-forks]
     if: github.event_name == 'pull_request'
     strategy:
       matrix:
-        fork: [frontier, homestead, tangerine_whistle, spurious_dragon, byzantium, constantinople, istanbul, berlin, london, paris, shanghai, cancun, prague, osaka]
+        fork: ${{ fromJson(needs.changed-forks.outputs.changed-forks) }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -95,3 +95,67 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install project with test dependencies
+        run: pip install -e '.[test]'
+      - name: Install mutmut
+        run: pip install mutmut==3.3.0
+      - name: Apply mutmut patch
+        run: |
+          git clone https://github.com/souradeep-das/mutate-execution-specs.git
```
souradeep-das: should i copy the patch to this repo instead?

.github/workflows/test.yaml
```diff
@@ -95,3 +95,67 @@ jobs:
...
       - name: Run optimized tests
         run: tox -e optimized
+
+  mutation-test:
+    runs-on: ubuntu-latest
+    needs: static
+    if: github.event_name == 'pull_request'
+    strategy:
+      matrix:
+        fork: [frontier, homestead, tangerine_whistle, spurious_dragon, byzantium, constantinople, istanbul, berlin, london, paris, shanghai, cancun, prague, osaka]
```
souradeep-das: some forks do not have a direct tests/{fork_name} file - what should we do for them?

.github/workflows/test.yaml
souradeep-das, line 1: The mutation tests might take a while to run - so I'm wondering how we should run them - per PR or commit on main

If several files have changes the time to run on CI also might be unrealistic. In which case I was thinking if we should add more limitations on when to run these tests (or patch mutmut to be able to run mutation tests for specific functions (as opposed to files).
Attached a test run on the main comment ^

.github/workflows/test.yaml
SamWilsn: Instead of patching, you can install a git dependency:

```bash
pip install git+https://github.com/souradeep-das/mutmut.git@<commit hash>
```

I think that would be easier to maintain than a patch file.

---

On a separate note, would something like this work (and maybe we can upstream)?

```patch
diff --git a/mutmut/__main__.py b/mutmut/__main__.py
index 0fb1ad9..15b16ef 100644
--- a/mutmut/__main__.py
+++ b/mutmut/__main__.py
@@ -407,6 +407,8 @@ class PytestRunner(TestRunner):
         if mutmut.config.debug:
             params = ['-vv'] + params
             print('python -m pytest ', ' '.join(params))
+        if mutmut.config.pytest_extra_args:
+            params += mutmut.config.pytest_extra_args
         exit_code = int(pytest.main(params, **kwargs))
         if mutmut.config.debug:
             print('    exit code', exit_code)
@@ -676,6 +678,7 @@ class Config:
     debug: bool
     paths_to_mutate: List[Path]
     tests_dir: List[str] = None
+    pytest_extra_args: List[str] = None
 
     def should_ignore_for_mutation(self, path):
         if not str(path).endswith('.py'):
@@ -756,6 +759,7 @@ def load_config():
             for y in s('paths_to_mutate', [])
         ] or guess_paths_to_mutate(),
         tests_dir=s('tests_dir', []),
+        pytest_extra_args=s('pytest_extra_args', []),
     )
```

.github/workflows/test.yaml
SamWilsn: This is likely going to change with @gurukamath's recent work on refactoring the tests.

.github/workflows/test.yaml
```diff
@@ -95,3 +95,67 @@ jobs:
...
+        run: |
+          git clone https://github.com/souradeep-das/mutate-execution-specs.git
+          cp mutate-execution-specs/mutmut.patch .
+          MUTMUT_FILE_TO_PATCH=$(python3 -c "import mutmut, os; print(os.path.join(os.path.dirname(mutmut.__file__), '__main__.py'))")
+          patch $MUTMUT_FILE_TO_PATCH < mutmut.patch
+      - name: Find changed files for fork
+        id: changed
+        run: |
+          git fetch origin ${{ github.event.pull_request.base.ref }}
+          CHANGED_FILES=$(git diff --name-only origin/${{ github.event.pull_request.base.ref }} ${{ github.sha }} | grep "^src/ethereum/${{ matrix.fork }}/" || true)
```
SamWilsn: Do you need to grep, or can you just do something like `git diff ... -- src/ethereum/${{ matrix.fork }}/`?

.github/workflows/test.yaml
```diff
@@ -95,3 +95,67 @@ jobs:
...
+          echo "changed_files<<EOF" >> $GITHUB_OUTPUT
+          echo "$CHANGED_FILES" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
```
SamWilsn:
```suggestion
          CHANGED_FILES=$()
          echo "changed_files<<EOF" >> $GITHUB_OUTPUT
          ( git diff --name-only origin/${{ github.event.pull_request.base.ref }} ${{ github.sha }} | grep "^src/ethereum/${{ matrix.fork }}/" || true ) >> $GITHUB_OUTPUT
          echo "EOF" >> $GITHUB_OUTPUT
```

Can avoid storing the list in a temporary variable.

.github/workflows/test.yaml
```diff
@@ -95,3 +95,67 @@ jobs:
...
+        if: steps.changed.outputs.changed_files == ''
+        run: echo "No changes for ${{ matrix.fork }}, skipping mutmut."
+      - name: Configure mutmut for changed files
+        if: steps.changed.outputs.changed_files != ''
+        run: |
+          echo '[tool.mutmut]' > pyproject.toml
+          FILES=$(echo "${{ steps.changed.outputs.changed_files }}" | awk '{printf "\"%s\", ", $0}' | sed 's/, $//')
+          echo "paths_to_mutate = [ $FILES ]" >> pyproject.toml
+          echo "tests_dir = [\"tests/${{ matrix.fork }}/\"]" >> pyproject.toml
+          echo "debug = true" >> pyproject.toml
```
SamWilsn: Hm, this is... unusual.

So first, this might be easier to implement with [`shell: python`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-running-an-inline-python-script) instead of bash. `tomllib` is in the standard library.

Second, we shouldn't need to modify our project configuration at all to get this to work. I think it would make sense to try and get a patch into mutmut that allows passing pytest arguments via the command line (unlike my patch above) or maybe via an environment variable. We can't be the only project with stupidly long running tests.

.github/workflows/test.yaml
```diff
@@ -95,3 +95,67 @@ jobs:
       - uses: ./.github/actions/setup-env
       - name: Run optimized tests
         run: tox -e optimized
+
+  mutation-test:
+    runs-on: ubuntu-latest
```
SamWilsn, line 100:
```suggestion
    runs-on: [self-hosted-ghr, size-xl-x64]
```

This is the magic sauce that runs the job on a beefier machine.

.github/workflows/test.yaml
SamWilsn: If you generate the matrix from the output of a previous job (i.e. `${{ needs.changed-forks.outputs.changed-forks }}` where `changed-forks` is a preceding job that stores the changed forks), you can cut down own the number of runs plus use `ethereum_spec_tools.forks:Hardfork` to dynamically generate the list instead of having to hard code it.

.github/workflows/test.yaml
souradeep-das: > On a separate note, would something like this work (and maybe we can upstream)?

Is the patch you mentioned above, for any specific cases that we have in mind? or is it just intended to supply in options like test_dir?

I updated mutmut to take in options through the cli - and its on a PR upstream

.github/workflows/test.yaml
souradeep-das: thanks - that sounds great!

.github/workflows/test.yaml
souradeep-das: also looks like mutmut is not particularly leaning towards enabling the cli options

souradeep-das: https://github.com/boxed/mutmut/pull/422#issuecomment-3184889126

felix314159:
Any update on this? Is it still relevant?

Issue #1209, opened 2025-05-04
Create an abstract interface for accessing state


petertdavies:
Currently the spec hardcodes the state as being stored in memory in the `State` object with entire MPT recomputed from scratch.This is not a good design and we already have the "optimised" state which monkey patches this code to allow storing the state on disk.

We should switch to a system where a object used to answer questions about the state is passed as a parameter to `state_transition()`. This would allow various types of state management strategies to be used flexibly with the spec. It would also make the transition to stateless easier, because the precise details of how the new state commitment (e.g. verkle trie) are likely to be highly specialised cryptography not suited to the skill set of the EELS. A clear boundary interface would allow the Statelessness team to develop those specs separately.

The interface would probably look like:

```python
class CantAnswer(Exception):
    """
    Thrown by state oracles when they can't answer a state request.

    Intended for witness based oracles that can't answer queries not witnessed
    to.
    """
 
class MerkleOracle(Protocol):
    def get_account(self, address: Bytes20) -> Optional[Account]:
        """
        Query an account in the state.

        NOTE: Must answer multiple identical queries efficiently.
        """
        ...

    def get_storage(self, address: Bytes20, key: Bytes32) -> Bytes32:
        """
        Query a storage key in the state.

        NOTE: Must answer multiple identical queries efficiently.
        """
        ...

    def write(self, accounts: Dict[Bytes20, Optional[Account]], storage: Dict[Tuple[Bytes20, Bytes32]]) -> None:
        """
        Bulk write accounts and storage keys. Typically called once per block.

        NOTE: This method may be expensive and should be called sparingly.
        """
        ...

    def state_root(self) -> Bytes32:
        """
        Compute the state root.
        """
        ...

    def get_chunk(self, seq_num: Uint, stride_length: Uint) -> List[Tuple[Bytes20, Account]]:
        """
        Used during the state transition. Details TBC.
        """
        ...


class FlatStateOracle(Protocol):
    def get(self, key: Bytes32) -> Bytes32:
        """
        Query a key in the state.

        NOTE: Must answer multiple identical queries efficiently.
        """
        ...

    def write(self, values: Dict[Bytes32, Bytes32]) -> None:
        """
        Bulk write accounts and storage keys. Typically called once per block.

        NOTE: This method may be expensive and should be called sparingly.
        """
        ...

    def state_root(self) -> Bytes32:
        """
        Compute the state root.
        """
        ...
```

[...]

PR #1371, opened 2025-08-17
feat: pass StateOracle as a parameter to `state_transition`

kevaundray:
### What was wrong?

Not mergable in current state

Related to https://github.com/ethereum/execution-specs/issues/1209

### How was it fixed?

This initial PR is made to start discussion on implementing this. Something like this would be useful for both:

- zkEVMs that need to create merkle proofs based on state that has been read
- Block level access lists that need to keep track of state changes across the block

The idea being that an oracle/state object can be wrapped to record whenever state has been accessed/modified. 

### Notes

- The original issue defined a quite minimal interface, this PR tries to reduce the diff by first introducing the idea of an Oracle and making all of the previous state management functions become methods. 

To get what the original PR described, I believe we would move most of the methods into something like an `EVMOracle` structure that depends on a `MerkleOracle`

- This PR does not modify the monkey patching that happens with the optimized impl for State


src/ethereum/osaka/fork.py
```diff
@@ -189,7 +182,7 @@ def get_last_256_block_hashes(chain: BlockChain) -> List[Hash32]:
     return recent_block_hashes
 
 
-def state_transition(chain: BlockChain, block: Block) -> None:
+def state_transition(chain: BlockChain, block: Block, oracle=None) -> None:
     """
```
kevaundray: main TLDR; the object that is used to fetch state is now an explicit parameter to the state transition function instead of it being implicit

src/ethereum/osaka/fork.py
```diff
@@ -870,7 +869,7 @@ def process_transaction(
         tx=tx,
     )
 
-    sender_account = get_account(block_env.state, sender)
+    sender_account = block_env.oracle.get_account(sender)
```
kevaundray: Most changes look like this:

Before:

```
func(state, params)
```

After:

```
oracle.func(params)
```

src/ethereum/state_oracle/interface.py
```diff
@@ -0,0 +1,213 @@
...
+Abstract interface for state oracle implementations.
+"""
+
+from typing import Any, Callable, Dict, Optional, Protocol, Tuple
+
+from ethereum_types.bytes import Bytes20, Bytes32
+
+# Use generic types for compatibility across forks
+Account = Any
+Address = Bytes20
```
kevaundray, line 12: I initially thought that this interface would fork agnostic, but perhaps it makes sense to have one per fork?

src/ethereum/state_oracle/interface.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Extensions needed for complete EVM instruction support
+    def get_storage_original(self, address: Address, key: Bytes32) -> Bytes32:
+        """
+        Get original storage value before current transaction started.
+
+        This is required for SSTORE gas calculations per EIP-2200.
+        The implementation should use state snapshots/checkpoints to track
+        pre-transaction values.
+        TODO: The oracle does not have a `begin_transaction` method, so it kind of breaks here.
```
kevaundray: Note: TODO

src/ethereum/state_oracle/memory_oracle.py
```diff
@@ -0,0 +1,191 @@
...
+to provide the oracle interface. 
+This is mainly done as a first pass to make the diff small. 
+"""
+
+from typing import Any, Dict, Optional, Tuple
+
+from ethereum_types.bytes import Bytes20, Bytes32
+from ethereum_types.numeric import Uint
+
+# TODO: This file is current Osaka specific -- we could move state.py into here to mitigate this.
```
kevaundray: Note: TODO

kevaundray:
cc @SamWilsn  @gurukamath since you were tagged in that issue -- would be good to know if this is directionally what you were thinking first, since there is likely some cleanup required

src/ethereum/state_oracle/interface.py
```diff
@@ -0,0 +1,213 @@
...
+
+        Returns None if the account doesn't exist.
+        Use this when you need to distinguish between non-existent and
+        empty accounts.
+        """
+
+    def get_storage(
+        self, address: Address, key: Bytes32  # noqa: U100
+    ) -> Bytes32:  # noqa: U100
+        """Get storage value at `key` for the given `address`."""
```
kevaundray, line 47: noqa: U100 was needed for tox lint

src/ethereum/state_oracle/interface.py
```diff
@@ -0,0 +1,213 @@
...
+
+        Used during contract creation to check if address is available.
+        """
+
+    def account_has_storage(self, address: Address) -> bool:  # noqa: U100
+        """
+        Check if account has any storage slots.
+
+        Used during contract creation to check if address is available.
+        """
```
kevaundray, line 112: As noted in the PR description; methods like these can be moved into something like an "EVMOracle", since they can be implemented using get_account plus a few extra LOC to check if there is code/nonce/storage.

The user would then only implement MerkleOracle, which then gets wrapped in an EVMOracle so that the EVM can use the relevant methods

src/ethereum/osaka/vm/instructions/environment.py
```diff
@@ -19,7 +19,6 @@
 from ethereum.utils.numeric import ceil32
```
kevaundray, line 20: blockhash was not modified, but that should also likely be a part of "state" since 2935 has been implemented

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -132,14 +132,24 @@ def block_environment(self) -> Any:
...
         """
+        # Handle oracle vs state for different forks
+        # TODO: Hack imo
+        if self.fork.is_after_fork("ethereum.osaka"):
+            from ethereum.state_oracle import MemoryMerkleOracle
+
+            oracle = MemoryMerkleOracle(self.alloc.state)
+            state_or_oracle = {"oracle": oracle}
+        else:
+            state_or_oracle = {"state": self.alloc.state}
```
kevaundray: Since the change currently only applies to Osaka

kevaundray: There are quite a few places in the `ethereum_spec_tools` directory that rely on accessing the state parameter explicitly, so I'll wait to see if this is directionally correct before modifying those

SamWilsn:
Our primary concern is readability, and I fear feeding the state oracle through everything will hurt that. Would an approach like our [trace](https://github.com/ethereum/execution-specs/blob/7ad27ac822b75e47c42e14cb4c6fece69ac0263b/src/ethereum/trace.py#L225-L234) implementation work?

Otherwise I think this looks pretty good!

kevaundray:
> Our primary concern is readability, and I fear feeding the state oracle through everything will hurt that. Would an approach like our [trace](https://github.com/ethereum/execution-specs/blob/7ad27ac822b75e47c42e14cb4c6fece69ac0263b/src/ethereum/trace.py#L225-L234) implementation work?
> 
> Otherwise I think this looks pretty good!

I have no strong opinion on how its implemented :) -- when you say feeding the state oracle through everything, do you mean places outside of BlockEnv? Since in most places we are replacing `block_env.state` with `block_env.oracle`

kevaundray:
@SamWilsn have added a commit to an approach that is closer to evm_trace; with a global oracle -- was this what you had in mind?

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -132,11 +132,21 @@ def block_environment(self) -> Any:
...
+
+            set_state_oracle(MemoryMerkleOracle(self.alloc.state))
+
         kw_arguments = {
             "block_hashes": self.env.block_hashes,
             "coinbase": self.env.coinbase,
             "number": self.env.block_number,
             "time": self.env.block_timestamp,
+            # TODO: Remove this, since its not being used.
             "state": self.alloc.state,
```
kevaundray, line 150: TODO

kevaundray: Need to remove state from BlockEnvironment to finish this

src/ethereum/state_oracle/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+
+    Returns the previous oracle if any.
+    """
+    global _state_oracle
+    old = _state_oracle
+    _state_oracle = oracle
+    return old
+
+
+def get_state_oracle() -> MerkleOracle:
```
SamWilsn, line 28: You're still returning an object here. We want this to be as flat as possible. For each function on the `MerkleOracle`, create a function in this file exposing it.

src/ethereum/state_oracle/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Get the current global state oracle.
+
+    Raises RuntimeError if no oracle has been set.
+    """
+    global _state_oracle
+    if _state_oracle is None:
+        raise RuntimeError(
+            "No global state oracle set. Call set_state_oracle() first."
+        )
```
SamWilsn, line 38: There should always be a default. That way the reader can easily trace the implementation of the unmodified spec. The average reader isn't going to be going through the whole codebase (instead just reading a function or implementing an EIP), so locality is very important.

src/ethereum/state_oracle/interface.py
```diff
@@ -0,0 +1,213 @@
...
+
+from ethereum_types.bytes import Bytes20, Bytes32
+from ethereum_types.numeric import U256
+
+# Use generic types for compatibility across forks
+Account = Any
+Address = Bytes20
+
+
+class MerkleOracle(Protocol):
```
SamWilsn, line 15: Hm, so we have a vague rule that if something changes between forks, it must be placed into the fork's folder. For example, `keccak256` is used in every fork and isn't likely to disappear, so it lives in `ethereum.crypto`, where as `Address` might get bumped to 32 bytes in the future, so it's defined per fork.

If the MPT is going to get replaced (or is fairly likely), it might be best to define it per fork.

src/ethereum/state_oracle/interface.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    Oracle interface for Merkle Patricia Trie based state.
+    """
+
+    def get_account(
+        self, address: Address  # noqa: U100
+    ) -> Account:  # noqa: U100
```
SamWilsn, line 22: I think if you `raise NotImplementedError` it mutes the linter, but I'm not 100% sure on that.

src/ethereum/state_oracle/interface.py
SamWilsn, line 15: "Oracle" is a bit of a foreign term for non-cryptography folks. The base protocol could just be `State`.

That said, we've managed to keep protocols and most OOP out of the spec so far, so we might want to rethink how we organize this.

src/ethereum/state_oracle/interface.py
SamWilsn: This is the motivation behind the minimal state interface in #1209. Everything that isn't off the shelf that can be done in the specs should be done there, even if the performance is terrible.

src/ethereum/state_oracle/__init__.py
kevaundray: I think this implies that the state_oracl folder will be copied between forks, unless we have a fork agnostic default

src/ethereum/state_oracle/interface.py
kevaundray: related to https://github.com/ethereum/execution-specs/pull/1371#discussion_r2286680828

src/ethereum/state_oracle/__init__.py
kevaundray: I also think that we would need some state to create the default oracle with

kevaundray:
Discussed with @SamWilsn offline and this may not be strictly better than what is already on main wrt the design constraints for the execution specs. Putting this back in draft.

The main change I'm looking to add is state tracking, which this should have made easier. If not, we can close it and make a separate PR for state tracking

SamWilsn:
Should we close this or keep it open and port it to `forks/amsterdam`?

kevaundray:
Think we can close it and whomever is working on #1865 can pull in anything relevant

Issue #1357, opened 2025-07-31
Improve `ethereum-spec-new-fork` to support blob parameter only hardforks


SamWilsn:
With the introduction of [EIP-7892](https://eips.ethereum.org/EIPS/eip-7892), we need to support Blob Parameter Only (BPO) hardforks.

Although they are similar to the glacier forks in that they only tune constants (and make no other changes), unlike glacier forks they are going to get more thorough testing.

I'd like to rewrite `ethereum-spec-new-fork` so that it:

1. Is more robust;
1. Requires no human-tuning after running;
1. Supports setting `FORK_CRITERIA`; and
1. Supports changing BPO parameters.

The end goal is to have t8n use the tool to generate a sequence of forks based on a BPO schedule.

PR #1372, opened 2025-08-18
Rewrite new fork tool to modify blob parameters

SamWilsn:
### What was wrong?

Related to Issue #1357 

### How was it fixed?

Rewrote the new fork tool to be more than just find-and-replace. It now uses libcst to transform the source somewhat intelligently.

#### Cute Animal Picture


src/ethereum_spec_tools/new_fork/cli.py
```diff
@@ -0,0 +1,216 @@
...
+    fork_criteria.add_argument(
+        "--at-block",
+        type=lambda x: ByBlockNumber(int(x)),
+        dest="fork_criteria",
+        help="mark the new fork as beginning at the given block number",
+    )
+
+    blob_parameters = parser.add_argument_group()
+
+    blob_parameters.add_argument(
```
gurukamath, line 89: This could get a bit more complicated since post-Osaka, we won't have TARGET_BLOB_GAS_PER_BLOCK as ain independent constant. See [this](https://github.com/ethereum/execution-specs/pull/1370)

src/ethereum_spec_tools/new_fork/cli.py
```diff
@@ -0,0 +1,216 @@
...
+
+def main(args: Sequence[str] | None = None) -> None:
+    """
+    Command-line entry point into `ForkBuilder`.
+    """
+    parser = _make_parser()
+    options = parser.parse_args(args)
+
+    if options.template_fork == "spurious_dragon":
+        raise NotImplementedError(
```
gurukamath, line 148: Tangerine Whistle and Dao Fork?

src/ethereum_spec_tools/new_fork/cli.py
SamWilsn: Should we disable this option post-osaka or allow users to modify it at their own risk?

src/ethereum_spec_tools/new_fork/cli.py
SamWilsn: Ugh, if there are more instances of "a self-reference that shouldn't be changed" I don't think I can leave this unimplemented 🤣

SamWilsn: Hm, from a quick look, it seems like spurious dragon is the only fork that's referenced by other forks (eg. [here](https://github.com/ethereum/execution-specs/blob/51cabd86502df7af596f5c78a0c4e00a4f92822c/src/ethereum/osaka/vm/interpreter.py#L185)).

SamWilsn:
What a finicky tool. Should be good to review now.

gurukamath (approved):

gurukamath (approved):

gurukamath (approved):

marioevz (approved):
This is great, I think this in combination with #1453 is going to make testing BPOs a breeze with minimal rework to be done.

Issue #641, opened 2022-11-09
Type soundness issue with `_parse`: claims to return `Sequence[str]` but can return anything.


SamWilsn:
_Originally posted by @SamWilsn in https://github.com/ethereum/execution-specs/pull/639#discussion_r1018155497_

PR #1375, opened 2025-08-19
New lint

SamWilsn:
### What was wrong?

Related to Issue #1357

There are relative imports that reach outside of the current fork. In my next PR, I'll be moving the forks into new directories.

### How was it fixed?

Added a check to the import hygiene lint that prevents relative imports like, for example, `..crypto.hash`.

#### Cute Animal Picture


gurukamath (approved):
LGTM!! This also closes #641 right?

SamWilsn:
It does!

PR #1377, opened 2025-08-20
refactor(weld): Full WELD

gurukamath:
### What was wrong?
The EELS and EEST repositories need to be merged and are currently separate

### How was it fixed?
Move the framework from eest to `src/ethereum_spec_tests` and the eest tests to `tests`

#### Cute Animal Picture
Cute Animals - 1 of 1


gurukamath:
I have temporarily disabled some of the static checks on the framework and tests folders since that is not the focus of this PR (See the updates to `tox -e static` in tox.ini). These will be re-enabled once the tooling is aligned. I will create an issue to track this

SamWilsn:
IMO, we should import EEST with `git subtree` if we can.

pyproject.toml
```diff
@@ -152,6 +239,12 @@ packages = [
...
 "ethereum_spec_tools" = ["py.typed"]
+"ethereum_test_forks" = ["forks/contracts/*.bin"]
+"pytest_plugins.execute" = ["eth_config/networks.yml"]
+"pytest_plugins" = ["eels_resolutions.json"]
+"cli.eest.make" = ["templates/*.j2"]
+"cli.pytest_commands" = ["pytest_ini_files/*.ini"]
+"ethereum_test_types" = ["kzg_trusted_setup.txt"]
 
 [project.optional-dependencies]
 test = [
```
SamWilsn, line 249: Should we define a `fill` target separately from `test`?

pyproject.toml
```diff
@@ -152,6 +239,12 @@ packages = [
     "assets/blocks/block_12964999.json",
 ]
 "ethereum_spec_tools" = ["py.typed"]
+"ethereum_test_forks" = ["forks/contracts/*.bin"]
+"pytest_plugins.execute" = ["eth_config/networks.yml"]
+"pytest_plugins" = ["eels_resolutions.json"]
```
SamWilsn: Do we resolver any more?

eest_tests/conftest.py
SamWilsn, line 1: Shouldn't there be a modification to `.gitmodules` in this commit?

gurukamath:
> IMO, we should import EEST with `git subtree` if we can.

At some point though, we would have to move the files over. So I just decided to pull the trigger in this PR. Do you think we should take the subtree approach for now and have it run that way for a while?

Also, there are a few minor updates to the files from `execution-spec-tests` which might make maintaining the subtree a bit challenging.

pyproject.toml
gurukamath: Updated

pyproject.toml
gurukamath: Ah. I missed the config here. Updated

eest_tests/conftest.py
gurukamath: Updated

src/ethereum_spec_tests/ethereum_clis/clis/execution_specs.py
```diff
@@ -126,3 +131,58 @@ def evaluate(
...
+        TransactionException.NONCE_IS_MAX: "NonceOverflowError",
+        TransactionException.GAS_ALLOWANCE_EXCEEDED: "GasUsedExceedsLimitError",
+        TransactionException.GAS_LIMIT_EXCEEDS_MAXIMUM: "TransactionGasLimitExceededError",
+        BlockException.SYSTEM_CONTRACT_EMPTY: "System contract address",
+        BlockException.SYSTEM_CONTRACT_CALL_FAILED: "call failed:",
+        BlockException.INVALID_DEPOSIT_EVENT_LAYOUT: "deposit",
+    }
+    mapping_regex: ClassVar[Dict[ExceptionBase, str]] = {
+        TransactionException.INSUFFICIENT_MAX_FEE_PER_GAS: (
+            r"InsufficientMaxFeePerGasError|InvalidBlock"  # Temporary solution for issue #1981.
```
Carsons-Eels, line 187: [ethereum/execution-spec-tests#1981](https://github.com/ethereum/execution-spec-tests/issues/1981) is now closed in EEST. Should probably raise a corresponding issue in this codebase, and update the comment to reference that issue # instead.

Carsons-Eels:
One downside of copy-pasting the code is that it wipes out the git history of EEST inside our repo. `git blame` will now just say you are responsible for every line in those migrated files. `git subtree` would preserve the full commit history of EEST (as long as we don't `--squash` the merge commit that subtree generates), which I think is important. Once the EEST project history is fully integrated into our own, we can just ask everyone to re-open their existing EEST PRs against the EELS codebase and stop making changes to EEST right?

I see your point about the changes to existing `ethereum-spec-tests` code though, we'd have to work out how to cleanly move it, possibly by spliting this into 2 commits. One to import the subtree, another to adjust the files necessary to link everything together into a working project.

src/ethereum_spec_tests/ethereum_clis/clis/execution_specs.py
gurukamath: Have updated the PR using the subtree approach and the latest commits in eest including this one.

gurukamath:
@SamWilsn @Carsons-Eels Updated the PR using the subtree approach. That's a lot of files updated and commits added :)

SamWilsn (approved):
As far as I can tell, this looks good to me!

gurukamath:
> As far as I can tell, this looks good to me!

Awesome. Since this is such a huge PR, I am not gonna merge it just yet. Will just let it sit for a couple of days so people can come back and take a look if needed.

PR #1381, opened 2025-08-21
feat: Block-Level Access List support for Amsterdam

fselmo:
### What was wrong?

This builds on #1338 but moves Block Access List (BAL) support to Amsterdam. I worked on this branch in conjunction with the EEST branch (ethereum/execution-spec-tests#2067) to get them talking to each other and they are currently working well. This is meant to be an initial implementation for BAL and for Amsterdam support in general. I fully expect that we will iterate on these changes as new PRs on top of `forks/amsterdam`.

### How was it fixed?

Notable updates from #1338:

- Added the index tracking to t8n tool
- Passes the BAL object via t8n to be consumed by an appropriate pydantic model on EEST side
- Passes the BAL hash over t8n
- Any updates to the spec are being added here as well

#### Cute Animal Picture

Screenshot 2025-08-27 at 16 11 26


src/ethereum/amsterdam/fork.py
```diff
@@ -0,0 +1,1119 @@
...
+
+    # Set system transaction index for pre-execution system contracts
+    # EIP-7928: System contracts use bal_index 0
+    set_transaction_index(change_tracker, 0)
+
+    process_unchecked_system_transaction(
+        block_env=block_env,
+        target_address=BEACON_ROOTS_ADDRESS,
+        data=block_env.parent_beacon_block_root,
+        change_tracker=change_tracker,
```
gurukamath: Would it make sense to add the `change_tracker` to the `block_env` instead of having it as a new paramater to the all the transaction processing functions? That way most of the logic lives in the spec code and we have to make minimal adaptions to the `t8n` code

src/ethereum/amsterdam/vm/instructions/storage.py
```diff
@@ -0,0 +1,209 @@
...
+                    GAS_STORAGE_UPDATE - GAS_COLD_SLOAD - GAS_WARM_ACCESS
+                )
+
+    charge_gas(evm, gas_cost)
+    if evm.message.is_static:
+        raise WriteInStaticContext
+
+    # Track storage write BEFORE modifying state
+    # so we capture the correct pre-value
+    if evm.message.change_tracker:
```
gurukamath: Are there going to be instances in Amsterdam that would not have a `change_tracker`?

src/ethereum/amsterdam/block_access_lists/tracker.py
```diff
@@ -0,0 +1,341 @@
...
+    """
+
+    pre_storage_cache: Dict[tuple, U256] = field(default_factory=dict)
+    """
+    Cache of pre-transaction storage values, keyed by (address, slot) tuples.
+    This cache is cleared at the start of each transaction to track values
+    from the beginning of the current transaction.
+    """
+
+    current_block_access_index: int = 0
```
gurukamath: This could be a `Uint` instead of int

src/ethereum/amsterdam/vm/instructions/environment.py
```diff
@@ -0,0 +1,617 @@
...
+    else:
+        evm.accessed_addresses.add(address)
+        charge_gas(evm, GAS_COLD_ACCOUNT_ACCESS)
+
+    # OPERATION
+    # Non-existent accounts default to EMPTY_ACCOUNT, which has balance 0.
+    balance = get_account(evm.message.block_env.state, address).balance
+
+    if evm.message.change_tracker:
+        from ...block_access_lists.tracker import track_address_access
```
gurukamath: If a module level import of this does not create any circular dependency, we should just go with module level imports

src/ethereum/amsterdam/vm/instructions/storage.py
gurukamath: This change has to be made in multiple parts of the code.

src/ethereum/amsterdam/block_access_lists/builder.py
```diff
@@ -0,0 +1,414 @@
...
+        ref:ethereum.amsterdam.vm.instructions.environment.balance
+    [`EXTCODESIZE`] :
+        ref:ethereum.amsterdam.vm.instructions.environment.extcodesize
+    [`EXTCODECOPY`] :
+        ref:ethereum.amsterdam.vm.instructions.environment.extcodecopy
+    """
+    ensure_account(builder, address)
+
+
+def build(builder: BlockAccessListBuilder) -> BlockAccessList:
```
gurukamath: Since this is a standalone function and not a method, a more descriptive name would be helpful. `build_block_access_list` for example

src/ethereum/amsterdam/fork.py
```diff
@@ -0,0 +1,1119 @@
...
+    process_unchecked_system_transaction(
+        block_env=block_env,
+        target_address=HISTORY_STORAGE_ADDRESS,
+        data=block_env.block_hashes[-1],  # The parent hash
+        change_tracker=change_tracker,
+    )
+
+    # EIP-7928: Transactions use bal_index 1 to len(transactions)
+    for i, tx in enumerate(map(decode_transaction, transactions)):
+        set_transaction_index(change_tracker, i + 1)
```
gurukamath: `set_transaction_index` call here should perhaps become a part of the `process_transaction` function.

src/ethereum/amsterdam/fork.py
```diff
@@ -0,0 +1,1119 @@
...
+        process_transaction(
+            block_env, block_output, tx, Uint(i), change_tracker
+        )
+
+    # EIP-7928: Post-execution uses bal_index len(transactions) + 1
+    post_execution_index = len(transactions) + 1
```
gurukamath: If we decide to make the index into a `Uint`, we can use the `ulen` function from `ethereum_types.numeric`

src/ethereum/amsterdam/fork.py
```diff
@@ -0,0 +1,1119 @@
...
+    set_transaction_index(change_tracker, post_execution_index)
```
gurukamath: Should this `set_transaction_index` call happen before or after the withdrawals and general purpose requests? Here it is being called before whereas in the `t8n` tool it is being called after. See [here](https://github.com/fselmo/execution-specs/blob/70c0f9e5c6685e57e76b4698129a7fdd1e67796c/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L366)

src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -323,6 +329,91 @@ def update(self, t8n: "T8N", block_env: Any, block_output: Any) -> None:
             self.requests = block_output.requests
             self.requests_hash = t8n.fork.compute_requests_hash(self.requests)
 
+        if hasattr(block_output, "block_access_list_builder"):
+            from ethereum.amsterdam.block_access_lists import (
```
gurukamath: Any function/dataclasses that the `t8n` tool needs access to should be added to `ethereum_spec_tools.evm_tools.loaders.fork_loader.Fork` and can then be accessed via `t8n.fork.XYZ`. See `compute_requests_hash` above. Otherwise this will always import from `amsterdam` which won't be right for the future forks.

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -12,6 +12,13 @@
 from ethereum_types.numeric import U64, U256, Uint
 
 from ethereum import trace
+from ethereum.amsterdam.block_access_lists import (
```
gurukamath: Fork specific imports need to be accessed via the fork loader. This is similar to the case I highlighted in `ethereum_spec_tools.evm_tools.t8n.t8n_types`

src/ethereum/amsterdam/block_access_lists/rlp_utils.py
```diff
@@ -0,0 +1,230 @@
...
+                nonce_changes_list,
+                code_changes_list,
+            ]
+        )
+
+    encoded = rlp.encode(cast(Extended, account_changes_list))
+    return Bytes(encoded)
+
+
+def validate_block_access_list_against_execution(
```
fselmo, line 121: @nerolation this doesn't look like it's currently being used anywhere. Was there a goal to use this within the spec?

src/ethereum/amsterdam/block_access_lists/builder.py
fselmo: Agreed. Done 👍🏼

src/ethereum/amsterdam/fork.py
fselmo: ~~I fixed this in t8n a while ago. That isn't pointing to the correct code. This is the current state of t8n [here](https://github.com/fselmo/execution-specs/blob/feat/amsterdam-fork-and-block-access-lists/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L290-L321).~~

fselmo: Correction, I'm very wrong. I fixed another index issue. You fixed this 😆 in your commit. Marking as resolved, the t8n change is the correct one 👍🏼

fselmo:
@gurukamath Thanks for the review 🙏🏼. I went through all your outstanding comments here and made the changes or marked resolved what you fixed in your commit. Everything should be resolved from your first pass here 🙂

src/ethereum/amsterdam/block_access_lists/rlp_utils.py
nerolation: My original thought was that it might be helpful for debugging.
You can check a BAL is correct by building it during execution, hashing it, and comparing the hashes. You won't know what's wrong though.
Happy to remove it!

src/ethereum_spec_tools/evm_tools/loaders/fork_loader.py
```diff
@@ -125,7 +125,7 @@ def signing_hash_155(self) -> Any:
     @property
     def build(self) -> Any:
         """build function of the fork"""
-        return self._module("block_access_lists").build
+        return self._module("block_access_lists").build_block_access_list
```
gurukamath, line 135: Conventionally, we have kept the method name in `Fork` to be same as the function we are importing. That way it is easier to read what is being used in `t8n`. So if we rename the imported function `build_block_access_list`, we should also update the method name.

src/ethereum/amsterdam/rlp_types.py
```diff
@@ -0,0 +1,133 @@
+"""
```
gurukamath, line 1: The name of this module might be a  bit misleading. It does not make it clear that these are block access list specific types. Wondering if these dataclasses should live in the `block_access_lists` init file

src/ethereum/amsterdam/fork.py
```diff
@@ -0,0 +1,1093 @@
...
+
+    Parameters
+    ----------
+    block_env :
+        The block scoped environment.
+    target_address :
+        Address of the contract to call.
+    data :
+        Data to pass to the contract.
+    change_tracker :
```
gurukamath: `change_tracker` is no longer a parameter to the function

src/ethereum/amsterdam/rlp_types.py
nerolation: I think we should rename it but avoid putting everything into the init file as we'd get into a circular import (which is fine if we add the respective # noqa flag but not really clean).
I'd prefer moving them into the block_access_lists dir? Wdyt?

src/ethereum_spec_tools/evm_tools/loaders/fork_loader.py
nerolation: Changes ready here:
https://github.com/fselmo/execution-specs/pull/5

kevaundray:
@fselmo on 

>  If we want to refactor this to be for every fork, we can do so either here or create an issue and track in another PR once this is merged for easier tracking.

Let me know what you think is best -- the reason we want it for every fork is so that we can run the full test suite on the "zk EL"

fselmo:
> the reason we want it for every fork is so that we can run the full test suite on the "zk EL"

I think I just wanted to know if we did want this for every fork. Is the answer here yes then? I'm in the process of moving amsterdam fork out of here and cleaning this up with a rebase. Once this is done I think we can start shaping this up a bit more. But yeah, clarity on that from you @kevaundray and @SamWilsn would be good and we can refactor it here.

kevaundray:
> > the reason we want it for every fork is so that we can run the full test suite on the "zk EL"
> 
> I think I just wanted to know if we did want this for every fork. Is the answer here yes then? I'm in the process of moving amsterdam fork out of here and cleaning this up with a rebase. Once this is done I think we can start shaping this up a bit more. But yeah, clarity on that from you @kevaundray and @SamWilsn would be good and we can refactor it here.

State tracking would be useful for the zkEVM team or to be more precise, the ability to eventually generate an ExecutionWitness akin to reth's `debug_executionWitness` -- Geth also has this endpoint though they are not compatible with Reth, which is one of the reasons we'd like to have the specs be the truth.

Would leave Sam to opine on the execution specs side

fselmo:
This will have to get cleaned up again once #1425 goes in to `forks/amsterdam` but it is at least now rebased from it at the moment. I think this is safe to add to again now that we went from `osaka` -> `amsterdam` -> `forks/amsterdam` 🎢 and it's back to passing all tests with EEST.

@nerolation please let me know the order of changes you'd like me to look at if there are any outstanding PRs that are still relevant. Sorry for the mess here but we needed to clean things up with some large `forks/osaka` changes 😅.

src/ethereum/amsterdam/rlp_types.py
fselmo: I moved it into `/block_access_lists` 👍🏼

nerolation:
This bug fix can be reviewed:
https://github.com/fselmo/execution-specs/pull/6

fselmo:
replaced by #1719

PR #1383, opened 2025-08-22
chore: State tracking in State object

kevaundray:


src/ethereum/cancun/state_tracking.py
```diff
@@ -0,0 +1,135 @@
...
+        tracker.main_trie_accessed_keys.add(address)
+    elif access_type in [STORAGE_READ, STORAGE_WRITE]:
+        if address not in tracker.storage_accessed_keys:
+            tracker.storage_accessed_keys[address] = set()
+        if key is not None:
+            tracker.storage_accessed_keys[address].add(key)
+
+# Dummy method
+def generate_merkle_proof_requests(state) -> Tuple[List[Address], List[Tuple[Address, Bytes32]]]:
+    """
```
kevaundray, line 120: Left this here to show what the next addition may look like -- will delete

src/ethereum/cancun/state_tracking.py
```diff
@@ -0,0 +1,135 @@
...
+        value_after=value_after,
+    )
+    tracker.accesses.append(access)
+    
+    if access_type in [ACCOUNT_READ, ACCOUNT_WRITE]:
+        tracker.main_trie_accessed_keys.add(address)
+    elif access_type in [STORAGE_READ, STORAGE_WRITE]:
+        if address not in tracker.storage_accessed_keys:
+            tracker.storage_accessed_keys[address] = set()
+        if key is not None:
```
jsign, line 115: Might be safer to also raise an error if `key is None`?
For a `STORAGE_READ` or `STORAGE_WRITE` sounds like if this sitaution happens, it should be an error?

kevaundray: Yep thats true -- @fselmo , this or the previous one were PRs that I was trying to merge into the execution specs, just want to confirm again that you will be adding the same functionality with your BAL PR?

fselmo: > Yep thats true -- @fselmo , this or the previous one were PRs that I was trying to merge into the execution specs, just want to confirm again that you will be adding the same functionality with your BAL PR?

Thanks for the ping! I brought this up recently in the STEEL meeting. I've mostly been coordinating that PR with the testing side and @nerolation has been implementing most of the specs for it. It would make sense to me though to get these changes here dialed in and approved and then rebase our changes off of these to use the same tracker? What do you think?

fselmo: We already have a working implementation for BALs. I think we should consolidate the ideas from here and in the BALs PR. I'm not sure what the differences in use / design are between the implementations but it would be worth not duplicating work and syncing on this.

kevaundray: I believe the BALs work is a superset of what we need -- I can setup a group with Me + Ignacio + Toni + other STEEL members so we can dial this in

nerolation: Happy to join that group, yeah.

SamWilsn:
Should we close this or leave it open and port it to `forks/amsterdam`?

kevaundray:
Think we can close it and whomever is working on #1865 can pull in anything relevant

PR #1384, opened 2025-08-24
feat(specs): Add EIP-7981 implementation

nerolation:
First draft implementation of EIP-7981, which is going to be proposed for Glamsterdam.


src/ethereum/osaka/transactions.py
```diff
@@ -568,6 +584,40 @@ def validate_transaction(tx: Transaction) -> Tuple[Uint, Uint]:
...
+    Returns a tuple of (zero_bytes, nonzero_bytes).
+    
+    [EIP-7981]: https://eips.ethereum.org/EIPS/eip-7981
+    """
+    zero_bytes = 0
+    nonzero_bytes = 0
+    
+    for access in access_list:
+        # Count bytes in address (20 bytes)
+        for byte in access.account:
```
gurukamath: Counting zero and non-zero bytes is being used in a couple of other places as well, including for call data in `calculate_intrinsic_cost`. Perhaps it would be a good idea to refactor it into a function in the `ethereum.utils.byte` module.

nerolation: yeah agree, makes sense

gurukamath:
@nerolation Does #2144 supersede this? If yes, I think this can be closed. wdyt?

PR #1400, opened 2025-09-03
chore(packaging): add all eest entrypoints; rename `ethereum-spec-fill` -> `fill`

danceratopz:
### What was wrong?

All of EEST's entrypoints need to be exposed for the full weld.

### How was it fixed?

Added the entrypoints.

**Note:** This is required in some shape or form, but it breaks EELS convention of prefixing `ethereum-spec-` to entrypoint names. This PR can serve as a discussion whether that's ok :)

#### Cute Animal Picture

image


SamWilsn:
> This is required in some shape or form, but it breaks EELS convention of prefixing `ethereum-spec-` to entrypoint names. This PR can serve as a discussion whether that's ok :)

IIRC the `ethereum-spec-` prefix grew up around not conflicting with geth's `evm`. I don't mind keeping your original executable names.

SamWilsn (approved):

PR #1401, opened 2025-09-03
feat(spec-specs): Implement EIP-7778 Block Gas Accounting without Refunds

nerolation:
This is the implementation of EIP-7778, which is proposed for Glamsterdam.


jochem-brouwer:
I love one-liner EIPs :heart_eyes:

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -980,7 +980,7 @@ def process_transaction(
     for address in tx_output.accounts_to_delete:
         destroy_account(block_env.state, address)
 
-    block_output.block_gas_used += tx_gas_used_after_refund
+    block_output.block_gas_used += tx_gas_used_before_refund
```
gurukamath: The EIP says [here](https://github.com/ethereum/EIPs/blob/251ca7dd04c7dfac19ce22261ab119379d40b126/EIPS/eip-7778.md?plain=1#L36) that "receipts store the post-refund `gas_used` value" but in this implementation, they store the cumulative gas used before the refund has been applied.

nerolation: Good catch, yeah, you're right. We have discussed this in December and specified in the EIP but seems like I forgot to update this PR too.

nerolation:
I'm wondering how important having a 100% correct gas_used (after refuds) in the receipt is for users/devs.
Putting the gas_used before refunds into it would make it easier to validate that the block gas limit was obeyed - just by using receipts. This is useful for parallel execution. Otherwise clients would need to keep track of the refunds of each transactions and add them to the gas_used field in the receipt if they want to check that a block is valid after executing it in parallel.

gurukamath:
> I'm wondering how important having a 100% correct gas_used (after refuds) in the receipt is for users/devs. Putting the gas_used before refunds into it would make it easier to validate that the block gas limit was obeyed - just by using receipts. This is useful for parallel execution. Otherwise clients would need to keep track of the refunds of each transactions and add them to the gas_used field in the receipt if they want to check that a block is valid after executing it in parallel.

Yeah, it makes sense in my mind to keep these two in sync. Unless there is some other use case here that I am overlooking.
@jochem-brouwer Is there any use case that you had in mind when you first raised this?

jochem-brouwer:
CC @nerolation since it's their quote :wink: @gurukamath 

I'll nevertheless chime in here:

> Otherwise clients would need to keep track of the refunds of each transactions and add them to the gas_used field in the receipt if they want to check that a block is valid after executing it in parallel.

Right, but this has to be done anyways, right? If we want to retrieve this from the receipts, we have to either get the receipts from other clients (IIRC this is only done in sync mode) or we have to re-run the block. In the later case, we quickly find out the block is invalid. In the sync phase, we can also invalidate the block, but this is a historical block, so our CL should not point us to that head.

However, we might think if we have EIP-7778 in combination with BALs EIP-7928 if it makes sense to change this field in the receipts. For instance, can we retrieve these fields from BALs? Or is this not what we want (since BALs have a limited (?) time to be queried?).
If we want to figure out how much the EOA paid vs how much gas was used (minus refunds) we could attempt to query the balance form the BAL which is burned, or what is sent to coinbase. However, the latter might not work in all cases, sine for 7702-accounts we could send "extra" ETH to coinbase, which yields extra transfers from the EOA to coinbase, and might look from an external point of view as extra gas spent :thinking: 

I think it might be beneficial to think of the combination of this EIP with BALs to see if we can improve both EIPs if those are shipped in the same fork.

nerolation:
yeah I agree. I was confusing the response of eth_getTransactionReceipt with the `Receipt` including the cumulative gas used.
In this case we may want to again move back to adding the gas_used before refunds to the cumulative gas used. Then we can use it for enforcing the block gas limit under BALs.

gurukamath:
Linking the EIPs [PR](https://github.com/ethereum/EIPs/pull/11075)

danceratopz:
@nerolation @gurukamath just a heads up, I just created a new target branch `eips/amsterdam/eip-7778` for this EIP and changed the PR's target branch to point to that.

src/ethereum/forks/amsterdam/fork.py
gurukamath: What happens if `calldata_floor_gas_cost` > `tx_gas_used_before_refund`?

nerolation: In that case the `calldata_floor_cost` must be used

nerolation: Let me investigate that further

nerolation:
@gurukamath @jochem-brouwer would love your review of the latest commit.

gurukamath:
> @gurukamath @jochem-brouwer would love your review of the latest commit.

The new commit handles the `calldata_floor_gas_cost` concern I had in my earlier comment. Needs to be clarified in the EIP text.

Edit: Please see my next comment

gurukamath:
> > @gurukamath @jochem-brouwer would love your review of the latest commit.
> 
> The new commit handles the `calldata_floor_gas_cost` concern I had in my earlier comment. Needs to be clarified in the EIP text.

@nerolation Actually, on closer thought, if we have a case where `tx_gas_used_after_refund < calldata_floor_gas_cost < tx_gas_used_before_refund`, we will end up in a situation where the user has used `calldata_floor_gas_cost` gas but the `block_gas_limit` is reduced by `tx_gas_used_before_refund`. I think this is a disconnect

nerolation:
> ... we will end up in a situation where the user uses `calldata_floor_gas_cost` wheras the block_gas_limit is reduced by `tx_gas_used_before_refund` which I think is a disconnect

True but I'd say that's fine. We still won't allow users to use refunded gas of earlier transaction and users would still pay the floor cost in case it is larger than the regular gas cost. In case the floor is larger than the tx_gas_used_before_refund we may want to substract the higher floor cost from the available gas.

nerolation:
> The new commit handles the `calldata_floor_gas_cost` concern I had in my earlier comment. Needs to be clarified in the EIP text.

True yeah. Did it here now:
https://github.com/ethereum/EIPs/pull/11090

gurukamath:
> True but I'd say that's fine. We still won't allow users to use refunded gas of earlier transaction and users would still pay the floor cost in case it is larger than the regular gas cost. In case the floor is larger than the tx_gas_used_before_refund we may want to substract the higher floor cost from the available gas.

Makes sense. Definitely a candidate here for potential test cases.

nerolation:
One thing we must consider here is how `eth_getTransactionReceipt`, which contains the `gas_used` , would look like.
We must make sure that the `gas_used` in the response is the gas that the user pays for (after the refund).

There are two options:
1) If we go for the gas used BEFORE the refund in the receipt's cumulative gas used, then, in order to return the "right" gas used for the user in `eth_getTransactionReceipt` clients would need to re-execute the transaction, determine the refund and subtract the refund from `CumulativeGasUsed[i] - CumulativeGasUsed[i-1]`.

2) If we go for gas used AFTER the refund in the receipt's cumulative gas used, then, in order to return the "right" gas used for the user in `eth_getTransactionReceipt` clients could just do `CumulativeGasUsed[i] - CumulativeGasUsed[i-1]` for determining the tx's gas used. This also means we need to enforce the block gas limit separetely (not using the gas used after refunds).

For BALs, txs execute on different cores and we might want to use the receipts to enforce the gas limit.
This works smoothly if we go with (1) as we could directly use the cumulative gas used value.
If we go with (2), we cannot directly use it for enforcing the block gas limit. Instead, we would need to add the refund of each transaction to its gas_gas used value (`CumulativeGasUsed[i] - CumulativeGasUsed[i-1]` + refund) again.

Another thing we could do is adding refund_amount to receipts. This would solve it too.

morph-dev:
I would like to mention that there are two more options, both with significant complications:
1. We can add another field to the receipt (e.g. `gasPaid`)
2. We can add refund as a Log

nerolation:
The latest commit now implements option (2), using the gas used AFTER refunds for receipt.
EIP reflects those changes too: https://github.com/ethereum/EIPs/pull/11075
cc @yperbasis

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -980,11 +980,14 @@ def process_transaction(
     for address in tx_output.accounts_to_delete:
         destroy_account(block_env.state, address)
 
-    block_output.block_gas_used += tx_gas_used_after_refund
+    block_output.block_gas_used += max(
```
gurukamath: I am just wondering if the variable names are starting to get a bit confusing. What do you think about renaming `tx_gas_used_after_refund` in Line 953 as `tx_gas_used` (i.e. `tx_gas_used = max(tx_gas_used_after_refund, calldata_floor_gas_cost`))? And then defining `block_gas_used_in_tx = max(tx_gas_used_before_refund, calldata_floor_gas_cost)` in the very next line.

Then proceeding to use these.

nerolation: yeah, agree! lmk what you think about the last commit.

benaadams:
> I would like to mention that there are two more options, both with significant complications:
> 
> 1. We can add another field to the receipt (e.g. `gasPaid`)

Or repurpose the fields:
```
typeByte || RLP([status, cumulativeGasUsed, logsBloom, logs])
```
=>
```
typeByte || RLP([status, gasUsed, gasSpent, logs])
```

nerolation:
> Or repurpose the fields:
> 
> ```
> typeByte || RLP([status, cumulativeGasUsed, logsBloom, logs])
> ```
> 
> =>
> 
> ```
> typeByte || RLP([status, gasUsed, gasSpent, logs])
> ```

Right but this requires changing the receipt format which we should try to avoid if possible to keep complexity low.
As we're doing it now, we're fine and we might not need any further, deeper changes to object structures.

yperbasis:
Speaking from Erigon's perspective. It's much easier for us to go with Option 1 (`CumulativeGasUsed` in receipts is consistent with block gas limit accounting) and replace `logsBloom` in receipts with `gasSpent`. I also think it's the safest/more logical/BAL-friendly option.

nerolation:
> Speaking from Erigon's perspective. It's much easier for us to go with Option 1 (`CumulativeGasUsed` in receipts is consistent with block gas limit accounting) and replace `logsBloom` in receipts with `gasSpent`. I also think it's the safest/more logical/BAL-friendly option.

Ah, then I misunderstood you. I thought you were advocating for option 2, having the cumulative gas in the receipt being consistent with what users pay.
in order to populate the `gas_used` in `eth_getTransactionReceipt` without re-execution you'd then need this gas_spent value, coming with changing the receipt.

nerolation:
Just to confirm @yperbasis , the current specs in this PR are NOT what you advocated for?

yperbasis:
> Just to confirm @yperbasis , the current specs in this PR are NOT what you advocated for?

Correct. I advocated for CumulativeGasUsed in receipts to be consistent with block gas limit accounting.

gurukamath:
Additional context from the ACDT discussion from today. The clients seem to be ok with the current spec of adding a *new* field (`gas_spent`) to the receipt.

gurukamath (approved):

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -550,6 +550,7 @@ def make_receipt(
     tx: Transaction,
     error: Optional[EthereumException],
```
fdtwd8vv45-sketch, line 551: Great

PR #1405, opened 2025-09-03
chore(weld): update latest json and EEST commits

gurukamath:
### What was wrong?
The EEST subtrees should be updated to include the latest commits

### How was it fixed?
Update subtrees and update the json fixtures

#### Cute Animal Picture

Cute Animals - 1 of 1


src/ethereum_spec_tests/ethereum_clis/clis/execution_specs.py
```diff
@@ -186,4 +186,4 @@ class ExecutionSpecsExceptionMapper(ExceptionMapper):
         TransactionException.INSUFFICIENT_MAX_FEE_PER_GAS: (
```
gurukamath, line 181: Note to reviewers: This is the only file which needs to be updated after pulling in the latest commits. The code here was adapted to work directly with the EELS repo post merge. All other updates are just pulling in the latest EEST commits

SamWilsn (approved):

PR #1409, opened 2025-09-04
fix: block difficulty calc frontier

souradeep-das:
### What was wrong?

Frontier block difficulty calculation was broken

### How was it fixed?

Type conversion

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/frontier/fork.py
+++ b/src/ethereum/frontier/fork.py
@@ -768,7 +768,7 @@ def calculate_block_difficulty(
     # See https://github.com/ethereum/go-ethereum/pull/1588
     num_bomb_periods = (int(block_number) // 100000) - 2
     if num_bomb_periods >= 0:
-        difficulty += 2**num_bomb_periods
+        difficulty += Uint(2**num_bomb_periods)
 
     # Some clients raise the difficulty to `MINIMUM_DIFFICULTY` prior to adding
     # the bomb. This bug does not matter because the difficulty is always much


src/ethereum/frontier/fork.py
```diff
@@ -768,7 +768,7 @@ def calculate_block_difficulty(
     # See https://github.com/ethereum/go-ethereum/pull/1588
     num_bomb_periods = (int(block_number) // 100000) - 2
     if num_bomb_periods >= 0:
-        difficulty += 2**num_bomb_periods
+        difficulty += Uint(2**num_bomb_periods)
```
SamWilsn, line 771:
```python
num_bomb_periods = (block_number // Uint(100000)) - Uint(2)
if num_bomb_periods >= 0:
    difficulty += Uint(2) ** num_bomb_periods
```

Kinda just thinking out loud, but would it be better in your opinion to do the math as `Uint`, or do the conversion at the end?

SamWilsn (commented):
Huh, is this broken all the way up to the merge?

souradeep-das:
the other forks seems to be okay with the calculation - looks like only a problem with frontier

src/ethereum/frontier/fork.py
souradeep-das: yeah was thinking the same! that would make it consistent with the other forks. Also do you think in that case converting difficulty to int for doing the calculations is better to be consistent?

souradeep-das:
```
difficulty = int(difficulty)
num_bomb_periods = (int(block_number) // 100000) - 2
if num_bomb_periods >= 0:
    difficulty += 2**num_bomb_periods

return max(Uint(difficulty), MINIMUM_DIFFICULTY)
```

is this better?

SamWilsn: Does Frontier have a minimum difficulty?

souradeep-das: seems like it - https://github.com/ethereum/execution-specs/blob/forks/osaka/src/ethereum/frontier/fork.py#L776C28-L776C46

SamWilsn (approved):

SamWilsn:
What you had is fine, now that I've looked a bit more deeply at it. Post-Frontier forks convert everything to `int` before doing the math, so if we want to fix this, we should do it for all the forks.

PR #1410, opened 2025-09-04
Move point at infinity check before curve check for p256 verify

kclowes:
### What was wrong?

[Code coverage](https://app.codecov.io/gh/ethereum/execution-specs/pull/1388/blob/src/ethereum/osaka/vm/precompiled_contracts/p256verify.py?dropdown=coverage#L76) highlighted that we aren't hitting the point at infinity check because `is_on_curve_secp256r1` catches the point at infinity and returns before it can reach the check for (0, 0).

### How was it fixed?

Moved the point at infinity check ahead of the `is_on_curve...` check, so we don't run through the method if we already know it won't pass. The other option is to remove the check, but I think this makes it more explicit that we're checking that point at infinity, and may be a tiny bit more performant.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/osaka/vm/precompiled_contracts/p256verify.py
+++ b/src/ethereum/osaka/vm/precompiled_contracts/p256verify.py
@@ -67,13 +67,13 @@ def p256verify(evm: Evm) -> None:
     if public_key_y < U256(0) or public_key_y >= SECP256R1P:
         return
 
+    # Point should not be at infinity (represented as (0, 0))
+    if public_key_x == U256(0) and public_key_y == U256(0):
+        return
+
     # Point validity: The point (qx, qy) MUST satisfy the curve equation
     # qy^2 ≡ qx^3 + a*qx + b (mod p)
     if not is_on_curve_secp256r1(public_key_x, public_key_y):
-        return
-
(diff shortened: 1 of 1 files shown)


src/ethereum/osaka/vm/precompiled_contracts/p256verify.py
```diff
@@ -67,15 +67,15 @@ def p256verify(evm: Evm) -> None:
     if public_key_y < U256(0) or public_key_y >= SECP256R1P:
```
SamWilsn: Huh, these `< 0` checks are impossible. You can't construct a negative `U256`. Should we remove them?

SamWilsn (commented):
This looks good. Does it need to be ported to prague as well?

kclowes:
> Does it need to be ported to prague as well?

This precompile got added in Osaka, so I don't think so.

src/ethereum/osaka/vm/precompiled_contracts/p256verify.py
kclowes: Sure! Will update!

SamWilsn (approved):
Duh, I don't know why I thought this was introduced in Prague :sweat_smile:

PR #1411, opened 2025-09-04
feat: add support for EIP-7782 (6 sec slots)

morph-dev:
### What was wrong?

We don't have a spec implementation for [EIP-7782](https://eips.ethereum.org/EIPS/eip-7782) (also known as 6-second slots).

### How was it fixed?

Added initial spec implementation, to get things started.

I implemented these changes in geth ([github.com/morph-dev/go-ethereum/tree/eip7782](https://github.com/morph-dev/go-ethereum/tree/eip7782)).

**Note:** I would also like to highlight one thing. In `check_gas_limit` function, we need to have custom logic for the fork block and I wasn't sure how exactly to do it. I used `FORK_CRITERIA`, but this doesn't really work if fork is not yet scheduled.

#### Cute Animal Picture

image

**_Disclaimer: It's my first time contributing to execution specs_**


src/ethereum/osaka/fork.py
```diff
@@ -110,11 +110,10 @@
 HISTORY_STORAGE_ADDRESS = hex_to_address(
     "0x0000F90827F1C53a10cb7A02335B175320002935"
 )
-MAX_BLOCK_SIZE = 10_485_760
-SAFETY_MARGIN = 2_097_152
+MAX_BLOCK_SIZE = 5_242_880 # Halved for EIP-7782 (was 10_485_760)
+SAFETY_MARGIN = 1_048_576 # Halved for EIP-7782 (was 2_097_152)
```
benaadams, line 114: These values don't need to change; should be controlled by gas pricing they are just backstops to ensure that the blocks fit in the p2p message sizes which wouldn't be changing

felix314159:
Since Ben's feedback was still not addressed I'm closing this for being stale.

Issue #1112, opened 2025-02-07
Create a tool that syncs the chain while watching code coverage


SamWilsn:
Would probably live in [`sync.py`](https://github.com/ethereum/execution-specs/blob/master/src/ethereum_spec_tools/sync.py).

The goal of this tool would be to detect areas of our codebase that are not covered by EEST that are touched on mainnet.

Would probably look like:

1. an import phase, where the code coverage information from the tests is imported (or failing that, run the tests as part of the tool)
2. a sync phase, where transactions that increase code coverage are logged for later examination
3. automated test case generation (EEST has a tool for this I think)

Additional resources:

- https://github.com/plasma-umass/slipcover

PR #1412, opened 2025-09-05
feat: add sync with coverage tool

souradeep-das:
### What was wrong?

Related to Issue https://github.com/ethereum/execution-specs/issues/1112

### How was it fixed?
Added a script that runs live coverage of execution-specs while syncing mainnet blocks, and finds out blocks that hit code paths not covered by the baseline (json_infra tests).

This script:
1. Generates a Baseline coverage by running json_infra tests if coverage.xml doesn't exist
2. Converts coverage.xml to baseline_coverage.json if needed
3. Runs slipcover along with the sync tool

Note:
To be able to pause/run from a midpoint, `persist` needs to work (on the sync tool). If state is persisted as a part of the sync tool, the syncing (and coverage) continues from that the last persisted state. This PR does not fix `persist`

#### Cute Animal Picture

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -302,6 +302,9 @@ optimized = [
     "rust-pyspec-glue>=0.0.9,<0.1.0",
     "ethash>=1.1.0,<2",
 ]
+coverage = [
+    "slipcover",
+]
 
 [tool.setuptools.dynamic]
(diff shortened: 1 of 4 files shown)


felix314159:
IMO it is pointless to leave it open when no one is working on it

PR #1417, opened 2025-09-10
Implement Gnosis post-merge spec

dapplion:
Implements https://github.com/gnosischain/specs/

- [x] Collect base fee post paris
- [x] Custom withdrawals
- [x] Collect blob fee post prague


dapplion:
Wrong target, closing

PR #1425, opened 2025-09-12
Forks/amsterdam

fselmo:
### What was wrong?

I created Amsterdam so that we could move block-level access lists into there but this should be separated into a separate PR.

Related to Issue #1381, which will need to be rebased once this is merged to `forks/amsterdam`

### How was it fixed?

- This is all the Amsterdam setup logic by itself
- #1381 will be used to track block-level access lists support coming into `forks/amsterdam` and should be rebased on top of these changes.

#### Cute Animal Picture

Screenshot 2025-09-12 at 14 55 35


fselmo:
@gurukamath @SamWilsn ok this should be rebased with `forks/` switch and ready to go. Includes index addition for `Unscheduled` forks.

src/ethereum/forks/amsterdam/utils/address.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def to_address_masked(data: Uint | U256) -> Address:
+    """
+    Convert a Uint or U256 value to a valid address (20 bytes).
+
+    Parameters
+    ----------
+    data :
+        The numeric value to be converted to address.
```
gurukamath, line 32: Would you happen to know why we need to correct these things manually after using the new fork tool? If I look at the Osaka fork, this docstring has the correct text. I would have expected the new fork tool to simple copy over the correct text.

src/ethereum/forks/amsterdam/vm/instructions/block.py
```diff
@@ -0,0 +1,255 @@
...
+    # STACK
+    block_number = Uint(pop(evm.stack))
+
+    # GAS
+    charge_gas(evm, GAS_BLOCK_HASH)
+
+    # OPERATION
+    max_block_number = block_number + Uint(256)
+    current_block_number = evm.message.block_env.number
+    if (
```
gurukamath, line 48: Same concern with the new fork tool here I guess.

src/ethereum/forks/amsterdam/utils/address.py
fselmo: I think this would've been corrected but I had already run the fork tool and had to fix things so I just did a manual diff here instead to tweak after rebasing with osaka. Is this OK?

fselmo: I had to cherry pick the new fork tool updates (they aren't merged yet), run the tool, remove the commit (so I don't introduce it here), then cherry-pick the other things. Instead I just updated on top of my previous run of the new fork tool.

gurukamath (approved):

Issue #671, opened 2022-12-17
chore(doc): ensure all doc-strings are in markdown


gurukamath:
I see that the other docstrings use the former syntax. If we have to change the syntax, I guess it will be a good idea to create a new issue and change the syntax for all the docstrings.

_Originally posted by @gurukamath in https://github.com/ethereum/execution-specs/pull/627#discussion_r1051381885_
      

The criteria mentioned [here](https://github.com/ethereum/execution-specs/pull/1428#issuecomment-3323253209) should also be kept in mind while re-writing the doc-strings.

Khushi-Chaudhary04:
Hi @SamWilsn, I'd like to work on this issue. Could you please assign it to me?

Pazil24:
Hi @gurukamath is this issue open?

Anuoluwapo25:
hi @SamWilsn can i work on this issue?

PR #1428, opened 2025-09-16
chore(tooling): Update static checks

### LINKED ISSUE #1378: Re-enable all static checks
I have temporarily disabled some of the static checks on the framework and tests folders since that is not the focus of this PR (See the updates to `tox -e static` in tox.ini). These will be re-enabled once the tooling is aligned. I will create an issue to track this

_Originally posted by @gurukamath in https://github.com/ethereum/execution-specs/issues/1377#issuecomment-3207204050_

### LINKED ISSUE #1397: chore(tooling): add missing code formatting checks (`ruff format --check`)
#1326 replaced `flake8`, `isort` and `black` with `ruff` 🚀 

While `tox -e static` runs `ruff check` which handles the checks `isort` and `flake8` were previously responsible for:
https://github.com/ethereum/execution-specs/blob/51cabd86502df7af596f5c78a0c4e00a4f92822c/tox.ini#L11-L22

`tox` doesn't have the checks that `black` previously performed, i.e., we should still add:
```bash
ruff format --check src --exclude src/ethereum_spec_tests
```

Unfortunately, as it stands, this would reformat 389 files.

### LINKED ISSUE #1403: chore(tooling): agree on common post-weld ruff rules
To keep post-Weld tooling simple, it would help to align on a shared set of ruff rules without any deviations across EEST and EELS codebases.

### Common Ignored Rules (Both repos)

| Rule | Description | Notes |
|------|-------------|-------|
| D205 | Missing blank line after summary | Common to both (STEEL) |
| D203 | 1 blank line required before class docstring | Common to both (STEEL) |
| D212 | Multi-line docstring summary should start at the first line | Common to both (STEEL) |
| D415 | First line should end with a ".", "?", or "!" | Common to both (STEEL) |

### execution-specs Specific Ignored Rules

| Rule | Description | Notes |
|------|-------------|-------|
| D107 | Missing docstring in __init__ | EELS specific |
| D200 | One-line docstring should fit on one line with quotes | EELS specific |
| D400 | First line should end with a period | EELS specific |
| D401 | First line should be in imperative mood | EELS specific |
| D410 | Missing blank line after last section | EELS specific |
| D411 | Missing blank line before last section | EELS specific |
| D412 | No blank lines between sections | EELS specific |
| D413 | Missing blank line after last section | EELS specific (same as D410) |
| D414 | Section should end with a period | EELS specific |
| D416 | Section names should be in the correct order | EELS specific |
| E203 | Whitespace before ':' | EELS specific |

### execution-spec-tests Specific Ignored Rules

| Rule | Description | Notes |
|------|-------------|-------|
| C901 | Function is too complex | Complexity* handled differently |
| A005 | Builtin shadowing | execution-spec-tests specific |
| C420 | Unnecessary comprehension | execution-spec-tests specific |

Regarding complexity: This is configured to 7 in EELS, it's the default in EEST. 7 is fine :)

gurukamath:
(closes #1378)
(closes #1397)
(closes #1403 )

### What was wrong?
Some of the current static checks are not run on the tests folder. Also, `ruff format` is not being run.

Related to Issue #1378
Related to Issue #1397
Related to Issue #1403 

### How was it fixed?
Extend static checks to the tests folder. Only exclude the eest related code

Add `ruff format` check.

#### Cute Animal Picture

Cute Animals - 1 of 1


danceratopz (commented):
The updated tox static env looks good, @gurukamath! We just need to make sure that we're running with the appropriate ruff config, to avoid another mass format PR.

A few points on config relevant for `ruff format` :

1. To have the `ruff format`-specific config, as close to the target post-weld config as possible, I'd suggest removing this EELS-specific `ignore`:

    https://github.com/ethereum/execution-specs/blob/5ddb904fa7ba27daeff423e78466744c51e8cb6a/pyproject.toml#L483

2. What do you think about removing `E501: line-too-long` for all of `./tests/` for consistency (my preference: as we're doing this for spec and unit tests in https://github.com/ethereum/execution-spec-tests/pull/2152)? This will not only be a reformat; docstrings, comments will cause trouble.
    https://github.com/ethereum/execution-specs/blob/5ddb904fa7ba27daeff423e78466744c51e8cb6a/pyproject.toml#L492

3. We should evaluate the target post-weld docstring rules. Perhaps it's ok to subsequently apply the changes for this ruleset as a separate PR? It will need a little alignment and potentially quite some effort. This is potentially another mass reformat.

Another suggestion, not strictly required in this PR: We will need to update the path(s) here to exclude EEST. So that when we enable EEST ruff commands, these rules aren't applied to tests/eest:
https://github.com/ethereum/execution-specs/blob/5ddb904fa7ba27daeff423e78466744c51e8cb6a/pyproject.toml#L486-L492

pyproject.toml
```diff
@@ -463,34 +451,16 @@ fixable = [
...
     "D400",     # First line should end with a period
     "D401",     # First line should be in imperative mood ("Do", not "Does")
-    "D410",     # Missing blank line after last section (Args, Returns, Raises)
-    "D411",     # Missing blank line before last section
-    "D412",     # No blank lines between sections
-    "D413",     # Missing blank line after last section (same as 410)
-    "D414",     # Section should end with a period
-    "D416",     # Section names should be in the correct order
-    "E203",     # Whitespace before ':'
+    "D415",     # First line should end with a ".", "?", or "!"
```
danceratopz: D400 is stronger than D415, so D415 is moot and can be removed (also in EEST's pyproject.toml).

I would like to keep D400 though, if that's possible on your side. I did try it out on top of this PR and there are some issues with restructuredText and many the ruff auto-fixes will be need to reviewed and manually fixed-up, i.e., the output of this command is not perfect, but a good start:
```
uv run ruff check src tests --exclude src/ethereum_spec_tests,tests/eest --fix --unsafe-fixes
```

Please see the messages I posted outside of Github for more info.

danceratopz (commented):
Hey @gurukamath, I only reviewed the changes to pyproject.toml and compared them to EEST's ruff config. The `[tool.ruff.lint]` section looks great; I copied it over to EEST (enabled D400, enabled D415; disabled "A" which will be fixed by Sam's https://github.com/ethereum/execution-spec-tests/pull/2166) and it runs cleanly on our code.

Just one request below to enable D400 :pray:

pyproject.toml
gurukamath: Sure Dan. I am going to take a look at this and add D400.

danceratopz:
> I copied it over to EEST (enabled D400, enabled D415; disabled "A" which will be fixed by Sam's https://github.com/ethereum/execution-spec-tests/pull/2166) and it runs cleanly on our code.

Small correction to the above: EEST will need to enable the other ARG checks to have parity with EELS:
- https://github.com/ethereum/execution-spec-tests/issues/2176

gurukamath:
@danceratopz @Carsons-Eels Have updated the PR to stop ignoring D400

pyproject.toml
```diff
@@ -463,34 +451,15 @@ fixable = [
...
-    "D400",     # First line should end with a period
     "D401",     # First line should be in imperative mood ("Do", not "Does")
-    "D410",     # Missing blank line after last section (Args, Returns, Raises)
-    "D411",     # Missing blank line before last section
-    "D412",     # No blank lines between sections
-    "D413",     # Missing blank line after last section (same as 410)
-    "D414",     # Section should end with a period
-    "D416",     # Section names should be in the correct order
-    "E203",     # Whitespace before ':'
+    "D415",     # First line should end with a ".", "?", or "!"
```
danceratopz: I think you can safely remove this line without requiring any other changes, @gurukamath, as D400 is stronger and now enabled.

danceratopz (commented):
Thanks for enabling D400, @gurukamath. At a first glance the changes look good.

Could you just comment whether the changes were fully automated or not? I.e., how much manual resolution did you do? I assume you manually removed or scripted the `^^^` restructuredText underneath headers? What else was required apart from `ruff format`?

Can you check the comment below? If that change is fine, then I'll do a more thorough review of the changes, based on your answers to the questions (about "how") above.

pyproject.toml
gurukamath: Thanks Dan. Updated

gurukamath:
> Thanks for enabling D400, @gurukamath. At a first glance the changes look good.
> 
> Could you just comment whether the changes were fully automated or not? I.e., how much manual resolution did you do? I assume you manually removed or scripted the `^^^` restructuredText underneath headers? What else was required apart from `ruff format`?

Yes. It is a ruff check automated fix + some manual changes. An overwhelming majority of these changes are removing the underline from restructured text. There might be 2-3 slightly more specific cases of manual change. For example like the re-ordering of the restructured test anchor [here](https://github.com/ethereum/execution-specs/blob/fde14aba5b6601d967fb7dcb07cda1291d8372ce/src/ethereum/forks/dao_fork/fork.py#L2). Unfortunately, I do not recollect all such cases right now but if you want, I can re-make the commits such that all the automated changes and manual ones are in different commits. Do let me know.

danceratopz (approved):
LGTM @gurukamath. Here's what I checked:

1. I used the `ruff` settings from this PR in EEST with two exceptions:
    i. `select`: without "ARG"; this is still WIP in EEST, 
    ii. `line-length` 99 instead of 79.
2. I reviewed a random subset of the changes manually: Small comment below.
3. I asked Claude to review the PR on a per-file basis: All good.

About 2.: Tldr: It's fine from my side, but.. The only comment I have is that the titles in the rendered doc might look a little strange with a period at the end of them. We could potentially modify `docc` to remove them again? It is kinda crazy to add a period for a linter only to remove them again in the doc flow, BUT, it does allow for consistent lint rules.

The review in 1. was worth it, I found two bugs in EEST's `ruff` config :smile: 
1. We select "C", which doesn't exist. It appears ruff enables both `complexity` (C90) and `flake8-comprehensions` (C4) with this -> we should be more explicit and only enable "C4". EELS should add "C4" to its select (27 issues; separate PR)? There are other rules that start with "C", looking into whether these are enabled, or not.
2. We reverted enabling of "C420" by accident, will fix.

gurukamath:
> About 2.: Tldr: It's fine from my side, but.. The only comment I have is that the titles in the rendered doc might look a little strange with a period at the end of them. We could potentially modify `docc` to remove them again? It is kinda crazy to add a period for a linter only to remove them again in the doc flow, BUT, it does allow for consistent lint rules.

@SamWilsn What do you think of the point made here about `docc`?

gurukamath:
> 1. We select "C", which doesn't exist. It appears ruff enables both `complexity` (C90) and `flake8-comprehensions` (C4) with this -> we should be more explicit and only enable "C4". EELS should add "C4" to its select (27 issues; separate PR)? There are other rules that start with "C", looking into whether these are enabled, or not.

Here are my thoughts on the C-space rules. I am ok with ignoring C901 as we had earlier agreed. But I think we should also ignore `C401` and `C408`. The generators and calls which these rules deal with might be a bit of an overhead but they definitely make for more readable code in the spec. So I am in favour of enabling the entire `C` space and selectively disabling `C401`, `C408` and `C901`

danceratopz:
> Here are my thoughts on the C-space rules. I am ok with ignoring C901 as we had earlier agreed. But I think we should also ignore `C401` and `C408`. The generators and calls which these rules deal with might be a bit of an overhead but they definitely make for more readable code in the spec. So I am in favour of enabling the entire `C` space and selectively disabling `C401`, `C408` and `C901`

Thanks for looking at this @gurukamath. This change would be outside the scope of this PR. I haven't looked at how difficult it'd be to able the whole C space, but let's follow up on that elsewhere.

src/ethereum/crypto/blake2.py
```diff
@@ -52,40 +51,38 @@ class Blake2:
 
     @property
     def max_word(self) -> Uint:
-        """
-        Largest value for a given Blake2 flavor.
-        """
+        """Largest value for a given Blake2 flavor."""
```
SamWilsn: I am simply complaining to get this off my chest. We should go with whatever the linter prefers.

But man, do I ever _hate_ this, both stylistically and semantically. No where else in Python would you recommend using triple quotes when they are unnecessary. Like if you saw the following in a pull request, you'd tell the author to use single quotes:

```python
some_string = """hello world"""
```

You only use triple quotes when you want to avoid escaping special characters. When the docstring contains no special characters, why would you use triple quotes? Multi-line docstrings, obviously, span multiple lines, so you need the triple quotes to capture the newlines. Single line docstrings don't!

Not only that, you effectively lose four characters you could use for semantic meaning, harming information density.

The only argument I see for using triple quotes is consistency, and that's a weak argument here at best. [PEP 257](https://peps.python.org/pep-0257/)'s only justification is that it "makes it easy to later expand". Bullshit. You already have to insert a newline after the opening quote, so you're adding four extra characters. That's really not that much more work.

src/ethereum/__init__.py
```diff
@@ -1,5 +1,5 @@
 """
-### Introduction
+Introduction to Ethereum.
```
SamWilsn: If we're going to enforce that the first line be a summary, always, then it should look like this:

```suggestion
Ethereum base package.

### Introduction
```

This would render to something like:

```html
<h2>ethereum</h2>
    <p>Ethereum base package.</p>
    <h3>Introduction</h3>
        <p>Seeing as internet connections...</p>
```

src/ethereum/crypto/blake2.py
```diff
@@ -52,40 +51,38 @@ class Blake2:
...
         return Uint(2) ** self.w
 
     @property
     def w_R1(self) -> Uint:
         """
         (w - R1) value for a given Blake2 flavor.
-        Used in the function G
+        Used in the function G.
```
SamWilsn, line 63: Why are we ignoring D205? If we're going to mandate summaries, we should mandate the blank line as well.

src/ethereum/crypto/blake2.py
```diff
@@ -153,8 +149,7 @@ def G(
         self, v: List, a: Uint, b: Uint, c: Uint, d: Uint, x: Uint, y: Uint
     ) -> List:
         """
-        The mixing function used in Blake2
-        https://datatracker.ietf.org/doc/html/rfc7693#section-3.1
+        The mixing function used in Blake2.
```
SamWilsn, line 156:
```suggestion
        The mixing function used in Blake2.
        
        See [RFC 7693] for more details.
        
        [RFC 7693]: https://datatracker.ietf.org/doc/html/rfc7693#section-3.1
```

SamWilsn (approved):

SamWilsn:
> > About 2.: Tldr: It's fine from my side, but.. The only comment I have is that the titles in the rendered doc might look a little strange with a period at the end of them. We could potentially modify `docc` to remove them again? It is kinda crazy to add a period for a linter only to remove them again in the doc flow, BUT, it does allow for consistent lint rules.
> 
> @SamWilsn What do you think of the point made here about `docc`?

Titles should never have a period, whether it's in the rendered output or the source input. The practical effect of this is that we always need to have a summary line before the first title in a docstring, but I think that actually makes a lot of sense.

src/ethereum/crypto/blake2.py
danceratopz: I just came to this PR to share our findings regarding D200 and found Sam's comment. Although we're coming from a different angle, it still seems the best place to add our thoughts on this...

I (and @felix314159, I believe) have come to the same conclusion during our battle to reduce EEST docstrings to 79 from 99 in:
- https://github.com/ethereum/execution-spec-tests/pull/2180

The EELS change above due to [D200](https://docs.astral.sh/ruff/rules/unnecessary-multiline-docstring/). 

We've made all the reasonable reformatting changes to EEST docstrings, but  we still hit 263 D200 errors. An example fail is:

```python
    """
    CLI wrapper to an index of all the fixtures in the specified directory.
    """
```

which is of length < 79, i.e., D200 is purely a formatting rule.

To fix this you have two options:
1. Re-write it so that it's < 79 with 6 quotation marks on one line. Then you have less space for your docstring!
2. Add a newline at an arbitrary position, e.g.,
    ```
        """
        CLI wrapper to an index of all the fixtures in the specified
        directory.
        """
    ```

Both are silly options, imo.

@gurukamath sorry about this, as I think there are probably a non-trivial number of changes in this PR that use 2. to fix D200 (probably using `check --fix`?), but I would prefer to disable D200 in our STEEL config, it has nothing to do with line-length.

danceratopz: About 2. I just realized that this is an invalid fix if you enable D205; you only have option 1 (re-write to be < 79 with quotes). See below.

src/ethereum/crypto/blake2.py
danceratopz: I thought I'd check the consequence of enabling D205. This example errors:
```python
    """
    Generate an index file (index.json) of all the fixtures in the specified
    directory.
    """
```
As the D checks seemingly expect that the summary is on a single line, i.e., it thinks "directory" is the description. 

This makes no sense, but it fixes D205:
```python
    """
    Generate an index file (index.json) of all the fixtures in the specified

    directory.
    """
```

I'm inclined to disable D200, but enable D205. But it's late and will re-visit tomorrow.

danceratopz: Ok, enabling D205 gives 1241 errors. These would not be trivial to fix. We typically launch into multiline docstrings without a summary, so they'd have to re-written. It would result in nicer documentation though.

e.g.
```python
"""
A small pytest plugin that shows the a concise help string that only contains
the options defined by the plugins defined in execution-spec-tests.
"""
```

danceratopz: Funny example, needs fixing anyway :laughing:

src/ethereum/crypto/blake2.py
gurukamath: Sure. I think the `D200` related changes are all in their own commit, so this should be pretty easy to reverse. I'll take a look

gurukamath:
> Titles should never have a period, whether it's in the rendered output or the source input. The practical effect of this is that we always need to have a summary line before the first title in a doc-string, but I think that actually makes a lot of sense.

I am going to add this to the issue #671 which is aimed at converting all the doc-strings into markdown. I think we should handle this post weld and enforce the following.

1. All doc-strings should have the a summary at the start.
2. The summary should be a single line and should end with a `.`
3. There should be a blank line after the summary.

src/ethereum/crypto/blake2.py
gurukamath: Updated to revert the D200 related changes

src/ethereum/crypto/blake2.py
gurukamath: Will need to be handled as part of the markdown conversion. See #671

Carsons-Eels: Agreed, it does need to be fixed but can be rolled into #671

src/ethereum/exceptions.py
```diff
@@ -3,7 +3,7 @@
 """
 
 
-class EthereumException(Exception):     # noqa N818
+class EthereumException(Exception):  # noqa N818
```
Carsons-Eels, line 6: Why do we want it to read `EthereumException` rather than `EthereumError`?

Carsons-Eels (approved):

PR #1436, opened 2025-09-23
bug(tool): update path to include forks folder

gurukamath:
What was wrong?

The path prefix in the patch tool is outdated after all the fork related spec code was moved into the `forks` folder.

How was it fixed?
Add the `forks` folder to the path prefix.


SamWilsn (approved):
~~Probably worth looking at #1372~~

PR #1455, opened 2025-10-06
Tune test memory usage; disable index on pypy

SamWilsn:
## 🗒️ Description

**This might require waiting until [`11b96b80f1a9dad80766d3fedb51c3cc463b2011`](https://github.com/ethereum/execution-spec-tests/commit/11b96b80f1a9dad80766d3fedb51c3cc463b2011) is pulled in a subtree**

Four notable tweaks in here:

- Switch to `--dist=loadgroup`. See pytest-xdist's [distribution modes](https://pytest-xdist.readthedocs.io/en/latest/distribution.html#distribution-modes).
- Lower the pypy garbage collector limit to 2G (from 10G).
- Increase the max number of workers to 10.
- Disable test index generation in pypy (this takes more than 2G of memory).

On my machine, as measured by `systemd-run`, this 

#### Cute Animal Picture

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -45,7 +45,7 @@ extras =
 commands =
     fill \
         -m "not slow and not zkevm and not benchmark" \
-        -n auto --maxprocesses 6 \
+        -n auto --maxprocesses 10 --dist=loadgroup \
         --basetemp="{temp_dir}/pytest" \
         --clean \
         --until Osaka \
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -74,6 +77,7 @@ extras =
     optimized
 passenv =
     PYPY_GC_MAX
+    PYPY_GC_MIN
```
gurukamath, line 80: Since the `optimized` workflow uses CPython, I am not sure if we need to pass `PYPY_GC_MAX` and `PYPY_GC_MIN`

gurukamath (approved):

gurukamath (commented):
LGTM!

PR #1639, opened 2025-10-21
chore(pytest): Fix/enable `packages/tests` unit tests

marioevz:
## 🗒️ Description
Creates tox workflow to run `packages/tests` unit tests, and fixes/skips failing tests:
- Disable using `--t8n-server-url` flag in unit tests because we no longer support the resolver for unit tests.
- Use a relative path to tests/eest in `packages/tests/src/cli/tests/test_pytest_fill_command.py` (~~Needs issue~~ Now fixed).
- Implements `is_installed` for `ExecutionSpecsTransitionTool` in order for it to work as the default T8N tool for unit tests.
- Always pass `--output.basedir` to T8N in `ExecutionSpecsTransitionTool` (#1652).
- Disables unit tests that require `docs` folder in `packages/tests/src/ethereum_test_checklists/tests/test_checklist_template_consistency.py` and `packages/tests/src/pytest_plugins/filler/tests/test_eip_checklist.py` (#1645).
- Disables BPO tests in `packages/tests/src/pytest_plugins/forks/tests/test_markers.py` (#1646).

## 🔗 Related Issues or PRs
N/A.

#### Cute Animal Picture

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 min_version = 4.0
 requires =
     tox-uv >=0.2.0
-envlist = py3,pypy3,json_infra,static
+envlist = py3,pypy3,json_infra,static,tests_pytest
 
 [testenv]
 package = uv
@@ -21,6 +21,14 @@ commands =
     mypy
     ethereum-spec-lint
     uv lock --check
+
+
+[testenv:tests_pytest]
+# run inside the package directory so pytest picks up packages/tests/pyproject.toml
+changedir = {toxinidir}/packages/tests
+# install the package in editable mode with its test extras
+deps = -e .[test,lint]  # lint is required because gentest uses ruff format
+commands = pytest src -n auto
 
 [testenv:json_infra]
 extras =
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 min_version = 4.0
 requires =
     tox-uv >=0.2.0
-envlist = py3,pypy3,json_infra,static,tests_pytest
+envlist = py3,pypy3,json_infra,static,tests_pytest_py3,tests_pytest_pypy3
 
 [testenv]
 package = uv
@@ -23,12 +23,30 @@ commands =
     uv lock --check
 
 
-[testenv:tests_pytest]
+[testenv:tests_pytest_py3]
 # run inside the package directory so pytest picks up packages/tests/pyproject.toml
 changedir = {toxinidir}/packages/tests
 # install the package in editable mode with its test extras
 deps = -e .[test,lint]  # lint is required because gentest uses ruff format
-commands = pytest src -n auto
+commands =
+    pytest \
+        -n auto --maxprocesses 6 \
+        --basetemp="{temp_dir}/pytest" \
+        src
+
+[testenv:tests_pytest_pypy3]
+passenv =
+    PYPY_GC_MAX
+    PYPY_GC_MIN
+# run inside the package directory so pytest picks up packages/tests/pyproject.toml
+changedir = {toxinidir}/packages/tests
+# install the package in editable mode with its test extras
+deps = -e .[test,lint]  # lint is required because gentest uses ruff format
+commands =
+    pytest \
+        -n auto --maxprocesses 6 \
+        --basetemp="{temp_dir}/pytest" \
+        src 
 
 [testenv:json_infra]
 extras =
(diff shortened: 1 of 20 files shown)


tox.ini
```diff
@@ -2,7 +2,7 @@
 min_version = 4.0
 requires =
     tox-uv >=0.2.0
-envlist = py3,pypy3,json_infra,static
+envlist = py3,pypy3,json_infra,static,tests_pytest
```
danceratopz: In EEST's `tox.ini`, we used hyphens instead of underscore (e.g. `tests-deployed`, `tests-deployed-benchmark)`: 
```suggestion
envlist = py3,pypy3,json_infra,static,tests-pytest
```

tox.ini
```diff
@@ -22,6 +22,14 @@ commands =
     ethereum-spec-lint
     uv lock --check
 
+
+[testenv:tests_pytest]
```
danceratopz:
```suggestion
[testenv:tests-pytest]
```

danceratopz (commented):
LGTM! `changedir` is a good find. I think the remaining tasks are fine as follow-up issues. Just a minor nit about the tox env naming :slightly_smiling_face:

spencer-tb (approved):
LGTM as well! Nice one!
Image

I think we should find another name over pytest. I would suggest `eest-unit-tests` but maybe we should stop using EEST. Dan mentioned to me earlier that the "EEST" package will be called `ethereum-execution-tests` so what about `eet-unit-tests`?

Created 2 issues for the skips:
- https://github.com/ethereum/execution-specs/issues/1646
- https://github.com/ethereum/execution-specs/issues/1645

tox.ini
SamWilsn: Does this need a new env? Doesn't this mean that the EEST tests only run under CPython?

SamWilsn (commented):
If this is intended to be merged as-is, we should create issues for the TODO and skips.

tox.ini
marioevz: The only reason I think is due to the:
```
changedir = {toxinidir}/packages/tests
```
Because otherwise we need to change a couple of directory path references both in the code and in the pyproject.toml.

I could try (a) see if we can put it in static, or (b) create pypy3_tests_pytests?

marioevz: I just realized that we now have a discrepancy because "json_infra" is here, which uses underscores, I think we stick to underscores to not break naming in this file.

marioevz: After second thoughts, `static` does not make sense because pytest executes.

spencer-tb: Created the issues here:
https://github.com/ethereum/execution-specs/issues/1646
https://github.com/ethereum/execution-specs/issues/1645

marioevz:
Converting to draft until all issues are created, and ci workflows are added to the changes.

tox.ini
SamWilsn: EELS has a fixture for resolving paths inside tests:

https://github.com/ethereum/execution-specs/blob/fe377e704b14b4076a6b07a5e9f20f50a1105511/tests/json_infra/conftest.py#L33-L43

We added this so that pytest invocations are current directory agnostic. Would something like that work for EEST?

SamWilsn: I suppose another reason to make a new env is speeding up the code-test cycle for framework code.

spencer-tb:
Awesome work. Happy to merge from my side.

Does it make sense to make the new pytest environments run on a new workflow file filtered by `tests/`, `packages/tests`, `tox.ini` & `<workflow_file>.yaml`? Or the existing workflow file with the `dorny/paths-filter@v3` action?
```yaml
detect-changes:
    runs-on: ubuntu-latest
    outputs:
      tests: ${{ steps.filter.outputs.tests }}
    steps:
      - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955
      - uses: dorny/paths-filter@v3
        id: filter
        with:
          filters: |
            tests:
              - 'tests/**'
              - 'packages/tests/**'
              - 'tox.ini'
              - '.github/workflows/test.yaml`         
...

tests_pytest_py3:
    runs-on: [self-hosted-ghr, size-xl-x64]
    needs: [static, detect-changes]
    if: needs.detect-changes.outputs.tests == 'true' || github.event_name == 'workflow_dispatch'
    steps:
    ...
```

Personally I prefer a new workflow file. Can also be in a future PR.

tox.ini
danceratopz: > I suppose another reason to make a new env is speeding up the code-test cycle for framework code.

Yes, I (and I think EEST in general) prefer one tox environment per "task" (to allow targeted local checks) and then add a workflow job that matches that granularity to CI (for easier reporting and reduce the need to study logs to see what failed).

PR #1657, opened 2025-10-22
chore(ci): use locking tox-uv runner

SamWilsn:
#### Cute Animal Picture


SamWilsn:
@marioevz are my changes to your new tox environments appropriate?

packages/tests/pyproject.toml
```diff
@@ -36,7 +36,7 @@ dependencies = [
     "coincurve>=20.0.0,<21",
     "trie>=3.1.0,<4",
     "semver>=3.0.1,<4",
-    "pydantic>=2.11.0,<3",
+    "pydantic>=2.12.3,<3",
```
fselmo, line 39: What was the reasoning here? Do we need to narrow this dep range?

SamWilsn: The locked version wouldn't pass type checking

fselmo: ~~Interesting I didn't have this issue locally~~ [see next comment]. I think we can at least use `2.12` here since that's when Python 3.14 support came in ~~but even `uv run --python=3.14 tox -e static` didn't break for me here with `2.11.0`~~.

fselmo: Ah wait, no I see it. But yeah `2.12.0` fixes it. Does `2.12.0` lower pin work for you locally? 🤔

fselmo: From our offline convo, I think this is fine. Ship it :shipit:

fselmo (approved):
lgtm 👍🏼

PR #1666, opened 2025-10-23
refactor(tests): Use pytest collection to load JSON fixtures

marioevz:
## 🗒️ Description
This PR refactors the blockchain and state test infrastructure to leverage pytest's native collection mechanism via pytest_collect_file, eliminating redundant JSON file reads and improving test execution efficiency.

#### Key Improvements

1. Native pytest Collection
  - Implements pytest_collect_file hook to collect tests directly from JSON files during pytest's discovery phase
  - Each JSON file is now read exactly once during collection, rather than being read multiple times during parameterization and execution
  - Test fixtures are created as pytest Item objects (e.g., BlockchainTestFixture, StateTestFixture) that encapsulate all test data

2. Eliminated Redundant File I/O
  - Before: JSON files were read during test parameterization (fetch_blockchain_tests) and again during test execution (run_blockchain_st_test)
  - After: JSON files are read once in FixturesFile.collect(), and test data is stored in fixture objects for later execution
  - Removes intermediate dictionaries passing file paths that triggered repeated file reads

3. Cleaner Architecture
  - Introduces Fixture base class for shared fixture behavior
  - Test execution logic moved into runtest() methods of fixture classes
  - Test metadata (markers, fork info) configured during collection rather than parameterization
  - Eliminates the need for custom idfn functions - pytest handles naming automatically

#### Performance Impact

This refactoring significantly reduces I/O overhead for large test suites where the same JSON files contain multiple test cases across different forks.

#### Open Issues

Some failing tests still that need to be investigated, for now I'd like to start running this in CI and see how it improves execution speed.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    uvx --with=tox-uv tox -e static
    ```

#### Cute Animal Picture


tests/json_infra/conftest.py
```diff
@@ -260,6 +267,17 @@ def pytest_sessionstart(session: Session) -> None:
...
+            # importing them.
+            for python_file in glob(
+                os.path.join(fixture_path, "**/*.py"), recursive=True
+            ):
+                try:
+                    os.unlink(python_file)
+                except FileNotFoundError:
+                    # Not breaking error, another process deleted it first
+                    pass
+
```
SamWilsn, line 320: This feels... strange? I can't quite put my finger on why.

Like, why do the fixtures contain python files at all? Is there another way we could accomplish the same thing (like excluding a directory)?

I dunno, this just triggers my spidey sense :rofl:

tests/json_infra/helpers/__init__.py
```diff
@@ -1 +1,10 @@
 """Helpers to load tests from JSON files."""
+
+from .fixtures import ALL_FIXTURE_TYPES, Fixture, FixturesFile
+from .load_blockchain_tests import BlockchainTestFixture
+from .load_state_tests import StateTestFixture
+
+ALL_FIXTURE_TYPES.append(BlockchainTestFixture)
+ALL_FIXTURE_TYPES.append(StateTestFixture)
```
SamWilsn, line 8: Do these get executed when importing only, for example, `.load_state_tests`? From my limited knowledge of Python's import machinery, I would guess yes, but I'm just checking.

tests/json_infra/helpers/exceptional_test_patterns.py
```diff
@@ -20,6 +21,7 @@ class TestPatterns:
     big_memory: Tuple[Pattern[str], ...]
 
 
+@lru_cache
```
SamWilsn: How often is this called to require an `lru_cache`? O.o

Depending on when the cache is populated (in worker vs. in master), using `lru_cache` can explode memory: each worker has its own cache.

tests/json_infra/conftest.py
marioevz: This is the culprit: https://github.com/ethereum/legacytests/tree/1f581b8ccdc4c63acf5f2c5c1b155c690c32a8eb/src/LegacyTests/Cancun/GeneralStateTestsFiller/Pyspecs

Checking out `ethereum/tests` at this commit, when submodules are included, results in these python files being checked out too, and when collecting `./tests/json_infra/fixtures` for JSON files, pytest tries to collect these files too.

tests/json_infra/helpers/__init__.py
marioevz: Yes that's correct, it gets executed only when importing from `.helpers`. If we were to, for example, import directly from `.helpers.fixtures`, this logic would not be executed and `ALL_FIXTURE_TYPES` would be empty, so it is indeed a bit brittle if being honest.

tests/json_infra/helpers/exceptional_test_patterns.py
marioevz: I removed it thinking it might reduce the memory footprint and it did by half a GB, but it still consumes around 30GB+ because all fixtures are in memory when running.

tests/json_infra/conftest.py
SamWilsn: Don't we exclude that directory on the command line?

tests/json_infra/helpers/__init__.py
SamWilsn: Oh really? I thought parent modules were implicitly imported. I'm glad I checked!

tests/json_infra/helpers/exceptional_test_patterns.py
SamWilsn: image

tests/json_infra/conftest.py
marioevz: I removed that because with this approach the files are collected directly by pytest, as opposed to doing a glob in the test itself.

SamWilsn:
I was thinking briefly about this. I also know next to nothing about pytest, so this might not make any sense at all, but...

What if we use an LRU cache for the JSON files (one per worker), and [loadgroup](https://pytest-xdist.readthedocs.io/en/latest/distribution.html#distribution-modes) all the tests that come from the same file?

So you'd read once during collection, find all the tests and group them by file, then while running the tests you minimize the number of times you need to re-read the same file.

tests/json_infra/helpers/load_blockchain_tests.py
```diff
@@ -133,96 +58,150 @@ def add_block_to_chain(
...
+    def __init__(
+        self,
+        *args: Any,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize a single blockchain test fixture from a JSON file."""
+        super().__init__(*args, **kwargs)
+        self.fork_name = self.test_dict["network"]
+        self.add_marker(pytest.mark.fork(self.fork_name))
+        self.add_marker("evm_tools")
```
gurukamath: We should probably get rid of the `evm_tools` marker. This was used back when we needed to distinguish between our blockchain tests and state test runs. For context, we were using blockchain tests to test our specs directly and state tests to test the `t8n` implementation (hence the `evm_tools` name).

Also including the `evm_tools` marker here means that we mark all blockchain tests under json infra. This has the side effect of not running blockchain tests under the `optimized` environment. This explains why so few tests are run under `optimized`. See [this line](https://github.com/marioevz/execution-specs/blob/011051133789d4e4057503ce983e9cbbf11b1035/tox.ini#L100)

tests/json_infra/helpers/load_blockchain_tests.py
```diff
@@ -133,96 +58,150 @@ def add_block_to_chain(
...
+            else:
+                add_block_to_chain(chain, json_block, load, mock_pow)
+
+        last_block_hash = hex_to_bytes(json_data["lastblockhash"])
+        assert (
+            keccak256(rlp.encode(chain.blocks[-1].header)) == last_block_hash
+        )
+
+        expected_post_state = load.json_to_state(json_data["postState"])
+        assert chain.state == expected_post_state
```
gurukamath, line 196: I think this is currently not set up to catch any tests where the blocks themselves do not throw any exceptions but the overall state comparison fails . This I think is causing the current CI failure

gurukamath (commented):
Even though this is a much larger re-factor than #1730, I do like this approach since it uses more of the pytest native patterns. So a one-time larger change might be worth it.

tests/json_infra/helpers/load_blockchain_tests.py
gurukamath: Updated

gurukamath:
This PR is almost ready for review. However, I'm moving this to draft in order to resolve the discrepancy between the collectd vs run tests in CI for `json_infra`.

gurukamath:
@SamWilsn This is now ready for review. The failing tests `json_infra` tests are unrelated and should be fixed with #1813

.github/workflows/test.yaml
```diff
@@ -105,13 +105,35 @@ jobs:
...
+          # On push or force push to the feature branch
+            BASE_SHA="${{ github.event.before }}"
+            HEAD_SHA="${{ github.sha }}"
+          fi
+
+          echo "Diffing commits: $BASE_SHA..$HEAD_SHA"
+
+          # Get changed files and save to disk
+          FILE_LIST="changed_files.txt"
+          git diff --name-only "$BASE_SHA" "$HEAD_SHA" > "$FILE_LIST"
```
SamWilsn, line 130: Is `BASH_SHA` going to be the head of the base branch, or the merge-base of the two branches?

.github/workflows/test.yaml
```diff
@@ -105,13 +105,35 @@ jobs:
...
+          echo "Changed files saved to $FILE_LIST"
+          echo "file_list=$FILE_LIST" >> $GITHUB_OUTPUT
+          echo "List of files changed in the PR"
+          cat $FILE_LIST
       - name: Run json infra tests
-        run: tox -e json_infra
+        run: tox -e json_infra -- --file-list=${{ steps.get-changed-files.outputs.file_list }}
```
SamWilsn:
```suggestion
        run: tox -e json_infra -- --file-list="${{ steps.get-changed-files.outputs.file_list }}"
```

maybe?

tests/json_infra/conftest.py
```diff
@@ -66,10 +74,33 @@ def pytest_addoption(parser: Parser) -> None:
     )
 
     parser.addoption(
-        "--fork",
-        dest="fork",
+        "--from",
+        action="store",
+        dest="forks_from",
```
SamWilsn, line 78: Would it make sense to make a range class and store `fork`/`until`/`from` in the same `dest`?

tests/json_infra/conftest.py
```diff
@@ -66,10 +74,33 @@ def pytest_addoption(parser: Parser) -> None:
...
+        action="store",
+        dest="single_fork",
+        default="",
+        help="Only fill tests for the specified fork.",
+    )
+    parser.addoption(
+        "--file-list",
+        action="store",
+        dest="file_list",
+        help="Only fill tests for the specified fork.",
```
SamWilsn: Needs an update?

tests/json_infra/helpers/fixtures.py
```diff
@@ -0,0 +1,118 @@
...
+    @cached_property
+    def data(self) -> Dict[str, Any]:
+        """Return the JSON data of the full file."""
+        # loaded once per worker per file (thanks to cached_property)
+        with self.fspath.open("r", encoding="utf-8") as f:
+            return json.load(f)
+
+    def clear_data_cache(self) -> None:
+        """Drop the data cache."""
+        self.__dict__.pop("data", None)
```
SamWilsn: Would [this](https://stackoverflow.com/a/62662941) work/be more idiomatic?

tests/json_infra/helpers/select_tests.py
```diff
@@ -0,0 +1,88 @@
...
+    if not changed_files_file.exists():
+        print(f"File list file {files_path} does not exist or is empty!!")
+        return all_forks
+
+    with open(changed_files_file, "r") as f:
+        changed_files = [line.strip() for line in f if line.strip()]
+
+    # Extract affected forks
+    affected_forks = set()
+    repo_root = Path.cwd()  # Assuming running from repo root
```
SamWilsn: You can get the root from pytest.

SamWilsn (commented):
Partial review so far

.github/workflows/test.yaml
gurukamath: `BASE_SHA` in this case would be the head of the base branch

tests/json_infra/conftest.py
gurukamath: I'm ok with either approach on this. The current method is seems simple and straightforward to me. Whereas a single destination might make it easier to handle mutual exclusivity of the options. No strong leaning either way for me.

tests/json_infra/helpers/select_tests.py
gurukamath: Updated

tests/json_infra/helpers/fixtures.py
gurukamath: Updated

tests/json_infra/conftest.py
gurukamath: Ah. Yes. Updated

.github/workflows/test.yaml
gurukamath: Updated

.github/workflows/test.yaml
SamWilsn: Hm, so a file added in the base branch would get tested here, even if no changes were made to it in this pull request?

tests/json_infra/conftest.py
SamWilsn: We can keep it the way it is.

.github/workflows/test.yaml
gurukamath: The added file itself would not be tested but the addition might trigger a broader set of tests than what the PR explicitly changes. Perhaps this is not desirable and we should stick to comparing with the merge-base of the two branches. I'll give it a bit more thought

SamWilsn (approved):

PR #1670, opened 2025-10-23
fix(consume): consume sync validation relies on FCU only

fselmo:
## 🗒️ Description

RPC is not yet available for some clients during the sync verification. I tried using `eth_getBlockByNumber("safe")` instead of retrieving by `blockhash` but this is also returning `None` after the `VALID` FCU response is returned from the sync client. I don't think sync tests need to test JSON-RPC and I think FCU `VALID` is enough to tell us that the client has validated the payload.

I'm open to other suggestions here but this tests the acceptance of the sync block and with these changes all tests pass for the current sync tests (block rlp limit, EIP-7934) without flakiness. This resolves a very annoying flaky test issue due to some clients not making the block available immediately via JSON-RPC.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,12 +14,14 @@ Test fixtures for use by clients are available for each release on the [Github r
 
 #### `consume`
 
+- 🐞 Fix a bug with `consume sync` tests where some clients don't have JSON-RPC immediately available after syncing and can't yet serve the synced block ([#1670](https://github.com/ethereum/execution-specs/pull/1670)).
+
 ### 📋 Misc
 
 ### 🧪 Test Cases
 
-- 🐞Fix BALs opcode OOG test vectors by updating the Amsterdam commit hash in specs and validating appropriately on the testing side ([#2293](https://github.com/ethereum/execution-spec-tests/pull/2293)).
-- ✨Fix test vector for BALs SSTORE with OOG by pointing to updated specs; add new boundary conditions cases for SSTORE w/ OOG ([#2297](https://github.com/ethereum/execution-spec-tests/pull/2297)).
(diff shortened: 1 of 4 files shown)


packages/tests/src/pytest_plugins/consume/simulators/simulator_logic/test_via_sync.py
```diff
@@ -503,44 +503,77 @@ def test_blockchain_via_sync(
...
+                        f"{sync_block['stateRoot']}"
+                    )
+                    break
+                elif attempt < max_attempts - 1:
+                    # Blocks not available yet, wait and retry
+                    missing = []
+                    if sync_block is None:
+                        missing.append("sync_client")
+                    if client_block is None:
+                        missing.append("client_under_test")
```
spencer-tb: We could bring this up a level to reduce duplication.

spencer-tb (approved):
LGTM! One small comment. Thanks so much for fix this.

marioevz (approved):
LGTM.

One thing I just noticed is that we have a few hardcoded constants, like the amount of time to wait for the client to sync, the amount of retries, etc. I think we should create parameters for all of this to make these tests more customizable and allow users to reduce the flakiness of tests themselves.

Should be left for a follow up PR though.

fselmo:
> One thing I just noticed is that we have a few hardcoded constants, like the amount of time to wait for the client to sync, the amount of retries, etc. I think we should create parameters for all of this to make these tests more customizable and allow users to reduce the flakiness of tests themselves.

I like this idea. The hard-coded values are mostly semi-conservative and I don't believe any client really takes the full length of time here. I think that trying to retrieve the `eth_getBlockByHash` at the end, for clients that never serve this within the time, we just waste 5 seconds per test. Perhaps we can add this extra check as a parameter even, if we know the client serves JSON-RPC in time. I'm not sure. I think we could get rid of it altogether as well. Open to streamlining this more in a separate PR for sure. Let's focus on getting hive passing here.

PR #1688, opened 2025-10-27
chore(deps,tooling): move `dev` extras to `dependency-group`

danceratopz:
## tldr

1. Fix `testing` package by adding `ruff` as a dep (required by `gentest`).
2. Move developer dependencies from "extras" (optional dependencies) to PEP 753 "[dependency groups](https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups)".

## What changes?

Developers need to run:
```
uv sync
```
to get all dev dependencies (previously it was `uv sync --all-extras`).

## More Background

These changes follow a retroactive review of #1671, sorry, I was too slow @fselmo!

I think the standard way of adding `dev` dependencies is not as an "extra", but rather as a `dependency-group`:

> uv reads development dependencies from the [dependency-groups] table (as defined in [PEP 735](https://peps.python.org/pep-0735/)).
> 
> The dev group is special-cased and synced by default.
From https://docs.astral.sh/uv/concepts/projects/sync/#syncing-development-dependencies

So this PR makes that change and additionally removes `tox` as a direct `dev` dependency; imo, it should be ran as a "tool" via `uvx`; the doc changes already made in 
- https://github.com/ethereum/execution-specs/pull/1668

reflect that.

Before this PR:
```
uv sync --extras dev  # unnecessary
uv run tox -e static  # feels a bit weird
```

With this PR:
```
uv sync  # installs dev deps by default
uvx tox -e static  # treat tox as a tool
```

It's great that we get to drop the `--with=tox-uv` flag!

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```
- [ ] ~~All: Considered adding an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).~~ skipped

#### Cute Animal Picture

image


pyproject.toml
```diff
@@ -187,10 +187,10 @@ optimized = [
     "rust-pyspec-glue>=0.0.9,<0.1.0",
     "ethash>=1.1.0,<2",
 ]
+
+[dependency-groups]
 dev = [
-    "ethereum-execution[test,lint,tools,doc]",
-    "tox>=4.0",
-    "tox-uv>=1.29",
+  "ethereum-execution[test,lint,tools,doc]",
```
SamWilsn: Should we instead move the entire `lint` and `doc` extras here? There is no downstream reason to depend on them.

danceratopz: Yes, considering this too. Discussing via the server right now :)

fselmo: Yeah I think we can slim these extras down and include an all-encompassing `dev` dep group so that a `uv sync` can  set up the whole dev environment.

fselmo: `test` may not even make sense as an install extra if there are no unit tests for the specs. I think anyone wanting to test things will be a developer running `uv` commands. It would make sense if you wanted to `uv pip install . && pytest ...`... but given our flow, I think anyone testing would want to `uv sync && uv run {pytest,fill,consume,execute} ...`.

Really I think `optimized` might be the only one that makes sense as optional install extra? 🤔

danceratopz: Awesome, I've removed all extras (except "optimized") to `dependency-groups` for both the base package and `testing`.

Btw, this isn't `uv` specific, this is now part of the regular python packaging guidelines:
https://packaging.python.org/en/latest/specifications/dependency-groups

It's even better than I originally realized, we can still keep these groups organized by tool group ("lint", "test", etc.) and then use `include-group` in the "dev" dependency group.

The only extra I left is optimized, as I believe the intent is to optionally allow package users to also use `ethereum_optimized` package via `ethereum-execution[optimized]` - is that right @SamWilsn?

Also, is it ok to remove the `doc` extra and just have it as a dependency group?

tox.ini
```diff
@@ -1,19 +1,14 @@
...
 [testenv]
 runner = uv-venv-lock-runner
 uv_seed = false
 
 [testenv:static]
-extras =
-    lint
-    optimized
-    test
-    doc
```
danceratopz, line 16: @SamWilsn @fselmo I went for the simpler option here: Just let uv install **all** dev deps. It's nice to keep this simple and mirrors the `.venv` that developers will use locally. It seemed the most appropriate. As `uv` is so fast, speed is not a consideration.

The alternative would be to explicitly list the groups that uv should install. That would be:
```ini
dependency_groups =
    lint
    test
    doc
extras =
    optimized
```
In which case, `tox` runs:
```bash
uv sync --locked --python-preference system --extra optimized --no-default-groups --group doc --group lint --group test -p /home/dtopz/code/github/new/execution-specs/.tox/.tox/bin/python
```

It does mean though, that the "lint", "test" `dependency-groups` in `pyproject.toml` are just for aesthetics; they're not tested. If someone makes a mistake and puts a `doc` dep in `lint`, we'll never know. I can live with that and would favor keeping the groups, for a better overview.

tox.ini
```diff
@@ -26,34 +21,28 @@ commands =
...
-    lint
 commands =
     pytest \
         -n auto --maxprocesses 6 \
         --basetemp="{temp_dir}/pytest" \
         src
 
 [testenv:tests_pytest_pypy3]
+dependency_groups =
+    test
```
danceratopz, line 33: We do need to specify the `dependency-groups` explicitly here (or any tox envs that use `pypy`), otherwise installing the `optimized` extra breaks the build:
```
       Built ethereum-execution-testing @ file:///opt/actions-runner/_work/execution-specs/execution-specs/packages/testing
  × Failed to build `ethash==1.1.0`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta.build_wheel` failed (exit status: 1)
```
From https://github.com/ethereum/execution-specs/actions/runs/18871754191/job/53851753115#step:5:44.

danceratopz: This additionally has the consequence that devs that use `pypy` can't run `uv sync`, they will have to run 
```
uv sync --no-default-groups --group test
```
to avoid pulling in all dev deps, specifically, the `optimized` extra.

danceratopz:
@fselmo @SamWilsn let me know if you like it, I can make a follow-up to update the docs.

tox.ini
```diff
@@ -26,34 +21,28 @@ commands =
...
 passenv =
     PYPY_GC_MAX
     PYPY_GC_MIN
 # run inside the package directory so pytest picks up packages/testing/pyproject.toml
 changedir = {toxinidir}/packages/testing
 package = editable
-extras =
-    test
-    lint
```
danceratopz, line 47: Removing the `lint` extras caused the `gentest` unit tests to fail as `gentest` requires `ruff` (it formats the generated files for convenience).

packages/testing/pyproject.toml
```diff
@@ -51,6 +51,7 @@ dependencies = [
     "joblib>=1.4.2",
     "ckzg>=2.1.3,<3",
     "tenacity>=9.0.0,<10",
+    "ruff==0.13.2",
```
danceratopz: This is a bit crazy, but currently `gentest` runs `ruff`. It didn't really matter when we all just worked in dev mode with `uv sync --all-extras`, but this is now required to fix our package.

Perhaps there should be a `gentest` extra which only adds this dependency?

tox.ini
danceratopz: Finding this bug through tighter (and only listing the expected) dependency groups now makes me lean to keeping it more explicit everywhere. But this `ruff` dep issue is somewhat of an edge case.

packages/testing/pyproject.toml
danceratopz: Here's where we call `ruff` btw:
https://github.com/danceratopz/execution-specs/blob/a4c85ce62218f569cc7a8c5b0235c74c85f7f609/packages/testing/src/execution_testing/cli/gentest/source_code_generator.py#L78-L85

danceratopz (commented):
I fixed up a packaging bug in `testing` that cropped up due to these changes; it was missing `ruff` as a dependency.

pyproject.toml
SamWilsn: > The only extra I left is optimized, as I believe the intent is to optionally allow package users to also use `ethereum_optimized` package via `ethereum-execution[optimized]` - is that right @SamWilsn?

Correct. `optimized` should actually be an extra.



> Also, is it ok to remove the `doc` extra and just have it as a dependency group?

Yep! It's also a dev tool.

packages/testing/pyproject.toml
SamWilsn: > Perhaps there should be a `gentest` extra which only adds this dependency?

Yeah, that makes sense. Could even make the formatting step optional, and only run if `ruff` is available?

pyproject.toml
```diff
@@ -156,6 +156,12 @@ packages = [
 "ethereum_spec_tools" = ["py.typed"]
 
 [project.optional-dependencies]
+optimized = [
+    "rust-pyspec-glue>=0.0.9,<0.1.0",
+    "ethash>=1.1.0,<2",
+]
+
+[dependency-groups]
```
SamWilsn, line 166: I'm not sure if tools should be a dependency group. `ethereum_spec_tools` is installed as part of `ethereum-execution`, so I think these are real dependencies.

pyproject.toml
```diff
@@ -183,14 +189,12 @@ doc = [
...
-]
 dev = [
-    "ethereum-execution[test,lint,tools,doc]",
-    "tox>=4.0",
-    "tox-uv>=1.29",
+    {include-group = "test"},
+    {include-group = "lint"},
+    {include-group = "tools"},
+    {include-group = "doc"},
+    "ethereum-execution[optimized]",
```
SamWilsn, line 194: I don't think `optimized` is a mandatory requirement for `dev`.

tox.ini
SamWilsn: Alternatively, I think we could `apt` install cmake plus a compiler and ethash would compile itself.

pyproject.toml
danceratopz: I'd agree, but we need to include it to run mypy cleanly on all sources including `src/ethereum_optimized/`.

```
static: commands[3]> mypy
src/ethereum_optimized/state_db.py:22: error: Cannot find implementation or library stub for module named "rust_pyspec_glue"  [import-not-found]
src/ethereum_optimized/fork.py:26: error: Cannot find implementation or library stub for module named "ethash"  [import-not-found]
```

pyproject.toml
danceratopz: Good catch, I deleted the `tools` dependency group and moved them up to the package dependencies.

tox.ini
```diff
@@ -26,34 +21,28 @@ commands =
...
 commands =
     pytest \
         -n auto --maxprocesses 6 \
         --basetemp="{temp_dir}/pytest" \
         src
 
 [testenv:json_infra]
-extras =
```
fselmo, line 55: Dumb question, why does this env not need `dependency_groups = test`?

fselmo: Ah nevermind, I see why the other one _needs_ it now. The rest is pulled from `uv.lock` by the runner as expected 👍🏼

packages/testing/pyproject.toml
fselmo: I agree, I think the optional formatting if available makes sense. In most cases it will be available because anyone developing will have `dev` deps installed. If it isn't, you have to lint your code before merging anyway so would just be an extra step.

tox.ini
danceratopz: :+1: added to the doc issue:
- https://github.com/ethereum/execution-specs/issues/1713

tox.ini
danceratopz: Definitely not a dumb question! You're absolutely right, but it's not at all obvious that pypy3 envs should not install `optimized`. I added a comment to make this more obvious!

danceratopz: Also added a follow-up issue which will address the pypy and optimized issue:
- #1713

packages/testing/pyproject.toml
danceratopz: Good points! Reverted the commit and created a follow-up issue to handle it correctly for the case when the command is ran in a package:
- https://github.com/ethereum/execution-specs/issues/1715

fselmo (approved):
lgtm once CI passes 👍🏼

danceratopz:
> lgtm once CI passes 👍🏼

Thanks, now we're failing testing unit tests ([here](https://github.com/ethereum/execution-specs/actions/runs/18921094965/job/54017267486?pr=1688)) due to the issue that @felix314159 is addressing here 
- #1711

fselmo:
Yep sounds good on CI fix in separate PR if you're good with it 👍🏼

tox.ini
danceratopz: `uvx` allows ephemeral venvs and should allow running a tool on a package without polluting the package. If it worked, it seems like the cleanest way to go. Had a play, documented it here:
- https://github.com/ethereum/execution-specs/issues/1716

PR #1692, opened 2025-10-27
feat(ci): add hive consume test workflow

spencer-tb:
## Description
Adds a CI workflow to validate hive integration with consume commands. Previously, we lacked CI for consume, which has led to breaking changes going undetected until they impacted downstream EL clients.

This workflow provides early detection of consume-related regressions, preventing disruption to client teams who depend on our test fixtures.

## What's Tested

The workflow runs four parallel test jobs:
- `ethereum/eels/consume-engine`
- `ethereum/eels/consume-rlp`
- `ethereum/eels/consume-sync`
- `ethereum/eels/consume-engine` Dev mode validation

All tests use `go-ethereum` as the reference client and filter to a subset of Osaka fork tests to keep CI runtime reasonable while still providing meaningful coverage. We pull the latest `go-ethereum` image from `ethpandaops`.

We run a single test just to check for no fails. It assumes `go-ethereum` will pass this test always. It takes no longer that 5 minutes to run the workflow.

We may consider expanding the filter for when we run this workflow, to include any changes to `packages/testing`.

### DIFF
--- a/.github/workflows/hive-consume.yaml
+++ b/.github/workflows/hive-consume.yaml
@@ -40,26 +40,26 @@ jobs:
     steps:
       - name: Checkout execution-specs
         if: matrix.mode == 'dev'
-        uses: actions/checkout@v5
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           path: execution-specs
 
       - name: Checkout Hive
-        uses: actions/checkout@v5
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
           repository: ethereum/hive
           ref: master
           path: hive
 
       - name: Setup go env and cache
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
           go-version: '>=1.24'
           cache-dependency-path: hive/go.sum
 
       - name: Setup Python
         if: matrix.mode == 'dev'
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
           python-version: '3.12'
 
--- a/.github/workflows/hive-consume.yaml
+++ b/.github/workflows/hive-consume.yaml
@@ -7,9 +7,12 @@ on:
   pull_request:
     paths:
       - '.github/workflows/hive-consume.yaml'
-      - 'packages/tests/src/pytest_plugins/consume/**'
-      - 'packages/tests/src/pytest_plugins/pytest_hive/**'
+      - 'packages/testing/src/execution_testing/cli/pytest_commands/consume.py'
+      - 'packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-consume.ini'
       - 'packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/**'
+      - 'packages/testing/src/execution_testing/cli/pytest_commands/plugins/pytest_hive/**'
+      - 'packages/testing/src/execution_testing/fixtures/consume.py'
+      - 'packages/testing/src/execution_testing/rpc/**'
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/hive-consume.yaml
```diff
@@ -0,0 +1,124 @@
...
+  push:
+    branches:
+      - 'forks/**'
+  pull_request:
+    # TODO: After initial testing, uncomment these paths to only run on relevant changes:
+    # paths:
+    #   - '.github/workflows/hive-consume.yaml'
+    #   - 'packages/tests/src/pytest_plugins/consume/**'
+    #   - 'packages/tests/src/pytest_plugins/pytest_hive/**'
+    #   - 'packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/**'
```
spencer-tb: Will uncomment once we re-trigger in EELS and see that it works.

.github/workflows/hive-consume.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - name: Run simulator tests
+        if: matrix.mode == 'simulator'
+        run: |
+          cd hive
+          ./hive --sim '${{ matrix.simulator }}' \
+            --sim.parallelism=1 \
+            --client go-ethereum \
+            --client-file clients.yaml \
+            --sim.buildarg fixtures=develop@latest \
+            --sim.limit=".*test_block_at_rlp_limit_with_logs.*Osaka.*" \
```
spencer-tb, line 98: I used a single test here. The same for all. Just so we can check it passes.

spencer-tb:
Passing them here: https://github.com/ethereum/execution-specs/actions/runs/18876034305/job/53865955924?pr=1692

I added the filtering back now so this is ready for review!

.github/workflows/hive-consume.yaml
```diff
@@ -0,0 +1,123 @@
...
+on:
+  push:
+    branches:
+      - 'forks/**'
+  pull_request:
+    paths:
+      - '.github/workflows/hive-consume.yaml'
+      - 'packages/tests/src/pytest_plugins/consume/**'
+      - 'packages/tests/src/pytest_plugins/pytest_hive/**'
+      - 'packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/**'
```
danceratopz: These are out of date. I think we should add:
``` 
      - 'packages/testing/src/execution_testing/fixtures/consume.py'
      - `packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-consume.ini`
```

danceratopz (commented):
Looks like we need a rebase - I will do this and address some issues directly :)

.github/workflows/hive-consume.yaml
danceratopz: Done.

.github/workflows/hive-consume.yaml
```diff
@@ -0,0 +1,123 @@
...
+          done
+
+      - name: Run consume in dev mode
+        if: matrix.mode == 'dev'
+        working-directory: execution-specs
+        env:
+          HIVE_SIMULATOR: http://127.0.0.1:3000
+        run: |
+          uv sync --all-extras
+          uv run consume ${{ matrix.consume_command }} --input develop@latest -k "Osaka and test_block_at_rlp_limit_with_logs"
```
danceratopz: I would pin the release to reduce flakiness. Also, if the test id changes, this could fail silently (no tests found). 
```suggestion
          uv run consume ${{ matrix.consume_command }} --input develop@v5.3.0 -k "Osaka and test_block_at_rlp_limit_with_logs"
```

.github/workflows/hive-consume.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+      - name: Run simulator tests
+        if: matrix.mode == 'simulator'
+        run: |
+          cd hive
+          ./hive --sim '${{ matrix.simulator }}' \
+            --sim.parallelism=1 \
+            --client go-ethereum \
+            --client-file clients.yaml \
+            --sim.buildarg fixtures=develop@latest \
```
danceratopz: @copilot 
```suggestion
            --sim.buildarg fixtures=develop@v5.3.0 \
```

danceratopz (approved):
@spencer-tb this is awesome, we've been missing this for a while!  :sweat_smile: 
I rebased so I could fix-up the `paths`, added a couple more paths to trigger the workflow and updated to hte `astral-sh/update-uv` action.

One other suggestion below.

.github/workflows/hive-consume.yaml
spencer-tb: Few nice catch. Thx

PR #1707, opened 2025-10-28
fix(testing): format `eest make test` output with project Ruff config

### LINKED ISSUE #1675: Ensure `uv run eest make test` calls `ruff` and formats the generated test module correctly
`uv run eest make test` runs correctly, but the generated test module is not correctly formatted.

From the copilot review:

The code now references `AppConfig().ROOT_DIR.parent` instead of `AppConfig().ROOT_DIR` to find pyproject.toml. This change suggests the ROOT_DIR definition may have changed during the refactor. Verify that ROOT_DIR still points to the correct location (packages/testing/src/execution_testing) and document why .parent is now needed.
```suggestion
        # Call ruff to format the file
        # AppConfig().ROOT_DIR points to 'packages/testing/src/execution_testing'.
        # The project's pyproject.toml is located in the parent directory: 'packages/testing/src'.
        # Therefore, we use .parent to correctly reference the config file.
```

_Originally posted by @Copilot in https://github.com/ethereum/execution-specs/pull/1654#discussion_r2459545228_

bodier123:
## Description
`eest make test` now formats the generated file with Ruff using the **project** config:

--config (AppConfig().ROOT_DIR.parent.parent / "pyproject.toml")

After #1654, `ROOT_DIR` points to `packages/testing/src/execution_testing`, while the project `pyproject.toml` lives at `packages/testing/`. Passing `--config` ensures the generated file matches repo style. The formatter runs right after the file is written.

## Changes
- Add `_ruff_cmd()`, `_project_pyproject()`, `_ruff_format_file(path)`.
- Call `_ruff_format_file(module_path)` after writing the file.
- Guard: only pass `--config` if the file exists (installed envs won’t error).
- Test: `test_make_test_formatting.py` asserts the Ruff CLI args.

## Checks
- [x] `uvx tox -e static` (ruff, ruff format --check, mypy, ethereum-spec-lint)
- [x] Unit test passes

Closes #1675.

### DIFF
--- a/packages/testing/src/execution_testing/cli/eest/make/commands/test.py
+++ b/packages/testing/src/execution_testing/cli/eest/make/commands/test.py
@@ -7,14 +7,16 @@ generated test file is saved in the appropriate directory with a rendered
 template using Jinja2.
 """
 
-import os
 import sys
 from pathlib import Path
+import subprocess
+import shutil
 
 import click
 import jinja2
 
 from execution_testing.config.docs import DocsConfig
+from execution_testing.config import AppConfig
 from execution_testing.forks import get_development_forks, get_forks
 
 from ....input import input_select, input_text
@@ -26,6 +28,26 @@ template_env = jinja2.Environment(
     trim_blocks=True,
     lstrip_blocks=True,
 )
+
+
+def _ruff_cmd() -> list[str]:
+    r = shutil.which("ruff")
+    return [r] if r else [sys.executable, "-m", "ruff"]
+
+
+def _project_pyproject() -> Path:
+    # ROOT_DIR -> packages/testing/src/execution_testing
+    # pyproject  -> packages/testing/pyproject.toml
+    return (AppConfig().ROOT_DIR.parent.parent / "pyproject.toml").resolve()
+
+
+def _ruff_format_file(target: Path) -> None:
+    cfg = _project_pyproject()
+    cmd = _ruff_cmd() + ["format"]
+    if cfg.exists():
+        cmd += ["--config", str(cfg)]
+    cmd += [str(target)]
+    subprocess.run(cmd, check=True)
 
 
 def exit_now() -> None:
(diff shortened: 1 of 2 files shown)


danceratopz:
Thanks @bodier123. I made a mistake in the title and description of the issue #1675 - sorry about that. The original issue should have referred to the `gentest` command, not the `eest make test` command. Most `make test` formatting issues should be resolved by the jinja2 templates, although some user input will break formatting (module title), so this PR still makes sense, imo, especially as we already have `ruff` as a depedency.

However, I would ask you to reuse and improve the function that calls `ruff` from `gentest`:
https://github.com/danceratopz/execution-specs/blob/15b5cfdfecdb66b525ca78379e9ad61bf54365bf/packages/testing/src/execution_testing/cli/gentest/source_code_generator.py#L54-L105

I would suggest that we move the `format_code` function to a shared module `packages/testing/src/execution_testing/cli/shared/format_code.py`. Then use it for both `gentest` and `make test`.

About the required improvements. Previously, `gentest` and `make test` were only intended to be ran within a clone in developer mode. If we want to make sure that they run correctly when executed from `execution_testing` as an installed package, then we need to remove the assumption that a `pyproject.toml` exists in one of the parent directories and we will need add the `ruff` required configuration directly to the `format_code` module. To do: Is it enough to specify the target line-length of 79? Or is there other ruff config we need to add? I.e., does the behavior of `ruff format --check path/to/generated/source.py` with minimal config match that with the pyproject.toml config?

Sorry for the confusion. I hope this helps clear things up. Lmk if you'd still like to work on this, if not I can help finish it off!

danceratopz:
Hey @bodier123. As you might have noticed, a lot this repo is in flux at the moment due to [The Weld](https://steel.ethereum.foundation/blog/blog_posts/2025-09-11_weld-announcement/). Another PR triggered this issue, which directly affects this PR. Please take it into consideration, if you work further on this :pray: 
- https://github.com/ethereum/execution-specs/issues/1715

PR #1708, opened 2025-10-28
fix(testing/specs): Fix benchmark pre-alloc grouping

marioevz:
## 🗒️ Description

This PR addresses issues when generating pre-allocation groups with the `BenchmarkTest` spec type.

There were two main issues that had to be addressed for this:

### `BenchmarkTest.get_genesis_environment` not using `BlockchainTest.get_genesis_environment`

`BenchmarkTest.get_genesis_environment` was directly generating the genesis from `env` without adding the extra required argument that `BlockchainTest.get_genesis_environment` uses.

One potential proper (follow-up) fix could be that `BenchmarkTest` inherits `BlockchainTest`.

### `BenchmarkTest.generate` modifying `pre`

`BenchmarkTest` modified an invariant that the pre-alloc grouping relied on: `pre` is not touched after the test function completes.

This resulted in, correctly, test filling failing on the test running step because the base `Alloc` does not implement `deploy_contract` because there should not be any more deployments.

But the problem was that `BenchmarkTest` required the `fork` instance in order to determine how many blocks to use (because of the transaction gas limit cap), and this value was only passed when the `generate` or the `execution` functions are called.

Solution was to include `fork` in the `BaseTest` class and make it mandatory for the filler, execute, or any other code that instantiated a test spec, to pass the fork during the initialization:
https://github.com/marioevz/execution-specs/blob/fc2bd475b8e7caf506b39ec6453e9d049e8f418e/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py#L1353
Example in an unit test:
https://github.com/marioevz/execution-specs/blob/fc2bd475b8e7caf506b39ec6453e9d049e8f418e/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_prealloc_group.py#L56

This resulted in `fork` being removed from `generate`, `execute` and other instance methods within `BaseTest`, which is a nice bonus.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/eofwrap.py
+++ b/packages/testing/src/execution_testing/cli/eofwrap.py
@@ -323,6 +323,7 @@ class EofWrapper:
 
         test = BlockchainTest(
             genesis_environment=env,
+            fork=EOFv1,
             pre=pre.root,
             post=fixture.post_state.root if fixture.post_state else {},
             blocks=[],
@@ -373,7 +374,6 @@ class EofWrapper:
 
         result = test.generate(
             t8n=t8n,
-            fork=EOFv1,
             fixture_format=BlockchainFixture,
         )
         assert isinstance(result, BlockchainFixture)
--- a/packages/testing/src/execution_testing/cli/fuzzer_bridge/blocktest_builder.py
+++ b/packages/testing/src/execution_testing/cli/fuzzer_bridge/blocktest_builder.py
@@ -68,7 +68,6 @@ class BlocktestBuilder:
         # Generate fixture
         fixture = test.generate(
             t8n=self.t8n,
-            fork=fork,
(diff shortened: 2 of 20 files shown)


fselmo (approved):
lgtm! 🚀 

> One potential proper (follow-up) fix could be that BenchmarkTest inherits BlockchainTest.

I'd be curious to explore this and see if it makes sense but I think the current flow makes more sense with separation of concerns. It would sort of violate Liskov principle since `BenchmarkTest` seems more like a blockchain test factory - `blocks` is required in `BlockchainTest` but can be generated and isn't required for `BenchmarkTest`, etc.

I think I prefer to leave them separate tbh 🤔.

PR #1722, opened 2025-10-31
fix(specs): Ensure tracking before first access

nerolation:
### What was wrong?

Related to Discussion [here](https://discord.com/channels/595666850260713488/1432349632616337520/1433523424181092504).

Opcodes were tracking addresses after dynamic state checks instead of after static gas costs, causing addresses to be omitted from the Block Access List when operations failed due to OOG.

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -110,6 +110,9 @@ def generic_create(
 
     evm.accessed_addresses.add(contract_address)
 
+    # Track address access for BAL
+    track_address_access(state.change_tracker, contract_address)
+
     if account_has_code_or_nonce(
         state, contract_address
     ) or account_has_storage(state, contract_address):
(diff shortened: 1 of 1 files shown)


fselmo (approved):
lgtm 👌🏼

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -110,6 +110,9 @@ def generic_create(
 
     evm.accessed_addresses.add(contract_address)
 
+    # Track address access for BAL
+    track_address_access(state.change_tracker, contract_address)
```
SamWilsn, line 114: Random comment, not a serious review, but if this is an operation on the "state", then should `track_address_access` only take the state itself as the first argument? Alternatively, if this is an operation on the "state tracker", then does it make sense for the tracker to be a property of `State` (or should it be a property of, for example, `BlockEnvironment`)?

nerolation: Let me evaluate this but now that you say it, it feels you're right on architectural inconsistencies.

PR #1728, opened 2025-11-02
feat(tests): Opcode tests: stack overflow & BLOCKHASH

pdobacz:
## 🗒️ Description

Two additions:

1. It seems there weren't tests covering stack overflow for some opcodes
2. Other one is an addition to the existing BLOCKHASH test, which didn't cover the situation where the instruction attempts to read beyond the 256 recent blocks. Catch here is that it comes up as MC/DC coverage gap even after it is filled. But I validated manually modifying the code - removing the decision's condition does indeed cause the new tests to fail. Not sure why would the report got this wrong.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/frontier/opcodes/test_all_opcodes.py
+++ b/tests/frontier/opcodes/test_all_opcodes.py
@@ -3,7 +3,7 @@ Call every possible opcode and test that the subcall is successful if the
 opcode is supported by the fork supports and fails otherwise.
 """
 
-from typing import Dict
+from typing import Dict, Iterator
 
 import pytest
 from execution_testing import (
@@ -19,6 +19,7 @@ from execution_testing import (
     Transaction,
     UndefinedOpcodes,
 )
+from execution_testing.forks import Byzantium
 
 REFERENCE_SPEC_GIT_PATH = "N/A"
 REFERENCE_SPEC_VERSION = "N/A"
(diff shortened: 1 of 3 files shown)


tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -135,3 +137,64 @@ def test_cover_revert(state_test: StateTestFiller, pre: Alloc) -> None:
...
+
+def fork_opcodes_increasing_stack(
+    fork: Fork,
+) -> Iterator[Op]:
+    """
+    Yields opcodes which are valid for `fork` and increase the operand stack.
+    """
+    for opcode in fork.valid_opcodes():
+        if opcode.pushed_stack_items > opcode.popped_stack_items:
+            yield opcode
```
LouisTsai-Csie, line 149: I like this approach!

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -135,3 +137,64 @@ def test_cover_revert(state_test: StateTestFiller, pre: Alloc) -> None:
...
+
+
+@pytest.mark.valid_from("Frontier")
```
LouisTsai-Csie: I think we do not need this here?
```suggestion
```

tests/frontier/opcodes/test_all_opcodes.py
```diff
@@ -135,3 +137,64 @@ def test_cover_revert(state_test: StateTestFiller, pre: Alloc) -> None:
...
+    state_test: StateTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    opcode: Op,
+    fails: bool,
+    env: Environment,
+) -> None:
+    """Test that opcodes which leave new items on the stack can overflow."""
+    if cap := fork.transaction_gas_limit_cap():
+        env.gas_limit = ZeroPaddedHexNumber(cap)
```
LouisTsai-Csie: Why do we need this? I removed it and tried locally, but it seems still working.

tests/frontier/opcodes/test_blockhash.py
```diff
@@ -10,11 +10,19 @@
...
     Transaction,
 )
+from execution_testing.forks import Byzantium
+from execution_testing.forks.helpers import Fork
 
 
 @pytest.mark.valid_from("Frontier")
+@pytest.mark.parametrize(
+    "first_block", ["none", "empty", "with_tx", "over_256"]
+)
```
LouisTsai-Csie, line 25: Could we parametrize `(setup_blocks_num, empty_block)` instead?

```python
@pytest.mark.parametrize(
    "setup_blocks_num,empty_block",
    [
        pytest.param(0, True, id="no_blocks"),
        pytest.param(1, False, id="one_empty_block"),
        pytest.param(1, True, id="one_block_with_tx"),
        pytest.param(256, True, id="256_empty_blocks"),
    ],
)
def test_genesis_hash_available(...):
    ...
    blocks = [
        Block(
            txs=[
                Transaction(
                    sender=sender,
                    to=contract,
                    protected=fork >= Byzantium,
                )
            ]
        )
        if not empty_block
        else Block()
        for _ in range(setup_blocks_num)
    ]
```
It would be more flexible if we want to add more cases!

tests/frontier/opcodes/test_blockhash.py
```diff
@@ -29,45 +37,58 @@ def test_genesis_hash_available(
     Bug context: revm blockchaintest runner wasn't inserting block_hashes,
     causing failures in tests with BLOCKHASH-derived addresses.
     """
-    storage = Storage()
+    Storage()
```
LouisTsai-Csie:
```suggestion
```

LouisTsai-Csie (changes requested):
Thanks for the PR, leave some suggestion!

tests/frontier/opcodes/test_all_opcodes.py
pdobacz: Indeed we don't. I've noticed earlier it was necessary, not sure, maybe my mistake.

LouisTsai-Csie (approved):
LGTM! Thanks for contribution

PR #1730, opened 2025-11-03
fix(tests): optimize json_infra fixture reads

gurukamath:
### What was wrong?
Running the `json_infra` does a lot of I?O that can potentially be avoided.

Related to Issue #

### How was it fixed?
Optimise file reads from disk using some caching techniques and queueing the test fixtures in a better order.

### DIFF
--- a/tests/json_infra/helpers/__init__.py
+++ b/tests/json_infra/helpers/__init__.py
@@ -1,1 +1,30 @@
 """Helpers to load tests from JSON files."""
+
+import json
+from functools import lru_cache
+from typing import Dict
+
+
+@lru_cache(maxsize=100)
+def load_json_file(test_file: str) -> Dict:
+    """
+    Load and cache a JSON fixture file.
+
+    Uses LRU (Least Recently Used) cache to avoid re-reading the same
+    JSON files multiple times during test execution. This is especially
+    important when running tests in parallel, as multiple test cases
+    from the same file would otherwise cause redundant file I/O.
+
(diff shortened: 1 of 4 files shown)


SamWilsn (approved):
@marioevz used `--dist=loadfile` in #1666, would that make sense here?

gurukamath:
> @marioevz used `--dist=loadfile` in #1666, would that make sense here?

I have implemented `--dist=loadgroup` by defining all the tests in a particular json fixture file as part of the same xdist group. That allows each file to be read exactly once and makes the LRU caching technique maximally useful. Just waiting to see what the CI runtime from this is

gurukamath:
> @marioevz used `--dist=loadfile` in #1666, would that make sense here?

Next, I will try out some of the ideas from that PR and see what works best

SamWilsn:
> I have implemented `--dist=loadgroup` by defining all the tests in a particular json fixture file as part of the same xdist group.

I saw! Was wondering if `loadfile` would save you having to manually create those groups.

gurukamath:
> I saw! Was wondering if `loadfile` would save you having to manually create those groups.

Ah. I see. From what I understand `--dist=loadfile` groups the tests by the python test file (.py file) and has no way of distinguishing the different json fixtures (unless we explicitly tell it). This is why I decided to explicitly define the groups and run the `--dist=loadgroup` which should achieve the intended effect.

gurukamath:
Closing this in favor of #1666

PR #1741, opened 2025-11-03
fix(spec-tests): Fix new fork tool test to use last fork dynamically

fselmo:
## 🗒️ Description

When we have two `Unscheduled` forks, which will be more frequent moving forward, while Osaka is `Unscheduled` and Amsterdam is too, we cannot resolve the fork criteria for the e2e test for the fork tool (e.g. [here](https://github.com/ethereum/execution-specs/actions/runs/18994902166/job/54253329761?pr=1727#step:5:4900)).

- We can check in `builder.py` if the fork is `Unscheduled` as well as if it's `forks[-1]` and raise the `order_index` to come after it as we already do.
- We can always use the last fork dynamically in this e2e test so that we always resolve and don't hard-code Osaka into it.

I'm not tied to the last change. If we want to make hard-coding Osaka work, we can change some other things around as well but this change does get this test passing on Amsterdam as it uses `Amsterdam` as the last fork.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels) --- question below

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/new_fork/builder.py
+++ b/src/ethereum_spec_tools/new_fork/builder.py
@@ -330,7 +330,7 @@ class ForkBuilder:
             self.output = Path(template_path).parent
 
         # Try to make a sane guess for the activation criteria.
-        if found is forks[-1]:
+        if found is forks[-1] or isinstance(found.criteria, Unscheduled):
             order_index = 0
 
             # check template fork's criteria and bump order_index if needed
--- a/tests/json_infra/test_tools_new_fork.py
+++ b/tests/json_infra/test_tools_new_fork.py
@@ -5,11 +5,21 @@ Tests for the ethereum-spec-new-fork CLI tool.
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
+import pytest
+
 from ethereum_spec_tools.forks import Hardfork
 from ethereum_spec_tools.new_fork.cli import main as new_fork
 
 
-def test_end_to_end() -> None:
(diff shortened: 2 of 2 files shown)


fselmo:
Are we perhaps missing a label for `spec-tests` or something that signifies `json_infra/` tests?

SamWilsn (commented):
Might be worth having a `test_end_to_end_latest` and keeping `test_end_to_end` as is. That way we can tell if the new fork tool breaks for old forks.

fselmo:
> Are we perhaps missing a label for spec-tests or something that signifies json_infra/ tests?

~~I added `A-spec-tests`. Feel free to change the description or remove it but I felt a good label for this was missing.~~ [Hmm... I think `C-test` then `A-spec-tool` works here?]

fselmo:
> Might be worth having a `test_end_to_end_latest` and keeping `test_end_to_end` as is. That way we can tell if the new fork tool breaks for old forks.

@SamWilsn I split these to have both a hard-coded Osaka and the latest fork. I didn't see a strong reason to change the parameters for these but let me know if you think we should.

fselmo:
Just sanity checked this by cherry-picking these commits over in Amsterdam bals branch and they pass 👌🏼

spencer-tb (approved):
LGTM! @SamWilsn for merge?

SamWilsn (approved):

Issue #1604, opened 2025-10-15
Port benchmark test from eip-7904 research


LouisTsai-Csie:
# Summary

EIP-7904 is proposed in `Glamsterdam` for gas repricing, which will update the current gas model in a significant way. The current proposed gas constant is based on this research: gas-cost-estimator. Since the end goal is to maintain a single source of test cases in EEST framework, this issue will explore how their benchmark works and how should it be ported.

# Background

Thanks to the contributions from zkEVM and Nethermind team, we now have a comprehensive compute-intensive benchmark test suite. The EEST infrastructure generates fixtures for the transaction setup, which are then provided to clients for execution.

The current benchmark tests are designed to:
1. Maximize the cycle count within the block gas limit.
2. Represent the worst-case scenario for specific opcodes, helping evaluate whether the current gas limits or defined gas costs could pose risks to network stability or zkEVM execution.
3. Execute the transaction and measure the total execution time of the payload.

However, this gas-repricing research takes a different approach, instead of exhausting the available gas in a block, it focuses on measuring the actual computational effort required for each operation.

# Analysis
According to the Stage 2 report, there are three measurement methods: **measure_total**, **measure_marginal**, and **measure_argument**. Note that these measurements target the EVM interpreter execution (e.g., [Geth](https://github.com/ethereum/go-ethereum/blob/7c107c2691fa66a1da60e2b95f5946c3a3921b00/core/vm/interpreter.go#L97)), not the transaction payload execution.

- **measure_total**: Measures the total execution time of the entire program.
- **measure_marginal**: Estimates the marginal gas cost by increasing the opcode count:
**measure_total_time**  = b + a * op_count,
where a is the marginal cost and b is the intercept.
- **measure_argument**: Evaluates how argument size or type affects opcode performance:
measure_total_time = b + a * op_count * weight(arg) + c * arg + d * op_count

# Comparison
To fairly compare different opcodes within a specific client implementation, we need to eliminate the impact from parameter interactions (e.g. `PUSH`)

For example, when comparing `ADDMOD` and `ADD`, we must ensure that stack setup operations do not introduce noise. Since `ADDMOD` requires 3 params while `ADD` only requires 2, we should prepend additional `PUSH` operations so that both sequences contain the same number of stack setup instructions. This ensures the measured time difference reflects the true execution cost of the target opcode.

| Instruction | Execution Count |
|--------------|------------------|
| `PUSH1` | `max_op_count` times |
[...]

PR #1747, opened 2025-11-04
feat(benchmark): support `--fixed-opcode-count` flag and tests

LouisTsai-Csie:
### Fixed Opcode Count Benchmark
This update introduces a fixed opcode count benchmark scenario.
A new flag, `--fixed-opcode-count`, and a new test marker, `gas_ref`, have been added. Only tests marked with `gas_ref` support the fixed opcode count feature.

Example command:
```
fill -v tests/benchmark/compute/instruction/<test> \
    --fixed-opcode-count 20 \
    --clean -m benchmark
```

**Technical Notes**
1. Regular and gas repricing reference tests should run under the normal benchmark command.
2. When specifying `--fixed-opcode-count`, only gas repricing reference tests will be executed.
3. Currently, this feature is supported only for benchmark tests written using the benchmark test wrapper and code generator.
4. The current benchmark process is not fully optimized, but it can be refactored later.
5. When running `--fixed-opcode-count` command, the gas limit would be configured as 1000M gas limit by default. Manually configure the `--gas-benchmark-values` will trigger an error.

**Benchmark Pattern**
1. Extract the parameters (`setup`, `attack_block`) from the benchmark test wrapper and generate a contract that iterates the attack_block 1000 times.
2. Generate another contract that calls the first contract `fixed-opcode-count` times.
3. This ensures the total opcode execution count equals 1000 × fixed-opcode-count.

**Example**
Setting `--fixed-opcode-count 200` means executing the opcode 200 × 1000 = 200,000 times in total.
The first contract runs 1000 iterations per call, while the second contract repeats those calls 200 times.

## 🔗 Related Issues or PRs

issue #1604

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -26,7 +26,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         dest="fixed_opcode_count",
         type=str,
         default=None,
-        help="Specify fixed opcode counts for benchmark tests as a comma-separated list.",
+        help="Specify fixed opcode counts (in thousands) for benchmark tests as a comma-separated list.",
     )
 
 
@@ -44,30 +44,35 @@ def pytest_configure(config: pytest.Config) -> None:
 def pytest_collection_modifyitems(
     config: pytest.Config, items: list[pytest.Item]
 ) -> None:
-    """Skip non-gas_ref tests when --fixed-opcode-count is specified."""
+    """Remove non-gas_ref tests when --fixed-opcode-count is specified."""
     fixed_opcode_count = config.getoption("fixed_opcode_count")
     if not fixed_opcode_count:
         # If --fixed-opcode-count is not specified, don't filter anything
         return
 
-    # Filter: keep only tests with gas_ref marker
-    for item in items:
-        has_gas_ref = item.get_closest_marker("gas_ref") is not None
-        has_benchmark = item.get_closest_marker("benchmark") is not None
-
-        # Skip benchmark tests that don't have gas_ref marker
-        if has_benchmark and not has_gas_ref:
-            item.add_marker(
-                pytest.mark.skip(
-                    reason="Test does not have gas_ref marker (required for --fixed-opcode-count)"
-                )
-            )
+    items[:] = [
+        item
+        for item in items
+        if not (
+            item.get_closest_marker("benchmark") is not None
+            and item.get_closest_marker("gas_ref") is None
+        )
+    ]
 
 
 def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     """Generate tests for the gas benchmark values and fixed opcode counts."""
+    gas_benchmark_values = metafunc.config.getoption("gas_benchmark_value")
+    fixed_opcode_counts = metafunc.config.getoption("fixed_opcode_count")
+
+    # Ensure mutual exclusivity
+    if gas_benchmark_values and fixed_opcode_counts:
+        raise pytest.UsageError(
+            "--gas-benchmark-values and --fixed-opcode-count are mutually exclusive. "
+            "Use only one at a time."
+        )
+
     if "gas_benchmark_value" in metafunc.fixturenames:
-        gas_benchmark_values = metafunc.config.getoption("gas_benchmark_value")
         if gas_benchmark_values:
             gas_values = [
                 int(x.strip()) for x in gas_benchmark_values.split(",")
@@ -89,9 +94,6 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
             metafunc.definition.get_closest_marker("gas_ref") is not None
         )
         if has_gas_ref:
-            fixed_opcode_counts = metafunc.config.getoption(
-                "fixed_opcode_count"
-            )
             if fixed_opcode_counts:
                 opcode_counts = [
                     int(x.strip()) for x in fixed_opcode_counts.split(",")
@@ -99,7 +101,7 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
                 opcode_count_parameters = [
                     pytest.param(
                         opcode_count,
-                        id=f"opcount_{opcode_count}",
+                        id=f"opcount_{opcode_count}K",
                     )
                     for opcode_count in opcode_counts
                 ]
@@ -116,6 +118,10 @@ def gas_benchmark_value(request: pytest.FixtureRequest) -> int:
     if hasattr(request, "param"):
         return request.param
 
+    # If --fixed-opcode-count is specified, use high gas limit to avoid gas constraints
+    if request.config.getoption("fixed_opcode_count"):
+        return HIGH_GAS_LIMIT
+
     return EnvironmentDefaults.gas_limit
 
 
@@ -129,6 +135,7 @@ def fixed_opcode_count(request: pytest.FixtureRequest) -> int | None:
 
 
 BENCHMARKING_MAX_GAS = 1_000_000_000_000
+HIGH_GAS_LIMIT = 1_000_000_000
 
 
 @pytest.fixture
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -35,7 +35,7 @@ def pytest_configure(config: pytest.Config) -> None:
     """Configure the fill and execute mode to benchmarking."""
     config.addinivalue_line(
         "markers",
-        "gas_ref: Mark test as a gas reference test for gas repricing analysis",
+        "repricing: Mark test as reference test for gas repricing analysis",
     )
     if config.getoption("gas_benchmark_value"):
         config.op_mode = OpMode.BENCHMARKING  # type: ignore[attr-defined]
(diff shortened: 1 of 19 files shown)


jsign:
If the `--opcode-count` isn't provided, it would do the same behavior as today of filling the block with the asked gas limits? Or those tests will always need an opcode-count flag?

LouisTsai-Csie:
@jsign Yes it would by default use the same behavior as it is now. But if the flag being specified, it would switch to the new fixed opcode count scenario. Do you think this is straightforward, or how could we improve the workflow here? Thanks

jsign:
> @jsign Yes it would by default use the same behavior as it is now. But if the flag being specified, it would switch to the new fixed opcode count scenario. Do you think this is straightforward, or how could we improve the workflow here? Thanks

I think we'll be interested in both styles, one for worst-case block gas limit and the other for regression-like analysis as planned. 

I think this optional flag and defaulting to worst-case-gas-limit is quite good, so sgtm!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -20,17 +20,52 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+
+    # Filter: keep only tests with gas_ref marker
+    for item in items:
+        has_gas_ref = item.get_closest_marker("gas_ref") is not None
+        has_benchmark = item.get_closest_marker("benchmark") is not None
+
+        # Skip benchmark tests that don't have gas_ref marker
+        if has_benchmark and not has_gas_ref:
+            item.add_marker(
+                pytest.mark.skip(
```
marioevz: I suggest we rather remove the tests from the list instead of skipping them, because you can clearly get the count of tests with `--collect-only` that way (Example: https://github.com/ethereum/execution-specs/blob/10784130e7094b76b081a74a9ffdf3c13e4721a4/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py#L1511).

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -48,6 +83,32 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
                 "gas_benchmark_value", gas_parameters, scope="function"
             )
 
+    if "fixed_opcode_count" in metafunc.fixturenames:
```
marioevz, line 107: My only concern here is that `--gas-benchmark-values` can be used with this new parameter, however I don't understand the interplay between these two. I.e. can you still guarantee that a certain amount of gas is used if you constrain the amount of times an opcode is executed? I feel these two are mutually exclusive, so IMO we should only parametrize by one of them.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -20,17 +20,52 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default=None,
         help="Specify gas benchmark values for tests as a comma-separated list.",
     )
+    evm_group.addoption(
+        "--fixed-opcode-count",
+        action="store",
+        dest="fixed_opcode_count",
+        type=str,
+        default=None,
+        help="Specify fixed opcode counts for benchmark tests as a comma-separated list.",
```
marioevz: IIUC, this value is times one thousand, so we should specify that here.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -30,6 +30,7 @@
 )
 
 
+@pytest.mark.gas_ref
```
marioevz: What's the origin of the "gas_ref" name for this marker?

Also, I think we should try to see if we can automatically detect the characteristic that makes a test candidate for this marker, rather than having to manually set it, because we will eventually forget to set it for new tests.

marioevz (commented):
In general looks good to me, but I left a couple of comments I feel we should address.

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: This comes from our discussion during the gas lighting call. We need to designate the `baseline` benchmark test as the reference test, but this should be configured manually since we only want a single combination. I’ll look into how to implement this properly (and use a better naming for this marker)

Also, the current name isn’t straightward, using a `baseline` marker might probably be more clear.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -20,17 +20,52 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+    )
 
 
 @pytest.hookimpl(tryfirst=True)
 def pytest_configure(config: pytest.Config) -> None:
     """Configure the fill and execute mode to benchmarking."""
+    config.addinivalue_line(
+        "markers",
+        "gas_ref: Mark test as a gas reference test for gas repricing analysis",
```
spencer-tb:
```suggestion
        "repricing": Mark test as a gas reference test for gas repricing analysis",
```

tests/benchmark/compute/instruction/test_account_query.py
spencer-tb: Baseline is good. I suggest repricing above.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
spencer-tb: Fully agree here. We should only allow one flag. If both are supplied, fill should error out before filling.

Given the above I think we should set the `--gas-benchmark-value` default really high (maybe 1000), when `--fixed-opcode-count` is supplied. This allows us to set `--fixed-opcode-count` to very large values that ignore the block gas limit.

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -290,6 +357,22 @@ def generate_blocks_from_code_generator(self) -> List[Block]:
...
+    def generate_fixed_opcode_count_transactions(self) -> List[Block]:
+        """Generate transactions with a fixed opcode count."""
+        if self.code_generator is None:
+            raise Exception("Code generator is not set")
+        self.code_generator.deploy_fix_count_contracts(
+            pre=self.pre, fork=self.fork
+        )
+        gas_limit = (
+            self.fork.transaction_gas_limit_cap() or self.gas_benchmark_value
+        )
```
spencer-tb, line 373: We should add a TODO here for Osaka support or consideration. If the number of fixed opcodes need more than 16M gas then we will fail. It would be nice to fail early here if we reach the tx gas limit cap.

I'm not sure if splitting into multiple txs makes much sense here for the gas repricing regression use case. Adding multiple txs would skew the timings I feel. Maybe they could be split into 2 categories. Single tx and multi tx. It would be good to know if multi txs should be supported here for forks >= Osaka.

The other consideration would be if you wanted to split them into multi txs equally... or have 1/many full and with the final tx partially full.

We could potentially find a way to disable the cap for testing in clients, but this would add a security concern.

spencer-tb (commented):
Thanks. Added some comments.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
spencer-tb: Just another naming idea.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
LouisTsai-Csie: I think giga gas is working well here, included in the new commits!

packages/testing/src/execution_testing/specs/benchmark.py
LouisTsai-Csie: AFAIK, we’re planning to disable the transaction gas limit cap for the repricing experiment. Maria hope to exclude the cap during timing measurements, and Nethermind and Geth will add support for this in internal testing.

LouisTsai-Csie:
Thanks for the review @marioevz , @spencer-tb . I update according to the comment, and i am ready for second review

I receive some feedback from Kamil. For gas repricing, we only want to run a specific test (only one parameter combination is enough). So i add extract filter logic in the `repricing` marker, such that you could select the benchmark parameter comnbination.

Example:
```python
@pytest.mark.repricing(
    size=1024 * 1024,
    non_zero_data=True,
    zeros_topic=False,
    fixed_offset=True,
)
@pytest.mark.parametrize(
    "size,non_zero_data",
)
@pytest.mark.parametrize(
    "zeros_topic",
)
@pytest.mark.parametrize("fixed_offset")
def test_log(...)
...
```

This is flexible, for normal scenario we could simply label the marker without configuring parameter:

```python
@pytest.mark.repricing
def test_codesize(
) -> None:
    """Benchmark CODESIZE instruction."""
    benchmark_test(
        code_generator=ExtCallGenerator(attack_block=Op.CODESIZE),
    )
```

jochem-brouwer:
I will check the existing benchmark tests to see if we can convert the existing tests to use this "opcode count" metric in it. This would avoid adding this CLI option and would instead add this option to the tests for maintainability of the framework (vs. maintainability of the tests).

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -101,9 +146,18 @@ def generate_repeated_code(
         available_space = max_code_size - overhead
         max_iterations = available_space // len(repeated_code)
 
+        # Use fixed_opcode_count if provided, otherwise fill to max
+        if self.fixed_opcode_count is not None:
+            max_iterations = min(max_iterations, 1000)
+
+        print(f"max_iterations: {max_iterations}")
```
marioevz, line 154: This print could be a bit annoying, we should remove IMO.

marioevz (approved):
LGTM, thanks for the changes!

Branch needs a rebase and then we can merge.

LouisTsai-Csie:
Hi @jochem-brouwer , currently i reuse the `code_generator` feature in the benchmark test wrapper. Since this type of test structure as:

- setup: initial stack element
- attack_block: repeat certain opcode sequence as benchmark target. This is hardcoded as 1000 now.
- cleanup: We skip this now since we do not need to clean up the stack as there is no more iteration in the current call frame. 

A low hanging fruit would be convert the existing tests in this format. But also we could find other ways for this feature.

spencer-tb:
I don't want to block this as LGTM from myside.
One comment would be to add some simple CI for the `--fixed-opcode-count` flag?
Inspired by: #1779

PR #1751, opened 2025-11-04
fix(spec-specs): split `charge_gas` around account access

fselmo:
## 🗒️ Description

Related to #1722, we should charge for account access costs before actually marking the account accessed. If there is not enough gas for the access, we should never expend the resources to access the account.

Test cases for all of these boundaries should be added as well.

These changes should likely be introduced at some point into all forks if we decide this is the approach that we should take.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2025-11-04 at 13 57 21

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -407,16 +407,28 @@ def call(evm: Evm) -> None:
     ) = access_delegation(evm, code_address)
     access_gas_cost += delegated_access_gas_cost
 
+    # Charge delegation access cost before tracking delegation target
+    if delegated_access_gas_cost > Uint(0):
+        charge_gas(evm, delegated_access_gas_cost)
+
+    # Track delegation target if delegation occurred
+    if disable_precompiles:
+        track_address_access(
+            evm.message.block_env.state.change_tracker, code_address
+        )
+
     create_gas_cost = GAS_NEW_ACCOUNT
     if value == 0 or is_account_alive(evm.message.block_env.state, to):
         create_gas_cost = Uint(0)
     transfer_gas_cost = Uint(0) if value == 0 else GAS_CALL_VALUE
+
+    # Charge remaining costs (create, transfer, subcall)
     message_call_gas = calculate_message_call_gas(
         value,
         gas,
         Uint(evm.gas_left),
-        extend_memory.cost,
-        access_gas_cost + create_gas_cost + transfer_gas_cost,
+        Uint(0),  # Memory cost already charged
+        create_gas_cost + transfer_gas_cost,
     )
     charge_gas(evm, message_call_gas.cost + extend_memory.cost)
     if evm.message.is_static and value != U256(0):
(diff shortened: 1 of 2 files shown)


src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -654,12 +691,22 @@ def delegatecall(evm: Evm) -> None:
...
+    # Track delegation target if delegation occurred
+    if disable_precompiles:
+        track_address_access(
+            evm.message.block_env.state.change_tracker, code_address
+        )
+
+    # Charge remaining costs (subcall)
     message_call_gas = calculate_message_call_gas(
-        U256(0), gas, Uint(evm.gas_left), extend_memory.cost, access_gas_cost
+        U256(0), gas, Uint(evm.gas_left), Uint(0), Uint(0)
```
fselmo: I think this kind of thing is only relevant now with BAL or state tracking in general but this should likely be reflected back to all forks right? There are things in BAL implementation that may impact the logic across all forks.

I think this plays into the state tracker conversation where this would have to be refactored for all forks. And it sort of renders memory and access costs in `calculate_message_call_gas` unnecessary (I believe).

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -498,17 +517,28 @@ def callcode(evm: Evm) -> None:
...
     transfer_gas_cost = Uint(0) if value == 0 else GAS_CALL_VALUE
+
+    # Charge remaining costs (transfer, subcall)
     message_call_gas = calculate_message_call_gas(
         value,
         gas,
         Uint(evm.gas_left),
-        extend_memory.cost,
-        access_gas_cost + transfer_gas_cost,
+        Uint(0),  # Memory cost already charged
```
fselmo, line 538: Note, should this be removed from `calculate_message_call_gas` if we always charge upfront for memory?

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -727,16 +778,26 @@ def staticcall(evm: Evm) -> None:
...
+        )
+
+    # Charge remaining costs (subcall)
     message_call_gas = calculate_message_call_gas(
         U256(0),
         gas,
         Uint(evm.gas_left),
-        extend_memory.cost,
-        access_gas_cost,
+        Uint(0),  # Memory cost already charged
```
fselmo, line 797: See prev comment in all these cases

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -392,10 +392,14 @@ def call(evm: Evm) -> None:
     if to in evm.accessed_addresses:
         access_gas_cost = GAS_WARM_ACCESS
     else:
-        evm.accessed_addresses.add(to)
         access_gas_cost = GAS_COLD_ACCOUNT_ACCESS
 
-    # Track address access for BAL
+    # Charge static costs (memory + access) before BAL tracking
+    charge_gas(evm, extend_memory.cost + access_gas_cost)
```
SamWilsn, line 398: This will output a double trace event for this opcode.

fselmo: Sorry Sam, I think we have a better implementation closer to what we talked about here now: https://github.com/ethereum/execution-specs/pull/1759

fselmo: We can focus on that PR and I will close this one

fselmo:
closed in favor of #1759

PR #1768, opened 2025-11-07
tests(benchmark): add missing scenario and optimization

LouisTsai-Csie:
## 🗒️ Description

Enhance the code generator for `BenchmarkTest` wrapper:
- Add initial balance setting for `JumpLoopGenerator`, which already supports in the other code generator, this could simplify `test_create`.
- Add safety checks for `ExtCallGenerator`, when calculating the max iterations, it should deduct the stack delta that comes from `setup` section, or it might lead to stack overflow issue.
- Most of the contract will be filled with repeated pattern, this PR further fills the remaining space with `STOP` opcode. This could help simplify some cases like `test_codecopy` & `test_return_revert`.
- For `ExtCallGenerator`, there are two different account: target contract and loop contract. The loop contract will repeatedly calls into target contract via `STATICCALL`. For some operations that involves `CALLDATA` interactions, we should forward the `CALLDATA` from loop contract to target contract via (1) forward the `CALLDATA` from loop contract and (2) Configure the memory from CALLDATA in target contract.

Test Case Enhancement:
- test_ext_account_query_warm: Add initial balance / initial storage parametrization.
- test_blockhash: Add extra cases for fixed block index and dynamic block index
- test_pc_op: Add missing benchmark test
- test_keccak: Add this case to parametrize memory size / offset for gas repricing effort.
- test_tload: Optimize the benchmark by switching to `ExtCallGenerator`
- test_block_full_of_ether_transfers: Add scenario that the receiver is not an empty account.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/specs/benchmark.py
+++ b/packages/testing/src/execution_testing/specs/benchmark.py
@@ -104,6 +104,8 @@ class BenchmarkCodeGenerator(ABC):
         # TODO: Unify the PUSH0 and PUSH1 usage.
         code = setup + Op.JUMPDEST + repeated_code * max_iterations + cleanup
         code += Op.JUMP(len(setup)) if len(setup) > 0 else Op.PUSH0 + Op.JUMP
+        # Pad the code to the maximum code size.
+        code += Op.STOP * (max_code_size - len(code))
         self._validate_code_size(code, fork)
 
         return code
--- a/packages/testing/src/execution_testing/specs/benchmark.py
+++ b/packages/testing/src/execution_testing/specs/benchmark.py
@@ -53,6 +53,7 @@ class BenchmarkCodeGenerator(ABC):
     setup: Bytecode = field(default_factory=Bytecode)
     cleanup: Bytecode = field(default_factory=Bytecode)
     tx_kwargs: Dict[str, Any] = field(default_factory=dict)
+    code_padding_opcode: Op | None = None
     _contract_address: Address | None = None
 
     @abstractmethod
@@ -105,7 +106,8 @@ class BenchmarkCodeGenerator(ABC):
         code = setup + Op.JUMPDEST + repeated_code * max_iterations + cleanup
         code += Op.JUMP(len(setup)) if len(setup) > 0 else Op.PUSH0 + Op.JUMP
         # Pad the code to the maximum code size.
-        code += Op.STOP * (max_code_size - len(code))
+        if self.code_padding_opcode is not None:
+            code += self.code_padding_opcode * (max_code_size - len(code))
         self._validate_code_size(code, fork)
 
         return code
--- a/tests/benchmark/compute/instruction/test_account_query.py
+++ b/tests/benchmark/compute/instruction/test_account_query.py
@@ -381,7 +381,14 @@ def test_ext_account_query_warm(
     for an account.
     """
     # Setup
-    target_addr = pre.empty_account() if initial_balance else pre.fund_eoa()
+    target_addr = Address()
+    if not initial_balance and not initial_storage:
+        target_addr = pre.empty_account()
+    elif initial_balance or initial_storage:
+        target_addr = pre.fund_eoa(
+            storage={0: 0x1337} if initial_storage else {0: 0}
+        )
+
     post = {}
     if not empty_account:
         code = Op.STOP + Op.JUMPDEST * 100
--- a/tests/benchmark/compute/instruction/test_account_query.py
+++ b/tests/benchmark/compute/instruction/test_account_query.py
@@ -4,6 +4,7 @@ current executing account or on a target account.
 """
 
 import math
+from typing import Dict
 
 import pytest
 from execution_testing import (
@@ -24,6 +25,7 @@ from execution_testing import (
     While,
     compute_create2_address,
 )
+from execution_testing.base_types.conversions import NumberConvertible
 
 from tests.benchmark.compute.helpers import (
     XOR_TABLE,
@@ -72,7 +74,6 @@ def test_codesize(
 )
 def test_codecopy(
     benchmark_test: BenchmarkTestFiller,
-    pre: Alloc,
     fork: Fork,
     max_code_size_ratio: float,
     fixed_src_dst: bool,
@@ -86,25 +87,11 @@ def test_codecopy(
     src_dst = 0 if fixed_src_dst else Op.MOD(Op.GAS, 7)
     attack_block = Op.CODECOPY(src_dst, src_dst, Op.DUP1)  # DUP1 copies size.
 
-    code = JumpLoopGenerator(
-        setup=setup, attack_block=attack_block
-    ).generate_repeated_code(
-        repeated_code=attack_block, setup=setup, fork=fork
+    benchmark_test(
+        code_generator=JumpLoopGenerator(
+            setup=setup, attack_block=attack_block
+        )
     )
-
-    # Pad the generated code to ensure the contract size matches the maximum
-    # The content of the padding bytes is arbitrary.
-    code += Op.INVALID * (max_code_size - len(code))
-    assert len(code) == max_code_size, (
-        f"Code size {len(code)} is not equal to max code size {max_code_size}."
-    )
-
-    tx = Transaction(
-        to=pre.deploy_contract(code=code),
-        sender=pre.fund_eoa(),
-    )
-
-    benchmark_test(tx=tx)
 
 
 @pytest.mark.parametrize(
@@ -361,7 +348,21 @@ def test_extcodecopy_warm(
     ],
 )
 @pytest.mark.parametrize(
-    "absent_target",
+    "empty_account",
+    [
+        True,
+        False,
+    ],
+)
+@pytest.mark.parametrize(
+    "initial_balance",
+    [
+        True,
+        False,
+    ],
+)
+@pytest.mark.parametrize(
+    "initial_storage",
     [
         True,
         False,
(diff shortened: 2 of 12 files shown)


tests/benchmark/compute/instruction/test_control_flow.py
```diff
@@ -22,6 +22,15 @@ def test_gas_op(
     )
 
 
+def test_pc_op(
```
LouisTsai-Csie, line 25: We miss the benchmark for `PC` opcode

tests/benchmark/compute/instruction/test_keccak.py
```diff
@@ -60,7 +61,24 @@ def test_keccak(
...
-            setup=Op.PUSH20[optimal_input_length],
+            setup=Op.PUSH20(optimal_input_length),
             attack_block=Op.POP(Op.SHA3(Op.PUSH0, Op.DUP1)),
         ),
     )
+
+
+@pytest.mark.parametrize("mem_alloc", [b"", b"ff", b"ff" * 32])
+@pytest.mark.parametrize("offset", [0, 31, 1024])
+def test_keccak(
```
LouisTsai-Csie, line 72: Instead of finding the optimal length, add a new case that parametrized initial memory layout and access offset. Request by gas repricing effort.

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -126,13 +128,12 @@ def ether_transfer_case(
         "diff_acc_to_diff_acc",
     ],
 )
+@pytest.mark.parametrize("balance", [0, 1])
 def test_block_full_of_ether_transfers(
```
LouisTsai-Csie, line 133: Add a new scenario for ether transfer benchmark: whether the receiver is an empty account.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -33,36 +34,25 @@ def test_tload(
...
-        attack_block = Op.POP(Op.TLOAD(Op.CALLVALUE))
+        attack_block = Op.TLOAD(Op.CALLVALUE)
 
-    cleanup = code_key_mut + code_val_mut
-    tx_value = start_key if not key_mut and val_mut else 0
+    tx_value = 42 if not key_mut and val_mut else 0
 
     benchmark_test(
-        code_generator=JumpLoopGenerator(
+        code_generator=ExtCallGenerator(
```
LouisTsai-Csie, line 49: By updating fro `JumpLoopGenerator` to `ExtCallGenerator`, we could get rid of the `POP` opcode each operation, enhancing the `TLOAD` density.

tests/benchmark/compute/instruction/test_block_context.py
```diff
@@ -36,14 +36,29 @@ def test_block_context_ops(
...
+)
 def test_blockhash(
     benchmark_test: BenchmarkTestFiller,
+    index: int | None,
 ) -> None:
     """Benchmark BLOCKHASH instruction accessing oldest allowed block."""
     # Create 256 dummy blocks to fill the blockhash window.
     blocks = [Block()] * 256
 
+    block_number = Op.AND(Op.GAS, 0xFF) if index is None else index
```
LouisTsai-Csie, line 57: Adding extra BLOCKHASH operation for benchmarking, as it was one of the slowest operations.
Cases:
- Valid block index
- Invalid block index
- Dynamic block index

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -371,27 +372,40 @@ def test_ext_account_query_warm(
     benchmark_test: BenchmarkTestFiller,
     pre: Alloc,
     opcode: Op,
-    absent_target: bool,
+    empty_account: bool,
+    initial_balance: bool,
+    initial_storage: bool,
```
LouisTsai-Csie: Add new test scenario for gas repricing effort:
- Accessing empty / non-empty account
- Accessing account contains zero / non-zero balance
- Accessing account contains zero / non-zero storage

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -104,6 +104,8 @@ def generate_repeated_code(
         # TODO: Unify the PUSH0 and PUSH1 usage.
         code = setup + Op.JUMPDEST + repeated_code * max_iterations + cleanup
         code += Op.JUMP(len(setup)) if len(setup) > 0 else Op.PUSH0 + Op.JUMP
+        # Pad the code to the maximum code size.
+        code += Op.STOP * (max_code_size - len(code))
```
marioevz: We should make this optional via a new flag in the `BenchmarkCodeGenerator` class:
```
code_padding_opcode: Op | None = None
```
So only tests that need to pad the code to the max size do this.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -86,26 +87,12 @@ def test_codecopy(
...
-    assert len(code) == max_code_size, (
-        f"Code size {len(code)} is not equal to max code size {max_code_size}."
-    )
-
-    tx = Transaction(
-        to=pre.deploy_contract(code=code),
-        sender=pre.fund_eoa(),
+    benchmark_test(
+        code_generator=JumpLoopGenerator(
+            setup=setup, attack_block=attack_block
```
marioevz:
```suggestion
            setup=setup, attack_block=attack_block, code_padding_opcode=Op.STOP
```
Following previous suggestion.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -361,7 +348,21 @@ def test_extcodecopy_warm(
     ],
 )
 @pytest.mark.parametrize(
-    "absent_target",
+    "empty_account",
```
marioevz: The meaning of "empty_account" seems confusing IMO. From what I'm reading below, it seems like it's rather "empty_code"?

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -371,27 +372,47 @@ def test_ext_account_query_warm(
...
     """
     # Setup
-    target_addr = pre.empty_account()
+    target_addr = Address()
+    if not initial_balance and not initial_storage:
+        target_addr = pre.empty_account()
+    elif initial_balance or initial_storage:
+        target_addr = pre.fund_eoa(
+            storage={0: 0x1337} if initial_storage else {0: 0}
+        )
```
marioevz: This unconditionally sets the balance to a non-zero value, even when `initial_balance==False`.

Also, `{0: 0}`, given the current `fund_eoa` logic, is still going to try to touch the storage (arguably a bug but besides the point).

We should instead build a `kwargs` and then call `pre.fund_eoa(**kwargs)` to make this cleaner.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -371,27 +372,47 @@ def test_ext_account_query_warm(
...
     """
     Test running a block with as many stateful opcodes doing warm access
     for an account.
     """
     # Setup
-    target_addr = pre.empty_account()
+    target_addr = Address()
+    if not initial_balance and not initial_storage:
+        target_addr = pre.empty_account()
+    elif initial_balance or initial_storage:
```
marioevz: I think we should refactor this logic a bit:
```suggestion
    if not initial_balance and not initial_storage and empty_code:
        target_addr = pre.empty_account()
    else:
```
And in the `else` branch we construct the `kwargs` to either call `pre.fund_eoa` or `pre.deploy_contract`.

tests/benchmark/compute/instruction/test_block_context.py
```diff
@@ -36,14 +36,29 @@ def test_block_context_ops(
...
 
 
+@pytest.mark.parametrize(
+    "index",
+    [
+        0,
+        1,
+        256,
+        257,
+        None,
```
marioevz:
```suggestion
        pytest.param(None, id="random"),
```
Or similar to show a better test id.

tests/benchmark/compute/instruction/test_keccak.py
```diff
@@ -60,7 +61,24 @@ def test_keccak(
 
     benchmark_test(
         code_generator=JumpLoopGenerator(
-            setup=Op.PUSH20[optimal_input_length],
+            setup=Op.PUSH20(optimal_input_length),
```
marioevz:
```suggestion
            setup=Op.PUSH20[optimal_input_length],
```
Original should be ok.

tests/benchmark/compute/instruction/test_memory.py
```diff
@@ -18,14 +18,10 @@ def test_msize(
...
-    """
-    Benchmark MSIZE instruction.
-
-    - mem_size: by how much the memory is expanded.
-    """
+    """Benchmark MSIZE instruction."""
     benchmark_test(
         code_generator=ExtCallGenerator(
-            setup=Op.MLOAD(Op.SELFBALANCE) + Op.POP,
+            setup=Op.POP(Op.MLOAD(Op.SELFBALANCE)),
```
marioevz, line 24: Nice, I'm guessing this helps with the stack push/pop elements calculations.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -33,36 +34,25 @@ def test_tload(
...
-    start_key = 41
-    code_key_mut = Bytecode()
-    code_val_mut = Bytecode()
     setup = Bytecode()
     if key_mut and val_mut:
-        setup = Op.PUSH1(start_key)
-        attack_block = Op.POP(Op.TLOAD(Op.DUP1))
-        code_key_mut = Op.POP + Op.GAS
-        code_val_mut = Op.TSTORE(Op.DUP2, Op.GAS)
+        setup = Op.GAS + Op.TSTORE(Op.DUP2, Op.GAS)
```
marioevz, line 36: While trying to read this code, I realize that `key_mut` -> `fixed_key`, `val_mut` -> `fixed_value`, and flipping the logic, would make it easier to read IMO.

tests/benchmark/compute/instruction/test_storage.py
marioevz: This might break the cases using `CALLVALUE` because we are now running this in the context of the subcall, and since we are calling using `STATICCALL` to call that subcontract, `CALLVALUE` is always going to be zero.
Perhaps replacing this with `CALLDATASIZE`, since we are passing along the data from the transaction?

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -462,16 +451,10 @@ def test_return_revert(
     mem_preparation = (
         Op.CODECOPY(size=return_size) if return_non_zero_data else Bytecode()
     )
-    executable_code = mem_preparation + opcode(size=return_size)
-    code = executable_code
-    if return_non_zero_data:
```
marioevz, line 475: This is lost because we are now by default returning code full of `Op.STOP`, which is all zeros. This can also be addressed by using the `code_padding_opcode` I mentioned in a previous comment (doing `code_padding_opcode=Op.INVALID` instead).

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -72,17 +74,17 @@ def get_single_sender_list(pre: Alloc) -> Generator[Address, None, None]:
...
+    pre: Alloc, balance: int
+) -> Generator[Address, None, None]:
     """Get a list of single receiver accounts."""
-    receiver = pre.fund_eoa(0)
+    receiver = pre.fund_eoa(balance)
     while True:
         yield receiver
 
 
-@pytest.fixture
```
marioevz, line 82: This could still work as a fixture if we parametrize `balance=0` on the rest of the test cases.

marioevz (commented):
Thanks!

In general, I liked the refactoring and changes, I've just left comments on multiple files just to make everything more polished.

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: Nice catch, i do not notice `STOP` is `0x00`. This is helpful.

LouisTsai-Csie:
Sorry for adding complexity to this PR, I’ve added one more commit that optimizes the `DUP` and `CALLDATASIZE` operations.
- `test_dup`: The setup stack delta is now handled in `ExtCallGenerator`, which simplifies the test logic.
- `test_calldatasize`: This test is also simplified using `ExtCallGenerator`, as the generator now forwards calldata to the target contract.

marioevz (approved):
Excellent job on all these optimizations. LGTM, thanks!

Issue #1755, opened 2025-11-05
Add extra storage read & write benchmark cases


LouisTsai-Csie:
**SSTORE -> https://github.com/ethereum/execution-specs/pull/1774** @LouisTsai-Csie 
- [x] Add a variant in the existing test that warms all storage slots via the Access List (AL).
- [x] In the [existing benchmark PR](https://github.com/ethereum/execution-specs/pull/1734/files), include different contract sizes: just created, small, medium, and XEN-sized.
- [x] Perform SLOAD on 1–200 storage positions before writing to each of them, covering all possible transition variants.
- [x] Deploy multiple contracts (1–5 parametrized instances) and perform cold storage writes on each.

**SLOAD -> https://github.com/ethereum/execution-specs/pull/1769** @jochem-brouwer 
- [ ] Repeated SLOAD on the same slot range (1–200) for four contract sizes (just created, small, medium, XEN), both with and without Access List, and with existing vs non-existing values.
- [ ] SLOAD on non-existing keys (1–200) for the same four sizes, tested with and without Access List.
- [ ] SLOAD on existing keys (1–200) for the same four sizes, tested with and without Access List.

jochem-brouwer:
I'm adding the SLOAD cases here. (Will link PR)

I think there is a feature being added to `execution-specs` which is a `num_opcodes` CLI argument - right? We should add this functionality directly to these tests, and also to the linked PR in the description (https://github.com/ethereum/execution-specs/pull/1734/files). I am building on top of that PR.

jochem-brouwer:
There is a practical problem with the small/big contracts here which we should address. I might raise an issue which we have discussed before here, if that is the case, sorry about this. 

It is about this: pointing to:

> (just created, small, medium, XEN)

contract sizes.

I just realized (and maybe someone else realized this before) we can use 7702 for this. Here is the takeaway:

**We can setup 4 7702 accounts with the desired storage size**

If we then want to execute specific code or operations which need to run on different sized contracts (so *(just created, small, medium, XEN)* sizes) we instead delegate the 7702 account **to that code** and then run the test.

**Here is the gotcha**: in the cleanup phase of the test, **iff the test used an SSTORE** this SSTOREs should be reset to the original values, **such that we can re-run tests on top of exactly the same contract** in storage. (This applies to situations where we run tests using `uv run execute` i.e. against remote networks, and where we cannot pre-fill / reset contracts in the genesis).

[...]

PR #1774, opened 2025-11-11
feat: add extra sstore benchmark cases

LouisTsai-Csie:
## 🗒️ Description

Add extra SSTORE benchmark cases, more details linked in the issue #1755 .

## 🔗 Related Issues or PRs

Issue #1755 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -364,3 +368,191 @@ def test_storage_access_warm(
...
+    loop += Op.JUMP
+    # [counter + 1, value, start_slot, num_slots]
+
+    # Cleanup: Stop
+    cleanup += Op.JUMPDEST
+    cleanup += Op.STOP
+
+    assert len(setup) - 1 == start_marker
+    assert len(setup) + len(loop) == end_marker
+    print(f"setup: {len(setup)}, loop: {len(loop)}, cleanup: {len(cleanup)}")
```
spencer-tb: Should we remove the print?
```suggestion
```

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -364,3 +368,191 @@ def test_storage_access_warm(
...
+            access_list = [
+                AccessList(
+                    address=contract_addr,
+                    storage_keys=storage_keys,
+                )
+            ]
+
+        contract_gas_limit = base_gas_per_contract
+        if contract_idx == 0:
+            contract_gas_limit += gas_remainder
```
spencer-tb: Why do we give the remainder to the first contract? Does the last contract make more sense?

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -364,3 +368,191 @@ def test_storage_access_warm(
...
+    txs = []
+    post = {}
+
+    base_gas_per_contract = gas_benchmark_value // num_contracts
+    gas_remainder = gas_benchmark_value % num_contracts
+
+    for contract_idx in range(num_contracts):
+        initial_storage = Storage()
+
+        start_slot = contract_idx * slot_count
```
spencer-tb: I think using `slot_count` instead of `slots_per_contract` creates storage gaps (for example contract 0 uses slots 0-9, contract 1 uses 100-109, contract 2 uses 200-209). With the latter we get, 0-9, 10-19, 20-29 etc
```suggestion
        start_slot = contract_idx * slot_per_contract
```

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -364,3 +368,191 @@ def test_storage_access_warm(
...
+@pytest.mark.parametrize("num_contracts", [1, 5, 10])
+@pytest.mark.parametrize(
+    "initial_value,write_value",
+    [
+        pytest.param(0, 0, id="zero_to_zero"),
+        pytest.param(0, 0xDEADBEEF, id="zero_to_nonzero"),
+        pytest.param(0xDEADBEEF, 0, id="nonzero_to_zero"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_nonzero"),
+    ],
+)
```
spencer-tb: This will create 384 seperate tests. Do we need this many here? Maybe we should remove some parameterization. I say this with the benchmark release process taking a long time in mind.

Maybe contract size and num contracts can have one element removed from each.

If these are required in your opinion please keep them. Just a thought.

spencer-tb (commented):
Thanks some comments! :)

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: Nice catch! I agree.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -354,3 +358,190 @@ def test_storage_access_warm(
...
+
+@pytest.mark.parametrize("slot_count", [50, 100])
+@pytest.mark.parametrize("use_access_list", [True, False])
+@pytest.mark.parametrize(
+    "contract_size",
+    [
+        pytest.param(0, id="just_created"),
+        pytest.param(1024, id="small"),
+        pytest.param(12 * 1024, id="medium"),
+        pytest.param(24 * 1024, id="xen"),
```
CPerezz: To make this compatible with Kamil's tool you should set IDs that have the name of the test _ contract name.

See my file as an example to understand the naming convention: https://gist.github.com/CPerezz/51e292676f2420cfb601d251a7ad72e7

.github/configs/feature.yaml
```diff
@@ -11,11 +11,11 @@ develop:
 
 benchmark:
   evm-type: benchmark
-  fill-params: --fork=Prague --gas-benchmark-values 1,10,30,45,60,100,150 -m "benchmark and not state_test" ./tests/benchmark
+  fill-params: --fork=Prague --gas-benchmark-values 5,10,30,45,60,100,150 -m "benchmark and not state_test" ./tests/benchmark
```
CPerezz: I think `1` is needed. But can be maybe done on the side? 

I recall Maria wanted it

jsign: Support the 5M setup -- for zkVM would be useful (so there isn't such a big gap between 1M and 10M). The 1M is also useful for use to as a very quick way to run things, so I would prefer to keep it.

CPerezz:
@LouisTsai-Csie @marioevz should we merge this such that we can start taking metrics for repricings?

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: Yes, the requirements are documented [here](https://github.com/ethereum/execution-specs/issues/1755), so all of them are necessary.

.github/configs/feature.yaml
LouisTsai-Csie: Some of the current setting (too many slot count, using access list), this would increase the gas usage and exceeds 1M. Include this benchmark will break our current benchmark building CI.

For this, i would suggest moving this benchmark under stateful folder, excluding it from compute benchmark.

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: Thanks, I’ll reach out to Kamil once he’s back from vacation.

One thing i am still not sure is how the contract size here work? Does it mean the size of `sstore` contract, which is  the attack contract in the benchmark?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,308 @@ def test_sstore_erc20_approve(
...
+@pytest.mark.parametrize("num_contracts", [1, 5, 10])
+@pytest.mark.parametrize(
+    "initial_value,write_value",
+    [
+        pytest.param(0, 0, id="zero_to_zero"),
+        pytest.param(0, 0xDEADBEEF, id="zero_to_nonzero"),
+        pytest.param(0xDEADBEEF, 0, id="nonzero_to_zero"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_nonzero"),
+    ],
+)
```
LouisTsai-Csie, line 540: This covers the scenario that edits the slots from zero to zero

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,308 @@ def test_sstore_erc20_approve(
...
+
+@pytest.mark.parametrize("slot_count", [50, 100])
+@pytest.mark.parametrize("use_access_list", [True, False])
+@pytest.mark.parametrize(
+    "contract_size",
+    [
+        pytest.param(0, id="just_created"),
+        pytest.param(1024, id="small"),
+        pytest.param(12 * 1024, id="medium"),
+        pytest.param(24 * 1024, id="xen"),
```
jochem-brouwer: This contract size has to go deeper, because the XEN size is of order size ~10 million storage slots. (This is thus also something we should not put directly in these tests, because the setup phase of these size of tests is extremely long)

For benchmarks, we have multiple degrees of freedom which likely influence the execution speed:

- The total size of the contract we work on
- The amount of MPT nodes we have to hit/change in order to calculate the state root (the "depth" of each individual storage node)
- The total size of the **network** we work on (or, more generally: the total database size. Maybe more correct here is how "dense" the database is packed. This will also depend on the database backend and the write/read search algorithms)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,308 @@ def test_sstore_erc20_approve(
...
+    loop += Op.DUP1
+    loop += Op.DUP4
+    loop += Op.ADD
+    loop += Op.DUP3
+    # [value, start_slot+counter, counter, value, start_slot, num_slots]
+
+    if sloads_before_sstore:
+        loop += Op.DUP2
+        loop += Op.SSTORE
+        loop += Op.SLOAD
```
jochem-brouwer, line 503: The SLOAD is after the SSTORE here, does not do as `sloads_before_sstore` would suggest.

As side question, what is the motivation/reason we want to have this case in the benchmark? (I am fine adding it but I do not see what case this could influence (?))

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,308 @@ def test_sstore_erc20_approve(
...
+    loop += Op.JUMP
+    # [counter + 1, value, start_slot, num_slots]
+
+    # Cleanup: Stop
+    cleanup += Op.JUMPDEST
+    cleanup += Op.STOP
+
+    assert len(setup) - 1 == start_marker
+    assert len(setup) + len(loop) == end_marker
+    return setup + loop + cleanup
```
jochem-brouwer, line 524: Note: did not thoroughly review this bytecode

tests/benchmark/stateful/bloatnet/test_single_opcode.py
jochem-brouwer: Ah ok I see that this is used here as contract code size, not storage size. This should have almost no impact on the execution speed: the only time we would incur (extra) execution time is when reading it for the first time and performing the JUMPDEST analysis on the first JUMP.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,308 @@ def test_sstore_erc20_approve(
...
+)
+@pytest.mark.parametrize("sloads_before_sstore", [True, False])
+@pytest.mark.parametrize("num_contracts", [1, 5, 10])
+@pytest.mark.parametrize(
+    "initial_value,write_value",
+    [
+        pytest.param(0, 0, id="zero_to_zero"),
+        pytest.param(0, 0xDEADBEEF, id="zero_to_nonzero"),
+        pytest.param(0xDEADBEEF, 0, id="nonzero_to_zero"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_nonzero"),
```
jochem-brouwer:
```suggestion
        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_nonzero"),
         pytest.param(0xDEADBEEF, 0xDEADBEEF, id="nonzero_to_nonzero_equal"),
```

This adds an extra test case where the value itself does not change (same as 0 -> 0 except it is now nonzero -> nonzero)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,308 @@ def test_sstore_erc20_approve(
...
+    gas_remainder = tx_gas_limit % num_contracts
+
+    for contract_idx in range(num_contracts):
+        initial_storage = Storage()
+
+        start_slot = contract_idx * slots_per_contract
+        for i in range(slots_per_contract):
+            initial_storage[start_slot + i] = initial_value
+
+        contract_addr = pre.deploy_contract(
```
jochem-brouwer, line 583: Ok, this works, but there is a problem here with the Fusaka EIP-7825 tx gas limit. If we start to create more storage slots then we are either going to hit the max initcode size limit (48 KiB) or we are going to need more gas than the gas limit. For a new storage slot, we need at least 20k gas, so `(2 ** 24) / 20_000` is 838 slots at most (in practice it will be lower)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,308 @@ def test_sstore_erc20_approve(
...
+
+    if len(base_contract) < contract_size:
+        padded_contract += Op.INVALID * (contract_size - len(base_contract))
+
+    slots_per_contract = slot_count // num_contracts
+
+    txs: list[Transaction] = []
+    post = {}
+
+    base_gas_per_contract = tx_gas_limit // num_contracts
```
jochem-brouwer: This splits the `tx_gas_limit` (I assume this is the EIP-7825 limit) over those contracts. I think this should be `min(tx_gas_limit, env.gas_limit // num_contracts)` such that we can target with multiple txs different contracts but still bounded by block gas limit (`env.gas_limit`) vs the tx gas limit (EIP-7825)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,308 @@ def test_sstore_erc20_approve(
...
+            + write_value.to_bytes(32, "big")
+        )
+
+        access_list = None
+        if use_access_list:
+            storage_keys = [
+                Hash(start_slot + i) for i in range(slots_per_contract)
+            ]
+            access_list = [
+                AccessList(
```
jochem-brouwer, line 600: Side note: this will increase the intrinsic gas cost of the tx. However, the execution gas used for the storage slots the execution hits will be lower. So if the access list is correct this should yield the "correct" slots to read/write to.

jochem-brouwer (commented):
Some comments, I'll review the final test for SLOAD tomorrow 😄 👍

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: I think I misunderstood the meaning of _storage size_ here. I originally thought it referred to the contract’s code size, but it actually seems to mean the storage footprint of the contract.

I’ll remove the logic related to contract code size and instead look into how we can explicitly configure the contract’s storage.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: I thought this was requested by Maria, no? I created this PR based on CPerezz’s PR and issue #1755 , which I understood had been discussed with Maria.

I would keep it as is (but change to SLOAD before SSTORE logic), if necessary i could remove it in the future.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: Added!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: You’re right, I see the issue now. Using EIP-7702 could give us much more flexibility here.
- Create an EOA that delegates to a contract responsible for pre-filling the storage.
- Then delegate that same EOA to the contract we want to benchmark.

This approach would be much more convenient and flexible.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: Good catch, i need to check again. But a small note here, the block gas limit should be `gas_benchmark_values` instead of `env.gas_limit`.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+@pytest.mark.parametrize("sloads_before_sstore", [True, False])
+@pytest.mark.parametrize("num_contracts", [1, 5, 10])
+@pytest.mark.parametrize(
+    "initial_value,write_value",
+    [
+        pytest.param(0, 0, id="zero_to_zero"),
+        pytest.param(0, 0xDEADBEEF, id="zero_to_nonzero"),
+        pytest.param(0xDEADBEEF, 0, id="nonzero_to_zero"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_diff"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_same"),
```
CPerezz, line 538: Here we aren't doing nonzero to same right? We are writing a different value.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+        post[contract_addr] = Account(
+            code=padded_contract,
+            storage=expected_storage,
+        )
+
+    benchmark_test(
+        blocks=[Block(txs=txs)],
+        post=post,
+        skip_gas_used_validation=True,
+    )
```
CPerezz, line 632: Wouldn't we need at least to pass `pre` here? I wonder why don't we even need to deploy the OG contract prior doing the rest of the test.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+    gas_remainder = tx_gas_limit % num_contracts
+
+    for contract_idx in range(num_contracts):
+        initial_storage = Storage()
+
+        start_slot = contract_idx * slots_per_contract
+        for i in range(slots_per_contract):
+            initial_storage[start_slot + i] = initial_value
+
+        contract_addr = pre.deploy_contract(
```
CPerezz, line 583: If `pre` is not passed later, this won't be deployed right?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+) -> None:
+    """
+    Benchmark SSTORE instruction with various configurations.
+
+    Variants:
+    - use_access_list: Warm storage slots via access list
+    - sloads_before_sstore: Number of SLOADs per slot before SSTORE
+    - num_contracts: Number of contract instances (cold storage writes)
+    - initial_value/write_value: Storage transitions
+      (zero_to_zero, zero_to_nonzero, nonzero_to_zero, nonzero_to_nonzero)
```
CPerezz, line 712: This might be copy-paste? But the test is about SLOADs and the docstring mentions SSTOREs

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+            access_list = [
+                AccessList(
+                    address=contract_addr,
+                    storage_keys=storage_keys,
+                )
+            ]
+
+        contract_gas_limit = base_gas_per_contract
+        if contract_idx == len(txs) - 1:
+            contract_gas_limit += gas_remainder
```
CPerezz, line 608: `txs` is being built, so this is always False initially.

At the point of this check, txs is being populated in the loop, so len(txs) - 1 will always be less than contract_idx on the first iterations. The condition should be:

`if contract_idx == num_contracts - 1:` is what we should have??

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+    Variants:
+    - use_access_list: Warm storage slots via access list
+    - sloads_before_sstore: Number of SLOADs per slot before SSTORE
+    - num_contracts: Number of contract instances (cold storage writes)
+    - initial_value/write_value: Storage transitions
+      (zero_to_zero, zero_to_nonzero, nonzero_to_zero, nonzero_to_nonzero)
+    """
+    slots: set[int] = set()
+    if storage_keys_set:
+        slots = {i * incrementer for i in range(num_slots)}
```
CPerezz, line 716: If `incrementer = 0`, all slots collapse to {0}, meaning only 1 slot is accessed regardless of `num_slots`. Is this the intended behaviour?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+    # - CALLDATA[0..31]: Number of slots to access
+    # - CALLDATA[32..63]: Starting slot index
+    # - CALLDATA[64..95]: Value to write
+    """
+    setup = Bytecode()
+    loop = Bytecode()
+    cleanup = Bytecode()
+
+    start_marker = 10
+    end_marker = 30 + (2 if sloads_before_sstore else 0)
```
CPerezz, line 473: Can we give const values or add a comment to avoid magic numbers in the test?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+        access_list=access_lists,
+        data=calldata,
+        sender=pre.fund_eoa(),
+    )
+
+    benchmark_test(
+        pre=pre,
+        blocks=[Block(txs=[tx])],
+        skip_gas_used_validation=True,
+    )
```
CPerezz, line 748: Wouldn't it be healthy to make sure we are indeed reading something? So basically, add a minimal check here? It won't even run on the measured block so should not be an issue.

CPerezz (changes requested):
PR tittle should change to: `feat(benchmark): add extra sstore benchmark cases`

Some minor things here and there. But looks good overall!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
spencer-tb: Yeah this should be `pytest.param(0xDEADBEEF, 0xDEADBEEF, id="nonzero_to_same")`

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+
+    slots_per_contract = slot_count // num_contracts
+
+    txs: list[Transaction] = []
+    post = {}
+
+    base_gas_per_contract = min(
+        tx_gas_limit, gas_benchmark_values // num_contracts
+    )
+    gas_remainder = tx_gas_limit % num_contracts
```
spencer-tb, line 574: Should we do something like this?
```suggestion
if gas_benchmark_values // num_contracts <= tx_gas_limit:
    base_gas_per_contract = gas_benchmark_values // num_contracts
    gas_remainder = gas_benchmark_values % num_contracts
else:
    base_gas_per_contract = tx_gas_limit
    gas_remainder = 0
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+@pytest.mark.parametrize(
+    "initial_value,write_value",
+    [
+        pytest.param(0, 0, id="zero_to_zero"),
+        pytest.param(0, 0xDEADBEEF, id="zero_to_nonzero"),
+        pytest.param(0xDEADBEEF, 0, id="nonzero_to_zero"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_diff"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_same"),
+    ],
+)
```
spencer-tb, line 540: Do we need to mark these valid from a fork like the other tests?
```suggestion
@pytest.mark.valid_from("Prague")
)
```

It be might be better to define this once at the top of the file.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: Which does the OG contract here mean?

We use `benchmark_test` here, which has a default `pre` instance. You could also use this in your bloatnet test also.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: Same as above - with `benchmark_test`, we do not need the pre instance.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+    # Cleanup: Stop
+    cleanup += Op.JUMPDEST
+    cleanup += Op.STOP
+
+    assert len(setup) - 1 == start_marker
+    assert len(setup) + len(loop) == end_marker
+    return setup + loop + cleanup
+
+
+@pytest.mark.parametrize("slot_count", [50, 100])
```
marioevz, line 527: I think my main issue with these tests is that they hard-code a number of slots to be accessed and completely disregard gas usage, which makes these tests incompatible with the rest of the benchmarking tests IMO.

Is this a specific requirement for bloatnet or can we scrap this parameter and make the slot count a function of the gas to be used?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+        pytest.param(0xDEADBEEF, 0, id="nonzero_to_zero"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_diff"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_same"),
+    ],
+)
+def test_sstore_variants(
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    tx_gas_limit: int,
+    gas_benchmark_values: int,
```
marioevz, line 545:
```suggestion
    gas_benchmark_value: int,
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
jochem-brouwer, line 538: This writes a different value, not the same value

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -450,3 +454,295 @@ def test_sstore_erc20_approve(
...
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    warm_slots: bool,
+    storage_keys_set: bool,
+    num_slots: int,
+    incrementer: int,
+    tx_gas_limit: int,
+) -> None:
+    """
+    Benchmark SSTORE instruction with various configurations.
```
jochem-brouwer, line 705:
```suggestion
    Benchmark SLOAD instruction with various configurations.
```

LouisTsai-Csie:
Done by PR #2130

Issue #1599, opened 2025-10-14
Unify gas costs source for faster repricing cycles


LouisTsai-Csie:
# Description

[EIP-7904](https://eips.ethereum.org/EIPS/eip-7904), proposed for the upcoming fork, introduces a set of eips that will update gas costs with actual computational overhead for each opcode.

However, the current setup relies on two separate data sources for gas models, requiring engineers to update both systems manually each time gas values change.

# Current Structure

## EELS
Defined in `src/ethereum/forks/{fork_name}/vm/gas.py` -> [Example](https://github.com/ethereum/execution-specs/blob/forks/osaka/src/ethereum/forks/osaka/vm/gas.py)

For each opcode, charge specific gas amount based on the opcode's gas category:

```python
def add(evm: Evm) -> None:
    x = pop(evm.stack)
    y = pop(evm.stack)

    charge_gas(evm, GAS_VERY_LOW) # ADD is under GAS_VERY_LOW category
    result = x.wrapping_add(y)

    push(evm.stack, result)
    evm.pc += Uint(1)
```

## EEST
Defined in `src/ethereum_spec_tests/ethereum_test_forks/forks/forks.py` -> [Example](https://github.com/ethereum/execution-specs/blob/5b5e026dc2ac7c1a7517bd684693b059c375cbeb/src/ethereum_spec_tests/ethereum_test_forks/forks/forks.py#L80)

In benchmark tests, this is used by hardcoded the value, similar to the EELS implementation here:
```python
loop_cost = (
    gas_costs.G_KECCAK_256  # KECCAK static cost
    + math.ceil(85 / 32) * gas_costs.G_KECCAK_256_WORD  # KECCAK dynamic
    # cost for CREATE2
    + gas_costs.G_VERY_LOW * 3  # ~MSTOREs+ADDs
    + gas_costs.G_COLD_ACCOUNT_ACCESS  # CALL to self-destructing contract
    + gas_costs.G_SELF_DESTRUCT
    + 63  # ~Gluing opcodes
)
```

## Difference
The difference between the variables definition, their naming or related analysis could be found [here](https://hackmd.io/@QmVpC8TxQ8a1nTCW46EsEQ/Sy_RzMULxl).

## Potential Issue
[...]

PR #1776, opened 2025-11-11
feat(tests): unify gas cost variable

LouisTsai-Csie:
## 🗒️ Description

Unify the gas cost variable name from Cancun to Osaka. Introduce a fork-dependent gas table.

The mapping for the EELS & EEST gas constant, I replace all the EEST variable name with the EELS naming, as this requires less effort. This documents the mapping: https://hackmd.io/@QmVpC8TxQ8a1nTCW46EsEQ/B1932RWy-x

Note: This PR is only a prototype and will be split into smaller PRs.

## 🔗 Related Issues or PRs

Issue #1599 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


packages/testing/src/execution_testing/vm/tests/test_opcode_gas_calculator.py
```diff
@@ -0,0 +1,208 @@
...
+    """Test that Prague and Osaka have same base costs for common opcodes."""
+    code = Op.PUSH1(0x01) + Op.PUSH1(0x02) + Op.ADD
+
+    prague_calc = OpcodeGasCalculator(Prague)
+    osaka_calc = OpcodeGasCalculator(Osaka)
+
+    prague_gas = prague_calc.calculate(code)
+    osaka_gas = osaka_calc.calculate(code)
+
+    assert prague_gas == osaka_gas == 9  # 3 + 3 + 3
```
LouisTsai-Csie, line 208: This is the example usage:

```python
code = Op.PUSH0 + Op.PUSH0 + Op.ADD
prague_calc = OpcodeGasCalculator(Prague)
prague_gas = prague_calc.calculate(code)
# the total gas cost for the code sequnece
```

This could be extended to:
```python
code = Op.PUSH0 + Op.PUSH0 + Op.ADD
calc = OpcodeGasCalculator(fork)
gas =calc.calculate(code)
```
This version could help us avoid hardcoding the gas cost. As i believe in the near future there would be gas repricing for ZkVM, and we do not want to update all the gas cost again.

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -2309,6 +2354,59 @@ def blob_base_cost(
...
+        for field in fields(GasCosts):
+            if hasattr(g, field.name):
+                kwargs[field.name] = int(getattr(g, field.name))
+
+        return GasCosts(**kwargs)
+
+    @classmethod
+    def op_cost(
+        cls, opcode: Opcodes, *, block_number: int = 0, timestamp: int = 0
+    ) -> int:
```
LouisTsai-Csie, line 2375: We need to define this `op_cost` function for every fork to enable fork-dependent gas table feature.

SamWilsn:
:us:

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
 GAS_BLS_G2_MUL = Uint(22500)
 GAS_BLS_G2_MAP = Uint(23800)
 
+# EEST Specific Variables
+G_TX_DATA_ZERO = 4
+G_TX_DATA_NON_ZERO = 16
+
+# Transaction costs (from transactions.py)
```
SamWilsn, line 91: Comments should describe the current state of the code. The git log gives us historical context.

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
 GAS_BLS_G2_MUL = Uint(22500)
 GAS_BLS_G2_MAP = Uint(23800)
 
+# EEST Specific Variables
+G_TX_DATA_ZERO = 4
+G_TX_DATA_NON_ZERO = 16
```
SamWilsn, line 89: Should these be typed?

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+TX_BASE_COST = Uint(21_000)
+TX_CREATE_COST = Uint(32_000)
+TX_ACCESS_LIST_ADDRESS_COST = Uint(2_400)
+TX_ACCESS_LIST_STORAGE_KEY_COST = Uint(1_900)
+STANDARD_CALLDATA_TOKEN_COST = Uint(4)
+FLOOR_CALLDATA_COST = Uint(10)
+
+# Authorization costs (from eoa_delegation.py)
+PER_EMPTY_ACCOUNT_COST = 25_000
+PER_AUTH_BASE_COST = 12_500
```
SamWilsn, line 101: Should these be typed?

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+
+# Opcode Costs
+G_STOP = GAS_ZERO
```
SamWilsn, line 104: We try to avoid abbreviations unless they are extremely common (eg. tx -> transaction). I'm not firmly opposed to g -> gas, but I did want to mention it.

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+G_ADD = GAS_VERY_LOW
+G_MUL = GAS_LOW
+G_SUB = GAS_VERY_LOW
+G_DIV = GAS_LOW
+G_SDIV = GAS_LOW
+G_MOD = GAS_LOW
+G_SMOD = GAS_LOW
+G_ADDMOD = GAS_MID
+G_MULMOD = GAS_MID
+G_EXP = GAS_EXPONENTIATION
```
SamWilsn, line 114: Might be worth adding a docstring to each of these constants that explains the difference?

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+G_AND = GAS_VERY_LOW
+G_OR = GAS_VERY_LOW
+G_XOR = GAS_VERY_LOW
+G_NOT = GAS_VERY_LOW
+G_BYTE = GAS_VERY_LOW
+G_SHL = GAS_VERY_LOW
+G_SHR = GAS_VERY_LOW
+G_SAR = GAS_VERY_LOW
+G_CLZ = GAS_LOW
+G_KECCAK256 = GAS_KECCAK256
```
SamWilsn, line 131: Same comment here about explaining the difference.

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+G_ADDRESS = GAS_BASE
+G_BALANCE = GAS_WARM_ACCESS  # or GAS_COLD_ACCOUNT_ACCESS if cold
```
SamWilsn, line 133: Hm, I'm not sold on this. If the price of an opcode is dynamic, should the associated gas constant have a different naming scheme? Like `G_BASE_BALANCE` or `G_MIN_BALANCE`? I'm just musing here, a change is not necessarily required.

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+G_GAS = GAS_BASE
+G_JUMPDEST = GAS_JUMPDEST
+G_TLOAD = GAS_WARM_ACCESS
+G_TSTORE = GAS_WARM_ACCESS
+G_MCOPY = GAS_VERY_LOW
+G_PUSH0 = GAS_BASE
+G_PUSHx = GAS_VERY_LOW
+G_DUPx = GAS_VERY_LOW
+G_SWAPx = GAS_VERY_LOW
+G_LOGx = GAS_LOG
```
SamWilsn, line 178:
```suggestion
G_PUSH = GAS_VERY_LOW
G_DUP = GAS_VERY_LOW
G_SWAP = GAS_VERY_LOW
G_LOG = GAS_LOG
```

The lowercase "x" is unusual in Python constants.

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+G_TX_DATA_ZERO = 4
+G_TX_DATA_NON_ZERO = 16
+
+# Transaction costs (from transactions.py)
+TX_BASE_COST = Uint(21_000)
+TX_CREATE_COST = Uint(32_000)
+TX_ACCESS_LIST_ADDRESS_COST = Uint(2_400)
+TX_ACCESS_LIST_STORAGE_KEY_COST = Uint(1_900)
+STANDARD_CALLDATA_TOKEN_COST = Uint(4)
+FLOOR_CALLDATA_COST = Uint(10)
```
SamWilsn, line 97: IIUC, these constants are _not_ used by anything under the `ethereum.forks.osaka.vm` package, right? It seems odd that they live in here if they aren't related to the virtual machine itself.

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+
+# Authorization costs (from eoa_delegation.py)
```
Carsons-Eels, line 99: Same as [here](https://github.com/ethereum/execution-specs/pull/1776/files#r2566116532), the "(from eoa_delegation.py)" is probably not required.

src/ethereum/forks/osaka/vm/gas.py
```diff
@@ -84,6 +84,109 @@
...
+PER_EMPTY_ACCOUNT_COST = 25_000
+PER_AUTH_BASE_COST = 12_500
+
+# Opcode Costs
```
Carsons-Eels, line 103: Right now these are in order of their opcode number, but the opcode numbers are not listed so it's not easy to reference them in that way. I'm wondering if there is a better way to organize this. Maybe by common cost?

src/ethereum/forks/osaka/vm/gas.py
Carsons-Eels: I think at the very least we would want to document that so it's clear. I'm not sure that saving two characters is worth the loss in readability though.

src/ethereum/forks/osaka/vm/gas.py
Carsons-Eels: I count 7 of these, so it wouldn't be a huge addition to document each. If we do end up reorganizing this list into groups by gas cost, then these would all be in one block which would make for easy reading.

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -2309,6 +2354,59 @@ def blob_base_cost(
...
+        """Return the gas costs for the fork."""
+        from ethereum.forks.osaka.vm import gas as g
+        from dataclasses import fields
+
+        kwargs = {}
+        for field in fields(GasCosts):
+            if hasattr(g, field.name):
+                kwargs[field.name] = int(getattr(g, field.name))
+
+        return GasCosts(**kwargs)
```
Carsons-Eels, line 2370: Should we return an error here if the field is not found rather than skipping it silently? Not sure if this is intended.

src/ethereum/forks/osaka/vm/instructions/arithmetic.py
```diff
@@ -18,11 +18,18 @@
...
+    G_DIV,
+    G_EXP,
+    G_MOD,
+    G_MUL,
+    G_MULMOD,
+    G_SDIV,
+    G_SIGNEXTEND,
+    G_SMOD,
+    G_SUB,
     GAS_EXPONENTIATION_PER_BYTE,
```
Carsons-Eels, line 32: Why does this one keep it's `GAS_` prefix rather than aligning with the others?

Carsons-Eels (commented):
I'm not sure that I like mixing the use of the "G" and "GAS" prefixes (`G_OPCODE` vs `GAS_OPCODE`) in the codebase. If I understand correctly, the intent is to use "G" in the EELS part of the code and "GAS" in the EEST part of the code? In my mind that should be reversed, because a IMO higher priority should be placed on the how readable the spec is. Should we maybe consider simply aligning the naming of them all? 

I'll keep looking this over in the morning, this is great stuff 🚀

src/ethereum/forks/osaka/vm/gas.py
LouisTsai-Csie: I support unifying the gas cost constants to the `GAS_OPCODE` format, it’s more clear.

src/ethereum/forks/osaka/vm/gas.py
LouisTsai-Csie: @Carsons-Eels What are the 7 categories you’re referring to here? I counted 16 unique gas categories, so I may be misunderstanding something.

```
1. GAS_ZERO
2. GAS_VERY_LOW
3. GAS_LOW
4. GAS_MID
5. GAS_EXPONENTIATION
6. GAS_KECCAK256
7. GAS_BASE
8. GAS_WARM_ACCESS
9. GAS_COLD_ACCOUNT_ACCESS
10. GAS_COLD_SLOAD
11. GAS_BLOCK_HASH
12. GAS_BLOBHASH_OPCODE
13. GAS_JUMPDEST
14. GAS_LOG
15. GAS_CREATE
16. GAS_SELF_DESTRUCT
```

And this is my idea for docstring as comment, is this the format you'd preferred?

```python
"""Cost for zero-cost operations."""
GAS_ZERO = Uint(0)

"""Cost for the JUMPDEST opcode."""
GAS_JUMPDEST = Uint(1)

"""Base cost for simple environment and stack operations."""
GAS_BASE = Uint(2)

"""Cost for very low-complexity operations."""
GAS_VERY_LOW = Uint(3)

"""Cost for low-complexity operations."""
GAS_LOW = Uint(5)

"""Cost for medium-complexity operations."""
GAS_MID = Uint(8)

"""Cost for high-complexity operations."""
GAS_HIGH = Uint(10)

"""Warm account or storage access cost."""
GAS_WARM_ACCESS = Uint(100)
...
```

src/ethereum/forks/osaka/vm/gas.py
LouisTsai-Csie: If we adopt the `GAS_OPCODE` naming pattern, do we still need this assignments for consistency? And do we still need docstring here?

```
GAS_SAR = GAS_VERY_LOW
GAS_CLZ = GAS_LOW
GAS_KECCAK256 = GAS_KECCAK256
...
```

src/ethereum/forks/osaka/vm/gas.py
LouisTsai-Csie: I’d like to discuss more on this, because I’m not fully convinced my current approach is ideal either

Many opcodes use a dynamic cost model, and assigning something like the `BALANCE` opcode to a fixed category is confusing. I’m leaning toward your suggestion, `GAS_BASE_BALANCE`, but I’m open to other naming options if you prefer.

Same rules apply to memory operations that contains memory expansion cost, like `GAS_BASE_MLOAD` for `GAS_MLOAD`

src/ethereum/forks/osaka/vm/gas.py
LouisTsai-Csie: Yes, this is not used in `vm` package but defined in `transaction.py`. 

My current approach is to put these gas cost constant together in `gas.py` only, and then import them in the testing framework, like the example below.

```python
@classmethod
def gas_costs(
  cls, *, block_number: int = 0, timestamp: int = 0
) -> GasCosts:
  """
  On Prague, the standard token cost and the floor token costs are
  introduced due to EIP-7623.
  """
  from ethereum.forks.prague.vm import gas as g
  from dataclasses import fields

  kwargs = {}
  for field in fields(GasCosts):
      if hasattr(g, field.name):
          kwargs[field.name] = int(getattr(g, field.name))

  return GasCosts(**kwargs)
```

I copy these gas cost constant from `transaction.py` / `eoa_delegation.py` so in the import logic i do not need to import two files for the gas costs definition.

But i could refactor the logic to something below, wdyt?

```python
@classmethod
def gas_costs(
  cls, *, block_number: int = 0, timestamp: int = 0
) -> GasCosts:
  """Return the gas costs for the fork."""
  from dataclasses import fields
  from ethereum.forks.osaka.vm import gas as g
  from ethereum.forks.osaka.vm import eoa_delegation as d
  from ethereum.forks.osaka import transactions as tx

  modules = (g, tx, d)
  kwargs = {}

  for field in fields(GasCosts):
      for mod in modules:
          if hasattr(mod, field.name):
              kwargs[field.name] = int(getattr(mod, field.name))
              break
      else:
          # executed only if the inner loop completed without a break
          raise AttributeError(
              f"Field '{field.name}' not found in gas, transactions, or eoa_delegation modules."
          )

  return GasCosts(**kwargs)
```

I am so shocked i do not know there is for-else statement before.

src/ethereum/forks/osaka/vm/gas.py
LouisTsai-Csie: Does this [approach](https://github.com/ethereum/execution-specs/pull/1776#discussion_r2576239101) resolve the problem? Based on the multiple file import scheme we do not need to copy these variable from `eoa_delegation.py` and `transaction.py`.

packages/testing/src/execution_testing/forks/forks/forks.py
LouisTsai-Csie: Yes, we should return an error here like [this](https://github.com/ethereum/execution-specs/pull/1776#discussion_r2576239101)

src/ethereum/forks/osaka/vm/instructions/arithmetic.py
LouisTsai-Csie: I agree to unify the pattern, i cannot even remember why i do this before tbh

src/ethereum/forks/osaka/vm/gas.py
LouisTsai-Csie: Do you mean organize them by something like this?

```python
"""Operations for very low gas cost"""
G_ADD = GAS_VERY_LOW
G_SUB = GAS_VERY_LOW
G_LT = GAS_VERY_LOW
G_GT = GAS_VERY_LOW
G_SLT = GAS_VERY_LOW
...

"""Operations for low gas cost"""
G_MUL = GAS_LOW
G_DIV = GAS_LOW
G_SDIV = GAS_LOW
G_MOD = GAS_LOW
G_SMOD = GAS_LOW
...

"""Operations for medium gas cost"""
G_ADDMOD = GAS_MID
G_MULMOD = GAS_MID
...
# continue on different gas categories...
```

src/ethereum/forks/osaka/vm/gas.py
Carsons-Eels: I believe I was referring to variables assigned to existing variables, rather than a specific number. Probably where the misunderstanding stems from. 

I don't believe there is a specific PEP standard that deals with this question. I think if it's just going to be a simple single line comment, then we should default to the simpler standard comment.
```python
# This comment explains the constant
THE_CONSTANT = Uint(0)
```
If it requires a more detailed multi-line explanation, then tripple quotes seem better. You could also group them all into a single block comment that explains all the constants below it like this:
```python
"""
Gas configuration constants.

Constants:
    GAS_ZERO: Cost for zero-cost operations.
    GAS_JUMPDEST: Cost for the JUMPDEST opcode.
    ... etc
"""
GAS_ZERO = Uint(0)
GAS_JUMPDEST = Uint(1)
... etc
```

src/ethereum/forks/osaka/vm/instructions/arithmetic.py
Carsons-Eels: Heh, no worries, sounds good :+1: :smile:

src/ethereum/forks/osaka/vm/gas.py
Carsons-Eels: Yeah, that's pretty much exactly what I was thinking!

SamWilsn:
How does this related to https://github.com/ethereum/execution-specs/pull/2094 ? Can one of them be closed?

Carsons-Eels:
> How does this related to #2094 ? Can one of them be closed?

I believe this is superseded by #2094 and the changes @marioevz made to cast cost calculation. I can be closed if there are no objections from @LouisTsai-Csie

LouisTsai-Csie:
@Carsons-Eels Yes, let's continue on PR #2094 and close this one

PR #1787, opened 2025-11-13
refactor(test-benchmark): optimize benchmark test cases

LouisTsai-Csie:
## 🗒️ Description

Optimize some benchmark test cases to enable code generator in benchmark test wrapper.

## 🔗 Related Issues or PRs

None

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/compute/instruction/test_account_query.py
+++ b/tests/benchmark/compute/instruction/test_account_query.py
@@ -327,31 +327,20 @@ def test_extcodecopy_warm(
     benchmark_test: BenchmarkTestFiller,
     pre: Alloc,
     copied_size: int,
-    gas_benchmark_value: int,
 ) -> None:
     """Benchmark EXTCODECOPY instruction."""
     copied_contract_address = pre.deploy_contract(
         code=Op.JUMPDEST * copied_size,
     )
 
-    execution_code = (
-        Op.PUSH10(copied_size)
-        + Op.PUSH20(copied_contract_address)
-        + While(
-            body=Op.EXTCODECOPY(Op.DUP4, 0, 0, Op.DUP2),
-        )
-    )
-    execution_code_address = pre.deploy_contract(code=execution_code)
-    tx = Transaction(
-        to=execution_code_address,
-        gas_limit=gas_benchmark_value,
-        sender=pre.fund_eoa(),
+    benchmark_test(
+        code_generator=JumpLoopGenerator(
+            setup=Op.PUSH10(copied_size) + Op.PUSH20(copied_contract_address),
+            attack_block=Op.EXTCODECOPY(Op.DUP4, 0, 0, Op.DUP2),
+        ),
     )
 
-    benchmark_test(tx=tx)
 
-
-@pytest.mark.repricing(absent_target=False)
 @pytest.mark.parametrize(
     "opcode",
     [
--- a/tests/benchmark/compute/instruction/test_call_context.py
+++ b/tests/benchmark/compute/instruction/test_call_context.py
@@ -22,11 +22,9 @@ from execution_testing import (
     Fork,
     JumpLoopGenerator,
     Op,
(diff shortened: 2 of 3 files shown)


tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -327,31 +327,20 @@ def test_extcodecopy_warm(
...
+        code_generator=JumpLoopGenerator(
+            setup=Op.PUSH10(copied_size) + Op.PUSH20(copied_contract_address),
+            attack_block=Op.EXTCODECOPY(Op.DUP4, 0, 0, Op.DUP2),
+        ),
     )
 
-    benchmark_test(tx=tx)
-
 
-@pytest.mark.repricing(absent_target=False)
```
spencer-tb, line 354: Should this be removed?

tests/benchmark/compute/instruction/test_call_context.py
```diff
@@ -64,44 +62,50 @@ def test_calldatasize(
     )
 
 
-@pytest.mark.repricing(non_zero_value=True, from_origin=True)
```
spencer-tb, line 67: Same here. Should this be removed?

tests/benchmark/compute/instruction/test_call_context.py
```diff
@@ -64,44 +62,50 @@ def test_calldatasize(
...
 @pytest.mark.parametrize("non_zero_value", [True, False])
-@pytest.mark.parametrize("from_origin", [True, False])
-def test_callvalue(
```
spencer-tb, line 70: My understanding of the this test has changed I think. The original looks like it was specifically measuring how expensive CALLVALUE is when you're inside a CALL frame by doing many CALLVALUEs per CALL.

The new test (`test_callvalue_from_call`) is more like how expensive is a CALL when the callee does some CALLVALUEs.

Just want to check this is intentional too!

spencer-tb (approved):
Approved from my side. Feel free to merge. Just some comments on sanity checks.

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: I remove all repricing markers, since they’re incorrect atm We should be using the least-performant benchmark  parameter combination (Based on gas-benchmark from Nethermind) as the reference test. With that in mind, I’ve removed some of the repricing markers in this PR.

tests/benchmark/compute/instruction/test_call_context.py
LouisTsai-Csie: Same as above discussion!

tests/benchmark/compute/instruction/test_call_context.py
LouisTsai-Csie: The original `test_callvalue` has been split into two more focused tests:
- `test_callvalue_from_call`: executes the `POP(CALLVALUE)` pattern repeatedly within a single contract, maximizing the number of iterations in one `CALL` frame.
- `test_callvalue_from_origin`: Uses two contracts, (1) a target contract that performs repeated `CALLVALUE` operations until hitting the stack depth limit, and (2) a loop contract that repeatedly calls into the target contract.

> how expensive CALLVALUE is when you're inside a CALL frame by doing many CALLVALUEs per CALL.

From the original implementation, although it tries to run as much `CALLVALUE` as possible, we could only run up to 1025 iterations before hitting the stack limit, where the final iteration would overflow the stack. With the split tests: ` test_callvalue_from_origin` can run 1024 iterations without triggering an invalid CALL.

So i think the performance / intention is not too different, wdyt?

spencer-tb: Sounds good to me! Makes sense - thanks for clarifying! :D

PR #1789, opened 2025-11-13
enhance(spec-tool): create blob parameter-only hardforks on the fly

### LINKED ISSUE #1453: Generate ephemeral BPO test-only forks during CI fixture building
While reviewing #1372, it became clear that the best way to handle test-only BPO forks is to generate them on the fly during the CI fixture building process using the `new_fork` tool.

This approach avoids:
- The need to modify the t8n interface to support `new_fork` before running tests, which would otherwise complicate the workflow. Instead, the process can be integrated after welding because everything is in the same repo and branch.
- Having to commit these test-only forks into any branch, or having to rebase these BPO forks if anything changes in the base fork.

Proposed workflow:
1. Define a `bpo_test_forks.yaml` file containing the parameters for each BPO fork.
  - This file is parsed only during CI fixture building.
2. Before running fill (still in the CI environment), invoke the `new_fork` tool once per BPO fork to generate each required EELS BPO fork folder.
3. Run fill to generate the BPO tests.

Outcome:
- These BPO test-only forks are generated ephemerally and never need to be committed to any EELS branch.
- The solution keeps BPO forks isolated to the testing process, reducing maintenance overhead.

SamWilsn:
## 🗒️ Description

This pull request hooks up the new fork tool to the `Hardfork` class, uses that integration to generate new forks when the blob parameters are changed by the tests.

## 🔗 Related Issues or PRs

closes #1453 

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/__init__.py
@@ -6,20 +6,22 @@ import argparse
 import fnmatch
 import json
 import os
-from typing import Any, Final, TextIO, Type, TypeVar
+from contextlib import AbstractContextManager
+from typing import Any, Final, TextIO, Tuple, Type, TypeVar
 
 from ethereum_rlp import rlp
 from ethereum_types.numeric import U64, U256, Uint
+from typing_extensions import override
 
 from ethereum import trace
 from ethereum.exceptions import EthereumException, InvalidBlock
-from ethereum_spec_tools.forks import Hardfork
+from ethereum.fork_criteria import ByBlockNumber, ByTimestamp, Unscheduled
+from ethereum_spec_tools.forks import Hardfork, TemporaryHardfork
 
 from ..loaders.fixture_loader import Load
-from ..loaders.fork_loader import ForkLoad
 from ..utils import (
     FatalError,
-    get_module_name,
+    find_fork,
     get_stream_logger,
     parse_hex_or_int,
 )
@@ -46,6 +48,12 @@ def t8n_arguments(subparsers: argparse._SubParsersAction) -> None:
     )
     t8n_parser.add_argument(
         "--input.txs", dest="input_txs", type=str, default="txs.json"
+    )
+    t8n_parser.add_argument(
+        "--input.blobParams",
+        dest="blob_parameters",
+        type=str,
+        default=None,
     )
     t8n_parser.add_argument(
         "--output.alloc", dest="output_alloc", type=str, default="alloc.json"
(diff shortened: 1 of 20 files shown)


src/ethereum_spec_tools/evm_tools/utils.py
```diff
@@ -26,6 +27,9 @@
 W = TypeVar("W", Uint, U64, U256)
 
 EXCEPTION_MAPS = {
+    "BPO1": {
+        "fork_blocks": [("osaka", 0)],
+    },
```
spencer-tb, line 32: We need this to fill all the tests!
```suggestion
    "BPO1": {
        "fork_blocks": [("osaka", 0)],
    },
    "BPO2": {
        "fork_blocks": [("osaka", 0)],
    },
    "BPO3": {
        "fork_blocks": [("osaka", 0)],
    },
    "BPO4": {
        "fork_blocks": [("osaka", 0)],
    },
```

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -81,32 +89,142 @@ def t8n_arguments(subparsers: argparse._SubParsersAction) -> None:
...
+        target_blob_gas_per_block: U64 | None = None,
+        gas_per_blob: U64 | None = None,
+        min_blob_gasprice: Uint | None = None,
+        blob_base_fee_update_fraction: Uint | None = None,
+        max_blob_gas_per_block: U64 | None = None,
+        blob_schedule_target: U64 | None = None,
+        blob_schedule_max: U64 | None = None,
+    ) -> Hardfork:
+        """
+        Search the cache for a maching hardfork, or create one if it doesn't
```
spencer-tb:
```suggestion
        Search the cache for a matching hardfork, or create one if it doesn't
```

spencer-tb (commented):
Thanks for adding this. Just a couple small comments!

I filled all our BPO tests with this and they pass on Geth :D

~~Would you be happy adding [this](https://github.com/SamWilsn/eth1.0-specs/compare/dynamic-bpo...spencer-tb:execution-specs:dynamic-bpo-tweak) `feature.yaml` change, is so once merged we should create a quick patch release on top of [v5.4.0](https://github.com/ethereum/execution-spec-tests/releases/tag/v5.4.0)?~~ edited: added here https://github.com/ethereum/execution-specs/pull/1845

spencer-tb (approved):
LGTM!

FYI I am filling with:
```
uv run fill --from=Osaka --until=BPO4 --clean -n 4 -k BPO -v
```

SamWilsn:
@spencer-tb should I add `--until BPO4` to our `tox.ini`?

spencer-tb:
> @spencer-tb should I add `--until BPO4` to our `tox.ini`?

Yes please! Totally forgot 😓

SamWilsn:
Done :3

spencer-tb:
cc-ing from Discord:

I think this is the cause for the CI fails.
```
pytest-xdist -> fill command -> Hardfork.clone()
-> libcst_tool() -> spawns process pool -> PicklingError
```

The AI tells me to add this [here](https://github.com/SamWilsn/eth1.0-specs/blob/3ec7644f74b16bf6f32b68428ad9ce1d0bde5e75/src/ethereum_spec_tools/new_fork/builder.py#L450-L453).
```python
  exit_code = libcst_tool(
      "<libcst_tool>",
      [args[0], "-j", "1"] + args[1:],
  )
```

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -81,32 +89,142 @@ def t8n_arguments(subparsers: argparse._SubParsersAction) -> None:
...
+            target_blob_gas_per_block,
+            gas_per_blob,
+            min_blob_gasprice,
+            blob_base_fee_update_fraction,
+            max_blob_gas_per_block,
+            blob_schedule_target,
+            blob_schedule_max,
+        )
+        if all(x is None for x in cache_key[1:]):
+            return template
```
marioevz, line 136: We currently pass the blob parameters unconditionally from the tests, i.e. as long as the fork supports blobs we pass the parameters. And I think we are generating an unnecessary exact copy of the fork in that case.

We could skip this for forks that are defined in the transition tool and only pass the blob parameters if the BPO fork is undefined.

I.e. for Osaka, BPO1 and BPO2, which are defined `src/ethereum/forks` (BPO1 and BPO2 not yet but eventually), we pass directly the name and skip the BPO parameters. But for example, if we want to test BPO3, which is still undefined in the specs, then in that case we do pass the parameters along with the Osaka fork name.

We have to guarantee that the BPO parameter values match in the specs and the tests but I think that's a completely valid requirement.

marioevz (commented):
Thanks, just one comment that I think should skip having to generate temporary forks if the BPO values are unchanged.

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
SamWilsn: So something like pre-populating the fork cache with the forks that exist, or possibly just checking if the temporary fork's values are the same as the template fork's?

marioevz:
Found the issue on why we are currently failing to fill tests for BPO1:

We are calling `ForkCache.get` with `blob_schedule_max` set but `max_blob_gas_per_block` set to `None`, so the temporary fork updates the `vm.gas.BLOB_SCHEDULE_MAX` but `fork.MAX_BLOB_GAS_PER_BLOCK` remains intact.

So we might need to calculate `max_blob_gas_per_block` on the fly if `blob_schedule_max` is passed (which depends on `vm.gas.GAS_PER_BLOB` which can also be modified so it might be not as straightforward).

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
marioevz: Yes exactly, that would be even better, if the template already matches the values, just return the template.

spencer-tb:
> Found the issue on why we are currently failing to fill tests for BPO1:

Are we failing to fill? It should be running in CI with the tox change I think

SamWilsn:
This PR changes `MAX_BLOB_GAS_PER_BLOCK` to be calculated dynamically:

https://github.com/ethereum/execution-specs/pull/1789/files#diff-819115894be2a859a2ee56a9c942931f9fe0ac1820cc6b8c8d8686079fc988e8R101

gurukamath (approved):

marioevz (approved):
LGTM, thanks for this needed change :D

Only downside is that the startup time for fill went up since we do the copy of the template regardless of BPO values:
```
===================================================== 48 passed, 10 warnings in 116.29s (0:01:56) ======================================================
```

from:
```
=========================================================== 48 passed, 10 warnings in 5.05s ============================================================
```

Added a P-High label to #1864 based on this.

PR #1790, opened 2025-11-14
enhance(test-benchmark): use config file for fixed opcode count scenarios

spencer-tb:
## 🗒️ Description

This PR adds a CLI tool `benchmark_parser` to automatically scan benchmark tests and generate a configuration file `.fixed_opcode_counts.json` for the `--fixed-opcode-count` feature from #1747.

**Key Changes**

- New CLI tool: `uv run benchmark_parser`
  - Uses Python AST to scan `tests/benchmark/` for tests with `@pytest.mark.repricing` marker
  - Extracts opcode patterns from `@pytest.mark.parametrize` decorators
  - Generates `.fixed_opcode_counts.json` at repo root with opcode counts mapping
  - Supports `--check` mode for CI validation: `uv run benchmark_parser --check`
- Config file format: `.fixed_opcode_counts.json`
  - Gitignored (user-local configuration)
  - All patterns default to [1] (1K opcodes)
  - Users can customize counts per pattern, [1, 10, 100] for 1K, 10K, 100K, by manually editing the file
  - Custom counts are preserved when re-running the parser.
- Help text improvements:
  - Added benchmark options to fill --fill-help and execute remote --execute-remote-help
  - Simplified help text with examples for --gas-benchmark-values and --fixed-opcode-count
- Test updates:
  - Renamed op parameters to opcode in `test_arithmetic.py` for consistency

**Usage**

1. Generate/update config (first time or after benchmark test changes): `uv run benchmark_parser`
2. Customize counts by editing `.fixed_opcode_counts.json`:
```json
{
  "scenario_configs": {
    "test_codecopy.*": [
      1
    ],
    ...
}
```
3. Run with configured opcode counts:
```bash
# Fill fixtures (useful for fast one shot checks if count is 1)
uv run fill --fixed-opcode-count --fork Prague -m repricing tests/benchmark

# Execute on remote RPC
uv run execute remote --fixed-opcode-count --fork Prague -m repricing tests/benchmark --rpc-seed-key <key> --rpc-endpoint <url> --chain-id <id>
```

Fill works correctly and I tested execute remote on Hoodi with 1K opcode count set, the latest txs: https://hoodi.etherscan.io/address/0x83fd666bfb2b345f932c3e4e04b6d85e5ed3568d

**Future Items**

- Add CI for fill/execute with `--fixed-opcode-count` after generating the config file.
- Verify `--fixed-opcode-count` with `debug_traceTransaction` using `execute hive`.
- Add documentation & framework tests.

## 🔗 Related Issues or PRs

#1747 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -183,54 +183,53 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
             )
 
     if "fixed_opcode_count" in metafunc.fixturenames:
-        # Only parametrize if test has repricing marker
-        has_repricing = (
-            metafunc.definition.get_closest_marker("repricing") is not None
-        )
-        if has_repricing:
-            opcode_counts_to_use = None
+        # Parametrize for any benchmark test when --fixed-opcode-count is provided
+        if fixed_opcode_counts_cli is None:
+            return
 
-            if fixed_opcode_counts_cli and fixed_opcode_counts_cli != "":
-                # CLI flag with value takes precedence
-                opcode_counts_to_use = [
-                    int(x.strip()) for x in fixed_opcode_counts_cli.split(",")
-                ]
-            elif fixed_opcode_counts_cli == "":
-                # Flag provided without value - load from config file
-                # Check if config data was already loaded in pytest_collection_modifyitems
-                config_data = getattr(
-                    metafunc.config, "_opcode_counts_config", None
-                )
+        opcode_counts_to_use = None
 
-                # If not loaded yet (pytest_generate_tests runs first), load it now
-                if config_data is None:
-                    config_data = load_opcode_counts_config(metafunc.config)
-                    if config_data:
-                        metafunc.config._opcode_counts_config = config_data  # type: ignore[attr-defined]
+        if fixed_opcode_counts_cli:
+            # CLI flag with value takes precedence
+            opcode_counts_to_use = [
+                int(x.strip()) for x in fixed_opcode_counts_cli.split(",")
+            ]
+        else:
+            # Flag provided without value - load from config file
+            # Check if config data was already loaded in pytest_collection_modifyitems
+            config_data = getattr(
+                metafunc.config, "_opcode_counts_config", None
+            )
 
+            # If not loaded yet (pytest_generate_tests runs first), load it now
+            if config_data is None:
+                config_data = load_opcode_counts_config(metafunc.config)
                 if config_data:
-                    # Look up opcode counts using regex pattern matching
-                    test_name = metafunc.function.__name__
-                    opcode_counts_to_use = get_opcode_counts_for_test(
-                        test_name,
-                        config_data.get("scenario_configs", {}),
-                        config_data.get("default_counts", [1]),
-                    )
+                    metafunc.config._opcode_counts_config = config_data  # type: ignore[attr-defined]
 
-            # Parametrize if we have counts to use
-            if opcode_counts_to_use:
-                opcode_count_parameters = [
-                    pytest.param(
-                        opcode_count,
-                        id=f"opcount_{opcode_count}K",
-                    )
-                    for opcode_count in opcode_counts_to_use
-                ]
-                metafunc.parametrize(
-                    "fixed_opcode_count",
-                    opcode_count_parameters,
-                    scope="function",
+            if config_data:
+                # Look up opcode counts using regex pattern matching
+                test_name = metafunc.function.__name__
+                opcode_counts_to_use = get_opcode_counts_for_test(
+                    test_name,
+                    config_data.get("scenario_configs", {}),
+                    config_data.get("default_counts", [1]),
                 )
+
+        # Parametrize if we have counts to use
+        if opcode_counts_to_use:
+            opcode_count_parameters = [
+                pytest.param(
+                    opcode_count,
+                    id=f"opcount_{opcode_count}K",
+                )
+                for opcode_count in opcode_counts_to_use
+            ]
+            metafunc.parametrize(
+                "fixed_opcode_count",
+                opcode_count_parameters,
+                scope="function",
+            )
 
 
 @pytest.fixture(scope="function")
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -109,14 +109,15 @@ def get_opcode_counts_for_test(
 def pytest_collection_modifyitems(
     config: pytest.Config, items: list[pytest.Item]
 ) -> None:
-    """Remove non-repricing tests when `--fixed-opcode-count` is specified."""
+    """Filter tests based on repricing marker when benchmark options are used."""
+    gas_benchmark_value = config.getoption("gas_benchmark_value")
     fixed_opcode_count = config.getoption("fixed_opcode_count")
 
-    # Only filter if --fixed-opcode-count flag was provided
-    if fixed_opcode_count is None:
+    # Only filter if either benchmark option is provided
+    if not gas_benchmark_value and fixed_opcode_count is None:
         return
 
-    # Load config data if flag provided without value
+    # Load config data if --fixed-opcode-count flag provided without value
     if fixed_opcode_count == "":
         config_data = load_opcode_counts_config(config)
         if config_data:
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -13,19 +13,23 @@ from .execute_fill import OpMode
 
 
 def pytest_addoption(parser: pytest.Parser) -> None:
-    """Add command line options for gas benchmark values."""
-    evm_group = parser.getgroup(
-        "evm", "Arguments defining evm executable behavior"
+    """Add command line options for benchmark tests."""
+    benchmark_group = parser.getgroup(
+        "benchmarking", "Arguments for benchmark test execution"
     )
-    evm_group.addoption(
+    benchmark_group.addoption(
         "--gas-benchmark-values",
         action="store",
         dest="gas_benchmark_value",
         type=str,
         default=None,
-        help="Specify gas benchmark values for tests as a comma-separated list.",
+        help=(
+            "Gas limits (in millions) for benchmark tests. "
+            "Example: '100,500' runs tests with 100M and 500M gas. "
+            "Cannot be used with --fixed-opcode-count."
+        ),
     )
-    evm_group.addoption(
+    benchmark_group.addoption(
         "--fixed-opcode-count",
         action="store",
         dest="fixed_opcode_count",
@@ -34,8 +38,10 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         nargs="?",
         const="",
         help=(
-            "Specify fixed opcode counts (in thousands) for benchmark tests as a comma-separated list. "
-            "If provided without a value, uses defaults from tests/benchmark/configs/fixed_opcode_counts.py."
+            "Opcode counts (in thousands) for benchmark tests. "
+            "Example: '1,10,100' runs tests with 1K, 10K, 100K opcodes. "
+            "Without value, uses .fixed_opcode_counts.json config. "
+            "Cannot be used with --gas-benchmark-values."
         ),
     )
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -1,5 +1,10 @@
 """The module contains the pytest hooks for the gas benchmark values."""
 
+import json
+import re
+from pathlib import Path
+from typing import Any
+
 import pytest
 
 from execution_testing.test_types import Environment, EnvironmentDefaults
@@ -26,7 +31,12 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         dest="fixed_opcode_count",
         type=str,
         default=None,
-        help="Specify fixed opcode counts (in thousands) for benchmark tests as a comma-separated list.",
+        nargs="?",
+        const="",
+        help=(
+            "Specify fixed opcode counts (in thousands) for benchmark tests as a comma-separated list. "
+            "If provided without a value, uses defaults from tests/benchmark/configs/fixed_opcode_counts.py."
+        ),
     )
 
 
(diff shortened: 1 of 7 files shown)


tests/benchmark/configs/fixed_opcode_counts.py
```diff
@@ -0,0 +1,212 @@
...
+"""
+
+import re
+from typing import List
+
+DEFAULT_OPCODE_COUNTS = [1]
+
+# Scenario configurations using test_name.*OPCODE.* patterns
+# Keys are regex patterns checked in order; first match wins
+SCENARIO_CONFIGS = {
```
LouisTsai-Csie: I suggest to move this mapping into a separate file so that we can maintain it automatically with the script.

I have this [prototype](https://gist.github.com/LouisTsai-Csie/11c16b052c975e65eff918116226d7b9), here’s a quick summary of the script that helps analyze differences between the old and new mappings:

1. It uses Python’s AST to analyze benchmark tests.
2. It filters only tests that use the `benchmark_test` wrapper with a `code_generator` attribute, since we currently only support fixed-opcode-count feature in code generator.
3. Based on current regex conventions, identifies the patterns:
  - If there is no opcode parameterization, the regex follows: `test_<opcode/precompile>_<modifier>.*` (e.g., `test_selfbalance`, `test_codesize`).
  - If the test is parametrized by opcode, the regex includes the opcode:
`test_name.*<opcode>.*`, e.g., `test_ext_account_query_warm.*<opcode>.*` where `<opcode>` is `BALANCE`, `EXTCODESIZE`, etc.
  - Note: Some parametrization names should be refactored to `opcode`. For example, in `test_mod_arithmetic`, `op` should be renamed to `opcode` so the script can detect it correctly.
4. The script supports several modes:
  - `check` mode: Detect new/missing entries without modifying files
  - `dry-run`: Show the generated configuration without applying changes
  - `update`: Update the config file with new entries
  - `no-filter`: Include all benchmark tests (not needed and can be removed)

However, I suggest the following structure, and update the script accordingly:
- the opcode-count mapping under `tests/benchmark/configs/fixed_opcode_counts.py`
- the parser and mapping update script under `tests/benchmark/configs/parser.py`

Regarding `--fixed-opcode-count`, I agree with the behavior this PR already had:
- If no values are passed, use the defaults from the mapping.
- If values are passed, they override the configuration for flexibility.

What do you think? Happy to discuss with different approach.

tests/benchmark/configs/fixed_opcode_counts.py
```diff
@@ -0,0 +1,212 @@
...
+    "test_tstore.*": [1, 10],
+    # SYSTEM OPERATIONS
+    "test_return_revert.*RETURN.*": [1, 10],
+    "test_return_revert.*REVERT.*": [1, 10],
+    # TRANSACTION CONTEXT OPERATIONS
+    "test_blobhash.*": [1, 10],
+}
+
+
+def get_opcode_counts_for_scenario(
```
LouisTsai-Csie: I propose to separate `SCENARIO_CONFIGS` variable and `get_opcode_counts_for_scenario` function in different files.

tests/benchmark/configs/fixed_opcode_counts.py
LouisTsai-Csie: Based on our discussion yesterday, add this script into CI to make sure no cases missing.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -110,16 +163,52 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
             metafunc.definition.get_closest_marker("repricing") is not None
         )
         if has_repricing:
-            if fixed_opcode_counts:
-                opcode_counts = [
-                    int(x.strip()) for x in fixed_opcode_counts.split(",")
+            opcode_counts_to_use = None
```
LouisTsai-Csie: This contains a few logic issues, but they will be resolved in my upcoming PR.

LouisTsai-Csie (changes requested):
Thanks a lot for this! I left some suggestions, but I’m happy to discuss further. I’ll share this with Kamil to confirm it aligns with their needs.

tests/benchmark/configs/parser.py
```diff
@@ -0,0 +1,487 @@
+#!/usr/bin/env python3
```
LouisTsai-Csie: Do we need this? I only found two other instances in the repo:
- `docs/scripts/generate_fill_help.py`
- `scripts/download_geth_linux.py`

I removed this and it seems still working

tests/benchmark/configs/parser.py
```diff
@@ -0,0 +1,487 @@
...
+"""
+Parser to analyze benchmark tests and maintain the opcode counts mapping.
+
+This script uses Python's AST to analyze benchmark tests and generate/update
+the SCENARIO_CONFIGS mapping in fixed_opcode_counts.py.
+
+Usage:
+    uv run python parser.py --check    # Check for new/missing entries (CI)
+    uv run python parser.py --dry-run  # Show config without changes
+    uv run python parser.py --update   # Update fixed_opcode_counts.py
```
LouisTsai-Csie: I am not sure but this seems not working for me. I still need to specify the path like:

`uv run tests/benchmark/configs/parser.py --check`

I also wonder whether we should place this script under `packages/testing/src/execution_testing/cli/`, since most related scripts, such as `extract_config` and `diff_opcode_count`, are located there. wdyt?

tests/benchmark/configs/parser.py
```diff
@@ -0,0 +1,487 @@
...
+    parser.add_argument(
+        "--update",
+        action="store_true",
+        help="Update fixed_opcode_counts.py with new entries",
+    )
+    parser.add_argument(
+        "--no-filter",
+        action="store_true",
+        help="Include all benchmark tests, not just those with code_generator",
+    )
```
LouisTsai-Csie: My personal view: we could remove this and related logic, wdyt?

Looking into the fixed-opcode-count logic, only those test that are (1) `BenchmarkTestFiller` and (2) `code_generator` could support this mode.

This constraint is added in PR #1810 with [comment](https://github.com/ethereum/execution-specs/pull/1810/files#r2567195068).

tests/benchmark/configs/parser.py
```diff
@@ -0,0 +1,487 @@
...
+
+        # Check if function has benchmark_test parameter
+        if not self._has_benchmark_test_param(node):
+            self.generic_visit(node)
+            return
+
+        # Optional: Filter for code generator usage
+        if self.filter_code_gen and not self._uses_code_generator(node):
+            self.generic_visit(node)
+            return
```
LouisTsai-Csie: I’m not sure whether we need `self.generic_visit(node)` in each of the if–else branches. My understanding is that once a condition matches, it will recursively visit the nested function anyway.

This seems relevant only for cases like:
```python
def test_function1(...):

    def test_function2(...):
        ...
```

If so, maybe we could remove the logic here:


```suggestion
        if not node.name.startswith("test_"):
            return

        # Check if function has benchmark_test parameter
        if not self._has_benchmark_test_param(node):
            return

        # Optional: Filter for code generator usage
        if self.filter_code_gen and not self._uses_code_generator(node):
            return
```

tests/benchmark/configs/parser.py
```diff
@@ -0,0 +1,487 @@
...
+            # Get parameter names (first arg)
+            param_names = decorator.args[0]
+            if isinstance(param_names, ast.Constant):
+                param_str = str(param_names.value).lower()
+            else:
+                continue
+
+            # Check if "opcode" or "op" is in parameter names
+            if "opcode" not in param_str and param_str not in ("op", "op,"):
+                continue
```
LouisTsai-Csie: I would suggest only use `opcode` as the keyword, instead of three combinations. Here we need a detailed documentation!

This would requires us to refactor the parametrization for the following tests:
- `test_arithmetic.py`: `test_mod` (`op`), `test_mod_arithmetic` (`op`),

tests/benchmark/configs/parser.py
```diff
@@ -0,0 +1,487 @@
...
+                elif isinstance(first_arg, ast.Tuple) and first_arg.elts:
+                    first_elem = first_arg.elts[0]
+                    if isinstance(first_elem, ast.Attribute):
+                        return first_elem.attr
+
+        # Case 3: Tuple like (Op.ADD, args)
+        if isinstance(node, ast.Tuple) and node.elts:
+            first_elem = node.elts[0]
+            if isinstance(first_elem, ast.Attribute):
+                return first_elem.attr
```
LouisTsai-Csie: More explanation for the logic here! Suggestion:

```
Case 1: Direct opcode reference
Example: @pytest.mark.parametrize("opcode", [Op.ADD, Op.MUL])
Result: ADD, MUL

Case 2a: Pytest param with tuple
Example: @pytest.mark.parametrize("opcode,args", [pytest.param((Op.ADD, 123))])
Result: ADD

Case 2b: Plain tuples
Example: @pytest.mark.parametrize("opcode,args", [(Op.ADD, 123), (Op.MUL, 456)])
Result: ADD, MUL
```

And here we assume the parametrization is always "opcode", this should be documented somewhere. 

---

IIUC, this would not be recognized
```python
@pytest.mark.parametrize("opcode_args,opcode", [pytest.param((123, Op.ADD))])
```

tests/benchmark/configs/parser.py
```diff
@@ -0,0 +1,487 @@
...
+        "CONTROL FLOW OPERATIONS": [],
+        "HASHING OPERATIONS": [],
+        "LOGGING OPERATIONS": [],
+        "MEMORY OPERATIONS": [],
+        "STACK OPERATIONS": [],
+        "STORAGE OPERATIONS": [],
+        "SYSTEM OPERATIONS": [],
+        "TRANSACTION CONTEXT OPERATIONS": [],
+        "OTHER": [],
+    }
```
LouisTsai-Csie: The current keyword-based solution requires manually update the mapping, but i notice a pattern here:

```python
"ACCOUNT QUERY OPERATIONS": [
      "selfbalance",
      "codesize",
      "ext_account",
      "BALANCE",
      "EXTCODE",
  ],
```

The key (`ACCOUNT QUERY OPERATIONS`) here could be derived from the file name:
`test_account_query_operations.py`. So when we loop over the benchmark test, we could also record its parent file (in `scan_benchmark_tests` function).

And then update the categorize_patterns function like this:

```python
    ...

    categories: dict[str, list[str]] = {}

    for pattern in config.keys():
        if pattern not in pattern_sources:
            # Fallback for patterns without source info
            category = "OTHER"
        else:
            source_file = pattern_sources[pattern]
            # Extract test file name: test_arithmetic.py -> arithmetic
            file_name = source_file.stem  # Gets filename without extension
            if file_name.startswith("test_"):
                category_name = file_name[5:]  # Remove "test_" prefix
                # arithmetic -> ARITHMETIC OPERATIONS
                category = (
                    f"{category_name.upper().replace('_', ' ')} OPERATIONS"
                )
            else:
                category = "OTHER"

        if category not in categories:
            categories[category] = []
        categories[category].append(pattern)

    # Sort patterns within each category and sort categories alphabetically
    return {k: sorted(v) for k, v in sorted(categories.items())}

```

LouisTsai-Csie: Based on this approach, we could remove `category_keywords`

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -110,16 +180,43 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
             metafunc.definition.get_closest_marker("repricing") is not None
         )
         if has_repricing:
```
LouisTsai-Csie: We might need a rebase as i remember i remove this logic

LouisTsai-Csie (changes requested):
Add some suggestion for the parser! Thanks

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -41,16 +57,71 @@ def pytest_configure(config: pytest.Config) -> None:
...
+    # Only filter if --fixed-opcode-count flag was provided
+    if fixed_opcode_count is None:
         return
 
+    # Load config data if flag provided without value
+    if fixed_opcode_count == "":
+        config_data = load_opcode_counts_config(config)
+        if config_data:
+            config._opcode_counts_config = config_data  # type: ignore[attr-defined]
+
```
LouisTsai-Csie, line 135:
```suggestion
def pytest_collection_modifyitems(
config: pytest.Config, items: list[pytest.Item]
) -> None:
    """Filter tests based on repricing marker"""
fixed_opcode_count = config.getoption("fixed_opcode_count")
gas_benchmark_value = config.getoption("gas_benchmark_value")

    if not gas_benchmark_value and not fixed_opcode_count:
return
```

Sorry, I thought this logic would be overridden once my PR was merged, but it seems still not working now. The filter logic should work as follows:
- There are two modes: `--gas-benchmark-values` and `--fixed-opcode-count`. Either mode will run the full benchmark test suite.
- We can specify `-m repricing` to restrict the benchmark suite to tests marked with the `repricing` marker.

With this in mind, the use of `-m repricing` below becomes limited to the `--fixed-opcode-count` mode only, while the `--gas-benchmark-values` mode cannot make use of this filter:
```python
# Only filter if --fixed-opcode-count flag was provided
if fixed_opcode_count is None:
    return
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -111,22 +182,54 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
-            ]
-            opcode_count_parameters = [
-                pytest.param(
-                    opcode_count,
-                    id=f"opcount_{opcode_count}K",
+        # Only parametrize if test has repricing marker
+        has_repricing = (
+            metafunc.definition.get_closest_marker("repricing") is not None
+        )
+        if has_repricing:
```
LouisTsai-Csie: Same logic as previous comment, we should not limit the fixed-opcode-count feature to a subset of benchmark tests, if user does not specify `-m repricing`

https://github.com/ethereum/execution-specs/blob/9523e75061bc9f989f0867e34f28364c719a3b07/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py#L113-L129

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -8,25 +13,36 @@
...
-    evm_group.addoption(
+    benchmark_group.addoption(
         "--gas-benchmark-values",
         action="store",
         dest="gas_benchmark_value",
         type=str,
         default=None,
-        help="Specify gas benchmark values for tests as a comma-separated list.",
+        help=(
+            "Gas limits (in millions) for benchmark tests. "
```
LouisTsai-Csie, line 28: This change would make this CI workflow fail:

https://github.com/ethereum/execution-specs/blob/9523e75061bc9f989f0867e34f28364c719a3b07/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_benchmarking.py#L63-L77

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -111,22 +182,54 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
-                    opcode_count,
-                    id=f"opcount_{opcode_count}K",
+        # Only parametrize if test has repricing marker
+        has_repricing = (
+            metafunc.definition.get_closest_marker("repricing") is not None
+        )
+        if has_repricing:
+            opcode_counts_to_use = None
+
+            if fixed_opcode_counts_cli and fixed_opcode_counts_cli != "":
```
LouisTsai-Csie: nit: i think this is the same logic

```suggestion
            if fixed_opcode_counts_cli:
```

LouisTsai-Csie (changes requested):
Thanks!! Only a small logic adjustment is needed.

Here is an example using this test:
```python
@pytest.mark.repricing(contract_balance=0)
@pytest.mark.parametrize("contract_balance", [0, 1])
def test_selfbalance(
    contract_balance: int,
) -> None:
    """Benchmark SELFBALANCE instruction."""
    benchmark_test(
        code_generator=ExtCallGenerator(
            attack_block=Op.SELFBALANCE,
            contract_balance=contract_balance,
        ),
    )
```

When running in pure `--fixed-opcode-count` or `--gas-benchmark-values` mode, both parameters should be executed (`contract_balance` = 0 and 1).
This results in 4 tests being run (including both blockchain test and blockchain engine test combinations).

**Example run (without repricing marker)**
The full benchmark test suite should run, producing 4 tests:
```python
fill -v tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance --gas-benchmark-values 10 -m benchmark --clean

fill -v tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance --fixed-opcode-count 1 -m benchmark --clean
```

**With the repricing marker**
Only the cases with contract_balance = 0 should run, so it should produce 2 tests:
```python
fill -v tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance --gas-benchmark-values 10 -m repricing --clean

fill -v tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance --fixed-opcode-count 1 -m repricing --clean
```

---
Edit:

Regarding the check: if no marker is present, the test should still run when using the `-m benchmark` flag, but it should be ignored when using the `-m repricing` flag.

Still take `test_selfbalance` as example, if you remove the repricing marker, both of this should still be able to run

```python
fill -v tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance --gas-benchmark-values 10 -m benchmark --clean

fill -v tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance --fixed-opcode-count 1 -m benchmark --clean
```

spencer-tb:
Thanks for this! I've tried to address all you comments. Let me know if I missed anything:

 **1. Fix repricing filter to work with both benchmark options** ([`0ca5109`](https://github.com/ethereum/execution-specs/pull/1790/commits/0ca5109c0))
  - `pytest_collection_modifyitems` now checks for both `--gas-benchmark-values` and `--fixed-opcode-count` before applying the `-m repricing` filter.
  - Added a test to verify repricing filter works with both options.

  **2. Allow fixed-opcode-count for all benchmark tests** ([`a18f800`](https://github.com/ethereum/execution-specs/pull/1790/commits/a18f800b7))
  - Removed the `has_repricing` check in `pytest_generate_tests` so `--fixed-opcode-count` works on all benchmark tests, not just repricing marked ones.

  **3. Warn when config file missing** ([`a20f2b7`](https://github.com/ethereum/execution-specs/pull/1790/commits/a20f2b783))
  - Added a `UserWarning` when `--fixed-opcode-count` is provided without a value but `.fixed_opcode_counts.json` doesn't exist.

  **4. Update test to match new help text** ([`dff1a46`](https://github.com/ethereum/execution-specs/pull/1790/commits/dff1a46a4))
  - Fixed the failing CI test by updating the expected help text string.

  **5. Remove unnecessary generic_visit call in parser** ([`33fa5ab`](https://github.com/ethereum/execution-specs/pull/1790/commits/33fa5ab4e))
  - Removed `self.generic_visit(node)` since we have early returns and don't use nested test functions.

LouisTsai-Csie (approved):
LGTM! Huge thanks for the effort!

PR #1803, opened 2025-11-20
fix(test-types): format `Transaction` values as readable hex-strings instead of raw bytes (#1803)

felix314159:
## 🗒️ Description
```
a = Transaction()
print(f"{a}")
```
Before this PR the resulting log is unreadable, you don't even see who is sending to who. With this PR you get readable addresses 0x...

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture
Description

### DIFF
--- a/packages/testing/src/execution_testing/cli/gentest/source_code_generator.py
+++ b/packages/testing/src/execution_testing/cli/gentest/source_code_generator.py
@@ -89,7 +89,6 @@ def format_code(code: str) -> str:
                 str(formatter_path),
                 "format",
                 str(input_file_path),
-                "--quiet",
                 "--no-cache",
                 "--config",
                 str(config_path),
@@ -100,7 +99,8 @@ def format_code(code: str) -> str:
         if result.returncode != 0:
             raise Exception(
                 f"Error formatting code using formatter '{formatter_path}': "
-                f"{result.stderr}"
+                f"returncode={result.returncode}, stdout={result.stdout!r}, "
+                f"stderr={result.stderr!r}"
             )
 
         # Return the formatted source code
--- a/packages/testing/src/execution_testing/test_types/transaction_types.py
+++ b/packages/testing/src/execution_testing/test_types/transaction_types.py
@@ -848,8 +848,8 @@ class Transaction(
         """
         Format a field value for string representation.
 
-        Uses hex encoding for any value that supports it
-        (Address, Bytes, Hash, HexNumber, etc).
+        Uses decimal for numeric values (int, HexNumber, etc.) and
+        hex encoding for Address, Bytes, Hash, etc.
         """
         if value is None:
             return "None"
--- a/packages/testing/src/execution_testing/test_types/transaction_types.py
+++ b/packages/testing/src/execution_testing/test_types/transaction_types.py
@@ -1,7 +1,6 @@
 """Transaction-related types for Ethereum tests."""
 
 from dataclasses import dataclass
-import decimal
 from enum import IntEnum
 from functools import cached_property
 import numbers
--- a/packages/testing/src/execution_testing/test_types/transaction_types.py
+++ b/packages/testing/src/execution_testing/test_types/transaction_types.py
@@ -856,11 +856,6 @@ class Transaction(
 
         # fields like 'value' should be shown as decimal number
         if isinstance(value, numbers.Number):
-            if isinstance(value, decimal.Decimal):
-                # Convert to string to avoid scientific notation (1E-9)
-                # while removing unnecessary trailing zeros (100.000000 -> 100)
-                return "{:f}".format(value).rstrip("0").rstrip(".")
-
             # Force decimal representation for int subclasses like HexNumber
             if isinstance(value, int):
                 return str(int(value))
--- a/packages/testing/src/execution_testing/test_types/transaction_types.py
+++ b/packages/testing/src/execution_testing/test_types/transaction_types.py
@@ -852,7 +852,7 @@ class Transaction(
         if value is None:
             return "None"
         elif hasattr(value, "hex"):
-            return value.hex()
+            return f'"{value.hex()}"'
         else:
             return repr(value)
 
--- a/packages/testing/src/execution_testing/test_types/transaction_types.py
+++ b/packages/testing/src/execution_testing/test_types/transaction_types.py
@@ -861,6 +861,10 @@ class Transaction(
                 # while removing unnecessary trailing zeros (100.000000 -> 100)
                 return "{:f}".format(value).rstrip("0").rstrip(".")
 
+            # Force decimal representation for int subclasses like HexNumber
+            if isinstance(value, int):
+                return str(int(value))
+
             return str(value)
 
         # fields like 'to' should be shown as hex string
--- a/packages/testing/src/execution_testing/test_types/transaction_types.py
+++ b/packages/testing/src/execution_testing/test_types/transaction_types.py
@@ -842,6 +842,37 @@ class Transaction(
         else:
             return gas_price * gas_limit + self.value
 
+    def _format_field_value(self, value: Any) -> str:
+        """
+        Format a field value for string representation.
+
+        Uses hex encoding for any value that supports it
+        (Address, Bytes, Hash, HexNumber, etc).
+        """
+        if value is None:
+            return "None"
+        elif hasattr(value, "hex"):
+            return value.hex()
+        else:
+            return repr(value)
+
+    def __repr__(self) -> str:
+        """
+        Return string representation with hex-encoded values for
+        applicable fields.
+        """
+        field_strs = []
+        for field_name in self.__class__.model_fields:
+            value = getattr(self, field_name)
+            formatted_value = self._format_field_value(value)
+            field_strs.append(f"{field_name}={formatted_value}")
+
+        return f"{self.__class__.__name__}({', '.join(field_strs)})"
+
+    def __str__(self) -> str:
+        """Return the repr string representation."""
+        return self.__repr__()
+
 
 class NetworkWrappedTransaction(CamelModel, RLPSerializable):
     """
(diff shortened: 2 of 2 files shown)


packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -730,6 +725,42 @@ def metadata_string(self) -> str | None:
             return None
         return self.metadata.to_json()
 
+    def _format_field_value(self, field_name: str, value: Any) -> str:
+        """
+        Format a field value for string representation.
+
+        Special handling for to, data, and sender fields to show hex encoding.
+        """
+        if field_name in ("to", "data", "sender"):
```
fselmo: What do you think about not being restrictive to specific fields here? Something like:

```py
def _format_field_value(self, value: Any) -> str:
    """
    Format a field value for string representation.
    
    Uses hex encoding for any value that supports it
    (Address, Bytes, Hash, HexNumber, etc).
    """
    if value is None:
       return "None"
    elif hasattr(value, "hex"):
       return value.hex()
    else:
       return repr(value)
```

I did a quick test and this works well. If we go with this, we just need to update the docstring descriptions in `__repr__()` and `__str__()`.

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -730,6 +725,42 @@ def metadata_string(self) -> str | None:
...
+        Return string representation with hex-encoded values for to, data, and
+        sender fields.
+        """
+        field_strs = []
+        for field_name in self.model_fields:
+            value = getattr(self, field_name)
+            formatted_value = self._format_field_value(field_name, value)
+            field_strs.append(f"{field_name}={formatted_value}")
+
+        return " ".join(field_strs)
```
fselmo: nit: is there a good reason to deviate from the usual pattern for `__repr__()` here with `ClassName(...)`?

```suggestion
        return f"{self.__class__.__name__}({', '.join(field_strs)})"
```

fselmo (commented):
I support better readability 👌🏼. Just a minor suggestion for generalizing it to other fields with hex support. Can do another pass after this is rebased.

packages/testing/src/execution_testing/test_types/transaction_types.py
felix314159: Fine with me. If we ever add sth like a `float` field to `Transaction` we can add an exception, for some datatypes it looks ugly e.g.:
```
0.1.hex()
'0x1.999999999999ap-4'
```

packages/testing/src/execution_testing/test_types/transaction_types.py
felix314159: Thanks I prefer your approach, it makes object boundaries more explicit. E.g. we might log an object whose fields include other objects and then it's nice that there is clear separation between the parent's fields and the nested object's representation. e.g.
```
child=a=1 b=2 status='ok'
```
vs
```
ParentWrapped(child=ChildWrapped(a=1, b=2), status='ok')
```
Thanks good change, I actually prefer being explicit here cuz it helps sometimes when debugging

felix314159:
TIL:
git commit -m "abc \`def\` ghi" shows up as "abc ghi"
but
git commit -m 'abc \`def\` ghi' shows up as "abc `def` ghi`

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -842,6 +842,37 @@ def signer_minimum_balance(self, *, fork: Fork) -> int:
...
+        """
+        Format a field value for string representation.
+
+        Uses hex encoding for any value that supports it
+        (Address, Bytes, Hash, HexNumber, etc).
+        """
+        if value is None:
+            return "None"
+        elif hasattr(value, "hex"):
+            return value.hex()
```
fselmo: Ah my fault on the currently failing tests. I think my code snippet was a bit off. The `__repr__` is generating an invalid hex when `data=b""` (empty bytes) because `Bytes(b"").hex()` returns `"0x"` which is an invalid hex (it needs to be something like `0x0` with one byte at least).

We can change this to the string representation directly and it should be OK:

```suggestion
            return f'"{value.hex()}"'
```

danceratopz:
@felix314159 just rebased on main to get the potential fix from:
- #2009

packages/testing/src/execution_testing/test_types/transaction_types.py
danceratopz: Yes, thanks @fselmo! This was necessary to fix the remaining failing tests!

packages/testing/src/execution_testing/cli/gentest/source_code_generator.py
```diff
@@ -100,7 +99,8 @@ def format_code(code: str) -> str:
         if result.returncode != 0:
             raise Exception(
                 f"Error formatting code using formatter '{formatter_path}': "
-                f"{result.stderr}"
+                f"returncode={result.returncode}, stdout={result.stdout!r}, "
```
danceratopz, line 102: Unrelated change, just an improvement to help debug the issues we were seeing in CI.

danceratopz (commented):
[capture_gentest.py](https://github.com/user-attachments/files/24587394/capture_gentest.py)

Thanks for adding this @felix314159! The CI fails were not due to a race condition while executing the gentest tests via `CliRunner`. `gentest` formats the generated code with `ruff`, which was failing because it generated invalid Python code until the fix from @fselmo was applied (`f'"{value.hex()}"'` instead of `value.hex()`, to ensure `0x0`, not `0x` was returned).

After seeing this fail (which is now fixed), I thought wed better check what else has changed with the output of `gentest` and these changes do greatly improve situation for `Address` objects (which was a long time coming, so thanks!), but actually regresses the situation for numeric values, such as `gas_price`. The following summarizes the situation.

The `_format_field_value` change improves readability for addresses and bytes data, but reduces readability for numeric fields.

## Examples from gentest output

**Addresses/Data (improved):**
```python
# Before: raw bytes (hard to read)
to=b'\xc6\xd9g\x86G\x7f\x82I\x1b\xfe\xad\x8f\x00\xb8)F\x88\xf7z\xbc'
data=b''

# After: hex strings (readable)
to="0xc6d96786477f82491bfead8f00b8294688f77abc"
data="0x"
```

**Numeric fields (regressed):**
```python
# Before: decimal (clear)
gas_price=10000000000  # obviously 10 gwei
value=668250000000000000
nonce=2

# After: hex strings (opaque)
gas_price="0x2540be400"  # what's this in gwei?
value="0x94619d59f27a000"
nonce="0x2"
```

## Suggestion

Consider making `_format_field_value` more nuanced:
- Addresses/Bytes → use hex string (current behavior, good)
- Integers/HexNumber → keep as decimal repr (better for readability)

This could be done by checking the type more specifically rather than just `hasattr(value, "hex")`.

I attached the script that Claude wrote to help compare the output.

felix314159:
Sorry I thought I had the SLOTNUM PR open when changing the target branch here :D

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -842,6 +844,48 @@ def signer_minimum_balance(self, *, fork: Fork) -> int:
...
+            return "None"
+
+        # fields like 'value' should be shown as decimal number
+        if isinstance(value, numbers.Number):
+            if isinstance(value, decimal.Decimal):
+                # Convert to string to avoid scientific notation (1E-9)
+                # while removing unnecessary trailing zeros (100.000000 -> 100)
+                return "{:f}".format(value).rstrip("0").rstrip(".")
+
+            return str(value)
```
danceratopz, line 862: I re-ran the script, we need this additional change to get decimal output, otherwise:
- nonce=0x2
- gas_price=0x2540be400
- value=0x94619d59f27a000

```suggestion
            # Force decimal representation for int subclasses like HexNumber
            if isinstance(value, int):
                return str(int(value))

            return str(value)
```

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -842,6 +844,48 @@ def signer_minimum_balance(self, *, fork: Fork) -> int:
         else:
             return gas_price * gas_limit + self.value
 
+    def _format_field_value(self, value: Any) -> str:
+        """
+        Format a field value for string representation.
+
+        Uses hex encoding for any value that supports it
+        (Address, Bytes, Hash, HexNumber, etc).
```
danceratopz:
```suggestion
        Uses decimal for numeric values (int, HexNumber, etc.) and
        hex encoding for Address, Bytes, Hash, etc.
```

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -842,6 +844,48 @@ def signer_minimum_balance(self, *, fork: Fork) -> int:
...
+        if value is None:
+            return "None"
+
+        # fields like 'value' should be shown as decimal number
+        if isinstance(value, numbers.Number):
+            if isinstance(value, decimal.Decimal):
+                # Convert to string to avoid scientific notation (1E-9)
+                # while removing unnecessary trailing zeros (100.000000 -> 100)
+                return "{:f}".format(value).rstrip("0").rstrip(".")
+
```
danceratopz: This seems overly defensive? We don't use Decimal anywhere in `package/testing/src/`.
```suggestion
```

danceratopz (commented):
Thanks @felix314159, here's few comments on the last change.

packages/testing/src/execution_testing/test_types/transaction_types.py
fselmo: Good catch!

danceratopz (approved):
Thanks for this improvement, @felix314159! LGTM!

fselmo (approved):
You've got my approval as well, thanks for this! Let's just make sure to `squash + merge` 😅.

Issue #1714, opened 2025-10-29
Mark `Osaka` mainnet tests


marioevz:
Use the `pytest.mark.mainnet` marker in new tests that are designated as a sanity check for when Osaka lands in mainnet.

Example of marked tests for Prague: https://github.com/ethereum/execution-specs/blob/forks/osaka/tests/prague/eip7702_set_code_tx/test_eip_mainnet.py

Required characteristics of these tests:
- Cheap: Consume a minimal amount of gas to test the EIP behavior.
- Single tx: No multi-tx tests preferably unless there's a good reason behind it.
- Succinct: Don't have to test all EIP's functionality, but they have to validate that the EIP got enabled correctly.
- Execute-compatible: All tests have to be compatible with the `execute` command.

If it's impossible to validate the EIP activation due to any of these constrains, we can skip validating that specific EIP, but we should clearly explain the reason.

Follow-up issue: Execute them on Fusaka activation.

felix314159:
Here is a list of my takes on live mainnet testing per EIP:

* EIP-7594: PeerDAS:
    * Send new cell-type blob after Osaka was activated 
    * Send cancun-style blob after the period where clients still accept it (TODO: find out how long that is for each client) to ensure it correctly is declined
    * Note: Creating this test in a live setting is currently a pain with our framework, will require more work than the others

---

* EIP-7823: MODEXP Size limit:
    * No test required, clients already have consumed our tests for invalid parameter lengths

---

* EIP-7825: Tx Gas Limit Cap:
    * Send one tx where 16,777,217 ($2^{24} + 1$) or more gas is used, it should be rejected (will not even be included in mempool, this test is 'free') 

---

* EIP-7883: ModExp Gas Cost Increase:
    * Make one call where any of the three params has exactly 32 bytes length (should still have old price, but it is the exact pricing edge case). Make another call where any of the three params has 33 bytes length (should have new, higher price), e.g.:
```
[...]

PR #1806, opened 2025-11-24
feat: add mainnet tests for osaka

felix314159:
## 🗒️ Description
Adds tests for modexp cost increase and introduction of secp256r1 precompile.
Blob tests still missing, standalone script still has to be integrated into framework for that.

Partially solves #1714 

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/osaka/eip7883_modexp_gas_increase/test_eip_mainnet.py
+++ b/tests/osaka/eip7883_modexp_gas_increase/test_eip_mainnet.py
@@ -1,0 +1,109 @@
+"""
+Mainnet marked tests for EIP-7883 ModExp gas cost increase.
+
+Tests for ModExp gas cost increase in
+[EIP-7883: ModExp Gas Cost Increase](https://eips.ethereum.org/EIPS/eip-7883).
+"""
+
+from typing import Dict
+
+import pytest
+from execution_testing import (
+    Alloc,
+    StateTestFiller,
+    Transaction,
+)
+
+from ...byzantium.eip198_modexp_precompile.helpers import ModExpInput
+from .spec import Spec, ref_spec_7883
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_7883.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7883.version
+
+pytestmark = [pytest.mark.valid_at("Osaka"), pytest.mark.mainnet]
+
+
+@pytest.mark.parametrize(
+    "modexp_input,modexp_expected",
+    [
+        pytest.param(
+            ModExpInput(
+                base="ff" * 31 + "fc",
+                exponent="02",
+                modulus="05",
+                declared_base_length=32,
+                declared_exponent_length=1,
+                declared_modulus_length=1,
+            ),
+            # expected result:
+            bytes.fromhex("04"),
+            id="32-bytes-long-base",
+        ),
+        pytest.param(
+            ModExpInput(
+                base="ff" * 32 + "fb",
+                exponent="02",
+                modulus="05",
+                declared_base_length=33,
+                declared_exponent_length=1,
+                declared_modulus_length=1,
+            ),
+            # expected result:
+            bytes.fromhex("01"),
+            id="33-bytes-long-base",  # higher cost than 32 bytes
+        ),
+        pytest.param(
+            ModExpInput(
+                base="ee" * 32,
+                exponent="ff" * Spec.MAX_LENGTH_BYTES,  # 1024 is upper limit
+                modulus="03",
+                declared_base_length=32,
+                declared_exponent_length=1024,
+                declared_modulus_length=1,
+            ),
+            # expected result:
+            bytes.fromhex("02"),
+            id="1024-bytes-long-exp",
+        ),
+        pytest.param(
+            ModExpInput(
+                base="e09ad9675465c53a109fac66a445c91b292d2bb2c5268addb30cd82f80fcb0033ff97c80a5fc6f39193ae969c6ede6710a6b7ac27078a06d90ef1c72e5c85fb5",
+                exponent="010001",
+                modulus="fc9e1f6beb81516545975218075ec2af118cd8798df6e08a147c60fd6095ac2bb02c2908cf4dd7c81f11c289e4bce98f3553768f392a80ce22bf5c4f4a248c6b",
+                declared_base_length=64,
+                declared_exponent_length=3,
+                declared_modulus_length=64,
+            ),
+            # expected result:
+            bytes.fromhex(
+                "c36d804180c35d4426b57b50c5bfcca5c01856d104564cd513b461d3c8b8409128a5573e416d0ebe38f5f736766d9dc27143e4da981dfa4d67f7dc474cbee6d2"
+            ),
+            id="nagydani-1-pow0x10001",
+        ),
+        pytest.param(
+            ModExpInput(
+                base="8d74b1229cc36912165d7ed62334d5ce0683ad12dbade86cdbd705f46693d6c0",
+                exponent="00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                modulus="8f70e8f94c5ad28ed971e258ea3854ebf57131ae4c842e5cafe1c70db8272caf",
+                declared_base_length=32,
+                declared_exponent_length=64,
+                declared_modulus_length=32,
+            ),
+            # expected result:
+            bytes.fromhex(
+                "0000000000000000000000000000000000000000000000000000000000000001"
+            ),
+            id="zero-exponent-64bytes",
+        ),
+    ],
+)
+def test_modexp_different_base_lengths(
+    state_test: StateTestFiller,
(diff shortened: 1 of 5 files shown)


tests/osaka/eip7883_modexp_gas_increase/test_eip_mainnet.py
```diff
@@ -0,0 +1,63 @@
...
+                declared_base_length=33,
+                declared_exponent_length=1,
+                declared_modulus_length=1,
+            ),
+            bytes([1]),
+            id="33-bytes-long-base",  # higher cost than 32 bytes
+        ),
+    ],
+)
+@pytest.mark.valid_at("Osaka")
```
spencer-tb: We should mark these tests as mainnet only like we do in: https://github.com/ethereum/execution-spec-tests/pull/1511/files#diff-a9add5727efb38c476ff59942c2a0732eb1f9dde4958dd3b573a36fe3444bf05R15
```suggestion
@pytest.mark.valid_at("Osaka")
@pytest.mark.mainnet
```

We only want to fill/execute these specifically for mainnet validation.

tests/osaka/eip7883_modexp_gas_increase/test_eip_mainnet.py
```diff
@@ -0,0 +1,63 @@
+"""
+EIP-7883 ModExp gas cost increase tests.
+
+Tests for ModExp gas cost increase in
+[EIP-7883: ModExp Gas Cost Increase](https://eips.ethereum.org/EIPS/eip-7883).
```
spencer-tb: Could we make it clear in the docstring that these are specifically for mainnet execute use only (fill is a sanity check).
```suggestion
Mainnet Marked Checklist Tests for ModExp gas cost increase in
[EIP-7883: ModExp Gas Cost Increase](https://eips.ethereum.org/EIPS/eip-7883).
```

tests/osaka/eip7951_p256verify_precompiles/test_eip_mainnet.py
```diff
@@ -0,0 +1,91 @@
+"""
+Tests for [EIP-7951: Precompile for secp256r1 Curve Support](https://eips.ethereum.org/EIPS/eip-7951).
```
spencer-tb: Similar here:
```suggestion
Mainnet Marked Checklist Tests for [EIP-7951: Precompile for secp256r1 Curve Support](https://eips.ethereum.org/EIPS/eip-7951).
```

tests/osaka/eip7951_p256verify_precompiles/test_eip_mainnet.py
```diff
@@ -0,0 +1,91 @@
...
+    StateTestFiller,
+    Transaction,
+)
+
+from .spec import H, R, S, Spec, X, Y, ref_spec_7951
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_7951.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7951.version
+
+
```
spencer-tb, line 20: If we add this here we can remove Osaka mark's below.
```suggestion
pytestmark = [pytest.mark.valid_at("Osaka"), pytest.mark.mainnet]
```

tests/osaka/eip7951_p256verify_precompiles/test_eip_mainnet.py
```diff
@@ -0,0 +1,91 @@
...
+                0xDA3C553A4215893E6D95D5818DF2519E13233A1E0E56E0EA7B4817A92A6973F9
+            )
+            + Y(
+                0xD8C8877A49383E20C3FDC21D4E8E280EF1FEEB72C333036770369A4387168D33
+            ),
+            id="valid_r1_sig",
+        ),
+    ],
+)
+@pytest.mark.valid_at("Osaka")
```
spencer-tb:
```suggestion
```

tests/osaka/eip7883_modexp_gas_increase/test_eip_mainnet.py
```diff
@@ -0,0 +1,63 @@
...
+            ModExpInput(
+                base="ff" * 32 + "fb",
+                exponent="02",
+                modulus="05",
+                declared_base_length=33,
+                declared_exponent_length=1,
+                declared_modulus_length=1,
+            ),
+            bytes([1]),
+            id="33-bytes-long-base",  # higher cost than 32 bytes
```
spencer-tb, line 53: I think we should remove this test and add 1/2 extras. These both currently test the minimum gas cost increase of 200 to 500 from the EIP.

1) If we add `nagydani-1-pow0x10001` from [here](https://github.com/ethereum/execution-specs/blob/forks/osaka/tests/osaka/eip7883_modexp_gas_increase/vector/vectors.json#L26) we additionally check that the 2x complexity multiplier works, `if max_length > 32: multiplication_complexity = 2 * words**2`, from the spec.


2) If we add `zero-exponent-64bytes` from [here](https://github.com/ethereum/execution-specs/blob/forks/osaka/tests/osaka/eip7883_modexp_gas_increase/vector/vectors.json#L222) we additionally test the 16x exponent multiplaction works, `iteration_count = (16 * (exponent_length - 32)) + ((exponent & (2**256 - 1)).bit_length() - 1)`, from the spec.


Please see below for modexp equation snippets:
https://eips.ethereum.org/EIPS/eip-7883#specification


They shouldn't cost too much more.

tests/osaka/eip7951_p256verify_precompiles/test_eip_mainnet.py
```diff
@@ -0,0 +1,91 @@
...
+def test_invalid(
+    state_test: StateTestFiller, pre: Alloc, post: dict, tx: Transaction
+) -> None:
+    """
+    Negative mainnet test for the P256VERIFY precompile.
+
+    The signature actually is a valid secp256k1 signature,
+    so this is an interesting test case.
+    """
+    state_test(env=Environment(), pre=pre, post=post, tx=tx)
```
spencer-tb, line 91: Nice. My understanding is that this is a valid tx but with an invalid signature?

spencer-tb (commented):
Thanks for starting on this! :)

Some comments on the existing tests below and ideas for the other EIPs.

I understand the logic for not adding tests for some of these EIPs but if we can I really think we should, mostly for completeness but also so we can sleep easy. It will additionally look better for announcing on twitter/x after the fork!

In another PR I think we should add mainnet marked tests for EIP-7954 and EIP-7918 (blob related EIPs).

### EIP-7823

While we already have some validation from the EIP-7883 tests, I do think it would be good to add a test right at the 1024-byte boundary. We can re-use [`base_boundary`](https://github.com/ethereum/execution-specs/blob/forks/osaka/tests/osaka/eip7823_modexp_upper_bounds/test_modexp_upper_bounds.py#L214-L222):
```python
pytest.param(
      ModExpInput(
          base=b"\x01" * Spec.MAX_LENGTH_BYTES,
          exponent=b"\x00",
          modulus=b"\x02",
      ),
```

### EIP-7939

I noted you stated not to add a tests for this but I think we should add at least one maybe 2 for sanity:
- One basic case with for example 8 leading zeros,
- One edge case with all zeros (256).

Something like the following cases:
```python
@pytest.mark.parametrize(
      "value,expected_clz",
      [
          pytest.param(    0x0100...0000,
              8,
              id="clz_8_leading_zeros",
          ),
          pytest.param(
              0x0000...0000,
              256,
              id="clz_all_zeros",
          ),
      ],
  )
```

### EIP-7825

Worth considering for completeness, but I won't push too hard on this one.

Adding a simple state test tx on exactly the cap (`2^24`).

Please see `test_gas_limit_cap_over`: https://github.com/ethereum/execution-specs/blob/forks/osaka/tests/osaka/eip7825_transaction_gas_limit_cap/test_tx_gas_limit.py#L68

tests/osaka/eip7951_p256verify_precompiles/test_eip_mainnet.py
felix314159: it is a valid ecdsa signature over a different curve, so the signature verification is expected to fail (but this is a good test because the other curve is also used by clients)

tests/osaka/eip7883_modexp_gas_increase/test_eip_mainnet.py
felix314159: The link you have put shows that these two test do NOT both test the same thing. The 32 bytes base test is cheaper than the 33 bytes test. If you remove one of them you lose coverage of that case. But I agree that we can add another test we an even larger number, e.g. the two cases u mentioned

felix314159:
thanks for the feedback, i added more cases you requested. let's see if ci passes

spencer-tb:
Nice! Thanks for adding the changes. You don't think we should cases add for 7823/7939/7825?

felix314159:
I can try to make a follow-up PR with more cases. Can we get this PR merged now?

spencer-tb (approved):
LGTM! Thanks :)

PR #1807, opened 2025-11-24
chore(tests): read test list from file

SamWilsn:
## 🗒️ Description

Teaches the json_infra tests how to read the set of tests to execute from a file. Useful in conjunction with [minimize-tests](https://github.com/SamWilsn/minimize-tests).

#### Cute Animal Picture

### DIFF
--- a/tests/json_infra/conftest.py
+++ b/tests/json_infra/conftest.py
@@ -80,6 +80,7 @@ def pytest_addoption(parser: Parser) -> None:
         type=str,
         help="Run tests from and including the specified fork.",
     )
+
     parser.addoption(
         "--until",
         action="store",
@@ -88,6 +89,7 @@ def pytest_addoption(parser: Parser) -> None:
         type=str,
         help="Run tests until and including the specified fork.",
     )
+
     parser.addoption(
         "--fork",
         action="store",
@@ -95,6 +97,7 @@ def pytest_addoption(parser: Parser) -> None:
         default="",
         help="Only run tests for the specified fork.",
     )
+
     parser.addoption(
         "--file-list",
         action="store",
@@ -104,6 +107,13 @@ def pytest_addoption(parser: Parser) -> None:
             "repository. This option specifies the path to a file which "
             "contains a list of relevant paths."
         ),
+    )
+
+    parser.addoption(
+        "--tests-file",
+        dest="tests_path",
+        type=Path,
+        help="Path to a file containing test ids, one per line",
     )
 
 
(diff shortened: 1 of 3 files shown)


tests/json_infra/conftest.py
```diff
@@ -125,6 +131,36 @@ def pytest_collection_modifyitems(config: Config, items: list[Item]) -> None:
...
+            test_ids.remove(item.nodeid)
+        else:
+            deselected.append(item)
+
+    if deselected:
+        config.hook.pytest_deselected(items=deselected)
+        items[:] = selected  # keep only what matches
+
+
+def pytest_collection_modifyitems(config: Config, items: list[Item]) -> None:
```
gurukamath: I think this hook will no longer be used if we go ahead with the changes proposed in #1666

SamWilsn: I dropped the part that's no longer used. Neat side effect of #1666 is that the testid actually points to the JSON file now!

tests/json_infra/conftest.py
```diff
@@ -175,6 +185,29 @@ def pytest_configure(config: Config) -> None:
...
 
 
+def pytest_collection_modifyitems(config: Config, items: list[Item]) -> None:
+    """Filter test items."""
+    tests_path = config.getoption("tests_path", None)
+    if tests_path is None:
+        return
+
+    with open(tests_path) as f:
+        test_ids = set(x[:-1] for x in f.readlines())
```
gurukamath:
```suggestion
        test_ids = set(line.strip() for line in f)
```
Otherwise, this could skip the last test in the file, if the file does not end with a new line

tests/json_infra/conftest.py
```diff
@@ -106,6 +109,13 @@ def pytest_addoption(parser: Parser) -> None:
         ),
     )
 
+    parser.addoption(
```
gurukamath, line 112: Some of these options might end up contradicting each other. For example, we might have `--fork=Frontier` and the tests file might have an entry called `tests/json_infra/fixtures/latest_fork_tests/fixtures/state_tests/berlin/eip2930_access_list/test_account_storage_warm_cold_state.json::tests/berlin/eip2930_access_list/test_acl.py::test_account_storage_warm_cold_state[fork_Shanghai-state_test-account_warm_True-storage_key_warm_True]::Shanghai::0` which is from `Shanghai`.

In this set-up, the Shanghai tests will not run. Is this what is intended? Alternatively, we could just make it invalid to pair the `--tests-file` option with the others.

tests/json_infra/conftest.py
```diff
@@ -175,6 +185,29 @@ def pytest_configure(config: Config) -> None:
     config.stash[desired_forks_key] = desired_forks
 
 
+def pytest_collection_modifyitems(config: Config, items: list[Item]) -> None:
```
gurukamath, line 188: This hook runs after everything has been parsed and collected and then filters items out. Just thinking out loud here but wonder if we should extract the json_paths from all the entries, only parse those and then let `pytest_collection_modifyitems` filter it further.

No strong opinion on this either way though.

tests/json_infra/conftest.py
SamWilsn: I think this makes sense. If I'm not mistaken, the following command wouldn't run any tests either, so this behaviour is consistent.

```bash
pytest --fork=Frontier \
    'tests/json_infra/fixtures/latest_fork_tests/fixtures/state_tests/berlin/eip2930_access_list/test_account_storage_warm_cold_state.json::tests/berlin/eip2930_access_list/test_acl.py::test_account_storage_warm_cold_state[fork_Shanghai-state_test-account_warm_True-storage_key_warm_True]::Shanghai::0'
```

tests/json_infra/conftest.py
SamWilsn: That would absolutely make it faster. I'll open an issue for that optimization.

gurukamath (approved):

PR #1810, opened 2025-11-26
refactor(test-benchmark): repricing filter logic

LouisTsai-Csie:
## 🗒️ Description

- Supported Command: `fill` &`execute`

- Modes:
  - `-m benchmark` (under `compute/` folder)
  - `-m stateful` (under `stateful/` folder)

- Flags
  - `--gas-benchmark-values` (specify the block gas limit)
  - `--fixed-opcode-count` (specify the opcode count)

- Options
  - `-m repricing` (run only the subset of benchmark tests for quick testing)

**Issue Description**
When running benchmark tests with the `--fixed-opcode-count` flag, the current behavior incorrectly restricts execution to only the benchmark cases marked with the repricing marker.  

However, `repricing` is meant to be an **independent option** that limits execution to a subset of the benchmark suite.  
Using `--fixed-opcode-count` should **not** imply repricing mode.

**Fix Applied in This PR**
This PR adjusts the logic so that:

- Running with `--fixed-opcode-count` executes the **full benchmark suite**: Four passes are executed (including `contract_balance` with values 0 and 1):
```bash
fill -v tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance --fixed-opcode-count 10 --clean -m benchmark
```

- Running with -m repricing executes only the repricing subset: Two passes are executed (only contract_balance = 0)
```bash
fill -v tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance --fixed-opcode-count 10 --clean -m repricing
```

Add a sanity check CI workflow for fixed opcode count feature.

## 🔗 Related Issues or PRs

None

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -115,26 +115,22 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
 
     if "fixed_opcode_count" in metafunc.fixturenames:
         # Only parametrize if test has repricing marker
-        has_repricing = (
-            metafunc.definition.get_closest_marker("repricing") is not None
-        )
-        if has_repricing:
-            if fixed_opcode_counts:
-                opcode_counts = [
-                    int(x.strip()) for x in fixed_opcode_counts.split(",")
-                ]
-                opcode_count_parameters = [
-                    pytest.param(
-                        opcode_count,
-                        id=f"opcount_{opcode_count}K",
-                    )
-                    for opcode_count in opcode_counts
-                ]
-                metafunc.parametrize(
-                    "fixed_opcode_count",
-                    opcode_count_parameters,
-                    scope="function",
+        if fixed_opcode_counts:
+            opcode_counts = [
+                int(x.strip()) for x in fixed_opcode_counts.split(",")
+            ]
+            opcode_count_parameters = [
+                pytest.param(
+                    opcode_count,
+                    id=f"opcount_{opcode_count}K",
                 )
+                for opcode_count in opcode_counts
+            ]
+            metafunc.parametrize(
+                "fixed_opcode_count",
+                opcode_count_parameters,
+                scope="function",
+            )
 
 
 @pytest.fixture(scope="function")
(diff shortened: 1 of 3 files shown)


packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -250,6 +250,11 @@ def model_post_init(self, __context: Any, /) -> None:
 
         blocks: List[Block] = self.setup_blocks
 
+        if self.fixed_opcode_count is not None and self.code_generator is None:
+            pytest.skip(
+                "Cannot run fixed opcode count tests without a code generator"
+            )
+
```
LouisTsai-Csie, line 255: Currently, the fixed opcode count feature is supported only by benchmark tests that use the code generator. If a test does not support this feature, it should be skipped.

There are two possible ways to filter such tests:

We could skip them during collection via `pytest_collection_modifyitems`, but this is more complex because it requires determining whether each test is a benchmark test and whether it uses the code generator.

Therefore, I chose to skip the unsupported tests directly in benchmark test wrapper instead.

Carsons-Eels: I think this is fine, it's a very fast check so the difference in time should be negligible. 👍

Carsons-Eels (approved):
I do not know the EEST code as well as EELS, but I think this looks good.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -44,10 +44,16 @@ def pytest_configure(config: pytest.Config) -> None:
...
-    if not fixed_opcode_count:
-        # If --fixed-opcode-count is not specified, don't filter anything
+
+    if not gas_benchmark_value and not fixed_opcode_count:
+        return
+
+    # Check if -m repricing marker filter was specified
+    markexpr = config.getoption("markexpr", "")
+    if "repricing" not in markexpr:
         return
```
spencer-tb, line 57:
```suggestion
    # Check if -m repricing marker filter was specified
    markexpr = config.getoption("markexpr", "")
    if "repricing" not in markexpr or "not repricing" in markexpr:
        return
```

spencer-tb (approved):
LGTM! One small change.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
LouisTsai-Csie: Done!

PR #1812, opened 2025-11-26
Qu0b/add bal test cases

qu0b:
## 🗒️ Description

Implement new BAL (Block Access List) tests with full test implementations:

- `test_bal_create2_collision` - CREATE2 address collision handling
- `test_bal_create_selfdestruct_to_self_with_call` - Init code with external call then selfdestruct to self
- `test_bal_selfdestruct_to_7702_delegation` - SELFDESTRUCT to EIP-7702 delegated account
- `test_bal_revert_insufficient_funds` - CALL failure due to insufficient balance
- `test_bal_lexicographic_address_ordering` - Strict byte-wise address ordering validation with endian-trap addresses
- `test_bal_transient_storage_not_tracked` - EIP-1153 transient storage exclusion
- `test_bal_selfdestruct_to_precompile` - SELFDESTRUCT with precompile beneficiary
- `test_bal_all_transaction_types` - All 5 tx types (Legacy, EIP-2930, EIP-1559, Blob, EIP-7702) in single block
- `test_bal_create_early_failure` - CREATE failure before track_address
- `test_bal_withdrawal_to_7702_delegation` - Withdrawal to EIP-7702 delegated account

image


tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -94,3 +94,10 @@
...
+| `test_bal_revert_insufficient_funds` | Ensure BAL handles CALL failure due to insufficient balance (not OOG) | Contract (balance=100, storage slot 0x02=0xDEAD) executes: `SLOAD(0x01), CALL(target, value=1000), SSTORE(0x02, result)`. CALL fails because 1000 > 100. Target address 0xDEAD (pre-existing with non-zero balance to avoid pruning). Note: slot 0x02 must start non-zero so SSTORE(0) is a change. | BAL **MUST** include: (1) Contract with `storage_reads` for slot 0x01, `storage_changes` for slot 0x02 (value=0, CALL returned failure). (2) Target (0xDEAD) with empty changes. Per EVM, CALL loads target address before balance check, so target **MUST** appear in BAL even though transfer failed. No `balance_changes` for target. | 🟡 Planned |
```
nerolation: Re `test_bal_revert_insufficient_funds` , we need to double check if the target address should be in the BAL or not. My intuition is that the BAL MUST NOT include the target if the balance of the caller is insufficient for the CALL.
Also, the sentence in the expectation ends abruptly.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -94,3 +94,10 @@
...
+| `test_bal_create2_collision` | Ensure BAL handles CREATE2 address collision correctly | Factory contract (nonce=1, storage slot 0=0xDEAD) executes `CREATE2(salt=0, initcode)` targeting address that already has `code=STOP, nonce=1`. Pre-deploy contract at calculated CREATE2 target address before factory deployment. | BAL **MUST** include: (1) Factory with `nonce_changes` (1→2, incremented even on failed CREATE2), `storage_changes` for slot 0 (0xDEAD→0, stores failure). (2) Collision address with empty changes (accessed during collision check, no state changes). CREATE2 returns 0. Collision address **MUST NOT** have `nonce_changes` or `code_changes`. | 🟡 Planned |
```
jochem-brouwer: Also ensure that there are collision checks for the creation transaction, the CREATE opcode, CREATE2 opcode. The collision "rules" are described in this EIP-684: https://eips.ethereum.org/EIPS/eip-684

@nerolation I think we also want a "collision" check for [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702)? Rule 5:

> Verify the code of authority is empty or already delegated.

If a check fails then:

> If any step above fails, immediately stop processing the tuple and continue to the next tuple in the list. When multiple tuples from the same authority are present, set the code using the address in the last valid occurrence.

I'm not sure if this could lead to new cases, but I was thinking of delegating to an account which has "normal" code (so not code like `0xef0100 || address`). However this would be an edge case which would require first deploying a contract to this address and then signing the authority on behalf of that address, which is super unlikely (but still a collision). Note that in the tests we can deploy this contract in genesis so we do not have to "mine" addresses.
The relevant test for this might already be in the 7702 tests. For this test we should only expand it with the expected BAL values.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -94,3 +94,10 @@
...
+| `test_bal_create_selfdestruct_to_self_with_call` | Ensure BAL handles init code that calls external contract then selfdestructs to itself | Factory executes `CREATE2` with endowment=100. Init code: (1) `CALL(Oracle, 0)` - Oracle writes to its storage slot 0x01. (2) `SSTORE(0x01, 0x42)` - write to own storage. (3) `SELFDESTRUCT(SELF)` - selfdestruct to own address. Contract created and destroyed in same tx. Note: Uses EXTCODECOPY from template contract for init code (too long for PUSH32). | BAL **MUST** include: (1) Factory with `nonce_changes`, `balance_changes` (loses 100). (2) Oracle with `storage_changes` for slot 0x01 (external call succeeded). (3) Template contract (EXTCODECOPY source). (4) Created address with `storage_reads` for slot 0x01 (aborted write becomes read) - **MUST NOT** have `nonce_changes`, `code_changes`, `storage_changes`, or `balance_changes` (ephemeral contract, SELFDESTRUCT to self = net zero). | 🟡 Planned |
```
jochem-brouwer: `CALL(Oracle, 0)` how should I read this? Is `Oracle` the target address and `0` the value sent?

Small note: the factory contract itself can already have the initcode embedded in it's code such that we do not need to `EXTCODECOPY` from the template contract (but both methods are fine by me, just noting that embedding initcode is also possible, this would remove the template contract from the expected BAL results).

Note: `SELFDESTRUCT` to self is currently the only way to burn ETH (I think this is what you mean by "net zero"?). This has to happen in the same transaction as where the contract was created, otherwise it is not burned (see https://eips.ethereum.org/EIPS/eip-6780)

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -94,3 +94,10 @@
...
+| `test_bal_lexicographic_address_ordering` | Ensure BAL enforces strict lexicographic byte-wise ordering | Pre-fund three addresses with specific byte patterns: `addr_low = 0x0000...0001`, `addr_mid = 0x0000...0100`, `addr_high = 0x0100...0000`. Contract touches them in reverse order: `BALANCE(addr_high), BALANCE(addr_low), BALANCE(addr_mid)`. | BAL account list **MUST** be sorted lexicographically by address bytes: `addr_low` < `addr_mid` < `addr_high`, regardless of access order. This catches endianness bugs where addresses are compared as big integers instead of byte arrays. Complements `test_bal_invalid_account_order` which tests rejection; this tests correct generation. | 🟡 Planned |
```
jochem-brouwer: Note: `addr_low` of `0x00..01` is a precompile, the others `addr_mid`, `addr_high` are not. Which I think is good because this hardens the test, but this would thus also test if precompile addresses are  treated "the same" as normal addresses (same lexicographic ordering). 

We should also include a test (if it's not in this file already) which checks the lexicographic ordering of the precompiles to ensure nothing funny happens there. (And maybe a mix with normal EOAs + precompiles, or only EOAs)

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -94,3 +94,10 @@
...
+| `test_bal_selfdestruct_to_precompile` | Ensure BAL captures SELFDESTRUCT with precompile as beneficiary | Contract with balance=100 wei executes `SELFDESTRUCT(0x0000...0001)` (ecrecover precompile). Precompile starts with balance=0. Note: Uses caller contract to trigger selfdestruct (post-Cancun SELFDESTRUCT only transfers balance on CALL). | BAL **MUST** include: (1) Contract with `balance_changes` (100→0, loses balance to selfdestruct). (2) Precompile address 0x01 with `balance_changes` (0→100, receives selfdestruct balance). Precompile **MUST NOT** have `code_changes` or `nonce_changes`. This complements `test_bal_withdrawal_to_precompiles` (withdrawal) and `test_bal_precompile_funded` (tx value). | 🟡 Planned |
```
jochem-brouwer: > (post-Cancun SELFDESTRUCT only transfers balance on CALL)

I think what is meant here is that the `SELFDESTRUCT` does not selfdestruct but only send all balance? (The `CALL` word is confusing here)

jochem-brouwer (commented):
Hey Stefan 😄 👍 We met at Devconnect!! 🥳 

Very impressive edge cases! I left a few comments on them, also some clarifications or just some notes. Thanks a lot 😄 👍

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
fselmo: > For this test we should only expand it with the expected BAL values.

The ideal way for this imo is to add the expected access list to the existing case instead of re-writing the same test just with an expectation. I think this is what you mean, no? We should avoid rewriting the same test cases just with an expected BAL.

See [here](https://github.com/fselmo/execution-specs/blob/abee0f4dabd3281ae172033c36b3edbfd27f9054/tests/cancun/eip6780_selfdestruct/test_selfdestruct_revert.py#L439-L493) and [here](https://github.com/fselmo/execution-specs/blob/abee0f4dabd3281ae172033c36b3edbfd27f9054/tests/frontier/opcodes/test_call_and_callcode_gas_calculation.py#L214-L272), defined as a fixture and passed to the test [here](https://github.com/fselmo/execution-specs/blob/abee0f4dabd3281ae172033c36b3edbfd27f9054/tests/frontier/opcodes/test_call_and_callcode_gas_calculation.py#L285).

jochem-brouwer: @fselmo Yup exactly, this makes us DRY ( :stuck_out_tongue_winking_eye: )! :smile: :+1: I think a pitfall might be now, but also for future EIPs, is to write a scenario which is already there in an old test. In that case then in that specific test there should be some "overlay" to the test which would then add the expected values of the EIP in question.
The downside of this approach (finding old tests) is that they are not in the relevant directory of that EIP (like `fork/eip-XXXX/test-eip-xxxx-scenarios.py`) but are rather at the relevant test which covers that situation for an older fork.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
qu0b: Yes, `CALL(Oracle, 0)` calls a contract with `0` as the value sent.

Thanks! I will remove the `EXTCODECOPY` part.

yes, the idea is to test the edge cases of calling `SELFDESTRUCT` on `SELF` within the same transaction that the contract was created.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
qu0b: Hmm yeah I know what you mean. Maybe we should add the balance check here: 

https://github.com/qu0b/execution-specs/blob/b172548eda946a2bb925125431a124267f5e95df/src/ethereum/forks/amsterdam/vm/instructions/system.py#L455

before the accessed_addresses list here:

https://github.com/qu0b/execution-specs/blob/b172548eda946a2bb925125431a124267f5e95df/src/ethereum/forks/amsterdam/vm/instructions/system.py#L411-L412

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
qu0b: I think the lexicographic ordering is a simple sorting problem in the clients, I mainly added it as a sanity check. most clients will do this in a post process step after creating the block level access list. The only addition that we would see as valuable would be to add endianness switching of the addresses i.e. have two addresses that are identical but their endianness is reversed.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
qu0b: you're right the note is confusing would this be resolved by just removing the note?

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
qu0b: @jochem-brouwer there is a lot of info in these comments, which is great, but also confusing, could you condense it down into what needs to be done specifically for `test_bal_create2_collision` test case above? Is it needed, is it a duplicate?

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -402,6 +402,19 @@ def call(evm: Evm) -> None:
...
+        evm.message.block_env.state, evm.message.current_target
+    ).balance
+    if sender_balance < value:
+        # Only charge transfer + memory (not cold access, since we didn't access)
+        charge_gas(evm, transfer_gas_cost + extend_memory.cost)
+        evm.memory += b"\x00" * extend_memory.expand_by
+        push(evm.stack, U256(0))
+        evm.return_data = b""
+        evm.pc += Uint(1)
+        return
```
qu0b: @nerolation @fselmo this would move the balance check up before reading the target address

qu0b:
Recent run with the new test cases against besu and nethermind.

image

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
jochem-brouwer: Yes, the test cases to be added for `test_bal_create2_collision` need to include `CREATE` opcode, transaction creation, and a collision caused by a 7702-delegation.
The discussion @fselmo and me had is about that these tests (collision tests), esp. the 7702-collision, are likely already part of the test suite in a different location. 
@fselmo How should we handle this? For people new to this repository we cannot ask that they first find if the test is a duplicate I'd think. Would this be something we'd mark as "possible duplicate" to clean this up later in a new PR? (With the danger it does not get cleaned and we have multiple duplicates, which I'm not sure if that's a super big problem)

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
jochem-brouwer: Yes, I think that'd work :+1:

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
fselmo: Yes I think submitting cases specifically into 7928 tests is just fine. I'd rather not put this burden on the contributor. We can try to clean these up later when we get some bandwidth as part of a cleanup once we have good coverage.

jochem-brouwer: Yes I agree with this sentiment. I'm also not sure if duplicate tests are a big problem. Cleaning tests up or removing them might actually remove subtle details of the tests which would not show up in current coverage decline if you would remove them, but could show up later if you would run this against a different version of client / python spec. So maybe it is actually good to "just write" tests with the risk of duplicates but for the sake of having good coverage.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
nerolation: It looks like the specs are correct.
But the test definition is potentially wrong.

In L412, we're only tracing the 2929 access list, nothing re BALs.
Then at L455ff we never call generic_call such that the target would never be tracked.

Based on the specs, the target of a call with value, but with insufficient balance, is only put in the BAL in case `generic_call` executes and calls `process_message`.

fselmo:
This needs a rebase with the recent additions from #1825 and #1829

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
```diff
@@ -2302,3 +2304,950 @@ def test_bal_cross_block_ripemd160_state_leak(
...
+            alice: Account(nonce=1),
+            factory: Account(nonce=2, balance=factory_balance - endowment),
+            oracle: Account(storage={0x01: 0x42}),
+            # Created address doesn't exist - destroyed in same tx
+            created_address: Account.NONEXISTENT,
+        },
+    )
+
+
+def test_bal_revert_insufficient_funds(
```
fselmo: We should add `call` to this test name and also move this file to `*_opcodes.py` as mentioned above.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
fselmo: I think this is resolved, yes? The target is always accessed if there is enough gas to get to the `is_account_alive` check and also checking whether the target code is a valid delegation designation. Specs are correct here and the target is in BAL.

fselmo (approved):
lgtm, ty @qu0b 🚀 

I moved some test cases to the `*_opcodes.py` for better organization and organized `test_cases.md`, added an explicit `None` check that was missing from an expectation ([here](https://github.com/ethereum/execution-specs/pull/1812/commits/b4f8109684ab780ec77c0c47fb3b9a678461bb86#diff-fe640a71f497572e788ba2601460ebc47862bae1852228008717b67aa25ca78bR2311)).

Issue #1756, opened 2025-11-05
chore(docs): re-enable benchmark test documentation


danceratopz:
#1681 removed the `__init__.py` files and contents from `tests/benchmarks/`; this broke the test case reference in the html docs.

LouisTsai-Csie:
This would be super helpful!!

Issue #1616, opened 2025-09-25
feat(fill,execute,benchmark): test selection via pytest's `testpaths` instead of markers via `--mode`


danceratopz:
We now have 3 classes of test:
1. Regular tests.
2. Computationally intensive tests (`./tests/benchmark/compute`[*])
3. State intensive tests (`./tests/benchmark/state/`).

( [*] These paths don't reflect the current state, but the proposal here:  https://github.com/ethereum/execution-specs/issues/1616 )

The requirement, as I understand, is:
1. `fill` should only fill regular tests by default (same for `execute`).
2. `fill -m benchmarks` should only run tests from `./tests/benchmarks/compute/` (same for `execute`).
3. `execute -m stateful` should only execute tests from `./tests/benchmarks/state/` (fill is currently not required/realistically possible).

This requirement is implemented (via pytest markers) here:
- ethereum/execution-spec-tests#1920, and
- ethereum/execution-spec-tests#2169 (WIP).

Using pytest markers for this case results in a lot of very complex logic, and, as it stands in ethereum/execution-spec-tests#2169, adds two additional modifications of `pytest_collect_modifyitems()` in two conftests under `./tests/`.

There seems to be a much simpler solution: Just collect tests from the applicable sub-directories. This will:
1. Result in much simpler pytest logic.
2. Result in much faster test collection times.

This can be achieved by modifying the pytest ini file `testpaths` config for the `fill` and `execute` commands. This config tells pytest where to search for tests.

#### UX and Corresponding Required Pytest Flags

I'd suggest adding a `--mode` flag, which is `None` by default.

Relevant pytest config and docs:
- `testpaths`
   - See our ini files, e.g. https://github.com/LouisTsai-Csie/execution-spec-tests/blob/7e4d3e214252fa878a7ac5c381fcab0d48d1c40a/src/cli/pytest_commands/pytest_ini_files/pytest-fill.ini#L5
   - https://docs.pytest.org/en/stable/reference/reference.html#confval-testpaths
- `-o`:
    ```
    -o OVERRIDE_INI, --override-ini=OVERRIDE_INI
                          Override ini option with "option=value" style, e.g. `-o
                          xfail_strict=True -o cache_dir=cache`.
   ```
- `-ignore`: 
    ```
    --ignore=path         Ignore path during collection (multi-allowed)
    ```

1. EIP testing (default behavior):
    ```
    fill
    ```
    should result in no changes to `testpaths` but automatically add these pytest flags:
[...]

PR #1821, opened 2025-11-28
refactor(benchmark): ignore `tests/benchmark` path in fill and drop markers

LouisTsai-Csie:
## 🗒️ Description

Continuing the discussion from this [comment](https://github.com/ethereum/execution-specs/issues/1616#issuecomment-3588049927): based on @danceratopz’s feedback, if we specify the `--ignore tests/benchmark` flag in the ini file for `fill`, pytest cannot later revert this behavior.

I looked into this and found a special hook, `pytest_ignore_collect`, which allows us to dynamically control which paths should be ignored during test collection. Using this hook, if a test path is not explicitly included under `tests/benchmark`, we can make all benchmark tests being ignored by default.

With this approach, we can remove the benchmark and stateful markers, as well as the filtering logic in `conftest.py`, which has already become overly complicated. 

I tested this draft PR in several combinations:

- fill vs. execute
- normal benchmark vs. fixed-opcode-count benchmark
- repricing marker or not

All combinations behave as expected with the new filtering logic.

My main concern is that this change may affect the current documentation structure, since the docs relies on the `benchmark` marker as a reference. But it seems the docs now is still under [maintenance](https://github.com/ethereum/execution-specs/issues/1756).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -30,31 +32,68 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+                # Explicitly specified, set op_mode and don't ignore
+                config.op_mode = OpMode.BENCHMARKING  # type: ignore[attr-defined]
+                return False
+            except ValueError:
+                continue
+        # Check relative paths containing 'benchmark'
+        elif "benchmark" in arg:
+            # Explicitly specified, set op_mode and don't ignore
+            config.op_mode = OpMode.BENCHMARKING  # type: ignore[attr-defined]
+            return False
```
LouisTsai-Csie, line 66: This step checks whether the benchmark path is explicitly specified, but i think this is not the best approach

tests/benchmark/conftest.py
```diff
@@ -32,62 +32,6 @@ def pytest_generate_tests(metafunc: Any) -> None:
...
-
-    if gen_docs:
-        for item in items:
-            if (
-                benchmark_dir in Path(item.fspath).parents
-                and not item.get_closest_marker("benchmark")
-                and not item.get_closest_marker("stateful")
-            ):
-                item.add_marker(benchmark_marker)
-        return
```
danceratopz, line 49: This code is pretty ugly, but this PR will have to add some "`gen_docs`" logic so that `./tests/benchmark` is collected if we're generating documentation, so that the docs contain all documentation.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -30,31 +32,68 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+    Returns True to ignore, False to collect, None for default behavior.
+    """
+    # Only handle paths within tests/benchmark/
+    try:
+        collection_path.relative_to(config.rootpath / "tests" / "benchmark")
+    except ValueError:
+        return None
+
+    # Check if benchmark tests explicitly specified in command line arguments
+    benchmark_path = config.rootpath / "tests" / "benchmark"
```
danceratopz, line 50: We could define this above the `try` block above and use it in the try block, too.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -30,31 +32,68 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+    Ignore tests/benchmark/ directory by default unless explicitly specified.
+
+    Returns True to ignore, False to collect, None for default behavior.
+    """
+    # Only handle paths within tests/benchmark/
+    try:
+        collection_path.relative_to(config.rootpath / "tests" / "benchmark")
+    except ValueError:
+        return None
+
```
danceratopz, line 48: To address the failing documentation generation workflow, if fill is running in documentation mode (i.e., `config.getoption("--gen-docs", default=False)` is `True`), then we can simply `return False` here to not ignore benchmark tests and have them included in the documentation.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -30,31 +32,68 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+        if arg_path.is_absolute():
+            try:
+                arg_path.relative_to(benchmark_path)
+                # Explicitly specified, set op_mode and don't ignore
+                config.op_mode = OpMode.BENCHMARKING  # type: ignore[attr-defined]
+                return False
+            except ValueError:
+                continue
+        # Check relative paths containing 'benchmark'
+        elif "benchmark" in arg:
```
danceratopz, line 63: This is a rather loose check, perhaps we could check for "benchmark/compute" or "benchmark/stateful". Of course, it'll break if we add another valid benchmark sub-folder here.

danceratopz (commented):
If we need to keep consensus test and benchmark test filling mutually exclusive, then I think this approach is about as good as it gets 👍 It's still a bit tricky, but much cleaner than the existing marker-based method!

I'll circle back to this PR later, but a few initial comments below.

The content of `./docs` aren't currently hosted anywhere, but we should updated them to reflect the changes in this PR. Can you update them and remove any references to `-m benchmark`? See `docs/writing_tests/benchmarks.md`.

danceratopz (commented):
Also, this functionality needs some unit tests to check collection is working as expected for the benchmark case. I think we can add update these tests:
`packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py`
to test the logic in this PR. We could add an additional valid test module in `tests/benchmark/` and test collection/fixture generation works as expected:
https://github.com/danceratopz/execution-specs/blob/519eb26e663f3bbb4d10a230c8cf0ac0f4c94f9b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py#L472-L474

PR #1822, opened 2025-11-28
enhance(test-cli-execute): Many Improvements to `execute remote` command

marioevz:
## 🗒️ Description
Introduces many enhancements to the `execute remote` and `execute hive` by porting https://github.com/ethereum/execution-spec-tests/pull/1566 to this repo and other fixes.

#### Deferred EOA Funding

Changes the behavior of `pre.fund_eoa` during execute to, instead of immediately sending a the transaction to fund the given EOA with a fixed amount of Eth in all tests, save the incomplete funding transaction into a list, and then process the rest of the transactions included in the test that have the given EOA as sender, then calculate the exact amount of Eth that is required to send all of them given current network gas prices, and finally fund the EOA before submitting the test transactions.

This behavior only kicks in if `pre.fund_eoa` is not called with the `amount` parameter explicitly set. I.e. only `pre.fund_eoa()` or `pre.fund_eoa(amount=None)` use this deferred funding method.

This results in:
- EOAs not being funded with excessive or insufficient Eth to execute the test transactions, rather they are always funded with the exact required amount to send all the transactions.
- Minimizing the amount of Eth spent sending the test transactions due to the gas price being fetched from the network and updated dynamically.
- More deterministic wait times for inclusion of the transactions since we fetch the gas prices from the network.

#### Dynamic Gas Prices

Execute command now dynamically fetches gas prices before the execution of each test, with the information becoming stale every 12 seconds.

The default gas prices (`gas_price`, `max_priority_fee_per_gas`, `max_fee_per_gas`, `max_fee_per_blob_gas`) are now automatically set to 1.5x the current network prices.

The 1.5x multiplier is currently constant, but if we deem necessary we could introduce another flag to change this.

To achieve this, we still set a number to the default gas price (7) during transaction instantiation, but we remove `gas_price`, `max_priority_fee_per_gas`, `max_fee_per_gas`, `max_fee_per_blob_gas` from `model_fields_set` if the user did not explicitly set them. Then `execute` peeks into `model_fields_set` and if the gas prices were not explicitly set, it updates them with the current network prices for all test transactions.

#### Dry-run Mode

This mode is enabled by flag `--dry-run` during `execute remote` in order to estimate how much Eth is going to be consumed when running these tests given the current network gas prices.

It does not send any transaction to the network, rather it just collects all transactions it would send and adds up the gas used and the gas prices.

#### Better Nonce Refresh for Worker Senders

This PR also removes `_refresh_sender_nonce` from the execute's `Alloc` class and moves this behavior to its own fixture. We now separate the `worker_key` (Prev `sender_key`) into two different fixtures:
- `session_worker_key`: Determines the key that will fund all contracts and EOAs for the current pytest worker. This fixture is "session" scoped and also cleans up the worker key by sending the funds back to the seed account.
- `worker_key`: This fixture takes the key from `session_worker_key` and updates its nonce from the current value from the network, then it's passed directly to the `pre` so no update of the nonce has to be performed there.

#### `loadscope` -> `load` for `execute remote` and `execute hive`

This is a minor change with the noticeable improvement when running tests from a single python file.

Using `loadscope` is convenient to `fill` because the module's fixtures are normally a bottleneck to test generation, and this way all the tests from the same file are assigned to the same pytest worker.

This is not the case for `execute` since the worst bottleneck is transaction inclusion in the chain. Now tests are assigned to whichever worker is available so it can run tests while the rest of the workers are waiting for their transactions to be included, even for tests in the same file.

Example running command `uv run execute remote --fork=Osaka ./tests/frontier/opcodes/test_dup.py --rpc-seed-key $RPC_SEED_KEY --rpc-endpoint $RPC_ENDPOINT --chain-id $RPC_CHAIN_ID -n 8` with a Kurtosis local network:

- `loadscope`: 16 passed, 9 warnings in 602.58s (0:10:02)
- `load`: 16 passed, 9 warnings in 97.46s (0:01:37)

Note: `test_dup.py` is currently broken in `forks/osaka` because it sets the `nonce` and the `gas_price`, so the command will definitely fail when trying to reproduce this command.

#### Increased Logging

`logger` is now used more frequently in `sender.py`, `pre_alloc.py` and `execute.py`, to print information about the transaction execution on chain during test execution, and also more verbose errors.

#### Blob Tests In Execute

All `blob_test` tests can now be run with `execute`, by skipping the `getBlobsVX` check which requires the Engine API.

This is a minor improvement mainly targeted to sending blob transactions on live networks, rather than attempting to test the `engine_getBlobsVX` endpoints.

The blob gas prices are also fetched from the network live during execution of the tests.

### ⚠️ Breaking Changes ⚠️

1) Flag `--eoa-fund-amount-default` is removed.
2) Flag `--sender-key-initial-balance` in `execute hive` is now `--seed-key-initial-balance`
3) Flags `--default-gas-price`, `--default-max-fee-per-gas` and `--default-max-priority-fee-per-gas` now default to `None` and ideally should be omitted because, when unset, the command now defaults to fetch the value from the network, which is a more reliable behavior.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/docs/running_tests/execute/hive.md
+++ b/docs/running_tests/execute/hive.md
@@ -16,7 +16,7 @@ Tests can be run using:
 ./hive --client besu --client-file ./configs/osaka.yaml --sim ethereum/eels/execute-blobs
 ```
 
-**Note**: If Engine RPC is unavailable, blob transactions will be sent but `getBlobsV*` validation is skipped.
+**Note**: If the Engine RPC is unavailable, blob transactions will be sent and `getBlobsV*` validation is skipped.
 
 See [Hive](../hive/index.md) for help installing and configuring Hive.
 
--- a/docs/running_tests/execute/hive.md
+++ b/docs/running_tests/execute/hive.md
@@ -4,11 +4,19 @@ Tests can be executed on a local hive-controlled single-client network by runnin
 
 ## The `eest/execute-blobs` Simulator
 
-The `blob_transaction_test` execute test spec sends blob transactions to a running client in order to verify its `engine_getBlobsVX` endpoint behavior. These tests can be run using:
+The `blob_transaction_test` execute test spec sends blob transactions to a running client. Blob transactions are fully supported in execute mode:
+
+- Blob transactions can be sent via `eth_sendRawTransaction`
+- Blob validation via `engine_getBlobsVX` endpoints (when Engine RPC available)
+- Automatic gas pricing for blob gas fees
+
+Tests can be run using:
 
 ```bash
 ./hive --client besu --client-file ./configs/osaka.yaml --sim ethereum/eest/execute-blobs
 ```
+
+**Note**: If Engine RPC is unavailable, blob transactions will be sent but `getBlobsV*` validation is skipped.
 
 See [Hive](../hive/index.md) for help installing and configuring Hive.
 
--- a/docs/running_tests/execute/hive.md
+++ b/docs/running_tests/execute/hive.md
@@ -13,7 +13,7 @@ The `blob_transaction_test` execute test spec sends blob transactions to a runni
 Tests can be run using:
 
 ```bash
-./hive --client besu --client-file ./configs/osaka.yaml --sim ethereum/eest/execute-blobs
+./hive --client besu --client-file ./configs/osaka.yaml --sim ethereum/eels/execute-blobs
 ```
 
 **Note**: If Engine RPC is unavailable, blob transactions will be sent but `getBlobsV*` validation is skipped.
--- a/docs/running_tests/execute/hive.md
+++ b/docs/running_tests/execute/hive.md
@@ -8,7 +8,7 @@ The `blob_transaction_test` execute test spec sends blob transactions to a runni
 
 - Blob transactions can be sent via `eth_sendRawTransaction`
 - Blob validation via `engine_getBlobsVX` endpoints (when Engine RPC available)
-- Automatic gas pricing for blob gas fees
+- Automatic gas pricing is used for the blob gas fees
 
 Tests can be run using:
 
--- a/docs/running_tests/execute/index.md
+++ b/docs/running_tests/execute/index.md
@@ -39,4 +39,12 @@ Reasoning behind the random generation of the sender keys is that one can execut
 
 The `fill` plugin will use a fixed and minimum gas price for all the transactions it uses for testing, but this is not possible with the `execute` plugin, as the gas price is determined by the current state of the network.
 
-At the moment, the `execute` plugin does not query the client for the current gas price, but instead uses a fixed increment to the gas price in order to avoid the transactions to be stuck in the mempool.
+The `execute` plugin queries the network for current gas prices and defaults to 1.5x the network price to ensure transaction inclusion. Gas prices can be overridden via command-line flags (`--default-gas-price`, `--default-max-fee-per-gas`, `--default-max-priority-fee-per-gas`).
+
+### Deferred EOA Funding
+
+EOAs are funded after gas prices are determined, enabling accurate balance calculations based on actual network conditions. This ensures sufficient funds are allocated for all test transactions.
+
+### Blob Transaction Support
+
+Blob transactions are fully supported in execute mode, including automatic gas pricing for blob gas fees and validation via `engine_getBlobsVX` endpoints when Engine RPC is available.
--- a/docs/running_tests/execute/index.md
+++ b/docs/running_tests/execute/index.md
@@ -47,4 +47,4 @@ EOAs are funded after gas prices are determined, enabling accurate balance calcu
 
 ### Blob Transaction Support
 
-Blob transactions are fully supported in execute mode, including automatic gas pricing for blob gas fees and validation via `engine_getBlobsVX` endpoints when Engine RPC is available.
+Blob transactions are fully supported in execute mode, including automatic gas pricing for blob gas fees and validation via `engine_getBlobsVX` endpoints when the Engine RPC is available.
--- a/docs/running_tests/execute/remote.md
+++ b/docs/running_tests/execute/remote.md
@@ -24,6 +24,41 @@ One last requirement is that the `--chain-id` flag is set to the chain id of the
 uv run execute remote --fork=Prague --rpc-endpoint=https://rpc.endpoint.io --rpc-seed-key 0x000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f --chain-id 12345
 ```
 
+## Test Accounts and Contracts
+
+Since `execute remote` does not control the blockchain where it runs the tests, and therefore cannot modify the genesis pre-allocation like `fill` does, all accounts and contracts need to be deployed via transactions sent to the network.
+
+These transactions are created from the seed or worker accounts provided via the command flags.
+
+When the test is executed, all instances of `pre.fund_eoa` and `pre.deploy_contract` calls, instead of placing the account directly in the `pre` object like `fill` does, they generate transactions that result in the creation of the accounts when executed on chain.
+
+The transactions are collected and only sent after the test function finishes execution. This is done in order to perform optimizations based on the transactions that the test requires to perform its verifications.
+
+One optimization is the deferred calculation of the funding amount for the EOA, which is calculated on the fly depending the test transactions that use the account as sender, and this amount is the minimum balance that the account would need in order for the transactions to be included given the current network gas prices.
+
+### Dry Run Mode
+
+Calculate minimum balance requirements without executing any transactions on chain:
+
+```bash
+uv run execute remote --fork=Prague --rpc-endpoint=https://rpc.endpoint.io --dry-run ./tests/prague/eip7702_set_code_tx/
+```
+
+This outputs the minimum balance needed and total gas consumption per test, useful for:
+- Estimating costs before execution
+- Verifying seed account has sufficient funds
+- Planning parallel execution funding requirements
+
+### Limit Gas Used By Tests
+
+A limit of the total gas consumption per test can be specified with the `--max-gas-per-test` flag:
+
+```bash
+uv run execute remote --fork=Prague --rpc-endpoint=https://rpc.endpoint.io --max-gas-per-test 30000000 --rpc-seed-key 0x... --chain-id 12345
+```
+
+Tests exceeding this limit will fail with an assertion error and will not send any transactions to the chain.
+
 ## Engine RPC Endpoint (Optional)
 
 By default, the `execute remote` command assumes that the execution client is connected to a beacon node and the chain progresses automatically. However, you can optionally specify an Engine RPC endpoint to drive the chain manually when new transactions are submitted.
--- a/docs/running_tests/execute/remote.md
+++ b/docs/running_tests/execute/remote.md
@@ -30,7 +30,7 @@ Since `execute remote` does not control the blockchain where it runs the tests, 
 
 These transactions are created from the seed or worker accounts provided via the command flags.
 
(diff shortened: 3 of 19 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -346,24 +356,21 @@ def deploy_contract(
         deploy_gas_limit = min(deploy_gas_limit * 2, 30_000_000)
         print(f"Deploying contract with gas limit: {deploy_gas_limit}")
 
-        self._refresh_sender_nonce()
```
LouisTsai-Csie, line 349: Removing this logic leads to issues for stateful benchmark testing and may also affect consensus testing.

We first construct a transaction (e.g., a contract deployment). This nonce assignment happens in the post-model initialization phase of the `Pydantic` model:

```python
if "nonce" not in self.model_fields_set and self.sender is not None:
    self.nonce = HexNumber(self.sender.get_nonce())
```

The `get_nonce` function assumes that the nonce always increases sequentially. However, this assumption breaks in several scenarios:

1. If two tests run sequentially, the framework assigns nonces 1 and 2. If the first transaction fails, the second transaction will remain pending indefinitely because its nonce will never become valid.
2. If multiple workers submit transactions concurrently and accidentally reuse the same nonce, the duplicate nonce will invalidate one or more of those transactions.

Point 1 is more critical, because we currently cannot ensure that all benchmark tests will run successfully. However, we still want the test suite to continue running without any stop.

This concern was previously raised and solved in [PR](https://github.com/ethereum/execution-spec-tests/pull/2284/files) from @kamilchodola , but removing the logic here would reintroduce the same problems and may also impact his tooling.

One potential solution here is to update the nonce value based on the network's state:

```
if "nonce" not in self.model_fields_set and self.sender is not None:
    self.nonce = HexNumber(elf._eth_rpc.get_transaction_count(self._sender, block_number="pending"))
```

I've tried this [approach](https://github.com/ethereum/execution-spec-tests/pull/2284#issuecomment-3384323198) before, but i encounter a circular import issue back at that time.

marioevz: Yep, agreed, (1) can be addressed by updating the nonce just a single time before running the test function so the nonce is up to date when the test starts requesting contracts and EOAs.

Number (2) I don't see how can it happen since all workers use different senders, otherwise concurrency would fail instantly.

Will fix (1), thanks for pointing out.

LouisTsai-Csie: Thanks for the change! I've reviewed again and this should be resolved.

LouisTsai-Csie:
Thanks for the enhancement! Could we also update the related docs so I can provide a reference to Kamil and Jochem?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
```diff
@@ -255,43 +290,178 @@ def transactions_per_block(
...
-    return request.config.getoption("default_max_fee_per_gas")
+    max_fee_per_gas = request.config.getoption("default_max_fee_per_gas")
+    if max_fee_per_gas is not None:
+        logger.info(
+            f"Using configured default max fee per gas: {max_fee_per_gas / 10**9:.2f} Gwei"
+        )
+    else:
+        logger.info(
+            "No default max fee per gas configured, will use network gas price * 1.5"
+        )
```
spencer-tb: Should these be `logger.debug`?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
```diff
@@ -255,43 +290,178 @@ def transactions_per_block(
...
+        "default_max_priority_fee_per_gas"
+    )
+    if max_priority_fee_per_gas is not None:
+        logger.info(
+            f"Using configured default max priority fee per gas: {max_priority_fee_per_gas / 10**9:.2f} Gwei"
+        )
+    else:
+        logger.info(
+            "No default max priority fee per gas configured, will use network max priority fee * 1.5"
+        )
```
spencer-tb: Similar here: `logger.debug`?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
```diff
@@ -255,43 +290,178 @@ def transactions_per_block(
...
+        "default_max_fee_per_blob_gas"
+    )
+    if max_fee_per_blob_gas is not None:
+        logger.info(
+            f"Using configured default max fee per blob gas: {max_fee_per_blob_gas / 10**9:.2f} Gwei"
+        )
+    else:
+        logger.info(
+            "No default max fee per blob gas configured, will use network blob base fee * 1.5"
+        )
```
spencer-tb, line 365: Similar here: `logger.debug`?

docs/running_tests/execute/hive.md
```diff
@@ -4,12 +4,20 @@ Tests can be executed on a local hive-controlled single-client network by runnin
 
 ## The `eest/execute-blobs` Simulator
 
-The `blob_transaction_test` execute test spec sends blob transactions to a running client in order to verify its `engine_getBlobsVX` endpoint behavior. These tests can be run using:
+The `blob_transaction_test` execute test spec sends blob transactions to a running client. Blob transactions are fully supported in execute mode:
+
+- Blob transactions can be sent via `eth_sendRawTransaction`
+- Blob validation via `engine_getBlobsVX` endpoints (when Engine RPC available)
+- Automatic gas pricing for blob gas fees
```
spencer-tb:
```suggestion
- Automatic gas pricing is used for the blob gas fees
```

docs/running_tests/execute/hive.md
```diff
@@ -4,12 +4,20 @@ Tests can be executed on a local hive-controlled single-client network by runnin
...
+
+Tests can be run using:
 
 ```bash
 ./hive --client besu --client-file ./configs/osaka.yaml --sim ethereum/eest/execute-blobs
```
spencer-tb:
```suggestion
./hive --client besu --client-file ./configs/osaka.yaml --sim ethereum/eels/execute-blobs
```

spencer-tb: We also need to merge this: https://github.com/ethereum/hive/pull/1365

docs/running_tests/execute/hive.md
```diff
@@ -4,12 +4,20 @@ Tests can be executed on a local hive-controlled single-client network by runnin
...
 ```
 
+**Note**: If Engine RPC is unavailable, blob transactions will be sent but `getBlobsV*` validation is skipped.
```
spencer-tb:
```suggestion
**Note**: If the Engine RPC is unavailable, blob transactions will be sent and `getBlobsV*` validation is skipped.
```

docs/running_tests/execute/index.md
```diff
@@ -39,4 +39,12 @@ Reasoning behind the random generation of the sender keys is that one can execut
...
-At the moment, the `execute` plugin does not query the client for the current gas price, but instead uses a fixed increment to the gas price in order to avoid the transactions to be stuck in the mempool.
+The `execute` plugin queries the network for current gas prices and defaults to 1.5x the network price to ensure transaction inclusion. Gas prices can be overridden via command-line flags (`--default-gas-price`, `--default-max-fee-per-gas`, `--default-max-priority-fee-per-gas`).
+
+### Deferred EOA Funding
+
+EOAs are funded after gas prices are determined, enabling accurate balance calculations based on actual network conditions. This ensures sufficient funds are allocated for all test transactions.
+
+### Blob Transaction Support
+
+Blob transactions are fully supported in execute mode, including automatic gas pricing for blob gas fees and validation via `engine_getBlobsVX` endpoints when Engine RPC is available.
```
spencer-tb:
```suggestion
Blob transactions are fully supported in execute mode, including automatic gas pricing for blob gas fees and validation via `engine_getBlobsVX` endpoints when the Engine RPC is available.
```

docs/running_tests/execute/remote.md
```diff
@@ -24,6 +24,42 @@ One last requirement is that the `--chain-id` flag is set to the chain id of the
...
+
+When the test is executed, all instances of `pre.fund_eoa` and `pre.deploy_contract` calls, instead of placing the account directly in the `pre` object like `fill` does, they generate transactions that result in the creation of the accounts when executed on chain.
+
+The transactions are collected and only sent after the test function finishes execution. This is done in order to perform optimizations based on the transactions that the test requires to perform its verifications.
+
+One optimization is the deferred calculation of the funding amount for the EOA, which is calculated on the fly depending the test transactions that use the account as sender, and this amount is the minimum balance that the account would need in order for the transactions to be included given the current network gas prices.
+
+### Dry Run Mode
+
+Calculate minimum balance requirements without executing any transactions on chain:
```
spencer-tb:
```suggestion
Dry run mode calculates the minimum balance required without executing any transactions on chain:
```

docs/running_tests/execute/remote.md
```diff
@@ -24,6 +24,42 @@ One last requirement is that the `--chain-id` flag is set to the chain id of the
...
+
+```bash
+uv run execute remote --fork=Prague --rpc-endpoint=https://rpc.endpoint.io --dry-run ./tests/prague/eip7702_set_code_tx/
+```
+
+This outputs the minimum balance needed and total gas consumption per test, useful for:
+
+- Estimating costs before execution
```
spencer-tb:
```suggestion
- Estimating gas costs before execution
```

docs/running_tests/execute/remote.md
```diff
@@ -24,6 +24,42 @@ One last requirement is that the `--chain-id` flag is set to the chain id of the
...
+- Verifying seed account has sufficient funds
```
spencer-tb:
```suggestion
- Verifying the seed account has sufficient funds
```

docs/running_tests/execute/remote.md
```diff
@@ -24,6 +24,42 @@ One last requirement is that the `--chain-id` flag is set to the chain id of the
 uv run execute remote --fork=Prague --rpc-endpoint=https://rpc.endpoint.io --rpc-seed-key 0x000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f --chain-id 12345
 ```
 
+## Test Accounts and Contracts
+
+Since `execute remote` does not control the blockchain where it runs the tests, and therefore cannot modify the genesis pre-allocation like `fill` does, all accounts and contracts need to be deployed via transactions sent to the network.
+
+These transactions are created from the seed or worker accounts provided via the command flags.
+
+When the test is executed, all instances of `pre.fund_eoa` and `pre.deploy_contract` calls, instead of placing the account directly in the `pre` object like `fill` does, they generate transactions that result in the creation of the accounts when executed on chain.
```
spencer-tb:
```suggestion
When the test is executed, all `pre.fund_eoa` and `pre.deploy_contract` calls generate transactions that create the accounts on chain, instead of placing them directly in the `pre` object like `fill` does.
```

spencer-tb (approved):
LGTM! Minor pedantic changes to docs and logging level now. Dry run works really well for. me. The speed up with load <- loadscope is insane :)

spencer-tb:
Should we update the title -> `enhance(test-cli-execute): ...`?

Where we have the following with the most important A label in the title:
```
<C-tag>(<A-tag>): <description>
```

I haven't really pushed hard but I think its nice to start with the efforts that went into adding the labels. I should have mentioned in the STEEL meeting 😆

PR #1825, opened 2025-11-30
test(eip7928): add EXTCODECOPY OOG memory expansion BAL test

qu0b:
## Summary

Tests verify that target address should NOT appear in BAL when EXTCODECOPY fails due to OOG at memory expansion

## Bug Description

Nethermind's `AddAccountRead()` is called BEFORE checking gas for memory expansion. When OOG occurs, the account read entry persists in BAL because it's not journaled.


fselmo (approved):
@qu0b this looks great to me, thanks for adding this!

I added a commit on top of yours that expands on what was already an existing test for extcodecopy with OOG. This leaves the `boundary` case and the `oog_at_cold_access` case sort of the same (gas_left = 1) but from different calculations (one has non mem cost but subtracts `1` from total cost so no acct load can happen, the other has mem cost and leaves calculation also with `1` gas less than needed). I am good with this from an implementation check POV since one has mem expansion and one doesn't. Any thoughts before I merge?

PR #1827, opened 2025-12-01
enhance(ci): add fast benchmark artifact workflow

spencer-tb:
## 🗒️ Description

A follow up request from the benchmarking meeting in Devconnect Argentina. Adds new CI workflow `artifact_benchmark_fast.yaml` that builds benchmark fixtures with 100M gas on push to `forks/osaka` and `forks/amsterdam` branches, only for changes relevant to benchmark tests.

Renames release workflows with `release_` prefix for better organization (`fixture_full_release.yaml` -> `release_fixture_full.yaml`, etc.)

To download the artifact we should be able to run the following below, to get the latest artifact:
```
gh run download -R ethereum/execution-specs -n fixtures_benchmark_fast
```

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


LouisTsai-Csie:
Nice! Hope we could present this during gas lighting call!

.github/workflows/artifact_benchmark_fast.yaml
danceratopz, line 1: The renaming of the other fixture release workflows [release_fixture_feature.yaml](https://github.com/ethereum/execution-specs/pull/1827/files#diff-ee5953c97e43890ef5598d09c9738e0aa885e1277643a62966c56f4c60fb7095)
[release_fixture_full.yaml](https://github.com/ethereum/execution-specs/pull/1827/files#diff-2c21869b3936cb9f71a931cae6268d546ab8689284a51299f4e566114729bddb) makes sense 👍 

However, the naming of this workflow, `.github/workflows/release_benchmark_fast.yaml`, seems inconsistent with the other workflows, as this workflow doesn't generate a tagged release, rather just an artifact.

.github/workflows/artifact_benchmark_fast.yaml
```diff
@@ -0,0 +1,50 @@
+name: Build Benchmark Fixtures
+
+on:
+  push:
+    branches:
+      # List specific branches for now.
+      - "forks/osaka"
+      - "forks/amsterdam"
```
danceratopz, line 7: I think we should only generate for the default fork, especially as we only generate benchmark fixtures for Prague. I guess this will be Amsterdam very soon :)

.github/workflows/release_benchmark_fast.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - "tests/prague/eip2537_bls_12_381_precompiles/spec.py"
+      - "tests/osaka/eip7951_p256verify_precompiles/spec.py"
+      # Benchmark related EEST framework files 
+      - "packages/testing/src/execution_testing/benchmark/**"
+      - "packages/testing/src/execution_testing/specs/**"
+      - "packages/testing/src/execution_testing/test_types/**"
+      - "packages/testing/src/execution_testing/fixtures/**"
+      - "packages/testing/src/execution_testing/forks/**"
+      - "packages/testing/src/execution_testing/vm/**"
+      - "packages/testing/src/execution_testing/cli/pytest_commands/fill.py"
```
danceratopz: This could trigger a lot of unnecessary builds, but happy to leave as-is for now and see how it goes.

danceratopz (commented):
Thanks for adding this, couple of comments below! 

I do  think this PR should be considered blocked by:
- https://github.com/ethereum/execution-specs/pull/1804

I'm slightly concerned that relying on latest artifacts (as-is) will make it difficult to trace issues in downstream infra. One idea to mitigate this would be to enable artifact downloading in `consume cache` (which should then be used by Kamil, etc.) and enable logging of the execution-specs git ref (resp. URL) corresponding to the artifact being downloaded. Currently, this ref is already reported in the fixture JSON and in `fixtures/.meta/fixtures.ini` but something even more obvious wouldn't hurt. 

E.g.:
- Q "I couldn't run the latest arfifact, is it broken?" 
- A "What's the execution-specs ref in the download log for me to verify locally?

.github/workflows/artifact_benchmark_fast.yaml
spencer-tb: Updated!

.github/workflows/release_benchmark_fast.yaml
spencer-tb: Updated. Removed some of these.

.github/workflows/artifact_benchmark_fast.yaml
spencer-tb: Agree here. There is not an automatic option for default fork here so leaving as is for now.

spencer-tb:
> I'm slightly concerned that relying on latest artifacts (as-is) will make it difficult to trace issues in downstream infra. One idea to mitigate this would be to enable artifact downloading in `consume cache` (which should then be used by Kamil, etc.) and enable logging of the execution-specs git ref (resp. URL) corresponding to the artifact being downloaded. Currently, this ref is already reported in the fixture JSON and in `fixtures/.meta/fixtures.ini` but something even more obvious wouldn't hurt.

As far as I understand the git ref is traceable via:
```
gh run list -R ethereum/execution-specs -w "Build Benchmark Fixtures" --limit 5
```

Then can be downloaded from a specific run if needed:
```
gh run download -R ethereum/execution-specs <run-id> -n fixtures_benchmark_fast
```

I agree adding more obvious logging in consume cache would be a nice improvement, but I'd
prefer to address that in a follow-up PR!

marioevz (approved):
Amazing, thanks!

PR #1829, opened 2025-12-01
test(eip7928): add cross-block precompile state leak test

qu0b:
This test recreates the issue where nethermind would include the 0x03 pre-compile in its block level access list even though the pre-compile is never called in that block, but instead is called in a parent block.


fselmo (approved):
This lgtm. Really nice find 🔥

PR #1841, opened 2025-12-03
refactor(spec-specs): Refactor state changes and frame hierarchy

fselmo:
## 🗒️ Description

- Refactor state tracking frames to be part of `BlockEnvironment`, `TransactionEnvironment`, and `Message`. This removes much of the unnecessary complexity related to parent frame tracking.
- Remove `state_changes` from `State` functions, opt to capture changes before and after (from comments here)
- General cleanup in spec code - Note: builder and rlp_utils have not been cleaned up here

## 🔗 Related Issues or PRs

#1836 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2025-12-03 at 12 48 20

### DIFF
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -69,12 +69,13 @@ from .state_tracker import (
     commit_transaction_frame,
     create_child_frame,
     get_block_access_index,
-    handle_in_transaction_selfdestruct,
     increment_block_access_index,
     merge_on_success,
     normalize_balance_changes_for_transaction,
     track_address,
     track_balance_change,
+    track_nonce_change,
+    track_selfdestruct,
 )
 from .transactions import (
     AccessListTransaction,
@@ -647,7 +648,11 @@ def process_system_transaction(
         authorizations=(),
         index_in_block=None,
         tx_hash=None,
+        state_changes=system_tx_state_changes,
     )
+
+    # Create call frame as child of tx frame
+    call_frame = create_child_frame(tx_env.state_changes)
 
     system_tx_message = Message(
         block_env=block_env,
@@ -667,14 +672,14 @@ def process_system_transaction(
         accessed_storage_keys=set(),
         disable_precompiles=False,
         parent_evm=None,
-        transaction_state_changes=system_tx_state_changes,
+        state_changes=call_frame,
     )
 
     system_tx_output = process_message_call(system_tx_message)
 
-    # Merge system transaction changes back to block frame
+    # Commit system transaction changes to block frame
     # System transactions always succeed (or block is invalid)
-    merge_on_success(system_tx_state_changes)
+    commit_transaction_frame(tx_env.state_changes)
 
     return system_tx_output
 
@@ -909,7 +914,9 @@ def process_transaction(
     # The frame will read the current block_access_index from the block frame
     increment_block_access_index(block_env.block_state_changes)
     tx_state_changes = create_child_frame(block_env.block_state_changes)
+    block_access_index = get_block_access_index(block_env.block_state_changes)
 
+    # Capture coinbase pre-balance for net-zero filtering
     coinbase_pre_balance = get_account(
         block_env.state, block_env.coinbase
     ).balance
@@ -947,16 +954,30 @@ def process_transaction(
     effective_gas_fee = tx.gas * effective_gas_price
 
     gas = tx.gas - intrinsic_gas
-    increment_nonce(block_env.state, sender, tx_state_changes)
+
+    # Track sender nonce increment
+    increment_nonce(block_env.state, sender)
+    sender_nonce_after = get_account(block_env.state, sender).nonce
+    track_nonce_change(
+        tx_state_changes, sender, U64(sender_nonce_after), block_access_index
+    )
+
+    # Track sender balance deduction for gas fee
+    sender_balance_before = get_account(block_env.state, sender).balance
+    track_address(tx_state_changes, sender)
+    capture_pre_balance(tx_state_changes, sender, sender_balance_before)
 
     sender_balance_after_gas_fee = (
         Uint(sender_account.balance) - effective_gas_fee - blob_gas_fee
     )
     set_account_balance(
-        block_env.state,
+        block_env.state, sender, U256(sender_balance_after_gas_fee)
+    )
+    track_balance_change(
+        tx_state_changes,
         sender,
         U256(sender_balance_after_gas_fee),
-        tx_state_changes,
+        block_access_index,
     )
 
     access_list_addresses = set()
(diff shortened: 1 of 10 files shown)


src/ethereum/forks/amsterdam/vm/__init__.py
```diff
@@ -142,7 +143,8 @@ class Message:
     accessed_storage_keys: Set[Tuple[Address, Bytes32]]
     disable_precompiles: bool
     parent_evm: Optional["Evm"]
-    transaction_state_changes: StateChanges
+    is_create: bool = False
```
gurukamath: I think in case of `Message` and `TransactionEnvironment`, I'd prefer not to rely on default values of the `dataclass`. I think the specs are more readable when we explicitly define all the attributes of these objects at the time of instantiation. Same for `BlockEnvironment`

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -467,37 +530,9 @@ def commit_transaction_frame(tx_frame: StateChanges) -> None:
...
-        parent_frame.storage_reads.add((address, key))
-
-    # Note: balance_changes, nonce_changes, and code_changes are NOT
-    # merged on failure - they are discarded
-
-
 def create_child_frame(parent: StateChanges) -> StateChanges:
     """
-    Create a child frame for nested execution.
+    Create a child frame linked to the given parent.
```
gurukamath, line 478: The other thing we need to consider here is if we are ok with the `_block_access_index` defaulting to `0` for every child frame. I think every child frame should get the same `_block_access_index` as it parent.

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -263,10 +209,19 @@ def process_create_message(message: Message) -> Evm:
     # added to SELFDESTRUCT by EIP-6780.
     mark_account_created(state, message.current_target)
 
-    parent_frame = get_parent_frame(message)
-    create_frame = create_child_frame(parent_frame)
+    block_access_index = get_block_access_index(
+        message.block_env.block_state_changes
```
gurukamath: Maybe rename this to `block_env.state_changes` to stay consistent with the `tx_env.state_changes` and `message.state_changes`

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -263,10 +209,19 @@ def process_create_message(message: Message) -> Evm:
...
+    )
+
+    increment_nonce(state, message.current_target)
+    nonce_after = get_account(state, message.current_target).nonce
+    track_nonce_change(
+        message.state_changes,
```
gurukamath, line 211: If we make the `_block_access_index` directly available in every child frame, we would not need to pass it as a function parameter

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -280,19 +235,29 @@ def process_create_message(message: Message) -> Evm:
...
-            merge_on_failure(create_frame)
+            merge_on_failure(message.state_changes)
             evm.gas_left = Uint(0)
             evm.output = b""
             evm.error = error
         else:
-            set_code(
-                state, message.current_target, contract_code, create_frame
+            # Track code change for contract creation
+            pre_code = get_account(state, message.current_target).code
```
gurukamath: I think this will always be b`` since we check if the target has code or nonce before going ahead with contract creation

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -1129,20 +1151,20 @@ def increase_recipient_balance(recipient: Account) -> None:
...
 
     # EIP-7928: Normalize balance changes after all withdrawals
     # Filters out net-zero changes
-
     normalize_balance_changes_for_transaction(
         block_env.block_state_changes,
-        BlockAccessIndex(
-            get_block_access_index(block_env.block_state_changes)
-        ),
+        block_access_index,
```
gurukamath: If we intend to call this function beyond a transaction processing, might be best to rename it to `normalize_balance_changes` instead

src/ethereum/forks/amsterdam/state.py
```diff
@@ -517,18 +509,22 @@ def move_ether(
     sender_address: Address,
     recipient_address: Address,
     amount: U256,
-    state_changes: StateChanges,
 ) -> None:
```
gurukamath, line 512: Note to self: Discuss the reasoning behind separating this out.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -356,43 +406,35 @@ def merge_on_success(child_frame: StateChanges) -> None:
...
-            capture_pre_code(parent_frame, addr, code)
-
     # Merge storage operations, filtering noop writes
     parent_frame.storage_reads.update(child_frame.storage_reads)
     for (addr, key, idx), value in child_frame.storage_writes.items():
         # Only merge if value actually changed from pre-state
-        if (addr, key) in child_frame.pre_storage:
-            if child_frame.pre_storage[(addr, key)] != value:
+        if (addr, key) in tx_frame.pre_storage:
+            if tx_frame.pre_storage[(addr, key)] != value:
```
gurukamath: I am thinking of the following scenario. `tx_frame.pre_storage` is 1. The `parent_frame` of the current frame changes it to 2 and the current frame changes it back to 1. Would the changes made by the current frame be ignored because the value is the same as in the `tx_frame` but the changes in the `parent_frame` be captured. Thus leading us to believe the overall change was `1->2` while in reality it was `1->2->1`

fselmo: Yes! I think this is definitely broken and I wrote a test to make sure [here](https://github.com/ethereum/execution-specs/pull/1854) :). I think this should be simplified quite a bit. The children should just merge all their changes to the parent's changes to be honest, to update the state of the parent after processing the child frame, and the net-zero filtering should only be done at the `commit_transaction_frame` level. That's the reason that has its own method anyhow since the only difference between the top-level transaction frame committing and `merge_on_success` for child frames is that net-zero changes need to be filtered out.

src/ethereum/forks/amsterdam/state_tracker.py
fselmo: Agreed. I also think this can have the underscore removed? I will include this here.

src/ethereum/forks/amsterdam/fork.py
fselmo: I think we can normalize all net-zero filtering here once we remove that from `merge_on_success` as I mentioned before. And yeah, agreed, this is called on withdrawals too so the naming should be better here.

src/ethereum/forks/amsterdam/state_tracker.py
fselmo: I did check this before and after and it was broken but now passes those nested tests. Good tests to have 👍🏼

src/ethereum/forks/amsterdam/vm/__init__.py
fselmo: I fixed this for `is_create`. Do you think the same for `state_changes` or do you think this is fine to instantiate with a blank slate there?

gurukamath (approved):

PR #1846, opened 2025-12-04
eip7928: add SELFDESTRUCT OOG BAL test

qu0b:
Found this issue when running a kurtosis network. This test currently fails against Besu but other clients tested (nethermind, reth) pass this test.

image


fselmo (approved):
lgtm! I added a change to the spec (was wrong, need to check gas before account access) and refactored the test to parametrize around the gas boundary. I also moved these `self_destruct` tests into a the appropriate file within `test_block_access_lists_opcodes.py`.

@qu0b lmk if these changes look good to you and I'll merge.

Issue #1695, opened 2025-10-27
Refactor benchmark test for Osaka support


LouisTsai-Csie:
### Benchmark Test Status Update

After introducing the benchmark test wrapper, most benchmark tests now support EIP-7825, though a few still require attention.

Scope: Compute Scenario

- [x] `tests/benchmark/compute/instruction/test_account_query.py::test_extcode_ops` @jsign 
- [x] `tests/benchmark/compute/instruction/test_account_query.py::test_ext_account_query_cold` @jsign - https://github.com/ethereum/execution-specs/pull/1862
- [x] `tests/benchmark/compute/instruction/test_storage.py::test_storage_access_cold` @jsign 
- [x] `tests/benchmark/compute/instruction/test_storage.py::test_storage_access_warm` @jsign - https://github.com/ethereum/execution-specs/pull/1915
- [x] `tests/benchmark/compute/instruction/test_system.py::test_xcall` @jsign - https://github.com/ethereum/execution-specs/pull/1850
- [x] `tests/benchmark/compute/instruction/test_system.py::test_selfdestruct_existing` @jsign - https://github.com/ethereum/execution-specs/pull/1904
- [x] `tests/benchmark/compute/instruction/test_system.py::test_selfdestruct_created` @jsign - https://github.com/ethereum/execution-specs/pull/1906
- [x] `tests/benchmark/compute/instruction/test_system.py::test_selfdestruct_initcode` @jsign - https://github.com/ethereum/execution-specs/pull/1914
- [ ] `tests/benchmark/compute/instruction/test_tx_context.py::test_blobhash` @LouisTsai-Csie 
- [x] `tests/benchmark/compute/scenario/test_transaction_types.py::test_auth_transaction` @LouisTsai-Csie - https://github.com/ethereum/execution-specs/pull/1927

Scope: Stateful Scenario

- [ ] `tests/benchmark/stateful/bloatnet/test_multi_opcode.py::test_bloatnet_balance_extcodesize`
- [ ] `tests/benchmark/stateful/bloatnet/test_multi_opcode.py::test_bloatnet_balance_extcodecopy`
- [ ] `tests/benchmark/stateful/bloatnet/test_multi_opcode.py::test_bloatnet_balance_extcodehash`
- [ ] `tests/benchmark/stateful/bloatnet/test_multi_opcode.py::test_mixed_sload_sstore`
[...]

PR #1850, opened 2025-12-05
feat(benchmarks): fix bytecode attack for CALL-like opcodes to work in Osaka

jsign:
## 🗒️ Description
This PR fixes the `test_xcall` benchmarks to work with Osaka. `test_xcall` has a CALL-like variant of the bytecode attack exploiting unchunkified code, highly relevant for zkVM worst-cases.

I think the solution should feel quite natural — I’ll add some review comments.

I checked this is doing what we want with two independent checks:

1. Hack a bit the Geth codebase to inspect how many distinct bytecodes were accessed during the block execution.
2. Actually running the worst-case block for zkVM execution and inspecting that indeed the witness to be sent to the guest program indeed is blowed up — since bytecode must be part of the input to the guest program.

For 1. and `--block-gas-limit 30M`, my Geth reports accessing ~260MiB of bytecodes were the upper limit is ~280MiB (30M/2600 — but this isn’t accounting for glue opcodes, intrinsic costs of splitting in txs, etc). So this looks fine.

For 2. and `--block-gas-limit 30M`, I also confirmed it really blows up the guest program witness, which is prob the best ultimate confirmation this is doing what we wanted. i.e: the json-encoded files have ~524MiB size which makes sense since bytes are hex encoded (with other things).

## 🔗 Related Issues or PRs

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -55,7 +55,6 @@ def test_xcall(
     pre: Alloc,
     fork: Fork,
     opcode: Op,
-    env: Environment,
     gas_benchmark_value: int,
     tx_gas_limit: int,
 ) -> None:
@@ -99,6 +98,8 @@ def test_xcall(
         num_contracts_per_tx = int(tx_gas_limit * 0.9) // (
             gas_costs.G_CODE_DEPOSIT_BYTE * max_contract_size
         )
+        if num_contracts_per_tx == 0:
+            pytest.skip("tx_gas_limit too low to deploy max-size contract")
         setup_txs = math.ceil(num_contracts / num_contracts_per_tx)
 
         contracts_deployment_txs = []
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -109,16 +109,33 @@ def test_xcall(
             "during the setup phase of this test."
         )
 
-    initcode, factory_address, factory_caller_address = deploy_max_contract_factory(pre, fork)
+    initcode, factory_address, factory_caller_address = (
+        deploy_max_contract_factory(pre, fork)
+    )
 
     with TestPhaseManager.setup():
-        contracts_deployment_tx = Transaction(
-            to=factory_caller_address,
-            gas_limit=env.gas_limit,
-            gas_price=10**6,
-            data=Hash(num_contracts),
-            sender=pre.fund_eoa(),
+        # We require deploying num_contracts by calling the factory contract. Each call can only
+        # use up to the transaction gas limit cap, thus we need to split the deployment into
+        # multiple transactions. The factory contract uses a storage slot to keep track of the last
+        # seed used to generate bytecodes, so it is safe to call it multiple times in different txs.
+        num_contracts_per_tx = (
+            3  # TODO: Try generalizing for any tx gas limit cap. For 17M is 3.
         )
+        num_contract_creation_txs = math.ceil(
+            num_contracts / num_contracts_per_tx
+        )
+
+        contracts_deployment_txs = []
+        for _ in range(num_contract_creation_txs):
+            contracts_deployment_txs.append(
+                Transaction(
+                    to=factory_caller_address,
+                    gas_limit=fork.transaction_gas_limit_cap(),
+                    gas_price=10**6,
+                    data=Hash(num_contracts_per_tx),
+                    sender=pre.fund_eoa(),
+                )
+            )
 
     post = {}
     deployed_contract_addresses = []
@@ -174,16 +191,17 @@ def test_xcall(
         pre=pre,
         post=post,
         blocks=[
-            Block(txs=[contracts_deployment_tx]),
-            Block(txs=[opcode_tx]),
+            Block(txs=contracts_deployment_txs),
+            # Block(txs=[opcode_tx]),
         ],
         exclude_full_post_state_in_output=True,
     )
 
+
 def deploy_max_contract_factory(
     pre: Alloc,
     fork: Fork,
-) -> tuple[Bytecode, Address, Address]: 
+) -> tuple[Bytecode, Address, Address]:
     max_contract_size = fork.max_code_size()
 
     # The initcode will take its address as a starting point to the input to
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -59,19 +59,16 @@ def test_xcall(
     gas_benchmark_value: int,
 ) -> None:
     """Benchmark a system execution where a single opcode execution."""
-    # The attack gas limit is the gas limit which the target tx will use The
-    # test will scale the block gas limit to setup the contracts accordingly to
-    # be able to pay for the contract deposit. This has to take into account
-    # the 200 gas per byte, but also the quadratic memory expansion costs which
-    # have to be paid each time the memory is being setup
+    # The attack gas limit represents the transaction gas limit cap or
+    # the block gas limit. If eip-7825 is applied, the test will create
+    # multiple transactions for contract deployment. It should account
+    # for the 200 gas per byte cost and the quadratic memory-expansion
+    # costs, which must be paid each time memory is initialized.
     attack_gas_limit = gas_benchmark_value
     max_contract_size = fork.max_code_size()
 
     gas_costs = fork.gas_costs()
     tx_gas_limit_cap = fork.transaction_gas_limit_cap()
-    assert tx_gas_limit_cap is not None, (
-        "This benchmark requires a tx gas limit cap"
-    )
 
     # Calculate the absolute minimum gas costs to deploy the contract This does
     # not take into account setting up the actual memory (using KECCAK256 and
@@ -126,17 +123,24 @@ def test_xcall(
         # If this estimation is incorrect in the future (i.e. tx gas limit cap)
         # is increased or cost per byte, the post-state check will detect it
         # and can be adjusted with a more complex formula.
-        num_contracts_per_tx = tx_gas_limit_cap // (
-            gas_costs.G_CODE_DEPOSIT_BYTE * max_contract_size
+        num_contracts_per_tx = (
+            tx_gas_limit_cap
+            // (gas_costs.G_CODE_DEPOSIT_BYTE * max_contract_size)
+            if tx_gas_limit_cap
+            else num_contracts
         )
-        attack_txs = math.ceil(num_contracts / num_contracts_per_tx)
+        attack_txs = (
+            math.ceil(num_contracts / num_contracts_per_tx)
+            if num_contracts_per_tx
+            else 1
+        )
 
         contracts_deployment_txs = []
         for _ in range(attack_txs):
             contracts_deployment_txs.append(
                 Transaction(
                     to=factory_caller_address,
-                    gas_limit=tx_gas_limit_cap,
+                    gas_limit=tx_gas_limit_cap or env.gas_limit,
                     data=Hash(num_contracts_per_tx),
                     sender=pre.fund_eoa(),
                 )
@@ -176,13 +180,17 @@ def test_xcall(
     attack_address = pre.deploy_contract(code=attack_code)
 
     with TestPhaseManager.execution():
-        full_txs = attack_gas_limit // tx_gas_limit_cap
-        remainder = attack_gas_limit % tx_gas_limit_cap
+        full_txs = (
+            attack_gas_limit // tx_gas_limit_cap if tx_gas_limit_cap else 1
+        )
+        remainder = (
+            attack_gas_limit % tx_gas_limit_cap if tx_gas_limit_cap else 0
+        )
 
         num_targeted_contracts_per_full_tx = (
             # Base available gas:
             # TX_GAS_LIMIT - intrinsic - (out of loop MSTOREs)
-            tx_gas_limit_cap
+            (tx_gas_limit_cap or attack_gas_limit)
             - intrinsic_gas_cost_calc()
             - gas_costs.G_VERY_LOW * 4
         ) // loop_cost
@@ -192,7 +200,7 @@ def test_xcall(
             opcode_txs.append(
                 Transaction(
                     to=attack_address,
-                    gas_limit=tx_gas_limit_cap,
+                    gas_limit=tx_gas_limit_cap or attack_gas_limit,
                     data=Hash(contract_start_index),
                     sender=pre.fund_eoa(),
                 )
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -98,7 +98,7 @@ def test_xcall(
         # complexity and potential brittleness.
         num_contracts_per_tx = int(tx_gas_limit * 0.9) // (
             gas_costs.G_CODE_DEPOSIT_BYTE * max_contract_size
-        ) 
+        )
         setup_txs = math.ceil(num_contracts / num_contracts_per_tx)
 
         contracts_deployment_txs = []
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -91,7 +91,7 @@ def test_xcall(
         + gas_costs.G_COLD_ACCOUNT_ACCESS  # Opcode cost
         + 30  # ~Gluing opcodes
     )
-    # Calculate the number of contracts to be targeted
+    # Calculate an upper bound of the number of contracts to be targeted
     num_contracts = (
         # Base available gas = GAS_LIMIT - intrinsic - (out of loop MSTOREs)
         attack_gas_limit - intrinsic_gas_cost_calc() - gas_costs.G_VERY_LOW * 4
@@ -121,12 +121,10 @@ def test_xcall(
         num_contracts_per_tx = (
             3  # TODO: Try generalizing for any tx gas limit cap. For 17M is 3.
         )
-        num_contract_creation_txs = math.ceil(
-            num_contracts / num_contracts_per_tx
-        )
+        attack_txs = math.ceil(num_contracts / num_contracts_per_tx)
 
         contracts_deployment_txs = []
-        for _ in range(num_contract_creation_txs):
+        for _ in range(attack_txs):
             contracts_deployment_txs.append(
                 Transaction(
                     to=factory_caller_address,
@@ -138,7 +136,6 @@ def test_xcall(
             )
 
     post = {}
-    deployed_contract_addresses = []
     for i in range(num_contracts):
         deployed_contract_address = compute_create2_address(
             address=factory_address,
@@ -146,7 +143,6 @@ def test_xcall(
             initcode=initcode,
         )
         post[deployed_contract_address] = Account(nonce=1)
-        deployed_contract_addresses.append(deployed_contract_address)
 
     attack_call = Bytecode()
     if opcode == Op.EXTCODECOPY:
@@ -162,7 +158,7 @@ def test_xcall(
         # 0xFF+[Address(20bytes)]+[seed(32bytes)]+[initcode keccak(32bytes)]
         Op.MSTORE(0, factory_address)
         + Op.MSTORE8(32 - 20 - 1, 0xFF)
-        + Op.MSTORE(32, 0)
+        + Op.MSTORE(32, Op.CALLDATALOAD(0))
         + Op.MSTORE(64, initcode.keccak256())
         # Main loop
         + While(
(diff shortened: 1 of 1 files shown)


tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +109,127 @@ def test_xcall(
             "during the setup phase of this test."
         )
 
+    initcode, factory_address, factory_caller_address = (
+        _deploy_max_contract_factory(pre, fork)
+    )
```
jsign, line 90: I extracted the deployment of the factory contract to a separate function, mainly to avoid such a big method here which hurts readability, but mostly because I think for the upcoming `test_extcode_ops` I'll fix, that code is duplicated so I can re-use the function and we can delete code.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +109,127 @@ def test_xcall(
...
+
+    # Deploy num_contracts via multiple txs (each capped by tx gas limit).
```
jsign, line 92: In the setup phase, we can't create all the required bytecodes in a single tx because this is disallowed by Osaka (unfortuantely for us here).

This means we have to split the creation in multiple transactions. This will have an unavoidable hit in filling performance, since we actually used the previous strategy for that reason.

For 60M it takes ~2m in my machine, which maybe isn't as bad -- but still I'm not sure we can workaround this with Osaka limitations.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +109,127 @@ def test_xcall(
...
+        for _ in range(attack_txs):
+            contracts_deployment_txs.append(
+                Transaction(
+                    to=factory_caller_address,
+                    gas_limit=fork.transaction_gas_limit_cap(),
+                    gas_price=10**6,
+                    data=Hash(num_contracts_per_tx),
+                    sender=pre.fund_eoa(),
+                )
+            )
```
jsign, line 114: I preferred to go heavy in comments for future maintainability.

But basically we do a rough estimation of how many 24KiB contracts we can create per tx, and then do that. Note I do `ceil` in `attack_txs` instead of trying to do precise creation, just because it doesn't hurt plus it avoids complexity in this code. Since this is the setup phase and the tx is gas capped, I think it is okay.

The factory contract has a new CALLDATA parameter which is how many contracts to create. Although we might only be interested in this being always the max, I think it is better to leave it generic so we can try reusing it as much as possible. It just changes the contract creation loop count to whatever CALLDATALOAD says.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +109,127 @@ def test_xcall(
...
+
+    attack_call = Bytecode()
+    if opcode == Op.EXTCODECOPY:
+        attack_call = Op.EXTCODECOPY(
+            address=Op.SHA3(32 - 20 - 1, 85), dest_offset=96, size=1000
+        )
+    else:
+        # For the rest of the opcodes, we can use the same generic attack call
+        # since all only minimally need the `address` of the target.
+        attack_call = Op.POP(opcode(address=Op.SHA3(32 - 20 - 1, 85)))
```
jsign, line 133: All this code is the same as before -- just that the diff is a bit unfortunate.

Something I noticed is that this has this `opcode == Op.EXTCODECOPY` which shouldn't be relevant for this method (look at test parametrization). Originally when I created this benchmark, the same test name was used for CALL-like opcodes and also EXTCODESIZE... but I think somebody split them after and maybe this was left here?

I'll leave this as is for now, and when I fix `test_extcode_ops` I'll try to re-use the logic since I'm pretty sure it was repeated there when it was splitted.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +109,127 @@ def test_xcall(
...
+        # 0xFF+[Address(20bytes)]+[seed(32bytes)]+[initcode keccak(32bytes)]
+        Op.MSTORE(0, factory_address)
+        + Op.MSTORE8(32 - 20 - 1, 0xFF)
+        + Op.MSTORE(32, Op.CALLDATALOAD(0))
+        + Op.MSTORE(64, initcode.keccak256())
+        # Main loop
+        + While(
+            body=attack_call + Op.MSTORE(32, Op.ADD(Op.MLOAD(32), 1)),
+        )
+    )
```
jsign, line 147: This is the same attack code as before but with a twist. The previous version started with `seed=0` and simply ran the loop until it ran out of  gas, since we could do all in a singe tx.

Now we have to split the attack in multiple txs, thus I allow reading the seed from CALLDATALOAD(0) so we can create the attack txs in a way that makes sense.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +109,127 @@ def test_xcall(
...
+
+    if len(attack_code) > max_contract_size:
+        # TODO: A workaround could be to split the opcode code into multiple
+        # contracts and call them in sequence.
+        raise ValueError(
+            f"Code size {len(attack_code)} exceeds maximum "
+            f"code size {max_contract_size}"
+        )
+    attack_address = pre.deploy_contract(code=attack_code)
```
jsign: Unchanged code.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +109,127 @@ def test_xcall(
...
+        post=post,
+        blocks=[
+            Block(txs=contracts_deployment_txs),
+            Block(txs=opcode_txs),
+        ],
+        exclude_full_post_state_in_output=True,
+    )
+
+
+def _deploy_max_contract_factory(
```
jsign, line 193: The way the contract factory was created remains unchanged.

Note that the `factory_code` already used storage slot 0 as the internal seed. This was quite nice since it means that when we call the factory mutliple times now, it keeps creating the next bytecode in a coherent way from the previous tx.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -67,6 +68,10 @@ def test_xcall(
     max_contract_size = fork.max_code_size()
 
     gas_costs = fork.gas_costs()
+    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
+    assert tx_gas_limit_cap is not None, (
+        "This benchmark requires a tx gas limit cap"
+    )
```
jsign: Could we mark this test for Osaka forward only? (Or maybe that already happens since AFAIK in all the benchmark releases only Prague was filled, since I'm not interested we want backwards performance runs. For zkVM also makes sense that previous forks aren't interesting for this attack).

I think we could make it work backwards, it would just blow up a the logic below reg tx preparations for setup and attack. This test is already quite complex so not sure it is worth it.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +113,128 @@ def test_xcall(
...
+        if remainder > 0:
+            opcode_txs.append(
+                Transaction(
+                    to=attack_address,
+                    gas_limit=remainder,
+                    gas_price=10**9,
+                    data=Hash(contract_start_index),
+                    sender=pre.fund_eoa(),
+                )
+            )
```
jsign: TL;DR: creating the attack txs to execute in the same block, such that none surpas the transaction gas limit cap. In contract_start_index we keep the offset we pass as CALLDATALOAD for the attack contract that I explained in my previous comment.

jochem-brouwer: Wanted to mention this (did not check other logic): this looks almost similar to the `split_transactions` logic (https://github.com/ethereum/execution-specs/blob/843178f3bd6764158de095c088fcb4b3658f01dc/packages/testing/src/execution_testing/specs/benchmark.py#L344)  of the `BenchmarkTestFiller`, however - that split txs would just split the tx (without calldata) up in multiple. Which we cannot do here due to the test logic.
@LouisTsai-Csie we could maybe think of adding cases to the split txs logic here? I feel that inserting calldata is something we will see in other benchmark tests often (for instance as a "pointer" to where a certain tx should pick up where the other tx has exited - we could also do this in the contract code itself, but this means spending gas on this logic, which we want to spend on the target (the scenario/opcode we are benching))

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +113,128 @@ def test_xcall(
             "during the setup phase of this test."
         )
```
LouisTsai-Csie: I think this could be removed, because the logic below no longer packs all deployments into one transaction within a single block with a high gas limit.

There will be some problems to `env.gas_limit` as it is today:
- We use `gas_benchmark_value` to represent the gas limit we actually want to benchmark. While `env.gas_limit` is set to a very high value only to enable generating a single genesis file.
- Such a large gas limit creates issues when running in [execute-remote](https://eest.ethereum.org/main/running_tests/execute/remote/?h=execute+remote) mode on live networks. Since `env.gas_limit` is currently set to [1000G gas](https://github.com/ethereum/execution-specs/blob/9523e75061bc9f989f0867e34f28364c719a3b07/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py#L154), benchmarking a network with a lower gas limit would cause an “intrinsic gas too high” error.

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: That's nice, if you want to share the helper function you could place it under `tests/benchmark/compute/helpers.py`

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: In my personal view, ~2 minutes isn’t too bad for 60M. When I tried this refactoring earlier, the setup + attack phases took more than 20 minutes (I don’t recall the exact number), this makes it very hard to check the result. The recent fill optimizations have already made a huge difference.

If this does become a bottleneck again, we can profile the fill process and identify more optimization opportunities!

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +113,128 @@ def test_xcall(
...
+        )
+        attack_txs = math.ceil(num_contracts / num_contracts_per_tx)
+
+        contracts_deployment_txs = []
+        for _ in range(attack_txs):
+            contracts_deployment_txs.append(
+                Transaction(
+                    to=factory_caller_address,
+                    gas_limit=tx_gas_limit_cap,
+                    gas_price=10**6,
```
LouisTsai-Csie:
```suggestion
```

This should be removed. When using `execute remote` targeting a live network, it will automatically fetch the current network gas price.

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: Thanks! this is clear

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: I didn’t notice this when splitting the tests based on the folder structure. It would be good to remove it once this refactor is complete.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +113,128 @@ def test_xcall(
...
+        )
+    )
+
+    if len(attack_code) > max_contract_size:
+        # TODO: A workaround could be to split the opcode code into multiple
+        # contracts and call them in sequence.
+        raise ValueError(
+            f"Code size {len(attack_code)} exceeds maximum "
+            f"code size {max_contract_size}"
+        )
```
LouisTsai-Csie:
```suggestion
        # TODO: A workaround could be to split the opcode code into multiple
        # contracts and call them in sequence.
```

This check could be removed (we could leave the `TODO` as it is now). Since in both fill & execute mode, there is contract size limit check:

**Fill**
https://github.com/ethereum/execution-specs/blob/9523e75061bc9f989f0867e34f28364c719a3b07/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py#L188-L191

**Execute**
https://github.com/ethereum/execution-specs/blob/9523e75061bc9f989f0867e34f28364c719a3b07/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py#L325-L327

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +113,128 @@ def test_xcall(
...
+            - gas_costs.G_VERY_LOW * 4
+        ) // loop_cost
+        contract_start_index = 0
+        opcode_txs = []
+        for _ in range(full_txs):
+            opcode_txs.append(
+                Transaction(
+                    to=attack_address,
+                    gas_limit=tx_gas_limit_cap,
+                    gas_price=10**9,
```
LouisTsai-Csie:
```suggestion
```

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -108,6 +113,128 @@ def test_xcall(
...
+                    sender=pre.fund_eoa(),
+                )
+            )
+            contract_start_index += num_targeted_contracts_per_full_tx
+        if remainder > 0:
+            opcode_txs.append(
+                Transaction(
+                    to=attack_address,
+                    gas_limit=remainder,
+                    gas_price=10**9,
```
LouisTsai-Csie:
```suggestion
```

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: This should be updated as part of the EIP-7825 compatibility refactor:

```python
The attack gas limit represents the transaction gas limit cap or the block gas limit.
If eip-7825 is applied, the test will create multiple transactions for contract deployment.
It should account for the 200 gas per byte cost and the quadratic memory-expansion costs, which must be paid each time memory is initialized.
```
Feel free to adjust the comment for clarity.

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie:
```suggestion
    tx_gas_limit_cap = fork.transaction_gas_limit_cap() or gas_benchmark_value
```

I believe this should be sufficient to support both pre-Osaka and post-Osaka scenarios. If `fork.transaction_gas_limit_cap()` returns `None` (e.g., in Prague), it will fall back to `gas_benchmark_value`, and the transaction preparation logic will result in a single transaction:

```python
full_txs = attack_gas_limit // tx_gas_limit_cap   # becomes 1
remainder = attack_gas_limit % tx_gas_limit_cap   # becomes 0
```

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: Yes, this is a very common pattern in stateful scenarios as well. I need to think about it a bit more, maybe a new code generator type could help.

LouisTsai-Csie (changes requested):
Huge thanks to the refactor! Leave some comments now but i will add more as i see some failure in `execute remote` scenario!

tests/benchmark/compute/instruction/test_system.py
jsign: The setup phase also requires some touching since should use `env.gas_limit` instead of `tx_gas_limit_cap`; that's mainly what I didn't like that much if not going to be used, but I can do it if feels worthwhile. I'll try and report later.

tests/benchmark/compute/instruction/test_system.py
jsign: Sounds good. I think I'll remove the comment since leaving it alone feels pretty weird without context.

tests/benchmark/compute/instruction/test_system.py
jsign: Cool, I'll move it there if re-used when fixing `test_extcode_ops`.

tests/benchmark/compute/instruction/test_system.py
jsign: @LouisTsai-Csie, generalized it in the last commit. Tested both again (Prague and Osaka) and looks fine, so now it works on any fork.

tests/benchmark/compute/instruction/test_system.py
jsign: Ah, just as a note, I usually fill go-ethereum. Prob with evmone will be as fast, but with spec prob will be much slower.

jsign:
@LouisTsai-Csie, addressed all the review comments. I think you wanted to do some remote tests, so LMK if there is any other blocker.

LouisTsai-Csie (changes requested):
Please help update the base branch to `forks/amsterdam` and update the `blockchain_test` fixture to `benchmark_test`. Also, please check this comment, thanks!!

https://github.com/ethereum/execution-specs/pull/1850#discussion_r2596820194

jsign:
> Please help update the base branch to `forks/amsterdam` and update the `blockchain_test` fixture to `benchmark_test`. Also, please check this comment, thanks!!
> 
> [#1850 (comment)](https://github.com/ethereum/execution-specs/pull/1850#discussion_r2596820194)

@LouisTsai-Csie, rebased for `forks/amsterdam` and also changed some things to use `tx_gas_limit` instead of other calculations as to avoid `env.gas_limit`.

Reg the `blockchain_test`, I'm not sure we can use `bnechmark_test` since in this test we rely on the `exclude_full_post_state_in_output=True` flag. This flag is used so the post-state isn't include in the fixture (since for this benchmark it would mean putting hundred of megabytes in the fixture). I think this flag isn't available in `benchmark_test`?

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -90,23 +80,118 @@ def test_xcall(
...
+    else:
+        # For the rest of the opcodes, we can use the same generic attack call
+        # since all only minimally need the `address` of the target.
+        attack_call = Op.POP(opcode(address=Op.SHA3(32 - 20 - 1, 85)))
+    attack_code = (
+        # Setup memory for later CREATE2 address generation loop.
+        # 0xFF+[Address(20bytes)]+[seed(32bytes)]+[initcode keccak(32bytes)]
+        Op.MSTORE(0, factory_address)
+        + Op.MSTORE8(32 - 20 - 1, 0xFF)
+        + Op.MSTORE(32, Op.CALLDATALOAD(0))
```
jochem-brouwer:
```suggestion
        + Op.MSTORE(32, Op.CALLDATALOAD(0)) # Calldata is the starting value of the CREATE2 salt
```

I think it is helpful to add we encode the starting salt of the tx in the calldata 😄 👍

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -90,23 +80,118 @@ def test_xcall(
...
+    attack_address = pre.deploy_contract(code=attack_code)
+
+    with TestPhaseManager.execution():
+        full_txs = attack_gas_limit // tx_gas_limit
+        remainder = attack_gas_limit % tx_gas_limit
+
+        num_targeted_contracts_per_full_tx = (
+            # Base available gas:
+            # TX_GAS_LIMIT - intrinsic - (out of loop MSTOREs)
+            tx_gas_limit - intrinsic_gas_cost_calc() - gas_costs.G_VERY_LOW * 4
```
jochem-brouwer, line 158: This does not account for the calldata gas cost which got introduced here. So this will target more contracts than will be reached in reality, which is not a problem, as this thus guarantees that all txs target unique addresses.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -90,23 +80,118 @@ def test_xcall(
...
+        _deploy_max_contract_factory(pre, fork)
+    )
+
+    # Deploy num_contracts via multiple txs (each capped by tx gas limit).
+    with TestPhaseManager.setup():
+        # Rough estimate (rounded down) of contracts per tx based on dominant
+        # cost factor only, and up to 90% of the block gas limit.
+        # The goal is to involve the minimum amount of gas pricing to avoid
+        # complexity and potential brittleness.
+        num_contracts_per_tx = int(tx_gas_limit * 0.9) // (
```
jochem-brouwer, line 98: This var is used in the execution phase also, can you scope it such it is available both in setup and in execution phase?

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -90,23 +80,118 @@ def test_xcall(
...
+            opcode_txs.append(
+                Transaction(
+                    to=attack_address,
+                    gas_limit=tx_gas_limit,
+                    data=Hash(contract_start_index),
+                    sender=pre.fund_eoa(),
+                )
+            )
+            contract_start_index += num_targeted_contracts_per_full_tx
+        if remainder > 0:
```
jochem-brouwer: This should also account for the fact if intrinsic costs cannot be paid (for instance if remainder is 1). It's a nit though, but this will happen at specific gas limits (for instance 2^16 + 1 will yield 1 remainder  gas).

jochem-brouwer (commented):
Some minor comments, looks very good 😄 👍

tests/benchmark/compute/instruction/test_system.py
jsign: `num_contracts_per_tx` is only used inside `with TestPhaseManager.setup():` scope. Maybe I'm missing something?

tests/benchmark/compute/instruction/test_system.py
jsign: Good point -- I changed this line. LMK if this addresses your point to double check?

jochem-brouwer: Ah cool yes this should work :smile: :+1: It also nicely takes care of the 32 bytes calldata. So this should not run into any exceptions when using this test.

jsign:
Rebased again on top of the forced pushed base.

LouisTsai-Csie:
Working good for 100M but failing for our 5M CI check. Not sure what's the best way going forward now.

```bash
_ test_xcall[fork_Prague-benchmark-gas-value_1M-blockchain_test_engine_x-opcode_CALLCODE] _
[gw4] linux -- Python 3.11.14 /opt/actions-runner/_work/execution-specs/execution-specs/.tox/benchmark/bin/python3
tests/benchmark/compute/instruction/test_system.py:102: in test_xcall
    setup_txs = math.ceil(num_contracts / num_contracts_per_tx)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   ZeroDivisionError: division by zero
```

jsign:
@LouisTsai-Csie, yeah, this is tricky.

For Prague they fail, since the `tx_gas_limit` received as parameter is resolved as 5M (or 1M). A single 24KiB deployment costs more than 5M, which means in both cases the amount of contracts you can deploy is zero.

But for Osaka, both 1M and 5M fill fine because `tx_gas_limit` is 16M. So the math works out since you have enough room. Although when you do the attack transaction you use 5M (or 1M) and tries to call on "best effort".

I think the way `tx_gas_limit` is defined in the framework, looks like already assumes in a way that the underlying block gas limit will be at least 16M (for Osaka)? (maybe sounds liek a safe assumption since the 17M number in the EIP was thinking about mainnet when was proposed, not a 1M or 5M chain). And I think you mentioned we can't use `env.gas_limit` since it might not be a fair representation of remote chains.

I think my conclusion here is (and pushed a commit with this), is if `num_contracts_per_tx` is zero, then we mark the fixture as skipped. Doing this attack for 1M or 5M targets doesn't sounds like important cases. For example for re-execution the main goal of this attack is to maximize IO on realistic sceanrios, and for zkVMs is to try to blow up the guest program input size (sum of bytecodes) to see if zkVM crash (which they do, but in higher gas limits) and also hashing these bytecodes.

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: I think it is only used in setup phase?

LouisTsai-Csie (approved):
Thanks for your explanation and the final fix! LGTM

PR #1852, opened 2025-12-05
bug(test-cli-fill): Use default evm-dump-dir for traces; gitignore it

fselmo:
## 🗒️ Description

- `--traces` with `--evm-dump-dir` is currently broken due to improper `LazyAlloc` handling. This PR fixes it by validating with `get()` and properly serializing via `model_dump_json`.
- Use a default `--evm-dump-dir` of `traces/` which is also in `.gitignore` so we don't have to keep removing it from the working dir. Previously, `--traces` did absolutely nothing without a specified `--evm-dump-dir` but it also didn't tell you anything in the cli if you did this, which is quite confusing DevEx.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2025-12-05 at 11 53 25


packages/testing/src/execution_testing/client_clis/file_utils.py
```diff
@@ -22,7 +24,16 @@ def dump_files_to_directory(output_path: str, files: Dict[str, Any]) -> None:
...
-            if isinstance(file_contents, BaseModel) or isinstance(
+            if isinstance(file_contents, LazyAlloc):
+                # Validate and serialize the lazy alloc
+                f.write(
+                    file_contents.get().model_dump_json(
+                        indent=4,
+                        exclude_none=True,
+                        by_alias=True,
+                    )
+                )
```
marioevz: Perhaps we can make the distinction here between `LazyAllocJson` and `LazyAllocStr` and write `file_contents.raw` directly (`json.dumps(file_contents.raw)` in case of `LazyAllocJson` perhaps) to avoid parsing the json before dumping it.

marioevz (changes requested):
Just one change I think should be a good optimization to make.

packages/testing/src/execution_testing/client_clis/file_utils.py
fselmo: I added this in here but maybe I'm missing the point. Isn't it JSON from t8n all the way until here? So don't we want to validate it before we dump it? It is an extra step in that case but we'd be guaranteeing that it's validated. I thought that was the point of the `LazyAlloc`, to delay validation until it is necessary - which this would end up being the very last step so it would get validated here. Let me know if I missed something.

.gitignore
```diff
@@ -76,6 +76,9 @@ tests/t8n_testdata
 
 fixtures/
 
+# Trace output (generated by --traces flag)
+traces/
```
SamWilsn: This effectively means we can have no directory anywhere called `traces`. Is that acceptable?

packages/testing/src/execution_testing/client_clis/file_utils.py
```diff
@@ -22,7 +24,19 @@ def dump_files_to_directory(output_path: str, files: Dict[str, Any]) -> None:
...
+                    dump(file_contents.raw, f, ensure_ascii=True, indent=4)
+                else:
+                    # Validate and serialize LazyAlloc
+                    f.write(
+                        file_contents.get().model_dump_json(
+                            indent=4,
+                            exclude_none=True,
+                            by_alias=True,
+                        )
+                    )
```
marioevz:
```suggestion
                    f.write(file_contents.raw)
```
I think in this case we can write raw directly since it's text.

.gitignore
fselmo: Open to any other names here tbh. It would be nice to have something that is git-ignored so we don't constantly have to clean it when we are debugging and adding changes to be committed. Any other suggestion is welcomed.

packages/testing/src/execution_testing/client_clis/file_utils.py
fselmo: @marioevz don't we want to validate here though? See my comment here https://github.com/ethereum/execution-specs/pull/1852#discussion_r2599479518

.gitignore
SamWilsn: I'd say it'd be better to ignore the specific file name patterns if we can (eg. `traces/*.txt` or whatever), or use an absolute `/traces/`. This is really a _tiny_ issue that'll get fixed if someone ever does try to include a `traces` directory in the future :rofl:

SamWilsn (approved):

.gitignore
fselmo: Ah, I see your point. I changed it to `/traces/` 👌🏼. More specific.

marioevz (approved):
LGTM, thanks!

PR #1862, opened 2025-12-08
feat(benchmarks): fix benchmark for BALANCE scenario for Osaka

jsign:
## 🗒️ Description
This PR fixes the `est_ext_account_query_cold` benchmark to work in Osaka.

To validate this still does what we want, tested with 30M attack:

- Before this PR with Prague:
    - `test_account_query.py::test_ext_account_query_cold[fork_Prague-blockchain_test-absent_accounts_False-opcode_BALANCE]` = 214M cycles
    - `test_account_query.py::test_ext_account_query_cold[fork_Prague-blockchain_test-absent_accounts_True-opcode_BALANCE]` = 98M cycles
- With this PR:
    - Filled with Osaka:
        - `test_account_query.py::test_ext_account_query_cold[fork_Prague-blockchain_test-absent_accounts_False-opcode_BALANCE]` = 213M cycles
        - `test_account_query.py::test_ext_account_query_cold[fork_Prague-blockchain_test-absent_accounts_True-opcode_BALANCE]` = 96M cycles
        - Conclusion: the difference sounds reasonable reg the gas lost in partitioning in txs.
    - Filled with Prague:
        - `test_account_query.py::test_ext_account_query_cold[fork_Prague-blockchain_test-absent_accounts_False-opcode_BALANCE]` = 214M cycles
        - `test_account_query.py::test_ext_account_query_cold[fork_Prague-blockchain_test-absent_accounts_True-opcode_BALANCE]` = 98M cycles
        - Conclusion: same as pre-PR.

## 🔗 Related Issues or PRs

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/benchmark/compute/instruction/test_account_query.py
+++ b/tests/benchmark/compute/instruction/test_account_query.py
@@ -448,15 +448,13 @@ def test_ext_account_query_cold(
 
     gas_costs = fork.gas_costs()
     intrinsic_gas_cost_calc = fork.transaction_intrinsic_cost_calculator()
+    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
     # For calculation robustness, the calculation below ignores "glue" opcodes
     # like  PUSH and POP. It should be considered a worst-case number of
     # accounts, and a few of them might not be targeted before the attacking
     # transaction runs out of gas.
     num_target_accounts = (
         attack_gas_limit - intrinsic_gas_cost_calc()
-    ) // gas_costs.G_COLD_ACCOUNT_ACCESS
-    max_target_per_tx = (
-        fork.transaction_gas_limit_cap() - intrinsic_gas_cost_calc()
     ) // gas_costs.G_COLD_ACCOUNT_ACCESS
 
     blocks = []
@@ -474,11 +472,17 @@ def test_ext_account_query_cold(
             + gas_costs.G_CALL_VALUE
             + gas_costs.G_NEW_ACCOUNT
         )
-        # To avoid brittle/tight gas calculations of glue opcodes, we take 90% of the maximum tx capacity.
-        # Even if this calculation fails in the future, it would be catched by the post-state check.
-        max_creations_per_tx = int(
-            (fork.transaction_gas_limit_cap() * 0.9) // account_creation_gas
-        )
+        # To avoid brittle/tight gas calculations of glue opcodes, we take
+        # 90% of the maximum tx capacity. Even if this calculation fails
+        # in the future, it will be caught by the post-state check.
+        # Also, this is only for the setup phase, so being optimal is
+        # not critical.
+        if tx_gas_limit_cap:
+            max_creations_per_tx = int(
+                (tx_gas_limit_cap * 0.9) // account_creation_gas
+            )
+        else:
+            max_creations_per_tx = num_target_accounts
         factory_code = (
             Op.CALLDATALOAD(0)  # addr_start
             + Op.PUSH4(max_creations_per_tx)  # counter
@@ -513,7 +517,7 @@ def test_ext_account_query_cold(
                     Transaction(
                         to=factory_address,
                         data=Hash(addr_start),
-                        gas_limit=fork.transaction_gas_limit_cap(),
+                        gas_limit=tx_gas_limit_cap or env.gas_limit,
                         sender=pre.fund_eoa(),
                     )
                 )
@@ -542,6 +546,13 @@ def test_ext_account_query_cold(
 
     execution_txs = []
     with TestPhaseManager.execution():
+        if tx_gas_limit_cap:
+            max_target_per_tx = (
+                tx_gas_limit_cap - intrinsic_gas_cost_calc()
+            ) // gas_costs.G_COLD_ACCOUNT_ACCESS
+        else:
+            max_target_per_tx = num_target_accounts
+
         num_execution_txs = math.ceil(num_target_accounts / max_target_per_tx)
         for i in range(num_execution_txs):
             address_start = i * int(max_target_per_tx)
@@ -551,7 +562,7 @@ def test_ext_account_query_cold(
                 Transaction(
                     to=op_address,
                     data=Hash(address_start) + Hash(num_to_query),
-                    gas_limit=fork.transaction_gas_limit_cap(),
+                    gas_limit=tx_gas_limit_cap or env.gas_limit,
                     sender=pre.fund_eoa(),
                 )
             )
--- a/tests/benchmark/compute/instruction/test_account_query.py
+++ b/tests/benchmark/compute/instruction/test_account_query.py
@@ -440,6 +440,7 @@ def test_ext_account_query_cold(
     absent_accounts: bool,
     env: Environment,
     gas_benchmark_value: int,
+    tx_gas_limit: int,
 ) -> None:
     """
     Benchmark stateful opcodes accessing cold accounts.
@@ -448,7 +449,6 @@ def test_ext_account_query_cold(
 
     gas_costs = fork.gas_costs()
     intrinsic_gas_cost_calc = fork.transaction_intrinsic_cost_calculator()
-    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
     # For calculation robustness, the calculation below ignores "glue" opcodes
     # like  PUSH and POP. It should be considered a worst-case number of
     # accounts, and a few of them might not be targeted before the attacking
@@ -477,12 +477,9 @@ def test_ext_account_query_cold(
         # in the future, it will be caught by the post-state check.
         # Also, this is only for the setup phase, so being optimal is
         # not critical.
-        if tx_gas_limit_cap:
-            max_creations_per_tx = int(
-                (tx_gas_limit_cap * 0.9) // account_creation_gas
-            )
-        else:
-            max_creations_per_tx = num_target_accounts
+        max_creations_per_tx = int(
+            (tx_gas_limit * 0.9) // account_creation_gas
+        )
         factory_code = (
             Op.CALLDATALOAD(0)  # addr_start
             + Op.PUSH4(max_creations_per_tx)  # counter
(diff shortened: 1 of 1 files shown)


tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -477,52 +478,106 @@ def test_ext_account_query_cold(
...
+        # in the future, it will be caught by the post-state check.
+        # Also, this is only for the setup phase, so being optimal is
+        # not critical.
+        if tx_gas_limit_cap:
+            max_creations_per_tx = int(
+                (tx_gas_limit_cap * 0.9) // account_creation_gas
+            )
+        else:
+            max_creations_per_tx = num_target_accounts
+        factory_code = (
```
jsign, line 482: TL;DR: the factory code that creates the accounts (`if not absent_accounts`), just creates the maximum allowed per tx in this setup phase. If in `>= Osaka`, it does a safe estimation so the setup phase doens't fail. If `< Osaka`, just does one transaction as before.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -477,52 +478,106 @@ def test_ext_account_query_cold(
...
+            Op.CALLDATALOAD(0)  # addr_start
```
jsign, line 483: For `>= Osaka`, we need this extra `CALLDATALOAD` so the setup phase with multiple txs keeps creating the target accounts in a continuous way.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -477,52 +478,106 @@ def test_ext_account_query_cold(
...
-        body=Op.POP(opcode(Op.ADD(addr_offset, Op.DUP1))),
-        condition=Op.PUSH1(1)
-        + Op.SWAP1
-        + Op.SUB
-        + Op.DUP1
-        + Op.ISZERO
-        + Op.ISZERO,
+    op_code = (
+        Op.CALLDATALOAD(0)  # address_start
+        + Op.CALLDATALOAD(32)  # num_to_query
```
jsign, line 529: For `>= Osaka`, the attack code now has two new parameters via CALLDATALOAD, which are required for all the attacking txs to target their respective (non-overlapping) accounts.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -466,52 +467,100 @@ def test_ext_account_query_cold(
...
+            (tx_gas_limit * 0.9) // account_creation_gas
+        )
+        factory_code = (
+            Op.CALLDATALOAD(0)  # addr_start
+            + Op.PUSH4(max_creations_per_tx)  # counter
+            # Stack: [counter, addr_start]
+            + While(
+                body=Op.POP(
+                    Op.CALL(
+                        address=Op.ADD(addr_offset, Op.ADD(Op.DUP7, Op.DUP7)),
```
jochem-brouwer, line 489: What do these two additions do? This is a constant I think? Do these stack values (`DUP7`) ever change?

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -466,52 +467,100 @@ def test_ext_account_query_cold(
...
+                        value=10,
+                    )
+                ),
+                condition=Op.PUSH1(1)
+                + Op.SWAP1
+                + Op.SUB
+                + Op.DUP1
+                + Op.ISZERO
+                + Op.ISZERO,
```
jochem-brouwer, line 498: The `ISZERO` is intended as loop condition, but I think if you put it in the body and not explicitly as condition it will not be used as it (so this is an infinite loop?).

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -466,52 +467,100 @@ def test_ext_account_query_cold(
...
+        num_execution_txs = math.ceil(num_target_accounts / max_target_per_tx)
+        for i in range(num_execution_txs):
+            address_start = i * int(max_target_per_tx)
+            remaining = num_target_accounts - address_start
+            num_to_query = min(int(max_target_per_tx), remaining)
+            execution_txs.append(
+                Transaction(
+                    to=op_address,
+                    data=Hash(address_start) + Hash(num_to_query),
+                    gas_limit=tx_gas_limit,
```
jochem-brouwer: This also has the problem it will only submit txs with the tx gas limit as limit, thus possibly exceeding the block gas limit we target.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -466,52 +467,100 @@ def test_ext_account_query_cold(
...
+        + Op.CALLDATALOAD(32)  # num_to_query
+        # Stack: [num_to_query, address_start]
+        + While(
+            body=Op.POP(opcode(Op.ADD(addr_offset, Op.ADD(Op.DUP2, Op.DUP2)))),
+            condition=Op.PUSH1(1)
+            + Op.SWAP1
+            + Op.SUB
+            + Op.DUP1
+            + Op.ISZERO
+            + Op.ISZERO,
```
jochem-brouwer, line 538: This while loop is likely also an infinite loop.

tests/benchmark/compute/instruction/test_account_query.py
jochem-brouwer: Ah ok it should resolve to (?) `addr_offset + counter + addr_start`, and the counter is a stack value which decreases by one each loop.

I'm not sure if DUP7 is correct here though, I don't think we have 7 stack items at this point.

jochem-brouwer (commented):
Some comments 😄 👍

tests/benchmark/compute/instruction/test_account_query.py
jsign: > Ah ok it should resolve to (?) addr_offset + counter + addr_start, and the counter is a stack value which decreases by one each loop.

Yep

> I'm not sure if DUP7 is correct here though, I don't think we have 7 stack items at this point.

The `Op.CALL` in L489 already pushed for `retSize`, `retOffset`, `argSize`, `argsOffset` and `value` (call parameters before `address`). Then on top of that we have the `CALDATALOAD(0)` (L484) and `PUSH4` (L485).

Thus the first DUP7 is targeting the `CALDATALOAD` and then DUP7 again would be `PUSH4` (since previous DUP shifted by one).

Yea, it isn't obvious and when I wrote this had to debug in Geth to be convinced. But hopefully this clarifies, since `Op.CALL` has many arguments that although aren't required, still push default values.

tests/benchmark/compute/instruction/test_account_query.py
jsign: (Parallel note: this code wasn't changed in this PR -- but still okay to review everything ofc)

Not sure I understood your comment 😅. What this is doing is substracting 1 from the `PUSH4`, using DUP1 to repeat the value for a first ISZERO, and then doing anoher ISZERO to make the condition be "is not zero --> jump to the loop again" (i.e. run the loop until the count is zero).

LMK if this makes sense, or why you think there can be an infinite loop here.

tests/benchmark/compute/instruction/test_account_query.py
jsign: Fixed!

tests/benchmark/compute/instruction/test_account_query.py
jsign: Made a comment in previous review comment to understand better.

tests/benchmark/compute/instruction/test_account_query.py
jochem-brouwer: For some reason I missed the `condition=` and I thought this While loop only had a body, and no condition (so then it is an infinite loop). However this is not the case, it all looks fine :smile: :+1:

tests/benchmark/compute/instruction/test_account_query.py
jochem-brouwer: Ah right, yes this looks fine. I did not realize this was in the `CALL` and instead thought it was "before" the call (prior to setting up the stack for the operation). 

Yes with these previous args this looks fine :+1: 

We might also think a little bit how we can improve the DevEx here with this stack manipulation or other tools we could introduce to help writing these tests :sweat_smile:

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -548,18 +548,21 @@ def test_ext_account_query_cold(
         ) // gas_costs.G_COLD_ACCOUNT_ACCESS
 
         num_execution_txs = math.ceil(num_target_accounts / max_target_per_tx)
+        gas_used = 0
         for i in range(num_execution_txs):
             address_start = i * int(max_target_per_tx)
             remaining = num_target_accounts - address_start
             num_to_query = min(int(max_target_per_tx), remaining)
+            gas_limit = min(tx_gas_limit, attack_gas_limit - gas_used)
```
jochem-brouwer, line 555: This should also introduce the check to see if the intrinsic gas of the tx is covered (only for final tx, and skip tx if it is not covered).

jsign: Added check.

tests/benchmark/compute/instruction/test_account_query.py
jsign: Yeah, it is quite rough. Although the main thing I think is the most hairy are many of hte manual gas cost calculations. 

I've been trying us to stay more away from that for checks (e.g. these 90% capacity block gas for setup txs to avoid glue opcodes precise calcs), since whenever there are repricings, it can backfire in all these benchmarks. e.g. some opcode now uses another constant G_BASE -> G_VERY_LOW or similar.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -440,6 +440,7 @@ def test_ext_account_query_cold(
     absent_accounts: bool,
     env: Environment,
```
LouisTsai-Csie:
```suggestion

```

LouisTsai-Csie (approved):
Nice!!! Appreciate the help!

tests/benchmark/compute/instruction/test_account_query.py
jsign: Interesting -- locally I run `uv run ruff format && uvx --with=tox-uv tox -e static` and wasn't raised as a lint error. :)
Fixed.

jsign:
@LouisTsai-Csie, fixed the last nit.

PR #1866, opened 2025-12-08
Merge `forks/osaka` into `mainnet`

SamWilsn:
#### Cute Animal Picture

image


SamWilsn:
Don't merge this one directly. I'll do a fast-forward-only push if we're happy with it.

Issue #1835, opened 2025-12-03
Augment `BenchmarkTestFiller` with opcode under test and its expected count


danceratopz:
Currently, the test framework doesn't know which opcode is under test or the expected number of opcode evaluations the test should perform. For example:
https://github.com/ethereum/execution-specs/blob/85c6fff98df4bc54b834deac39273ab7bfc51d27/tests/benchmark/compute/instruction/test_arithmetic.py#L151-L158

If we gave this information to the test spec, `BenchmarkTestFiller`, e.g., for the case of a "fixed opcode count" test (see #1834):
```python
    benchmark_test(
        opcode_under_test=opcode,  # in test_arithmetic.py, this is a test parameter
        # after #1834 `opcode_count` is a test parameter; could be None for worst case tests (skip check or 
        # test framework specifies it automatically based on gaslimit.
        expected_opcode_count=opcode_count,
        code_generator=JumpLoopGenerator(
            setup=setup,
            attack_block=attack_block,
            cleanup=cleanup,
            tx_kwargs={"data": tx_data},
        ),
    )
```

then we could leverage evmone's  [`--opcode.count`](https://github.com/ipsilon/evmone/issues/1256) bookkeeping functionality to verify the opcode evaluations during `fill`.

LouisTsai-Csie:
Hi Dan, i will closed this one as PR #1869 is done. Thanks for the instruction!

PR #1869, opened 2025-12-09
feat(test-benchmark): implement opcode count verification

LouisTsai-Csie:
## 🗒️ Description

There are several enhancements in this PR:
- Implement opcode-count verification for `--fixed-opcode-count` mode using the opcount field provided when generating tests with `evmone` with 5% deviation. 

- Re-label the worst-case scenarios for the repricing marker based on the Grafana [dashboard](https://grafana.observability.ethpandaops.io/d/feo4ronhsqv40d/opcodes-benchmarking?orgId=1&from=2025-12-02T11:44:15.682Z&to=2025-12-02T12:01:58.956Z&timezone=browser&var-posgreSQL=benuragv7iuwwb&var-TableName=gas_limit_benchmarks&var-ClientName=nethermind&var-ClientName=geth&var-TestTitle=$__all&viewPanel=panel-24) data and this [script](https://gist.github.com/LouisTsai-Csie/17a57d0500466261935c6919cdb2e80e) that sort the cases based on the MGas result.
  - The repricing marker under` tests/benchmark/compute/instruction/` are udpated.

Based on our discussion about the verification mechanism during sync up call, specifically whether to remove the `target_opcode` field. I would suggest not removing it for now, since this [feature](https://github.com/ethereum/execution-specs/pull/1898) depends on it, which is very helpful for me to show coverage to others, without manually update the Notion page.

## 🔗 Related Issues or PRs

Issue #1835 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture
Gemini_Generated_Image_d360w8d360w8d360

### DIFF
--- a/packages/testing/src/execution_testing/cli/benchmark_parser.py
+++ b/packages/testing/src/execution_testing/cli/benchmark_parser.py
@@ -268,17 +268,20 @@ def categorize_patterns(
 def generate_config_json(
     config: dict[str, list[int]],
     pattern_sources: dict[str, Path],
+    default_counts: list[int],
 ) -> OpcodeCountsConfig:
     """Generate the JSON config file content."""
     categories = categorize_patterns(config, pattern_sources)
 
     scenario_configs: dict[str, list[int]] = {}
-    output = OpcodeCountsConfig(scenario_configs=scenario_configs)
     for _, patterns in categories.items():
         for pattern in patterns:
-            output.scenario_configs[pattern] = config[pattern]
+            scenario_configs[pattern] = config[pattern]
 
-    return output
+    return OpcodeCountsConfig(
+        scenario_configs=scenario_configs,
+        default_counts=default_counts,
+    )
 
 
 def main() -> int:
@@ -352,8 +355,9 @@ def main() -> int:
         if pattern in merged:
             del merged[pattern]
 
-    content = generate_config_json(merged, pattern_sources)
-    content.default_counts = existing_file.default_counts
+    content = generate_config_json(
+        merged, pattern_sources, existing_file.default_counts
+    )
     config_file.write_text(
         content.model_dump_json(exclude_defaults=True, indent=2)
     )
--- a/packages/testing/src/execution_testing/cli/benchmark_parser.py
+++ b/packages/testing/src/execution_testing/cli/benchmark_parser.py
@@ -11,9 +11,12 @@ Usage:
 
 import argparse
 import ast
-import json
 import sys
 from pathlib import Path
+
+from execution_testing.cli.pytest_commands.plugins.shared.benchmarking import (
+    OpcodeCountsConfig,
+)
 
 
 def get_repo_root() -> Path:
@@ -227,16 +230,11 @@ def scan_benchmark_tests(
     return config, pattern_sources
 
 
-def load_existing_config(config_file: Path) -> dict[str, list[int]]:
+def load_existing_config(config_file: Path) -> OpcodeCountsConfig:
     """Load existing config from .fixed_opcode_counts.json."""
     if not config_file.exists():
-        return {}
-
-    try:
-        data = json.loads(config_file.read_text())
-        return data.get("scenario_configs", {})
-    except (json.JSONDecodeError, KeyError):
-        return {}
+        return OpcodeCountsConfig()
+    return OpcodeCountsConfig.model_validate_json(config_file.read_bytes())
 
 
 def categorize_patterns(
@@ -270,18 +268,17 @@ def categorize_patterns(
 def generate_config_json(
     config: dict[str, list[int]],
     pattern_sources: dict[str, Path],
-) -> str:
+) -> OpcodeCountsConfig:
     """Generate the JSON config file content."""
     categories = categorize_patterns(config, pattern_sources)
 
     scenario_configs: dict[str, list[int]] = {}
+    output = OpcodeCountsConfig(scenario_configs=scenario_configs)
     for _, patterns in categories.items():
         for pattern in patterns:
-            scenario_configs[pattern] = config[pattern]
+            output.scenario_configs[pattern] = config[pattern]
 
-    output = {"scenario_configs": scenario_configs}
-
-    return json.dumps(output, indent=2) + "\n"
+    return output
 
 
 def main() -> int:
@@ -307,7 +304,8 @@ def main() -> int:
     detected, pattern_sources = scan_benchmark_tests(benchmark_dir)
     print(f"Detected {len(detected)} opcode patterns")
 
-    existing = load_existing_config(config_file)
+    existing_file = load_existing_config(config_file)
+    existing = existing_file.scenario_configs
     print(f"Loaded {len(existing)} existing entries")
 
     detected_keys = set(detected.keys())
@@ -355,7 +353,10 @@ def main() -> int:
             del merged[pattern]
 
     content = generate_config_json(merged, pattern_sources)
-    config_file.write_text(content)
+    content.default_counts = existing_file.default_counts
+    config_file.write_text(
+        content.model_dump_json(exclude_defaults=True, indent=2)
+    )
     print(f"\nUpdated {config_file}")
     return 0
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -432,6 +432,8 @@ class Alloc(BaseAlloc):
         fund_tx: PendingTransaction | None = None
         if delegation is not None or storage is not None:
             if storage is not None:
+                if not isinstance(storage, Storage):
+                    storage = Storage.model_validate(storage)
                 logger.debug(
                     f"Deploying storage contract for EOA {eoa} with {len(storage)} storage slots"
                 )
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -432,14 +432,18 @@ class Alloc(BaseAlloc):
         fund_tx: PendingTransaction | None = None
         if delegation is not None or storage is not None:
             if storage is not None:
+                # Handle both Storage objects and plain dicts
+                storage_dict = (
+                    storage.root if isinstance(storage, Storage) else storage
+                )
                 logger.debug(
-                    f"Deploying storage contract for EOA {eoa} with {len(storage.root)} storage slots"
+                    f"Deploying storage contract for EOA {eoa} with {len(storage_dict)} storage slots"
                 )
                 sstore_address = self.deploy_contract(
                     code=(
                         sum(
                             Op.SSTORE(key, value)
-                            for key, value in storage.root.items()
+                            for key, value in storage_dict.items()
                         )
                         + Op.STOP
                     )
@@ -451,6 +455,7 @@ class Alloc(BaseAlloc):
                 set_storage_tx = PendingTransaction(
                     sender=self._sender,
                     to=eoa,
+                    value=0,
                     authorization_list=[
                         AuthorizationTuple(
                             chain_id=self._chain_id,
@@ -498,7 +503,7 @@ class Alloc(BaseAlloc):
                 fund_tx = PendingTransaction(
                     sender=self._sender,
                     to=eoa,
-                    value=amount,
+                    value=amount if amount is not None else 0,
                     authorization_list=[
                         AuthorizationTuple(
                             chain_id=self._chain_id,
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -408,7 +408,7 @@ class Alloc(BaseAlloc):
         self,
         amount: NumberConvertible | None = None,
         label: str | None = None,
-        storage: Storage | None = None,
+        storage: Storage | StorageRootType | None = None,
         delegation: Address | Literal["Self"] | None = None,
         nonce: NumberConvertible | None = None,
     ) -> EOA:
@@ -432,18 +432,14 @@ class Alloc(BaseAlloc):
         fund_tx: PendingTransaction | None = None
         if delegation is not None or storage is not None:
             if storage is not None:
-                # Handle both Storage objects and plain dicts
-                storage_dict = (
-                    storage.root if isinstance(storage, Storage) else storage
-                )
                 logger.debug(
-                    f"Deploying storage contract for EOA {eoa} with {len(storage_dict)} storage slots"
+                    f"Deploying storage contract for EOA {eoa} with {len(storage)} storage slots"
                 )
                 sstore_address = self.deploy_contract(
                     code=(
                         sum(
                             Op.SSTORE(key, value)
-                            for key, value in storage_dict.items()
+                            for key, value in storage.items()
                         )
                         + Op.STOP
                     )
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -124,7 +124,7 @@ def pytest_collection_modifyitems(
     #
     # Here we filter out tests that do not use the benchmark_test fixture.
     # Note: At this stage we cannot filter based on whether a code generator is used.
-    if fixed_opcode_count:
+    if fixed_opcode_count or fixed_opcode_count == "":
         filtered = []
         for item in items:
             if (
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -118,6 +118,22 @@ def pytest_collection_modifyitems(
     if not gas_benchmark_value and fixed_opcode_count is None:
         return
 
+    # In --fixed-opcode-count mode, we only support tests that meet all of the following:
+    #   - The test uses the benchmark_test fixture
+    #   - The benchmark test uses a code generator
+    #
+    # Here we filter out tests that do not use the benchmark_test fixture.
+    # Note: At this stage we cannot filter based on whether a code generator is used.
+    if fixed_opcode_count:
+        filtered = []
+        for item in items:
+            if (
+                hasattr(item, "fixturenames")
+                and "benchmark_test" in item.fixturenames
+            ):
+                filtered.append(item)
+        items[:] = filtered
+
     # Load config data if --fixed-opcode-count flag provided without value
     if fixed_opcode_count == "":
         config_data = load_opcode_counts_config(config)
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -1,14 +1,16 @@
 """The module contains the pytest hooks for the gas benchmark values."""
 
-import json
 import re
 import warnings
+from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any
+from typing import ClassVar, Dict, List, Self
 
 import pytest
+from pydantic import BaseModel, Field, RootModel
 
 from execution_testing.test_types import Environment, EnvironmentDefaults
+from execution_testing.tools import ParameterSet
 
 from .execute_fill import OpMode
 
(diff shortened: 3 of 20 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -44,25 +44,40 @@ def pytest_configure(config: pytest.Config) -> None:
...
         return
 
     filtered = []
     for item in items:
-        if not item.get_closest_marker("benchmark"):
-            continue
-
         repricing_marker = item.get_closest_marker("repricing")
         if not repricing_marker:
+            filtered.append(item)
```
spencer-tb: More of a question. Is the intent to keep tests without the marker?
```suggestion
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -44,25 +44,40 @@ def pytest_configure(config: pytest.Config) -> None:
...
+    # If the -m repricing marker was not specified,
+    # we should not filter any tests, so skip the remaining logic.
     markexpr = config.getoption("markexpr", "")
     if "repricing" not in markexpr or "not repricing" in markexpr:
         return
 
     filtered = []
     for item in items:
-        if not item.get_closest_marker("benchmark"):
-            continue
```
spencer-tb, line 144: Also a question, we only check for tests with rerpricing marker now? The aim here is to filter by fixture usage and not markers moving forward?

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -82,7 +89,7 @@ def deploy_fix_count_contracts(self, *, pre: Alloc, fork: Fork) -> Address:
             prefix
             + Op.JUMPDEST
             + Op.POP(
-                Op.STATICCALL(
+                Op.DELEGATECALL(
```
spencer-tb: To allow state modifications?

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -392,6 +421,31 @@ def generate_blockchain_test(self) -> BlockchainTest:
...
+        assert self.target_opcode is not None, "target_opcode is not set"
+        assert self.fixed_opcode_count is not None, (
+            "fixed_opcode_count is not set"
+        )
+
+        # fixed_opcode_count is in thousands units
+        expected = self.fixed_opcode_count * 1000
+
+        actual = opcode_count.root.get(self.target_opcode, 0)
+        tolerance = expected * 0.05  # 5% tolerance
```
spencer-tb, line 464: We should try to reduce this as much as we can imo. Lets start with 5% though.

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -148,11 +155,32 @@ def generate_repeated_code(
...
+        # 1c. If the calculated N is less than 1000, use 500 as the fallback count.
+
+        # --- 2. Determine Outer Iterations (M) ---
+        # The Loop Contract's call count (M) is set to ensure the final total execution is consistent.
+        #
+        # 2a. If N is 1000: Set M = fixed_opcode_count. (Total ops: fixed_opcode_count * 1000)
+        # 2b. If N is 500: Set M = fixed_opcode_count * 2. (Total ops: (fixed_opcode_count * 2) * 500 = fixed_opcode_count * 1000)
         if self.fixed_opcode_count is not None:
-            max_iterations = min(max_iterations, 1000)
+            inner_iterations = 1000 if max_iterations >= 1000 else 500
```
spencer-tb, line 197: We might need to consider the case where `max_iterations` < 500 in the future

spencer-tb (commented):
Awesome work on implementing this, nice solution to a tricky problem!

Added some questions below. I think we should rebase once #1790 is merged, as there will be some conflicts. :D

packages/testing/src/execution_testing/specs/benchmark.py
LouisTsai-Csie: Yes this is for state modification, but I am not sure whether this is a good approach going forward.

In the current fixed-opcode-count design, there are two contracts: a loop contract repeatedly calls into the target contract. Originally, i `STATICALL`s into the target contract, but this would be a problem if the target contract contains state-changing operations, thus making the tests not working as expected. 

I have three options:
1. Change it to `CALL` operation but it would require 1 more parameter (`value`).
2. Change it to `DELEGATECALL` operation but this will execute logic in target contract on the context of loop contract, not sure if this is a good idea.
3. If the target contract contains state-changing operations, use `CALL`, otherwise, use `STATICCALL`

packages/testing/src/execution_testing/specs/benchmark.py
LouisTsai-Csie: Sure! The only consideration is that some of the benchmark opcode are also used during the test construction. For example:

- JUMPDEST
- JUMP
- POP
- DUP
- PUSH

This should be much less than 5%, i think at most 1% deviation is enough, but i need to dive deeper into this analysis.

packages/testing/src/execution_testing/specs/benchmark.py
LouisTsai-Csie: This is now a problem for RETURN/REVERT scenario where each target contract could only execute the opcode once. But this is a special case.

So far setting `max_iteratinos` to 500 could run most of the benchmark test, i would suggest leave it as it is now. We could update the value if the test breaks in the future. (Hopefully now we have such verification mechanism.)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
LouisTsai-Csie: This was incorrect, I revisited the filter logic and removed it. However, I was curious why the behavior still looked correct even when this logic was here.

After reviewing into `pytest` implementation, I found there is an automatic selection mechanism: when you pass a marker (e.g. `-m repricing`), `pytest` will only select tests that have this marker. So even though we were adding these tests to the filtered list here, tests without the marker were still being deselected by `pytest` by default.

Related logic:
```python
def deselect_by_keyword(items: list[Item], config: Config) -> None:
    keywordexpr = config.option.keyword.lstrip()
    if not keywordexpr:
        return

    expr = _parse_expression(keywordexpr, "Wrong expression passed to '-k'")

    remaining = []
    deselected = []
    for colitem in items:
        if not expr.evaluate(KeywordMatcher.from_item(colitem)):
            deselected.append(colitem)
        else:
            remaining.append(colitem)

    if deselected:
        config.hook.pytest_deselected(items=deselected)
        items[:] = remaining
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
LouisTsai-Csie: Remove this logic as it is unnecessary, i update the filter logic part with more detailed comment!

packages/testing/src/execution_testing/specs/benchmark.py
jochem-brouwer: You are right regarding these options. I would advise to use `CALL` for this. When writing the tests, I think most writers will assume that the code runs in a separate environment (so in a separate contract). While `CALL` needs one more parameter (the call value), this thus "wastes" more gas on the loop (as overhead) (namely: a PUSH0 which is 2 gas). However, if we use `DELEGATECALL`, we run the target code in the current context. This also means that the storage you are reading/writing to is the storage of the current contract (the loop contract). This might not be what is assumed.
Instead, I would add an option to this benchmark loop code to use `DELEGATECALL` or `STATICCALL` for testers to override the standard `CALL`.
Note that with the `CALL` we thus run the code in a separate context, and we can for instance also target "bloated" storage contracts. This is not possible with `DELEGATECALL`, because this will use the storage of the current contract (which thus starts empty).

spencer-tb (approved):
Approved from my end (as going ooo)! :D

packages/testing/src/execution_testing/specs/benchmark.py
LouisTsai-Csie: I go with option 3 at the end, so if the attack block contains state changing operations, i would switch to `CALL` otherwise `STATICCALL`, this way we do not increase the overhead for most of the test, but only adjust these who changes state.

- SSTORE, TSTORE
- CREATE, CREATE2, SELFDESTRUCT
- CALL, CALLCODE
- LOG0-4

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -432,14 +432,18 @@ def fund_eoa(
         fund_tx: PendingTransaction | None = None
         if delegation is not None or storage is not None:
             if storage is not None:
+                # Handle both Storage objects and plain dicts
+                storage_dict = (
+                    storage.root if isinstance(storage, Storage) else storage
+                )
                 logger.debug(
-                    f"Deploying storage contract for EOA {eoa} with {len(storage.root)} storage slots"
+                    f"Deploying storage contract for EOA {eoa} with {len(storage_dict)} storage slots"
```
LouisTsai-Csie: Solving failing benchmark scenario due to typing issue

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -72,17 +74,45 @@ def deploy_fix_count_contracts(self, *, pre: Alloc, fork: Fork) -> Address:
...
+            Op.LOG3,
+            Op.LOG4,
+        ]
+
+        # Select CALL for state-changing opcodes, STATICCALL otherwise
+        uses_state_changing_opcode = any(
+            bytes(opcode) in bytes(self.attack_block)
+            for opcode in is_state_changing_set
+        )
+        call_opcode = Op.CALL if uses_state_changing_opcode else Op.STATICCALL
```
LouisTsai-Csie, line 109: Handling scenario like:

- Loop contract calls into target contract
- Target contracts contains state-changing operations.

If so, then loop contract could not use `STATICCALL` here, it should use `CALL` instead.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
fselmo: I wonder if instead of manual coercion / handling here, we can make use of pydantic more directly. It wasn't immediately obvious (and I couldn't find) what test was giving issues here but it would be really nice if this worked:

```py
@validate_call(config=ConfigDict(arbitrary_types_allowed=True))
def fund_eoa(
    self,
    ...
) -> EOA:
    ...

```
What this does is it will validate all the inputs as the models that they are defined, with the same type coercion that pydantic gives to model instantiation. I wouldn't spend a ton of time on it but it would be nice to use pydantic for all of it's bells and whistles if we can.

Do you know what test was giving issues here?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -110,14 +110,30 @@ def get_opcode_counts_for_test(
...
     if not gas_benchmark_value and fixed_opcode_count is None:
         return
 
+    # In --fixed-opcode-count mode, we only support tests that meet all of the following:
+    #   - The test uses the benchmark_test fixture
+    #   - The benchmark test uses a code generator
+    #
+    # Here we filter out tests that do not use the benchmark_test fixture.
+    # Note: At this stage we cannot filter based on whether a code generator is used.
+    if fixed_opcode_count or fixed_opcode_count == "":
```
fselmo: nit: I think this can be simplified to:

```suggestion
    if fixed_opcode_count is not None:
```

right?

tests/benchmark/compute/instruction/test_arithmetic.py
```diff
@@ -158,7 +159,7 @@ def test_arithmetic(
     )
 
 
-@pytest.mark.repricing(mod_bits=255)
+@pytest.mark.repricing(mod_bits=127)
```
fselmo, line 162: From the perspective of someone coming in without knowing the full flow well enough, it feels like these kinds of changes / settings could use some context. Just curious, is the grafana + script flow documented somewhere?

Don't let this hold up the changes though, just something to consider 🙂

packages/testing/src/execution_testing/specs/benchmark.py
fselmo: > This should be much less than 5%, i think at most 1% deviation is enough, but i need to dive deeper into this analysis.

Curious if this was considered in this PR already or if this is a thought for a follow-up PR.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -266,7 +286,6 @@ def fixed_opcode_count(request: pytest.FixtureRequest) -> int | None:
 
 
 BENCHMARKING_MAX_GAS = 1_000_000_000_000
-HIGH_GAS_LIMIT = 1_000_000_000
```
fselmo, line 269: Assuming this was agreed to somewhere, just looking for a bit of context on this change 👀

fselmo (commented):
From an implementation standpoint, this looks good on my end. I tested the tolerance and the counts, `CALL` vs `STATICCALL`, looks like it works as intended. The review looks good from an outsider's perspective, I just left some things to think about and some things for me to understand a bit better as I'm not so used to this flow yet.

I will take a look again tomorrow with fresh eyes. Really nice work 👌🏼

marioevz:
Rebased to latest `forks/amsterdam` to make review easier.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: I really like this idea, although this would basically touch every single test created, so perhaps for a follow up PR.

What I would fix in this PR though is to make it explicit that this can receive a Dict in that argument.

marioevz:
Created a PR on top of this to use pydantic: https://github.com/LouisTsai-Csie/execution-specs/pull/1

Other than that it looks good to me! @LouisTsai-Csie

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
LouisTsai-Csie: This breaks the `test_ext_account_query_warm` benchmark, its behavior is different under `fill` and `execute remote`. Now it passes during `fill` but fails when using `execute remote`, so now our CI cannot catch the bug!

tests/benchmark/compute/instruction/test_arithmetic.py
LouisTsai-Csie: More on this: the original repricing marker just randomly select parameters combination (i am sorry for this) to test out the functionality. After the infra structure is more well-organized, what we would need now is to label the worst case scenario based on the grafana dashboard. IIUC, this could help repricing group narrow down the test run.

If the new pricing could not even pass the worst case, then there is no need to run the second-worst (not sure if this is correct grammar) scenario.

So my approach is to check the grafana dashboard and sort the performance, pick the least param combination for each test and label them with repricing marker.

I am concerned here and highly likely i will re-do the process again, since the dashboard used an outdated naming/folder structure, that i need automatically maps the old test to new one. This makes the peer-review very hard. But now the test name is correctly updated!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
LouisTsai-Csie: For some of the tests, the `HIGH_GAS_LIMIT` is too low to run. And i notice we do not need two separate systems for the maximum gas limit, which will make the logic more complicated.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -54,103 +56,211 @@ def pytest_configure(config: pytest.Config) -> None:
...
+            )
+            for gas_value in self.root
+        ]
+
+
+class OpcodeCountsConfig(BaseModel, BenchmarkParametrizer):
+    """Opcode counts configuration object."""
+
+    scenario_configs: Dict[str, List[int]] = Field(default_factory=dict)
+    default_counts: List[int] = Field(default_factory=lambda _: [1])
```
danceratopz: This lambda doesn't take a positional argument. I think pydantic introspects and passes a value (which avoids a runtime `TypeError`). But I think it's cleaner to omit it.

```suggestion
    default_counts: List[int] = Field(default_factory=lambda : [1])
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
danceratopz: Unmodified in this PR, but this is dead code due to L258-L260.
```suggestion
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -54,103 +56,211 @@ def pytest_configure(config: pytest.Config) -> None:
...
+        # If the test does not have the repricing marker, skip it
         repricing_marker = item.get_closest_marker("repricing")
         if not repricing_marker:
             continue
 
+        # If the test has the repricing marker but no specific kwargs,
+        # include the entire parametrized test in the filtered list.
         if not repricing_marker.kwargs:
             filtered.append(item)
             continue
```
danceratopz, line 255: I.e., the check is already here.

packages/testing/src/execution_testing/cli/benchmark_parser.py
```diff
@@ -270,18 +268,17 @@ def categorize_patterns(
 def generate_config_json(
     config: dict[str, list[int]],
     pattern_sources: dict[str, Path],
-) -> str:
+) -> OpcodeCountsConfig:
```
danceratopz, line 272: I think setting the `default_counts` via `OpcodeCountsConfig` would be a little cleaner...
```suggestion
    default_counts: list[int],
) -> OpcodeCountsConfig:
```

packages/testing/src/execution_testing/cli/benchmark_parser.py
```diff
@@ -270,18 +268,17 @@ def categorize_patterns(
...
     """Generate the JSON config file content."""
     categories = categorize_patterns(config, pattern_sources)
 
     scenario_configs: dict[str, list[int]] = {}
+    output = OpcodeCountsConfig(scenario_configs=scenario_configs)
```
danceratopz: ...then:
```suggestion
    output = OpcodeCountsConfig(
        scenario_configs=scenario_configs,
        default_counts=default_counts,
    )
```

packages/testing/src/execution_testing/cli/benchmark_parser.py
```diff
@@ -355,7 +353,10 @@ def main() -> int:
             del merged[pattern]
 
     content = generate_config_json(merged, pattern_sources)
-    config_file.write_text(content)
+    content.default_counts = existing_file.default_counts
```
danceratopz: ...means that we can avoid this change which appears as an afterthought.
```suggestion
```

danceratopz (commented):
This is looking good. A few very minor nits below!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
LouisTsai-Csie: Not sure what to change, could you provide more details here?

danceratopz: Nothing to change in this section of the code, just showing why we can remove the code in the above https://github.com/ethereum/execution-specs/pull/1869#discussion_r2650894489.

LouisTsai-Csie: Thanks! I remove the dead code

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -54,192 +56,229 @@ def pytest_configure(config: pytest.Config) -> None:
...
+            else:
+                warnings.warn(
+                    "--fixed-opcode-count was provided without a value, but "
+                    f"{cls.default_config_file_name} was not found. "
+                    "Run 'uv run benchmark_parser' to generate it, or provide "
+                    "explicit values (e.g., --fixed-opcode-count 1,10,100).",
+                    UserWarning,
+                    stacklevel=1,
+                )
+                return cls()
```
danceratopz: I think an explicit fail is more obvious behavior than a warning and running with an unexpected default value.
```suggestion
                pytest.UsageError(
                    "--fixed-opcode-count was provided without a value, but "
                    f"{cls.default_config_file_name} was not found. "
                    "Run 'uv run benchmark_parser' to generate it, or provide "
                      "explicit values (e.g., --fixed-opcode-count 1,10,100)."
                )
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -1,14 +1,16 @@
 """The module contains the pytest hooks for the gas benchmark values."""
 
-import json
 import re
 import warnings
```
danceratopz:
```suggestion
```

danceratopz (commented):
Thanks @LouisTsai-Csie! One more minor nit!

danceratopz (approved):
Thanks @LouisTsai-Csie! LGTM!

PR #1872, opened 2025-12-09
feat(tests): add benchmark for transfers to non-existent receivers

MariusVanDerWijden:
## 🗒️ Description
Adds a benchmark that touches many non-existent accounts

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -37,6 +37,59 @@ from execution_testing import (
 )
 
 
+def test_contract_calling_many_addresses(
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    fork: Fork,
+) -> None:
+    """
+    Benchmark a contract that calls many addresses.
+
+    Creates a contract that generates addresses dynamically using KECCAK256
+    and calls each one with 1 wei.
+    Stop when the gas is less than the loop cost.
+    """
+    gas_costs = fork.gas_costs()
+    loop_cost = (
+        gas_costs.G_JUMPDEST  # 1
+        + gas_costs.G_VERY_LOW * 4  # 4× PUSH1 for CALL zero args = 12
+        + gas_costs.G_BASE * 2  # 2× GAS = 4
+        + gas_costs.G_VERY_LOW * 5  # PUSH1 value, DUP6, SWAP1, SUB, GT = 15
+        + gas_costs.G_VERY_LOW * 2  # PUSH1 dest, PUSH2 loop_cost = 6
+        + gas_costs.G_HIGH  # JUMPI = 10
+        + gas_costs.G_COLD_ACCOUNT_ACCESS  # 2600
+        + gas_costs.G_CALL_VALUE  # 9000
+        + gas_costs.G_NEW_ACCOUNT  # 25000
+        - gas_costs.G_CALL_STIPEND  # -2300 (refunded when callee has no code)
+    )
+    # Build contract code that generates addresses and calls them
+    # Pattern:
+    # 1. Pushes the current blockhash on the stack
+    # 2. JUMPDEST (loop start)
+    # 3. Call into the address at the top of the stack with 1 wei
+    # 4. Subtract 1 from the blockhash
+    # 5. Jump to the start of the loop if the gas is greater than the loop cost
+    # 6. otherwise, return 0
+
+    setup = Op.BLOCKHASH(0)
+    code = (
+        setup
+        + Op.JUMPDEST  # [BLOCKHASH]
+        + Op.CALL(address=Op.DUP6, value=1)  # [1, BLOCKHASH]
+        + Op.SWAP1  # [BLOCKHASH, 1] // Assume every CALL is successful
+        + Op.SUB  # [BLOCKHASH-1]
+        + Op.JUMPI(Op.GT(Op.GAS, loop_cost), len(setup))
+        + Op.RETURN(0, 0)
+    )
+
+    benchmark_test(
+        tx=Transaction(
+            to=pre.deploy_contract(code=code),
+            sender=pre.fund_eoa(),
+        )
+    )
+
+
 @pytest.mark.repricing(max_code_size_ratio=0)
 @pytest.mark.parametrize(
     "opcode",
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -17,7 +17,9 @@ import math
 
 import pytest
 from execution_testing import (
+    AccessList,
     Account,
+    Address,
     Alloc,
     BenchmarkTestFiller,
     Block,
@@ -37,55 +39,59 @@ from execution_testing import (
 )
 
 
+@pytest.mark.parametrize("transfer_amount", [0, 1])
+@pytest.mark.parametrize("opcode", [Op.CALL, Op.CALLCODE])
+@pytest.mark.parametrize("access_warm", [True, False])
 def test_contract_calling_many_addresses(
     benchmark_test: BenchmarkTestFiller,
     pre: Alloc,
     fork: Fork,
+    transfer_amount: int,
+    opcode: Op,
+    access_warm: bool,
+    tx_gas_limit: int,
 ) -> None:
-    """
-    Benchmark a contract that calls many addresses.
+    """Benchmark a contract that calls many addresses."""
+    warm_start_addr = 2**80 - 1
+    setup = Op.PUSH20(warm_start_addr) if access_warm else Op.GAS
 
-    Creates a contract that generates addresses dynamically using KECCAK256
-    and calls each one with 1 wei.
-    Stop when the gas is less than the loop cost.
-    """
-    gas_costs = fork.gas_costs()
-    loop_cost = (
-        gas_costs.G_JUMPDEST  # 1
-        + gas_costs.G_VERY_LOW * 4  # 4× PUSH1 for CALL zero args = 12
-        + gas_costs.G_BASE * 2  # 2× GAS = 4
-        + gas_costs.G_VERY_LOW * 5  # PUSH1 value, DUP6, SWAP1, SUB, GT = 15
-        + gas_costs.G_VERY_LOW * 2  # PUSH1 dest, PUSH2 loop_cost = 6
-        + gas_costs.G_HIGH  # JUMPI = 10
-        + gas_costs.G_COLD_ACCOUNT_ACCESS  # 2600
-        + gas_costs.G_CALL_VALUE  # 9000
-        + gas_costs.G_NEW_ACCOUNT  # 25000
-        - gas_costs.G_CALL_STIPEND  # -2300 (refunded when callee has no code)
-    )
-    # Build contract code that generates addresses and calls them
-    # Pattern:
-    # 1. Pushes the current blockhash on the stack
-    # 2. JUMPDEST (loop start)
-    # 3. Call into the address at the top of the stack with 1 wei
-    # 4. Subtract 1 from the blockhash
-    # 5. Jump to the start of the loop if the gas is greater than the loop cost
-    # 6. otherwise, return 0
+    def loop(threshold: int) -> Bytecode:
+        return (
+            Op.JUMPDEST
+            + opcode(address=Op.DUP6, value=transfer_amount)
+            + Op.SWAP1
+            + Op.SUB
+            + Op.JUMPI(Op.GT(Op.GAS, threshold), len(setup))
+        )
 
-    setup = Op.BLOCKHASH(0)
-    code = (
-        setup
-        + Op.JUMPDEST  # [BLOCKHASH]
-        + Op.CALL(address=Op.DUP6, value=1)  # [1, BLOCKHASH]
-        + Op.SWAP1  # [BLOCKHASH, 1] // Assume every CALL is successful
-        + Op.SUB  # [BLOCKHASH-1]
-        + Op.JUMPI(Op.GT(Op.GAS, loop_cost), len(setup))
-        + Op.RETURN(0, 0)
-    )
+    cost = loop(0xFFFF).gas_cost(fork)
+    code = setup + loop(cost)
+
+    access_list = None
+    if access_warm:
+        gas_costs = fork.gas_costs()
+        intrinsic_cost_calc = fork.transaction_intrinsic_cost_calculator()
+
+        iterations = (tx_gas_limit - intrinsic_cost_calc()) // (
+            gas_costs.G_ACCESS_LIST_ADDRESS + cost
+        )
+
+        access_list = [
+            AccessList(
+                address=Address(warm_start_addr - i),
+                storage_keys=[],
+            )
+            for i in range(iterations)
+        ]
 
     benchmark_test(
         tx=Transaction(
-            to=pre.deploy_contract(code=code),
+            to=pre.deploy_contract(
+                code=code,
+                balance=10**18 if transfer_amount > 0 else 0,
+            ),
             sender=pre.fund_eoa(),
+            access_list=access_list,
         )
     )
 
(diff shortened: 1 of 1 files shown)


tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -67,6 +67,14 @@ def get_distinct_receiver_list(
         yield pre.fund_eoa(balance)
 
 
+def get_distinct_unfunded_receiver_list(
+    pre: Alloc,
+) -> Generator[Address, None, None]:
+    """Get a list of distinct unfunded receiver accounts."""
+    while True:
+        yield pre.empty_account()
```
jochem-brouwer: Should be checked if this account does not end up in genesis as empty ([EIP-161 "empty" ](https://eips.ethereum.org/EIPS/eip-161))

jochem-brouwer (commented):
This test touches and creates (value sent is 1) many accounts via a transaction. Might also want to expand on touching those (but not creating them -> sending 0 value) and doing this in a transaction by creating them via a CALL loop.

MariusVanDerWijden:
@jochem-brouwer I think the parameterization of the benchmark ("balance", [0,1]) already does the second part (touching but not creating them)
I pushed a second test for the the first part (creating from within a contract)

jochem-brouwer:
You are right, I'm not sure why I missed the part where it also "sends" 0 :sweat_smile:

MariusVanDerWijden:
Hmm I don't get why this fails:
```
 E   execution_testing.base_types.composite_types.Account.BalanceMismatchError: unexpected balance for account 0x57d5192d3682925609bb7f6188a5dc9daeff09ff: want 0x02, got 0x01
 ```

**Comment by jochem-brouwer** [comment]
For the other tests, where the target account(s) exist, the account is initialized with a balance. The test checks if the expected balance is met for the targets. However, in this test we create empty accounts, so only balance = 0 is valid. Balance = 1 will make the test think it yields accounts (which also get 1 more balance) which now have balance of 2. This is incorrect. For the empty test case, skip the test with balance=1:

```python
    if case_id == "a_to_nonexistent" and balance == 1:
        pytest.skip("balance=1 is invalid for a_to_nonexistent")
```

(Should be added to `test_block_full_of_ether_transfers`)

jochem-brouwer:
Actually I think the test for one account to many non-existent already exists: this is the case for balance = 0. It targets these non-existent accounts. So the case `a_to_diff_acc` with `balance=0` is the same as the test created in `test_transaction_types`. So `a_to_nonexistent` does not add a new test but is a duplicate.

The `balance` used here is the balance where the accounts are initialized with (the receivers). It is not the value sent (this is always 1 in the `test_block_full_of_ether_transfers` test). The balances are parametrized as [0, 1]. The zero case is the non-existent account case.

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -151,6 +165,7 @@ def test_block_full_of_ether_transfers(
     - diff_acc_to_b: multiple senders → one receiver
     - a_to_diff_acc: one sender → multiple receivers
     - diff_acc_to_diff_acc: multiple senders → multiple receivers
+    - a_to_nonexistent: one sender → non-existent receivers
```
jochem-brouwer: This is the same as the `test_block_full_of_ether_transfers[fork_Osaka-benchmark-gas-value_1M-blockchain_test-balance_0-case_id_a_to_diff_acc]` test 😄 👍

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -247,6 +247,42 @@ def test_xcall(
...
+def test_contract_calling_many_addresses(
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    fork: Fork,
+) -> None:
+    """
+    Benchmark a contract that calls many addresses.
+
+    Creates a contract that generates addresses dynamically using KECCAK256
+    and calls each one with 1 wei. The contract loops until it runs out of gas.
```
jochem-brouwer: Is the OOG intended? This rollbacks all created accounts, so the state root calculation is rather easy.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -247,6 +247,42 @@ def test_xcall(
...
+    # Build contract code that generates addresses and calls them
+    # Pattern:
+    # 1. Store NUMBER in memory[0] (32 bytes)
+    # 2. JUMPDEST (loop start)
+    # 3. Update memory[0] with KECCAK256(memory[0], 32)
+    # 4. CALL address from memory[0] (first 20 bytes) with 1 wei
+    # 5. JUMP back to JUMPDEST
+    setup = Op.MSTORE(0, Op.NUMBER())  # Initialize with block number
+    attack_block = (
+        Op.MSTORE(0, Op.SHA3(0, 32))  # Generate new address via hash
```
jochem-brouwer: Can also hash once to pick a starting point and then increment the address by 1 each loop. This saves gas by not spending it on the hashing. (Can use the `1` status from the CALL to directly feed in the addition)

jochem-brouwer: Note: this targets the same addresses if the block consists of multiple transactions (these are split up due to Fusaka's 7825 gas limit). So if two transactions are in the block, these will both OOG and will both attempt to create the same accounts (if the gas limit of both txs is equal)

jochem-brouwer (commented):
Some more comments 😄 👍

MariusVanDerWijden:
> The zero case is the non-existent account case.
I thought that as well, but I somehow convinced myself that its not. I don't remember the specifics, but iirc the other tests have the accounts in the state already, so they are not non-existent, they are existent with 0 balance

jochem-brouwer:
If I check this test: `test_block_full_of_ether_transfers[fork_Osaka-benchmark-gas-value_1M-blockchain_test-balance_0-case_id_a_to_diff_acc]` fixture output then I see a pre state which is almost empty and a post state which has a lot of new accounts. I might also be on some outdated version of EELS though :thinking:

MariusVanDerWijden:
Ah okay, yeah so I will remove this one

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -276,6 +285,8 @@ def test_contract_calling_many_addresses(
             ),
         ),
     )
+    attack_block += Op.JUMPI(Op.GT(Op.GAS, loop_cost), destination)
+    attack_block += Op.RETURN(0, 0)
 
     benchmark_test(
         code_generator=JumpLoopGenerator(
```
MariusVanDerWijden: This is not really correct, we don't need a JumpLoopGenerator here anymore, since I added the loop earlier. I am thinking about how to make it cleaner

MariusVanDerWijden: I wonder if it makes sense to have a similar generator to `JumpLoopGenerator` that basically does what I'm doing, have an if condition in the loop that stops on some condition.
Basically:
```
JumpLoopNoOOGGenerator(
    setup=setup,
    attack_block=attack_block,
    condition=Op.GT(Op.GAS, loop_cost)
    teardown=Op.Return(0,0)
)
```
Which would do something like 
```
setup + jumpdest + attack_block + condition + jumpi (condition, jumpdest) + teardown
```

wdyt @jochem-brouwer @LouisTsai-Csie

jochem-brouwer: @LouisTsai-Csie I think an exit condition for the loop and teardown is useful for the situations which we don't want to run OOG in the JumpLoopGenerator.

Without this feature you can also implement a While loop, as seen here:

https://github.com/ethereum/execution-specs/blob/3a85f3db8fd88c22f8889db6b03e2997a788ed16/tests/benchmark/compute/instruction/test_system.py#L590-L596

LouisTsai-Csie: For this scenario, I think we do not need an additional benchmark code generator. The contract logic itself already handles the OOG condition, so we can simply pass a single transaction to the benchmark test wrapper like this:

```python
benchmark_test(
  tx=Transaction(
      to=pre.deploy_contract(code=code, balance=10**8),
      sender=pre.fund_eoa(),
  )
)
```

The wrapper will then create a set of transactions that follows eip7825, and the execution will naturally stop once the transaction runs out of gas. But we need to be really careful with the gas usage here.

https://github.com/ethereum/execution-specs/blob/dca59f62f16089e6a3256979b9d218976570140f/packages/testing/src/execution_testing/specs/benchmark.py#L322-L350

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -247,6 +247,55 @@ def test_xcall(
...
+        + gas_costs.G_LOW * 3  # ~Gluing opcodes
+    )
+    # Build contract code that generates addresses and calls them
+    # Pattern:
+    # 1. Store Keccak256(NUMBER) in memory[0] (32 bytes)
+    # 2. JUMPDEST (loop start)
+    # 4. CALL address from memory[0] (first 20 bytes) with 1 wei
+    # 5. Update memory[0] with ADD(memory[0], 1)
+    # 6. JUMP back to JUMPDEST
+    setup = Op.MSTORE(0, Op.NUMBER())  # Initialize with block number
```
LouisTsai-Csie: As mentioned in this comment, if we are running a block with multiple transactions, each transactions will start by the same index, thus sending ether to the same address, which is not our intended logic.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -247,6 +247,55 @@ def test_xcall(
...
+    destination = Op.JUMPDEST()
+    attack_block = Op.MSTORE(
+        0,
+        Op.ADD(
+            1,
+            Op.CALL(
+                address=Op.MLOAD(0),  # Load address from memory
+                value=1,  # Send 1 wei
+            ),
+        ),
```
LouisTsai-Csie: This would send ether to the same addr every time, since the `CALL` result is either 0 or 1. The result of `ADD` would therefore be either (1 + 0) or (1 + 1), meaning that memory slot 0 would always be the same value.

Instead of the relying on the memory layout, i propose using a stack-based solution instead. More details in the following comment.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -247,6 +247,55 @@ def test_xcall(
...
+    )
+    attack_block += Op.JUMPI(Op.GT(Op.GAS, loop_cost), destination)
```
LouisTsai-Csie:
```suggestion
    attack_block += Op.JUMPI(Op.GT(Op.GAS, loop_cost), len(setup))
```

The program count provided should be an index instead of an opcode.

LouisTsai-Csie (changes requested):
I would suggest a stack-based solution for this scenario, this is my idea:

```python
code = (
  Op.SELFBALANCE # [BALANCE]
  + Op.JUMPDEST # [BALANCE]
  + Op.CALL(address=Op.DUP6, value=1) # [1, BALANCE] Here assume every CALL is successful
  + Op.SWAP1 # [BALANCE, 1] 
  + Op.SUB # [BALANCE-1]
  + Op.JUMPI(Op.GT(Op.GAS, loop_cost), len(Op.SELFBALANCE))
)
```

We could pick a reasonable initial balance as the starting address. Since the balance is decreasing every time, we could ensure the target address differs at each call. Even there are multiple transaction, this logic still prevents duplicate ether transfer target.

To prevent OOG issue, we need to make sure this loop cost is calculated precisely. For my implementation, the loop cost would be:

```python
loop_cost = (
    gas_costs.G_JUMPDEST  # 1
    + gas_costs.G_VERY_LOW * 4  # 4× PUSH1 for CALL zero args = 12
    + gas_costs.G_BASE * 2  # 2× GAS = 4
    + gas_costs.G_VERY_LOW * 5  # PUSH1 value, DUP6, SWAP1, SUB, GT = 15
    + gas_costs.G_VERY_LOW * 2  # PUSH1 dest, PUSH2 loop_cost = 6
    + gas_costs.G_HIGH  # JUMPI = 10
    + gas_costs.G_COLD_ACCOUNT_ACCESS  # 2600
    + gas_costs.G_CALL_VALUE  # 9000
    + gas_costs.G_NEW_ACCOUNT  # 25000
    - gas_costs.G_CALL_STIPEND  # -2300 (refunded when callee has no code)
)

```

Please correct me if this is wrong.

---

I think we could even replace every address calculation with a simple `SELFBALANCE` like this:

```python
Op.JUMPDEST
+ Op.POP(Op.CALL(address=Op.SELFBALANCE, value=1))
+ Op.JUMPI(Op.GT(Op.GAS, loop_cost), len(Op.SELFBALANCE))
```

LouisTsai-Csie:
One question - what happens if OOG? Does receiver still get ether or everything is reverted?

MariusVanDerWijden:
We can use the stack instead of the memory. I don't think we should base it on balance though, because then it wouldn't be replay-able twice on bloatnet

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -267,6 +267,59 @@ def _deploy_max_contract_factory(
...
+        + Op.CALL(address=Op.DUP6, value=1)  # [1, BLOCKHASH]
+        + Op.SWAP1  # [BLOCKHASH, 1] // Assume every CALL is successful
+        + Op.SUB  # [BLOCKHASH-1]
+        + Op.JUMPI(Op.GT(Op.GAS, loop_cost), len(setup))
+        + Op.RETURN(0, 0)
+    )
+
+    benchmark_test(
+        tx=Transaction(
+            to=pre.deploy_contract(code=code),
```
LouisTsai-Csie:
```suggestion
            to=pre.deploy_contract(code=code, balance=10**18),
```

We should fund this contract, so it could transfer ether to accounts.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -267,6 +267,59 @@ def _deploy_max_contract_factory(
...
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    fork: Fork,
+) -> None:
+    """
+    Benchmark a contract that calls many addresses.
+
+    Creates a contract that generates addresses dynamically using KECCAK256
+    and calls each one with 1 wei.
+    Stop when the gas is less than the loop cost.
```
LouisTsai-Csie: We are using BLOCKHASH(0) to count the address now, not using KECCAK256.

LouisTsai-Csie (changes requested):
left a few small changes. Also, maybe we could clean up some of the comments—I think the code itself is already very clear.

LouisTsai-Csie (approved):
Add some fixes

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -37,6 +39,63 @@
 )
 
 
+@pytest.mark.parametrize("transfer_amount", [0, 1])
+@pytest.mark.parametrize("opcode", [Op.CALL, Op.CALLCODE])
```
jochem-brouwer, line 43: Why no `DELEGATECALL` and `STATICCALL`? For the value transfers, skip those tests 👍

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -37,6 +39,63 @@
...
+        gas_costs = fork.gas_costs()
+        intrinsic_cost_calc = fork.transaction_intrinsic_cost_calculator()
+
+        iterations = (tx_gas_limit - intrinsic_cost_calc()) // (
+            gas_costs.G_ACCESS_LIST_ADDRESS + cost
+        )
+
+        access_list = [
+            AccessList(
+                address=Address(warm_start_addr - i),
```
jochem-brouwer: Adding items to the access list will increase the tx cost and will therefore impact the iterations. Due to Fusaka the transactions will now be split up in multiple ones, but since the setup always pushes the same address, every transaction will interact with the same addresses (which is not what is intended to test here since we want as many unique targets as possible)

tests/benchmark/compute/instruction/test_system.py
spencer-tb: Agree here, happy with the skips. I think we should clean up all the skips in a follow up PR though. We are skipping 1/3 of all benchmark tests that are collected now lol

spencer-tb (approved):

PR #1882, opened 2025-12-10
feat(test-tests): Expand BAL CALL opcode OOG boundary test cases

fselmo:
## 🗒️ Description

Expand on the OOG boundary cases for CALL* opcodes to parametrize for:

- value / no value (CALL and CALLCODE only)
- mem expansion / no expansion
- 7702 delegation / no delegation
  - COLD / WARM delegation
- COLD / WARM target
- empty / existing target (CALL only - affects create_cost)

Test both before and after state access gas boundaries to make sure client implementations match accordingly. `oog_after_target_access` is only meaningful when there's a gap between first and second gas check:
  - CALL: gap exists only when `value > 0` AND `target is empty (G_NEW_ACCOUNT = 25000)`
  - CALLCODE/DELEGATECALL/STATICCALL: no gap without 7702 delegation
  - All 7702 tests: always have a gap (delegation_cost = 2600 cold / 100 warm)

One scenario I considered but didn't explore is:

- Beyond the first static check for state access, we are currently only checking `static_check + 1`. We do not test `static_check + delegation_cost - 1`. I think this is OK because the gap between the two shouldn't matter with implementation here... but I can be convinced that if we provide just enough for the delegation minus `1` that this could potentially be a meaningful boundary to check.

---
Note:

- `test_bal_code_with_value_transfer` was removed as this case is now covered in the new parametrization. 
- `test_bal_callcode_with_value_transfer` -> `test_bal_callcode_nested_value_transfer` renamed to better match that we track the code execution of the CALL from the CALLCODE which we do not do in the gas boundary tests - so this test remains unique.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -47,6 +47,46 @@
 pytestmark = pytest.mark.valid_from("Amsterdam")
 
 
+@pytest.fixture
+def call_memory_expansion_cost(fork: Fork) -> Callable[[int], int]:
```
marioevz: There's a function in `Fork` that does this and is updated with each fork:
https://github.com/ethereum/execution-specs/blob/cc82819f277c014ef900250f9954c964011389e8/packages/testing/src/execution_testing/forks/base_fork.py#L358-L367

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,259 +416,1614 @@ def test_bal_extcodesize_and_oog(
...
+        pre.empty_account()
+        if target_is_empty
+        else pre.deploy_contract(code=Op.STOP)
     )
 
-    call_contract = pre.deploy_contract(code=call_contract_code)
+    # warmup code if needed
+    warmup_code = Bytecode()
+    if target_is_warm:
+        warmup_code = Op.POP(Op.EXTCODESIZE(target))
```
raxhvl: BAL will always have `target` if warmed up via `EXTCODESIZE`. Which means this check will true even if the `CALL` fails:

```sh
 account_expectations[target] = BalAccountExpectation.empty()
```

This is evident for warm up + no value transfer, we cannot determine if the `CALL` went through.

You can set `gas_limit` to `gas_limit -1` and this combination will still return a false positive (even though `CALL` goes OOG).

```sh
uv run fill 'tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py::test_bal_call_success_no_delegation' --watch --clean --until Amsterdam --traces --evm-dump-dir logs -k "no_value and warm_target"
```

fselmo: That's a good point. We can certainly check that it went through (it does) but we should have a better check on the test. I will add this.

raxhvl: We could warm up using 2390 since BAL DOES NOT include them by default unless touched by evm.

jochem-brouwer: Yes and this could also open up more edge cases. We need tests which put the targets in the 2930 access list and we need cases where it needs to be added to BAL and not.
To "warm up" an account we should thus do this via the access list and not inside EVM, because it's a good point, this will put them in the BAL.

fselmo: Working on this now, agreed all around 👍🏼

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,259 +416,1614 @@ def test_bal_extcodesize_and_oog(
...
-    push_cost = gas_costs.G_VERY_LOW * 7
-    call_cold_cost = gas_costs.G_COLD_ACCOUNT_ACCESS
-    tx_gas_limit = intrinsic_gas_cost + push_cost + call_cold_cost
-
-    if fails_at_call:
-        # subtract 1 gas to ensure OOG at CALL
-        tx_gas_limit -= 1
+    # caller contract
+    call_code = warmup_code + Op.CALL(
+        gas=0, address=target, value=value, ret_size=ret_size
```
jochem-brouwer: For completeness I'd also add here the output offset being zero here. If somehow the output offset would get encoded as 64, and the ret size is also 32, we are thus now having a memory size of 96 and not 32.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,259 +416,1614 @@ def test_bal_extcodesize_and_oog(
...
-        sender=alice,
-        to=call_contract,
-        gas_limit=tx_gas_limit,
+    intrinsic_cost = fork.transaction_intrinsic_cost_calculator()()
+
+    # Warmup cost: bytecode (PUSH20 + POP) + state access (EXTCODESIZE cold)
+    warmup_total = (
+        (call_warmup_bytecode_cost + gas_costs.G_COLD_ACCOUNT_ACCESS)
+        if target_is_warm
+        else 0
```
jochem-brouwer: Does this include the base cost? Even if the account is warm, CALLs still cost 100 gas (where is this included?)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,259 +416,1614 @@ def test_bal_extcodesize_and_oog(
...
-                # Bob should only appear if CALL succeeded
-                **(
-                    {bob: None}
-                    if fails_at_call
-                    else {bob: BalAccountExpectation.empty()}
-                ),
-            }
-        ),
+    # Bytecode cost: 7 pushes for Op.CALL + warmup
+    bytecode_cost = gas_costs.G_VERY_LOW * 7 + warmup_total
```
jochem-brouwer: This is correct if when push 0 is encoded as `PUSH1(0)` (3 gas) (or, anything which is not `PUSH0`, so `PUSH20(0)` also works) and not encoded as `PUSH0` (2 gas)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,259 +416,1614 @@ def test_bal_extcodesize_and_oog(
...
+
+    # Access cost for CALL
+    access_cost = (
+        gas_costs.G_WARM_ACCOUNT_ACCESS
+        if target_is_warm
+        else gas_costs.G_COLD_ACCOUNT_ACCESS
```
jochem-brouwer, line 470: Hmm I believe if not `target_is_warm` `G_COLD_ACCOUNT_ACCESS` is added twice here? See line 466

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,259 +416,1614 @@ def test_bal_extcodesize_and_oog(
...
+
+    if value_transferred:
+        account_expectations[caller] = BalAccountExpectation(
+            balance_changes=[BalBalanceChange(tx_index=1, post_balance=0)]
+        )
+        account_expectations[target] = BalAccountExpectation(
+            balance_changes=[BalBalanceChange(tx_index=1, post_balance=value)]
+        )
+    else:
+        account_expectations[caller] = BalAccountExpectation.empty()
```
jochem-brouwer: Why is caller empty here? We target it and run its code - right?

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,259 +416,1614 @@ def test_bal_extcodesize_and_oog(
...
+        else gas_costs.G_COLD_ACCOUNT_ACCESS
     )
+    transfer_cost = gas_costs.G_CALL_VALUE if value > 0 else 0
+    memory_cost = fork.memory_expansion_gas_calculator()(new_bytes=ret_size)
+
+    # static gas cost (before state access): access + transfer + memory
+    static_gas_cost = access_cost + transfer_cost + memory_cost
+
+    # Fail at before any state access
+    gas_limit = intrinsic_cost + bytecode_cost + static_gas_cost - 1
```
jochem-brouwer: I believe this test should be merged with the previous one and then add a parametrization for the OOG case. This test looks the same except that it has different gas limit (correct me if I'm wrong 😄) It should be tested at exactly the boundary, where either a BAL item is added or not. 

(So: `gas_limit - 1` -> no item in BAL, `gas_limit` -> item in BAL)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
jochem-brouwer: Nvm empty bal is an empty item, not a non-existent BAL item, which is `None`. TIL

jochem-brouwer (commented):
I have not reviewed all test cases, but I am getting the impression that we can parameterize all these test cases. I got the feeling that these tests are now duplicates of the test previously, with a change (7702 delegation, or OOG rules).

I think we should parametrize on (besides what is already added as param)

- Target is 7702 delegated or not
- "Reason" for OOG (or non-OOG)
  - Insufficient gas in general
  - Enough gas to load `target`, but insufficient to load the delegated account
  - Enough gas to call both target and the 7702-delegation

To do so, we essentially need to calculate the necessary gas and then parametrize on the OOG or non-OOG (so this is the gas needed minus 1, or the actual gas needed). The difference here is between those two gas limits that an item is added to the BAL, or not.

We need to calculate the exact gas necessary to perform these operations. We would do so also for the parameter if the target is 7702 is delegated or not.

If it is not delegated, then `target` is in BAL if we have enough gas to call it. `target` is still in BAL if it is 7702-delegated, but the delegated account is NOT in BAL. If we add the sufficient gas (this is to either warm or cold), then this is at the gas boundary of adding the delegated account to BAL or not.

See also this part from [EIP-7702:](https://eips.ethereum.org/EIPS/eip-7702)

> If a code executing instruction accesses a cold account during the resolution of delegated code, add an additional [EIP-2929](https://eips.ethereum.org/EIPS/eip-2929) `COLD_ACCOUNT_READ_COST` cost of `2600` gas to the normal cost and add the account to `accessed_addresses`. Otherwise, assess a `WARM_STORAGE_READ_COST` cost of `100`.

As final remark, to warm an account, we should not use `POP(EXTCODESIZE(T))` as mentioned by @raxhvl, as this adds `T` to BAL which we in some cases do not want. Instead if the target is warmed we should add it as EIP-2930 access list address.

NOTE: I also realize that adding a lot of parameters might blow up the test cases here. However I now get the feeling we are duplicating too much in the tests.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
raxhvl: > Yes and this could also open up more edge cases. We need tests which put the targets in the 2930 access list and we need cases where it needs to be added to BAL and not.


@jochem-brouwer We have covered those [edge cases here.](https://github.com/ethereum/execution-specs/blob/3a85f3db8fd88c22f8889db6b03e2997a788ed16/tests/amsterdam/eip7928_block_level_access_lists/test_cases.md?plain=1#L15-L18)

image

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: Yes! I agree. This separation of tests was an effort to untangle some repetitive cases but I agree these two should be put back together as they are essentially the same minus 1 gas 👍🏼

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: The `100` gas is the warm access base cost. It's accounted for in the `access_cost`.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: Yes, the framework uses `PUSH1()` to prevent any ambiguity and for compatibility across forks that cannot use `PUSH0` instruction ([here](https://github.com/fselmo/execution-specs/blob/e58986d12f2586435066ab98c051050f93aa9e7f/packages/testing/src/execution_testing/vm/opcodes.py#L57-L60)).

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: I believe this is stale now? 👀

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: Agreed, used access lists here. Very nice idea! ❤️

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: These default to `0`, would you prefer to see this explicitly set here?

fselmo: added :)

fselmo:
@jochem-brouwer, I have added the extra gas boundary for delegated cases of `success_minus_1` gas. This means that when we have a delegation, if we OOG with exact gas minus `1` we should not have the delegation in BAL. I added this and then re-read your comment and I think this is what you meant as well? Let me know what you think. And thanks for the comments!

I thought about adding this to the `no_delegation_oog_after_state_access` tests but there really is no other boundary here. If we OOG with just enough for static gas (state access), we reach into the account (this is tested). If we OOG with exact gas minus `1` in this case, there is no new meaningful difference here in the BAL, target is still in BAL and nothing else could've been added to it so nothing else to check. Wdyt?

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
jochem-brouwer: Yup stale 👍 I should permalink those lines instead of reffing them like this 😓

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,159 +406,1216 @@ def test_bal_extcodesize_and_oog(
...
+
+    # static gas (before state access): access + transfer + memory
+    static_gas_cost = access_cost + transfer_cost + memory_cost
+    # second check includes create_cost
+    second_check_cost = static_gas_cost + create_cost
+
+    if oog_boundary == OutOfGasBoundary.OOG_BEFORE_TARGET_ACCESS:
+        gas_limit = intrinsic_cost + bytecode_cost + static_gas_cost - 1
+    else:  # SUCCESS
+        gas_limit = intrinsic_cost + bytecode_cost + second_check_cost
```
jochem-brouwer, line 488: This misses a test case. We have three cases here.

`OOG_BEFORE_TARGET_ACCESS`: insufficient gas for the static checks -> does not add `target` to BAL
`OOG_AFTER_TARGET_ACCESS`: insufficient gas for static checks + creating the account (this thus only applies for the situation where `create_cost` is nonzero) -> this adds `target` to BAL
`SUCCESS`: sufficient gas (should be at the exact gas boundary so setting -1 tx gas limit would in some cases remove `target` from BAL. Note: *some*,  because the case where we have enough for static cost, but not enough for static+create cost would in that case still add `target` to BAL. To remove it from BAL, we need to subtract 1, plus the `create_cost` fee, in order to remove it)

Man, those cases are complex 😅 

NOTE: The 4 cases as described here, I'm not sure if we can universally apply them, because these are edge cases. For instance, the missing test case here is an edge case which only applies to Op.CALL when we try to create a new account.

jochem-brouwer: Also, for a fuzzer, we'd want to call into this contract with a wide range of gas limits. One tx per block, and then to see if the BALs agree. @qu0b maybe this is something for you to pick up with the fuzzers? This fuzzer should parametrize the call value, the gas limit, the target, warm/cold target (access list) and if target exists or not (only relevant when callvalue != 0). And then for these situations execute it for a wide range of  gas limits to see if all clients agree on this (not some weird optimizations which could randomly add target to BAL or not). Not sure if this makes sense 🤔

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,159 +406,1216 @@ def test_bal_extcodesize_and_oog(
...
+    target = (
+        pre.empty_account()
+        if target_is_empty
+        else pre.deploy_contract(code=Op.STOP)
+    )
+
+    ret_size = 32 if memory_expansion else 0
+
+    call_code = Op.CALL(
+        gas=0, address=target, value=value, ret_size=ret_size, ret_offset=0
```
jochem-brouwer, line 451: Side-note: this only target out_size for the memory. We also have in_size, this should not change anything but this (😅 ) should also be parametrized. (So this allocates memory for the return value of the call. Also have to allocate memory for the in-value of the call. If memory does not yet "exist" it will fill it with zeros (and also init calldata like that) but we still have to pay for this "expansion")

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,159 +406,1216 @@ def test_bal_extcodesize_and_oog(
...
+
+    value_transferred = value > 0 and oog_boundary == OutOfGasBoundary.SUCCESS
+
+    post_state: Dict[Address, Account | None] = {alice: Account(nonce=1)}
+
+    if value_transferred:
+        post_state[target] = Account(balance=value)
+        post_state[caller] = Account(balance=0)
+    else:
+        post_state[caller] = Account(balance=value)
```
jochem-brouwer, line 530: Nit: in this case `value = 0`, so we can put `Account()`? (Note: it still has code)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
jochem-brouwer: Ah ok this missing test case might actually be the next test written 😅

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,159 +406,1216 @@ def test_bal_extcodesize_and_oog(
...
+    delegation_target = pre.deploy_contract(code=Op.STOP)
+    target = pre.fund_eoa(amount=0, delegation=delegation_target)
+
+    # memory expansion / no expansion
+    ret_size = 32 if memory_expansion else 0
+
+    # Use gas=1 to test that when we have enough for delegation_cost but not
+    # enough for delegation_cost + message_gas, we still don't read the
+    # delegation (the EVM does a static check upfront)
+    message_gas = 1
```
jochem-brouwer: This is incorrect. The GAS sent with the tx is that gas what ends up in the target contract. So this does not specify how much gas we can spend on, for instance, paying for 7702 delegations. Should be set to 0. (Or actually this value should not influence anything at all). 
What will influence the result is the gas send to this execution frame (so the tx gas limit)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,159 +406,1216 @@ def test_bal_extcodesize_and_oog(
...
+        gas_costs.G_WARM_ACCOUNT_ACCESS
+        if delegation_is_warm
+        else gas_costs.G_COLD_ACCOUNT_ACCESS
+    )
+
+    static_gas_cost = access_cost + transfer_cost + memory_cost
+    # The EVM's second check cost is static_gas + delegation_cost.
+    # When gas_left == extra_gas, calculate_message_call_gas caps the
+    # provided gas to 0, so message_gas doesn't contribute to the cost.
+    second_check_cost = static_gas_cost + delegation_cost
```
jochem-brouwer, line 745: Well this is actually interesting, because which check should go first? If we do a value transfer we have to check target account if it exists or not. We now check it, and in this case we see it is delegated. So we now have to also load the delegated account (to get the target code). This might yield interesting results, because what is checked first: if the account exists, or already following the path to load the delegated account?
I think this might be an extra edge case. So the question is what the order of checks here is. Likely: first check if target account is empty (it is not here). And then check the costs to load the delegation. I'm not sure if this could lead to divergences, because since the account is delegated we already know that it exists (so we are not paying the new account gas). Should sit down and figure out if there are interesting gas boundary values here to check 🤔

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
jochem-brouwer: Also, sending 0 gas still means we have to execute the code, even though 0 gas is available at the next execution frame. Why? 

If we run `Op.STOP` (or empty code) the call has succeeded (push 1 on stack to the execution frame invoking `Op.CALL`)

However - if we run anything else (or `Op.INVALID`) we run out of gas and we thus did not succeed and push 0 on stack of the execution frame invoking `Op.CALL`.

jochem-brouwer (commented):
I thoroughly checked the `Op.CALL` tests. The other tests seem similar (but are subject to cleanup so we don't `DRY` too much).

One thing we should add is parametrizing the memory, for both in and output. Both should be tested. I think insize [0, 32, 64] and outside [0,32,64] will catch all cases, both memory expansion, the case where out > in, out = in, and out < in.

I have to look over these tests again tomorrow 😄 👍 

Also apologies for some walls of text, I did a half brain dump on some possible extra edge cases here 😅

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -7,8 +7,11 @@
...
-| `test_bal_call_with_value_transfer` | Ensure BAL captures balance changes from `CALL` opcode with value transfer | Alice calls `Oracle` contract (200 wei balance) which uses `CALL` opcode to transfer 100 wei to Bob (0 wei balance). | BAL MUST include Alice (nonce changes), Oracle (balance change to 100 wei), and Bob (balance change to 100 wei). | ✅ Completed |
-| `test_bal_callcode_with_value_transfer` | Ensure BAL captures balance changes from `CALLCODE` opcode with value transfer | Alice calls `Oracle` contract (200 wei balance) which uses `CALLCODE` opcode to execute `TargetContract`'s code with 100 wei value transfer to Bob (0 wei balance). | BAL MUST include Alice (nonce changes), `Oracle` (balance change to 100 wei), Bob (balance change to 100 wei), and `TargetContract` (empty changes). | ✅ Completed |
+| `test_bal_call_success_no_delegation` | Ensure BAL captures successful CALL across all parameter combinations | Parametrized: target warm/cold, target empty/existing, value 0/1, memory expansion on/off. Uses precise gas calculations. | Target always in BAL. Balance changes recorded for caller/target when value > 0. | ✅ Completed |
```
raxhvl: The test descriptions needs update.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -375,159 +406,1216 @@ def test_bal_extcodesize_and_oog(
...
+
+    # Build access list for warming
+    access_list: list[AccessList] = []
+    if target_is_warm:
+        access_list.append(AccessList(address=target, storage_keys=[]))
+    if delegation_is_warm:
+        access_list.append(
+            AccessList(address=delegation_target, storage_keys=[])
+        )
+    access_list_or_none = access_list if access_list else None
```
raxhvl: `access_list_or_none` is not required, `access_list` will be empty if no address needs to be warmed and can be used directly to compute:

```py
 intrinsic_cost = fork.transaction_intrinsic_cost_calculator()(
        access_list=access_list
    )

```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: "second_check" can perhaps be better explained. This is the second `check_gas()` call in the specs. There is state access for the `target` in both the "delegation" check and the "is account alive" check. So far so good because we already checked for target access gas in the "first check gas" boundary.

- If the `target` is a delegation, we cannot have an empty account (`target` needs to have delegation designation `code`), so the "is account alive" check returns that there is `0` create cost always.
    - Then we have to see if we have enough gas to read the COLD or WARM delegation address for the `target`. This is where we do the second check here. At this point we know all the gas we will have to consume for a successful message call so we check ALL of that (along with the access for delegation). If there is not enough we should NOT have the delegation in the BAL - if there is, it's success case.
- If it's NOT a delegation, then we can either have create cost or not. None of these cases produce anything else that's affecting the BAL because we've already accessed the `target` at this point and there is nothing else to access.
    - If we don't have enough for create cost, there is no new boundary here as far as any new state access. No matter where we OOG here, it's the same case. 
    - If we do have the create cost, this is a successful scenario

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: Yes this gas doesn't influence anything other than I was experimenting passing through more gas or not (it used to be `0`) before the last changes. I think we can keep this at `0`, what do you think?

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: It's not always `0` here. `value_transferred = value > 0 and oog_boundary == OutOfGasBoundary.SUCCESS`. So (pseudocode):

```py
if success and value == 1 :  # we only parametrize `0` and `1`
        post_state[target] = Account(balance=value)
        post_state[caller] = Account(balance=0)
elif (
     OOG and value == 1,
     OOG and value == 0,
     success and value == 0,
):
        post_state[caller] = Account(balance=value)
        post_state[target] = (
            Account.NONEXISTENT
            if target_is_empty
            else Account(balance=0, code=Op.STOP)
        )
```

Would you want to track something else here?

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: Yes I think you're technically right. This will increase the number of tests quite a bit but we should parametrize this perhaps. What do you think? 😅

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: @jochem-brouwer can you summarize these here? I think we end up covering all boundaries across the tests. Is there anything else you still see that's missing? 👀

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: Back to `0` :)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
jochem-brouwer: Whoops, you are right!

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
jochem-brouwer: It is indeed the next test!

jochem-brouwer: (I cannot resolve this comment)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
jochem-brouwer: (I cannot resolve this comment)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: We decided to put this in a separate PR ^. I created an issue to track this [here](https://github.com/ethereum/execution-specs/issues/1910).

jochem-brouwer (approved):
LGTM. I might have to take a look at this one later, but for now this looks fine.

As next steps, I think we should address the duplication in a new PR to clean up these tests. We should also parametrize the memory.

Great work! 🚀 😄

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -57,6 +58,36 @@ class OutOfGasAt(Enum):
...
+    second check (delegation_cost). All 4 scenarios test
+    distinct boundaries.
+
+    Gas check order:
+    1. oog_before_target_access: access + transfer (if applicable) + memory.
+       OOG with not enough for this check - no state access.
+    2. oog_after_target_access: only enough for static check, state access
+       reads target into BAL, not enough for anything else.
+    3. oog_success_minus_1: exact gas minus 1. OOG here means target is in
+       BAL, but we have enough information to calculate delegation cost
```
jwasinger, line 75: if we calculated the delegation resolution cost component, that means the delegation target was read from the state and so it should be in the BAL in this case even if we ultimately go OOG.

fselmo: You don't need to reach into state to calculate this just as you don't need to reach into state to calculate the cold / warm charge of the target. You can check EVM accessed addresses and based on that you can charge warm / cold.

edit: I misunderstood. Yes, OOG with exact gas minus 1 has both in the BAL.

jwasinger: Actually, disregard this.  The delegation resolution cost only looks at the authority account's code.

PR #1885, opened 2025-12-10
feat: add depth-based worst-case attack benchmarks for execute mode

CPerezz:
This commit introduces a comprehensive worst-case attack benchmark suite that exploits Patricia Merkle Trie computational complexity through:

- Pre-mined CREATE2 addresses with configurable prefix sharing depth
- Deep storage slots with configurable trie depth
- Attack orchestrator contract for efficient batch attacks
- Execute mode support with pre-allocation pattern

Key changes:
- Add test_worst_case_attack_execute.py for execute mode testing
- Include AttackOrchestrator.sol for batch attack execution
- Add depth_9.sol and depth_10.sol contracts with deep storage
- Include pre-mined address data files (s9_acc3/4/5.json)
- Dynamic deep slot extraction based on storage_depth parameter
- Pre-allocation of all contracts and funds in pre-state
- Single-block attack transaction pattern for execute mode

The test deploys contracts at addresses with shared prefixes to maximize trie traversal overhead, then uses an orchestrator to efficiently attack deep storage slots, creating worst-case computational scenarios for benchmarking Ethereum client performance.


tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,505 @@
...
+    print(f"  Account depth: {account_depth}")
+    print(f"  Total auxiliary accounts: {NUM_CONTRACTS * account_depth}")
+
+    # Phase 1: Fund auxiliary accounts with 1 wei each to create trie depth
+    for i, contract_data in enumerate(contracts):
+        auxiliary_accounts = contract_data.get("auxiliary_accounts", [])
+        for aux_account in auxiliary_accounts:
+            aux_address = Address(aux_account)
+            # Fund with 1 wei to ensure the account exists in the trie
+            pre.fund_eoa(aux_address, value=1)
```
LouisTsai-Csie: I think this is why the current `execute remote` flow is failing. From my understanding of the implementation, here we intend to fund 1 wei to `aux_address`. However, `pre.fund_eoa()` does not work this way, it creates a new EOA instead of funding an existing account.

I experimented with manually injecting the account, but this only works in `fill` mode:
```python
pre[aux_address] = Account(address=aux_address, balance=1)
```

The correct solution might be to create funding transactions instead:
```python
funding_tx = Transaction(
    to=aux_address,
    value=1,  # 1 wei to create the account in the trie
    gas_limit=21_000,
    sender=deployer_eoa,
)
funding_txs.append(funding_tx)
```

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+        # Create deployment transaction to Nick's deployer
+        deploy_tx = Transaction(
+            to=NICK_DEPLOYER,
+            data=calldata,
+            gas_limit=deployment_gas_per_contract + 50000,  # Add buffer for CREATE2 operation
+            sender=deployer_address,
+            max_fee_per_gas=10_000_000_000,  # 10 gwei
+            max_priority_fee_per_gas=1_000_000_000,  # 1 gwei
+        )
+        deployment_txs.append(deploy_tx)
```
CPerezz, line 334: This sends 15k txs (as in execute mode even if we split them in blocks, they're all sent at the same time).
Thus, killink any RPC endpoint you have even locally.

For 1000 contracts this all works. But for 15k (the goal fo this tests) it just kills Anvil, Geth etc...

Unsure how to proceed..

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+        sstore_pattern = r'sstore\((0x[0-9a-fA-F]+),\s*1\)'
+        sstores = re.findall(sstore_pattern, content)
+        if sstores:
+            # The last sstore is the deepest slot
+            DEEP_SLOT = int(sstores[-1], 16)
+        else:
+            raise Exception(f"Could not find sstore operations in {sol_filename}")
+
+    print(f"\nDeep storage slot for depth {storage_depth}: {hex(DEEP_SLOT)}")
+    print(f"Expected: Changed from initial value of 1 to {attack_value} after attack")
```
CPerezz, line 513: I need to do all this crap such that I can get some values from inside the contract (cause I can't call RPCs directly through the tooling).

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+                storage={
+                    DEEP_SLOT: expected_value,  # Exact value from last attack
+                }
+            )
+
+    blockchain_test(
+        pre=pre,
+        blocks=blocks,
+        post=post,
+    )
```
CPerezz, line 539: This is (I assume) useless? As basically it just doesn't propagate blocks as I would expect it to. Neither it waits between blocks to not send all the txs at once.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+    json_path = Path(__file__).parent / json_filename
+
+    if not json_path.exists():
+        raise FileNotFoundError(f"Pre-mined data not found: {json_filename}")
+
+    with open(json_path, 'r') as f:
+        return json.load(f)
+
+
+def compile_attack_orchestrator():
```
CPerezz, line 73: We can get rid of this and just have the bytecode file.

CPerezz:
ping @danceratopz @LouisTsai-Csie 

Notice that I can simply take a turn and avoid deploying anything here. (Just the orchestator).

Thus, on this way, I can just send a single tx. And that would be the whole test. But, this would make this test only runnable on bloatnet. (Which is fine).

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
LouisTsai-Csie: We need some rate limit that wait until the first 1000 transactions are complete before proceeding the next 1000 transactions in `execute remote`. Thanks for sharing these number, i will try with 750 at first

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+        for aux_account in auxiliary_accounts:
+            aux_address = Address(aux_account)
+            # Create funding transaction with 1 wei to ensure the account exists in the trie
+            funding_tx = Transaction(
+                to=aux_address,
+                value=1,  # 1 wei to create the account in the trie
+                gas_limit=21_000,
+                sender=deployer_address,
+                max_fee_per_gas=10_000_000_000,  # 10 gwei
+                max_priority_fee_per_gas=1_000_000_000,  # 1 gwei
```
marioevz, line 260: This is automatically calculated by fill/execute, no need to add this. 
```suggestion
```

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+    init_code = Bytecode(init_code_bytes, popped_stack_items=0, pushed_stack_items=0)
+
+    # Calculate gas for deployment
+    # CREATE2 gas: 32000 base + 6 * init_code_size + execution
+    init_code_size = len(init_code)
+    # Constructor performs storage_depth SSTOREs to deep slots
+    # Each SSTORE to a new slot costs ~22100 gas (cold storage)
+    constructor_gas = storage_depth * 22100  # Dynamic based on storage depth
+    deployment_gas_per_contract = (
+        32000  # CREATE2 base cost
```
marioevz, line 280:
```suggestion
        gas_costs.G_TRANSACTION_CREATE  # CREATE2 base cost
```

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+
+    Args:
+        storage_depth: Depth of storage slots in the contract (e.g., 10)
+        account_depth: Depth of account address prefix sharing (e.g., 5)
+    """
+    # Load pre-mined CREATE2 data for the given depth parameters
+    try:
+        create2_data = load_create2_data(storage_depth, account_depth)
+    except FileNotFoundError as e:
+        pytest.skip(f"Skipping test: {e}")
```
marioevz, line 219: This should be a failure. If we modify parametrization and add a bunch of invalid tests for which json files do not exist we will not notice as they will only show up as skipped tests.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+        - contracts: List of dicts with 'salt' and 'auxiliary_accounts'
+    """
+    json_filename = f"s{storage_depth}_acc{account_depth}.json"
+    json_path = Path(__file__).parent / json_filename
+
+    if not json_path.exists():
+        raise FileNotFoundError(f"Pre-mined data not found: {json_filename}")
+
+    with open(json_path, 'r') as f:
+        return json.load(f)
```
marioevz, line 70: These files have a known format, we need to use Pydantic to load these into the proper types:
```python
class ContractInFile(base_model):
    salt: int
    contract_address: Address
    auxiliary_accounts: List[Address]

class ContractsFile(base_model):
    deployer: Address
    init_code_hash: Hash
    target_depth: int
    num_contracts: int
    total_time: float
    contracts: List[ContractInFile]
```
Or something like that, then you can do `ContractsFile.model_validate_json` to load the file's contents.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+
+    # Compile the AttackOrchestrator contract
+    print(f"  Compiling AttackOrchestrator.sol...")
+    orchestrator_bytecode = compile_attack_orchestrator()
+
+    # Phase 1: Create funding transactions for auxiliary accounts
+    # We need to use transactions, not pre-allocation, for execute mode
+    funding_txs = []
+    for i, contract_data in enumerate(contracts):
+        auxiliary_accounts = contract_data.get("auxiliary_accounts", [])
```
marioevz, line 250:
```suggestion
        auxiliary_accounts = contract_data.auxiliary_accounts
```
After we update the loaded files to use Pydantic.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+                max_priority_fee_per_gas=1_000_000_000,  # 1 gwei
+            )
+            funding_txs.append(funding_tx)
+
+    # Compile contract for the specified storage depth
+    print(f"  Compiling depth_{storage_depth}.sol...")
+    try:
+        init_code_bytes = compile_worst_case_contract(storage_depth)
+    except FileNotFoundError as e:
+        pytest.skip(f"Skipping test: {e}")
```
marioevz, line 269: Like the comment above, this should be a failure.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+
+    # Calculate how many contracts we can deploy per transaction
+    max_contracts_per_tx = MAX_GAS_PER_TX // deployment_gas_per_contract
+
+    print(f"  Init code size: {init_code_size} bytes")
+    print(f"  Deployment gas per contract: ~{deployment_gas_per_contract:,}")
+    print(f"  Max deployments per tx: {max_contracts_per_tx}")
+
+
+    from ethereum.crypto.hash import keccak256
```
marioevz, line 294: We never import inside of functions unless absolutely necessary. This specific import can be removed altogether though, see comment below.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+    deployment_txs = []  # Track deployment transactions
+
+    print(f"  Deploying {NUM_CONTRACTS} contracts via Nick's method...")
+    print(f"  Nick's deployer: {NICK_DEPLOYER}")
+
+    for i, contract_data in enumerate(contracts):
+        salt = contract_data["salt"]
+
+        # Convert salt to 32 bytes
+        salt_bytes = salt.to_bytes(32, 'big') if isinstance(salt, int) else bytes.fromhex(salt[2:] if salt.startswith("0x") else salt)
```
marioevz, line 307: This conversion can be skipped once we use pydantic.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+# Maximum gas per transaction (Fusaka EIP limit)
+MAX_GAS_PER_TX = 16_000_000
+
+NUM_CONTRACTS = 15000  # Full test with 15,000 contracts
+
+# Nick's deterministic deployer address (must be pre-deployed in execute mode)
+NICK_DEPLOYER = Address("0x4e59b44847b379578588920ca78fbf26c0b4956c")
+
+
+def load_create2_data(storage_depth, account_depth):
```
marioevz, line 51: This function should be a pytest fixture instead, which allows pytest to not have to reload the entire file each time for every test.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+        calldata = salt_bytes + bytes(init_code)
+
+        # Calculate CREATE2 address: keccak256(0xff ++ nick_deployer ++ salt ++ keccak256(init_code))
+        create2_input = (
+            bytes.fromhex("ff") +
+            bytes(NICK_DEPLOYER) +
+            salt_bytes +
+            keccak256(bytes(init_code))
+        )
+        contract_addr = keccak256(create2_input)[-20:]  # Last 20 bytes
```
marioevz, line 319: Use [`compute_create_address`](https://github.com/ethereum/execution-specs/blob/9e161b6e2f83fe41f3489d391a3584140184381a/packages/testing/src/execution_testing/test_types/helpers.py#L31-L57).

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+    )
+
+    # Calculate the orchestrator address based on nonce 0 (deployed FIRST)
+    import rlp
+    from eth_utils import keccak
+    # Nonce = 0 since orchestrator is deployed first
+    nonce_for_orchestrator = 0
+    deployer_bytes = bytes.fromhex(str(deployer_address)[2:])
+    orchestrator_address_bytes = keccak(rlp.encode([deployer_bytes, nonce_for_orchestrator]))[12:]
+    orchestrator_address = Address("0x" + orchestrator_address_bytes.hex())
```
marioevz, line 400:
```suggestion
    orchestrator_address = orchestrator_deploy_tx.created_contract
```

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
marioevz: Is this to counter the RPC crashing when sending 15k transactions at once? I think we need a proper fix in the `execute` command machinery to handle extreme cases like this (perhaps setting a max-txs-per-request and split [this line](https://github.com/ethereum/execution-specs/blob/9e161b6e2f83fe41f3489d391a3584140184381a/packages/testing/src/execution_testing/execution/transaction_post.py#L134) into multiple ones?)

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
marioevz: Can't we rather store metadata with the solidity source code? RPC calls will not work because filled tests, when consumed by the client test executors don't even spawn an RPC node.

marioevz (changes requested):
Left some comments with improvement suggestions. I'm still not completely sold on the need for compiling solidity, and that's because it's a pain to maintain and make the tests forward compatible.

tests/benchmark/stateful/bloatnet/depth_benchmarks/AttackOrchestrator.sol
```diff
@@ -0,0 +1,94 @@
...
+            // Loop through the range
+            for { let i := startIndex } lt(i, endIndex) { i := add(i, 1) } {
+                // Derive CREATE2 address
+                // Formula: keccak256(0xff ++ deployer ++ salt ++ initCodeHash)[12:]
+
+                // Store CREATE2 preimage in memory
+                // The layout should be: 0xff (1 byte) + deployer (20 bytes) + salt (32 bytes) + codeHash (32 bytes)
+                // Total: 85 bytes
+                let memPtr := 0x00
+                mstore8(memPtr, 0xff)                                    // 0xff prefix at byte 0
```
jochem-brouwer, line 53: There are constants stored in this for loop which do not need to be there. You can move these constant mstores before the loop (to init the constant parts of memory once)

tests/benchmark/stateful/bloatnet/depth_benchmarks/AttackOrchestrator.sol
```diff
@@ -0,0 +1,94 @@
...
+                mstore(add(memPtr, 0x01), shl(96, deployer))            // deployer address at bytes 1-20
+                mstore(add(memPtr, 0x15), i)                            // salt (i as uint256) at bytes 21-52
```
jochem-brouwer, line 55: For clarity I would not use hex here, but rather ints, so it is easier to check if these constants are correct.

tests/benchmark/stateful/bloatnet/depth_benchmarks/AttackOrchestrator.sol
```diff
@@ -0,0 +1,94 @@
...
+                mstore(add(memPtr, 0x35), codeHash)                     // init code hash at bytes 53-84
+
+                // Compute CREATE2 address
+                let target := and(
+                    keccak256(0x00, 0x55),            // Hash 85 bytes
+                    0xffffffffffffffffffffffffffffffffffffffff  // Mask to 20 bytes
```
jochem-brouwer, line 61: Masking is not necessary. If you `CALL` something then the EVM will mask it (so hide the topmost 12 bytes)

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
+"""
+abstract: BloatNet worst-case attack benchmark for maximum SSTORE stress (Execute Mode Version).
+
+This test implements a worst-case scenario for Ethereum block processing that exploits
+the computational complexity of Patricia Merkle Trie operations. It uses CREATE2 to deploy
+contracts at pre-mined addresses with shared prefixes, maximizing trie traversal depth.
```
jochem-brouwer, line 6: I think this should be somewhat more explicit: the deployed contracts will write to the storage there, but in such way that with a minimal set of key/values we get a certain trie depth (storage slots all share some prefix).

As a side-note, this is a good way to test the impact of "depth" but it does not test the impact of the actual contract (if it would actually have the depth if we would consider the MPT balanced - here it is completely the opposite and super unbalanced in one direction 😆 ). So if a client reads from a flat database (can directly load storage key without having to traverse the trie first), then if we read a storage key the more keys which "share" some prefix might have impact on the read speed (so the difference between the unbalanced/balanced tree on the state). This is obviously dependent on the database backend used and the db config.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+    BlockchainTestFiller,
+    Bytecode,
+    Fork,
+    Op,
+    Transaction,
+    Address,
+)
+
+# Attack function selector for WorstCaseERC20.attack(uint256)
+ATTACK_SELECTOR = 0x64dd891a  # attack(uint256) - verified with: cast sig "attack(uint256)"
```
jochem-brouwer, line 40: FYI, this is also `keccak256("attack(uint256)")[:4]` 😄 👍

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+
+# Maximum gas per transaction (Fusaka EIP limit)
+MAX_GAS_PER_TX = 16_000_000
```
jochem-brouwer, line 43:
```suggestion
MAX_GAS_PER_TX = 16_777_216
```

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+    Fork,
+    Op,
+    Transaction,
+    Address,
+)
+
+# Attack function selector for WorstCaseERC20.attack(uint256)
+ATTACK_SELECTOR = 0x64dd891a  # attack(uint256) - verified with: cast sig "attack(uint256)"
+
+# Maximum gas per transaction (Fusaka EIP limit)
```
jochem-brouwer, line 42:
```suggestion
# Maximum gas per transaction (EIP-7825 limit)
```

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+MAX_GAS_PER_TX = 16_000_000
+
+NUM_CONTRACTS = 15000  # Full test with 15,000 contracts
+
+# Nick's deterministic deployer address (must be pre-deployed in execute mode)
+NICK_DEPLOYER = Address("0x4e59b44847b379578588920ca78fbf26c0b4956c")
```
jochem-brouwer, line 48: FYI, this is not Nick's deployer, it is the deterministic deployment: https://github.com/Arachnid/deterministic-deployment-proxy/blob/master/source/deterministic-deployment-proxy.yul

That one was deployed using Nick's method - it is creating a transaction without a private key. This ensures that the transaction which we "send" cannot be overwritten (because no one has the key) and we can thus deterministically deploy a contract on most non-mainnet chains (and of course mainnet). Nick's method: https://medium.com/patronum-labs/nicks-method-ethereum-keyless-execution-168a6659479c

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+    """
+    sol_path = Path(__file__).parent / "AttackOrchestrator.sol"
+
+    if not sol_path.exists():
+        raise FileNotFoundError(f"AttackOrchestrator.sol not found at {sol_path}")
+
+    # Compile with optimization and no metadata for reproducibility
+    result = run(
+        [
+            "solc",
```
jochem-brouwer, line 88: To fully reproduce would also need to pin a solc version, as different versions can compile to different results.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+        + 200 * init_code_size  # Code deposit cost (200 gas per byte)
+        + constructor_gas  # Constructor execution (storage_depth SSTOREs to deep slots)
+    )
+
+    # Calculate how many contracts we can deploy per transaction
+    max_contracts_per_tx = MAX_GAS_PER_TX // deployment_gas_per_contract
+
+    print(f"  Init code size: {init_code_size} bytes")
+    print(f"  Deployment gas per contract: ~{deployment_gas_per_contract:,}")
+    print(f"  Max deployments per tx: {max_contracts_per_tx}")
```
jochem-brouwer, line 291: We only deploy one contract per tx, so this does not do anything (yet)?

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
jochem-brouwer, line 319: Isn't this part of the json already? `.contract_address` field? Or was that deployed on a different address (so not the `0x4e59b44847b379578588920ca78fbf26c0b4956c`)?

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
jochem-brouwer: Can also deploy more contracts in a single tx by creating a contract which calls multiple times in `0x4e59b44847b379578588920ca78fbf26c0b4956c` (CREATE2 factory). As calldata we give the starting salt, and for each call into CREATE2 it increases the salt. Alternatively can also not put that in calldata and just store the current salt in storage. Only call into the CREATE2 factory if there is enough gas left. Depending on the size of the contract which we are deploying (this one does not look super big) we will get at least 3x less txs but possibly even more.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+            gas_limit=min(MAX_GAS_PER_TX, (batch_end - batch_start) * GAS_PER_ATTACK + 21_200),
+            sender=deployer_address,
+            data=calldata,
+            max_fee_per_gas=10_000_000_000,  # 10 gwei
+            max_priority_fee_per_gas=1_000_000_000,  # 1 gwei
+        )
+        attack_txs.append(attack_tx)
+
+    # Create blocks with all transactions in proper order
+    # IMPORTANT: Deploy orchestrator FIRST (nonce 0) to have deterministic address
```
jochem-brouwer, line 435: Can deploy using Nick's method to be sure the deterministic address is equal 😉

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
jochem-brouwer: The depth contracts look alike: the difference is the storage written to in the constructor and the target storage slot. We can encode these as constructor arguments (so there is then also no need for `depth_9.sol` and `depth_10.sol`, the storage keys to write to and attack are then in a metadata list, which are thus the values provided currently in those files)

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_worst_case_attack_execute.py
```diff
@@ -0,0 +1,539 @@
...
+        if contract_address in last_written_values:
+            # Contract was attacked - verify the last written value
+            expected_value = last_written_values[contract_address]
+            if i < 3:
+                print(f"    Expected storage[{hex(DEEP_SLOT)}] = {expected_value}")
+            # Verify the attack worked - storage should be modified
+            post[contract_address] = Account(
+                # Contract should exist with code
+                storage={
+                    DEEP_SLOT: expected_value,  # Exact value from last attack
```
jochem-brouwer, line 531: Shouldn't this be `attack_value`?

jochem-brouwer (commented):
Some comments.

Regarding the need to compile solidity, for these kind of setups it certainly lowers the time spent on creating these contracts, because without solidity we have to write everything by hand. That would however be super optimized code (in most cases). For this one, ERC20 contracts with some extra logic have to be created, and it would be rather impossible to create such contract by hand without a compiler.

However, I'm not sure why the code here needs to be an ERC20 contract. If the attack here is a trie-depth based attack then we could also do with minimal code, right? This would write the storage in the constructor and at runtime it would run the `attack` (writing the value to the target storage slot).

These mined storage keys/account keys, how did you retrieve those? 😄 👍

tests/benchmark/stateful/bloatnet/depth_benchmarks/depth_10.sol
```diff
@@ -0,0 +1,79 @@
...
+    uint8 public constant decimals = 18;
+
+    constructor() {
+        // Mint total supply to deployer
+        totalSupply = 1_000_000_000 * 10 ** 18; // 1 billion tokens
+        balanceOf[msg.sender] = totalSupply;
+
+        // Set all mined storage slots to 1
+        assembly {
+            sstore(0xfeb1bc66963690bd7d902e86ccaf4e0fa1ea72277653d012a3fed288892770fc, 1)
```
jsign, line 22: Does the address miner also put as a constraint that the preimages have this "nibble by nibble" overlap for the required depth?

Mostly asking because what matters for this to be deep branches is that the hashes of the storage slot numbers satisfy this, not the storage numbers themselves (i.e. the storage slots are hashed to define which is the storage trie address).

(Maybe I'm missing something else from the whole PR setup!)

Edit: or maybe the code that generates this contract is putting the hashes instead of the preimages?

LouisTsai-Csie:
Close as PR #1976 done

PR #1894, opened 2025-12-10
fix(spec-tests): fix title->module conversion for bpo forks

SamWilsn:
#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/utils.py
+++ b/src/ethereum_spec_tools/evm_tools/utils.py
@@ -128,6 +128,8 @@ def find_fork(
                 current_fork_module = fork
                 current_fork_block = fork_block
 
+    current_fork_module = re.sub("^b_p_o", "bpo", current_fork_module)
+
     for fork in forks:
         if current_fork_module == fork.short_name:
             return fork, current_fork_block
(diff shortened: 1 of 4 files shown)


gurukamath (approved):

src/ethereum_spec_tools/evm_tools/utils.py
```diff
@@ -128,6 +128,10 @@ def find_fork(
                 current_fork_module = fork
                 current_fork_block = fork_block
 
+    stripped = current_fork_module.removeprefix("b_p_o")
+    if stripped != current_fork_module:
```
fselmo: Just a thought that might make this easier / quicker to understand for a reader, what if we did something like:

```py
current_fork_module = current_fork_module.replace("b_p_o", "bpo")
```

To replace any sort of `b_p_o` with `bpo`? Or, if we want to be more targeted:

```py
if current_fork_module.startswith("b_p_o"):
    # only at the start and replaces only the first occurrence
    current_fork_module = current_fork_module.replace("b_p_o", "bpo", 1)
```

fselmo (approved):
lgtm, just a tiny nit for readability.

PR #1897, opened 2025-12-11
refactor(spec-specs): Refactor specs to be more coherent wrt gas acco…

fselmo:
## 🗒️ Description

I went back through and refactored the specs to be more succinct wrt gas accounting, in an effort to move these changes out of this branch and into its own PR to be applied across all forks.

- Check all static gas we can before state accesses
- Early exit on static access checks

I have similar changes without the BAL tracking in its own branch and I think this will make it easier to agree on here, refactor there, and then eventually to rebase these changes back into here once we agree on a good design.

A good idea for reviewers here is to diff against `forks/amsterdam` instead of `eips/amsterdam-eip-7928` so we can see the minimal set of changes required to the underlying base spec.

---
Note: This is still a WIP as I'm going to look back over and make sure I didn't miss anything but putting it here early so we can have eyes on it.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


src/ethereum/forks/amsterdam/vm/instructions/storage.py
```diff
@@ -55,20 +55,16 @@ def sload(evm: Evm) -> None:
...
 
     # GAS
-    is_cold_access = (
-        evm.message.current_target,
-        key,
-    ) not in evm.accessed_storage_keys
-    gas_cost = GAS_COLD_SLOAD if is_cold_access else GAS_WARM_ACCESS
-
-    charge_gas(evm, gas_cost)
+    if (evm.message.current_target, key) in evm.accessed_storage_keys:
```
fselmo, line 58: This actually brings the changes back to what it used to be (no changes once PR'd to `forks/amsterdam`) as there is no need to alter the logic here.

src/ethereum/forks/amsterdam/vm/instructions/storage.py
```diff
@@ -91,35 +87,28 @@ def sstore(evm: Evm) -> None:
...
-        gas_cost += GAS_COLD_SLOAD
-
     state = evm.message.block_env.state
     original_value = get_storage_original(
         state, evm.message.current_target, key
     )
     current_value = get_storage(state, evm.message.current_target, key)
 
-    if is_cold_access:
+    gas_cost = Uint(0)
```
fselmo, line 106: This also returns the logic to what it is, no need to change this unless we want the `is_cold_access` check on all operations. I think we can decide this on the isolated PR as well. I'd rather not alter too much of the underlying code in our changes so we can more easily rebase up to eip-7928.

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -417,69 +417,56 @@ def call(evm: Evm) -> None:
...
-            access_gas_cost + create_gas_cost + transfer_gas_cost,
-        )
-        check_gas(evm, message_call_gas.cost + extend_memory.cost)
-        code = get_account(evm.message.block_env.state, to).code
-        final_address = to
-
-    code_address = final_address
-    disable_precompiles = is_delegated
-
+    if is_delegated:
```
fselmo, line 449: Much more succinct. All opcodes here should have a similar flow.

src/ethereum/forks/amsterdam/vm/eoa_delegation.py
```diff
@@ -155,51 +149,16 @@ def calculate_delegation_cost(
...
-    # Calculate gas cost for delegation target access
     if delegated_address in evm.accessed_addresses:
         delegation_gas_cost = GAS_WARM_ACCESS
     else:
         delegation_gas_cost = GAS_COLD_ACCOUNT_ACCESS
 
-    return True, address, delegated_address, delegation_gas_cost
-
-
-def read_delegation_target(evm: Evm, delegated_address: Address) -> Bytes:
```
fselmo, line 171: This is unnecessary. By the time we get code for the `code_address`, we already have either the original address or the delegated address as the `code_address`, so we can simply call `get_account(state, code_address).code`, essentially.

fselmo:
I'm going to merge this as it's in a much cleaner state than the current specs and actually fixes a bug for CALL in static context where we should do this static check as early as we can, once `value` is popped from the stack, and before any state access.

I added a test here for this as well. We should put up a subsequent PR if we'd like to make any amendments to this design before we factor out these changes into its own changeset to PR against `forks/amsterdam`.

We should make sure to reference these changes when reviewing a subsequent PR.

Issue #1865, opened 2025-12-08
State redesign tracking issue


fselmo:
There is broad agreement that the spec needs a redesigned state representation that allows us to efficiently answer questions about the state at any point in time. Multiple ongoing efforts (block-level access lists, statelessness, zkEVM) would benefit from a unified approach.

The current block-level access list implementation in `eips/amsterdam/eip-7928` has created its own `state_tracker` implementation that works well but can benefit from this "bigger picture" redesign. To avoid divergence and move toward a shared design, the next steps we identified are:

- [x] Reach consensus on a state re-design that answers questions about state - I'm linking #1209 here, which essentially attempts to answer this same question

- [x] Implement this state re-design in its own PR targeting the active development fork
(likely `forks/amsterdam`).

- [x] Refactor the `eips/amsterdam/eip-7928` implementation to use this new unified state design to build block-level access lists.

cc: @petertdavies, @gurukamath, @nerolation, @kevaundray

petertdavies:
I think the first step is implementing the main state tracking types. There are then lots of different features that should switch to state tracking (state accesses, BALs, access lists), but that can done in incremental and independent PRs.

Here is my sketch of the state tracking types:


```python
@dataclass
class TxStateTracking:
   parent: BlockStateTracking
   
   account_reads: Set[Address]
   storage_reads: Set[Tuple[Address, Bytes32]]
   
   account_writes: Dict[Address, Account]
   storage_writes: Dict[Address, Dict[Bytes32, U256]]

   created_accounts: Set[Address]
   
   
@dataclass
class BlockStateTracking:
   parent: StateOracle
   
   account_reads: Set[Address]
   storage_reads: Set[Tuple[Address, Bytes32]]
   
   account_writes: Dict[Address, List[Tuple[BlockAccessIndex, Account]]]
   storage_writes: Dict[Address, Dict[Bytes32, Tuple[BlockAccessIndex, Account]]]
```


* The version in the current BAL branch has a bunch of `pre_*` fields. I think these can be omitted, since they are effectively just caching queries to `parent`.
* I have gone with `account_write` rather than separating out the individual members of `Account`, because the current EELS code treats the 3 fields as a single datatype (we use `get_account()`). It should be easy to generate the BAL from this.
* I think having separate `BlockStateTracking` and `TxStateTracking` types is more elegant, but it may lead to problematic code duplication (there are so many helpers in `state.py`). I'm not sure what the best approach here is.

[...]

PR #1899, opened 2025-12-11
refactor(eip7928): refactor net zero filtering in BALs

gurukamath:
## 🗒️ Description
This PR refactors the net zero refactoring mechanism in BALs and tightens the code up by introducing the following changes

1. Make assertions such that capturing pre-state only happens in tx level frame or higher.
2. Use all post-values from the relevant `*_changes` attributes from the frame (instead of looking into the state)
3. Remove the `capture_pre_nonce` function since this is not used anywhere
4. Standardize the filtering code across balance, code and storage in the `filter_net_zero_frame_changes` function.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/src/ethereum/forks/amsterdam/state_tracker.py
+++ b/src/ethereum/forks/amsterdam/state_tracker.py
@@ -314,7 +314,7 @@ def track_code_change(
 
 
 def track_selfdestruct(
-    state_changes: StateChanges,
+    tx_frame: StateChanges,
     address: Address,
 ) -> None:
     """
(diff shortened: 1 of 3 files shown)


src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -521,44 +499,50 @@ def filter_net_zero_frame_changes(
     ----------
     tx_frame :
```
gurukamath, line 512: Since this can also be called on a block level frame (while processing withdrawals), we might need to rethink the `tx_frame` variable name and other related details.

fselmo (dismissed):
Nice cleanups. This lgtm!

I'm a bit hesitant spending too much more time on any more state changes here before the [larger state refactoring](https://github.com/ethereum/execution-specs/issues/1865#issuecomment-3642660766) changes are implemented since we will likely use that instead. What do you think is a good line to draw for us as far as cleaning things up here and waiting for the more general state refactor?

This is definitely an improvement though, let's get this in 👍🏼

gurukamath:
> What do you think is a good line to draw for us as far as cleaning things up here and waiting for the more general state refactor?

I don't think we should be actively looking to make any design changes right now. This particular change just crossed my mind as I was looking at the specs in order to write a test for a different scenario. And figured it was a quick win that tightens our implementation. (possibly eliminating any subtle bugs)

I wouldn't mind having minor adjustments to the spec implementation from time to time (especially to fix logical inconsistencies) but agree with you that most of our focus right now should be test implementation and releases.

fselmo:
> I wouldn't mind having minor adjustments to the spec implementation from time to time (especially to fix logical inconsistencies) but agree with you that most of our focus right now should be test implementation and releases.

Yeah, agreed 👍🏼. sgtm 😃

fselmo:
@gurukamath I found some failing tests with this refactor, specifically `test_selfdestruct_created_in_same_tx_with_revert`:

```
FAILED tests/cancun/eip6780_selfdestruct/test_selfdestruct_revert.py::test_selfdestruct_created_in_same_tx_with_revert[fork_Amsterdam-blockchain_test_from_state_test-outer_selfdestruct_before_inner_call-same_tx] - execution_testing.test_types.block_access_list.exceptions.BlockAccessListValidationError: Expected balance_changes to be empty but found [BalBalanceChange(rlp_override=None, block_access_index=1, post_balance=1)]
```

I haven't had a chance to dig into the issue yet but we should address this.

gurukamath:
> I haven't had a chance to dig into the issue yet but we should address this.

The issue here was that self-destruct was not being handled properly. We need to delete the balance changes (just like we do with nonce and code changes) for self-destructed accounts. I have updated the PR to make necessary changes.

fselmo:
> The issue here was that self-destruct was not being handled properly. We need to delete the balance changes (just like we do with nonce and code changes) for self-destructed accounts. I have updated the PR to make necessary changes.

It looks like now `test_bal_self_destruct` is failing. I also have a PR with a lot more boundary cases for selfdestruct. I expect some would fail there as well. PR #1954 also includes a fix for selfdestruct that I introduced in my [refactor PR](https://github.com/ethereum/execution-specs/pull/1897) (I moved the `track_address` into only the COLD path by accident... fixed [here](https://github.com/ethereum/execution-specs/pull/1954/changes#diff-28a1b575fd7c3d82832c0826cf58a881101643543d35c123c78ca65202152c23L643-R643)), but it looks like this failure is unrelated to that - even if I apply the one line fix here it will still fail `test_bal_self_destruct`.

gurukamath:
> It looks like now `test_bal_self_destruct` is failing. I also have a PR with a lot more boundary cases for selfdestruct. I expect some would fail there as well. PR #1954 also includes a fix for selfdestruct that I introduced in my [refactor PR](https://github.com/ethereum/execution-specs/pull/1897) (I moved the `track_address` into only the COLD path by accident... fixed [here](https://github.com/ethereum/execution-specs/pull/1954/changes#diff-28a1b575fd7c3d82832c0826cf58a881101643543d35c123c78ca65202152c23L643-R643)), but it looks like this failure is unrelated to that - even if I apply the one line fix here it will still fail `test_bal_self_destruct`.

Ah!! The curious case of the pre-funded account, created and self-destructed in the same tx. Don't know how I missed that. Have updated the PR.

Also, I ran this re-factor on top of what you have in #1954 and all the tests there are passing.

fselmo (approved):
Much cleaner, ty! 🚀

PR #1901, opened 2025-12-11
feat(test-benchmark): add labels to benchmark pre-alloc groups and fix extract config

spencer-tb:
### Main Fix
This PR fixes the `extract_config` bug where fork was accessed from genesis instead of `PreAllocGroup`:
https://github.com/ethereum/execution-specs/actions/runs/20071991321/job/57576383289

### Extra Changes
Adds automatic labeling to benchmark test pre-allocation groups to make it easy to identify the main group vs separated test groups. This consideration comes from the 6 pre-allocations groups now contained in benchmark releases, due to the seperate pre-alloc groups in `test_blockhash`:
```python
@pytest.mark.repricing
@pytest.mark.parametrize(
    "index,chain_length",
    [
        pytest.param(0, 256, id="genesis"),
        pytest.param(1, 256, id="block_1"),
        pytest.param(256, 256, id="block_256"),
        pytest.param(257, 256, id="current_block"),
        pytest.param(None, 256, id="random"),
    ],
)
@pytest.mark.slow("Generates long chain")
@pytest.mark.pre_alloc_group("separate", reason="Generates long chain")
def test_blockhash(
...
```

When `fill --generate-all-formats` is run for benchmark tests, pre-alloc groups are now:
  - Labeled in JSON content with `"label": "main"` or `"label": "single_test_<name>"`
  - Named with the label in filenames: `0x398fa63f55cfbafc_main.json` instead of just `0x398fa63f55cfbafc.json`

This makes it trivial to find the main benchmark pre-alloc group (593 tests) versus the 5 separated `test_blockhash` variants (1 test each).

Before:
```tree
  fixtures/blockchain_tests_engine_x/pre_alloc/
  ├── 0x398fa63f55cfbafc.json  (which one is main?)
  ├── 0x2e3463cfd832e146.json
  ├── ...
```
After:
```tree
  fixtures/blockchain_tests_engine_x/pre_alloc/
  ├── 0x398fa63f55cfbafc_main.json  ← Easy to identify!
  ├── 0x2e3463cfd832e146_single_test_test_blockhash.json
  ├── ...
```

## 🔗 Related Issues or PRs

#1799

## ✅ Checklist

    ```console
    uvx tox -e static
    ```


packages/testing/src/execution_testing/cli/extract_config.py
```diff
@@ -121,11 +121,12 @@ def extract_client_files(
 
 def create_genesis_from_fixture(
     fixture_path: Path,
-) -> Tuple[FixtureHeader, Alloc, int]:
+) -> Tuple[FixtureHeader, Alloc, int, Fork]:
     """Create a client genesis state from a fixture file."""
     genesis: FixtureHeader
     alloc: Alloc
     chain_id: int = 1
+    fork: Fork
```
spencer-tb, line 129: `extract_config` fix.

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
```diff
@@ -66,6 +70,21 @@ def add_test_alloc(self, test_id: str, new_pre: Alloc) -> None:
 
     def build(self) -> "PreAllocGroup":
         """Build the pre-alloc group."""
+        # Generate a label based on test count and test IDs
+        # Only label benchmark tests
+        label = None
+        if self.test_ids and "tests/benchmark/" in self.test_ids[0]:
```
marioevz, line 76: This seems too brittle, we could collect a test object that contains more metadata instead of trying to infer information that we already lost.

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
```diff
@@ -66,6 +70,21 @@ def add_test_alloc(self, test_id: str, new_pre: Alloc) -> None:
...
+            if len(self.test_ids) == 1:
+                # Single test - extract test name for label
+                test_id = self.test_ids[0]
+                # Format: "tests/benchmark/.../test_file.py::test_name[params]"
+                if "::" in test_id:
+                    test_name = test_id.split("::")[-1].split("[")[0]
+                    label = f"single_test_{test_name}"
+            elif len(self.test_ids) > 100:
```
marioevz, line 84: This hard-coded number heuristic is prone to error, as the number of tests grow, we will likely hit this at some point when we have more grouped tests.

Perhaps main should be equal to all tests that do not modify the `env`?

Also, if we plan to make benchmark fixtures for other forks, we are going to have a conflict, so perhaps we should include the name of the fork in the label too.

marioevz (changes requested):
I feel like the label should have not been included in the same PR as this fix because it needs more thinking it through.

Plus it is technically a breaking change, because benchmark test runners will have to be updated to be prepared for these file name changes.

marioevz:
We can remove the fix from this PR and merge #1947, and only leave this PR to discuss the labeling.

PR #1903, opened 2025-12-11
feat(tests): Add more EIP-7928 test cases

jochem-brouwer:
## 🗒️ Description
This PR adds two test cases to be added for BALs. Feel free to merge as-is or keep this open until these cases are added.

---

@fselmo updating:

Since we had coverage for both of these cases, I attempted to describe the missing coverage a little better here. I then ended up implementing the tests so this no longer updated the `test_cases.md` only, it also adds the missing coverage cases. PR title updated to reflect this.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


fselmo:
@jochem-brouwer can you put the target branch here to be `eips/amsterdam/eip-7928`?

marioevz:
Hey @jochem-brouwer is this PR still required or could we close it? Thanks.

jochem-brouwer:
@fselmo I believe these test cases are not yet written, so they should be added. If that's the case then let's merge this PR :smile:

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -134,4 +134,5 @@
...
-
+| `test_bal_static_mode` | Verify correct BAL in `static` mode | Invoke code `C` in `static` mode via `STATICCALL`. The code `C` should execute (with target `T` - brackets denote if `T` should be in BAL or not): `CALL` with nonzero value (no), `CALL` with zero value (yes), `CALLCODE` with (non)zero value (yes), `CREATE[2]` with (non)zero value (no), `SELFDESTRUCT` (no). Note: in cases with `value` being sent, also ensure the case with insufficient balance and sufficient balance in `C`. | Cases denote in brackets if target `T` must be in BAL or not. | 🟡 Planned |
+| `test_bal_call_just_enough_gas` | Tests if target is in BAL if there is 0 gas left for execution | Code C which invokes `CALL`/`CALLCODE`/`STATICCALL`/`DELEGATECALL`/`CREATE`/`CREATE2` with exactly enough gas to invoke the code at target `T` (there is thus 0 gas available in this newly created call frame). | `T` is always in BAL. | 🟡 Planned |
```
fselmo: I think these are extensively tested via [these tests](https://github.com/fselmo/execution-specs/blob/3b8575bbd79201a850f986f0d4f02eb18ecc328f/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py#L405-L405) where the `OOG` cases only subtract `1` from each parametrized test case so we can sanity check that we indeed have exact gas for the `SUCCESS` cases. We have the same for `DELEGATECALL`, `CALLCODE`, and `STATICCALL`. We check relevant 7702 paths where appropriate too. We checked when that PR was merged very diligently and we have test cases for every code path in these opcodes unless some major refactor happened recently there. Is this test case any different from these?

I have not checked, however, the `CREATE` and `CREATE2` cases but either way this needs to be updated to only reflect those if that's the only thing missing. We should check these.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -134,4 +134,5 @@
...
-
+| `test_bal_static_mode` | Verify correct BAL in `static` mode | Invoke code `C` in `static` mode via `STATICCALL`. The code `C` should execute (with target `T` - brackets denote if `T` should be in BAL or not): `CALL` with nonzero value (no), `CALL` with zero value (yes), `CALLCODE` with (non)zero value (yes), `CREATE[2]` with (non)zero value (no), `SELFDESTRUCT` (no). Note: in cases with `value` being sent, also ensure the case with insufficient balance and sufficient balance in `C`. | Cases denote in brackets if target `T` must be in BAL or not. | 🟡 Planned |
```
fselmo: Yes I think this also may be missing some coverage but we need to update this to be a bit more specific for missing cases. We have a test for `CALL` for example [here](https://github.com/fselmo/execution-specs/blob/3b8575bbd79201a850f986f0d4f02eb18ecc328f/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py#L1966-L1966). Let's see what's actually missing in the coverage and update the description here appropriately. Again, I can check more thoroughly in another pass, just updating on what I know is already tested :)

fselmo:
@jochem-brouwer, I took a look at the missing coverage and I believe my commit on top of this now bridges the gaps we had here. We had some older tests that captured some of these without BAL expectations but these are good to reference and have updated for Amsterdam with expectations.

@raxhvl can you maybe sanity check the tests here with a review? Don't want to merge my own test implementations and we can use another pass at checking full coverage on the intended cases here.

I will refrain from approving though I do approve 😅... so as not to accidentally merge before we get more eyes on these :)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -2549,6 +2838,181 @@ def test_bal_transient_storage_not_tracked(
     )
 
 
+@pytest.mark.parametrize(
+    "oog_boundary",
+    [
+        OutOfGasBoundary.OOG_BEFORE_TARGET_ACCESS,
+        OutOfGasBoundary.OOG_AFTER_TARGET_ACCESS,
```
fselmo, line 2779: This case is a nice-to-have for test readability since we can subtract just `1` and see that the target is not in the BAL and have some sanity check that the gas is exact for static check. AT this boundary we _read_ the account so target is in BAL but we OOG. Subtract one and the target is not in BAL because it fails our static checks.

^ This is extensively covered in the CALL tests but this boundary is useful for the readability aspect and sanity check I mentioned above.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1888,35 +1889,31 @@ def test_bal_create_oog_code_deposit(
...
-            address=contract_b,
-            args_offset=0,
-            args_size=0,
-            ret_offset=0,
-            ret_size=0,
-        )
-        + Op.POP  # pop the return value (0 = failure)
-        + Op.SSTORE(0, 1)  # this should succeed (non-static context)
+        code=Op.STATICCALL(gas=1_000_000, address=contract_b)
+        + Op.POP
```
raxhvl: issue: Store staticcall result

 The result of `STATICCALL` is central to the test and should not be popped away.  The current tests fails silently if `STATICCALL` fails for whatever reason.  Instead:

```
SSTORE(0, Op.STATICCALL(gas=1_000_000, address=contract_b))
```

Post-state` (storage={0: 1})` will assert  `STATICCALL` went through or test fails

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1888,35 +1889,31 @@ def test_bal_create_oog_code_deposit(
...
     pre: Alloc,
     blockchain_test: BlockchainTestFiller,
+    original_value: int,
 ) -> None:
     """
-    Ensure BAL does not record storage reads when SSTORE fails in static
-    context.
+    SSTORE in static context must not leak storage reads into BAL.
 
-    Contract A makes STATICCALL to Contract B. Contract B attempts SSTORE,
```
raxhvl, line 1899: nit: Consistent docstring

The dostring mentions contract B and not contract A, it feels incomplete. Keep "    Contract A makes STATICCALL to Contract B. Contract B attempts SSTORE,"

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1888,35 +1889,31 @@ def test_bal_create_oog_code_deposit(
     )
 
 
+@pytest.mark.parametrize(
+    "original_value", [0, 0x42], ids=["zero_original", "nonzero_original"]
```
raxhvl, line 1893: question: Is the parametrization redundant? `STATICCALL` fails before storage.

I would argue the original value of storage is irrelevant for the test; because storage and its value are on the state side of things and `STATICCALL` should fail before we get there. So regardless of what is in storage, we should not see a read in BAL.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1951,48 +1948,272 @@ def test_bal_sstore_static_context(
...
+        target = pre.fund_eoa(amount=target_starting_balance)
 
     caller_starting_balance = 10**18
     caller = pre.deploy_contract(
-        code=Op.CALL(gas=100_000, address=target, value=1) + Op.STOP,
+        code=Op.CALL(
+            gas=100_000,
+            address=target,
+            value=1,
+            address_warm=target_is_warm,
```
raxhvl: suggestion: metadata for `CALL` is unnecessary

This test is not concerned with gas calculation so we can skip the `address_warm` metadata for simplicity.
```suggestion
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1951,48 +1948,272 @@ def test_bal_sstore_static_context(
...
             contract_a: Account(storage={0: 1}),
-            contract_b: Account(storage={0: 0}),  # SSTORE failed
+            contract_b: Account(
+                storage={0: original_value} if original_value else {}
+            ),
         },
     )
 
 
+def build_static_context_test(
```
raxhvl: suggestion: Simplify the helper API

Every consumer does the same thing:
```py
 block, post = build_static_context_test(...)
 blockchain_test(pre=pre, blocks=[block], post=post)
```

The sole purpose of the return value is to run `blockchain_test`. If the helper took `blockchain_test` as a parameter, it could just call it internally and return nothing.

The param name `caller` is confusing. It is rather the target for `staticcall`, consider renaming. I recommend a `bal_expection` instead of `expected_not_in_bal` for full control over expectations.  So the API becomes:


```py
def blockchain_test_under_static_call(     
      pre: Alloc,    
      *,   
      static_call_target: Address,       # contract executed inside STATICCALL                    
      bal_expectations: Dict[Address, BalAccountExpectation | None],  # BAL entries beyond alice/static_caller           
      post: Dict[Address, Account | None] | None = None,  # post-state beyond static_caller     
      tx_access_list: list[AccessList] | None = None,  # warm addresses before execution   
  ) -> None:      
             
```

An example test with this API:

```py
@pytest.mark.parametrize(
    "target_is_warm", [False, True], ids=["cold_target", "warm_target"]          
)                                                                                
@pytest.mark.parametrize(                                                        
    "target_has_code", [False, True], ids=["eoa_target", "contract_target"]      
)                                                                                
def test_bal_call_with_value_in_static_context(                                  
    pre: Alloc,                                                                  
    target_is_warm: bool,                                                        
    target_has_code: bool,                                                       
) -> None:                                                                       
    target_starting_balance = 1022                                               
    if target_has_code:                                                          
        target = pre.deploy_contract(                                            
            code=Op.STOP, balance=target_starting_balance                        
        )                                                                        
    else:                                                                        
        target = pre.fund_eoa(amount=target_starting_balance)                    
                                                                                 
    caller_starting_balance = 10**18                                             
    caller = pre.deploy_contract(                                                
        code=Op.CALL(gas=100_000, address=target, value=1) + Op.STOP,            
        balance=caller_starting_balance,                                         
    )                                                                            
                                                                                 
    blockchain_test_under_static_call(                                           
        pre,                                                                     
        blockchain_test,                                                         
        static_call_target=caller,                                               
        bal_expectations={                                                       
            caller: BalAccountExpectation.empty(),                               
            target: None,                                                        
        },                                                                       
        post={                                                                   
            caller: Account(balance=caller_starting_balance),                    
            target: Account(balance=target_starting_balance),                    
        },                                                                       
        tx_access_list=(                                                         
            [AccessList(address=target, storage_keys=[])]                        
            if target_is_warm                                                    
            else None                                                            
        ),                                                                       
    ) 
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1888,35 +1889,31 @@ def test_bal_create_oog_code_deposit(
...
-            gas=1_000_000,
-            address=contract_b,
-            args_offset=0,
-            args_size=0,
-            ret_offset=0,
-            ret_size=0,
-        )
-        + Op.POP  # pop the return value (0 = failure)
-        + Op.SSTORE(0, 1)  # this should succeed (non-static context)
+        code=Op.STATICCALL(gas=1_000_000, address=contract_b)
```
raxhvl: suggestion: Use the static helper

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1951,48 +1948,272 @@ def test_bal_sstore_static_context(
...
+    """
+    SELFDESTRUCT in static context: beneficiary NOT in BAL.
+
+    Static check must fire before beneficiary access (warm/cold lookup)
+    or balance transfer.
+    """
+    beneficiary_balance = 1
+    beneficiary = pre.fund_eoa(amount=beneficiary_balance)
+
+    caller_balance = 100 if caller_has_balance else 0
```
raxhvl: Suggestion: Parameterize the value

Get rid of the intermediate conditional

```py
  @pytest.mark.parametrize(                                         
      "caller_value", [0, 100], ids=["no_balance", "with_balance"]             
  )  
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1951,48 +1948,272 @@ def test_bal_sstore_static_context(
...
+            beneficiary: Account(balance=beneficiary_balance),
+        },
+        tx_access_list=access_list,
+    )
+
+    blockchain_test(pre=pre, blocks=[block], post=post)
+
+
+@pytest.mark.with_all_call_opcodes
+def test_bal_call_opcode_succeeds_in_static_context(
```
raxhvl, line 2210: Suggestion: Use the helper

The proposed helper API makes this test quite short:

```py
def test_bal_call_opcode_succeeds_in_static_context(
    pre: Alloc,                                     
    call_opcode: Op,                                
) -> None:                                          
    target = pre.deploy_contract(code=Op.STOP)      
                                                    
    caller = pre.deploy_contract(                   
        code=call_opcode(address=target) + Op.STOP, 
    )                                               
                                                    
    blockchain_test_under_static_call(              
        pre,                                        
        blockchain_test,                            
        static_call_target=caller,                  
        bal_expectations={                          
            caller: BalAccountExpectation.empty(),  
            target: BalAccountExpectation.empty(),  
        },                                          
    )                                               

```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -2015,25 +2236,93 @@ def test_bal_call_with_value_in_static_context(
...
-            static_caller: Account(storage={0: 1}),
-            # no transfer occurred, balances unchanged
-            caller: Account(balance=caller_starting_balance),
-            target: Account(balance=target_starting_balance),
-        },
+        post={static_caller: Account(storage={0: 1})},
+    )
+
+
+def test_bal_callcode_with_value_in_static_context(
```
raxhvl, line 2237: Suggestion: Use helper to simplify

This test is quite simlar to the one above, but its cleaner to keep it separate, 


```py
def test_bal_callcode_with_value_in_static_context(                       
    pre: Alloc,                                                           
) -> None:                                                                
    """                                                                   
    CALLCODE with nonzero value succeeds in static context.               
                                                                          
    EIP-214 explicitly excludes CALLCODE from write-protection.           
    """                                                                   
    target = pre.deploy_contract(code=Op.STOP)                            
                                                                          
    caller = pre.deploy_contract(                                         
        code=Op.CALLCODE(gas=100_000, address=target, value=1) + Op.STOP, 
        balance=10**18,                                                   
    )                                                                     
                                                                          
    blockchain_test_under_static_call(                                    
        pre,                                                              
        blockchain_test,                                                  
        static_call_target=caller,                                        
        bal_expectations={                                                
            caller: BalAccountExpectation.empty(),                        
            target: BalAccountExpectation.empty(),                        
        },                                                                
    )                                                                     

```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1951,48 +1948,272 @@ def test_bal_sstore_static_context(
...
+    Build the block (with BAL expectations) and post-state for a static
+    context test. Deploys a static_caller that STATICCALLs caller and
+    continues with SSTORE to prove execution resumed.
+    """
+    alice = pre.fund_eoa()
+
+    static_caller = pre.deploy_contract(
+        code=Op.STATICCALL(gas=1_000_000, address=caller)
+        + Op.POP
+        + Op.SSTORE(0, 1)
```
raxhvl: duplicate issue: [see this comment](https://github.com/ethereum/execution-specs/pull/1903/changes#r3013417623)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: I did think about this but I think it's a good idea to test possible code paths in client implementations. It's kind of impossible to guess the different code paths that could be triggered so imo it may just be good to have the sanity check that only adds one extra test case. Thoughts?

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
fselmo: Yeah I had it like this originally. I chose to refactor it back out because it's easier for me to read what's actually going into the `blockchain_test` (as all tests have this behavior). But looking at it again I'm not sure which way I prefer tbh. Since you voiced a preference here, I think we can run with this approach 👍🏼

fselmo:
Thanks @raxhvl. I should have most of the comments implemented here. I think we should keep the parametrization of the storage pre-value as it adds one extra test and keeps a sanity check on possibly different existing client logic - can maybe be convinced that it's overkill 😄.

fselmo:
@raxhvl if you give the nod here, feel free to merge if you have permissions... lgtm 👍🏼

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -1951,89 +1959,311 @@ def test_bal_sstore_static_context(
...
     )
 
 
-def test_bal_call_with_value_in_static_context(
+def blockchain_test_under_static_call(
     pre: Alloc,
     blockchain_test: BlockchainTestFiller,
+    *,
+    static_call_target: Address,
+    expected_not_in_bal: list[Address] | None = None,
```
raxhvl: question:  do we need both `expected_not_in_bal` and  `bal_expectations`?

`bal_expectations`  can assert absense in BAL

```py
 blockchain_test_under_static_call(
        expected_not_in_bal=[target],
    )
```

can be written as 

```py
 blockchain_test_under_static_call(
        bal_expectations={                                                       
            target: None,                                                        
        }
    )
```

fselmo: Added

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
raxhvl: Sure - wouldn't harm to keep it.

raxhvl (approved):
All clear! I dont have perms to merge.

fselmo (approved):

PR #1904, opened 2025-12-11
feat(benchmarks): fix benchmark for SELFDESTRUCT of existing accounts for Osaka

jsign:
## 🗒️ Description
This PR fixes the `test_selfdestruct_existing` benchmark to work in Osaka.

To validate this still does coherent work, I compare in this PR filling for Osaka vs Prague, tested with 30M target:

- Osaka:
    - `test_system.py::test_selfdestruct_existing[fork_Osaka-blockchain_test-value_bearing_False]` 154M cycles
    - `test_system.py::test_selfdestruct_existing[fork_Osaka-blockchain_test-value_bearing_True]` 188M cycles
- Prague
    - `test_system.py::test_selfdestruct_existing[fork_Prague-blockchain_test-value_bearing_False]` 154 cycles
    - `test_system.py::test_selfdestruct_existing[fork_Prague-blockchain_test-value_bearing_True]` 189 cycles

So looks like it is working fine reg the amount of work in Prague vs Osaka in this PR. (Prague slightly doing more effort, but makes sense due to partitioning gas overheads)

## 🔗 Related Issues or PRs

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -630,7 +630,7 @@ def test_selfdestruct_existing(
                 Transaction(
                     to=code_addr,
                     gas_limit=tx_gas_limit,
-                    data=Hash(start) + Hash(count),
+                    data=Hash(start),
                     sender=pre.fund_eoa(),
                 )
             )
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -515,10 +515,10 @@ def test_selfdestruct_existing(
         + 63  # ~Gluing opcodes
     )
     final_storage_gas = (
-        gas_costs.G_VERY_LOW # ADD
-        + gas_costs.G_VERY_LOW * 3 # PUSHs
-        + gas_costs.G_COLD_SLOAD # SSTORE cold
-        + gas_costs.G_STORAGE_RESET # SSTORE new value
+        gas_costs.G_VERY_LOW  # ADD
+        + gas_costs.G_VERY_LOW * 3  # PUSHs
+        + gas_costs.G_COLD_SLOAD  # SSTORE cold
+        + gas_costs.G_STORAGE_RESET  # SSTORE new value
     )
     memory_expansion_cost = fork().memory_expansion_gas_calculator()(
         new_bytes=96
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -479,15 +479,14 @@ def test_selfdestruct_existing(
     fork: Fork,
     pre: Alloc,
     value_bearing: bool,
-    env: Environment,
     gas_benchmark_value: int,
+    tx_gas_limit: int,
 ) -> None:
     """
     Benchmark SELFDESTRUCT instruction for existing contracts.
     contracts.
     """
     attack_gas_limit = gas_benchmark_value
-    effective_tx_gas_limit = fork.transaction_gas_limit_cap() or env.gas_limit
     fee_recipient = pre.fund_eoa(amount=1)
 
     # Template code that will be used to deploy a large number of contracts.
@@ -576,9 +575,7 @@ def test_selfdestruct_existing(
     # The deployed code length is two-bytes. The dominant factor
     # is the static cost, plus a few glue opcodes/memory-expansion costs.
     estimated_gas_per_creation = gas_costs.G_CREATE + 3_000
-    max_creations_per_tx = int(
-        effective_tx_gas_limit // estimated_gas_per_creation
-    )
+    max_creations_per_tx = int(tx_gas_limit // estimated_gas_per_creation)
     num_setup_txs = math.ceil(num_contracts / max_creations_per_tx)
 
     setup_txs = []
@@ -590,7 +587,7 @@ def test_selfdestruct_existing(
             setup_txs.append(
                 Transaction(
                     to=factory_caller_address,
-                    gas_limit=effective_tx_gas_limit,
+                    gas_limit=tx_gas_limit,
                     data=Hash(count),
                     sender=pre.fund_eoa(),
                 )
@@ -624,13 +621,7 @@ def test_selfdestruct_existing(
     code_addr = pre.deploy_contract(code=code, storage={0: 1})
 
     # Calculate max targets per execution TX based on effective gas limit
-    gas_limit_cap = fork.transaction_gas_limit_cap()
-    effective_attack_gas_limit = (
-        min(attack_gas_limit, gas_limit_cap)
-        if gas_limit_cap is not None
-        else attack_gas_limit
-    )
-    max_targets_per_tx = (effective_attack_gas_limit - base_costs) // loop_cost
+    max_targets_per_tx = (tx_gas_limit - base_costs) // loop_cost
     num_exec_txs = math.ceil(num_contracts / max_targets_per_tx)
 
     exec_txs = []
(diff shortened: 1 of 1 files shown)


tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie, line 517: This should be adjusted since the loop implementation had changed.

This is the updated version i calculated:

```python
loop_cost = (
    gas_costs.G_KECCAK_256  # KECCAK static cost
    + math.ceil(85 / 32) * gas_costs.G_KECCAK_256_WORD  # KECCAK dynamic
    # cost for CREATE2
    + gas_costs.G_VERY_LOW * 3  # MSTORE + 2x ADD
    + gas_costs.G_COLD_ACCOUNT_ACCESS  # CALL to self-destructing contract
    + gas_costs.G_BASE  # COINBASE inside called contract
    + gas_costs.G_SELF_DESTRUCT
    + 88 # loop_glue
)
```

I get the loop_glue value by analyzing the traces, the exact calculation is as follows:

```python
# Loop condition glue opcodes: 2*MLOAD + 2*CALLDATALOAD + 2*GAS + LT + GT +
# AND + PC + POP + JUMPI + JUMPDEST + ~13*PUSH1 + PUSH2 + PUSH4 + SUB
loop_glue = (
    gas_costs.G_VERY_LOW * 2  # 2x MLOAD in loop condition
    + gas_costs.G_VERY_LOW * 2  # 2x CALLDATALOAD in loop condition
    + gas_costs.G_BASE * 2  # 2x GAS
    + gas_costs.G_VERY_LOW  # LT
    + gas_costs.G_VERY_LOW  # GT
    + gas_costs.G_VERY_LOW  # AND
    + gas_costs.G_BASE  # PC (for JUMPDEST target)
    + gas_costs.G_BASE  # POP (CALL return value)
    + gas_costs.G_HIGH  # JUMPI
    + gas_costs.G_JUMPDEST  # JUMPDEST
    + gas_costs.G_VERY_LOW * 13  # ~13 PUSH1 per iteration
    + gas_costs.G_VERY_LOW  # PUSH2
    + gas_costs.G_VERY_LOW  # PUSH4
    + gas_costs.G_VERY_LOW  # SUB (loop counter)
)
```

LouisTsai-Csie, line 523: For the first transaction, the cost is `G_STORAGE_RESET` (1 → 42). But starting from the second transaction, the operation becomes an `SSTORE` to the same value (42 → 42), which results in a lower gas cost.

There are three possible paths forward, what do you think is the best approach?
1. Remove the storage check entirely, since we already validate the expected gas usage in the benchmark tests (`expected_benchmark_gas_used`).
2. Initialize the storage slot to 42, so subsequent writes do not change the value.
3. Keep the initial storage slot at 1 and write a new value on each transaction to ensure the higher gas cost is consistently applied.

For simplicity i will go for solution 2 for the following comments.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -571,41 +572,71 @@ def test_selfdestruct_existing(
...
+                    Op.ADD(Op.CALLDATALOAD(0), Op.CALLDATALOAD(32)),
+                ),
+            ),
         )
         + Op.SSTORE(0, 42)  # Done for successful tx execution assertion below.
     )
     assert len(code) <= fork.max_code_size()
 
     # The 0 storage slot is initialize to avoid creation costs in SSTORE above.
     code_addr = pre.deploy_contract(code=code, storage={0: 1})
```
LouisTsai-Csie, line 618: Following the previous comment, i update the storage to `{0: 42}` here


```suggestion
    code_addr = pre.deploy_contract(code=code, storage={0: 42})
```

LouisTsai-Csie (changes requested):
I’m considering whether we still want to keep the sanity check using the `expected_benchmark_gas_used` approach, as it adds quite a bit of complexity.

If we stick with the current approach, we also need to account for calldata costs, since zero and non-zero bytes are priced differently. There are two possible options:
- Start the index at 0 and explicitly calculate the calldata cost for each case based on the number of zero bytes, or
- Start the index at 257 (0x101) to ensure all calldata consists of non-zero bytes, resulting in a consistent calldata cost.

Neither approach is easy to handle. For the second option, we would need to ensure that this assumption holds consistently across all relevant tests and does not introduce unintended side effects.

This is my current fix, which i start the index from 257 (`0x101`). We could also use the storage check as the only sanity check here, and we could use `skip_gas_used_validation` to ignore the gas usage check!

My fix: https://gist.github.com/05af49781f707aee93b6b6114da79d92.git

tests/benchmark/compute/instruction/test_system.py
jsign: Ah, lets go with option 3 actually, since is something applied in [another PR](https://github.com/ethereum/execution-specs/pull/1906/commits/2122f617394a87f7e8349e56d847dfe3037f9364).

I'll change this, and this `final_storage_bas` should have a predictable value on each tx.

Edit: Done.

tests/benchmark/compute/instruction/test_system.py
jsign: I'll keep this as is, since I switched to option 3. idea in my previous comment.

tests/benchmark/compute/instruction/test_system.py
jsign: @LouisTsai-Csie, I changed it. Thanks btw for the recalculation! (Edit: Looks like it is still failing)

I'm feeling this gas tracking for the expected gas is prob already getting quite hairy. In the coming repricings, all this part of benchmarks will get a big hit if constants for particular opcodes change (e.g. something being G_BASE switches to G_VERY_LOW or similar).

Are we getting enough return on investment on doing this, if we are already doing the post-state checks making sure the transactions aren't failing?

jsign: Sorry, saw your comment here :sweat_smile: :
> We could also use the storage check as the only sanity check here, and we could use skip_gas_used_validation to ignore the gas usage check!

I think I agree with this, since I switched to a stronger post-check ensuring all txs can't fail.

LouisTsai-Csie (approved):
LGTM, but i will merge later since we need to wait until `forks/amsterdam` is re-created.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -571,45 +573,73 @@ def test_selfdestruct_existing(
...
-            sender=pre.fund_eoa(),
-        )
+        for i in range(num_exec_txs):
+            start = i * max_targets_per_tx
+            count = min(max_targets_per_tx, num_contracts - start)
+            exec_txs.append(
+                Transaction(
+                    to=code_addr,
+                    gas_limit=tx_gas_limit,
+                    data=Hash(start) + Hash(count),
```
jochem-brouwer: The `count` is not being executed in the `code_addr`? (No `CALLDATALOAD(32)`)
I don't think we need it, since the attack loop exits based on the available gas. So the EVM will handle the "count" here. We should just provide the start, and these start values should be such that later txs do not call into addresses targeted before (but this should be the case due to the `max_targets_per_tx` calculation)

jochem-brouwer (commented):
One small comment!

tests/benchmark/compute/instruction/test_system.py
jsign: Agree! Removed.

jochem-brouwer (approved):
LGTM 😄 👍

LouisTsai-Csie:
LGTM！

PR #1906, opened 2025-12-11
feat(benchmarks): fix benchmark for SELFDESTRUCT of created accounts for Osaka

jsign:
## 🗒️ Description
This PR fixes the `test_selfdestruct_created` benchmark to work in Osaka.

Filling in this PR for Osaka and Prague for **45M** target:
- Osaka:
  - `test_system.py::test_selfdestruct_created[fork_Osaka-blockchain_test-value_bearing_False]` 36M cycles
  - `test_system.py::test_selfdestruct_created[fork_Osaka-blockchain_test-value_bearing_True]` 36M cycles
- Prague:
  - `test_system.py::test_selfdestruct_created[fork_Prague-blockchain_test-value_bearing_False]` 37M cycles
  - `test_system.py::test_selfdestruct_created[fork_Prague-blockchain_test-value_bearing_True]` 36M cycles

Although each fork is coherent with the other, felt quite low number of cycles to me. So I debugged in Geth further and got convinced all is good, since saw all the SELFDESTRUCT instructions being executed.

Then remembered two things:
- We already have a post-state check of storage slot 0 having the value 42, to check that the attack loop ended fine without unexpected EVM errors. So that is already quite good.
- Found the [original PR](https://github.com/ethereum/execution-spec-tests/pull/1678) that created this benchmark, and indeed the used gas was ~35M cycles. So that's good too. (This is why I filled with 45M above to make them comparable).

## 🔗 Related Issues or PRs

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -689,9 +689,10 @@ def test_selfdestruct_created(
         + memory_expansion_calc(new_bytes=32)
     )
     suffix_cost = (
-        gas_costs.G_COLD_SLOAD
+        gas_costs.G_VERY_LOW
+        + gas_costs.G_VERY_LOW * 3
+        + gas_costs.G_COLD_SLOAD
         + gas_costs.G_STORAGE_RESET
-        + (gas_costs.G_VERY_LOW * 2)
     )
 
     base_costs = prefix_cost + suffix_cost + intrinsic_gas_cost_calc()
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -753,7 +753,7 @@ def test_selfdestruct_created(
     benchmark_test(
         post=post,
         blocks=[
-            Block(txs=exec_txs),
+            Block(txs=exec_txs, fee_recipient=fee_recipient),
         ],
         expected_benchmark_gas_used=(
             max_iterations_per_tx * loop_cost + base_costs
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -681,7 +681,12 @@ def test_selfdestruct_created(
         + gas_costs.G_VERY_LOW * 6  # PUSHs, ADD, DUP, GT
         + gas_costs.G_HIGH  # JUMPI
         + gas_costs.G_JUMPDEST
-    )
+extra_costs = (
+    gas_costs.G_BASE * 2  # POP, GAS
+    + gas_costs.G_VERY_LOW * 5  # PUSHs, ADD, DUP, GT
+    + gas_costs.G_HIGH  # JUMPI
+    + gas_costs.G_JUMPDEST
+)
     loop_cost = create_costs + call_costs + extra_costs
 
     prefix_cost = (
@@ -725,11 +730,17 @@ def test_selfdestruct_created(
     )
     code = code_prefix + loop_body + code_suffix
     # The 0 storage slot is initialize to avoid creation costs in SSTORE above.
-    code_addr = pre.deploy_contract(
-        code=code,
-        balance=iterations if value_bearing else 0,
-        storage={0: 1},
-    )
+max_iterations_per_tx = (tx_gas_limit - base_costs) // loop_cost
+num_exec_txs = math.ceil(iterations / max_iterations_per_tx)
+
+total_expected_iterations = max_iterations_per_tx * num_exec_txs
+
+# The 0 storage slot is initialize to avoid creation costs in SSTORE above.
+code_addr = pre.deploy_contract(
+    code=code,
+    balance=total_expected_iterations if value_bearing else 0,
+    storage={0: 1},
+)
 
     max_iterations_per_tx = (tx_gas_limit - base_costs) // loop_cost
     num_exec_txs = math.ceil(iterations / max_iterations_per_tx)
@@ -753,7 +764,10 @@ def test_selfdestruct_created(
         blocks=[
             Block(txs=exec_txs),
         ],
-        expected_benchmark_gas_used=iterations * loop_cost + base_costs,
+        expected_benchmark_gas_used=(
+            max_iterations_per_tx * loop_cost + base_costs
+        )
+        * num_exec_txs
     )
 
 
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -639,6 +639,7 @@ def test_selfdestruct_created(
     fork: Fork,
     env: Environment,
     gas_benchmark_value: int,
+    tx_gas_limit: int,
 ) -> None:
     """
     Benchmark SELFDESTRUCT instruction for deployed contracts within same tx.
@@ -730,15 +731,7 @@ def test_selfdestruct_created(
         storage={0: 1},
     )
 
-    gas_limit_cap = fork.transaction_gas_limit_cap()
-    effective_attack_gas_limit = (
-        min(gas_benchmark_value, gas_limit_cap)
-        if gas_limit_cap is not None
-        else gas_benchmark_value
-    )
-    max_iterations_per_tx = (
-        effective_attack_gas_limit - base_costs
-    ) // loop_cost
+    max_iterations_per_tx = (tx_gas_limit - base_costs) // loop_cost
     num_exec_txs = math.ceil(iterations / max_iterations_per_tx)
 
     exec_txs = []
@@ -747,15 +740,14 @@ def test_selfdestruct_created(
             exec_txs.append(
                 Transaction(
                     to=code_addr,
-                    gas_limit=effective_attack_gas_limit,
+                    gas_limit=tx_gas_limit,
                     sender=pre.fund_eoa(),
                 )
             )
 
     post = {
         code_addr: Account(storage={0: len(exec_txs) + 1})
-    }  # Check for successful
-    # execution.
+    }  # Check for successful execution.
     benchmark_test(
         post=post,
         blocks=[
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -643,10 +643,9 @@ def test_selfdestruct_created(
     """
     Benchmark SELFDESTRUCT instruction for deployed contracts within same tx.
     """
-    # Avoid account creation costs in the SELFDESTRUCT receipient.
+    # Avoid account creation costs in the SELFDESTRUCT recipient.
     fee_recipient = pre.fund_eoa(amount=1)
     env.fee_recipient = fee_recipient
-
 
     # SELFDESTRUCT(COINBASE) contract deployment
     initcode = (
@@ -677,18 +676,17 @@ def test_selfdestruct_created(
         + gas_costs.G_BASE  #  Parameter GAS
     )
     extra_costs = (
-        gas_costs.G_BASE * 2  # POP, GAS
-        + gas_costs.G_VERY_LOW
-        * 18  # 6 PUSHs, MSTORE, MLOAD, ADD, GT, LT, AND, 2 CALLDATALOADs
+        gas_costs.G_BASE  # POP
+        + gas_costs.G_VERY_LOW * 6  # PUSHs, ADD, DUP, GT
         + gas_costs.G_HIGH  # JUMPI
         + gas_costs.G_JUMPDEST
     )
     loop_cost = create_costs + call_costs + extra_costs
 
     prefix_cost = (
-        gas_costs.G_VERY_LOW * 7  # 2 MSTOREs, 4 PUSHs, 1 CALLDATALOAD
-        + gas_costs.G_JUMPDEST
-        + memory_expansion_calc(new_bytes=64)  # Memory for 2 MSTORE slots
+        gas_costs.G_VERY_LOW * 3
+        + gas_costs.G_BASE
+        + memory_expansion_calc(new_bytes=32)
     )
     suffix_cost = (
         gas_costs.G_COLD_SLOAD
@@ -700,11 +698,7 @@ def test_selfdestruct_created(
 
     iterations = (gas_benchmark_value - base_costs) // loop_cost
 
-    code_prefix = (
-        Op.MSTORE(0, initcode.hex())
-        + Op.MSTORE(32, Op.CALLDATALOAD(0))
-        + Op.JUMPDEST
-    )
+    code_prefix = Op.MSTORE(0, initcode.hex()) + Op.PUSH0 + Op.JUMPDEST
     code_suffix = (
         Op.SSTORE(0, 42)  # Done for successful tx execution assertion below.
         + Op.STOP
@@ -719,17 +713,9 @@ def test_selfdestruct_created(
                 )
             )
         )
-        + Op.MSTORE(32, Op.ADD(Op.MLOAD(32), 1))
-        + Op.JUMPI(
-            len(code_prefix) - 1,
-            Op.AND(
-                Op.GT(Op.GAS, suffix_cost + loop_cost),
-                Op.LT(
-                    Op.MLOAD(32),
-                    Op.ADD(Op.CALLDATALOAD(0), Op.CALLDATALOAD(32)),
-                ),
-            ),
-        )
+        + Op.PUSH1[1]
+        + Op.ADD
+        + Op.JUMPI(len(code_prefix) - 1, Op.GT(Op.GAS, Op.DUP1))
     )
     code = code_prefix + loop_body + code_suffix
     # The 0 storage slot is initialize to avoid creation costs in SSTORE above.
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -677,16 +677,11 @@ def test_selfdestruct_created(
         + gas_costs.G_BASE  #  Parameter GAS
     )
     extra_costs = (
-        gas_costs.G_BASE  # POP
-        + gas_costs.G_VERY_LOW * 6  # PUSHs, ADD, DUP, GT
+        gas_costs.G_BASE * 2  # POP, GAS
+        + gas_costs.G_VERY_LOW * 5  # PUSHs, ADD, DUP, GT
         + gas_costs.G_HIGH  # JUMPI
         + gas_costs.G_JUMPDEST
-extra_costs = (
-    gas_costs.G_BASE * 2  # POP, GAS
-    + gas_costs.G_VERY_LOW * 5  # PUSHs, ADD, DUP, GT
-    + gas_costs.G_HIGH  # JUMPI
-    + gas_costs.G_JUMPDEST
-)
+    )
     loop_cost = create_costs + call_costs + extra_costs
 
     prefix_cost = (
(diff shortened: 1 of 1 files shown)


tests/benchmark/compute/instruction/test_system.py
```diff
@@ -643,9 +643,6 @@ def test_selfdestruct_created(
     """
     Benchmark SELFDESTRUCT instruction for deployed contracts within same tx.
     """
-    fee_recipient = pre.fund_eoa(amount=1)
```
jsign, line 646: I removed this `fee_receipient` since I can't make sense why this is relevant.
After removing, I didn't get any unexpected error or similar.

So, unless we can figure out why this was here, I would prefer to remove it to avoid confusing future readers.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -697,7 +695,11 @@ def test_selfdestruct_created(
 
     iterations = (gas_benchmark_value - base_costs) // loop_cost
 
-    code_prefix = Op.MSTORE(0, initcode.hex()) + Op.PUSH0 + Op.JUMPDEST
+    code_prefix = (
```
jsign: TL;DR of bytecode changes is using CALLDATA 0 to resume in desired address target due to tx partitioning, and CALLDATA 1 for the desired count.

jochem-brouwer:
Maybe idea to help quickly debug these (also for sanity check, for instance here check if SELFDESTRUCT is ran a lot of times), use the opcode counter report? https://github.com/ethereum/execution-specs/pull/1353

tests/benchmark/compute/instruction/test_system.py
jochem-brouwer: This inits COINBASE as an existing account and thus prevents paying the gas for creating a new account on the first SELFDESTRUCT (25k)

This is an optimization and this cost is amortized over large gas limits.

jochem-brouwer: CREATE opcode which is necessary before we can SELFDESTRUCT (to create the to-be-destructed account) already costs 32k base. So deducting to this 25k extra cost will thus, at most, allow one more SELFDESTRUCT when compared to not initializing the target COINBASE.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -723,18 +733,39 @@ def test_selfdestruct_created(
         balance=iterations if value_bearing else 0,
         storage={0: 1},
```
jochem-brouwer, line 736: I think we should store `SSTORE(0, ADD(SLOAD(0), 1)` here as sanity check to confirm all txs are valid (and not reverted)

jochem-brouwer: Since it is initialized to 1, the final value should thus be the txs ran + 1. (This is the post state condition)

tests/benchmark/compute/instruction/test_system.py
jochem-brouwer: What is the calldata encoding? We cannot resume at a desired address because we are using `Op.CREATE` (not `CREATE2`). 

I actually think for this test it is enough to just test for `Op.GT(Op.GAS, suffix_cost + loop_cost)`. We let the EVM decide if it has enough gas left to perform another loop (and pay enough for the SSTORE). So we do not have to encode any details into calldata.

(We can thus even use the "normal" tx splitter here without adding the calldata split logic)

jochem-brouwer (commented):
Some comments 😄 👍

tests/benchmark/compute/instruction/test_system.py
jsign: Ahh yes. I re-enabled it and added a comment to remember why we did this.

tests/benchmark/compute/instruction/test_system.py
jsign: Great idea. Added it.

tests/benchmark/compute/instruction/test_system.py
jsign: Ah, I wanted to be more strict in the expected number of SELFDESTRUCT; but I agree it is better to use `GAS` and run until the end. Changed it.

jsign: Ah, I wanted to be more strict in the expected number of SELFDESTRUCT; but I agree it is better to use `GAS` and run until the end. Changed it.

jsign:
> Maybe idea to help quickly debug these (also for sanity check, for instance here check if SELFDESTRUCT is ran a lot of times), use the opcode counter report? #1353

@jochem-brouwer , Oh, interesting.

Trying this this way:
- `uv run fill --from=Osaka --until=Osaka --evm-bin=/data/code-data/go-ethereum/evm -m "benchmark and blockchain_test" --block-gas-limit 45000000 -k "test_selfdestruct_created" --evm-dump-dir=dumpy` (I'm filling with Geth).
- `cat /data/code-data/execution-specs/dumpy/benchmark__compute__instruction__system__test_selfdestruct_created/fork_Osaka_blockchain_test_value_bearing_False/0/stdin.txt | uv run ethereum-spec-evm t8n \
  --input.alloc=stdin --input.txs=stdin --input.env=stdin \
  --output.result=stdout --output.alloc=stdout --output.body=stdout \
  --state.fork=Osaka --state.chainid=1 --state.reward=0 \
  --opcode.count stdout`
  
But I see:
```
"opcodeCount": {
        "CALLER": 1,
        "PUSH20": 1,
        "EQ": 4,
        "PUSH1": 11,
        "JUMPI": 6,
        "JUMPDEST": 8,
        "SLOAD": 4,
        "DUP1": 4,
        "DUP3": 3,
        "SUB": 1,
        "GT": 2,
        "PUSH0": 8,
        "DUP2": 2,
        "PUSH2": 5,
        "SWAP2": 1,
        "ADD": 2,
        "SWAP3": 1,
        "SWAP1": 1,
        "POP": 3,
        "SSTORE": 4,
        "PUSH32": 1,
        "ISZERO": 1,
        "JUMP": 1,
        "MUL": 1,
        "RETURN": 1
    }
```

This doesn't make much sense to me, since there is no SELFDESTRUCT there.
But if I debug with t8n in Geth with the same stdin.txt, I definitely see breakpoints in the SELFDESTRUCT instruction handler. (And clearly the post-state check is passing).

Feels I'm misusing the opcode counter or might have a bug?

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -714,7 +717,9 @@ def test_selfdestruct_created(
         )
         + Op.PUSH1[1]
         + Op.ADD
-        + Op.JUMPI(len(code_prefix) - 1, Op.GT(iterations, Op.DUP1))
+        + Op.JUMPI(
+            len(code_prefix) - 1, Op.GT(Op.GAS, suffix_cost + loop_cost)
```
jochem-brouwer, line 723: `suffix_cost` is now increased due to the SLOAD

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -723,18 +728,41 @@ def test_selfdestruct_created(
...
+    exec_txs = []
+    with TestPhaseManager.execution():
+        for i in range(num_exec_txs):
+            start = i * max_iterations_per_tx
+            count = min(max_iterations_per_tx, iterations - start)
+            exec_txs.append(
+                Transaction(
+                    to=code_addr,
+                    gas_limit=effective_attack_gas_limit,
+                    data=Hash(start) + Hash(count),
```
jochem-brouwer: I believe we do not have to use data at all here. The test will create new contracts due to `Op.CREATE` (there is no way to influence it. In fact, the account `nonce` will be used as live "tracker" for our current progress, since that will be used to generate the target address). So we dont need to track our progress at all, this will be done by the EVM (and we dont have a way to influence it either).

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -723,18 +728,41 @@ def test_selfdestruct_created(
...
+                    sender=pre.fund_eoa(),
+                )
+            )
+
+    post = {
+        code_addr: Account(storage={0: len(exec_txs) + 1})
```
jochem-brouwer, line 751: I think what we can do here is use the `benchmark_test` transaction splitter (see for instance here: https://github.com/ethereum/execution-specs/blob/2a2d0d5ed92d53fa64e87c69fd0e6537974d950c/tests/benchmark/compute/instruction/test_bitwise.py#L195). We do not need the calldata, so we can simply setup a tx to target `code_addr`. The only thing we need here to calculate the post state correctly is the amount of transactions generated. This might however be a problem, I'm not sure how to get this from the benchmark test runner. @LouisTsai-Csie might know if there is a simple way to get the tx count from it?

jochem-brouwer (commented):
Some more comments 😄 👍

tests/benchmark/compute/instruction/test_system.py
jsign: Sorry, this was a leftover.

tests/benchmark/compute/instruction/test_system.py
jsign: You mean for extra pushes? (if that's the case, I agree). But interestingly, we already have a G_COLD_SLOAD before that I think was incorrrect (but now would be fine), no?

tests/benchmark/compute/instruction/test_system.py
jsign: How would `benchmark_test(tx=tx) ` know how to split the tx into required transactions? Wouldn't need also to pass `num_exec_txs` as some parameter?

tests/benchmark/compute/instruction/test_system.py
jochem-brouwer: We have:

```
    suffix_cost = (
        gas_costs.G_COLD_SLOAD
        + gas_costs.G_STORAGE_RESET
        + (gas_costs.G_VERY_LOW * 2)
    )
```

This is the suffix cost of the previous code of `Op.SSTORE(0, 42)`. It accounts for: the two pushes, the changing of storage value (`G_STORAGE_RESET` -> this actually looks incorrect, it should be 5000 and not the 2900 per https://eips.ethereum.org/EIPS/eip-2200#sstore-changes) but also the fact that we are loading a cold key (this is, somewhat confusing, `G_COLD_SLOAD`). This accounts for having to load the storage, because for the gas calculation we need the value on disk to calculate the actual gas cost (creating new keys is more expensive). It is also cold (not in access list) so we charge `G_COLD_SLOAD`.

The confusion here comes from that the SSTORE gas cost thus here incurs a cost which has the name SLOAD in it. But this is actually correct per spec, see https://eips.ethereum.org/EIPS/eip-2929#sstore-changes 😄 👍 

Now we change to:

```
    code_suffix = (
        Op.SSTORE(
            0, Op.ADD(Op.SLOAD(0), 1)
        )  # Done for successful tx execution assertion below.
        + Op.STOP
    )
```

I believe correct suffix cost would be:

`gas_costs.G_VERY_LOW + gas_costs.G_VERY_LOW * 3 + gas_costs.G_COLD_SLOAD + gas_costs.G_SSTORE_RESET_GAS`

Note: `G_SSTORE_RESET_GAS` does not exist, it should be 5000 per EIP-2200. (also not sure why this term is called "reset"). The first `G_VERY_LOW` is the `ADD`, the other 3 are for PUSH, the `G_COLD_SLOAD` is for `SLOAD` and `G_SSTORE_RESET_GAS` is for the `SSTORE`.

jochem-brouwer:
RE: opcode counter. 

This seems to report the wrong block? It's just guessing here, I'd assume Geth is reporting on the wrong block. One of my hunches could be the system contracts (beacon root storing / blockhash storing), but CALLER is only applied once so that does not seem to be the case :thinking:

tests/benchmark/compute/instruction/test_system.py
jochem-brouwer: As far as I understand it, it takes the gas benchmark value and uses that to split txs, see;

https://github.com/ethereum/execution-specs/blob/2a2d0d5ed92d53fa64e87c69fd0e6537974d950c/packages/testing/src/execution_testing/specs/benchmark.py#L352-L359

https://github.com/ethereum/execution-specs/blob/2a2d0d5ed92d53fa64e87c69fd0e6537974d950c/packages/testing/src/execution_testing/specs/benchmark.py#L322

@LouisTsai-Csie might have more insights here :smile: :+1:

tests/benchmark/compute/instruction/test_system.py
jsign: > G_SSTORE_RESET_GAS does not exist, it should be 5000 per EIP-2200.

I agree with your last formula but I think instead of your (created) G_SSTORE_RESET_GAS, using `G_STORAGE_RESET` looks correct. [2929](https://eips.ethereum.org/EIPS/eip-2929#sstore-changes) isn't saying that the new value is `5000 - COLD_SLOAD_COST` which is G_STORAGE_RESET?. Your 5000 is basically G_STORAGE_RESET+G_COLD_SLOAD, no? (and we already included the G_COLD_SLOAD factor)

jsign: Another way to see it in evm.codes:
image
So your 5000 is the 2100 of cold plus the 2900. i.e G_STORAGE_RESET+G_COLD_SLOAD

tests/benchmark/compute/instruction/test_system.py
jsign: In the current code the number of transactions required is calculated as:
```
max_iterations_per_tx = (
        effective_attack_gas_limit - base_costs
    ) // loop_cost
num_exec_txs = math.ceil(iterations / max_iterations_per_tx)
```
So at a minimum this other helper would need `base_costs` and `loop_costs` to even try calculating the same.  :thinking:

tests/benchmark/compute/instruction/test_system.py
jochem-brouwer: Ah yes, that looks correct, so my 5k is actually wrong. So it should charge the 2900 I think for the SSTORE? Plus the other costs (cold sload + pushs + add)

tests/benchmark/compute/instruction/test_system.py
jochem-brouwer: Yes you are right. I think this test could also be written using the benchmark tx splitter (but still have to know the balance of the contract to fund the selfdestructs). This might actually be cleaner, can keep it like this.

jochem-brouwer (approved):
LGTM

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie, line 684:
```suggestion
extra_costs = (
    gas_costs.G_BASE * 2  # POP, GAS
    + gas_costs.G_VERY_LOW * 5  # PUSHs, ADD, DUP, GT
    + gas_costs.G_HIGH  # JUMPI
    + gas_costs.G_JUMPDEST
)
```

This should be changed accordingly since the loop body logic is changed. Adding one `GAS` and remove `DUP1`.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -723,18 +730,29 @@ def test_selfdestruct_created(
         balance=iterations if value_bearing else 0,
         storage={0: 1},
     )
```
LouisTsai-Csie, line 737: The iteration count calculated here does not reflect the actual number of iterations executed, as it is calculated from the block gas limit rather than the transaction gas limit cap:

```python
iterations = (gas_benchmark_value - base_costs) // loop_cost
```

We need to move the `num_exec_txs` calculation to this point and update the balance for `code_addr` accordingly. Otherwise, the balance may be insufficient, making the execution to fail.


```suggestion
max_iterations_per_tx = (tx_gas_limit - base_costs) // loop_cost
num_exec_txs = math.ceil(iterations / max_iterations_per_tx)

total_expected_iterations = max_iterations_per_tx * num_exec_txs

# The 0 storage slot is initialize to avoid creation costs in SSTORE above.
code_addr = pre.deploy_contract(
    code=code,
    balance=total_expected_iterations if value_bearing else 0,
    storage={0: 1},
)
```

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -723,18 +730,29 @@ def test_selfdestruct_created(
...
-        gas_limit=gas_benchmark_value,
-        sender=pre.fund_eoa(),
-    )
 
-    post = {code_addr: Account(storage={0: 42})}  # Check for successful
-    # execution.
-    state_test(
-        pre=pre,
+    max_iterations_per_tx = (tx_gas_limit - base_costs) // loop_cost
+    num_exec_txs = math.ceil(iterations / max_iterations_per_tx)
```
LouisTsai-Csie, line 729: Remove here since we already done in previous section

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -723,18 +730,29 @@ def test_selfdestruct_created(
...
+    post = {
+        code_addr: Account(storage={0: len(exec_txs) + 1})
+    }  # Check for successful execution.
+    benchmark_test(
         post=post,
-        tx=code_tx,
+        blocks=[
+            Block(txs=exec_txs),
+        ],
         expected_benchmark_gas_used=iterations * loop_cost + base_costs,
```
LouisTsai-Csie: This does not consider the multiple transactions implementation, we need to update as:


```suggestion
        expected_benchmark_gas_used=(
            max_iterations_per_tx * loop_cost + base_costs
        )
        * num_exec_txs
```

LouisTsai-Csie (changes requested):
Thanks @jsign and @jochem-brouwer , i add some comments for the root cause of the failing benchmark CI, there is some gas mismatch. After applying the suggested changes it should work as expected for both `fill` and `execute`!

LouisTsai-Csie:
@jsign @jochem-brouwer 

Regarding the opcode count feature, currently, when filling the tests with `evmone`, the opcode counts are recorded in the fixture files, please check the `_info` field in the corresponding files.

Given this, do you think we still need the opcode-count sanity check here? Personally, I think:
- The gas-usage check is sufficient for this scenario.
- Most of benchmark cases also cover the fixed-opcode-count tests, where the sanity check could be performed instead (Please check PR #1869 )

That said, if you feel additional checks are still necessary, I’m happy to explore further options here.

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: The current challenge is that most benchmark code generators do not support scenarios where the number of iterations are calculated dynamically based on gas consumption. For example:

```python
base_cost = XXX
loop_cost = XXX

iterations = (tx_gas_limit - base_cost) // loop_cost
```

This pattern is not currently supported. IMO this logic should be generalized so that all benchmark code generators can handle it in a compatible way.

tests/benchmark/compute/instruction/test_system.py
jsign: Thank you! Applied.

jsign:
> @jsign @jochem-brouwer
> 
> Regarding the opcode count feature, currently, when filling the tests with `evmone`, the opcode counts are recorded in the fixture files, please check the `_info` field in the corresponding files.
> 
> Given this, do you think we still need the opcode-count sanity check here? Personally, I think:
> 
> * The gas-usage check is sufficient for this scenario.
> * Most of benchmark cases also cover the fixed-opcode-count tests, where the sanity check could be performed instead (Please check PR [feat(test-benchmark): implement opcode count verification #1869](https://github.com/ethereum/execution-specs/pull/1869) )
> 
> That said, if you feel additional checks are still necessary, I’m happy to explore further options here.

I think as is it should be okay -- eventually we can keep adding checks if we feel they can provide extra safety checks.

jsign:
@LouisTsai-Csie, applied some extra changes reg the last review -- maybe you want to do a final check before merging!

LouisTsai-Csie (changes requested):
One more minor suggestion to fix the failing CI!

jsign:
> One more minor suggestion to fix the failing CI!

Ah, I think the comment might get lost?
Let me check.

Edit: found the problem, pushed the fix.

LouisTsai-Csie (approved):
Sorry, i indeed missed to submit the suggested changes...

LGTM, thanks a lot

PR #1908, opened 2025-12-12
enable BPO1 and BPO2 for Amsterdam

qu0b:
Geth is failing many tests as the BPOs 1 and 2 are hard coded


fselmo:
I'd like to wait on this until maybe next release. We are just now moving into a new system for filling tests with BPOs and this seems like it's just a trick to please Geth and no other client. So I'd rather wait to include this until we align everything on the testing side and with all client expectations for the BAL tests.

fselmo:
closing in favor of #1973

PR #1911, opened 2025-12-12
Qu0b/fix pre alloc group enginex

qu0b:
patch pre allocation issues


fselmo (approved):
lgtm! There may be a better way to rewrite these tests (perhaps not) but we can address them later. Thanks for the fixes!

PR #1914, opened 2025-12-12
feat(benchmarks): fix benchmark for SELFDESTRUCT of initcode contracts for Osaka

jsign:
## 🗒️ Description
This PR fixes the `test_selfdestruct_initcode` benchmark to work in Osaka.

Very similar fix as I did in https://github.com/ethereum/execution-specs/pull/1906

## 🔗 Related Issues or PRs

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -769,8 +769,8 @@ def test_selfdestruct_initcode(
         + gas_costs.G_INITCODE_WORD
     )
     extra_costs = (
-        gas_costs.G_BASE  # POP
-        + gas_costs.G_VERY_LOW * 6  # PUSHs, ADD, DUP, GT
+        gas_costs.G_BASE * 2  # POP, GAS
+        + gas_costs.G_VERY_LOW * 5  # PUSHs, ADD, GT
         + gas_costs.G_HIGH  # JUMPI
         + gas_costs.G_JUMPDEST
     )
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -741,7 +741,7 @@ def test_selfdestruct_created(
 
 @pytest.mark.parametrize("value_bearing", [True, False])
 def test_selfdestruct_initcode(
-    state_test: StateTestFiller,
+    benchmark_test: BenchmarkTestFiller,
     pre: Alloc,
     value_bearing: bool,
     fork: Fork,
@@ -749,6 +749,7 @@ def test_selfdestruct_initcode(
     gas_benchmark_value: int,
 ) -> None:
     """Benchmark SELFDESTRUCT instruction executed in initcode."""
+    # Avoid account creation costs in the SELFDESTRUCT recipient.
     fee_recipient = pre.fund_eoa(amount=1)
     env.fee_recipient = fee_recipient
 
@@ -780,9 +781,10 @@ def test_selfdestruct_initcode(
         + memory_expansion_calc(new_bytes=32)
     )
     suffix_cost = (
-        gas_costs.G_COLD_SLOAD
+        gas_costs.G_VERY_LOW
+        + gas_costs.G_VERY_LOW * 3
+        + gas_costs.G_COLD_SLOAD
         + gas_costs.G_STORAGE_RESET
-        + (gas_costs.G_VERY_LOW * 2)
     )
 
     base_costs = prefix_cost + suffix_cost + intrinsic_gas_cost_calc()
@@ -792,7 +794,9 @@ def test_selfdestruct_initcode(
     initcode = Op.SELFDESTRUCT(Op.COINBASE)
     code_prefix = Op.MSTORE(0, initcode.hex()) + Op.PUSH0 + Op.JUMPDEST
     code_suffix = (
-        Op.SSTORE(0, 42)  # Done for successful tx execution assertion below.
+        Op.SSTORE(
+            0, Op.ADD(Op.SLOAD(0), 1)
+        )  # Done for successful tx execution assertion below.
         + Op.STOP
     )
 
@@ -806,24 +810,45 @@ def test_selfdestruct_initcode(
         )
         + Op.PUSH1[1]
         + Op.ADD
-        + Op.JUMPI(len(code_prefix) - 1, Op.GT(iterations, Op.DUP1))
+        + Op.JUMPI(
+            len(code_prefix) - 1, Op.GT(Op.GAS, suffix_cost + loop_cost)
+        )
     )
     code = code_prefix + loop_body + code_suffix
 
     # The 0 storage slot is initialize to avoid creation costs in SSTORE above.
     code_addr = pre.deploy_contract(code=code, balance=100_000, storage={0: 1})
-    code_tx = Transaction(
-        to=code_addr,
-        gas_limit=gas_benchmark_value,
-        gas_price=10,
-        sender=pre.fund_eoa(),
-    )
 
-    post = {code_addr: Account(storage={0: 42})}  # Check for successful
+    gas_limit_cap = fork.transaction_gas_limit_cap()
+    effective_attack_gas_limit = (
+        min(gas_benchmark_value, gas_limit_cap)
+        if gas_limit_cap is not None
+        else gas_benchmark_value
+    )
+    max_iterations_per_tx = (
+        effective_attack_gas_limit - base_costs
+    ) // loop_cost
+    num_exec_txs = math.ceil(iterations / max_iterations_per_tx)
+
+    exec_txs = []
+    with TestPhaseManager.execution():
+        for _ in range(num_exec_txs):
+            exec_txs.append(
+                Transaction(
+                    to=code_addr,
+                    gas_limit=effective_attack_gas_limit,
+                    sender=pre.fund_eoa(),
+                )
+            )
+
+    post = {
+        code_addr: Account(storage={0: len(exec_txs) + 1})
+    }  # Check for successful
     # execution.
-    state_test(
-        pre=pre,
+    benchmark_test(
         post=post,
-        tx=code_tx,
+        blocks=[
+            Block(txs=exec_txs),
+        ],
         expected_benchmark_gas_used=iterations * loop_cost + base_costs,
     )
--- a/tests/benchmark/compute/instruction/test_system.py
+++ b/tests/benchmark/compute/instruction/test_system.py
@@ -819,8 +819,8 @@ def test_selfdestruct_initcode(
     # costs in SSTORE above.
     code_addr = pre.deploy_contract(code=code, balance=100_000, storage={0: 1})
 
-    max_iterations_per_tx = (tx_gas_limit - base_costs) // loop_cost
     exec_txs = []
+    expected_benchmark_gas_used = 0
     with TestPhaseManager.execution():
         used_gas = 0
         while used_gas < gas_benchmark_value:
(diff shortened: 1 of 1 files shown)


jsign:
@jochem-brouwer can't assign you (not appear in the list) but feel free to review, should be quite similar to the other one you already reviewed.

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie, line 776: Since the conditional jump in the `loop_body` is modified: 

Before:
```python
Op.JUMPI(len(code_prefix) - 1, Op.GT(iterations, Op.DUP1))

extra_costs = (
    gas_costs.G_BASE  # POP
    + gas_costs.G_VERY_LOW * 6  # PUSHs, ADD, DUP, GT
    + gas_costs.G_HIGH  # JUMPI
    + gas_costs.G_JUMPDEST
)
```

After:
```python
Op.JUMPI(
    len(code_prefix) - 1, Op.GT(Op.GAS, suffix_cost + loop_cost)
)

extra_costs = (
    gas_costs.G_BASE * 2 # POP, GAS
    + gas_costs.G_VERY_LOW * 5  # PUSHs, ADD, GT
    + gas_costs.G_HIGH  # JUMPI
    + gas_costs.G_JUMPDEST
)
```

The extra cost should be modified accordingly, this would affect the final result.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -806,24 +811,37 @@ def test_selfdestruct_initcode(
...
+    }  # Check for successful
     # execution.
-    state_test(
-        pre=pre,
+    benchmark_test(
         post=post,
-        tx=code_tx,
+        blocks=[
+            Block(txs=exec_txs),
+        ],
```
LouisTsai-Csie, line 848: This is a bit tricky, but we need to explicitly set `fee_recipient` for **every** block. Even if we pass `env` into the benchmark test, the fee recipient field is not configured correctly unless it is set on the block itself (AFAIK).


```suggestion
        blocks=[
            Block(txs=exec_txs, fee_recipient=fee_recipient),
        ],
```

If we don’t do this, the fee recipient is not overridden and remains the default value, which is why the current benchmark CI is failing.

Specifically, if the fee recipient is not configured manually, even though we do the following, it is not working...
```python
fee_recipient = pre.fund_eoa(amount=1)
env.fee_recipient = fee_recipient

...
benchmark_test(
    env=evn
    post=post,
    blocks=[
        Block(txs=exec_txs),
    ],
    expected_benchmark_gas_used=iterations * loop_cost + base_costs,
)
```

The fee recipient ends up being an empty account. When ether is transferred to this account, it triggers the value-to-empty-account gas cost (25,000 gas). To avoid introducing this additional complexity into the benchmarks, we apply this workaround (set `fee_recipient` to every block).

Note, as specified in [EIP-150:](https://eips.ethereum.org/EIPS/eip-150):

> If SELFDESTRUCT hits a newly created account, it triggers an additional gas cost of 25,000 (similar to CALLs).

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -806,24 +811,37 @@ def test_selfdestruct_initcode(
...
         expected_benchmark_gas_used=iterations * loop_cost + base_costs,
```
LouisTsai-Csie: This should be updated, since there are more than one transaction, so it would charge more than one `base_costs`.

the new total cost should be as follows:




```suggestion
        expected_benchmark_gas_used=(max_iterations_per_tx * loop_cost + base_costs) * num_exec_txs,
```

LouisTsai-Csie (changes requested):
Thanks for this PR, leave some comment for the failing benchmark due to gas mismatch!

LouisTsai-Csie:
One more thing! Please help change the base branch to `forks/amsterdam` instead of the current `forks/osaka`, sorry for the inconvenience.

tests/benchmark/compute/instruction/test_system.py
jsign: > Even if we pass env into the benchmark test, the fee recipient field is not configured correctly unless it is set on the block itself (AFAIK).

Is this intentional? I would have thought that doing `env.fee_recipient = fee_recipient` should be enough.

(I did the code change anyway to get us unblocked; but wondering if this should be fixed at the framework level since doesn't sounds obvious)

tests/benchmark/compute/instruction/test_system.py
jsign: Thank you! Changed it.

~Is there a way to detect this missing locally? When I fill, doesn't seems this to be checked. It is only checked in CI?~ I think this requries evmone.

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: This is not intentional and should be fixed, the current approach could be workaround tho. I will create an issue or PR for this.

LouisTsai-Csie (approved):
LGTM, will merge later!!

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -806,24 +811,40 @@ def test_selfdestruct_initcode(
...
-    code_tx = Transaction(
-        to=code_addr,
-        gas_limit=gas_benchmark_value,
-        gas_price=10,
-        sender=pre.fund_eoa(),
-    )
 
-    post = {code_addr: Account(storage={0: 42})}  # Check for successful
+    max_iterations_per_tx = (tx_gas_limit - base_costs) // loop_cost
+    num_exec_txs = math.ceil(iterations / max_iterations_per_tx)
```
jochem-brouwer: If we ceil this doesn't this attempt to fill the block in some cases with 1 more tx than is possible? (Does not fit in block gas limit anymore)?

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -806,24 +811,40 @@ def test_selfdestruct_initcode(
...
+
+    exec_txs = []
+    with TestPhaseManager.execution():
+        for _ in range(num_exec_txs):
+            exec_txs.append(
+                Transaction(
+                    to=code_addr,
+                    gas_limit=tx_gas_limit,
```
jochem-brouwer: This does not account for the "last" tx where we should not send all gas, but only what's currently left in the block.

jochem-brouwer (commented):
I think there is a small problem here, as in some cases we might try to fill the block with more txs than can fit?

tests/benchmark/compute/instruction/test_system.py
jsign: I changed a bit this logic in the last commit. I think the new style feels better. LMK!

tests/benchmark/compute/instruction/test_system.py
jsign: See prev comment!

jsign:
I see the base branch was force pushed :cry: I'll rebase again.

Edit: done

jochem-brouwer (approved):
LGTM 😄 👍

tests/benchmark/compute/instruction/test_system.py
LouisTsai-Csie: No, this should not required evmone, it should by default examine the gas usage as sanity check.

tests/benchmark/compute/instruction/test_system.py
```diff
@@ -817,22 +815,25 @@ def test_selfdestruct_initcode(
...
+                break
             exec_txs.append(
                 Transaction(
                     to=code_addr,
-                    gas_limit=tx_gas_limit,
+                    gas_limit=gas_limit,
                     sender=pre.fund_eoa(),
                 )
             )
+            used_gas += gas_limit
```
LouisTsai-Csie, line 837: This change will break the test under Osaka. Consider 120M gas limit, and 16M tx gas limit cap. This would require 7 full transactions (16M) + 1 small transaction (8M).

For the current approach, we have the `max_iterations_per_tx` calculation as above (`max_iterations_per_tx = (tx_gas_limit - base_costs) // loop_cost`), let's assume this value is 1K.

Our `expected_benchmark_gas_used` is calculated as follows, which assume each transaction could execute 1K iterations (`max_iterations_per_tx`), which is not true for the final transaction.
```python
expected_benchmark_gas_used=(
    max_iterations_per_tx * loop_cost + base_costs
)
* len(exec_txs),
```

This does not hold as the final transaction only have 8M gas limit. We might need to track this value like this:

```python
exec_txs = []
    expected_benchmark_gas_used = 0 # The actual gas usage for each transaction
    used_gas = 0
    while used_gas < gas_benchmark_value:
        gas_limit = min(tx_gas_limit, gas_benchmark_value - used_gas)
        if gas_limit < intrinsic_gas_cost_calc():
            break
        exec_txs.append(
            Transaction(
                to=code_addr,
                gas_limit=gas_limit,
                sender=pre.fund_eoa(),
            )
        )
        used_gas += gas_limit
        iterations = (gas_limit - base_costs) // loop_cost
        # This becomes the value passed to `expected_benchmark_gas_used` in `benchmark_test`
        expected_benchmark_gas_used += iterations * loop_cost + base_costs
```

LouisTsai-Csie (changes requested):
One more comment for the failing on Osaka fork. It would be nice if you could help run it to see if there is any gas mismatch, this is a default behavior for all benchmark tests if the `skip_gas_used_validation` is not configured to `True`. And this does not require filling with `evmone`.

By default, the fork running for benchmark tests are Prague now, you might need to explicitly specify it in your command (`--fork Osaka`):
```python
fill -v tests/benchmark/compute/instruction/test_system.py::test_selfdestruct_initcode --gas-benchmark-values 120 --fork Osaka --clean --evm-bin <path-to-geth-or-evmone> -m benchmark
```

BTW, I am running an execute remote sanity check using [anvil](https://getfoundry.sh/anvil/reference/) as a local node, you could try it our if you are interested about it but not a must.

This is my command:

Anvil:
```bash
anvil --accounts 1 --balance 30000 --block-gas-limit 1000000000
```

Note: if you do not configure the block gas limit to a high value, you might get intrinsic gas too low issue.


EELS:
```bash
execute remote --rpc-endpoint=http://127.0.0.1:8545 --rpc-seed-key=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --rpc-chain-id=31337 --gas-benchmark-values 120 tests/benchmark/compute/instruction/test_system.py::test_selfdestruct_initcode --fork Prague -m benchmark
```

tests/benchmark/compute/instruction/test_system.py
jsign: Oh yes, you're right. Fixed it.

jsign:
> One more comment for the failing on Osaka fork. It would be nice if you could help run it to see if there is any gas mismatch, this is a default behavior for all benchmark tests if the `skip_gas_used_validation` is not configured to `True`. And this does not require filling with `evmone`.
> 
> By default, the fork running for benchmark tests are Prague now, you might need to explicitly specify it in your command (`--fork Osaka`):
> 
> ```python
> fill -v tests/benchmark/compute/instruction/test_system.py::test_selfdestruct_initcode --gas-benchmark-values 120 --fork Osaka --clean --evm-bin <path-to-geth-or-evmone> -m benchmark
> ```

@LouisTsai-Csie, oh, but I'm always filling locally with geth in all PRs (and actually every commit fix). This is how I actually get the fixture to double check the zkVM cycles I usually report in PR heads. And I do it both for Prague and Osaka.

But I was filling with:
```
uv run fill --from=Prague --until=Prague --evm-bin=/data/code-data/go-ethereum/evm -m "benchmark and blockchain_test" --block-gas-limit 1000000
```
I notice that in your command that you use `--gas-benchmark-values`. 

I think that explains a lot of things, since with `--block-gas-limit` looks like the expected gas check isn't done when filling (not sure why?), but it does with `--gas-benchmark-values`. I was never understanding why I didn't get these filling failures, since I was always filling locally for both forks and many gas targets. I think `--block-gas-limit` is a flag that we used before and was changed later.

Thanks for sharing this btw! I think I couldn't have figured out this was a wrong flag usage.

LouisTsai-Csie:
No worries! My responsibility to spend more time on the benchmark docs and possibly organize a _Benchmark 101_ series as suggested during last ACDE call.

To generate a single genesis file for all benchmark tests, we use the `--gas-benchmark-values` flag instead. For example:

```python
--gas-benchmark-values 1,2,3
```
This will produce three benchmark tests with gas limits of 1M, 2M, and 3M respectively, while generating only one genesis file. The block gas limit would be configured to giga gas in the genesis, while the gas-benchmark-values controls the gas used in each block.

LouisTsai-Csie (approved):
LGTM!! Thanks for the contribution!

PR #1915, opened 2025-12-12
feat(benchmarks): fix benchmark for warm SSTORE/SLOAD for Osaka

jsign:
## 🗒️ Description
This PR fixes the `test_storage_access_warm` benchmark to work in Osaka.

For 20M gas benchmark value:
- `test_storage.py::test_storage_access_warm[fork_Osaka-blockchain_test-SLOAD]`: Osaka 310M cycles. Prague 311M cycles
- `test_storage.py::test_storage_access_warm[fork_Osaka-blockchain_test-SSTORE new value]`: Osaka 908M cycles. Prague 909M cycles.
- `test_storage.py::test_storage_access_warm[fork_Osaka-blockchain_test-SSTORE same value]`: Osaka 636M cycles. Prague 637M cycles.

Pretty brutal for only being 20M gas.

## 🔗 Related Issues or PRs

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/benchmark/compute/instruction/test_storage.py
+++ b/tests/benchmark/compute/instruction/test_storage.py
@@ -9,6 +9,7 @@ Supported Opcodes:
 """
 
 import math
+
 import pytest
 from execution_testing import (
     Alloc,
@@ -356,7 +357,9 @@ def test_storage_access_warm(
         num_exec_txs = math.ceil(gas_benchmark_value / tx_max_gas_limit)
         txs = []
         for i in range(num_exec_txs):
-            tx_gas_limit = min(tx_max_gas_limit, gas_benchmark_value - i * tx_max_gas_limit) 
+            tx_gas_limit = min(
+                tx_max_gas_limit, gas_benchmark_value - i * tx_max_gas_limit
+            )
             op_tx = Transaction(
                 to=contract_address,
                 gas_limit=tx_gas_limit,
--- a/tests/benchmark/compute/instruction/test_storage.py
+++ b/tests/benchmark/compute/instruction/test_storage.py
@@ -302,6 +302,7 @@ def test_storage_access_warm(
     fork: Fork,
     gas_benchmark_value: int,
     env: Environment,
+    tx_gas_limit: int,
 ) -> None:
     """
     Benchmark warm storage slot accesses.
@@ -335,12 +336,11 @@ def test_storage_access_warm(
         + Op.RETURN(0, Op.MSIZE)
     )
 
-    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
     with TestPhaseManager.setup():
         sender_addr = pre.fund_eoa()
         setup_tx = Transaction(
             to=None,
-            gas_limit=tx_gas_limit_cap or env.gas_limit,
+            gas_limit=tx_gas_limit,
             data=creation_code,
             sender=sender_addr,
         )
@@ -349,20 +349,15 @@ def test_storage_access_warm(
     contract_address = compute_create_address(address=sender_addr, nonce=0)
 
     with TestPhaseManager.execution():
-        tx_max_gas_limit = (
-            min(gas_benchmark_value, tx_gas_limit_cap)
-            if tx_gas_limit_cap is not None
-            else gas_benchmark_value
-        )
-        num_exec_txs = math.ceil(gas_benchmark_value / tx_max_gas_limit)
+        num_exec_txs = math.ceil(gas_benchmark_value / tx_gas_limit)
         txs = []
         for i in range(num_exec_txs):
-            tx_gas_limit = min(
-                tx_max_gas_limit, gas_benchmark_value - i * tx_max_gas_limit
+            gas_limit = min(
+                tx_gas_limit, gas_benchmark_value - i * tx_gas_limit
             )
             op_tx = Transaction(
                 to=contract_address,
-                gas_limit=tx_gas_limit,
+                gas_limit=gas_limit,
                 sender=pre.fund_eoa(),
             )
             txs.append(op_tx)
--- a/tests/benchmark/compute/instruction/test_storage.py
+++ b/tests/benchmark/compute/instruction/test_storage.py
@@ -8,6 +8,7 @@ Supported Opcodes:
 - TSTORE
 """
 
+import math
 import pytest
 from execution_testing import (
     Alloc,
@@ -297,6 +298,7 @@ def test_storage_access_warm(
     benchmark_test: BenchmarkTestFiller,
     pre: Alloc,
     storage_action: StorageAction,
+    fork: Fork,
     gas_benchmark_value: int,
     env: Environment,
 ) -> None:
@@ -305,25 +307,24 @@ def test_storage_access_warm(
     """
     blocks = []
 
-    # The target storage slot for the warm access is storage slot 0.
-    storage_slot_initial_value = 10
+    # The warm access is done in storage slot 0.
 
     # Contract code
     execution_code_body = Bytecode()
     if storage_action == StorageAction.WRITE_SAME_VALUE:
         execution_code_body = Op.SSTORE(0, Op.DUP1)
     elif storage_action == StorageAction.WRITE_NEW_VALUE:
-        execution_code_body = Op.PUSH1(1) + Op.ADD + Op.SSTORE(0, Op.DUP1)
+        execution_code_body = Op.SSTORE(0, Op.GAS)
     elif storage_action == StorageAction.READ:
         execution_code_body = Op.POP(Op.SLOAD(0))
 
-    execution_code = Op.PUSH1(storage_slot_initial_value) + While(
+    execution_code = Op.SLOAD(0) + While(
         body=execution_code_body,
     )
     execution_code_address = pre.deploy_contract(code=execution_code)
 
     creation_code = (
-        Op.SSTORE(0, storage_slot_initial_value)
+        Op.SSTORE(0, 42)
         + Op.EXTCODECOPY(
             address=execution_code_address,
             dest_offset=0,
@@ -333,11 +334,12 @@ def test_storage_access_warm(
         + Op.RETURN(0, Op.MSIZE)
     )
 
+    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
     with TestPhaseManager.setup():
         sender_addr = pre.fund_eoa()
         setup_tx = Transaction(
             to=None,
-            gas_limit=env.gas_limit,
+            gas_limit=tx_gas_limit_cap or env.gas_limit,
             data=creation_code,
             sender=sender_addr,
         )
@@ -346,11 +348,21 @@ def test_storage_access_warm(
     contract_address = compute_create_address(address=sender_addr, nonce=0)
 
     with TestPhaseManager.execution():
-        op_tx = Transaction(
-            to=contract_address,
-            gas_limit=gas_benchmark_value,
-            sender=pre.fund_eoa(),
+        tx_max_gas_limit = (
+            min(gas_benchmark_value, tx_gas_limit_cap)
+            if tx_gas_limit_cap is not None
+            else gas_benchmark_value
         )
-        blocks.append(Block(txs=[op_tx]))
+        num_exec_txs = math.ceil(gas_benchmark_value / tx_max_gas_limit)
+        txs = []
+        for i in range(num_exec_txs):
+            tx_gas_limit = min(tx_max_gas_limit, gas_benchmark_value - i * tx_max_gas_limit) 
+            op_tx = Transaction(
+                to=contract_address,
+                gas_limit=tx_gas_limit,
+                sender=pre.fund_eoa(),
+            )
+            txs.append(op_tx)
+        blocks.append(Block(txs=txs))
 
     benchmark_test(blocks=blocks)


tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -305,25 +307,24 @@ def test_storage_access_warm(
...
-    storage_slot_initial_value = 10
+    # The warm access is done in storage slot 0.
 
     # Contract code
     execution_code_body = Bytecode()
     if storage_action == StorageAction.WRITE_SAME_VALUE:
         execution_code_body = Op.SSTORE(0, Op.DUP1)
     elif storage_action == StorageAction.WRITE_NEW_VALUE:
-        execution_code_body = Op.PUSH1(1) + Op.ADD + Op.SSTORE(0, Op.DUP1)
+        execution_code_body = Op.SSTORE(0, Op.GAS)
```
jsign, line 319: Feels to me this is a better idea to get different values.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -305,25 +307,24 @@ def test_storage_access_warm(
...
     elif storage_action == StorageAction.READ:
         execution_code_body = Op.POP(Op.SLOAD(0))
 
-    execution_code = Op.PUSH1(storage_slot_initial_value) + While(
+    execution_code = Op.SLOAD(0) + While(
```
jsign, line 323: We always have to pay a cold access, so I think this looks cleaner.
This is only relevant for the `StorageAction.WRITE_SAME_VALUE` case, although for `StorageAction.READ` is kind of adding an extra loop "manually".

Since I change to `Op.GAS` for `WRITE_NEW_VALUE`, feels keeping the `PUSH1(init_value)` isn't adding much -- but happy to change if anyone has a strong opinion.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -297,6 +298,7 @@ def test_storage_access_warm(
     benchmark_test: BenchmarkTestFiller,
```
jsign, line 299: General note: I'm fixing this test just to not lose it -- but I think this SLOAD/SSTORE ideally should also target a deep branch.

Althought for warm access no EL client should be accessing the tree, so maybe is safe not to get into that can of worm setup for this case (and only do it in the other benchmark of cold accesses).

tests/benchmark/compute/instruction/test_storage.py
jsign: There is something else to potentially consider here, and is making these values different between transactions.

For that I could receive some seed via `CALLDATALOAD(0)` and do some `ADD(CALLDATALOAD(0), Op.GAS)` or something like that.

Doubt it would make a difference for the EL -- but if somebody has strong opinion, I can do something like that.

jsign:
cc @jochem-brouwer if you want to take a look

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: > Doubt it would make a difference for the EL

I’m not entirely sure, but I think we could add this. It could make the test more flexible.

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: I support the current approach!

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: @jochem-brouwer Could you help review this scenario to see if we’re missing anything? CPerezz added PR #1885, which targets the worst-case state access by maximizing trie traversal. Not sure if we want to add similar cases here.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -333,11 +335,12 @@ def test_storage_access_warm(
...
+    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
     with TestPhaseManager.setup():
         sender_addr = pre.fund_eoa()
         setup_tx = Transaction(
             to=None,
-            gas_limit=env.gas_limit,
+            gas_limit=tx_gas_limit_cap or env.gas_limit,
             data=creation_code,
             sender=sender_addr,
         )
```
LouisTsai-Csie:
```suggestion
    with TestPhaseManager.setup():
        sender_addr = pre.fund_eoa()
        setup_tx = Transaction(
            to=None,
            gas_limit=tx_gas_limit,
            data=creation_code,
            sender=sender_addr,
        )
```

We should avoid using `env.gas_limit` in **any** benchmark tests. It is intended to provide a single genesis configuration for clients, but it does not work in `execute remote` mode.

In the benchmark setup, `env.gas_limit` is configured to `1_000_000_000_000` (see link below). If the live network being used does not support such a high block gas limit, transactions will break with "intrinsic gas too low" error, thus failing the test.

https://github.com/ethereum/execution-specs/blob/2a2d0d5ed92d53fa64e87c69fd0e6537974d950c/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py#L154-L166

To solve this, we add a fixture that determines the maximum allowable gas limit per transaction. This fixture:
- Uses the transaction gas limit cap when one is enforced, and
- Falls back to the block gas limit (`gas_benchmark_values`) when no such cap exists.

https://github.com/ethereum/execution-specs/blob/2a2d0d5ed92d53fa64e87c69fd0e6537974d950c/tests/benchmark/conftest.py#L90-L93

Example Usage:

https://github.com/ethereum/execution-specs/blob/2a2d0d5ed92d53fa64e87c69fd0e6537974d950c/tests/benchmark/compute/instruction/test_keccak.py#L21-L26

This approach makes the benchmarks compatible with `execute remote` mode and avoids assumptions about unrealistically high block gas limits.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -346,11 +349,23 @@ def test_storage_access_warm(
...
     with TestPhaseManager.execution():
-        op_tx = Transaction(
-            to=contract_address,
-            gas_limit=gas_benchmark_value,
-            sender=pre.fund_eoa(),
+        tx_max_gas_limit = (
+            min(gas_benchmark_value, tx_gas_limit_cap)
+            if tx_gas_limit_cap is not None
+            else gas_benchmark_value
         )
```
LouisTsai-Csie: `tx_max_gas_limit` is equivalent to `tx_gas_limit`.

https://github.com/ethereum/execution-specs/blob/2a2d0d5ed92d53fa64e87c69fd0e6537974d950c/tests/benchmark/conftest.py#L90-L93

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -346,11 +349,23 @@ def test_storage_access_warm(
...
+            tx_gas_limit = min(
+                tx_max_gas_limit, gas_benchmark_value - i * tx_max_gas_limit
+            )
+            op_tx = Transaction(
+                to=contract_address,
+                gas_limit=tx_gas_limit,
+                sender=pre.fund_eoa(),
+            )
+            txs.append(op_tx)
+        blocks.append(Block(txs=txs))
```
LouisTsai-Csie:
```suggestion
        num_exec_txs = math.ceil(gas_benchmark_value / tx_gas_limit)
        txs = []
        for i in range(num_exec_txs):
            gas_limit = min(
                tx_max_gas_limit, gas_benchmark_value - i * tx_gas_limit
            )
            op_tx = Transaction(
                to=contract_address,
                gas_limit=gas_limit,
                sender=pre.fund_eoa(),
            )
            txs.append(op_tx)
        blocks.append(Block(txs=txs))
```

According to the previous comments

LouisTsai-Csie (changes requested):
Thanks for this PR! Leave some comments about the usage of `tx_gas_limit`.

https://github.com/ethereum/execution-specs/blob/2a2d0d5ed92d53fa64e87c69fd0e6537974d950c/tests/benchmark/conftest.py#L90-L93

One question for this test (`test_storage_access_warm`), I wonder if we should take care of the OOG issue for this test? It seems these benchmark test transaction would OOG, and my understanding is that the storage would revert due to OOG, will this affect the intention of this test?

Since this test  (`test_storage_access_cold`) do consider this scenario.

jsign:
> Leave some comments about the usage of tx_gas_limit.

Fixed! I'll also do this change in other open PRs.


> One question for this test (`test_storage_access_warm`), I wonder if we should take care of the OOG issue for this test? It seems these benchmark test transaction would OOG, and my understanding is that the storage would revert due to OOG, will this affect the intention of this test?
> 
> Since this test (`test_storage_access_cold`) do consider this scenario.

I think the non-OOG case is of main relevance for this warm flavor.. In this test we exhaust the gas limit in a single storage slot on warm accesses, so the potential post-state-root calculation would only imply a single storage slot. For the cold case, I added OOG and non-OOG since at a minimum non-OOG is quite relevant, since we target as many storage slots as possible, so those are many writes that will triger post-state root calculation.

tests/benchmark/compute/instruction/test_storage.py
jsign: Great, I think I wasn't aware of this helper value. I'll change it here, and also in other open PRs I have.

tests/benchmark/compute/instruction/test_storage.py
jsign: Note that as I mentioned in my comment, in these warm tests only one storage slot is accessed so the branch length isn't very relevant.

> CPerezz added PR https://github.com/ethereum/execution-specs/pull/1885, which targets the worst-case state access by maximizing trie traversal. Not sure if we want to add similar cases here.

Interesting. I think if this can be ported to be a normal benchmark instead of `test_storage_access_cold`, it would be very good. Can that be ported?

tests/benchmark/compute/instruction/test_storage.py
jsign: @LouisTsai-Csie, @jochem-brouwer, what do you guys thing if on each "attack tx", I can do something like:
```
op_tx = Transaction(
   to=contract_address,
   gas_limit=gas_limit - i,
   sender=pre.fund_eoa(),
)
```
Substracting `i` for each attack tx would make `Op.GAS` return a different set of values for each (basically offseting each by the "tx index").

I don't think substrating `i` changes things much since the tx gas cap is 17M, so even for an attack in 150M block gas limit, that's substracting in the worst case 8 gas (out of 150M) in the worst-case attack tx.

What I like about this idea is, apart from its simplicity, we also don't have to do things like `Op.ADD(op.CALLDATALOAD(..), Op.GAS)` or similar, which eats plenty of gas considering this is a "warm case" scenario (which costs 100)... so basically trying to do anything slightly fancy here introduces a lot of noise for the case we are trying to target (e.g. compared to other cases like cold access where any glue opcodes are small compared to the 2600-ish main operation).

LouisTsai-Csie: I like this idea a lot, and i'd like to check if i could do such refactoring to the rest of the tests!

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: It might be a bit complicated, since it requires some pre-mined address & smart contract. But we could try it out after PR #1885 has a stable version.

LouisTsai-Csie (approved):
We have recently bumped our default branch from `forks/osaka` to `forks/amsterdam`, could you please help update your base branch please. Otherwise, i have no other requested changes! Really appreciate the help, thanks a lot

tests/benchmark/compute/instruction/test_storage.py
jsign: I would say lets do this in a separate PR. I think it is better to avoid blocking PRs that fix benchmarks for Osaka for lateral improvements, because it would be good that we can make a release for Osaka benchmarks as soon as possible.

jsign:
@LouisTsai-Csie, all done.

LouisTsai-Csie:
Thanks a lot!! i will merge after `forks/amsterdam` recreated.

tests/benchmark/compute/instruction/test_storage.py
jochem-brouwer: For the benchmarks which specifically target the trie structure, I think it is not the place for here in the general benchmarks. The benchmarks which also take into account the trie structure might run either on predetermined tries, or just on a "size" trie (small contract, large contract). 
So as a "general benchmark" I think we should not add it here. I'll check out the linked PR also.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -297,33 +299,34 @@ def test_storage_access_warm(
...
     blocks = []
 
-    # The target storage slot for the warm access is storage slot 0.
-    storage_slot_initial_value = 10
+    # The warm access is done in storage slot 0.
 
     # Contract code
     execution_code_body = Bytecode()
     if storage_action == StorageAction.WRITE_SAME_VALUE:
         execution_code_body = Op.SSTORE(0, Op.DUP1)
```
jochem-brouwer, line 317: This will keep writing 0 to the slot (it will load from SLOAD(0) the value to write here)
I think we should init the storage of this contract to a nonzero value (or parametrize it to write  both 0s and nonzeros)

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -297,33 +299,34 @@ def test_storage_access_warm(
...
     elif storage_action == StorageAction.WRITE_NEW_VALUE:
-        execution_code_body = Op.PUSH1(1) + Op.ADD + Op.SSTORE(0, Op.DUP1)
+        execution_code_body = Op.SSTORE(0, Op.GAS)
     elif storage_action == StorageAction.READ:
         execution_code_body = Op.POP(Op.SLOAD(0))
 
-    execution_code = Op.PUSH1(storage_slot_initial_value) + While(
+    execution_code = Op.SLOAD(0) + While(
         body=execution_code_body,
```
jochem-brouwer, line 324: Hmm just noticing, this while loop has no exit condition. I don't think it matters here (since we only target one slot) but somewhat surprising that this tx thus always fails.

tests/benchmark/compute/instruction/test_storage.py
jochem-brouwer: I love this idea!!

jochem-brouwer (commented):
Minor comments!

tests/benchmark/compute/instruction/test_storage.py
jsign: The storage slot 0 is initialized to value 42, no? 
```
   creation_code = (
        Op.SSTORE(0, 42) <------
        + Op.EXTCODECOPY(
            address=execution_code_address,
            dest_offset=0,
```

tests/benchmark/compute/instruction/test_storage.py
jsign: Yeah, I think this is fine since given we only target a singe storage slot (since this is warm access flooding), should be okay.

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: @jsign Yes for sure, i mean i will review existing PR and make an issue for the refactor.

LouisTsai-Csie:
LGTM

PR #1917, opened 2025-12-13
feat(specs): EIP-7928 move bal from payload

nerolation:
🗒️ Description

This PR completes the EIP-7928 block access list (BAL) refactoring by moving BAL validation outside of the state transition function. 

The changes across the last two commits:
  - Previous commit: Moved BAL from the block body to be passed separately to state_transition
  - The second commit: Removed BAL parameter entirely from state_transition, keeping only hash validation

The state transition now:
  - Generates the BAL during block execution via apply_body()
  - Computes the hash of the generated BAL
  - Validates only that this hash matches block.header.block_access_list_hash

This simplifies the state transition by removing the need to pass and validate the full BAL, since the actual BAL validation (checking p2p received BAL matches the hash) is handled at the engine API layer.

  🔗 Related Issues or PRs

  N/A.

Alt text

### DIFF
--- a/src/ethereum/forks/amsterdam/blocks.py
+++ b/src/ethereum/forks/amsterdam/blocks.py
@@ -19,7 +19,6 @@ from ethereum_types.numeric import U64, U256, Uint
 
 from ethereum.crypto.hash import Hash32
 
-from .block_access_lists.rlp_types import BlockAccessList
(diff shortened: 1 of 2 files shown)


qu0b (approved):

fselmo (approved):
This lgtm, thank you!

PR #1921, opened 2025-12-15
use uint256 for storage key / values in RLP encoding

qu0b:
missing uint256 in RLP encoding: https://discord.com/channels/595666850260713488/1364000387195076608/1450030511841087671

Gemini_Generated_Image_7jgumy7jgumy7jgu


fselmo (commented):
I don't think this is necessary. Let me know if I'm missing something 👀

PR #1922, opened 2025-12-15
fix(test-tests): Use `ZeroPaddedHexNumber` instead of `HexNumber` for BALs

fselmo:
## 🗒️ Description

Use `ZeroPaddedHexNumber` for consistency with other fields in test vectors (e.g. `0x0c` instead of `0xc`)

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2025-12-15 at 10 55 30

### DIFF
--- a/packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
+++ b/packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
@@ -13,8 +13,8 @@ from execution_testing.base_types import (
     Address,
     Bytes,
     CamelModel,
-    HexNumber,
     RLPSerializable,
+    ZeroPaddedHexNumber,
 )
 
 
@@ -23,11 +23,11 @@ class BalNonceChange(CamelModel, RLPSerializable):
 
     model_config = CamelModel.model_config | {"extra": "forbid"}
 
-    block_access_index: HexNumber = Field(
-        HexNumber(1),
+    block_access_index: ZeroPaddedHexNumber = Field(
+        ZeroPaddedHexNumber(1),
         description="Transaction index where the change occurred",
     )
-    post_nonce: HexNumber = Field(
+    post_nonce: ZeroPaddedHexNumber = Field(
         ..., description="Nonce value after the transaction"
     )
 
@@ -39,11 +39,11 @@ class BalBalanceChange(CamelModel, RLPSerializable):
 
     model_config = CamelModel.model_config | {"extra": "forbid"}
 
-    block_access_index: HexNumber = Field(
-        HexNumber(1),
+    block_access_index: ZeroPaddedHexNumber = Field(
+        ZeroPaddedHexNumber(1),
         description="Transaction index where the change occurred",
     )
-    post_balance: HexNumber = Field(
+    post_balance: ZeroPaddedHexNumber = Field(
         ..., description="Balance after the transaction"
     )
 
@@ -55,8 +55,8 @@ class BalCodeChange(CamelModel, RLPSerializable):
 
     model_config = CamelModel.model_config | {"extra": "forbid"}
 
-    block_access_index: HexNumber = Field(
-        HexNumber(1),
+    block_access_index: ZeroPaddedHexNumber = Field(
+        ZeroPaddedHexNumber(1),
         description="Transaction index where the change occurred",
     )
     new_code: Bytes = Field(..., description="New code bytes")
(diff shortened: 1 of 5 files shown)


packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
```diff
@@ -23,11 +23,11 @@ class BalNonceChange(CamelModel, RLPSerializable):
 
     model_config = CamelModel.model_config | {"extra": "forbid"}
 
-    block_access_index: HexNumber = Field(
-        HexNumber(1),
+    block_access_index: ZeroPaddedHexNumber = Field(
+        ZeroPaddedHexNumber(1),
```
danceratopz, line 27: I don't have a strong opinion on this, but we could be less verbose here and simply write "1". Pydantic will coerce the type to `ZeroPaddedHexNumber`.

packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
```diff
@@ -111,7 +111,7 @@ class BalAccountChange(CamelModel, RLPSerializable):
     storage_changes: List[BalStorageSlot] = Field(
         default_factory=list, description="List of storage changes"
     )
-    storage_reads: List[HexNumber] = Field(
+    storage_reads: List[ZeroPaddedHexNumber] = Field(
```
danceratopz, line 114: I think `StorageKey` would be better (requires import!):

```suggestion
    storage_reads: List[StorageKey] = Field(
```
and consistent with:
 
- `BalAccountExpectation` - https://github.com/fselmo/execution-specs/blob/ab17e0642769c0c9bc4b3a4a34ff21dfa09e7d3d/packages/testing/src/execution_testing/test_types/block_access_list/expectations.py#L48-L50
- `BalAccountAbsentValues` - https://github.com/fselmo/execution-specs/blob/ab17e0642769c0c9bc4b3a4a34ff21dfa09e7d3d/packages/testing/src/execution_testing/test_types/block_access_list/account_absent_values.py#L94-L98

packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
```diff
@@ -13,8 +13,8 @@
     Address,
     Bytes,
     CamelModel,
-    HexNumber,
     RLPSerializable,
```
danceratopz, line 16:
```suggestion
    RLPSerializable,
    StorageKey,
```

danceratopz (commented):
Nice, this def looks like it addresses the issue at hand.

Added a dedicated BALs serialization roundtrip test that checks that values are zero-padded correctly here:
- https://github.com/fselmo/execution-specs/pull/33

Did you intentionally leave `HexNumber` in `test_block_access_list_t8n.py`? I think we should update them there, too:

https://github.com/fselmo/execution-specs/blob/ab17e0642769c0c9bc4b3a4a34ff21dfa09e7d3d/packages/testing/src/execution_testing/test_types/tests/test_block_access_list_t8n.py

Otherwise, a couple of small comments below.

packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
fselmo: `StorageKey` is exactly what we just moved out of (`Bytes32`) [here](https://github.com/ethereum/execution-specs/pull/1912/changes#diff-e154b39cdaa261fe7e49a21c81aff37f728ddfa19ae8d97fb9a0580125030df1L89-R88) and instead opted to store storage keys as `Uint256` (hence `ZeroPaddedHexNumber`). It is purposeful to _not_ have this as Bytes32.

packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
fselmo: This will fail `mypy` check, which is why I added it here :/

packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
fselmo: See comment about not wanting this as `Bytes32`

fselmo:
Thanks @danceratopz, can you run back through and make sure your comments are addressed and re-review? 🙏🏼

packages/testing/src/execution_testing/test_types/block_access_list/account_changes.py
fselmo: Here is the recent update to the EIP: https://github.com/ethereum/EIPs/commit/6c6b1e72412460d9888e47f3549b2c926cf13b19

danceratopz: Aaah, thanks @fselmo.

danceratopz (approved):
> Thanks @danceratopz, can you run back through and make sure your comments are addressed and re-review? 🙏🏼

There was just this remaining consistency nit:
> Did you intentionally leave HexNumber in test_block_access_list_t8n.py? I think we should update them there, too:
> https://github.com/fselmo/execution-specs/blob/ab17e0642769c0c9bc4b3a4a34ff21dfa09e7d3d/packages/testing/src/execution_testing/test_types/tests/test_block_access_list_t8n.py

But approving, as I don't want to block this PR any longer!

PR #1934, opened 2025-12-16
feat(test-execute): Implement `pre.deterministic_deploy_contract`

### LINKED ISSUE #1933: feat(test-cli-execute): Allow use of `address` during `pre.deploy_contract`
Some zkevm and stateless tests need a very specific address which are mined using https://github.com/CPerezz/worst_case_miner.

In this case, it would be possible to hard-code the contract in the pre at the given address for fill, but currently for `execute` if we pass an address like `pre.deploy_contract(address=0x1234...)` execute will complain because it cannot guarantee that the contract is going to end up in that address.

Specifically this line forbids it:
https://github.com/ethereum/execution-specs/blob/dca59f62f16089e6a3256979b9d218976570140f/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py#L273

But perhaps what we can do in this case is check the address in the current chain and check if the code matches what we were intending to deploy, and if it does, proceed with the test.

marioevz:
### `pre.deterministic_deploy_contract`

Implements a new function in the `pre` object called `deterministic_deploy_contract`, which uses https://github.com/Arachnid/deterministic-deployment-proxy to deploy contracts at deterministic addresses using `CREATE2`.

The function takes the following parameters:
- deploy_code: The bytecode to be deployed. Will be deployed at `keccak256( 0xff ++ 0x4E59B44847B379578588920CA78FBF26C0B4956C ++ salt ++ keccak256(init_code))[12:]`
- salt: Salt to be passed to the `CREATE2` opcode. Defaults to zero.
- initcode: Initcode that deploys the code. It will not be checked in any way against the value passed in `deploy_code` until the transactions is sent to the network using `execute`. Can be skipped and the initcode will be auto-generated.
- setup_calldata: If specified, a transaction will be sent to the contract after deployment with this calldata. Can be used to reset storage values before the test starts.

### [EIP-7997](https://eips.ethereum.org/EIPS/eip-7997)

EIP-7997 includes a predeploy that is meant to replace contracts like the one in `0x4E59B44847B379578588920CA78FBF26C0B4956C`.
To account for this and to be future compatible with the EIP, all added functions check the new `fork.deterministic_factory_predeploy_address()` method to check whether the currently active fork already has a deployment factory before deploying/using `0x4E59B44847B379578588920CA78FBF26C0B4956C`.

### `execute deploy-required-contracts`

New execute command that deploys all execute pre-requisite contracts to the given chain.

At the moment it only deploys `0x4E59B44847B379578588920CA78FBF26C0B4956C`, but we will add more useful contracts in the future.

## 🔗 Related Issues or PRs
Closes #1933

## ✅ Checklist
    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


marioevz:
One open question is that, in execute, we will not be able to run two different tests that use the same contract without causing issues.
One potential solution could be to use a lock for these contracts, but it could lead to tests dead-locking themselves.

marioevz:
Example test: https://github.com/ethereum/execution-specs/blob/dd68f48bd0400d35b0bbaf00fd8be016f523edf2/tests/constantinople/eip1014_create2/test_deterministic_deployment.py

I've verified this locally using fill and kurtosis, the test can be run over and over and the deployment only happens once.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -45,6 +46,16 @@
...
+DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS = Address(
+    0x4E59B44847B379578588920CA78FBF26C0B4956C
+)
+DETERMINISTIC_DEPLOYMENT_SIGNER_ADDRESS = Address(
+    0x3FAB184622DC19B6109349B94811493BF2A45362
+)
+DETERMINISTIC_DEPLOYMENT_TRANSACTION = Bytes(
+    "0xf8a58085174876e800830186a08080b853604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf31ba02222222222222222222222222222222222222222222222222222222222222222a02222222222222222222222222222222222222222222222222222222222222222"
+)
+
```
CPerezz: What are these hardcoded things? Why do we have a `DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS`?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -255,6 +267,176 @@ def code_pre_processor(
...
+                    action="deterministic_deploy_contract",
+                    target=f"{DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS}",
+                    ty=0,
+                    protected=False,
+                    nonce=0,
+                    gas_price=0x174876E800,
+                    gas_limit=0x0186A0,
+                    to=None,
+                    value=0,
+                    data=Bytes(
```
CPerezz: Why do we hardcode gas_price here to 0x174876E800 (100 Gwei) ?

CPerezz (changes requested):
Important things also:

- Missing max code size validation in execute mode. The filler implementation checks max_code_size (line 176), but execute mode doesn't. This should be patched.
- DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS is defined multiple times.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
CPerezz: Nevermind. I assume we will set Nick's deployer unless we provide another one? If that is the case, we can close this.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: The transaction that deploys the proxy must be exactly like this in order to deploy to the expected address: https://github.com/Arachnid/deterministic-deployment-proxy?tab=readme-ov-file#proxy-address. This is just how Nick's method works.

Could you confirm that your miner uses `0x4e59b44847b379578588920ca78fbf26c0b4956c` please? I used this as it was the one that was linked in @jochem-brouwer message in the original issue: https://github.com/ethereum/execution-specs/issues/1933#issuecomment-3661560912

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
LouisTsai-Csie: IIUC, these info comes from Nick’s [deterministic deployer](https://etherscan.io/address/0x4e59b44847b379578588920ca78fbf26c0b4956c):

- `DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS`: the contract address of Nick’s deployer (the factory address).
- `DETERMINISTIC_DEPLOYMENT_SIGNER_ADDRESS`: the address recovered via `ecrecover` from the randomly generated `(v, r, s)` values.
- `DETERMINISTIC_DEPLOYMENT_TRANSACTION`: the transaction used to deploy Nick’s factory contract.

The information documented [here](https://github.com/Arachnid/deterministic-deployment-proxy?tab=readme-ov-file#latest-outputs).

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
CPerezz: Yes yes. That's Nick's deployer. I was asking about gas price and gas_limit hardcoded here rather than anything else.

The address won';t be affected by that right?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -142,6 +149,61 @@ def code_pre_processor(
...
+
+        super().__setitem__(
+            contract_address,
+            Account(
+                nonce=1,
+                balance=minimum_balance,
+                code=deploy_code,
+                storage={},
+            ),
+        )
```
LouisTsai-Csie, line 212: Some feedback:

In `execute remote`, we will first check if the contract exists or not, i believe we should do the similar verification here, or we might override the existing contract if we do not have proper salt management. Moreover, in post-state object, it would not raise error when there is duplicate entry.

```python
def test_collision_detection(
    pre: Alloc,
) -> None:
    """
    Test collision detection for deterministic deployments.
    """
    deploy_code = Op.SSTORE(1, Op.CALLDATALOAD(0))

    addr_1 = pre.deterministic_deploy_contract(
        deploy_code=deploy_code,
        setup_calldata=Hash(0),
        label="contract_1",
    )
    
    # This would override the addr_1 account
    addr_2 = pre.deterministic_deploy_contract(
        deploy_code=deploy_code,
        setup_calldata=Hash(1),
        label="contract_2",
    )
    
    # If we are not aware the `addr_1` and `addr_2` are the same, the test might not behave as intended.
    assert addr_1 == addr_2, "Contract address is the same"

    sender = pre.fund_eoa()

    tx_1 = Transaction(
        sender=sender,
        to=addr_1,
        data=Hash(1),
        gas_limit=100_000,
    )

    tx_2 = Transaction(
        sender=sender,
        to=addr_2,
        data=Hash(2),
        gas_limit=100_000,
    )

    post = {
       # this is overriden by the `addr_2` account setting, as `addr_1` is the same as `addr_2`.
        addr_1: Account(
            deploy_code=deploy_code,
            storage={
                1: 1,
            },
        ),
        addr_2: Account(
            deploy_code=deploy_code,
            storage={
                1: 2,
            },
        ),
    }

    blockchain_test(pre=pre, post=post, blocks=[Block(txs=[tx_1, tx_2])])

```

LouisTsai-Csie: Moreover, the `setup_calldata` mechanism is not working for `fill` but `execute remote`:

Below is an example:

```python
def test_missing_setup_phase(
    pre: Alloc,
) -> None:
    """
    Test collision detection for deterministic deployments.
    """
    deploy_code = Op.SSTORE(1, Op.CALLDATALOAD(0))

    addr_1 = pre.deterministic_deploy_contract(
        deploy_code=deploy_code,
        salt=Hash(1),
        setup_calldata=Hash(0),
        label="contract_1",
    )

    addr_2 = pre.deterministic_deploy_contract(
        deploy_code=deploy_code,
        salt=Hash(2),
        setup_calldata=Hash(1),
        label="contract_2",
    )

    assert addr_1 != addr_2, "Contract addresses should be different"

    sender = pre.fund_eoa()

    tx_1 = Transaction(
        sender=sender,
        to=addr_1,
        data=Hash(2),
        gas_limit=100_000,
    )

    post = {
        addr_1: Account(
            deploy_code=deploy_code,
            storage={
                1: 2,
            },
        ),
        addr_2: Account(
            deploy_code=deploy_code,
            storage={
                1: 1,
            },
        ),
    }

    blockchain_test(pre=pre, post=post, blocks=[Block(txs=[tx_1])])

```

- expected behavior (Works under execute remote)
  - slot 1 in addr_1 increases from 1 to 2
  - slot 1 in addr_2 increases stays 1
- Current behavior (In fill mode)
  - slot 1 in addr_1 from 0 to 2
  - slot 1 in addr_2 stays 0

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -255,6 +267,176 @@ def code_pre_processor(
...
+                gas_costs.G_TRANSACTION + gas_costs.G_TRANSACTION_CREATE
+            )
+            deploy_gas_limit += (
+                len(deploy_code) * gas_costs.G_CODE_DEPOSIT_BYTE
+            )
+            deploy_gas_limit += memory_expansion_gas_calculator(
+                new_bytes=len(initcode)
+            )
+            deploy_gas_limit += calldata_gas_calculator(data=initcode)
+            deploy_gas_limit = min(deploy_gas_limit * 2, 30_000_000)
```
LouisTsai-Csie: This 30M hardcoded limit does not work for Osaka (eip-7825).

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -326,57 +520,44 @@ def deploy_contract(
...
         )
 
-        calldata_gas_calculator = self._fork.calldata_gas_calculator(
-            block_number=0, timestamp=0
-        )
-        deploy_gas_limit += calldata_gas_calculator(data=initcode)
+        deploy_gas_limit += calldata_gas_calculator(data=prepared_initcode)
 
         # Limit the gas limit
         deploy_gas_limit = min(deploy_gas_limit * 2, 30_000_000)
```
LouisTsai-Csie: Same argument

LouisTsai-Csie (dismissed):
Leave some comments! I now have a better understanding of the big picture, will revisit CPerezz' PR to see how to integrate.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Take a look at the original repo, it requires this hardcoded amount in the deployment transaction to produce the expected contract address.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
marioevz: Thanks for pointing this out. I've removed `setup_calldata` and `minimum_balance` since sending these, as extra transactions during execute, would result in unexpected side-effects. The test should handle this by some other way (providing a value in the transaction that interacts with the contract for example).

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Fixed in last commit.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Fixed in last commit.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
```diff
@@ -658,20 +658,20 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
                     if actual_code != expected_code:
-                        raise Exception(
+                        msg = (
                             f"Deployed test contract didn't match expected code at address "
                             f"{deployed_contract} (not enough gas_limit?).\n"
                             f"Expected: {expected_code}\n"
                             f"Actual: {actual_code}"
                         )
+                        logger.error(msg)
+                        raise Exception(msg)
```
fselmo: nit: could we use `LoggedError` here instead?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -651,6 +796,7 @@ def minimum_balance_for_pending_transactions(
                 # never sends a transaction during the test.
                 assert tx.to in sender_balances, (
                     "Sender balance must be set before sending"
```
fselmo: nit: should we add a space here for better separation?

```suggestion
                    "Sender balance must be set before sending "
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -45,6 +46,16 @@
 MAX_BYTECODE_SIZE = 24576
 MAX_INITCODE_SIZE = MAX_BYTECODE_SIZE * 2
 
+DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS = Address(
```
fselmo: nit: Can this be imported from `test_types.helpers` and only defined once instead? Maybe the same for all of these? Or, better yet, maybe we can have a `constants.py` defined somewhere for importing constants across files? Just a nit :)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -45,6 +46,16 @@
...
+    0x4E59B44847B379578588920CA78FBF26C0B4956C
+)
+DETERMINISTIC_DEPLOYMENT_SIGNER_ADDRESS = Address(
+    0x3FAB184622DC19B6109349B94811493BF2A45362
+)
+DETERMINISTIC_DEPLOYMENT_TRANSACTION = Bytes(
```
fselmo: nit: This isn't used anywhere. Is it worth keeping around? Maybe add to the `constants` / `helpers` file mentioned above?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -255,6 +267,155 @@ def code_pre_processor(
...
+                "match the expected code: "
+                f"Expected: {deploy_code}, "
+                f"Current: {chain_code}"
+            )
+        else:
+            # 2) Determine if the deployment contract is already on chain
+            deployment_contract_code = self._eth_rpc.get_code(
+                DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS
+            )
+            if deployment_contract_code == b"":
```
fselmo: Should we add a flag for `proxy_deployed` here if we enter this code block and then pass this to the `deploy_tx` below if we had to create the `deterministic_deploy_contract` first?

```suggestion
          proxy_deployed = False
          if deployment_contract_code == b"":
              proxy_deployed = True
```

Then pass this below:

```py
deploy_tx = self._add_pending_tx(
    action="deterministic_deploy_contract",
    target=label,
    to=DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS,
    data=Bytes(salt) + Bytes(initcode),
    gas_limit=deploy_gas_limit,
    requires_predecessor=proxy_deployed,  # pass it here
)
```

fselmo: I think we also may need to account for multiple contracts being deployed using the proxy, while the first contract will put the proxy deployment into the pending tx's, we shouldn't have each contract do so because this will cause things to fail probably (`code == b""` will still be `True` since it's still pending and would trigger this logic again for each contract).

We could have a private attr in the class, `_proxy_deployment_pending`

```py
# private attr in the class
_proxy_deployment_pending: bool = PrivateAttr(False)
```

revised with the above...

```py
proxy_deployed = False
if deployment_contract_code == b"" and not self._proxy_deployment_pending:
    proxy_deployed = True
    self._proxy_deployment_pending = True
```

This will only deploy the proxy on the first pass of this logic and each subsequent contract deployment would be added to the same batch, which still has the proxy deployment in the "parent" batch to them, so I think this would work well. Thoughts?

---
edit: Yeah, thinking on this I think this would produce a nice result.

assuming the proxy is not deployed:
  - fund proxy tx - `requires_predecessor = False`
  - proxy deploy - `requires_predecessor = True` (new batch)
  - contract_a - `requires_predecessor = proxy_deployed = True` (new batch)
  - contract_b - `requires_predecessor = proxy_deployed = False` (added to the latest batch)

```
batches = [[funding_tx], [proxy], [contract_a, contract_b, ...]]
```

fselmo (commented):
This flow is quite elegant, love this. I left some tiny nits (take or leave them) but also some things to think about when it comes to the proxy deployment.

- We can put the proxy deploy in a parent batch for safety so that at least the first contract that depends on it uses the `requires_predecessor` flag to start a new batch and to make sure the proxy deploy is confirmed first.
- We could have an internal flag that only deploys the proxy once, in the case that a test deploys multiple contracts using this API. They can still be in the same batch, whose parent is the proxy deployment and is only triggered to deploy one time.

I still haven't tested this in `execute` so I will take a pass at that next but it seems like it should work well. Great addition 👍🏼

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Removed! It was added at the start for me as a guidance, but now is not needed.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Added new line and printed the metadata if available. Proved helpful during debugging indeed.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -255,6 +303,136 @@ def code_pre_processor(
...
+        self._pending_txs.append(pending_tx)
+        return pending_tx
+
+    def deterministic_deploy_contract(
+        self,
+        *,
+        deploy_code: BytesConvertible,
+        salt: Hash | int = 0,
+        initcode: BytesConvertible | None = None,
+        storage: Storage | StorageRootType | None = None,
```
fselmo: This looks like it's not used, is this meant to be passed to `set_item` below? 

bonus: we could add some sanity unit tests to catch this

fselmo (approved):
Really cool feature! And really elegant implementation 👌🏼

lgtm 🚀!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -381,6 +377,16 @@ def deterministic_deploy_contract(
             ), "Deployment contract code is not found"
 
             # Deploy the actual contract.
+            max_code_size = self._fork.max_code_size()
```
fselmo, line 380: Ah yeah, good eye. Especially with EIPs that are looking to change these being proposed 👍🏼

CPerezz (approved):
Thanks for adding this!!

PR #1954, opened 2025-12-29
feat(test): OOG and success selfdestruct tests

fselmo:
## 🗒️ Description

Added comprehensive `SELFDESTRUCT` tests covering beneficiaries as EOA, contract, precompiles, and self. All tests consolidated in `TangerineWhistle` with fork-aware parametrization

  - is_success (exact_gas / exact_gas_minus_1) - tests success boundary
  - warm param uses `pytest.param(..., marks=valid_from("Berlin"))` so warm tests only run on `>=Berlin`
  - State access boundary tests (consensus check: beneficiary accessed before `G_NEW_ACCOUNT` evaluated)
  - same_tx vs pre-deployed contracts
  - originator_balance (0/1) - EIP-161 behavior difference
  - beneficiary_initial_balance (0/1) - dead/alive for `G_NEW_ACCOUNT`

Precompile coverage was sparse - some static tests hit 0x01, others 0x01-0x03. Now using `@pytest.mark.with_all_precompiles` for full coverage across all fork precompiles.

TangerineWhistle (EIP-150): All tests live here since this is when gas costs for `SELFDESTRUCT` were introduced. Fork-aware parametrization handles Berlin+ warm/cold distinction.

SpuriousDragon (EIP-161, EIP-155): Tests correctly handle the originator_balance > 0 requirement for G_NEW_ACCOUNT. ~~Also updated old tests to use >= SpuriousDragon for protected transactions since this is when EIP-155 was introduced, not Byzantium as other tests have been using in our codebase.~~ [This was then refactored as `fork.supports_protected_txs`, returning `True` beginning at `SpuriousDragon`].

Berlin (EIP-2929): Warm/cold coverage integrated via warm param with `valid_from("Berlin")` mark - no separate Berlin test file needed.

## 🔗 Related Issues or PRs

- Related to a Nethermind bug in BAL devnet 1, relevant PR [here](https://github.com/NethermindEth/nethermind/pull/10064)

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2025-12-29 at 14 19 44


fselmo:
I think we should move these to an earlier fork as it looks like we may not have coverage for this. I see _some_ coverage in the static tests but not for all opcodes. It would be good to have this in python either way and use the `pytest.mark.with_all_precompiles` to test every added precompile for every fork as they are added.

I'll work on porting this to an earlier fork.

jochem-brouwer:
Does this initialize the precompiles as empty and does it attempt to create the accounts via `SELFDESTRUCT`? 

This is definitely a case which should be added, but just wanted to note here that due to this bug:

image

(from the yellow paper) there is this weird behavior with precompiles, creating the account, the precompile going out-of-gas and then the account cleanup either cleaning up the account (or not). 

This bug caused that later devnets and all live chains just funded the addresses with 1 wei, such that the account is always there (and not the "0 balance, 0 nonce, no code, no storage", but still in trie case, which now is not possible anymore since it has a balance)

I wanted to share this because this might be the reason why the coverage report is low. I think (although this should be an EIP in that case) that the general consensus is to "just fund the precompiles" with 1 wei at genesis to get rid of this behavior, and that it is thus either unspecified or untested behavior.

fselmo:
> Does this initialize the precompiles as empty and does it attempt to create the accounts via SELFDESTRUCT?

`pytest.mark.with_all_precompiles` will parametrize for all precompile addresses that are mapped for the fork you are filling for, but it doesn't add anything to the `pre`. All the EVM knows is that this is the beneficiary for the self-destruct. Is there a missing case? 🤔

jochem-brouwer:
> `pytest.mark.with_all_precompiles` will parametrize for all precompile addresses that are mapped for the fork you are filling for, but it doesn't add anything to the `pre`. All the EVM knows is that this is the beneficiary for the self-destruct. Is there a missing case? 🤔

Ah sorry, I think my comment is confusing. The YP case is, IIRC, a case which is only possible if a precompile account is empty (no balance, 0 nonce, no code, no storage). I am not sure but I thought some tests pre-fill precompiles with a balance to make them non-empty, and wanted to explicitly mention this here. But I think this is already the case. So no missing cases found here.

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -947,7 +947,7 @@ class SpuriousDragon(Tangerine, ignore=True):
     pass
```
jochem-brouwer: Could you edit the description of the Spurious Dragon fork? It lists EIP-158 to be in there but this is not the case (this was a non-shipped variant of EIP-161).

Ref: https://eips.ethereum.org/EIPS/eip-607

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -935,7 +935,7 @@ class DAOFork(Homestead, ignore=True):
...
 
 
-class Tangerine(DAOFork, ignore=True):
+class Tangerine(DAOFork, transition_tool_name="TangerineWhistle"):
```
jochem-brouwer: Why is SpuriousDragon called SpuriousDragon and TangerineWhistle Tangerine?

packages/testing/src/execution_testing/forks/__init__.py
```diff
@@ -97,6 +98,7 @@
     "Frontier",
     "GrayGlacier",
     "Homestead",
+    "Tangerine",
```
jochem-brouwer: If this is the "official" fork name then this should be TangerineWhistle.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -105,11 +105,10 @@
...
-| `test_bal_selfdestruct_to_precompile` | Ensure BAL captures SELFDESTRUCT with precompile as beneficiary | Caller triggers victim contract (balance=100) to execute `SELFDESTRUCT(0x0000...0001)` (ecrecover precompile). Precompile starts with balance=0. | BAL **MUST** include: (1) Contract with `balance_changes` (100→0, loses balance to selfdestruct). (2) Precompile address 0x01 with `balance_changes` (0→100, receives selfdestruct balance). Precompile **MUST NOT** have `code_changes` or `nonce_changes`. This complements `test_bal_withdrawal_to_precompiles` (withdrawal) and `test_bal_precompile_funded` (tx value). | ✅ Completed |
+| `test_bal_selfdestruct_to_precompile_and_oog` | Ensure BAL captures SELFDESTRUCT to precompile at different gas boundaries | Victim executes `SELFDESTRUCT(precompile)`. Parameterized by all precompiles and three scenarios: (1) Success, (2) OOG before state access, (3) OOG after state access. | Success: victim and precompile have `balance_changes`. OOG before state access: precompile **NOT** in BAL. OOG after state access: precompile in BAL with empty changes. | ✅ Completed |
```
jochem-brouwer: (3) is only possible in this case if we send value, right?

jochem-brouwer (commented):
Not a review of the tests, just a small point regarding fork names and docs 😄 👍

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
fselmo: If we have >= `selfdestruct (5000)` cost, we reach into the `accessed_addresses` for `beneficiary` to see if it's alive or not and so we mark it warm. It's a bit of a pedantic case where it may be marked "warm" even though we don't _need_ to touch it (if `0` transfer). But as far as the access list is concerned, whether we touch it in state or not with optimizations, this means it's already warmed at this point and is in BAL. The EIP mentions this specifically in the [Scope and Inclusion](https://eips.ethereum.org/EIPS/eip-7928#scope-and-inclusion) section, under "addresses accessed without state changes, including":

> Beneficiary addresses for SELFDESTRUCT

packages/testing/src/execution_testing/forks/__init__.py
fselmo: Yes I agree. I just turned this on (it was already defined). I'm happy to change it here though as I don't think that would complicate anything since it is currently ignored.

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
jochem-brouwer: So the gas checks are: `base cost (5000)` + `warm/cold access` + `create account`.

Just iterating this for myself but there are thus 4 boundaries:

1. Not enough gas to pay base cost
2. Not enough gas to pay in case target is cold (to warm it up) (If target is warm, this check costs 0)
3. Not enough gas to pay for creating the account
4. Success

So one of the cases is missing here :thinking:

jochem-brouwer: Also linking EIP-2929: the warm/cold check is there even if the value of SELFDESTRUCT is 0 https://eips.ethereum.org/EIPS/eip-2929#selfdestruct-changes

fselmo: Yes! This is missing some cases for sure thank you. I was thinking about this from a BAL perspective as this was a BAL test ported to earlier forks (pre state access and post state access). But since we are missing coverage for this, we should be more thorough on these boundaries in general with these tests. Thanks!

fselmo: I have greatly extended the test cases now. Thanks!

fselmo:
@danceratopz this is ready for review now! Note that `json_infra` and `optimized` CI runs won't pass until the next bal test release is out and we point to it in the CI (blob params from the last release are at Osaka but the repo is now at BPO2).

packages/testing/src/execution_testing/client_clis/transition_tool.py
```diff
@@ -326,13 +327,19 @@ def _evaluate_filesystem(
         }
         output_paths["body"] = os.path.join("output", "txs.rlp")
 
+        # Get fork name and apply any tool-specific mapping
+        fork_name = (
```
fselmo, line 331: This fixed CI against EvmOne btw... for both forks that have a space in between. Even though I re-ignored them, I think this is a nice addition since it would just take turning the ignore on / off to turn these forks on again.

fselmo:
One thing here we'd have to eventually check is I believe this may cover a lot of cases that static tests currently cover. We can maybe, as a separate task, go through and check which tests are now covered here that we can possibly remove from `tests/static/`.

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -0,0 +1,1004 @@
...
+            Op.EXTCODECOPY(initcode_address, 0, 0, initcode_len)
+            + Op.CREATE(value=originator_balance, offset=0, size=initcode_len)
+            + Op.POP
+            + Op.CALL(gas=inner_call_gas, address=victim)
+        )
+        caller = pre.deploy_contract(
+            address=factory_address,
+            code=factory_code,
+            balance=originator_balance,
+        )
```
marioevz:
```suggestion
        factory_code = Op.EXTCODECOPY(
            initcode_address, 0, 0, initcode_len
        ) + Op.CALL(
            gas=inner_call_gas,
            address=Op.CREATE(
                value=originator_balance, offset=0, size=initcode_len
            ),
        )
        caller = pre.deploy_contract(
            code=factory_code,
            balance=originator_balance,
        )
        victim = compute_create_address(address=caller, nonce=1)
```
Should avoid the need to use `pre._contract_address_iterator` since this is not available in `execute`.

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -0,0 +1,1004 @@
...
+    alice = pre.fund_eoa()
+    victim_code = Op.SELFDESTRUCT(beneficiary)
+
+    if same_tx:
+        # Deploy and selfdestruct in same transaction via factory
+        initcode = Initcode(deploy_code=victim_code)
+        initcode_len = len(initcode)
+
+        # Pre-deploy initcode at separate address, then use EXTCODECOPY
+        initcode_address = pre.deploy_contract(initcode)
```
marioevz: `MSTORE` macro should allow us to skip this contract deployment:

```
from execution_testing import Macros as Om
```
```
factory_code = Om.MSTORE(initcode, 0) + ...
```

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -0,0 +1,1004 @@
...
+    [0, 1],
+    ids=["no_balance", "has_balance"],
+)
+@pytest.mark.parametrize(
+    "beneficiary_initial_balance",
+    [0, 1],
+    ids=["dead_beneficiary", "alive_beneficiary"],
+)
+@pytest.mark.valid_from("TangerineWhistle")
+def test_selfdestruct_to_precompile(
```
marioevz, line 583: We could add system contracts as another test, I think they are also always warm iirc.

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -0,0 +1,1004 @@
...
+    same_tx: bool,
+) -> None:
+    """
+    Test SELFDESTRUCT where beneficiary is the executing contract itself.
+
+    Uses Op.SELFDESTRUCT(Op.ADDRESS) - the victim selfdestructs to itself.
+
+    Key characteristics:
+    - Beneficiary is always warm (it's the executing contract)
+    - Beneficiary is always alive (has code)
```
marioevz: I think this does not hold in the case where self is a contract mid-creation (Initcode self-destructs).
We need to add such case.

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -0,0 +1,1004 @@
...
+    inner_call_gas = base_gas if is_success else base_gas - 1
+
+    if same_tx:
+        # Deploy and selfdestruct in same transaction via factory
+        initcode = Initcode(deploy_code=victim_code)
+        initcode_len = len(initcode)
+
+        initcode_address = pre.deploy_contract(initcode)
+
+        factory_address = next(pre._contract_address_iterator)  # type: ignore
```
marioevz: Same recommendation as prior comment.

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -0,0 +1,1004 @@
...
+                        nonce_changes=[
+                            BalNonceChange(block_access_index=1, post_nonce=1)
+                        ],
+                        code_changes=[
+                            BalCodeChange(
+                                block_access_index=1,
+                                new_code=bytes(victim_code),
+                            )
+                        ],
+                    )
```
marioevz: I think doing `victim_expectation.balance_changes.append(BalBalanceChange(...` might make it a bit easier to read instead of repeating almost the same code in two different blocks.

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -0,0 +1,1004 @@
...
+
+            if originator_balance > 0:
+                caller_expectation = BalAccountExpectation(
+                    nonce_changes=[
+                        BalNonceChange(block_access_index=1, post_nonce=2)
+                    ],
+                    balance_changes=[
```
marioevz: Same comment as above.

marioevz (commented):
Looks great! I made a couple suggestions, thanks!

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
fselmo: Ah, ty! I was hoping to not have to reach into the private API 👍🏼

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
fselmo: I learn some new utility every day it seems! Thanks, this is really cool 👍🏼

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
fselmo: I think it's worth it to add this but EIP-161 sets the nonce to `1` before the initcode runs ([section a](https://eips.ethereum.org/EIPS/eip-161#specification)). So I think in any case it's not really possible to test the "dead" case. I suppose the "has code" part doesn't hold, is this what you mean? I can make that more clear here and add this  test case as well 👍🏼.

fselmo:
@marioevz Added changes from the comments here. I also realized that the `valid_from("Berlin")` pytest marker on the param was not working as expected. I added a `pytest_collection_modifyitems` to the forks plugin in order to sort the markers better. It worked out well but I also added some unit tests to make sure it's sanity checked. Lmk your thoughts on that approach. It makes it so we can run `warm` tests from Berlin on but `cold`, for example, for `TangerineWhistle` on. It's a nice added flexibility imo, which allowed me to not copy the entirety of the test over to Berlin just to add warm / cold paths.

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
marioevz: Yeah, just the "has code" part is the thing that came to mind for me, it would not hurt to have it implemented, thanks!

fselmo: added here ([518c5c8](https://github.com/ethereum/execution-specs/pull/1954/commits/518c5c8797f31d6f4fc19aa50dbde89c0f06ab76))

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
fselmo: added here [518c5c8](https://github.com/ethereum/execution-specs/pull/1954/commits/518c5c8797f31d6f4fc19aa50dbde89c0f06ab76)

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -640,7 +639,8 @@ def selfdestruct(evm: Evm) -> None:
     state = evm.message.block_env.state
     if is_cold_access:
         evm.accessed_addresses.add(beneficiary)
-        track_address(evm.state_changes, beneficiary)
+
+    track_address(evm.state_changes, beneficiary)
```
gurukamath, line 643: This is related to line 669 but leaving a comment here.

Could we defer the balance tracking for the `originator` and `beneficiary` until after the `same_tx` check on line 685? That way we do not have to balance track the `originator` twice for `same_tx`

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -494,12 +494,12 @@ def extcodehash(evm: Evm) -> None:
     check_gas(evm, access_gas_cost)
     if is_cold_access:
         evm.accessed_addresses.add(address)
-    track_address(evm.state_changes, address)
```
gurukamath, line 497: For `balance`, `extcodesize`, `extcodecopy` and `extcodehash`, we don't need to call `check_gas` first and then call `charge_gas`. Could we directly call `charge_gas` in these cases?

src/ethereum/forks/amsterdam/vm/instructions/system.py
fselmo: This is here because we reach into the account in the very next line `is_account_alive()`. So this is not for the balance change, this is for checking if the account is alive. Does that make sense? I think functionally we could defer it, but semantically it makes sense to track it here.

src/ethereum/forks/amsterdam/vm/instructions/environment.py
fselmo: Yes, I agree on this. This was part of the initial implementation of `check_gas`. I haven't looked back at those since then but I think this makes more sense for sure. Are you good with changing that here?

fselmo: Added in latest commit [here](https://github.com/ethereum/execution-specs/pull/1954/commits/6cb77129260ced9d711899953d8ba8cc6591c0bd).

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -0,0 +1,1197 @@
...
+
+    Key characteristics:
+    - During initcode, the contract has no code yet
+    - Contract has nonce=1 (post-EIP-161) making it non-empty
+    - Beneficiary is always warm (it's the executing contract)
+    - No G_NEW_ACCOUNT charge (contract has nonce > 0)
+    - No cold access charge (>=Berlin)
+
+    Note: Gas boundary testing not possible for initcode since CREATE
+    doesn't accept a gas parameter - it uses all available gas.
```
marioevz, line 1142: See example in this PR: https://github.com/ethereum/execution-specs/pull/2002/files#diff-095178dc2c6d121067cadab23cb09943682751281a0489a73f90906e26855a6a

Basically we limit the gas passed to the factory contract, taking into account the 63/64 rule.

fselmo: Really nice use 😄. I created #2010 to track this 👍🏼

marioevz (approved):
Approving based on:
- Tracking issue #2010 will fix outstanding comment.
- CI failures require new BAL release.

PR #1960, opened 2025-12-31
feat(test): port static context static tests to python

fselmo:
## 🗒️ Description

Port STATICCALL tests with zero and non-zero value to precompiles. This has the added bonus of extending the precompiles, parametrized per fork, via `pytest.mark.with_all_precompiles`.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


tests/byzantium/eip214_staticcall/test_staticcall.py
```diff
@@ -0,0 +1,340 @@
...
+
+    Original bytecode from: https://github.com/dapphub/dapptools/pull/360
+    Issue: https://github.com/ethereum/tests/issues/683
+    """
+    alice = pre.fund_eoa()
+
+    target_code = bytes.fromhex(
+        "600080541515601d576001815580818283305afa15601b578081fd5b005b"
+        "80818283600160025af15050"
+    )
```
LouisTsai-Csie: Could we still update this test in EELS-style instead of using plain byte code? 

I've reviewed related issue and the test logic: the contract first `STATICCALL` itself, thus entering the static mode, and then `CALL` into a precompile with non-zero value.

This is the disassembled [result](https://gist.github.com/LouisTsai-Csie/591df83ebbd7dd22184f046c5ce87fbb) of the target byte code, but i propose this refactor for simplicity and readability:

```python
x = Op.JUMPI(0xFF, Op.ISZERO(Op.CALLVALUE)) # Assume the PC is 1 byte
y = Op.STATICCALL(gas=200_000, address=Op.ADDRESS) + Op.STOP
z = Op.JUMPDEST + Op.SSTORE(0, Op.CALL(gas=200_000, address=0x01, value=1))

pc = len(x) + len(y)
assert len(Op.PUSH1(pc)) == len(Op.PUSH1(0xFF)) # Assert PC is 1 byte or it would jump to random destination.

target_code = Op.JUMPI(pc, Op.ISZERO(Op.CALLVALUE)) + y + z
```

If the `CALLVALUE` is non-zero, which comes from the transaction origin, then it performs `STATICCALL`, then it enters the context again with zero `CALLVALUE`, and jumps to the `CALL` section.

Some benefits here:
- Enhance readability
- We could update the CALL target to all the precompiles by `with_all_precompiles` marker.

LouisTsai-Csie: We could also change `y` to be `Op.SSTORE(Op.STATICCALL(address=Op.ADDRESS)) + Op.STOP` to verify the failure of `STATICCALL`.

tests/byzantium/eip214_staticcall/test_staticcall.py
```diff
@@ -0,0 +1,340 @@
...
+@pytest.mark.ported_from(
+    [
+        "https://github.com/ethereum/tests/blob/v13.3/src/GeneralStateTestsFiller/stStaticFlagEnabled/CallWithZeroValueToPrecompileFromCalledContractFiller.yml",
+        "https://github.com/ethereum/tests/blob/v13.3/src/GeneralStateTestsFiller/stStaticFlagEnabled/CallWithNOTZeroValueToPrecompileFromCalledContractFiller.yml",
+    ],
+)
+def test_staticcall_nested_call_to_precompile(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+    fork: Fork,
```
LouisTsai-Csie: This seems to be unused if `protected=True` in Transaction object

tests/byzantium/eip214_staticcall/test_staticcall.py
```diff
@@ -0,0 +1,340 @@
...
+    contract_a = pre.deploy_contract(
+        code=(
+            Op.SSTORE(0, marker)
+            + Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE)
+            + Op.SSTORE(1, Op.CREATE(value=0, offset=0, size=Op.CALLDATASIZE))
+            + Op.SSTORE(2, marker)
+        ),
+        balance=contract_initial_balance,
+    )
+    created_contract = compute_create_address(address=contract_a, nonce=1)
```
LouisTsai-Csie: I’ve reviewed the legacy tests `CallWithZeroValueToPrecompileFromContractInitializationFiller` & `CallWithNOTZeroValueToPrecompileFromContractInitializationFiller`, and it seems that the current implementation uses `CREATE2` instead of `CREATE` here, but I think we could support both.

Proposed changes:
- Parameterize the opcode to support both `CREATE` and `CREATE2`.
- Replace `Op.CREATE(size=Op.CALLDATASIZE)` with `opcode(size=Op.CALLDATASIZE)`.
- For address calculation, continue using `compute_create_address`, but pass the `initcode` and `opcode` as parameters.

Example of this approach can be found [here](https://github.com/ethereum/execution-specs/blob/d618e5259a06179d7556af1961166cb7bc5da1e6/tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py#L646-L651).

LouisTsai-Csie (dismissed):
Leave some comments! It is really crazy that old test is written in `yml` format.

tests/byzantium/eip214_staticcall/test_staticcall.py
fselmo: Yes! Great point.

tests/byzantium/eip214_staticcall/test_staticcall.py
fselmo: I did think about this as well. I think the only way I'd want to approach this, since this is such a specific test, is to make sure that the bytecode is exactly the same from the provided test with an `assert` to provide even better readability and confidence that this is the exact test, with the exact same bytecode, that the original was testing.

We can then add the `SSTORE` after this assert. I think it will be a bit difficult to map this exactly but I can give it a shot, unless you have this already from your effort? @LouisTsai-Csie, lmk.

danceratopz: Happy to try and work on our workflows to gain confidence when making these types of changes. It's a lot of work to verify such refactors, but they'll definitely bring a lot more value to this work.

Small comment about "decompiling" to the `execution-testing` opcode minilang, jicymi, there's `evm_bytes`. For example:

```console
  uv run evm_bytes hex-string "600080541515601d576001815580818283305afa15601b578081fd5b005b80818283600160025af15050"
```
outputs:
```python
Op.PUSH1[0x0] + Op.DUP1 + Op.SLOAD + Op.ISZERO + Op.ISZERO + Op.PUSH1[0x1d] + Op.JUMPI + Op.PUSH1[0x1] + Op.DUP2 + Op.SSTORE + Op.DUP1 + Op.DUP2 + Op.DUP3 + Op.DUP4 + Op.ADDRESS + Op.GAS + Op.STATICCALL + Op.ISZERO + Op.PUSH1[0x1b] + Op.JUMPI + Op.DUP1 + Op.DUP2 + Op.REVERT + Op.JUMPDEST + Op.STOP + Op.JUMPDEST + Op.DUP1 + Op.DUP2 + Op.DUP3 + Op.DUP4 + Op.PUSH1[0x1] + Op.PUSH1[0x2] + Op.GAS + Op.CALL + Op.POP + Op.POP
```

LouisTsai-Csie: Wow this feature is so nice. i was abusing ChatGPT for this.

fselmo: Wow thanks @danceratopz. TIL :)... it is indeed really nice.

fselmo:
@LouisTsai-Csie great suggestion for readability. I ended up keeping the other test with a comparison to its `evm_bytes` output for readability and I added your variation, parametrizing for all precompiles as well as for `call_value` being [0, 1] where we can check the success when `0` and fail for `1`.

Lmk your  thoughts. Ready for another pass.

---
Note: As mentioned in the unresolved comment here, we will have to wait for the feature introduced in #1954 to parametrize Create2 cases from `Constantinople` ([EIP-1014](https://eips.ethereum.org/EIPS/eip-1014)) onwards as `SpuriousDragon` did not have this available.

tests/byzantium/eip214_staticcall/test_staticcall.py
fselmo: The last commit in this PR (dbad28aace536518c07c749eeca6dd83a20c38a2) fixes the current issues, but it will have to wait on [this commit](https://github.com/ethereum/execution-specs/pull/1954/changes/74129679a0eeac3958bf1df41573d896d0709e83) that introduced `pytest.param` based fork aware test cases. Coincidentally I added this just this week and it would work quite well here.

So we will have to wait until #1954 is merged first.

tests/byzantium/eip214_staticcall/test_staticcall.py
```diff
@@ -0,0 +1,487 @@
...
+    Account,
+    Address,
+    Alloc,
+    Block,
+    BlockchainTestFiller,
+    Op,
+    Transaction,
+    compute_create_address,
+)
+from execution_testing.forks import Byzantium
```
marioevz:
```suggestion
```
Not used.

tests/byzantium/eip214_staticcall/test_staticcall.py
```diff
@@ -0,0 +1,487 @@
...
+
+@pytest.mark.ported_from(
+    [
+        "https://github.com/ethereum/tests/blob/v13.3/src/GeneralStateTestsFiller/stStaticFlagEnabled/StaticcallForPrecompilesIssue683Filler.yml"
+    ],  # noqa: E501
+)
+@pytest.mark.valid_from("Byzantium")
+def test_staticcall_reentrant_call_with_value_to_precompile_issue683(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
```
marioevz:
```suggestion
    state_test: StateTestFiller,
```
Can be a state test.

tests/byzantium/eip214_staticcall/test_staticcall.py
```diff
@@ -0,0 +1,487 @@
...
+        + Op.JUMPDEST  # [0x00]
+        + Op.STOP
+        # 0x1d: Reentry path - CALL to precompile 0x02 with value=1
+        + Op.JUMPDEST  # [0x00] - jumped here because S[0]!=0
+        + Op.DUP1  # [0x00, 0x00]
+        + Op.DUP2  # [0x00, 0x00, 0x00]
+        + Op.DUP3  # [0x00, 0x00, 0x00, 0x00]
+        + Op.DUP4  # [0x00, 0x00, 0x00, 0x00, 0x00]
+        + Op.PUSH1[0x1]  # [0x01, 0x00, 0x00, 0x00, 0x00, 0x00] - value
+        + Op.PUSH1[0x2]  # [0x02, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00] - addr
```
marioevz: Call an empty account  (using `pre.fund_eoa(amount=0)`) to guarantee this can be run using `execute`.

tests/byzantium/eip214_staticcall/test_staticcall.py
```diff
@@ -0,0 +1,487 @@
...
+    - call_value=0: CALL succeeds in static context → STATICCALL returns 1
+    - call_value>0: CALL violates static context, reverts frame → STATICCALL
+      returns 0
+    """
+    alice = pre.fund_eoa()
+
+    pc_placeholder = 0xFF
+
+    # if CALLVALUE == 0 (inside STATICCALL), jump to reentry path
+    entry_check = Op.JUMPI(pc_placeholder, Op.ISZERO(Op.CALLVALUE))
```
marioevz: We can also use [`Conditional`](https://github.com/ethereum/execution-specs/blob/9aefc86f66ebc8e1a41c530827d8f9ba1bd50fd2/tests/prague/eip7702_set_code_tx/test_set_code_txs.py#L384-L392) to simplify this bytecode construction.

tests/byzantium/eip214_staticcall/test_staticcall.py
```diff
@@ -0,0 +1,487 @@
...
+    3. Back in outer: STATICCALL returned 0 (failed) → STOP
+
+    Result: S[0]=1 persists, proving CALL-with-value in static context fails.
+    """
+    target_code_raw = bytes.fromhex(
+        "600080541515601d576001815580818283305afa15601b578081fd5b005b"
+        "80818283600160025af15050"
+    )
+
+    target_code = (
```
marioevz: We should also use `Conditional` here to have more readable code, no need to try to match the exact original bytecode IMO.

marioevz (commented):
Some small comments that can be applied to all test functions in the PR. Thanks!

tests/byzantium/eip214_staticcall/test_staticcall.py
fselmo: See my other comment, if we don't care about preserving this test case for historical purpose then I think we can just keep `test_staticcall_reentrant_call_with_value_to_precompile` and remove this.

tests/byzantium/eip214_staticcall/test_staticcall.py
fselmo: The goal here was to preserve the original test and to get it in our API so it is _more_ readable. We then created a version of the test, `test_staticcall_reentrant_call_with_value_to_precompile()`, that is more readable and does a similar thing while extending coverage.

If we don't care about keeping the exact bytecode that caught a bug once upon a time, then we can keep the variant of the test that is most readable. To be clear, I would prefer to keep the actual bytecode for added safety and for historical reasons. But if we don't want to try to match the original bytecode, I think we can remove this test case entirely.

fselmo (commented):
@marioevz I applied all the changes you suggested here. I opted to remove the first test entirely since we don't care about preserving the actual bytecode. We should carefully review to make sure it's covering the same case as the original test if this is the route we want to take.

---
edit: I had to add an update to the `pytest.param()` support for `valid_from()` because docs pulls in all forks and this was erring with the marker for `with_all_valid_precompiles` in conjunction with `valid_from()` being in the params. It wouldn't have this fork range updated and some forks don't support precompiles at all. It's a minor change [here](https://github.com/ethereum/execution-specs/pull/1960/changes/0ad70615c54c0e80880145ca922d3b82f7aaf492#diff-2b91aff16cbd15e7e97e543003a887327ed7faf938fea1b9c9371a9c5d5871deR1114-R1171), but we should make sure it is reviewed appropriately.

fselmo:
@LouisTsai-Csie can you re-review and remove the requested changes if this is good? It will block a merge here. This is ready for another pass.

fselmo:
I also added BAL expectations here since these were of interest to BALs originally, which is why it caught my eye to port them in the first place.

More recently (since this PR was opened), geth caught an implementation issue because of these static tests ([relevant conversation here](https://discord.com/channels/595666850260713488/1364000387195076608/1462768360336134259)).

marioevz (approved):
LGTM. I made one last push to change the test types to `StateTest` while preserving the BAL checks.

PR #1961, opened 2026-01-01
feat(benchmark): add EXTCODESIZE bytecode size benchmark for cold access testing

CPerezz:
## 🗒️ Description
Test EXTCODESIZE with parametrized bytecode sizes using CREATE2 factory.

This test executes EXTCODESIZE operations against pre-deployed contracts
via factories, measuring the performance impact of different contract
sizes on EXTCODESIZE operations.
Designed for execute mode only - contracts must be pre-deployed.

The test maximizes cold EXTCODESIZE calls to stress client state loading:
1. Using CREATE2 address derivation to access many unique contracts
2. Filling block gas with transactions close to FUSAKA_TX_GAS_LIMIT (16M)
3. Verifying contracts exist by checking the last accessed contract's size

This benchmark measures the performance impact of `EXTCODESIZE` operations
on contracts of varying sizes (0.5KB to 24KB).
It stresses client state loading by maximizing **cold** EXTCODESIZE calls.

## Overview

The test deploys attack contracts that loop through thousands of unique
contract addresses, calling `EXTCODESIZE` on each.
By using CREATE2 address derivation, the test accesses pre-deployed
contracts without storing their addresses, maximizing the number of cold
state accesses per block.

┌──────────────────┐
│                        Test Block                             │
├──────────-────────┤
│  TX1: Verification (~30K gas)                     │
│    └─> Calls EXTCODESIZE on last contract, stores result      │
│                                                               │
│  TX2: Attack (~16M gas)                                       │
│    └─> Loops EXTCODESIZE on salts 0..5,878                    │
│                                                               │
│  TX3: Attack (~16M gas)                                       │
│    └─> Loops EXTCODESIZE on salts 5,879..11,757               │
│                                                               │
│  TX4: Attack (~16M gas)                                       │
│    └─> Loops EXTCODESIZE on salts 11,758..17,636              │
└───────────────────────────┘

### Execute a Single Size

```bash
uv run execute remote \\
  --fork Prague \\
  --rpc-endpoint http://127.0.0.1:8545 \\
  --rpc-seed-key <SEED_KEY> \\
  --rpc-chain-id 1337 \\
  --address-stubs tests/benchmark/stateful/bloatnet/stubs.json \\
  -- -m stateful --gas-benchmark-values 60 \\
  tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py \\
  -k '24KB' -v
```

### Execute All Sizes

```bash
uv run execute remote \\
  --fork Prague \\
  --rpc-endpoint http://127.0.0.1:8545 \\
  --rpc-seed-key <SEED_KEY> \\
  --rpc-chain-id 1337 \\
  --address-stubs tests/benchmark/stateful/bloatnet/stubs.json \\
  -- -m stateful --gas-benchmark-values 60 \\
  tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py -v
```

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

image


tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,470 @@
...
+        + Op.POP  # Stack: [num_deployed]
+        # === Step 3: Main loop - EXTCODESIZE operations ===
+        + While(
+            body=(
+                # CREATE2 address: keccak256(0xFF ++ factory ++ salt ++ hash)
+                Op.SHA3(11, 85)  # 85 bytes from offset 11
+                # Result is 32-byte hash, address is last 20 bytes (low bits)
+                # AND with address mask to extract address from hash
+                + Op.PUSH20(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF)
+                + Op.AND
```
jochem-brouwer: Nit: this masking is not necessary. EVM will do this automatically. See: https://github.com/ethereum/go-ethereum/blob/b635e0632ce675be3d7cc0b498e08df8dc6346d6/core/vm/instructions.go#L337

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,470 @@
...
+            address=factory_address,
+            args_offset=0,
+            args_size=0,
+            ret_offset=96,  # Store result at memory[96]
+            ret_size=64,  # 64 bytes for 2 uint256s
+        )
+        # Check if call succeeded (STATICCALL returns 1 on success)
+        # Stack: [starting_salt, success]
+        + Op.ISZERO
+        + Op.PUSH2(0x1000)  # Jump to end if failed
```
jochem-brouwer: This will (likely) jump to an invalid JUMPDEST, right? (So it will error / OOG). Comments seem to indicate that this is some designated area to jump to but this does not seem to be happening here?

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,470 @@
...
+        to=verification_address,
+        sender=sender,
+        data=b"",
+        value=0,
+    )
+    txs.append(verification_tx)
+
+    # Attack transactions: fill remaining block gas with ~16M gas txs
+    # Each tx uses a different starting salt to access unique contracts
+    for tx_index in range(num_attack_txs):
```
jochem-brouwer: Can also keep track of the remaining gas here: we have the verification transaction, so deduct this from the gas benchmark value. Now keep adding transactions with `FUSAKA_TX_GAS_LIMIT` until there is less than `FUSAKA_GAS_LIMIT` left. In that case, calculate if we can add a transaction (if the remaining gas is lower than the intrinsic cost, we cannot add a transaction). 
As side note, it would be super helpful if we can come up with a way to get gas cost from a "test run". So we would do something like `eth_call` here to figure out the gas cost, and use that to fill the test. The reason for this is that in here we assume that the transactions use all the gas, but due to the setup it is likely that not the entire gas limit is spent on the transaction (there is some left). I have no idea how to do this though (will bring this up at the benchmarking call next week, because this would make writing these tests much easier than having to hardcode all gas costs, like the `gas_costs.G_KECCAK_256 + gas_costs. ...` calculations)

jochem-brouwer (commented):
Saw this PR, some small comments/ideas 😄 👍

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,470 @@
+r"""
```
MariusVanDerWijden, line 1:
```suggestion
"""
```

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
MariusVanDerWijden: Jep a return here would probably be better

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: If I do that, the formatting linter complains. This is the only way that allows me to add MD clauses.
Is this a must-change?

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,470 @@
...
+sizes on EXTCODESIZE operations.
+Designed for execute mode only - contracts must be pre-deployed.
+
+The test maximizes cold EXTCODESIZE calls to stress client state loading:
+1. Using CREATE2 address derivation to access many unique contracts
+2. Filling block gas with transactions close to FUSAKA_TX_GAS_LIMIT (16M)
+3. Verifying contracts exist by checking the last accessed contract's size
+
+This benchmark measures the performance impact of `EXTCODESIZE` operations
+on contracts of varying sizes (0.5KB to 24KB).
```
jochem-brouwer, line 5: For the benches we need for code size increase we thus need the 32kb contract in here also. (This is part of predeploy so should not be a problem. However, predeploy thus needs to run on a fork with client which supports 32kb contracts)

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: Addressed in [341594c](https://github.com/ethereum/execution-specs/pull/1961/commits/341594c3c6e2e819bebc8ae333a6784d39782cba)

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: Addressed in [14cccd8](https://github.com/ethereum/execution-specs/pull/1961/commits/14cccd8426b2578e4c81b1654ff52e499d6c750e). 

I just discovered `Conditional` which makes handling this much easier than just jumping to an invalid destination to revert.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,470 @@
...
+    Transaction,
+    While,
+)
+from execution_testing.forks.gas_costs import GasCosts
+
+REFERENCE_SPEC_GIT_PATH = "DUMMY/bloatnet.md"
+REFERENCE_SPEC_VERSION = "1.0"
+
+# Fusaka transaction gas limit (16M gas)
+FUSAKA_TX_GAS_LIMIT = 16_000_000
```
jochem-brouwer:
```suggestion
FUSAKA_TX_GAS_LIMIT = 16_777_216
```

See https://eips.ethereum.org/EIPS/eip-7825

Even better, you can also use `fork.transaction_gas_limit_cap()` see e.g. https://github.com/ethereum/execution-specs/blob/d618e5259a06179d7556af1961166cb7bc5da1e6/packages/testing/src/execution_testing/specs/benchmark.py#L410

(not a blocker, but I think the fusaka tx gas limit should be fixed here)

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: I agree that this could be slightly more optimized. 

Nevertheless, this is the only way that doesn't either complicate the scenario more. Do you think I should make a bigger change?

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,463 @@
...
+        + Op.PUSH1(64)
+        + Op.MSTORE  # Stores init_code_hash at memory[64]
+        # Stack: [starting_salt, num_deployed]
+        # Store starting salt at memory[32]
+        + Op.SWAP1  # Stack: [num_deployed, starting_salt]
+        + Op.DUP1  # Stack: [num_deployed, starting_salt, starting_salt]
+        + Op.PUSH1(32)
+        + Op.MSTORE  # Store starting_salt at memory[32]
+        # Stack: [num_deployed, starting_salt]
+        + Op.POP  # Stack: [num_deployed]
```
jochem-brouwer: So the counter here is `num_deployed` - right? 
This means that all txs except the last one will out-of-gas. A tx will revert in case that the conditional fails. 

I'd propose a slightly other design. It's not perfect, it will add small overheads, but it saves the calculation of the gas per loop and the starting salt of the tx.

To do so, we read the starting salt from slot 0. We then enter the while loop, and the exit condition is that Op.GASLEFT is higher than some value (which is not too low, and also not too high. It should ensure we can at least afford one more loop, and the SSTORE at the end of the loop). When we exit the loop, we save the current salt in slot 0. Therefore, the next tx will read the init seed and thus start at the account we would read if the tx would not be split up into multiple txs (so essentially resuming the loop).

The test verification should check that these transactions do not OOG.

To verify that the accounts we queried exists, we can use the return value of the factory contract: the `num_contracts`. Before performing the SSTORE, we should check that the salt we are storing is less (strictly less, so Op.LT, `<`) than `num_deployed` as returned from the factory contract.

If a tx OOGs or reverts, it means either:

1. There was not enough gas left for the loop + SSTORE (therefore each tx will start at salt 0)
2. There were not enough contracts deployed

This setup removes the need for the gas calculations. It adds overhead to the EVM (the gas check and the final SSTORE) which thus "wastes" gas. However since we are querying cold accounts (at least 2600 gas) on Fusaka tx gas limit (16+M) the amount of extra accounts we could have queried without this overhead is much less than the accounts which we would query here, so neglegible. It therefore gives a good indication of the worst case, and it saves the error-prone gas calculations (if we edit the code, we need to edit the gas calculations again, and we might forget to do one of the two 😅 ).

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,463 @@
...
+        # === Step 3: Main loop - EXTCODESIZE operations ===
+        + While(
+            body=(
+                # CREATE2 address: keccak256(0xFF ++ factory ++ salt ++ hash)
+                Op.SHA3(11, 85)  # 85 bytes from offset 11
+                # Hash result - EVM auto-truncates to 20-byte address
+                # Call EXTCODESIZE and discard result (no storage writes!)
+                + Op.EXTCODESIZE
```
jochem-brouwer: With above suggestion, we can directly edit this test to perform the other operations also!
The EVM gas check loop will take care of  the gas calculations and the loop exits.

If we now have `Op.POP(Op.EXTCODESIZE(Op.SHA3(11, 85))` we can edit this to:

`Op.POP(Op.EXTCODEHASH(Op.SHA3(11, 85))))`
`Op.EXTCODECOPY(Op.SHA3(11, 85), 0, 1, 1))` -> note: have to copy at least one byte otherwise we do not strictly need to lookup the account. Also need to ensure we are not writing to the window we are using to calculate the CREATE2 address
`Op.POP(Op.CALL(address=Op.SHA3(11, 85),gas=1)` (`CALL` can be interchanged with `CALLCODE`, `DELEGATECALL`, `STATICCALL`)

The verification contract should stay the same (extcodesize > 0 verifies contract exists).

Note: for the ops which CALL into the target contract, we just send 1 gas, so it does not matter what code it executes. If it OOGs or not, it will "waste" at most 1 gas.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,463 @@
...
+                + Op.POP
+                # Increment salt for next iteration
+                + Op.MSTORE(32, Op.ADD(Op.MLOAD(32), 1))
+            ),
+            # Continue while counter > 0
+            # Decrement counter and check if non-zero
+            condition=(
```
jochem-brouwer, line 142: So this condition could change to the gas left check.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,463 @@
...
+                Op.PUSH1(1)
+                + Op.SWAP1
+                + Op.SUB
+                + Op.DUP1
+                + Op.ISZERO
+                + Op.ISZERO  # Convert to boolean: 1 if counter > 0
+            ),
+        )
+        + Op.POP  # Clean up remaining counter
```
jochem-brouwer: SSTORE the current salt here.

Note: the initial SSTORE will create the slot, so needs 20k gas for creation (plus warming it). So a ballpark overhead for the SSTORE could be ~30k. 

To be safe we can use gasleft > 50k. The "loss" here would be then <0.4% of gas spent on "unrelated operations" (50_000 / 16_000_000)

jochem-brouwer (commented):
General suggestion to remove the need for gas calculations. I think this kind of pattern is something we should at some point make into a template (not necessary now, but just noting for future cleanup). I think this kind of attack pattern (execute certain code against a CREATE2 factory) is something which we want to use more often and could reuse.
As mentioned in the comments it is not perfect, but the overhead is small enough compared to what we are measuring here (cold account access).

General Q: where do I find the initcode of the deployed contracts? I understand that we can choose any code here (via the stubs), but would like to see what contracts are initialized.

Have added a suggestion to directly add the other operations `EXTCODEx` and `xCALLx`s here 😄 👍

jochem-brouwer:
Let me know if anything is unclear, happy to help :smile: :+1:

LouisTsai-Csie:
This PR looks nice, i will just add some format suggestion later, and i wonder if we could remove some of the comments.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: I'll tackle this locally in my machine. 
So will only be able to do it with empty state. 


Sadly we would need all clients in bloatnet to support it and to hard-fork such that we can test it propperly

CPerezz:
> General Q: where do I find the initcode of the deployed contracts? I understand that we can choose any code here (via the stubs), but would like to see what contracts are initialized.

You can get this from the deployer scenario that you can find in: https://github.com/ethpandaops/spamoor/pull/161

LMK if something isn't clear.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: Addressed in [179b011](https://github.com/ethereum/execution-specs/pull/1961/commits/179b01136ad897ffdb49b421139db8223cfaa7a2)

I thought 50k was a lot. But indeed doesn't seem to crazily affect how optimal it is

jochem-brouwer:
Thanks for the spamoor URL :smile: :+1: 

I have one more argument why I think we should switch to storing the pointer of the next salt into the EVM. If we do this via the calldata approach, then we can parallelize these transactions in an optimistic way (we do not need to execute previous transactions in order to execute the next one. Even though they hit the same attack contract they do not read/write to same storage). So if we store the pointer into the EVM we force all clients to execute these transactions serially, because the second (attack) transaction has to "know" which salt it has to start from, and therefore has to execute the first (attack) transaction first. (Bonus: BALs fix this, making them parallelizable, because we can now read the pointer from the BAL).

Let me know if I should assist here, it's a rather big change in the logic, happy to help :smile: :+1:

CPerezz:
@jochem-brouwer that makes all the sense! Thanks for the suggestion. You are right. 

I think the latest commit should do what you suggested! Ofc, can't push here anything >24KB. But, will do this bench locally.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+    2. Loop exits when gas < 50K, saves salt for next TX
+    3. Each TX automatically resumes from where previous left off
+    4. No manual gas calculations needed - contract self-regulates
+
+    Verification TX checks that contracts exist by calling EXTCODESIZE
+    on salt 0 (first contract) and storing the result.
+    """
+    gas_costs = fork.gas_costs()
+    # Use fork's TX gas limit cap, or 16M fallback for pre-Osaka forks
+    tx_gas_limit = fork.transaction_gas_limit_cap() or 16_000_000
```
LouisTsai-Csie: We have this tx_gas_limit fixture already. Just import it like below:

```python
def test_extcodesize_bytecode_sizes(
    pre: Alloc,
    fork: Fork,
    bytecode_size_kb: float,
    gas_benchmark_value: int,
    tx_gas_limit: int # fork.transaction_gas_limit_cap() or gas_benchmark_value
) -> None:
```
And then you could reference it in the code.

The implementation could be found here:

https://github.com/ethereum/execution-specs/blob/3ac8ce8865e2239a0d5d576091770573accc04c8/tests/benchmark/conftest.py#L90-L93

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+                + Op.MSTORE  # Store size at memory[160] for verification
+                # Increment salt for next iteration
+                + Op.MSTORE(32, Op.ADD(Op.MLOAD(32), 1))
+            ),
+            # Continue while: gas > gas_reserve AND salt < num_deployed
+            condition=(
+                # Check gas > gas_reserve
+                Op.GAS
+                + Op.PUSH3(gas_reserve)
+                + Op.GT  # gas > gas_reserve
```
LouisTsai-Csie: If the comparison here is GAS > gas_reserve, i think this should be in reverse order?

```python
Op.PUSH3(gas_reserve)
Op.GAS
Op.GT
```

Or we have syntax like: `Op.GT(Op.GAS, gas_reserve)`

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+        return "bloatnet_factory_5kb"
+    elif size_kb == 10.0:
+        return "bloatnet_factory_10kb"
+    elif size_kb == 24.0:
+        return "bloatnet_factory_24kb"
+    else:
+        raise ValueError(f"Unsupported size: {size_kb}KB")
+
+
+def build_attack_contract(factory_address: Address) -> Bytecode:
```
LouisTsai-Csie, line 102: I appreciate the detailed comment, but i found some of them are repeating. Below i simplify the logic and comments a little, could you help check if this align with your intention, and if remains the readability?

```python
def build_attack_contract(factory_address: Address) -> Bytecode:
  """
  Benchmark EXTCODESIZE calls with gas-based loop exit.

  Storage Layout:
   - Slot 0: current salt (persists across transactions)
   - Slot 1: last EXTCODESIZE result (for verification)

  CREATE2 Memory Layout (85 bytes from offset 11):
   - MEM[11]    = 0xFF prefix
   - MEM[12-31] = factory address (20 bytes)
   - MEM[32-63] = salt (32 bytes)
   - MEM[64-95] = init_code_hash (32 bytes)
  """
  gas_reserve = 50_000  # Reserve for 2x SSTORE + cleanup

  return (
      # Call factory.getConfig() -> (num_deployed, init_code_hash)
      + Conditional(
          condition=Op.STATICCALL(
              gas=Op.GAS,
              address=factory_address,
              args_offset=0,
              args_size=0,
              ret_offset=96,  # MEM[96]=num_deployed, MEM[128]=init_code_hash
              ret_size=64,
          ),
          if_false=Op.REVERT(0, 0),
      )
      # Setup CREATE2 memory: keccak256(0xFF ++ factory ++ salt ++ init_code_hash)
      + Op.MSTORE(0, factory_address)
      + Op.MSTORE8(11, 0xFF)
      + Op.MSTORE(32, Op.SLOAD(0))        # Load salt directly to memory
      + Op.MSTORE(64, Op.MLOAD(128))      # init_code_hash
      + Op.MSTORE(160, 0)                 # Initialize last_size
      + Op.MLOAD(96)                      # Stack: [num_deployed]
      + While(
          body=(
              Op.MSTORE(160, Op.EXTCODESIZE(Op.SHA3(11, 85)))
              + Op.MSTORE(32, Op.ADD(Op.MLOAD(32), 1))
          ),  # Stack unchanged: [num_deployed]
          condition=(
              Op.GT(Op.GAS, gas_reserve)      # [num_deployed, (gas>reserve)]
              + Op.GT(Op.DUP2, Op.MLOAD(32))  # [num_deployed, (gas>res), (num_deployed>salt)]
              + Op.AND
          ),
      )
      + Op.POP
      + Op.SSTORE(0, Op.MLOAD(32))        # Save final salt
      + Op.SSTORE(1, Op.MLOAD(160))       # Save last result
      + Op.STOP
  )
```

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+        # Generate CREATE2 address
+        + Op.SHA3(11, 85)
+        # Result is 32-byte hash - EVM auto-truncates to 20-byte address
+        # Call EXTCODESIZE
+        + Op.EXTCODESIZE
+        # Store result in storage slot 0
+        + Op.PUSH1(0)
+        + Op.SSTORE
+        + Op.STOP
+    )
```
LouisTsai-Csie: Some format suggestion:

```python
def build_verification_contract(
      factory_address: Address, verification_salt: int
  ) -> Bytecode:
  """
  Verify EXTCODESIZE result for a specific salt by storing it in slot 0.

  CREATE2 Memory Layout (same as attack contract):
   - MEM[11]    = 0xFF prefix
   - MEM[12-31] = factory address
   - MEM[32-63] = salt
   - MEM[64-95] = init_code_hash
  """
  return (
      # Call factory.getConfig() -> (num_deployed, init_code_hash)
      Op.POP(Op.STATICCALL(
          gas=Op.GAS,
          address=factory_address,
          args_offset=0,
          args_size=0,
          ret_offset=96,  # MEM[96]=num_deployed, MEM[128]=init_code_hash
          ret_size=64,
      ))
      # Setup CREATE2 memory
      + Op.MSTORE(0, factory_address)
      + Op.MSTORE8(11, 0xFF)
      + Op.MSTORE(32, verification_salt)
      + Op.MSTORE(64, Op.MLOAD(128))
      # EXTCODESIZE on CREATE2 address, store result
      + Op.SSTORE(0, Op.EXTCODESIZE(Op.SHA3(11, 85)))
      + Op.STOP
  )
```

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
jochem-brouwer: This check checks `gas_reserve > gas`, `Op.GAS` and `Op.PUSH3(gas_reserve)` should be swapped

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+            # Continue while: gas > gas_reserve AND salt < num_deployed
+            condition=(
+                # Check gas > gas_reserve
+                Op.GAS
+                + Op.PUSH3(gas_reserve)
+                + Op.GT  # gas > gas_reserve
+                # Check salt < num_deployed (num_deployed > salt)
+                + Op.DUP2  # Stack: [..., (gas>reserve), num_deployed]
+                + Op.MLOAD(32)  # Stack: [..., (gas>res), num_deployed, salt]
+                + Op.GT  # Stack: [..., (gas>reserve), (num_deployed > salt)]
```
jochem-brouwer: Same here, we check now `salt > num_deployed`. We cannot swap `Op.DUP2` and `Op.MLOAD(32)` here because it now turns into an `Op.DUP3`

Instead, we can read `num_deployed` from memory (reading from memory is a great trick by the way, I was thinking to update this thing also for those SSTOREs with a lot of stack manipulation, but we can just read from memory!)

It should be:
`Op.MLOAD(32) + Op.MLOAD(96) + Op.GT` (we re-read the output of the factory, byte 96 should be the num_deployed)`

jochem-brouwer (commented):
Since we now verify the execution of this attack by inspecting the storage, the verification tx is not necessary anymore. Also really nice to use the memory instead of stack manipulation. I have two points in which the bytecode seems to be incorrect.

Let me add here how I tested this.

I added some extra code:

```python
    factory_stub = get_factory_stub_name(bytecode_size_kb)

    # FACTORY CODE DUMMY

    INITCODE = Op.STOP * expected_size_bytes
    INITCODE_HASH = Bytes(INITCODE).keccak256() # dummy hash
    NUM_DEPLOYED = 0xffffff

    factory_code = (Op.MSTORE(0, NUM_DEPLOYED) # num deployed
           + Op.MSTORE(32, INITCODE_HASH) # initcode hash
           + Op.RETURN(0, 64))

    # Deploy factory stub (address comes from stub file)
    factory_address = pre.deploy_contract(
        code=factory_code,  # Empty bytecode - address from stub
        stub=factory_stub,
    )

    for _ in range(16):
        pre.deploy_contract(address=compute_create2_address(initcode=INITCODE,
                                                            salt=_,
                                                            address=factory_address),
                            code=Op.STOP * expected_size_bytes)

    # Build and deploy the attack contract with storage initialized
```

(import `compute_create2_address` and `Bytes` from `execution_testing`)

This setups a dummy contract for the factory (simply returns the num deployed and the hash) and also setups the target contracts.

Testing with: `uv run fill -m stateful ./tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py --clean --evm-dump-dir=./dump --traces --block-gas-limit 20000000`

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+    factory_address = pre.deploy_contract(
+        code=Bytecode(),  # Empty bytecode - address from stub
+        stub=factory_stub,
+    )
+
+    # Build and deploy the attack contract with storage initialized
+    attack_code = build_attack_contract(factory_address)
+    attack_address = pre.deploy_contract(
+        code=attack_code,
+        storage={0: 0},  # Initialize salt counter to 0
```
LouisTsai-Csie:
```suggestion
```

I think it is by default empty storage.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+    print(f"\n{'=' * 60}")
+    print(f"EXTCODESIZE Benchmark: {bytecode_size_kb}KB contracts")
+    print(f"{'=' * 60}")
+    print(f"Block gas budget: {gas_benchmark_value:,}")
+    print(f"TX gas limit: {tx_gas_limit:,}")
+    print(f"Number of attack txs: {num_attack_txs}")
+    print(f"Verification tx gas: {verification_gas:,}")
+    print(f"Expected contract size: {expected_size_bytes} bytes")
+    print("Note: Using gas-based loop - each TX auto-resumes from storage")
+    print(f"{'=' * 60}\n")
```
LouisTsai-Csie: These are helpful, but not sure if we want to merge them or just for local debugging.

jochem-brouwer:
There is another problem with the post state verification:

```
    post = {
        verification_address: Account(
            storage={
                0: expected_size_bytes,  # EXTCODESIZE on salt 0
            }
        ),
        attack_address: Account(
            storage={
                # Slot 1: last EXTCODESIZE result should match expected size
                1: expected_size_bytes,
            }
        ),
    }
```

For `attack_address` we only want to verify key 1. In key 0 we store the current salt. EELS rejects because the post-state also contains the 0 key. (This is only for `uv run fill`, might not pop up in `uv run execute`)

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+    txs = []
+
+    # First transaction: verification (runs first, uses minimal gas)
+    verification_tx = Transaction(
+        gas_limit=verification_gas,
+        to=verification_address,
+        sender=sender,
+        data=b"",
+        value=0,
+    )
```
LouisTsai-Csie: Please correct me if i am wrong, is it necessary to pack verification_tx and other attach transaction within the same block? If not, you could create 2 blocks: the first one only do verification, and the second one only execute the attack, then you do not need to calculate the gas cost for the gas limit (verification_gas).


Something like this
```python
blocks = []

verification_tx = Transaction(
    gas_limit=verification_gas,
    to=verification_address,
    sender=sender,
)

blocks.append(Block(txs=[verification_tx]))
```

Some example: 

https://github.com/ethereum/execution-specs/blob/3ac8ce8865e2239a0d5d576091770573accc04c8/tests/benchmark/compute/instruction/test_account_query.py#L337-L345

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
```diff
@@ -0,0 +1,434 @@
...
+
+    # Attack transactions: all identical, no calldata needed
+    # Each TX reads salt from storage, loops until gas low, saves salt back
+    for _ in range(num_attack_txs):
+        attack_tx = Transaction(
+            gas_limit=tx_gas_limit,
+            to=attack_address,
+            sender=sender,
+            data=b"",  # No calldata - salt comes from storage
+            value=0,
```
LouisTsai-Csie:
```suggestion
```

LouisTsai-Csie (changes requested):
Hi @CPerezz i add some refactor suggestion, but do not change the logic much as it is quite good. The comment is very clear, but i found some of repeated comments, could you help clean up some of them? Thanks

jochem-brouwer:
For the storage to work with `uv run fill`, we need:

```python

    attack_storage = Storage({
                # Slot 1: last EXTCODESIZE result should match expected size
                1: expected_size_bytes,
            })
    attack_storage.set_expect_any(0)

    # Post-state verification:
    # 1. Verify that verification contract stored expected size (salt 0)
    # 2. Verify attack contract's last EXTCODESIZE returns expected size
    #    (proves the gas-based loop ran and accessed real contracts)
    post = {
        verification_address: Account(
            storage={
                0: expected_size_bytes,  # EXTCODESIZE on salt 0
            }
        ),
        attack_address: Account(
            storage=attack_storage
        ),
    }
```

This skips the storage key 0 :smile: (can have any value). We only care about key 1, which should have the EXTCODESIZE of the last salt we ran.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: Done - now using the fixture instead of manual fork.transaction_gas_limit_cap() call.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: Removed storage={0: 0}, data=b"", and value=0 - all defaults.

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: Removed - they were just for local debugging. MB

tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
CPerezz: I did it like this such that is easier for Kamil to integrate with his tool. Happy to change though.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -128,6 +128,17 @@ def test_sload_empty_erc20_balanceof(
     # Calculate gas costs
     intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(calldata=b"")
 
+    # Per-contract fixed overhead (setup + teardown for each contract's loop)
+    overhead_per_contract = (
```
jochem-brouwer: Are these changes intended for this PR?

jochem-brouwer (changes requested):
I ran the test locally using the dummy factory and dummy contracts as described above (with `uv run fill`), and it throws:

`GasUsedExceedsLimitError('gas used exceeds limit')`

This is because we have a verification tx and an attack tx, and in my particular setup the total gas of verification + attack goes over the block gas limit.
@LouisTsai-Csie suggests removing the verification tx and we can indeed do that, as the final verification is the storage check of the attack contract (checks that final salt queries an account with expected code size)

```diff
diff --git a/tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py b/tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
index ba66da5b5..52398ec08 100644
--- a/tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
+++ b/tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py
@@ -293,12 +293,12 @@ def test_extcodesize_bytecode_sizes(
     txs = []
 
     # First transaction: verification (runs first, uses minimal gas)
-    verification_tx = Transaction(
-        gas_limit=verification_gas,
-        to=verification_address,
-        sender=sender,
-    )
-    txs.append(verification_tx)
+    #verification_tx = Transaction(
+    #    gas_limit=verification_gas,
+    #    to=verification_address,
+    #    sender=sender,
+    #)
+    #txs.append(verification_tx)
 
     # Attack transactions: all identical, no calldata needed
     for _ in range(num_attack_txs):
@@ -319,7 +319,7 @@ def test_extcodesize_bytecode_sizes(
     attack_storage.set_expect_any(0)
 
     post = {
-        verification_address: Account(storage={0: expected_size_bytes}),
+        #verification_address: Account(storage={0: expected_size_bytes}),
         attack_address: Account(storage=attack_storage),
     }
 ```

Removing the verification tx makes it pass locally (please confirm if verification tx can indeed be removed)

File: tests/benchmark/stateful/bloatnet/test_single_opcode.py (hunk 20 above)
Comment (CPerezz) [reply tests/benchmark/stateful/bloatnet/test_single_opcode.py]: Nope.. Thanks for the catch. I was working on a worktree and got this mixed with my work on #1962

**Review by jochem-brouwer — APPROVED** [review]
LGTM, great work!

Tested locally, it passes 😄 👍 

Image

File: tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py (hunk 3 above)
Comment (LouisTsai-Csie) [reply tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py]: We could keep this as is for now. The approach Jochem mentioned needs some refactor in the EELS framework. As currently we construct all the test at the same time and send them.

File: tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py (hunk 4 above)
Comment (LouisTsai-Csie) [reply tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py:1]: We could keep this for now if CI is not complaining

File: tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py (hunk 18 above)
Comment (LouisTsai-Csie) [reply tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py]: We implement a feature for Kamil now, so user could specify the category of the transaction, and the info will be embedded in transaction field, which could be parsed and filtered later.

https://github.com/ethereum/execution-specs/blob/forks/amsterdam/docs/running_tests/execute/transaction_metadata.md

Some example:

```python
with TestPhaseManager.setup(): 
  # Some Benchmark Setup Transaction / Blocks
  # All the operations here would be marked as "SETUP"
  # The Real Benchmark Transaction / Blocks
  # Kamil only collect info with this label
```

https://github.com/ethereum/execution-specs/blob/a64f269f8f496063af924c9b94c70bbf04938c42/tests/benchmark/compute/instruction/test_account_query.py#L587-L611

I notice you remove the verification transaction, but if next time with the same scenario, you could try to use it.

LouisTsai-Csie (approved):
Thanks! LGTM

PR #1962, opened 2026-01-02
feat(benchmark): support tx gas limit cap in stateful benchmarks

CPerezz:
## 🗒️ Description

Update SLOAD/SSTORE and multi-opcode stateful benchmark tests to support Fusaka's 16M transaction gas limit. Tests now split large attack transactions into multiple smaller ones that each respect the `tx_gas_limit` cap while still filling the entire block gas budget.

Changes:
- Added `tx_gas_limit` fixture parameter to all affected test functions in `test_single_opcode.py` and `test_multi_opcode.py`
- Calculate `num_txs = max(1, gas_benchmark_value // tx_gas_limit)` to determine how many transactions are needed
- Split single large attack transaction into multiple smaller transactions in a list
- Updated gas calculations to be per-transaction rather than per-block

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

- [x] All: PR title adheres to the repo standard - it will be used as the squash commit message and should start `type(scope):`.
- [x] Tests: Verified SLOAD and SSTORE benchmarks work with `--transaction-gas-limit 16000000` and `--gas-benchmark-values 34` on Anvil

#### Cute Animal Picture

image

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_multi_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_multi_opcode.py
@@ -83,7 +83,6 @@ def test_bloatnet_balance_extcodesize(
     # Setup overhead (before loop): STATICCALL + result handling + memory setup
     setup_overhead = (
         gas_costs.G_COLD_ACCOUNT_ACCESS  # STATICCALL to factory (2600)
-        + 100  # STATICCALL base cost
         + gas_costs.G_VERY_LOW  # ISZERO (3)
         + gas_costs.G_VERY_LOW  # PUSH2 (3)
         + gas_costs.G_HIGH  # JUMPI (10)
@@ -273,7 +272,6 @@ def test_bloatnet_balance_extcodecopy(
     # Setup overhead (before loop): STATICCALL + result handling + memory setup
     setup_overhead = (
         gas_costs.G_COLD_ACCOUNT_ACCESS  # STATICCALL to factory (2600)
-        + 100  # STATICCALL base cost
         + gas_costs.G_VERY_LOW  # ISZERO (3)
         + gas_costs.G_VERY_LOW  # PUSH2 (3)
         + gas_costs.G_HIGH  # JUMPI (10)
@@ -470,7 +468,6 @@ def test_bloatnet_balance_extcodehash(
     # Setup overhead (before loop): STATICCALL + result handling + memory setup
     setup_overhead = (
         gas_costs.G_COLD_ACCOUNT_ACCESS  # STATICCALL to factory (2600)
-        + 100  # STATICCALL base cost
         + gas_costs.G_VERY_LOW  # ISZERO (3)
         + gas_costs.G_VERY_LOW  # PUSH2 (3)
         + gas_costs.G_HIGH  # JUMPI (10)
@@ -774,7 +771,7 @@ def test_mixed_sload_sstore(
         + gas_costs.G_VERY_LOW  # MLOAD for While condition check (3)
         + gas_costs.G_BASE  # ISZERO (2)
         + gas_costs.G_BASE  # ISZERO (2)
-        + gas_costs.G_MID  # JUMPI (8)
+        + gas_costs.G_HIGH  # JUMPI (10)
         + gas_costs.G_BASE  # POP to clean up at end (2)
     )
 
--- a/tests/benchmark/stateful/bloatnet/test_multi_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_multi_opcode.py
@@ -84,11 +84,11 @@ def test_bloatnet_balance_extcodesize(
     setup_overhead = (
         gas_costs.G_COLD_ACCOUNT_ACCESS  # STATICCALL to factory (2600)
         + 100  # STATICCALL base cost
-        + gas_costs.G_BASE  # ISZERO (3)
+        + gas_costs.G_VERY_LOW  # ISZERO (3)
         + gas_costs.G_VERY_LOW  # PUSH2 (3)
-        + gas_costs.G_MID  # JUMPI (10)
-        + gas_costs.G_LOW * 2  # MLOAD × 2 for factory results (3 * 2)
-        + gas_costs.G_LOW * 3  # MSTORE × 3 for memory setup (3 * 3)
+        + gas_costs.G_HIGH  # JUMPI (10)
+        + gas_costs.G_VERY_LOW * 2  # MLOAD × 2 for factory results (3 * 2)
+        + gas_costs.G_VERY_LOW * 3  # MSTORE × 3 for memory setup (3 * 3)
         + gas_costs.G_VERY_LOW  # MSTORE8 for 0xFF prefix (3)
         + gas_costs.G_VERY_LOW  # PUSH1 for memory position (3)
     )
@@ -98,14 +98,14 @@ def test_bloatnet_balance_extcodesize(
 
     # While loop condition overhead per iteration
     loop_condition_overhead = (
-        gas_costs.G_BASE  # DUP1 (3)
+        gas_costs.G_VERY_LOW  # DUP1 (3)
         + gas_costs.G_VERY_LOW  # PUSH1 (3)
         + gas_costs.G_VERY_LOW  # SWAP1 (3)
         + gas_costs.G_VERY_LOW  # SUB (3)
-        + gas_costs.G_BASE  # DUP1 (3)
-        + gas_costs.G_BASE  # ISZERO (3)
-        + gas_costs.G_BASE  # ISZERO (3)
-        + gas_costs.G_MID  # JUMPI (10)
+        + gas_costs.G_VERY_LOW  # DUP1 (3)
+        + gas_costs.G_VERY_LOW  # ISZERO (3)
+        + gas_costs.G_VERY_LOW  # ISZERO (3)
+        + gas_costs.G_HIGH  # JUMPI (10)
     )
 
     # Cost per contract access with CREATE2 address generation
@@ -116,10 +116,10 @@ def test_bloatnet_balance_extcodesize(
         + gas_costs.G_BASE  # POP first result (2)
         + gas_costs.G_WARM_ACCOUNT_ACCESS  # Warm access (100)
         + gas_costs.G_BASE  # POP second result (2)
-        + gas_costs.G_BASE  # DUP1 before first op (2)
-        + gas_costs.G_LOW  # MLOAD for salt (3)
+        + gas_costs.G_VERY_LOW  # DUP1 before first op (3)
+        + gas_costs.G_VERY_LOW  # MLOAD for salt (3)
         + gas_costs.G_VERY_LOW  # ADD for increment (3)
-        + gas_costs.G_LOW  # MSTORE salt back (3)
+        + gas_costs.G_VERY_LOW  # MSTORE salt back (3)
         + loop_condition_overhead  # While loop condition
     )
 
@@ -274,11 +274,11 @@ def test_bloatnet_balance_extcodecopy(
     setup_overhead = (
         gas_costs.G_COLD_ACCOUNT_ACCESS  # STATICCALL to factory (2600)
         + 100  # STATICCALL base cost
-        + gas_costs.G_BASE  # ISZERO (3)
+        + gas_costs.G_VERY_LOW  # ISZERO (3)
         + gas_costs.G_VERY_LOW  # PUSH2 (3)
-        + gas_costs.G_MID  # JUMPI (10)
-        + gas_costs.G_LOW * 2  # MLOAD × 2 for factory results (3 * 2)
-        + gas_costs.G_LOW * 3  # MSTORE × 3 for memory setup (3 * 3)
+        + gas_costs.G_HIGH  # JUMPI (10)
+        + gas_costs.G_VERY_LOW * 2  # MLOAD × 2 for factory results (3 * 2)
+        + gas_costs.G_VERY_LOW * 3  # MSTORE × 3 for memory setup (3 * 3)
         + gas_costs.G_VERY_LOW  # MSTORE8 for 0xFF prefix (3)
         + gas_costs.G_VERY_LOW  # PUSH1 for memory position (3)
     )
@@ -288,14 +288,14 @@ def test_bloatnet_balance_extcodecopy(
 
     # While loop condition overhead per iteration
     loop_condition_overhead = (
-        gas_costs.G_BASE  # DUP1 (3)
+        gas_costs.G_VERY_LOW  # DUP1 (3)
         + gas_costs.G_VERY_LOW  # PUSH1 (3)
         + gas_costs.G_VERY_LOW  # SWAP1 (3)
         + gas_costs.G_VERY_LOW  # SUB (3)
-        + gas_costs.G_BASE  # DUP1 (3)
-        + gas_costs.G_BASE  # ISZERO (3)
-        + gas_costs.G_BASE  # ISZERO (3)
-        + gas_costs.G_MID  # JUMPI (10)
+        + gas_costs.G_VERY_LOW  # DUP1 (3)
+        + gas_costs.G_VERY_LOW  # ISZERO (3)
+        + gas_costs.G_VERY_LOW  # ISZERO (3)
+        + gas_costs.G_HIGH  # JUMPI (10)
     )
 
     # Cost per contract with EXTCODECOPY and CREATE2 address generation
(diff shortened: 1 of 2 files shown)


jochem-brouwer:
I think we should update these tests to use `BenchmarkTest`, as this supports the logic which I think this PR also adds. See:

https://github.com/ethereum/execution-specs/blob/d618e5259a06179d7556af1961166cb7bc5da1e6/packages/testing/src/execution_testing/specs/benchmark.py#L393-L400

This works for tests if the transactions are just "copies" and there are no differences necessary per tx like changing calldata for instance (which in some cases we need to change as for instance calldata inputs to starting salts of the CREATE2 address calculator)

CC @LouisTsai-Csie

CPerezz:
> I think we should update these tests to use `BenchmarkTest`, as this supports the logic which I think this PR also adds. See:
> 
> https://github.com/ethereum/execution-specs/blob/d618e5259a06179d7556af1961166cb7bc5da1e6/packages/testing/src/execution_testing/specs/benchmark.py#L393-L400
> 
> This works for tests if the transactions are just "copies" and there are no differences necessary per tx like changing calldata for instance (which in some cases we need to change as for instance calldata inputs to starting salts of the CREATE2 address calculator)
> 
> CC @LouisTsai-Csie

- EXTCODESIZE/EXTCODECOPY/EXTCODEHASH attacks: The attack bytecode reads num_deployed_contracts from factory storage and iterates from salt 0 to that value. If we split into multiple identical transactions, each transaction would:
1. Start at salt 0
2. Access the same contracts again
3. Those contracts would be warm on subsequent transactions (defeating the purpose of cold access benchmarks)
- CALL attack: Same issue - iterates from salt 0 each time.
- SLOAD/SSTORE attacks (in test_single_opcode.py): These call ERC20 `balanceOf/approve` on a fixed set of contracts. The same contracts get accessed each transaction instead of varying

So, to elaborate, `BenchmarkTest.split_transaction` cannot be used for these benchmarks because:
- It creates identical transaction copies (same calldata)
- Each copy would access the same addresses starting from salt 0
- After the first transaction, those addresses are warm, not cold
- This would fundamentally change what's being benchmarked

The current manual transaction splitting approach in `test_multi_opcode.py` has the same issue though - the transactions are also identical. To properly benchmark cold access across multiple transactions, the attack contracts would need to:
- Accept a starting salt offset in calldata
- Each transaction would pass a different offset


JIC, can you confirm this makes sense @LouisTsai-Csie ?

jochem-brouwer:
Ah I see, yes I was giving this advice based on the diff, but as I now understand this has to be updated such that the attack contracts support a way to pick up the attack at a certain offset. Yes, for non-identical transactions the splitter cannot be used. However I think at some point we should include the calldata case where the transactions differ, since this is currently the go-to to make attacks start at a specific offset (so we could refactor later and add this feature to the transaction splitter, where txs have calldata based on some start offset and on an increase of this offset for each new tx).

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -154,23 +155,26 @@ def test_sload_empty_erc20_balanceof(
...
     )
     cold_warm_diff = (
         gas_costs.G_COLD_ACCOUNT_ACCESS - gas_costs.G_WARM_ACCOUNT_ACCESS
     )
 
+    # Calculate how many transactions we need to fill the block
+    num_txs = max(1, gas_benchmark_value // tx_gas_limit)
+
+    # Calculate gas budget per contract per transaction
+    available_gas_per_tx = tx_gas_limit - intrinsic_gas
```
danceratopz: I think we're missing the `overhead_per_contract * num_contracts` here that's present in `test_sstore_erc20_approve`?

In general, we could consider adding a helper (data)class here to help enforce more structure on these costs, which could also help with the calculation, something like:
```python
@dataclass
class LoopGasModel:
    """Gas model for benchmark attack loops."""

    overhead_per_contract: int
    """One-time cost per contract: counter init, memory expansion, JUMPDEST, cleanup."""

    cost_per_iteration: int
    """Per-iteration cost: loop body operations + condition check."""

    cold_warm_diff: int
    """Extra cost for first call vs subsequent (G_COLD - G_WARM)."""

    def calls_per_contract(
        self,
        available_gas: int,
        num_contracts: int,
    ) -> int:
        """Calculate how many calls fit per contract given available gas."""
        total_overhead = self.overhead_per_contract * num_contracts
        per_contract = (available_gas - total_overhead) // num_contracts
        return int((per_contract - self.cold_warm_diff) // self.cost_per_iteration)
```

I'm still digesting where this would fit in in terms of a potential larger refactor to add a more automated tx splitter as you discussed with Jochem above.

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -114,8 +117,9 @@ def test_bloatnet_balance_extcodesize(
 
     # Log test requirements - deployed count read from factory storage
     print(
-        f"Test needs {contracts_needed} contracts for "
-        f"{gas_benchmark_value / 1_000_000:.1f}M gas. "
+        f"Tx gas limit: {tx_gas_limit / 1_000_000:.1f}M gas. "
+        f"Number of txs: {num_txs}. "
```
danceratopz, line 146: We don't need to do this as part of this PR, but we could avoid some repetition to abstract some of this to a pytest fixture that calculates the `num_txs`, potentially some other auxillary variables and does (part of) the logging.

danceratopz (commented):
Hey, @CPerezz! Could you please take a look at the comment on the `overhead_per_contract` in `test_single_opcode`? Once that's addressed, I think we can merge as-is and address any refactors in a follow-up PR.

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
CPerezz: Open to any ideas! Could you file an issue for it? I'd be happy to take it!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
CPerezz: Good catch! Thanks!!!

CPerezz:
i think everything is addressed @danceratopz 

Could you merge if everything is indeed ok?

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -96,11 +97,13 @@ def test_bloatnet_balance_extcodesize(
...
-    available_gas = (
-        gas_benchmark_value - intrinsic_gas - 1000
-    )  # Reserve for cleanup
-    contracts_needed = int(available_gas // cost_per_contract)
+    # Calculate how many transactions we need to fill the block
+    num_txs = max(1, gas_benchmark_value // tx_gas_limit)
+
+    # Calculate how many contracts to access per transaction
+    # Reserve 1000 gas for cleanup
+    available_gas_per_tx = tx_gas_limit - intrinsic_gas - 1000
```
danceratopz: The `1000` seems arbitrary. Is it possible to make this more precise with reasonable effort? Also, this is only present for the tests in `test_multi_opcode` but not for `test_single_opcode`, was there a particular reason for that.

danceratopz (commented):
Thanks for the update @CPerezz! Excuse me being pedantic, but I was going to suggest some clean-up and came across the inconsistency below.

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
danceratopz: Ah this is related to the `overhead_per_contract` again :+1:

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -79,28 +80,56 @@ def test_bloatnet_balance_extcodesize(
...
+    # While loop condition overhead per iteration
+    loop_condition_overhead = (
+        gas_costs.G_BASE  # DUP1 (3)
+        + gas_costs.G_VERY_LOW  # PUSH1 (3)
+        + gas_costs.G_VERY_LOW  # SWAP1 (3)
+        + gas_costs.G_VERY_LOW  # SUB (3)
+        + gas_costs.G_BASE  # DUP1 (3)
+        + gas_costs.G_BASE  # ISZERO (3)
+        + gas_costs.G_BASE  # ISZERO (3)
+        + gas_costs.G_MID  # JUMPI (10)
```
danceratopz:
```suggestion
        gas_costs.G_VERY_LOW  # DUP1 (3)
        + gas_costs.G_VERY_LOW  # PUSH1 (3)
        + gas_costs.G_VERY_LOW  # SWAP1 (3)
        + gas_costs.G_VERY_LOW  # SUB (3)
        + gas_costs.G_VERY_LOW  # DUP1 (3)
        + gas_costs.G_VERY_LOW  # ISZERO (3)
        + gas_costs.G_VERY_LOW  # ISZERO (3)
        + gas_costs.G_HIGH  # JUMPI (10)
```

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -235,27 +270,57 @@ def test_bloatnet_balance_extcodecopy(
...
```
danceratopz:
```suggestion
        gas_costs.G_VERY_LOW  # DUP1 (3)
        + gas_costs.G_VERY_LOW  # PUSH1 (3)
        + gas_costs.G_VERY_LOW  # SWAP1 (3)
        + gas_costs.G_VERY_LOW  # SUB (3)
        + gas_costs.G_VERY_LOW  # DUP1 (3)
        + gas_costs.G_VERY_LOW  # ISZERO (3)
        + gas_costs.G_VERY_LOW  # ISZERO (3)
        + gas_costs.G_HIGH  # JUMPI (10)
```

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -642,13 +747,16 @@ def test_mixed_sload_sstore(
...
     )
 
-    # Calculate gas budget per contract
-    available_gas = gas_benchmark_value - intrinsic_gas
-    gas_per_contract = available_gas // num_contracts
+    # Calculate how many transactions we need to fill the block
+    num_txs = max(1, gas_benchmark_value // tx_gas_limit)
+
+    # Calculate gas budget per contract per transaction
+    available_gas_per_tx = tx_gas_limit - intrinsic_gas
```
danceratopz: We're missing the `overhead_per_contract` here.

danceratopz (commented):
Hi @CPerezz, I think some of the gas constants used in the calculations and comments are incorrect. This is a real pain - we need to absolutely try to remove as much of this accounting burden from test authors as possible. I thought it might be easier to PR my suggestions here:
- https://github.com/CPerezz/execution-specs/pull/1

Otherwise, could you please address the remaining missing contract overhead. Sorry that these have landed in drips and drabs!

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
danceratopz: Thanks for addressing this!

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
danceratopz: I will make a follow-up issue!

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
CPerezz: @danceratopz it should be merged and ready to be merged here!

danceratopz: I think we still need these changes @CPerezz?

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
danceratopz: Follow-up issue:
- https://github.com/ethereum/execution-specs/issues/1991

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
danceratopz: https://github.com/ethereum/execution-specs/issues/1991

CPerezz: My bad.. 3 PRs at the sime time is not nice to handle hahaha.

bf5724a should handle the remaining pointed things.

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -642,13 +767,30 @@ def test_mixed_sload_sstore(
...
+    overhead_per_contract = (
+        gas_costs.G_VERY_LOW  # MSTORE to initialize counter (3)
+        + gas_costs.G_JUMPDEST  # JUMPDEST at loop start (1)
+        + gas_costs.G_VERY_LOW  # MLOAD for While condition check (3)
+        + gas_costs.G_BASE  # ISZERO (2)
+        + gas_costs.G_BASE  # ISZERO (2)
+        + gas_costs.G_MID  # JUMPI (8)
+        + gas_costs.G_BASE  # POP to clean up at end (2)
+    )
+
```
danceratopz:
```suggestion
```

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -79,28 +80,56 @@ def test_bloatnet_balance_extcodesize(
     # Calculate gas costs
     intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(calldata=b"")
 
+    # Setup overhead (before loop): STATICCALL + result handling + memory setup
+    setup_overhead = (
+        gas_costs.G_COLD_ACCOUNT_ACCESS  # STATICCALL to factory (2600)
+        + 100  # STATICCALL base cost
```
danceratopz:
```suggestion
```

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -235,27 +270,57 @@ def test_bloatnet_balance_extcodecopy(
     # Calculate costs
     intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(calldata=b"")
 
+    # Setup overhead (before loop): STATICCALL + result handling + memory setup
+    setup_overhead = (
+        gas_costs.G_COLD_ACCOUNT_ACCESS  # STATICCALL to factory (2600)
+        + 100  # STATICCALL base cost
```
danceratopz:
```suggestion
```

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -396,28 +467,56 @@ def test_bloatnet_balance_extcodehash(
     # Calculate gas costs
     intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(calldata=b"")
 
+    # Setup overhead (before loop): STATICCALL + result handling + memory setup
+    setup_overhead = (
+        gas_costs.G_COLD_ACCOUNT_ACCESS  # STATICCALL to factory (2600)
+        + 100  # STATICCALL base cost
```
danceratopz:
```suggestion
```

danceratopz (commented):
Hi @CPerezz, let's try and get closure on this PR! 👍 

Looks like we both added commits to address the missing `overhead_per_contract`. That's the main remaining thing to address in the scope of this PR. I would suggest we remove the second one added in bf5724a4d, could you please check the suggestion below that removes it?

I don't want to be overcritical, but unfortunately there are still issues in the gas accounting for overhead cost estimation (some pre-date this PR). One obvious offender  is the superfluous `100` gas for the non-existant STATICCALL base cost. 

I'm not sure how precise these calculations need be for these tests, which target state growth (the attack is still valid). We should not address them in this PR, but if you think we should nail down exact costs, we could create a follow-up issue to review them. Perhaps you can pass comment on how important these costs are? A quick explanation in the next benchmarking call would be great. Thanks!

danceratopz:
@CPerezz just a follow-up comment and heads up about the gas accounting. @marioevz just proposed the following to help simplify gas cost calculations:
- https://github.com/ethereum/execution-specs/pull/2002

I would merge this PR with the suggestions above and then add an issue to refactor the gas accounting with #2002, once it's merged. Happy to help or even take the lead on that. How does that sound?

danceratopz (approved):
Thanks for adding this @CPerezz!

PR #1971, opened 2026-01-05
feat(benchmarks): fix bytecode attack for EXT-like opcodes to work in Osaka

jsign:
## 🗒️ Description
This PR builds on the work in https://github.com/ethereum/execution-specs/pull/1850 to fix the remaining `EXT*` mechanisms for accessing unchunkified bytecode.

After thinking it over, it makes sense to unify the benchmarks as initially done. The benchmark code is the same. I also think that the primary goal of the benchmark is not "generic usage" of these opcodes, but specifically on the unchunkified bytecode scenario, so I changed the benchmark name accordingly.

If this isn't convincing, I can keep them split and extract the whole internal logic to some other place and use it literally for both EXT* and CALL*. I just think that's quite awkard and why I ended up unifying and deciding on a more meaningful name on the scenario being benchmarked.

## 🔗 Related Issues or PRs

## ✅ Checklist

    ```console
    uvx tox -e static
    ```


LouisTsai-Csie:
Hi @jsign, i agree with your point about the unchunkified scenario, putting these opcodes together makes sense from a benchmark perspective, since they are testing very similar behavior.

One possible option is to move related test into` tests/benchmark/compute/scenario` and named as you proposed, `test_unchunkified_bytecode`. The existing files may not really fit this scenario, feel free to add a new one if necessary.

jsign:
@LouisTsai-Csie, done!

LouisTsai-Csie (approved):
LGTM!

PR #1937, opened 2025-12-17
feat(test-benchmark): add worst-case depth attack benchmarks for Ethereum state tries

CPerezz:
> This PR is a non-deployment-required refactor of #1936 such that we can discuss what is the best approach by comparing the 2 of them.

## Summary

- Add benchmark tests for worst-case depth attacks on Ethereum state and account tries
- Implement `AttackOrchestrator` contract for efficient batched SSTORE attacks via CREATE2 address derivation
- Add `Verifier` contract for post-attack storage verification
- Include pre-mined CREATE2 data for storage depths 9-10 and account depths 3-5

## Description

This PR introduces benchmark tests that measure the worst-case performance impact when modifying deep storage slots across thousands of contracts. The attack pattern:

1. **Pre-deployed contracts** with deep storage tries maximize trie traversal costs
2. **CREATE2-based addressing** enables deterministic contract addresses without on-chain deployment during test execution
3. **Optimized batched attacks** using `AttackOrchestrator` executes up to 1,980 attacks per transaction at 16M gas limit
4. **Verifier contract** confirms attack success by checking the deepest storage slot values

Key implementation details:
- Gas calculations empirically measured: ~8,050 gas per attack (2,750 overhead + 5,300 forwarded)
- Uses Nick's deterministic deployer (`0x4e59...956c`) for CREATE2 deployments
- Supports configurable storage depth (9-10) and account depth (3-5) via pre-mined JSON files

Deployment setup and instructions available at:
https://gist.github.com/CPerezz/44d521c0f9e6adf7d84187a4f2c11978

## Test Plan

- [x] Run `uv run execute remote` with `--gas-benchmark-values 60` to verify 7,104 contracts attacked successfully
- [x] Verify storage values changed from `0x01` to `0x2a` (42) via Verifier contract
- [ ] Run on production-grade client to measure actual performance impact

cc: @LouisTsai-Csie @marioevz

PR #1976, opened 2026-01-06
feat(test-benchmark): add worst-case depth attack benchmarks for Ethereum state tries using deterministic deploy

marioevz:
## 🗒️ Description
Based on #1937 but using #1934.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -20,19 +20,17 @@
 - Gas per attack call: ~8,050 gas (~2,742 overhead + 5,300 forwarded)
 
 Contract sources:
-- AttackOrchestrator.sol and Verifier.sol:
-  https://gist.github.com/CPerezz/8686da933fa5c045fbdf7c31e20e6c71
```
CPerezz: Why were those removed?? Just curious as they are indeed the contracts used within this test to perform the attack and verify the execution

CPerezz (commented):
I have been thinking on some of the suggestions from @jochem-brouwer in 
#1976 and I think some of them apply here. Would you like me to address them?

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
marioevz: For `AttackOrchestrator.sol`, Solidity was using more opcodes than necessary and made it difficult to see which ones were actually being used, which in turn made it difficult to estimate gas.
For `Verifier.sol`, we could do the verification via the `post` object in the test, so no need to create another contract and call it and use gas that could be instead used by the attacks.

docs/running_tests/execute/remote.md
```diff
@@ -32,6 +32,8 @@ These transactions are created from the seed or worker accounts provided via the
 
 When the test is executed, all `pre.fund_eoa` and `pre.deploy_contract` calls generate transactions that create the accounts on chain, instead of placing them directly in the `pre` object like `fill` does.
 
+If a test uses `pre.deterministic_deploy_contract`, the `execute` command first checks if the contract is already present on chain before attempting to deploy it. If the proxy used for deterministic deployment is not present on chain either, the command will automatically deploy it too. The address of these contracts is calculated using `keccak256( 0xff ++ 0x4E59B44847B379578588920CA78FBF26C0B4956C ++ salt ++ keccak256(init_code))[12:]`.
```
jochem-brouwer: This address looks like a magical address, could the context of https://github.com/Arachnid/deterministic-deployment-proxy/tree/master be added here for completeness? 😄

packages/testing/src/execution_testing/test_types/helpers.py
```diff
@@ -19,6 +19,13 @@
 Helper functions
 """
 
+DETERMINISTIC_DEPLOYMENT_CONTRACT_ADDRESS = Address(
+    0x4E59B44847B379578588920CA78FBF26C0B4956C
+)
+DETERMINISTIC_DEPLOYMENT_BYTECODE = Bytes(
+    "0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3"
+)
```
jochem-brouwer, line 48: I think this also should have a reference to https://github.com/Arachnid/deterministic-deployment-proxy/tree/master for context.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -255,6 +265,160 @@ def code_pre_processor(
...
+                    gas_price=0x174876E800,
+                    gas_limit=0x0186A0,
+                    to=None,
+                    value=0,
+                    data=Bytes(
+                        "0x604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3"
+                    ),
+                    v=0x1B,
+                    r=0x2222222222222222222222222222222222222222222222222222222222222222,
+                    s=0x2222222222222222222222222222222222222222222222222222222222222222,
```
jochem-brouwer: Could also benefit here to link the source of these ""magic"" values, e.g. https://github.com/Arachnid/deterministic-deployment-proxy/blob/be3c5974db5028d502537209329ff2e730ed336c/scripts/compile.ts#L87

jochem-brouwer:
@CPerezz I think you are refering to #1961 and you are right, those indeed apply here. I'll write comments on it and will publish a review in a hour or so :smile: :+1:

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,467 @@
...
+    end: int
+    initcode_hash: Hash
+
+    def calldata(self) -> bytes:
+        """Return the calldata that needs to be passed to the orchestrator."""
+        return Bytes(
+            self.value.to_bytes(32, "big")
+            + self.start.to_bytes(32, "big")
+            + self.end.to_bytes(32, "big")
+            + self.initcode_hash
```
jochem-brouwer: I believe all 4 of these can be removed from calldata as with the method in #1961. This also raises the point for a further refactor short- to mid-term where we create a template for the following pattern:

We execute certain code with one input argument: the target address. This target address is a target of a CREATE2 factory. This is done in a loop with the exit condition: if the gas is below some threshold then exit the loop. The salt is increased every time the loop is re-ran. After the loop, the next salt is stored into storage. The initial salt is read from this storage.

This has the following benefits:

1. We do not have to calculate the start salt and the end salt here. To calculate the start and end salt we need to know the gas costs of each loop (which are usually dynamic and hard to calculate)
2. Encoding this data into calldata makes the transactions parallelizable. Using storage this is not possible, because the next transaction depends on the initial storage slot left to the previous one.
3. Although we get slightly waste of gas in the EVM, this is negligible compared to the operations we use in these tests (storage writes)

The `value` here can be read from the attack contract (this seems to be the target key to write. Small note: the usage of the name "value" here is confusing, because it could also point to tx value e.g. `CALLVALUE`). Start/end bytes are handled by the contract (in EVM). `initcode_hash` is also a constant and is something we can hardcode in the contract (no need to put it in calldata)

What to do is to take the contract from #1961: these lines: https://github.com/ethereum/execution-specs/pull/1961/changes#diff-88ac263a5a41126dcb0c95cc6939a105f972f0a9fd526ecaae4f085f01f96d0aR118-R152 and edit it such that it hardcodes the initcode, and changes the EXTCODESIZE in the loop to Op.CALL which calls the `attack(uint256)` here.

jochem-brouwer: Note: this CREATE2 attack pattern is something we have seen in many places (and also in some small variations, with the same end goal) - so we should template this attack at some point so we can re-use it and iterate faster using the same code 😄 👍

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,467 @@
...
+        )
+
+    def calculate_inner_call_cost(self, fork: Fork) -> int:
```
jochem-brouwer: Note: with the new format this can thus be removed as we let EVM handle the logic if we want to do "one more loop" or if we want to exit

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,467 @@
...
+            + 2 * gas_costs.G_BASE  # PUSH0 operations
+            + 1
+            * (
+                gas_costs.G_COLD_SLOAD + gas_costs.G_STORAGE_RESET
+            )  # SSTORE operations
+        )
+        return inner_call_cost
+
+    def calculate_gas(self, fork: Fork) -> int:
+        """Calculate the exact gas this attack transaction will use."""
```
jochem-brouwer: Same here, gas calculations are not necessary anymore with this other logic

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,467 @@
...
+        inner_call_cost = self.calculate_inner_call_cost(fork)
+        return gas_cost + ((inner_call_cost * 64 // 63) - inner_call_cost)
+
+    def generate_transaction(self, fork: Fork, sender: EOA) -> Transaction:
+        """Generate the transaction to perform the attack."""
+        return Transaction(
+            to=ATTACK_ORCHESTRATOR_ADDRESS,
+            gas_limit=self.calculate_tx_gas_limit(fork),
+            sender=sender,
+            data=self.calldata(),
```
jochem-brouwer: If we switch here to new format, calldata is empty, and can thefore use BenchmarkTestFiller to handle the split logic

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,467 @@
...
+        )
+
+    def add_post_verification(
+        self, post: Alloc, mined_contract_file: MinedContractFile
+    ) -> None:
+        """Add the post-verification transaction to the post-state."""
```
jochem-brouwer: The post verification from #1961: 

- Writes current salt to slot 0
- Writes EXTCODESIZE of target to slot 1.

The EXTCODESIZE check verifies that the target contract is deployed. Additionally, can also verify that slot 0 is "at least some number" such that we know at least X amount of attacks have ran. 
(Note: if tx fails or OOGs these slots could thus be written but are then reverted)

jochem-brouwer (changes requested):
@CPerezz is I think referring to #1961. I think the format there cleans up the test and makes it easier to reason about. I left comments and pointers to there.

The main thing I think we should do ASAP is wrap this attack template in a helper file somewhere. So attacking specific contracts using CREATE2 factory is a common pattern, however the attack itself (use an EVM opcode on them, or CALL them with certain data) differs.

(If something is unclear let me know 😄 👍 Happy to help or give pointers)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
CPerezz: This should not be here right? This change should be in another PR where we add all the stuff. I asume after rebasig this  should dissapear?

docs/running_tests/execute/remote.md
marioevz: Added to #1934, thanks!

packages/testing/src/execution_testing/test_types/helpers.py
marioevz: Added comment to #1934 too.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Added to https://github.com/ethereum/execution-specs/pull/1934 too.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
marioevz: The downside of the code created in `build_attack_contract` is that we have to deploy a new fresh contract each time we run `execute` again, because it stores the last salt to storage so we cannot start from zero when the test is executed from the beginning.

This version can reuse the same contract not only across re-execution of the same test but in in all tests (as long as the factory pre-deploy address is the same).

I think the slightly higher calldata cost is worth it in order to speed up test execution.

LouisTsai-Csie:
Hi @marioevz , could we continue on this refactor by using the gas cost calculator?

marioevz:
@LouisTsai-Csie @CPerezz I've pushed a new and much simplified version using the latest tools that we have implemented, including `IteratingBytecode` and `Create2PreimageLayout`. Please take a look and let me know what you think.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,370 @@
...
+
+    def model_post_init(self, __context: Any) -> None:
+        """
+        Perform post-initialization checks.
+        """
+        if len(self.contracts) != self.num_contracts:
+            raise ValueError(
+                f"Number of storage keys ({len(self.storage_keys)}) does "
+                f"not match number of contracts ({self.num_contracts})"
+            )
```
CPerezz, line 122: The condition checks `len(self.contracts) !=
  self.num_contracts` but the error message reports
  `len(self.storage_keys)`.

  This will print completely wrong values when triggered making debugging harder

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,370 @@
...
+    """
+    Model to load information about a contract mined using using
+    https://github.com/CPerezz/worst_case_miner.
+    """
+
+    deployer: Address
+    initcode_hash: Hash = Field(..., alias="init_code_hash")
+    initcode: Bytes = Field(..., alias="init_code")
+    deploy_code: Bytes
+    storage_keys: List[Address]
```
CPerezz:
```suggestion
    storage_keys: List[StorageKey]
```

Don't we have this struct?

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,370 @@
...
+    ):
+        return int(time.time())
+    return DEFAULT_ATTACK_VALUE
+
+
+@pytest.mark.valid_from("Prague")
+@pytest.mark.parametrize(
+    "storage_depth,account_depth",
+    [
+        (10, 6),  # From .worst_case_miner/mined_assets
```
CPerezz: There are a lot more combinations available. We should add them all if possible. I think I only left 1 so that I could test easily

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,370 @@
...
+    1. Derives CREATE2 addresses from initcode_hash + Nick's deployer
+    2. Deploys AttackOrchestrator that calls attack() on each target
+    3. Fills blocks with 16M gas transactions attacking contracts
+    4. Adds a verification transaction at the end to confirm success
+
+    Args:
+        blockchain_test: The blockchain test filler
+        pre: Pre-state allocation
+        fork: The fork to test on
+        env: Environment object that will be used to fill/execute
```
CPerezz: Unsure this is an argument we need to use?? Is this stale?

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,370 @@
...
+    """
+
+    salt: int
+    contract_address: Address
+    auxiliary_accounts: List[Address]
+
+
+class MinedContractFile(BaseModel):
+    """
+    Model to load information about a contract mined using using
```
CPerezz:
```suggestion
    Model to load information about a contract mined using
```

CPerezz (commented):
Dunno much about Python.

But aren't we missing `__init__.py` in `depth_benchmarks/`

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
marioevz: Could you share a list of the ones you'd like added in this PR please?

CPerezz: https://github.com/CPerezz/worst_case_miner/tree/master/mined_assets

All of the ones that combine depth of account and storage.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
marioevz: We use `Hash`, but it's a bit inconvenient because the hex strings in the JSON files have a length of 20 bytes, so we have to left-pad here.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
marioevz: Added all combinations in the latest commit:
```python
        (10, 3),
        (10, 4),
        (10, 5),
        (10, 6),
        (10, 7),
        (11, 3),
        (11, 4),
        (11, 5),
        (11, 6),
        (11, 7),
        (12, 3),
        (12, 4),
        (12, 5),
```

marioevz: I've added these combinations but I noticed that some of the mined asset files did not contain the storage keys and thus the tests were failing.

I've created this PR as prerequisite: https://github.com/CPerezz/worst_case_miner/pull/5

marioevz:
@CPerezz I've addressed all the comments, but we need the PR in the [worst_case_miner](https://github.com/CPerezz/worst_case_miner) repo to ensure all mined assets contain the storage keys as a list in the JSON (some of them didn't have it).

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,385 @@
...
+        # Salt + 1 < End Index
+        condition=Op.LT(
+            Op.MLOAD(create2_preimage.salt_offset),
+            Op.DUP1,
+        ),
+    )
+    cleanup = Op.STOP
+
+    # The code was compiled by solidity, so these opcode counts were obtained
+    # from the traces.
```
LouisTsai-Csie, line 269: Could we add a comment clarifying that this is the Solidity function dispatching pattern & function logic?

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -0,0 +1,385 @@
...
+            start_iteration=0,
+            sender=sender,
+            to=attack_orchestrator_address,
+            calldata=calldata,
+        )
+    )
+
+    total_gas_cost = sum(tx.gas_cost for tx in attack_txs)
+
+    blockchain_test(
```
LouisTsai-Csie:
```suggestion
   benchmark_test(
```

LouisTsai-Csie (changes requested):
Thanks, left small changes.

Have you tried running this benchmark? We could ask Kamil to run it as well, in case there are any issues with the implementation.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
marioevz: Yes, I'm actually not 100% sure which part of the solidity code is actually being executed, that's an inherent problem of including solidity code in our tests, but we can write a small description of our best guess.

marioevz:
Ready for another (hopefully last one) review :) @LouisTsai-Csie

LouisTsai-Csie (approved):
LGTM!! Please let @jsign know when this is merged, he needs this benchmark for zkEVM also

PR #1985, opened 2026-01-07
feat(test-benchmark): updates and fixes for fixed opcode count

spencer-tb:
## 🗒️ Description

This PR fixes several bugs and adds features for `--fixed-opcode-count`.

#### Features
- Use longest match for pattern specificity with regex, more detail below.
- Add `-m repricing` marker to all precompile tests, apart from select few not required or broken.
- Remove `test_mod` from repricing as not compatible with foc (fixed-opcode-count).
- Adds unit tests for `OpcodeCountsConfig` pattern matching and validation.
- Support float values for sub-1K opcode counts (e.g., `0.25` = 250 opcodes, `0.5` = 500 opcodes).
  
#### Bug Fixes
- Raise `UsageError` when the fixed opcode count config file (`.fixed_opcode_counts.json`) is missing.
- Validate flag input to reject test paths accidentally consumed by argparse.
- **_Major_** - Fix config file pattern matching for parametrized tests, patterns like `test_bitwise.*AND.*` now correctly match against full test names instead of `test_bitwise.*` only.

#### Floats As Inputs
For fixed opcode count values < 1.0 (e.g., 0.25 = 250 opcodes), the inner iterations are set to the exact count with `outer = 1`, enabling precise low-count benchmarks.

#### Correct Pattern Matching With Config
Pattern matching now works with full test names, given both `"test_bitwise.*": [1]` & `"test_bitwise.*AND.*": [100]`:
- Before: All `test_bitwise` variants got 1K, pattern matches function name only.
- After: `test_bitwise[fork_Prague-opcode_AND]` gets 100K, others get 1K, pattern matches simulated full test ID so get different counts per parameter.

## 🔗 Related Issues or PRs

Follow ups and fixes from #1790.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

**Venusaur - 0003**

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -41,7 +41,12 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         const="",
         help=(
             "Opcode counts (in thousands) for benchmark tests. "
-            "Example: '1,10,100' runs tests with 1K, 10K, 100K opcodes. "
+            "Granularity rules (for ≤10%% CALL overhead): "
+            "cheap ops (1-2 gas): integers only, no sub-1K; "
+            "medium ops (3-5 gas): 0.5 increments, min 0.5K; "
+            "expensive ops (6+ gas): 0.25 increments, min 0.25K; "
+            "very expensive (100+ gas): 0.25 increments, min 0.01K. "
+            "Example: '0.5,1,2' runs 500, 1K, 2K opcodes. "
             "Without value, uses .fixed_opcode_counts.json config. "
             f"Cannot be used with {GasBenchmarkValues.flag}."
         ),
@@ -134,15 +139,13 @@ class GasBenchmarkValues(RootModel, BenchmarkParametrizer):
 
     @classmethod
     def from_parameter_value(
-        cls, config: pytest.Config, value: str
+        cls, _config: pytest.Config, value: str
     ) -> Self | None:
         """Given the parameter value and config, return the expected object."""
-        del config
         return cls.model_validate(value.split(","))
 
-    def get_test_parameters(self, test_name: str) -> list[ParameterSet]:
+    def get_test_parameters(self, _test_name: str) -> list[ParameterSet]:
         """Get benchmark values. All tests have the same list."""
-        del test_name
         return [
             pytest.param(
                 gas_value * 1_000_000,
@@ -155,8 +158,9 @@ class GasBenchmarkValues(RootModel, BenchmarkParametrizer):
 class OpcodeCountsConfig(BaseModel, BenchmarkParametrizer):
     """Opcode counts configuration object."""
 
-    scenario_configs: Dict[str, List[int]] = Field(default_factory=dict)
-    default_counts: List[int] = Field(default_factory=lambda: [1])
+    scenario_configs: Dict[str, List[float]] = Field(default_factory=dict)
+    default_counts: List[float] = Field(default_factory=lambda: [1.0])
+    uses_config_file: bool = Field(default=False)
 
     default_config_file_name: ClassVar[str] = ".fixed_opcode_counts.json"
     flag: ClassVar[str] = "--fixed-opcode-count"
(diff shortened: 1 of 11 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -169,43 +170,141 @@ def from_parameter_value(
...
                     f"{cls.default_config_file_name} was not found. "
                     "Run 'uv run benchmark_parser' to generate it, or provide "
                     "explicit values (e.g., --fixed-opcode-count 1,10,100)."
                 )
-        return cls.model_validate({"default_counts": value.split(",")})
+        # Validate that value looks like comma-separated integers
+        # This catches the case where argparse greedily consumes a test path
+        parts = value.split(",")
+        for part in parts:
+            if not part.strip().isdigit():
```
LouisTsai-Csie: just personal idea!
```suggestion
        if not all(part.strip().isdigit() for part in parts):
```

LouisTsai-Csie (changes requested):
Some initial comments, will continue on more feedback:
- The required precompile benchmarks are already labeled with repricing marker, we only label these with necessary configurations. So we could remove the repricing marker in precompiles benchmark you add.

We need to update the benchmark_parser also, or it would override some of the setting. For example, this is valid in current logic: `test_push*` but it would be removed in the parser logic, as parser always find all the entry for the fixed-opcode-count compatible test.

spencer-tb:
@LouisTsai-Csie I updated the parser and removed some precompiles: Modexp, BLS381, BN128 - as followed on discord.

tests/benchmark/compute/precompile/test_ecrecover.py
LouisTsai-Csie: Same comment, if the marker is applied to all the benchmarks, then i would remove this one.

tests/benchmark/compute/instruction/test_log.py
```diff
@@ -21,11 +21,11 @@
...
-        pytest.param(Op.LOG0, id="log0"),
-        pytest.param(Op.LOG1, id="log1"),
-        pytest.param(Op.LOG2, id="log2"),
-        pytest.param(Op.LOG3, id="log3"),
-        pytest.param(Op.LOG4, id="log4"),
+        pytest.param(Op.LOG0),
+        pytest.param(Op.LOG1),
+        pytest.param(Op.LOG2),
+        pytest.param(Op.LOG3),
+        pytest.param(Op.LOG4),
```
LouisTsai-Csie:
```suggestion
        Op.LOG0, Op.LOG1, Op.LOG2, Op.LOG3, Op.LOG4
```

I think we could refactor a bit here.

tests/benchmark/compute/instruction/test_log.py
```diff
@@ -88,11 +88,11 @@ def test_log(
...
```
LouisTsai-Csie:
```suggestion
        Op.LOG0, Op.LOG1, Op.LOG2, Op.LOG3, Op.LOG4
```

Same as prev.

tests/benchmark/compute/precompile/test_ecrecover.py
LouisTsai-Csie:
```suggestion
           id="ecrecover",
```

we could remove duplicate marker here.

tests/benchmark/compute/precompile/test_p256verify.py
LouisTsai-Csie, line 35:
```suggestion
```

if we want to add `repricing` marker to the entire test, then we could remove this marker here.

tests/benchmark/compute/precompile/test_point_evaluation.py
LouisTsai-Csie:
```suggestion
```

Same - remove duplicate marker

tests/benchmark/README.md
```diff
@@ -0,0 +1,298 @@
...
+uv run fill --fork Prague --gas-benchmark-values 1 -m benchmark ./tests/benchmark --evm-bin=evmone-t8n
+```
+
+This mode is useful for measuring how much work can be done within a given gas budget, simulating real-world block execution conditions.
+
+## Fixed Opcode Count
+
+The fixed opcode count mode runs benchmark tests with a predetermined number of opcode iterations rather than gas-based limits. This approach enables rapid iteration when analyzing gas costs for repricing proposals, as you can directly compare execution times across different opcode counts.
+
+**Important:** Tests must be marked with `@pytest.mark.repricing` to be compatible with fixed opcode count mode. This marker identifies tests that have been specifically designed for gas repricing analysis with proper code generators.
```
LouisTsai-Csie: This is only partially correct. The fixed-opcode-count feature is not limited to tests marked with the repricing marker.

Q: Which benchmark formats support the fixed-opcode-count feature?
A: Any test that uses the benchmark test wrapper (`benchmark_test`) together with a code generator (`code_generator`) can support the fixed-opcode-count feature, since we can generate the required contract logic dynamically.

Q: Why do we use the repricing marker?
A: During gas repricing analysis, Maria is typically interested in a specific subset of configurations. For example, in `test_calldatasize`, we assume that calldata size is the primary factor affecting performance, so we don’t focus on the `zero_data` parameter. The repricing marker allows us to configure and narrow down the relevant benchmark cases. We apply this same approach to other tests to limit the scope of benchmark runs.

The current implementation does **not** restrict the fixed-opcode-count feature to tests labeled with the repricing marker.

https://github.com/ethereum/execution-specs/blob/180dcecd7deb336b5a33c957bd860a5e98c3fab9/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py#L210

Summary: atm, we typically run the fixed-opcode-count feature together with the repricing marker, but this is a usage choice, not a technical limitation. From an implementation perspective, fixed-opcode-count benchmarks can run independently of the repricing marker.

LouisTsai-Csie: Example: below both work

```python
uv run fill -v --clean --gas-benchmark-values 30 tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance -m benchmark

uv run fill -v --clean --gas-benchmark-values 30 tests/benchmark/compute/instruction/test_account_query.py::test_selfbalance -m repricing
```

tests/benchmark/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+```bash
+# Generate or update .fixed_opcode_counts.json
+uv run benchmark_parser
+
+# Validate that config is in sync
+uv run benchmark_parser --check
+```
+
+The parser preserves any custom counts you've configured while adding new tests with default values.
```
LouisTsai-Csie: Suggestions:
- Move this section before the `Config File Format` section, so readers are aware of the tool before they start manually adding configuration entries.
- Provide more details on the override rules used by `benchmark_parser`, so users can understand whether their existing configurations will be preserved or overridden.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -169,48 +175,174 @@ def from_parameter_value(
...
+                self.get_test_parameters(test_name),
+                scope="function",
+            )
+        else:
+            # CLI mode: use function name matching (original behavior)
+            metafunc.parametrize(
+                self.parameter_name,
+                self.get_test_parameters(test_name),
+                scope="function",
+            )
```
LouisTsai-Csie:
```suggestion
        elif self.uses_config_file:
            # Config file mode, no existing params: match against function name
            metafunc.parametrize(
                self.parameter_name,
                self.get_test_parameters(test_name),
                scope="function",
            )
```

The `else` statement duplicate the logic of the one in `elif` statement

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -266,8 +402,60 @@ def pytest_collection_modifyitems(
...
+            continue
+
+        # Build simulated test ID WITHOUT the opcode count for pattern matching
+        # The test name format is: test_func[fork_X-fixture_format-params-opcount_Y]
+        # We need: test_func[fork_X-fixture_format-params]
+        test_name = item.name
+
+        # Remove the opcode count part from the test ID for pattern matching
+        # Pattern: -opcount_X.XK or -opcount_XK at the end before ]
+        import re
```
LouisTsai-Csie:
```suggestion
```
Already imported from the top

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -193,9 +196,40 @@ def generate_repeated_code(
         #
         # 2a. If N is 1000: Set M = fixed_opcode_count. (Total ops: fixed_opcode_count * 1000)
         # 2b. If N is 500: Set M = fixed_opcode_count * 2. (Total ops: (fixed_opcode_count * 2) * 500 = fixed_opcode_count * 1000)
```
LouisTsai-Csie: After review the logic below, i think the current fallback iteration is not 500 but 250, would be nice to be consistent here. It might be worth checking the entire comment again.

LouisTsai-Csie (changes requested):
Huge thanks for the help!! I have some comment below and wants to discuss several points:

- For the `README` file, could we integrate it into existing docs under the [benchmark](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/docs/writing_tests/benchmarks.md) section?

I have one more files to review, the `benchmark_parser.py` one, will udpated soon

spencer-tb:
@LouisTsai-Csie I am going to remove all the documentation from this PR. How about we merge your docs PR: https://github.com/ethereum/execution-specs/pull/2100, then I can add a follow that includes fixed opcode count among other things?

LouisTsai-Csie:
@spencer-tb sounds good!

LouisTsai-Csie (approved):
LGTM, thanks for the update

PR #2002, opened 2026-01-10
feat(testing/forks): Implement `bytecode.gas_cost(fork)`

marioevz:
## 🗒️ Description
Implements `bytecode.gas_cost(fork)` to calculate the exact gas cost of a bytecode for a given fork.

This is done by adding "metadata" to the opcodes that are used within bytecode.

Example:
```python
expand_memory_code = Op.MSTORE8(
    # Expand memory first
    offset=deposited_len - 1,
    value=0,
    new_memory_size=deposited_len,  # For gas accounting
)
sstore_code = Op.SSTORE(0, 1, key_warm=True)
return_code = Op.RETURN(
    offset=0,
    size=deposited_len,
    code_deposit_size=deposited_len,  # For gas accounting
)
initcode = expand_memory_code + sstore_code + return_code
initcode.gas_cost(fork)  # Returns the total gas cost
```
In this example `Op.MSTORE8` now contains `new_memory_size` metadata to include the memory expansion calculation when `bytecode.gas_cost` operates.

Also `Op.RETURN` includes a `code_deposit_size` to signal that this opcode is supposed to operate within initcode and returns bytecode that creates a contract, because otherwise the cost would be zero.

### Updated tests

Includes updates to the following tests:
- tests/berlin/eip2929_gas_cost_increases/test_call.py
- tests/cancun/eip5656_mcopy/test_mcopy_memory_expansion.py
- tests/frontier/create/test_create_deposit_oog.py
- tests/frontier/opcodes/test_all_opcodes.py
- tests/frontier/opcodes/test_exp.py
- tests/frontier/opcodes/test_log.py
- tests/osaka/eip7939_count_leading_zeros/test_count_leading_zeros.py

These were ported in order to debug and validate the gas calculations in the bytecodes included in them.

### Bytecode retains list of opcodes

The bytecode instances now include a list of all opcodes they contain.

It was done by creating an Opcode prototype to avoid circular dependencies.

This list is used to calculate the gas cost of each opcode.

### `fork.opcode_gas_calculator`

The fork base type now contains a method that returns a calculator which returns the gas cost of a given opcode, which inspects the opcode object "metadata" to be able to determine dynamic gas calculation.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/forks/base_fork.py
+++ b/packages/testing/src/execution_testing/forks/base_fork.py
@@ -26,11 +26,9 @@ from execution_testing.base_types import (
 from execution_testing.base_types.conversions import BytesConvertible
 from execution_testing.vm import (
     EVMCodeType,
+    ForkOpcodeInterface,
+    OpcodeBase,
     Opcodes,
-)
-from execution_testing.vm.bytecode import (
-    ForkPrototype,
-    OpcodePrototype,
 )
 
 from .base_decorators import prefer_transition_to_method
@@ -245,7 +243,7 @@ class BaseForkMeta(ABCMeta):
         return cls is other or BaseForkMeta._is_subclass_of(other, cls)
 
 
-class BaseFork(ForkPrototype, metaclass=BaseForkMeta):
+class BaseFork(ForkOpcodeInterface, metaclass=BaseForkMeta):
     """
     An abstract class representing an Ethereum fork.
 
@@ -367,11 +365,14 @@ class BaseFork(ForkPrototype, metaclass=BaseForkMeta):
     @abstractmethod
     def opcode_gas_map(
         cls, *, block_number: int = 0, timestamp: int = 0
-    ) -> Dict[OpcodePrototype, int | Callable[[OpcodePrototype], int]]:
+    ) -> Dict[OpcodeBase, int | Callable[[OpcodeBase], int]]:
         """
-        Return a mapping of opcodes to either:
+        Return a mapping of opcodes to either int or callable.
+
+        The values of the mapping can be as follow:
         - Constants (int): Direct gas cost values from gas_costs()
-        - Callables: Functions that take the opcode instance with metadata and return gas cost
+        - Callables: Functions that take the opcode instance with metadata and
+                     return gas cost
         """
         pass
 
--- a/packages/testing/src/execution_testing/forks/base_fork.py
+++ b/packages/testing/src/execution_testing/forks/base_fork.py
@@ -1,8 +1,9 @@
 """Abstract base class for Ethereum forks."""
 
-from abc import ABC, ABCMeta, abstractmethod
+from abc import ABCMeta, abstractmethod
 from typing import (
     Any,
+    Callable,
     ClassVar,
     Dict,
     List,
@@ -23,7 +24,14 @@ from execution_testing.base_types import (
     BlobSchedule,
 )
 from execution_testing.base_types.conversions import BytesConvertible
-from execution_testing.vm import EVMCodeType, Opcodes
+from execution_testing.vm import (
+    EVMCodeType,
+    Opcodes,
+)
+from execution_testing.vm.bytecode import (
+    ForkPrototype,
+    OpcodePrototype,
+)
 
 from .base_decorators import prefer_transition_to_method
 from .gas_costs import GasCosts
@@ -237,7 +245,7 @@ class BaseForkMeta(ABCMeta):
         return cls is other or BaseForkMeta._is_subclass_of(other, cls)
 
 
-class BaseFork(ABC, metaclass=BaseForkMeta):
+class BaseFork(ForkPrototype, metaclass=BaseForkMeta):
     """
     An abstract class representing an Ethereum fork.
 
@@ -353,6 +361,18 @@ class BaseFork(ABC, metaclass=BaseForkMeta):
         cls, *, block_number: int = 0, timestamp: int = 0
     ) -> GasCosts:
         """Return dataclass with the gas costs constants for the fork."""
+        pass
+
+    @classmethod
+    @abstractmethod
+    def opcode_gas_map(
+        cls, *, block_number: int = 0, timestamp: int = 0
+    ) -> Dict[OpcodePrototype, int | Callable[[OpcodePrototype], int]]:
+        """
+        Return a mapping of opcodes to either:
+        - Constants (int): Direct gas cost values from gas_costs()
+        - Callables: Functions that take the opcode instance with metadata and return gas cost
+        """
         pass
 
     @classmethod
--- a/packages/testing/src/execution_testing/forks/forks/forks.py
+++ b/packages/testing/src/execution_testing/forks/forks/forks.py
@@ -4,7 +4,16 @@ from dataclasses import replace
 from hashlib import sha256
 from os.path import realpath
 from pathlib import Path
-from typing import List, Literal, Mapping, Optional, Sized, Tuple
+from typing import (
+    Callable,
+    Dict,
+    List,
+    Literal,
+    Mapping,
+    Optional,
+    Sized,
+    Tuple,
+)
 
 from execution_testing.base_types import (
     AccessList,
@@ -15,6 +24,11 @@ from execution_testing.base_types import (
 )
 from execution_testing.base_types.conversions import BytesConvertible
 from execution_testing.vm import EVMCodeType, Opcodes
+from execution_testing.vm.bytecode import (
+    OpcodeGasCostCalculator,
+    OpcodePrototype,
+    OpcodeRefundCalculator,
+)
 
 from ..base_fork import (
     BaseFeeChangeCalculator,
@@ -128,6 +142,7 @@ class Frontier(BaseFork, solc_name="homestead"):
             G_WARM_SLOAD=100,
             G_COLD_SLOAD=2_100,
             G_STORAGE_SET=20_000,
+            G_STORAGE_UPDATE=5_000,
             G_STORAGE_RESET=2_900,
             R_STORAGE_CLEAR=4_800,
             G_SELF_DESTRUCT=5_000,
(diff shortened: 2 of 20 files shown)


LouisTsai-Csie:
Awesome! I’ve partially reviewed the PR, and the framework changes look good so far. I still need to take a closer look at the per-opcode gas cost calculations and the updated opcode tests.

I’ve also spent some time thinking about how to continue with our original [plan](https://github.com/ethereum/execution-specs/pull/1776#discussion_r2576239101) for gas repricing and integrate it into this feature.

fselmo:
Note, we decided not to go with an implementation that uses the exact gas returned from the specs / EVM trace. I will leave this commit hash diff here so we have a reference to it in case it is ever useful in the future: https://github.com/ethereum/execution-specs/compare/858336ecbcc085c9697823631aa432db37b9e658...fselmo:execution-specs:7770f14daa131b25450a67059f055978f1fe3e3d.

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -157,6 +171,459 @@ def gas_costs(
...
+            return gas_costs.R_STORAGE_CLEAR
+
+        return 0
+
+    @classmethod
+    def _calculate_sstore_gas(
+        cls, opcode: OpcodePrototype, gas_costs: GasCosts
+    ) -> int:
+        """Calculate SSTORE gas cost based on metadata."""
+        metadata = opcode.metadata
```
fselmo, line 588: I think this is missing two cases:

- If a storage value is set tp something and changes to the same thing (NOOP)
- if a storage value is at a non-zero value and gets set to another non-zero value (1 -> 2)

I asked Claude to come up with a refactor and this does work:

```diff
         metadata = opcode.metadata
+        cold_cost = 0 if metadata["key_warm"] else gas_costs.G_COLD_SLOAD
 
-        if metadata["key_warm"]:
-            if metadata["value_reset"]:
-                return gas_costs.G_STORAGE_RESET
-            elif metadata["value_zero"] and metadata["current_nonzero"]:
-                return gas_costs.G_STORAGE_RESET
-            else:
-                return gas_costs.G_STORAGE_SET
-        else:
-            base = gas_costs.G_COLD_SLOAD
-            if metadata["value_reset"]:
-                return base + gas_costs.G_STORAGE_RESET
-            elif metadata["value_zero"] and metadata["current_nonzero"]:
-                return base + gas_costs.G_STORAGE_RESET
-            else:
-                return base + gas_costs.G_STORAGE_SET
+        # No-op: new == original
+        if metadata["value_reset"]:
+            return cold_cost + gas_costs.G_WARM_SLOAD
+
+        # Modifying existing storage (original != 0)
+        if metadata["current_nonzero"]:
+            return cold_cost + gas_costs.G_STORAGE_RESET
+
+        # Creating new storage (original == 0)
+        return cold_cost + gas_costs.G_STORAGE_SET
```

But I think a better suggestion may be for us to follow the exact logic the specs do in case only the constant values change at some point. It looks like we currently don't have a `G_STORAGE_UPDATE` (5000) on the testing side. I think the most "correct" approach would be to add this to our gas costs constants and then we can follow the exact pattern the specs do, using the same constants.

fselmo: > G_STORAGE_UPDATE

Ah it looks like our `G_STORAGE_RESET` is a hard-coded equivalent of `G_STORAGE_UPDATE - G_COLD_SLOAD`. The latter does feel more "correct" imo and would be less likely to require a manual update if these change.

packages/testing/src/execution_testing/forks/tests/test_opcode_gas_costs.py
```diff
@@ -0,0 +1,502 @@
...
+            Osaka.gas_costs().G_LOW,
+            id="clz_osaka",
+        ),
+    ],
+)
+def test_opcode_gas_costs(fork: Fork, opcode: Op, expected_cost: int) -> None:
+    op_gas_cost_calc = fork.opcode_gas_calculator()
+    assert expected_cost == op_gas_cost_calc(opcode)
+
+
```
fselmo, line 424: These Claude tests (or similar tests) might help, based on my comment above. Note this doesn't consider the `RESET` -> `UPDATE` refactor I mentioned.

```py
@pytest.mark.parametrize(
    "opcode,expected_cost",
    [
        # No-op: new == current (value_reset=True on clean slot)
        pytest.param(
            Op.SSTORE(key_warm=True, value_reset=True),
            _gas.G_WARM_SLOAD,
            id="sstore_noop_zero_warm",  # 0 → 0
        ),
        pytest.param(
            Op.SSTORE(key_warm=False, value_reset=True),
            _gas.G_COLD_SLOAD + _gas.G_WARM_SLOAD,
            id="sstore_noop_zero_cold",  # 0 → 0
        ),
        pytest.param(
            Op.SSTORE(
                key_warm=True,
                value_reset=True,
                current_nonzero=True,
            ),
            _gas.G_WARM_SLOAD,
            id="sstore_noop_nonzero_warm",  # 5 → 5
        ),
        pytest.param(
            Op.SSTORE(
                key_warm=False,
                value_reset=True,
                current_nonzero=True,
            ),
            _gas.G_COLD_SLOAD + _gas.G_WARM_SLOAD,
            id="sstore_noop_nonzero_cold",  # 5 → 5
        ),
        # Create storage: 0 → X (original == 0)
        pytest.param(
            Op.SSTORE(key_warm=True),
            _gas.G_STORAGE_SET,
            id="sstore_create_warm",  # 0 → 5
        ),
        pytest.param(
            Op.SSTORE(key_warm=False),
            _gas.G_COLD_SLOAD + _gas.G_STORAGE_SET,
            id="sstore_create_cold",  # 0 → 5
        ),
        # Modify storage: X → Y (original != 0, new != 0, new != original)
        pytest.param(
            Op.SSTORE(key_warm=True, current_nonzero=True),
            _gas.G_STORAGE_RESET,
            id="sstore_modify_warm",  # 5 → 7
        ),
        pytest.param(
            Op.SSTORE(key_warm=False, current_nonzero=True),
            _gas.G_COLD_SLOAD + _gas.G_STORAGE_RESET,
            id="sstore_modify_cold",  # 5 → 7
        ),
        # Clear storage: X → 0 (original != 0, new == 0)
        pytest.param(
            Op.SSTORE(
                key_warm=True,
                value_zero=True,
                current_nonzero=True,
            ),
            _gas.G_STORAGE_RESET,
            id="sstore_clear_warm",  # 5 → 0
        ),
        pytest.param(
            Op.SSTORE(
                key_warm=False,
                value_zero=True,
                current_nonzero=True,
            ),
            _gas.G_COLD_SLOAD + _gas.G_STORAGE_RESET,
            id="sstore_clear_cold",  # 5 → 0
        ),
    ],
)
def test_sstore_gas_costs(opcode: Op, expected_cost: int) -> None:
    """Test SSTORE gas costs for all single-SSTORE scenarios."""
    assert opcode.gas_cost(Osaka) == expected_cost
```

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -157,6 +171,459 @@ def gas_costs(
...
+                return base + gas_costs.G_STORAGE_RESET
+            else:
+                return base + gas_costs.G_STORAGE_SET
+
+    @classmethod
+    def _calculate_call_gas(
+        cls, opcode: OpcodePrototype, gas_costs: GasCosts
+    ) -> int:
+        """Calculate CALL/DELEGATECALL/STATICCALL gas cost based on metadata."""
+        metadata = opcode.metadata
```
fselmo, line 615: We need to account for 7702 delegated COLD / WARM access costs in all relevant opcodes: `CALL`, `CALLCODE`, `DELEGATECALL`, `STATICCALL`. We need something like `is_target_delegated` and `is_delegation_warm`. Or we can use `Literal` strings or an enum or something like this.

Thoughts?

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -157,6 +171,459 @@ def gas_costs(
...
+
+        # Code deposit cost when returning from initcode
+        code_deposit_size = metadata["code_deposit_size"]
+        return gas_costs.G_CODE_DEPOSIT_BYTE * code_deposit_size
+
+    @classmethod
+    def _calculate_selfdestruct_gas(
+        cls, opcode: OpcodePrototype, gas_costs: GasCosts
+    ) -> int:
+        """Calculate SELFDESTRUCT gas cost based on metadata."""
```
fselmo, line 648: I ran into this recently, we need to account for [EIP-161](https://eips.ethereum.org/EIPS/eip-161) case where we check if the originator value != 0. This has been since SpuriousDragon [here](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/src/ethereum/forks/spurious_dragon/vm/instructions/system.py#L391-L394).

packages/testing/src/execution_testing/forks/forks/forks.py
marioevz: This is a great catch. I need to take a look on how many tests use `G_STORAGE_RESET`, to see if we need to do it in this PR or a follow up.

packages/testing/src/execution_testing/forks/forks/forks.py
fselmo: Maybe `is_warm_delegation` and `is_cold_delegation` or something like this 🤔? I'm admittedly not great at these naming decisions 😂.

packages/testing/src/execution_testing/forks/forks/forks.py
marioevz: This is implied from `account_new` now, and I added it to the description in the opcode's docstring.

marioevz:
@fselmo I think I've applied all comments plus other fixes, thanks for the review!

packages/testing/src/execution_testing/forks/base_fork.py
```diff
@@ -237,7 +245,7 @@ def __le__(cls, other: "BaseForkMeta") -> bool:
         return cls is other or BaseForkMeta._is_subclass_of(other, cls)
 
 
-class BaseFork(ABC, metaclass=BaseForkMeta):
+class BaseFork(ForkPrototype, metaclass=BaseForkMeta):
```
fselmo: Hmm my only thought left in this PR is that `Frontier` (first fork impl) is getting really busy. And that's not going anywhere. I was trying to brainstorm some ways where we could abstract not just abstract classes out that `Frontier` then implements... but what if we refactored some code out of `Frontier` entirely into `Mixin` classes that it then inherits and can still override methods based on the fork?

This is a larger refactoring thought, out of scope for this PR... but I created an issue to look into this #2025.

If we _do_ like this idea of mixins, though, we could start with this PR and name `ForkPrototype` -> `OpcodeGasMixin` - then have the actual base opcode gas implementations the gas calculator, refund calculator, gas map, etc in that class. We could have something like `forks/mixins.py` to abstract this out.

I think even with this mixin architecture, `forks.py` will still grow quite a bit as different forks require different overrides of mixin methods... but at least some of the setup can be abstracted out.

Curious on your thoughts here.

---

edited: `BaseFork` -> `Frontier` that's the first impl class so this is where mixins would be inherited.

packages/testing/src/execution_testing/vm/bytecode.py
```diff
@@ -12,6 +13,63 @@
 from execution_testing.base_types import Bytes, Hash
 
 
+class OpcodePrototype:
```
fselmo: nit: I'm curious on the "Prototype" name here as this confused me when I was reviewing. I usually associate "prototype" with something that isn't production ready, or is being tested.

Consider this very much a nit but maybe `BaseOpcode` or `OpcodeSpec`, `OpcodeInterface`? This is a bit different from the mixin idea for `ForkPrototype` since that's not a base but rather introduction of methods for a particular focus. But this one does feel more like a `BaseOpcode` class. Wdyt?

fselmo (approved):
Can't wait to use this 😄. Great changes here.

I left some very minor remaining thoughts to think about. As I scrolled through `forks.py` I can't help but feel there's a better way to organize things rather than overloading forks indefinitely. I'm good with the state of this PR so I'm going to approve it in order to not block... but I do wonder if starting a mixin design pattern here wouldn't help us trend in the right direction. I also left a nit on the "Prototype" naming you can feel free to take or leave 😃.

packages/testing/src/execution_testing/vm/bytecode.py
marioevz: Great suggestion, I changed them to `OpcodeBase` and `ForkOpcodeInterface`.

PR #2004, opened 2026-01-12
chore(ci): reduce docker hub rate limit usage in hive-consume workflow

spencer-tb:
### Problem
Docker hub rate limits pulls to 100 every 6hrs when unauthenitcated. Currently we trigger these pulls way to often. This has the result of landing us with the following CI [fails](https://github.com/ethereum/execution-specs/actions/runs/20861109952):
```
Waiting for Hive to be ready (pid=3561, timeout=120s)...
Jan  9 18:10:56.320 INF building image image=hive/hiveproxy nocache=false pull=false
Step 1/14 : FROM golang:1-alpine as builder
1-alpine: Pulling from library/golang
Jan  9 18:10:59.241 ERR image build failed image=hive/hiveproxy err="error from registry: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit"
error from registry: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```

### ~~Solution~~ (replaced by solution below)
~~This PR reduces the docker hub rate limit usage in the hive-consume workflow with the following changes:~~
- ~~Docker hub authentication via `--docker.auth` flag to hive commands. Gives us 200 every 6 hrs.~~
  - ~~**Note:** Requires `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` secrets to be configured.~~
- ~~`max-parallel: 1` to run matrix jobs sequentially in order to share docker cache (reduces pulls from 12 to 3 per workflow).~~

### Alternative Solution
Use github actions cache for docker images to avoid docker hub pulls entirely:
- **New `cache-docker-images` job** - pulls required images once and caches them via github actions cache
- **New reusable actions:**
  - `.github/actions/cache-docker-images` - pulls and saves images to cache
  - `.github/actions/load-docker-images` - restores and loads images from cache
- **Matrix jobs run in parallel** - each job loads from cache instead of pulling from docker hub
- **Configurable images** - `docker_images` input allows customizing which images to cache (default: `docker.io/ethereum/client-go:latest docker.io/alpine:latest`)

## 🔗 Related Issues or PRs

#1692 #1855

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Catching them all for each PR!
**Bulbasaur - 0001**

### DIFF
--- a/.github/actions/cache-docker-images/action.yaml
+++ b/.github/actions/cache-docker-images/action.yaml
@@ -14,14 +14,17 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Get week number
+      id: week
+      shell: bash
+      run: echo "num=$(date +%U)" >> $GITHUB_OUTPUT
+
     - name: Restore Docker image cache
       id: cache-restore
       uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
       with:
         path: /tmp/docker-images
-        key: ${{ inputs.cache-key-prefix }}-${{ github.run_id }}
-        restore-keys: |
-          ${{ inputs.cache-key-prefix }}-
+        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml') }}
 
     - name: Pull and save Docker images
       shell: bash
@@ -44,4 +47,4 @@ runs:
       uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
       with:
         path: /tmp/docker-images
-        key: ${{ inputs.cache-key-prefix }}-${{ github.run_id }}
+        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml') }}
--- a/.github/actions/cache-docker-images/action.yaml
+++ b/.github/actions/cache-docker-images/action.yaml
@@ -24,7 +24,7 @@ runs:
       uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
       with:
         path: /tmp/docker-images
-        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml') }}
+        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml', 'execution-specs/.github/actions/cache-docker-images/action.yaml') }}
 
     - name: Pull and save Docker images
       shell: bash
@@ -47,4 +47,4 @@ runs:
       uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
       with:
         path: /tmp/docker-images
-        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml') }}
+        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml', 'execution-specs/.github/actions/cache-docker-images/action.yaml') }}
--- a/.github/actions/cache-docker-images/action.yaml
+++ b/.github/actions/cache-docker-images/action.yaml
@@ -1,0 +1,47 @@
+name: Cache Docker Images
+description: Cache Docker images to avoid Docker Hub rate limits
+
+inputs:
+  images:
+    description: "Space-separated list of Docker images to cache"
+    required: true
+    default: "docker.io/ethereum/client-go:latest docker.io/alpine:latest docker.io/library/golang:1-alpine"
+  cache-key-prefix:
+    description: "Prefix for the cache key"
+    required: false
+    default: "docker-images"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Restore Docker image cache
+      id: cache-restore
+      uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      with:
+        path: /tmp/docker-images
+        key: ${{ inputs.cache-key-prefix }}-${{ github.run_id }}
+        restore-keys: |
+          ${{ inputs.cache-key-prefix }}-
+
+    - name: Pull and save Docker images
+      shell: bash
+      run: |
+        mkdir -p /tmp/docker-images
+        for image in ${{ inputs.images }}; do
+          # Create a safe filename from image name
+          filename=$(echo "$image" | sed 's/[\/:]/-/g').tar.gz
+          if [ ! -f "/tmp/docker-images/$filename" ]; then
+            echo "Pulling $image..."
+            docker pull "$image"
+            echo "Saving $image to /tmp/docker-images/$filename..."
+            docker save "$image" | gzip > "/tmp/docker-images/$filename"
+          else
+            echo "Cache hit for $image, skipping pull"
+          fi
+        done
+
+    - name: Save Docker image cache
+      uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      with:
+        path: /tmp/docker-images
+        key: ${{ inputs.cache-key-prefix }}-${{ github.run_id }}
--- a/.github/actions/cache-docker-images/action.yaml
+++ b/.github/actions/cache-docker-images/action.yaml
@@ -43,7 +43,8 @@ runs:
           fi
         done
 
-    - name: Save Docker image cache
+    - name: Save Docker image cache 
+      if: steps.cache-restore.outputs.cache-hit != 'true'
       uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
       with:
         path: /tmp/docker-images
--- a/.github/actions/load-docker-images/action.yaml
+++ b/.github/actions/load-docker-images/action.yaml
@@ -1,0 +1,33 @@
+name: Load Cached Docker Images
+description: Load Docker images from cache
+
+inputs:
+  cache-key-prefix:
+    description: "Prefix for the cache key"
+    required: false
+    default: "docker-images"
+
+runs:
+  using: "composite"
+  steps:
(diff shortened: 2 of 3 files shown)


.github/workflows/hive-consume.yaml
```diff
@@ -71,6 +74,12 @@ jobs:
           version: ${{ vars.UV_VERSION }}
           python-version: ${{ vars.DEFAULT_PYTHON_VERSION }}
 
+      - name: Login to Docker Hub to increase pull limits
+        uses: docker/login-action@v3
```
danceratopz: Versions must be pinned as defined in the repo's setttings. Surprised that this isn't flagged, will look into it.

danceratopz (commented):
LGTM. We need to pin the dep. Will do it.

.github/workflows/hive-consume.yaml
spencer-tb: Will add to my Claude.md :)

spencer-tb:
I have to create a branch in `ethereum` over mine its seems. GitHub blocks secrets on fork PRs to prevent malicious PRs from stealing them. We are failing the CI now as its not available. Let me bump the branch.

spencer-tb:
Please see: https://github.com/ethereum/execution-specs/actions/runs/20922714279 🫢
For the CI!

danceratopz:
> I have to create a branch in `ethereum` over mine its seems. GitHub blocks secrets on fork PRs to prevent malicious PRs from stealing them. We are failing the CI now as its not available. Let me bump the branch.

Unfortunately, that mean almost all PRs from contributors will be affected by the rate limit as there origin will be a fork :sob:.

Carsons-Eels:
It looks like the new [non-parallelized run](https://github.com/ethereum/execution-specs/actions/runs/20922714279) took 38m, wheras before it was taking 12-13m. So, about 3x slower.

spencer-tb:
> It looks like the new [non-parallelized run](https://github.com/ethereum/execution-specs/actions/runs/20922714279) took 38m, wheras before it was taking 12-13m. So, about 3x slower.

I looked into this and unfortunately most of this time is in cloning EELS, downloading the fixtures, and setting up consume, not the actual consume runtime! We should look into improving the latter in a future PR.

.github/actions/cache-docker-images/action.yaml
```diff
@@ -0,0 +1,47 @@
...
+  using: "composite"
+  steps:
+    - name: Restore Docker image cache
+      id: cache-restore
+      uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      with:
+        path: /tmp/docker-images
+        key: ${{ inputs.cache-key-prefix }}-${{ github.run_id }}
+        restore-keys: |
+          ${{ inputs.cache-key-prefix }}-
```
danceratopz: This means we'll cache images from every run which will blow up the cache very quickly (we cache 3 images for every run). 

How about this? If we add this instead, it will be stable across runs and we get automatic cache invalidation when we modify this file (if  the `images` input changes, for example):
```suggestion
        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml') }}
```
(see below and next comments about the week addition). 

We should also remove `restore-keys` to avoid restoring stale images. This means we require an exact match.

To get the week number, we'd need an additional step:
```yaml
  - name: Get week number
    id: week
    run: echo "num=$(date +%U)" >> $GITHUB_OUTPUT
```

.github/actions/load-docker-images/action.yaml
```diff
@@ -0,0 +1,33 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Restore Docker image cache
+      uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      with:
+        path: /tmp/docker-images
+        key: ${{ inputs.cache-key-prefix }}-${{ github.run_id }}
+        restore-keys: |
+          ${{ inputs.cache-key-prefix }}-
```
danceratopz: Use exact match only.
```suggestion
```

.github/actions/cache-docker-images/action.yaml
```diff
@@ -0,0 +1,47 @@
+name: Cache Docker Images
+description: Cache Docker images to avoid Docker Hub rate limits
+
+inputs:
+  images:
+    description: "Space-separated list of Docker images to cache"
+    required: true
+    default: "docker.io/ethereum/client-go:latest docker.io/alpine:latest docker.io/library/golang:1-alpine"
```
danceratopz, line 8: As we use `:latest` tags, after we've cached a particular version, we'll never update it (unless we didn't run this action for 7 days, in which case the cache would auto-expire, this is unlikely though).

This means we'll never check for updates and have outdated images. Perhaps that's ok for these images, but could be an issue if we add other images that require updates.

To resolve this, we could either: 
1. Pin a particular version and update manually when required.
2. Try and rotate out the images automatically using a time-based cache key, e.g.,
```
key: docker-images-week-${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml') }}
```

I think the time-base key is simple enough to add it. Wdyt?

.github/actions/load-docker-images/action.yaml
```diff
@@ -0,0 +1,33 @@
...
+        if [ -d /tmp/docker-images ]; then
+          for file in /tmp/docker-images/*.tar.gz; do
+            if [ -f "$file" ]; then
+              echo "Loading $file..."
+              gunzip -c "$file" | docker load
+            fi
+          done
+        else
+          echo "No cached images found"
+        fi
```
danceratopz, line 37: I wonder if we should treat "No cached images found" as an error, because 

If the cache restore failed silently, the script will just print a message and succeed, the job continues, Hive pulls from Docker Hub and we could still hit rate limits.

danceratopz (commented):
Thanks for addressing this @spencer-tb! I like the approach but think we'll currently overload the cache very quickly. I added a few comments below.

.github/actions/cache-docker-images/action.yaml
spencer-tb: Nice! Added below.

.github/actions/load-docker-images/action.yaml
spencer-tb: Added exact match below.

.github/actions/load-docker-images/action.yaml
spencer-tb: Added an error below.

.github/actions/cache-docker-images/action.yaml
spencer-tb: Agreed. Added weekly rotation below. Images should now refresh once per week.

.github/actions/load-docker-images/action.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Get week number
+      id: week
+      shell: bash
+      run: echo "num=$(date +%U)" >> $GITHUB_OUTPUT
+
+    - name: Restore Docker image cache
+      uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+      with:
+        path: /tmp/docker-images
+        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml') }}
```
danceratopz: We're hashing `.github/actions/cache-docker-images/action.yaml', but `hive-consume` checks out to a sub-dir `execution-specs`.

I suspect, this means that the cache keys won't match as we'll be hashing a path that doesn't exist and silently fail (we'll hash the empty string).

I think this fallback solution works, but worth double-checking:
```suggestion
        path: /tmp/docker-images
        key: ${{ inputs.cache-key-prefix }}-week${{ steps.week.outputs.num }}-${{ hashFiles('.github/actions/cache-docker-images/action.yaml', 'execution-specs/.github/actions/cache-docker-images/action.yaml') }}
```

If we want to be more robust, then we could have `hive-consume.yaml` calculate the hash (it knows where the workflow file is) and pass the cache key to the cache-docker-images and load-docker-images actions. Perhaps overkill.

danceratopz (commented):
Thanks @spencer-tb! I think we introduced one new issue by hashing the workflow file 😢

.github/actions/load-docker-images/action.yaml
spencer-tb: Just updated this in both the load/cache actions! Its not required in the cache action but I changed it there to keep things consistent!

danceratopz: Awesome, thanks!

danceratopz (approved):
LGTM! Thanks a lot for this! Enjoyed the fast iterations!

.github/actions/cache-docker-images/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+            docker pull "$image"
+            echo "Saving $image to /tmp/docker-images/$filename..."
+            docker save "$image" | gzip > "/tmp/docker-images/$filename"
+          else
+            echo "Cache hit for $image, skipping pull"
+          fi
+        done
+
+    - name: Save Docker image cache
+      uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
```
danceratopz: We could skip the save step when the cache was already restored. This change prevents redundant saves on cache hits (to avoid wasted upload time). It doesn't affect cache growth:
```suggestion
    - name: Save Docker image cache 
      if: steps.cache-restore.outputs.cache-hit != 'true'
      uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
```

danceratopz (commented):
One more minor improvement that might be worth having @spencer-tb!

danceratopz (approved):
Thanks @spencer-tb, assuming passing and clean CI (it might be worth checking the logs first), this LGTM!

spencer-tb:
On the last run: `Cache not found for input keys: docker-images-week02-2853ffffa54104f82532615b9050021c7fce7913a4e506ef63186669751cecf6` 🚀

Going to do some re-names so we have:
- Hive Consume / Engine
- Hive Consume / RLP
- Hive Consume / Sync
- Hive Consume / Dev Mode
- Hive Consume / Cache Docker Images

Then we can check the logs again to see if it finds the correct input key!

spencer-tb:
Okay nice it works: `Cache restored successfully
Cache restored from key: docker-images-week02-2853ffffa54104f82532615b9050021c7fce7913a4e506ef63186669751cecf6` 😮‍💨 

Feel free to merge if you get to it before me!

danceratopz:
Thanks for checking the logs, @spencer-tb!

PR #1989, opened 2026-01-07
chore(test-types): Add `extra="forbid"` to `CamelModel`; fix errors

fselmo:
## 🗒️ Description

I'm still seeing issues from time to time where test writers (including myself) can misspell a kwarg to a test class which could go unnoticed with unintended consequences. This is an attempt to tighten up the pydantic models in our testing framework and prevent these sorts of issues.

- Use `extra="forbid"` in `CamelModel` which our test types inherit from
- Fix serialization issues by removing computed fields from input dicts on relevant models with singular field issues. For internal fixture models that use combined models to build a simpler model (many fields don't exist), explicitly turn on `extra="allow"` with a clarifying message on why we want to / need to do it here.
- Remove redundant `model_config` for classes that inherit from `CamelModel`

## 🔗 Related Issues or PRs

- Closes #1612 

## ✅ Checklist

- [x] All: Ran fast `tox` checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    uvx tox -e static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [ ] All: Considered adding an entry to [CHANGELOG.md](/ethereum/execution-spec-tests/blob/main/docs/CHANGELOG.md).
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-spec-tests/blob/main/docs/) directory.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #2014, opened 2026-01-13
fix(test-types,ci): use `evm` in `packages/test` unit tests & fix `TransactionReceipt`

danceratopz:
## 🗒️ Description

<del>
- Replace hardcoded geth v1.10.8 with dynamic lookup via GitHub API in download geth script. This ensures CI uses the latest stable geth release.
- Added `--alltools` command-line flag to not only download the `geth` binary but additional tools (for `evm`).
- Updated the workflow to use `--alltools` so that `tests_pytest_py3` has the latest geth `evm` binary available.
- This triggered the fail in CI that I was seeing locally from `tests_pytest_py3` in  `src/execution_testing/client_clis/tests/test_transition_tools_support.py::test_t8n_support[GethTransitionTool-*]`:
  ```python
   E       pydantic_core._pydantic_core.ValidationError: 1 validation error for Result
   E       receipts.0.blockNumber
   E         Extra inputs are not permitted [type=extra_forbidden, input_value='0x1', input_type=str]
   E           For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
  ```
</del>

- Fix: Remove `blockNumber` and `type` before validation in `TransactionReceipt`. 

I think this behavior changed due to:
- #1989

I only noticed it locally as I had `evm` in my PATH, but we didn't have an `evm` available in the PATH when `tests_pytest_py3` env ran.

## 🔗 Related Issues or PRs
 
- #1989
- #2000

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

image

### DIFF
--- a/packages/testing/src/execution_testing/test_types/receipt_types.py
+++ b/packages/testing/src/execution_testing/test_types/receipt_types.py
@@ -49,6 +49,9 @@ class TransactionReceipt(CamelModel):
             # t8n tool may return 'post_state' which is not part of this model
             data.pop("post_state", None)
             data.pop("postState", None)
+            # geth (1.16+) returns extra fields in receipts
+            data.pop("type", None)
+            data.pop("blockNumber", None)
         return data
 
     transaction_hash: Hash | None = None


danceratopz:
@fselmo can you please check the changes to test-types? Here are the same fails I was experiencing locally before the fix:
- https://github.com/ethereum/execution-specs/actions/runs/20959245299/job/60231804865

@gurukamath I assume it's ok to bump the version of geth here? The previous version was from 2021 :) I did try to add a way to take the latest release version from geth's github releases and then download it from gethstore.blob.core.windows.net, but the latest stable release wasn't available there so fell back to a hard-coded version (definitely less flaky).

packages/testing/src/execution_testing/test_types/receipt_types.py
```diff
@@ -49,6 +49,9 @@ def strip_extra_fields(cls, data: Any) -> Any:
             # t8n tool may return 'post_state' which is not part of this model
             data.pop("post_state", None)
             data.pop("postState", None)
+            # geth (1.16+) returns extra fields in receipts
+            data.pop("type", None)
```
fselmo, line 53: Hmm yeah, if this wasn't used as a test-facing class to set the `expected_receipt` in some tests, I'd say we could explicitly turn on `{"extra": "ignore"}`. I don't know the right answer here... this may end up being more annoying than helpful, though it has already proven helpful to find unwanted bugs in the tests.

We can do `TransactionReceipt.model_validate(extra="ignore")` in cases where we don't really care to pass extras through... but I think at t8n we would want to catch things and ignore them explicitly. I'm just thinking out loud here. Maybe this here is the right decision based on where this being used... but we can also be conscious that we can use `extra="ignore"` within `model_validate()` itself and we don't need to set it on the `model_config` globally.

TL;DR: "lgtm 👍🏼" 🤣

fselmo (approved):
lgtm 👍🏼

gurukamath:
> @gurukamath I assume it's ok to bump the version of geth here? The previous version was from 2021 :) I did try to add a way to take the latest release version from geth's github releases and then download it from gethstore.blob.core.windows.net, but the latest stable release wasn't available there so fell back to a hard-coded version (definitely less flaky).

The geth binary is required for one of the ethash tests from the POW forks. I don't see much of an issue with bumping the version as long as it supports this

gurukamath (approved):
LGTM!

packages/testing/src/execution_testing/test_types/receipt_types.py
danceratopz: Yeah, I agree we should try and be as strict as possible when interacting with `t8n`. Thanks for the additional info. As I understand, I think this solution is the softest way of preventing the error.

danceratopz:
> > @gurukamath I assume it's ok to bump the version of geth here? The previous version was from 2021 :) I did try to add a way to take the latest release version from geth's github releases and then download it from gethstore.blob.core.windows.net, but the latest stable release wasn't available there so fell back to a hard-coded version (definitely less flaky).
> 
> The geth binary is required for one of the ethash tests from the POW forks. I don't see much of an issue with bumping the version as long as it supports this

Thanks for the context, @gurukamath! I just checked and the `makedag` subcommand was cleaned out with PoW in [v1.12.0](https://github.com/ethereum/go-ethereum/releases/tag/v1.12.0). This means that the following would fail with newer geth's (confirmed locally):

```bash
uv run pytest tests/json_infra/test_ethash_general.py -k test_dataset_generation_random_epoch
```

This wasn't caught by CI in this PR as the test module (` pytest tests/json_infra/test_ethash_general.py` is not in the list of changed files, see this (and previous step):
https://github.com/ethereum/execution-specs/actions/runs/20964454224/job/60250939820?pr=2014#step:6:1

But if it was, bumping the geth version would cause CI to fail :sob: 

I've reverted the changes to the script and the workflow file to avoid that. The main issue I was trying to resolve the failing tests when running `tests_pytest_py3` locally (with `evm` in my path), which is addressed. But it would be good to provide a go-ethereum evm binary to `tests_pytest_py3` in CI which will enable these tests:
```
src/execution_testing/client_clis/tests/test_transition_tools_support.py::test_t8n_support[GethTransitionTool-*]
```

Follow-up issue:
- https://github.com/ethereum/execution-specs/issues/2020

PR #2016, opened 2026-01-13
feat(fill): Customize fill with `--chain-id`

### LINKED ISSUE #1394: Can eels t8n accept chain_id != 1
Currently, the chain_id is hardcoded in certain places in the `t8n` tool. We need to look into making it more flexible.

See [this](https://github.com/ethereum/execution-specs/blob/forks/osaka/src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py#L217-L219) and [this](https://github.com/ethereum/execution-specs/blob/forks/osaka/src/ethereum_spec_tools/evm_tools/loaders/transaction_loader.py#L43-L45) for example.

pdobacz:
## 🗒️ Description

YMMV but this works for me if I want to fill for chainid other than `1`. 

## 🔗 Related Issues or PRs

Maybe closes #1394 , please verify. There are two links posted in the description there, the latter is handled, ~the former I think also seems like should be doable, but I'm not sure if that `secretKey` is used anywhere at all, please advise~ I think the latter one is also handled, CI made me do it.

## ✅ Checklist

- [x] All: Ran fast `tox` 
- [x] All: PR title adheres 

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -582,7 +582,10 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         dest="chain_id",
         type=int,
         default=None,
-        help=("Specify the chain ID for the test filling."),
+        help=(
+            "Specify the chain ID for the test filling. "
+            f"Default: {ChainConfigDefaults.chain_id}."
+        ),
     )
     evm_group.addoption(
         "--traces",
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -69,6 +69,7 @@ from execution_testing.forks import (
 from execution_testing.specs import BaseTest
 from execution_testing.specs.base import FillResult, OpMode
 from execution_testing.test_types import EnvironmentDefaults
+from execution_testing.test_types.chain_config_types import ChainConfigDefaults
 from execution_testing.tools.utility.versioning import (
     generate_github_url,
     get_current_commit_hash_or_tag,
@@ -576,6 +577,14 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         ),
     )
     evm_group.addoption(
+        "--chain-id",
+        action="store",
+        dest="chain_id",
+        type=str,
+        default=None,
+        help=("Specify the chain ID for the test filling."),
+    )
+    evm_group.addoption(
         "--traces",
         action="store_true",
         dest="evm_collect_traces",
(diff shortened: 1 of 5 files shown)


src/ethereum_spec_tools/evm_tools/t8n/t8n_types.py
```diff
@@ -215,9 +215,9 @@ def sign_transaction(self, json_tx: Any) -> None:
             if t8n.fork.has_signing_hash_155:
                 if protected:
                     signing_hash = t8n.fork.signing_hash_155(
-                        tx_decoded, U64(1)
+                        tx_decoded, self.t8n.chain_id
                     )
-                    v_addend = U256(37)  # Assuming chain_id = 1
+                    v_addend = U256(36) + U256(self.t8n.chain_id)
```
fselmo: I think `v_addend` is wrong here? 👀 

```suggestion
                    # EIP-155: CHAIN_ID * 2 + 35
                    v_addend = U256(self.t8n.chain_id) * U256(2) + U256(35)
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -516,6 +517,14 @@ def pytest_addoption(parser: pytest.Parser) -> None:
             "framework tests/ci; not intended for regular CLI use."
         ),
     )
+    evm_group.addoption(
+        "--chain-id",
+        action="store",
+        dest="chain_id",
+        type=str,
```
fselmo: we could use `int` directly here

```suggestion
        type=int,
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -858,6 +867,16 @@ def pytest_configure(config: pytest.Config) -> None:
         f"<code>{command_line_args}</code>"
     )
 
+    chain_id = config.getoption("chain_id")
+
+    if chain_id is None:
+        # Try to get the chain ID from the environment variable
+        chain_id = os.environ.get("CHAIN_ID")
```
fselmo: If we use `int` directly above, we could convert here:

```py
if chain_id is None:                                                                                                                                                                                           
      env_chain_id = os.environ.get("CHAIN_ID")                                                                                                                                                                
      if env_chain_id is not None:                                                                                                                                                                               
          chain_id = int(env_chain_id)
```

fselmo (commented):
Untested so there may be a bit of fanangling to use `int` instead of `str` but I think it may make more sense. Biggest thing to get a sanity check on is the `v_addend` to make sure it's defined appropriately 👀.

Also needs to resolve conflicts.

pdobacz:
Thank you, I've applied the suggestions. As for the `/docs` it seems the auto generated list of options should do in this case.

pdobacz:
@fselmo re the CI failure, it seems to be caused by some mishandling of a BAL exception? I have no idea how to fix this.

fselmo:
> @fselmo re the CI failure, it seems to be caused by some mishandling of a BAL exception? I have no idea how to fix this.

Sorry for the delay here. Can you try fetching latest and rebasing with `forks/amsterdam`? We recently changed the CI run for `json_infra` and I'd like to get this to re-run for a sanity check.

pdobacz:
@fselmo Done, seems green now

fselmo (approved):
lgtm, ty!

PR #2030, opened 2026-01-16
refactor(testing): Implement IteratingBytecode, FixedIterationsBytecode

marioevz:
## 🗒️ Description

Implements some changes to the benchmark tests to use `bytecode.gas_cost` instead of manually adding `fork.gas_costs` cost values.

### `IteratingBytecode`, `FixedIterationsBytecode`

Two new classes that subclass `Bytecode` but contain it in three different sections: `setup`, `iterating` (could be renamed to `iteration` perhaps?) and `cleanup`.

This allows the implementation of a new function `gas_cost_by_iteration_count` which returns the gas cost given a number of iterations of the `iterating` bytecode.

Also included is a field called `warm_iterating` which defaults to `None` but when set will be used as the iterating bytecode from the second iteration and onwards.

This allows passing bytecodes with different gas behaviors on the first run when compared to the second run and after. E.g. `SSTORE` accessing a cold key at the beginning but starting from the second iteration it becomes warm.

`FixedIterationsBytecode` is very similar but simply contains another field `iteration_count` which makes it so that `bytecode.gas_cost` for this instance uses this iteration count and returns the gas used after all the iterations are run.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


marioevz:
cc @LouisTsai-Csie this is a work in progress but I wanted to highlight some ideas in this PR that could benefit benchmarking, and also so we don't duplicate work.

cc @jochem-brouwer also PTAL, I would appreciate your feedback on these new ideas!

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -151,91 +151,80 @@ def test_storage_access_cold(
...
-        else:
-            loop_cost += gas_costs.G_WARM_SLOAD
-    elif storage_action == StorageAction.READ:
-        loop_cost += 0  # Only G_COLD_SLOAD is charged
-
     # Contract code
     execution_code_body = Bytecode()
     if storage_action == StorageAction.WRITE_SAME_VALUE:
         # All the storage slots in the contract are initialized to their index.
         # That is, storage slot `i` is initialized to `i`.
```
LouisTsai-Csie:
```suggestion
```
This is only true if `absent_slots=False`, we could also simplify this:

```
 if absent_slots:
execution_code_body = Op.SSTORE(
    Op.DUP2, 0, original_value=0, new_value=0
)
else:
execution_code_body = Op.SSTORE(
    Op.DUP1, Op.DUP1, original_value=1, new_value=1
)
```

into this:

```python
execution_code_body = Op.SSTORE(
    Op.DUP1, Op.DUP1, original_value=int(not absent_slots), new_value=1
)
```

Not sure if this breaks the readability.

tests/benchmark/compute/helpers.py
```diff
@@ -204,3 +219,517 @@ def calculate_optimal_input_length(
...
+                dest_offset=0,
+                offset=0,
+                size=len(initcode),
+                # Gas accounting
+                address_warm=False,
+                data_size=len(initcode),
+                new_memory_size=len(initcode),
+            )
+            + Op.ADD(1, Op.CALLDATALOAD(32))
+            + Op.CALLDATALOAD(0)
```
LouisTsai-Csie, line 355: It would be nice if we have some comment for the CALLDATA layout here.

CALLDATA[0..31] = start_index
CALLDATA[32..63] = end_index

LouisTsai-Csie (changes requested):
I like this idea, especially the `txs_with_gas_limit_cap` helper that uses a sliding-window approach to handle contract creation and the transaction gas limit cap. I think this could be useful for several `bloatnet` scenarios as well. After this PR (and PR #1962 ) merge, I can follow up by refactoring the bloatnet scenarios accordingly.

Also, for `IteratingBytecode` and `FixedIterationsBytecode`: since these are not limited to compute scenarios, what do you think about moving them under `packages/testing/src/execution_testing/benchmark/`? That would let us reuse them as templates for additional helper objects (e.g., `MaxSizedContractInitcode`, `MaxSizedContractFactory`, etc.).

LouisTsai-Csie:
Hi @marioevz could we try to get this merged first so we don’t get file conflict ? I’d like to refactor the `test_storage_access_cold` benchmark to make it Osaka-compatible using your approach after this PR merged

tests/benchmark/compute/instruction/test_storage.py
marioevz: I've updated the comment, but the optimization does not quite work because if we use `Op.DUP1` with zero->zero case, we duplicate the index which is not zero.

tests/benchmark/compute/instruction/test_keccak.py
LouisTsai-Csie:
```suggestion
            attack_block=iterating_bytecode,
```

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -470,3 +469,444 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
+                iteration_count=mid,
+                **intrinsic_cost_kwargs,
+            )
+
+            if tx_gas <= gas_limit_cap:
+                best_iterations = mid
+                best_iterations_gas = tx_gas
+                low = mid + 1
+            else:
+                high = mid - 1
```
LouisTsai-Csie: I found this logic duplicate in `tx_iterations_by_gas_limit_cap` and `tx_iterations_by_gas_limit`, could we replace with this `_binary_search_iterations` helper function?

```python
def _binary_search_iterations(
    self,
    *,
    fork: Fork,
    gas_limit: int,
    initial_low: int = 0,
    initial_high: int = 1,
    **intrinsic_cost_kwargs: Any,
) -> Tuple[int, int]:
    """
    Binary search for the maximum iterations that fit within a gas limit.
    """
    low = initial_low
    high = initial_high

    # Exponential search to find upper bound
    high_gas_cost = self.tx_gas_limit_by_iteration_count(
        fork=fork,
        iteration_count=high,
        **intrinsic_cost_kwargs,
    )
    while high_gas_cost < gas_limit:
        low = high
        high *= 2
        high_gas_cost = self.tx_gas_limit_by_iteration_count(
            fork=fork,
            iteration_count=high,
            **intrinsic_cost_kwargs,
        )

    # Binary search for exact fit
    best_iterations = 0
    best_iterations_gas = 0

    while low <= high:
        mid = (low + high) // 2
        if mid == 0:
            break

        tx_gas = self.tx_gas_limit_by_iteration_count(
            fork=fork,
            iteration_count=mid,
            **intrinsic_cost_kwargs,
        )

        if tx_gas <= gas_limit:
            best_iterations = mid
            best_iterations_gas = tx_gas
            low = mid + 1
        else:
            high = mid - 1

    return best_iterations, best_iterations_gas

# Updated version
def tx_iterations_by_gas_limit_cap(
    self,
    *,
    fork: Fork,
    **intrinsic_cost_kwargs: Any,
) -> Tuple[int, int]:
    """
    Calculate the number of iterations needed to reach the fork's
    transaction gas limit cap.
    """
    gas_limit_cap = fork.transaction_gas_limit_cap()
    if gas_limit_cap is None:
        raise ValueError("Fork has no gas limit cap.")

    return self._binary_search_iterations(
        fork=fork,
        gas_limit=gas_limit_cap,
        initial_low=0,
        initial_high=1,
        **intrinsic_cost_kwargs,
    )
```

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -470,3 +469,444 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
+        fork: Fork,
+        **intrinsic_cost_kwargs: Any,
+    ) -> Tuple[int, int]:
+        """
+        Calculate the number of iterations needed to reach the fork's
+        transaction gas limit cap.
+        """
+        gas_limit_cap = fork.transaction_gas_limit_cap()
+        if gas_limit_cap is None:
+            raise ValueError("Fork has no gas limit cap.")
```
LouisTsai-Csie: Some benchmarks are still running under the Prague fork like fixed-opcode-count feature. With the current structure, Prague benchmark should use `tx_iterations_by_gas_limit`, while Osaka uses `tx_iterations_by_gas_limit_cap`.

We could simplify this by unifying the logic in `tx_iterations_by_gas_limit_cap`: if `gas_limit_cap` is None, fall back to using the block gas limit. This would avoid having separate code paths for Prague vs. Osaka.

Example: 

https://github.com/ethereum/execution-specs/blob/c6c380acffa586cfd683384fb2e09348e9a97c03/tests/benchmark/conftest.py#L90-L93

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -470,3 +469,444 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
+        Calculate how to split a total number of iterations across multiple
+        transactions so that each transaction fits within the gas limit cap.
+
+        Returns a list where each element represents the number of iterations
+        for that transaction, and the sum equals total_iterations.
+        """
+        gas_limit_cap = fork.transaction_gas_limit_cap()
+        if gas_limit_cap is None:
+            # No limit, all iterations fit in a single transaction.
+            return [total_iterations]
```
LouisTsai-Csie: We should be careful with this approach. In some cases, we use an optimistic estimation, and not all operations may fit into a single transaction. In those scenarios, we may need to split the execution into two transactions across separate blocks.

LouisTsai-Csie (changes requested):
I’ve reviewed most of the changes, except for` test_iterating_bytecode.py`, I’ll continue on it later today. I left a few comments, but overall it looks really good. Amazing work!

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -470,3 +469,444 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
+
+        if max_iterations_per_tx == 0:
+            raise ValueError(
+                "No iterations possible with the given gas limit cap."
+            )
+
+        full_txs = remaining_iterations // max_iterations_per_tx
+        result = [max_iterations_per_tx] * full_txs
+        remaining_iterations -= max_iterations_per_tx * full_txs
+        return result + [remaining_iterations]
```
LouisTsai-Csie: What if there is no remaining iteration, there might be trailing zero in the list.

LouisTsai-Csie (changes requested):

tests/benchmark/compute/instruction/test_keccak.py
marioevz: Fixed with:
```python
Op.POP(
                Op.SHA3(Op.PUSH0, Op.DUP1, data_size=optimal_input_length)
            )
```
because `iteration_bytecode` does not always point to the correct bytecode.

marioevz:
@LouisTsai-Csie thanks for the review. I think I've addressed all comments and took the opportunity to implement two new methods: `transactions_by_gas_limit` and `transactions_by_total_iteration_count`.

Please take a look at `tests/benchmark/compute/scenario/test_unchunkified_bytecode.py` to see these in action and how they simplify the benchmark test writing even further.

marioevz:
Need to rebase before trying to merge, there seems to be a new conflict.

LouisTsai-Csie (approved):
LGTM, i am very bullish on this feature!!!

LouisTsai-Csie:
Running the test locally with release configuration before merging, to ensure not breaking the process again.

PR #2032, opened 2026-01-17
feat(tests): port EXTCODEHASH dynamicAccountOverwriteEmpty

chfast:
## 🗒️ Description
Port the legacy static test:
stExtCodeHash/dynamicAccountOverwriteEmpty_Paris
for EIP-1052 EXTCODEHASH testing.

The test verifies EXTCODEHASH correctly updates when an empty account is overwritten by CREATE2. Modified from original: target account has no storage to avoid EIP-7610 collision behavior.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,125 @@
...
+# keccak256 of empty data
+EMPTY_KECCAK = keccak256(b"")
+
+
+@pytest.mark.valid_from("Constantinople")
+@pytest.mark.ported_from(
+    [
+        "https://github.com/ethereum/tests/tree/v13.3/src/GeneralStateTestsFiller/stExtCodeHash/dynamicAccountOverwriteEmpty_ParisFiller.yml",  # noqa: E501
+    ]
+)
```
LouisTsai-Csie, line 413:
```suggestion
@pytest.mark.ported_from(
    [
        "https://github.com/ethereum/tests/tree/v13.3/src/GeneralStateTestsFiller/stExtCodeHash/dynamicAccountOverwriteEmpty_ParisFiller.yml",  # noqa: E501
    ],
    pr=["https://github.com/ethereum/execution-specs/pull/2032"],
)
```

Adding this might be helpful, i discovered that other porting PR also have this [marker](https://github.com/ethereum/execution-specs/blob/4ef381a0f75c96b52da635653ab580e731d3882a/tests/frontier/precompiles/test_ripemd.py#L20).

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,125 @@
...
+@pytest.mark.pre_alloc_modify
```
LouisTsai-Csie:
```suggestion
```

I’m not sure if we need this for the test, maybe I’m missing something here.

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,125 @@
...
+            10,
+            Op.CREATE2(value=0, offset=0, size=len(create2_initcode), salt=0),
+        )
+        # Call the deployed contract to execute its code (SSTORE 80, 11)
+        + Op.SSTORE(3, Op.CALL(50000, target_address, 0, 0, 0, 0, 0))
+        # Check stats after CREATE2
+        + Op.SSTORE(4, Op.EXTCODEHASH(target_address))
+        + Op.SSTORE(5, Op.EXTCODESIZE(target_address))
+        + Op.STOP
+    )
```
LouisTsai-Csie, line 506: Some suggestion for this call sequence:

- We could use the `Storage` object here instead of specifying the slot for `SSTORE`.

Example:

```python
from execution_testing import Storage

storage = Storage()

Op.SSTORE(storage.store_next(EMPTY_KECCAK), Op.EXTCODEHASH(target_address))
```

`store_next` would handle the slot management for us. And the benefit is we could simplify the post storage by passing the entire `Storage` object:

```python
caller_address: Account(storage=storage),
```

- In the original test, it do the `EXTCODEHASH`, `EXTCODESIZE`, `EXTCODECOPY`, `CALLCODE` to the target address before / after the contract creation. We might need to do the same check here to avoid any coverage loss.

- In the original test, it performs a `CALLCODE` to the target account. Since `CALLCODE` executes the callee’s code in the context of the caller, any code present at the target account could potentially modify the caller’s storage. The test currently checks whether the call succeeds and whether the caller’s storage remains unchanged, but this is insufficient here because there are two possible scenarios:
  - There is no code at the target account, so nothing is executed.
  - Code execution succeeds, but the logic does not modify storage.
To distinguish these cases, I suggest passing **zero gas** to the call. This ensures that if a success value is returned, no execution could have taken place at all, making the test behavior unambiguous.


This is my suggested changes, please let me know if it is unclear or does not make sense, thanks:
```python
caller_code = (
    # EXTCODEHASH Check
    Op.SSTORE(
        storage.store_next(EMPTY_KECCAK), Op.EXTCODEHASH(target_address)
    )
    # EXTCODESIZE Check
    + Op.SSTORE(storage.store_next(0), Op.EXTCODESIZE(target_address))
    # EXTCODECOPY Check
    + Op.EXTCODECOPY(target_address, 0, 0, 32)
    + Op.SSTORE(storage.store_next(0), Op.MLOAD(0))
    # CALL* Ops Check
    + Op.SSTORE(
        storage.store_next(1),
        Op.CALLCODE(
            gas=0,  # Passing zero gas to ensure no execution
            address=target_address,
        ),
    )
    # Create the contract via CREATE2
    + Op.MSTORE(0, bytes(create2_initcode).ljust(32, b"\0"))
    + Op.SSTORE(
        storage.store_next(target_address),
        Op.CREATE2(size=len(create2_initcode)),
    )
    # EXTCODEHASH Check
    + Op.SSTORE(
        storage.store_next(deploy_code.keccak256()),
        Op.EXTCODEHASH(target_address),
    )
    # EXTCODESIZE Check
    + Op.SSTORE(
        storage.store_next(len(deploy_code)),
        Op.EXTCODESIZE(target_address),
    )
    # EXTCODECOPY Check
    + Op.EXTCODECOPY(target_address, 0, 0, 32)
    + Op.SSTORE(
        storage.store_next(bytes(deploy_code).ljust(32, b"\0")),
        Op.MLOAD(0),
    )
    # CALL* Ops Check
    + Op.SSTORE(
        storage.store_next(1),
        Op.CALL(gas=gas_cost, address=target_address),
    )
)
```

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,125 @@
...
+                    0: EMPTY_KECCAK,
+                    1: 0,  # EXTCODESIZE = 0
+                    # CALL to deployed contract succeeded
+                    3: 1,
+                    # CREATE2 returned the new address
+                    10: target_address,
+                    # After CREATE2: account has code
+                    4: deploy_code.keccak256(),
+                    5: len(deploy_code),
+                },
```
LouisTsai-Csie: Continue the last comment, we could simplify this storage comparison to


```suggestion
                storage=storage
```

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
LouisTsai-Csie, line 506: Some suggestion for this call sequence:

- We could use the `Storage` object here instead of specifying the slot for `SSTORE`.

Example:

```python
from execution_testing import Storage

storage = Storage()

Op.SSTORE(storage.store_next(EMPTY_KECCAK), Op.EXTCODEHASH(target_address))
```

`store_next` would handle the slot management for us. And the benefit is we could simplify the post storage by passing the entire `Storage` object:

```python
caller_address: Account(storage=storage),
```

- In the original test, it do the `EXTCODEHASH`, `EXTCODESIZE`, `EXTCODECOPY`, `CALLCODE` to the target address before / after the contract creation. We might need to do the same check here to avoid any coverage loss.

- In the original test, it performs a `CALLCODE` to the target account. Since `CALLCODE` executes the callee’s code in the context of the caller, any code present at the target account could potentially modify the caller’s storage. The test currently checks whether the call succeeds and whether the caller’s storage remains unchanged, but this is insufficient here because there are two possible scenarios:
  - There is no code at the target account, so nothing is executed.
  - Code execution succeeds, but the logic does not modify storage.
To distinguish these cases, I suggest passing **zero gas** to the call. This ensures that if a success value is returned, no execution could have taken place at all, making the test behavior unambiguous.

- Consider the potential changes to the storage gas cost, i would suggest not to hardcode the gas cost in `CALL` operations. Now we have a powerful gas cost calculation feature, we could do something like this:

```python
deploy_code = Op.SSTORE(
    key=80,
    value=11,
    key_warm=False,
    original_value=0,
    current_value=None,
    new_value=11,
)
gas_cost = deploy_code.gas_cost(fork)
...
Op.CALL(
  gas=gas_cost, # Use this calculated value instead of the hardcoded one
  address=target_address
)
```


This is my suggested changes, please let me know if it is unclear or does not make sense, thanks:
```python
caller_code = (
    # EXTCODEHASH Check
    Op.SSTORE(
        storage.store_next(EMPTY_KECCAK), Op.EXTCODEHASH(target_address)
    )
    # EXTCODESIZE Check
    + Op.SSTORE(storage.store_next(0), Op.EXTCODESIZE(target_address))
    # EXTCODECOPY Check
    + Op.EXTCODECOPY(target_address, 0, 0, 32)
    + Op.SSTORE(storage.store_next(0), Op.MLOAD(0))
    # CALL* Ops Check
    + Op.SSTORE(
        storage.store_next(1),
        Op.CALLCODE(
            gas=0,  # Passing zero gas to ensure no execution
            address=target_address,
        ),
    )
    # Create the contract via CREATE2
    + Op.MSTORE(0, bytes(create2_initcode).ljust(32, b"\0"))
    + Op.SSTORE(
        storage.store_next(target_address),
        Op.CREATE2(size=len(create2_initcode)),
    )
    # EXTCODEHASH Check
    + Op.SSTORE(
        storage.store_next(deploy_code.keccak256()),
        Op.EXTCODEHASH(target_address),
    )
    # EXTCODESIZE Check
    + Op.SSTORE(
        storage.store_next(len(deploy_code)),
        Op.EXTCODESIZE(target_address),
    )
    # EXTCODECOPY Check
    + Op.EXTCODECOPY(target_address, 0, 0, 32)
    + Op.SSTORE(
        storage.store_next(bytes(deploy_code).ljust(32, b"\0")),
        Op.MLOAD(0),
    )
    # CALL* Ops Check
    + Op.SSTORE(
        storage.store_next(1),
        Op.CALL(gas=gas_cost, address=target_address),
    )
)
```

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,125 @@
...
+    pre[caller_address] = Account(
+        balance=10**18,
+        code=caller_code,
+        nonce=1,
+    )
+
+    # Pre-existing empty account at CREATE2 target.
+    pre[target_address] = Account(
+        balance=10,
+    )
```
LouisTsai-Csie: This works under the `fill` mode, but i would suggest using a more dynamic approach, so that we could run the test under `execute remote` in a live network.

We could wait until PR #1996  merged to do the refactor again, and we would no longer need the hardcoded `caller_address`:

```python
caller_address = Address(0x095E7BAEA6A6C7C4C2DFEB977EFAC326AF552D87)
```

LouisTsai-Csie (changes requested):
Thanks for the PR! I’ve left a few comments suggesting some refactoring using helper functions. Overall, the test logic looks great.

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
chfast: The `caller_address` is fixed, so you will not be able to deploy it in a live network. Maybe we should make it random: I don't think it will modify the main flow of the test.

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
chfast: Added.

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
chfast: I applied suggestions and even extended the workflow. In the end the check hash, size, and code itself in 3 states: non-existient, no-code and CREATE2-deployed. The no-code account is created via CALL so we can avoid fixed-address pre-state.

We also call the target account code via CALL and DELEGATECALL.

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
chfast: This is resolved now by creating an account with `CALL`.

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
chfast: Done.

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
chfast: Resolved. See above.

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,158 @@
...
+@pytest.mark.ported_from(
+    [
+        "https://github.com/ethereum/tests/tree/v13.3/src/GeneralStateTestsFiller/stExtCodeHash/dynamicAccountOverwriteEmpty_ParisFiller.yml",  # noqa: E501
+    ],
+    pr=["https://github.com/ethereum/execution-specs/pull/2032"],
+)
+@pytest.mark.parametrize(
+    "target_exists",
+    [True, False],
+)
```
LouisTsai-Csie, line 417: The new test format looks good, but in the legacy test the target account (created via `CREATE2`) had pre-existing storage. We’ve removed that setup, do we lose the coverage here?

I now understand why `pytest.mark.pre_alloc_modify` is required: we want to configure an account storage via `pre`.

I’m thinking of parameterizing the cases like this:
```python
@pytest.mark.parametrize(
    "exist_balance, exist_storage",
    [
        pytest.param(True, True, id="collision", marks=pytest.mark.pre_alloc_modify),
        pytest.param(True, False, id="balance_only"),
        pytest.param(False, False, id="non_existent"),
    ],
)
```

And configuring `pre` like:

```python
if exist_storage:
    pre[target_address] = Account(balance=10, storage={0: 1})
elif exist_balance:
    pre.fund_address(target_address, 1)
```

However, this setting keeps failing in `fill`: it looks like `t8n` doesn’t handle the `initial storage + nonce=0` scenario as I expected. Am I missing something in how pre-state is applied here?

LouisTsai-Csie: We could add multiple markers like this, for the pre-alloc modification, we should also skip the execute-remote mode.

```python
pytest.param(
  True,
  True,
  id="collision",
  marks=[
      pytest.mark.pre_alloc_modify,
      pytest.mark.execute(
          pytest.mark.skip(reason="pre-alloc modified")
      ),
  ],
),
```

LouisTsai-Csie (changes requested):
I leave a comment about the potential missing coverage for non-empty storage scenario. In the deleted `tests/static/state_tests/stExtCodeHash/dynamicAccountOverwriteEmpty_ParisFiller.ym` file, i found this pre-alloc setting:

```py
# non-empty storage for target account
c5691dc90d9fd2a2e9a5fa5bd28bf77ffd60aa78:
  balance: '10'
  code: ''
  nonce: '0'
  storage: {
     "0x01": '1'
  }
```

while in the test this logic is removed.

chfast:
> I leave a comment about the potential missing coverage for non-empty storage scenario. In the deleted `tests/static/state_tests/stExtCodeHash/dynamicAccountOverwriteEmpty_ParisFiller.ym` file, i found this pre-alloc setting:
> 
> ```python
> # non-empty storage for target account
> c5691dc90d9fd2a2e9a5fa5bd28bf77ffd60aa78:
>   balance: '10'
>   code: ''
>   nonce: '0'
>   storage: {
>      "0x01": '1'
>   }
> ```
> 
> while in the test this logic is removed.

I dropped the "empty account with storage" because it actually prevents the original scenario. If there is storage, CREATE2 fails. Moreover, the EIP-7610 which describes what happens in this scenario is not final. This is my main motivation for the porting.

I believe if you want test `EXTCODEHASH` against "empty account with storage" we need separate test. Something that don't overwrite the account with `CREATE2`.

chfast:
Ping.

tests/constantinople/eip1052_extcodehash/__init__.py
```diff
@@ -0,0 +1 @@
+"""Cross-client EIP-1052 EXTCODEHASH Tests."""
```
danceratopz:
```suggestion
Tests for [EIP-1052: EXTCODEHASH](https://eips.ethereum.org/EIPS/eip-1052).
```

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,158 @@
...
+    deploy_code = Op.SSTORE(target_storage_slot, 1)
+    create2_initcode = Initcode(deploy_code=deploy_code)
+
+    caller_code = (
+        # EXTCODEHASH of the pre-CREATE2 target account.
+        Op.SSTORE(
+            caller_storage.store_next(keccak256(b"") if target_exists else 0),
+            Op.EXTCODEHASH(Op.CALLDATALOAD(0)),
+        )
+        # EXTCODESIZE of non-existent target account.
```
danceratopz: This is not strictly true when `target_exists=True`.
```suggestion
        # EXTCODESIZE of target account.
```

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,158 @@
...
+        + Op.SSTORE(
+            caller_storage.store_next(0), Op.EXTCODESIZE(Op.CALLDATALOAD(0))
+        )
+        # EXTCODECOPY of non-existent target account.
```
danceratopz:
```suggestion
        # EXTCODECOPY of target account.
```

tests/constantinople/eip1052_extcodehash/test_extcodehash.py
```diff
@@ -0,0 +1,158 @@
...
+        + Op.EXTCODECOPY(Op.CALLDATALOAD(0), 0, 0, 32)
+        + Op.SSTORE(caller_storage.store_next(0), Op.MLOAD(0))
+        # DELEGATECALL the non-existent target account.
```
danceratopz:
```suggestion
        # DELEGATECALL the target account.
```

danceratopz (commented):
Thanks @chfast! I checked the traces and indeed the original tests went OOG and didn't even test the intended behavior. The new Python versions do. The comments below are v minor nits.

I rebased to forks/amsterdam to add these on top of #2237 and #2249, but don't have rights to push to this branch. Please enable the rights or rebase this branch.

@LouisTsai-Csie could you please re-review and retract your "Requested Changes"?

LouisTsai-Csie (approved):
LGTM, thank you!

danceratopz (approved):
Thanks @chfast!

PR #2033, opened 2026-01-18
feat(tests): add worst-case BAL read test

nerolation:
## 🗒️ Description

Add EIP-7928 BAL test with maximum SLOADs in reverse lexicographic order.

- 7 transactions × 7,793 slots = 54,551 total SLOADs per block
- First access: highest slot, last access: slot 0

## ✅ Checklist

- [x] All: Ran `uvx tox -e static`
- [x] All: PR title adheres to repo standard


tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_max_sloads.py
```diff
@@ -0,0 +1,211 @@
...
+)
+
+from .spec import ref_spec_7928
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_7928.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7928.version
+
+pytestmark = pytest.mark.valid_from("Amsterdam")
+
+MAX_TX_GAS = 2**24  # 16,777,216
```
fselmo: You can import `fork` fixture into the test as a param and use it:

```py
def test_bal_max_sloads_reverse_order(
    pre: Alloc,
    fork: Fork,
) -> None:
    ...

    calculate_test_parameters(
        block_gas_limit, fork
    )
```

And then we can make these changes, so that they are updated if they ever change:

- `MAX_TX_GAS` -> `fork.transaction_gas_limit_cap()`
- `G_TRANSACTION` -> `gas_costs = fork.gas_costs(); gas_costs.G_TRANSACTION`

This makes the test fork agnostic, rather than using static vars.

fselmo (commented):
Just a minor comment as I saw this come through. Will take a look here this week 👀

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_compute_then_sload.py
```diff
@@ -0,0 +1,262 @@
...
+
+GAS_PER_COMPUTE_ITERATION = 60
+GAS_PER_SLOAD_ITERATION = 2_200
+
+
+def create_compute_then_sload_contract() -> Bytecode:
+    """
+    Create bytecode that first computes, then SLOADs.
+
+    Calldata: computation_iterations (32) + start_slot (32) + sload_count (32)
```
jochem-brouwer: Putting these pointers in the calldata means in this test that the transactions can be executed in parallel in the block without the need of BALs. To solve this (to force transactions being executed sequentially without BALs or to make parallel execution available if a BAL is provided) we need to make the transactions depend on the previous transaction.

The simplest way to do this is to just use the EVM storage to store the current start slot in (initialize this slot to 0xffff..ff so we start at the highest slot). For the SLOAD loop, use a `While` loop and the condition is a simple check to see if gasleft is enough to perform one more loop and the termination logic (which stores the current slot into the storage).

Can also be added to the computation loop. We can hardcode the "gasleft" there which is the threshold to continue the loop. For instance if the compute loop is 50%, then we just take 50% of the tx gas limit and use that as `GASLEFT` > `x` as `x` value for the condition.

Note that if we use the gas threshold in the EVM we do not have to manually calculate gas per loop. We also ensure that transactions depend on each other, since the second transaction has to start at the slot where the first one ended (cannot be executed in parallel without BALs). The EVM loss in cycles is nonzero, but it is negligible compared to the total SLOAD count.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_compute_then_sload.py
```diff
@@ -0,0 +1,262 @@
...
+def calculate_test_parameters(
+    block_gas_limit: int, fork: Fork, compute_fraction: float
+) -> tuple[int, int, int, int]:
+    """Calculate (num_transactions, compute_iters, sload_count, total_slots)."""
+    gas_costs = fork.gas_costs()
+    max_tx_gas = fork.transaction_gas_limit_cap()
+    assert max_tx_gas is not None
+
+    num_transactions = block_gas_limit // max_tx_gas
+    intrinsic_overhead = gas_costs.G_TRANSACTION + 600
```
jochem-brouwer: 600 magic number? Is this enough to cover the calldata?

I think we don't have to calculate this, we can construct the transaction and then calculate the intrinsic fee from there.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_compute_then_sload.py
```diff
@@ -0,0 +1,262 @@
...
+        num_transactions,
+        compute_iters_per_tx,
+        sload_count_per_tx,
+        total_slots,
+    ) = calculate_test_parameters(block_gas_limit, fork, compute_fraction)
+
+    storage = Storage({i: i + 1 for i in range(total_slots)})  # type: ignore
+
+    contract_code = create_compute_then_sload_contract()
+    contract = pre.deploy_contract(code=contract_code, storage=storage)
```
jochem-brouwer: I wonder if this works if we deploy a contract with a rather large storage (which cannot be deployed in one transaction)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_compute_then_sload.py
```diff
@@ -0,0 +1,262 @@
...
+    compute_fraction = compute_percent / 100.0
+
+    (
+        num_transactions,
+        compute_iters_per_tx,
+        sload_count_per_tx,
+        total_slots,
+    ) = calculate_test_parameters(block_gas_limit, fork, compute_fraction)
+
+    storage = Storage({i: i + 1 for i in range(total_slots)})  # type: ignore
```
jochem-brouwer: Doesn't this setup the storage up in a way 0 -> 1 -> 2? We want to read backwards now right?

jochem-brouwer (commented):
I did not review the tests in-depth, but in general: the current setup which uses calldata does not make BALs a requirement in order to execute the transactions in parallel (without BALs can run them in parallel). I suggest wasting some gas to store these pointers in the EVM. It saves gas calculation and we also directly get the transaction dependencies on the previous transactions (to make the BAL a requirement in order to execute them in parallel).

These tests also follow mostly the same format, I would also suggest to create a helper method or some kind of template to run these tests. The difference seems to be the attack code and the storage setup, but the files mostly (?) look the same.

More general question, should these tests be added in a BAL folder under benchmarks? I think these are not specifically consensus tests, they are rather targeted to perform DoS-like operations. @LouisTsai-Csie what do you think?

jochem-brouwer:
Oh, hmm, I just realized:

The max amount of cold SLOADs is 2 ** 24 / 2100 = 7989 per transaction. If we create a storage where the value of the storage points to the next key (I think this idea is already applied here?) then **we do not need a loop** because we can fit 8k bytes into the 24kb limit. (Code for the sload block is just `SLOAD SLOAD SLOAD SLOAD`)

Note that for different compute fractions this would thus influence the length of this SLOAD byte sequence.

(Even if slots are added to access list for the discount 2 ** 24 / 1900 = 8830 would still fit in the 24 kb)

tests/benchmark/stateful/bal/test_block_access_lists_compute_then_sload.py
```diff
@@ -0,0 +1,230 @@
...
+)
+from .spec import ref_spec_7928
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_7928.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7928.version
+
+pytestmark = pytest.mark.valid_from("Amsterdam")
+
+GAS_PER_COMPUTE_ITERATION = 60
+GAS_PER_SLOAD_ITERATION = 2_200
```
jochem-brouwer: This should be changed to use the dynamic gas calculations, on the target bytecode use `.gas_cost(fork)`

jochem-brouwer (commented):
The main thing to do here is to change the constants to calculate gas costs dependent on the fork. I will address this. Then I think we can make this review-ready and merge it 😄 🎉

tests/benchmark/stateful/bal/test_block_access_lists_compute_then_sload.py
nerolation: Great thanks! I'll then let you take over from here on

jochem-brouwer:
TODOs (I am working on this)

- [ ] Verify locally tests work as intended. (do not yet merge)
- [ ] Ensure these tests have to be run sequential if BAL not available (remove the calldata as input and instead store pointers to where next tx should start in storage)

fselmo:
@jochem-brouwer please feel free to ping me when this is ready for a pass 👀

jochem-brouwer:
@fselmo having to battle ruff/mypy, I hope CI passes but this is ready for review

tests/benchmark/stateful/eip7928_block_level_access_lists/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def cursor_read() -> Bytecode:
+    """PUSH1(CURSOR_SLOT) + SLOAD → stack: [..., cursor]."""
+    return Op.PUSH1(CURSOR_SLOT) + Op.SLOAD
+
+
+def cursor_write() -> Bytecode:
+    """PUSH1(CURSOR_SLOT) + SSTORE ← stack: [..., cursor]."""
+    return Op.PUSH1(CURSOR_SLOT) + Op.SSTORE
```
fselmo: The cursor slot is warm (setup `SLOAD`s it) and it's nonzero-to-nonzero. We need to update the metadata here. We can use `key_warm=True` which will allow us to squeeze in some more `SLOAD`s in the `test_max_sloads` calc.

```suggestion
    return Op.PUSH1(CURSOR_SLOT) + Op.SSTORE(
        key_warm=True,
        original_value=1,
        current_value=1,
        new_value=2,
    )
```

I patched this in locally and compared fills and this is what we get from the `--traces`:

```
  ┌──────────────────────┬───────────────────┬─────────────────┬─────────┐
  │        Metric        │ Original          │ Patch           │  Delta  │
  ├──────────────────────┼───────────────────┼─────────────────┼─────────┤
  │ Gas used             │ 979,300           │ 998,623         │ +19,323 │
  ├──────────────────────┼───────────────────┼─────────────────┼─────────┤
  │ Storage reads in BAL │ 453               │ 462             │ +9      │
  ├──────────────────────┼───────────────────┼─────────────────┼─────────┤
  │ Storage changes      │ 3                 │ 3               │ same    │
  ├──────────────────────┼───────────────────┼─────────────────┼─────────┤
  │ BAL accounts         │ 7                 │ 7               │ same    │
  └──────────────────────┴───────────────────┴─────────────────┴─────────┘


                                                                                                                                                                                                                 
  ┌───────────────────────────────┬──────────────────────┬───────────────────────┬─────────────────────────────────────────────────┐                                                                             
  │             Test              │ Uses cursor_write()  │ Uses plan_benchmark() │                     Benefit                     │
  ├───────────────────────────────┼──────────────────────┼───────────────────────┼─────────────────────────────────────────────────┤                                                                             
  │ test_bal_max_sloads           │ via default_teardown │          yes          │ +9 SLOADs/tx + corrected iteration estimate     │
  ├───────────────────────────────┼──────────────────────┼───────────────────────┼─────────────────────────────────────────────────┤                                                                             
  │ test_bal_max_pointer_chase    │ via default_teardown │          yes          │ +9 SLOADs/tx + corrected iteration estimate     │                                                                             
  ├───────────────────────────────┼──────────────────────┼───────────────────────┼─────────────────────────────────────────────────┤                                                                             
  │ test_bal_compute_then_sload   │ via default_teardown │          yes          │ +9 iterations/tx + corrected iteration estimate │                                                                             
  ├───────────────────────────────┼──────────────────────┼───────────────────────┼─────────────────────────────────────────────────┤                                                                             
  │ test_bal_max_account_access   │ via custom _teardown │          yes          │ +9 BALANCEs/tx + corrected iteration estimate   │
  ├───────────────────────────────┼──────────────────────┼───────────────────────┼─────────────────────────────────────────────────┤                                                                             
  │ test_bal_cross_contract_chase │    in dispatcher     │   no (own gas math)   │ longer chain per tx (~19K more gas for hops)    │
  └───────────────────────────────┴──────────────────────┴───────────────────────┴─────────────────────────────────────────────────┘  
```

Note, this can perhaps use an update to the docstring to explain why this metadata is important to help a reader of these tests. I'm a big proponent of spelling things out that may not be so obvious in order to make things more readable / accessible.

tests/benchmark/stateful/eip7928_block_level_access_lists/helpers.py
```diff
@@ -0,0 +1,248 @@
...
+        + Op.PUSH3(0)
+        + Op.GT
+        + Op.ISZERO
+        + Op.PUSH2(0)
+        + Op.JUMPI
+    )
+    loop_footer = Op.PUSH2(0) + Op.JUMP
+    overhead = loop_header.gas_cost(fork) + loop_footer.gas_cost(fork)
+    teardown_gas = teardown.gas_cost(fork)
+    gas_opcode_offset = (Op.JUMPDEST + Op.GAS).gas_cost(fork)
```
fselmo: I think there's something off here where we are not accounting for the `PUSH3(0)` because the bytecode here is re-built in an incorrect order from the bytecode in the `gas_check_loop_contract`.

Instead of re-building the bytecode, why don't we define this a bit more directly? Instead of applying a bandaid fix (adding the `PUSH3(0)`, we should instead define these bytecode sections as standalone helpers that can be put together and recalled when they need their `gas_cost(fork)` calculated. I can push this change here.

fselmo (commented):
Publishing some comments I had before I pushed the changes, for clarity. Will add some more to clarify some things.

tests/benchmark/stateful/eip7928_block_level_access_lists/helpers.py
```diff
@@ -39,8 +39,18 @@ def cursor_read() -> Bytecode:
...
-    """PUSH1(CURSOR_SLOT) + SSTORE ← stack: [..., cursor]."""
-    return Op.PUSH1(CURSOR_SLOT) + Op.SSTORE
+    """
+    PUSH1(CURSOR_SLOT) + SSTORE ← stack: [..., cursor].
+
+    SSTORE metadata reflects runtime: cursor slot is warm
+    (setup SLOADs it) and nonzero-to-nonzero.
+    """
+    return Op.PUSH1(CURSOR_SLOT) + Op.SSTORE(
+        key_warm=True,
```
fselmo, line 49: The metadata here is important for gas calc

tests/benchmark/stateful/eip7928_block_level_access_lists/test_block_access_lists_cross_contract_chase.py
```diff
@@ -123,19 +123,87 @@ def _compute_tx_gas_limits(
     return gas_limits
 
 
+def _max_chain_depth(
```
fselmo: Fixing the metadata in `SSTORE` (fixing the gas calc) caused this test to not fill. The calc here was not accounting for 63/64 rule chaining through.

fselmo (approved):
This lgtm now @jochem-brouwer. I'm going to approve but will wait for your final sanity check here to merge 👀

tests/benchmark/stateful/eip7928_block_level_access_lists/helpers.py
fselmo: Note: I added this in a commit (old notes / comment)

tests/benchmark/stateful/eip7928_block_level_access_lists/test_block_access_lists_cross_contract_chase.py
jochem-brouwer: Weird, on my local test it did fill (but I only tested for gas limit 1M so that might have been the problem)

fselmo: @jochem-brouwer it filled I think for some scenarios before the `metadata` fix which uses the correct gas calc for `SSTORE` based on it being warm and nonempty -> nonempty (if I'm remembering correctly). It was a bit of a circumstantial thing that it filled - I don't think for the right reasons.

fselmo: From Claude:


> The old _calculate_params used create_dispatcher_contract().gas_cost(fork) for dispatcher_cost. Before the metadata fix, cursor_write() used default SSTORE metadata — cold key, 0→nonzero — costing ~22,100 gas instead of the actual ~2,900.                                                               
                                                                                                                                                                                                                                                                                                                            
> That ~19,200 gas overestimate inflated dispatcher_cost, which shrank available, which shrank chain_by_gas = available // chain_hop_cost. The resulting chain_length happened to be small enough that all contracts were reachable despite the 63/64 compounding.                                                          
                                                                                                                                                                                                                                                                                                                            
> Once cursor_write() was fixed to report accurate gas (warm, nonzero→nonzero = 2,900), dispatcher_cost dropped, available grew, chain_length jumped to 100 (MAX_CALL_DEPTH), and the deep contracts couldn't be reached — test broke.                                                                                       
                             
> Two bugs canceling out: inflated SSTORE masked missing 63/64 accounting.

tests/benchmark/stateful/eip7928_block_level_access_lists/helpers.py
fselmo: Note, the above comment is old. I pushed a commit with this refactor.

jochem-brouwer:
There is a problem with `test_block_access_lists_cross_contract_chase.py`, I am going to remove it from this PR and open a new one, the others look fine, can we merge those? Opened #2588

fselmo:
> There is a problem with `test_block_access_lists_cross_contract_chase.py`, I am going to remove it from this PR and open a new one, the others look fine, can we merge those? Opened #2588

Yes I agree. I think this is 63/64 rule compounding with each nested call. We probably need to figure out the actual gas limit used since it doesn't use the per-tx-gas-cap. Then, we can squeeze more transactions in the block.

I figured we would tweak these a bit but I'm good with breaking it out and thinking through it more 👍🏼

jochem-brouwer:
Yes I think we should iterate on adding these benchs and not making one bench block the others (which look fine) to get merged. We should be able to run these benchs soon (TM) and that means that these benchs which we just merged can enter the feedback loop of actually checking what the clients do. If behavior is unexpected, we figure if the client is behaving weird or if the bench is weird :stuck_out_tongue_winking_eye: 

Thanks for merging, great work, this was an amazing collective effort :smile: :+1: @nerolation @fselmo @LouisTsai-Csie :heart_hands:

PR #2042, opened 2026-01-19
feat(ci): fixture release artifacts

spencer-tb:
### Problem

The existing fixture build process is slow, and requires us to wait a long time to generate new releases. Every time we re-release we are required to update the hive [workflow](https://github.com/ethpandaops/bal-devnets/blob/master/.github/workflows/hive-devnet-2.yaml#L71), and as this is not updated instantly, we end up with delays in test failures on clients during development.

### Solution

An overarching solution to this set of problems is to create fixture artifacts via github action artifacts. Given that these will be building every day, it will help us catch issues with fixture building for all features that we would otherwise miss. During development hive can be set to point at the latest develop (devnet) fixture artifact indefinitely. As less updates are required to the hive workflow, a faster feedback loop will occur when testing against clients. Under certain commit/path conditions test releases can happen instantly or significantly faster by downloading the artifacts instead of building them on the fly.

TLDR; this PR aims to unlock the following:
- Faster hive feedback loop by enabling automatic test/fixture updates, using fixture artifacts with the gh cli.
- Speed up the release process, where fixtures can use artifacts instead of having to rebuild manually for every release.
- Proactively catch fixture build failures for changes that have historically been missed when creating new releases.

### Compute Cost Analysis

The artifact builder is designed to be efficient. Here's how it compares to existing CI:

| Metric | test.yaml (existing CI) | fixture_build.yaml (new) |
|--------|------------------------|--------------------------|
| Runs per month | ~680 | ~180 checks, ~35-52 builds |
| Runner size | xl | gigachungus |
| % of CI volume | 100% | ~7% |

**Why so few builds?**
- Cron runs 6x/day but `check_artifacts.py` skips builds when:
  - Artifact already exists for current SHA
  - No relevant paths changed since last artifact
- `develop` branch rarely updates (~0 builds/month)
- `benchmark` + `benchmark_fast` share triggers

**Estimated builds per feature (based on last 30 days):**
| Feature | Builds/month | Reason |
|---------|-------------|--------|
| stable | ~18 | Active development on forks/amsterdam |
| develop | ~0 | devnets/bal/2 rarely updated |
| benchmark | ~17 | Triggered by common_paths changes |
| benchmark_fast | ~17 | Same triggers as benchmark |

### Release Speed Analysis

Based on the last 30 EEST releases (6 months):

| Release Type | Count | With Caching |
|--------------|-------|--------------|
| Full (vX.Y.Z) | 5 | Likely instant (stable branch) |
| bal/devnet | 16 | Instant if branch tracked |
| benchmark | 4 | Always instant (infrequent changes) |
| fusaka (untracked) | 5 | Always needs build |

**Estimate:** ~53% of releases would be instant (cached), ~47% would need builds. Untracked feature branches (like fusaka) always need builds since they're not in `feature.yaml`.

### New Artifacts

Artifacts are defined with the `feature.yaml` file, exactly as releases types are currently.

- `stable` — full test suite up to mainnet fork.
- `develop` — full test suite up to the development fork (devnet feature).
- `benchmark` — benchmark tests with multiple gas values (1, ..., 150).
- `benchmark_fast` — benchmark tests with single gas value (100).

### Feature Config

The `feature.yaml` config now supports artifacts and includes some new fields:
- `enabled` - whether artifacts are built for the feature, by default this is true.
- `branch` - the git branch to build from, might be a devnet/feature branch.
- `fill-params` - parameters passed to the fill command (use `--fill-static-tests` to enable lllc/solc).
- `test_paths` - feature specific file changes that trigger artifact rebuilds.

All features enabled for artifact building additionally use the `common_paths` filters (on top of `test_paths`). These together are the paths that are used to determine whether a new artifact should be built. For example if only the docs/readme changes in the most recent commit range since the last artifact build then we dont want to rebuild the artifact.

### Fixture Release Workflow

The `fixture_release.yaml` workflow consolidates the prior release workflows, by merging `release_fixture_full.yaml` and `release_fixture_feature.yaml` into a single file and offloading the fixture building logic to `fixture_build.yaml`. The latter lets releases take advantage of the fixture artifacts during the release flow.

Releases are now triggered via `workflow_dispatch` using the `gh` CLI:

```bash
# Full release
gh workflow run fixture_release.yaml -f version=1.0.0 -f release_type=full

# Feature release
gh workflow run fixture_release.yaml -f version=1.0.0 -f release_type=feature -f feature=benchmark_fast

# Force rebuild specific features (others use cache)
gh workflow run fixture_release.yaml -f version=1.0.0 -f release_type=full -f force_features=stable

# Force rebuild multiple features
gh workflow run fixture_release.yaml -f version=1.0.0 -f release_type=full -f force_features="stable,develop"
```

Release type determines which features are built:
- `full`: builds `stable` + `develop` + `benchmark`
- `feature`: builds a single specified feature

Since releases are now triggered via `workflow_dispatch` instead of tag push, tags are only created after a successful build/release. Previously, if a tag-triggered release failed, the tag would need to be manually deleted before retrying.

Before building, the workflow checks for cached artifacts per feature:
- Checks for an exact SHA match: `<feature>-<current_sha>`
- If no exact match, finds the most recent cached artifact
- Compares file changes between cached and current SHA
- Uses cached artifact if no relevant paths changed, otherwise triggers a fresh build
- Features in `force_features` list skip cache and always rebuild

#### Future Consideration

We could extend the gh cli trigger to allow us create feature releases without requiring any changes to the `feature.yaml` file, where we provide `branch`, `evm-type`, and `fill-params` as flags to the gh cli command.

### Fixture Build Workflow

The `fixture_build.yaml` workflow is the core CI workflow in this PR that will build our fixture artifacts moving forward. It runs every 4 hours (currently).

First it checks each enabled feature (stable, develop, etc) from `feature.yaml` for changes, starting with the `check_artifacts.py` script, which does the following:
- Get the current SHA for the features branch, `forks/amsterdam` for `stable`.
- Find the most recent cached artifact for the feature.
- Compare SHAs between cached artifact and branch.
- Identify which files changed between SHAs with `git diff`.
- Check whether the changed files match `test_paths` or `common_paths` within `feature.yaml` specified for the feature, the files that trigger a new artifact build.
- A list will be drafted for each feature that has changed and requires a new build.
- If none require a new artifact build the workflow will exit.

For every feature that requires a build, a parallel matrix job will be created that will do everything that the existing release flow already did:
- Reads required filling config options from `feature.yaml`: branch, evm-type, fill-params.
- Checks out the feature branch, for example `forks/amsterdam`.
- If `fill-params` contains `--fill-static-tests`, fetches `lllc`/`solc` (derived from the flag, no separate config field needed).
- Builds the EVM based on `evm-type`, for example evmone for benchmarking.
- Fills the fixtures to produce: `fixtures_<feature>.tar.gz`.
- If its a benchmark feature, it will generate a genesis tar: `benchmark_genesis.tar.gz`.
- These are then uploaded as artifacts with the following naming `<feature>-<sha>`.
  - This naming is what allows `check_artifacts.py` to extract the cached SHA for comparison on subsequent runs.
- Finally it will cleans up any old artifacts, keeping only the last 5 per feature.

### Hive Artifact Usage

Hive workflows can download the latest artifact directly using `gh run download` from the most recent fixture artifact, eliminating the need to wait for official releases or update workflow files when fixtures change. Hive will automatically run the most recent tests.

For the develop artifacts that are updated every time the `devnets/bal/2` branch has changes merged:
```
gh run download --repo ethereum/execution-specs --name develop-* --dir ./fixtures
```
This flow aligns nicely with the newer branching flow we have chosen for fork development, as devnet branches are only updated once all EIPs are merged and force pushed. That is a single push (change) each time.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


spencer-tb:
Closing. No longer needed.

PR #1930, opened 2025-12-16
refactor(test-cli-consume,rpc): use tenacity for retry logic

danceratopz:
## 🗒️ Description

This PR was intended to be a pure refactor, but streamlining the simulator retry logic was more involved than expected: 

- Add tenacity-based retry helpers to the RPC module for:
  -  forkchoice updates, 
  - block retrieval,
  - peer connection waiting.
- Consolidate duplicate `LoggedError` definitions into `helpers/exceptions.py`.
- Refactor `test_via_engine.py` and `test_via_sync.py` to use the new helpers, reducing boilerplate retry code.

**Note:** This is not a pure refactor. The retry mechanism in `test_via_sync.py` changes from exponential backoff to fixed intervals for simplicity and consistency; see "Retry Behavior Changes".

#### Bug Fixes

An incorrect enum comparison in the `consume sync` sync wait loop has been fixed: `if status.VALID:` → `if status == PayloadStatusEnum.VALID`. The old code accessed the enum class attribute via the instance (`status.VALID` returns `PayloadStatusEnum.VALID`), which is always truthy - causing the loop to break immediately regardless of the actual status.

TIL: Python enums allow accessing class members through instances. status.VALID is equivalent to type(status).VALID.

#### Speed-up

Speed-up in `consume sync` test execution, `consume sync -k "fork_Osaka-blockchain_test_sync" --input=develop@v5.4.0` with go-ethereum, [before](https://github.com/ethereum/execution-specs/commit/cafe03264f4629f179a1a8fdf9e5437dd3612654): ~30 seconds, after: ~17 seconds for 4 tests. 8 seconds come from removing two hard-coded 1 second sleeps which was to allow time for peers to connect. I think the rest might come from faster polling in the initial genesis fcu update, but didn't check more closely.

#### RPC Module (`rpc/rpc.py`)

- Add `forkchoice_updated_with_retry()` function that retries while status is `SYNCING` and returns on any terminal status (`VALID`, `INVALID`, `ACCEPTED`).
- Add `EthRPC.get_block_by_hash_with_retry()` method that retries while block is `None`.
- Add `NetRPC.wait_for_peer_connection()` method that polls `net_peerCount` until peers are connected.
- Add exceptions with diagnostic info: `ForkchoiceUpdateTimeoutError`, `BlockNotAvailableError`, `PeerConnectionTimeoutError`.

#### Simulator Logic

- Replace manual retry loops with tenacity-based helpers.
[...]

PR #2056, opened 2026-01-21
feat(tools): retry rpc calls on server-side error

SamWilsn:
## 🗒️ Description

The EF's RPC endpoints return 500 errors pretty often, so this pull request adds a retry mechanism.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/sync.py
+++ b/src/ethereum_spec_tools/sync.py
@@ -188,6 +188,34 @@ class BlockDownloader(ForkTracking):
                     except Full:
                         pass
 
+    def _make_request(self, req: request.Request) -> Any:
+        backoff = 1.75
+        start = time.monotonic()
+        last_error = None
+        delay = 1.0
+        while True:
+            now = time.monotonic()
+            elapsed = now - start
+            remaining = (60.0 * 60.0) - elapsed
+
+            if 0.0 >= remaining:
+                assert last_error is not None
+                raise last_error
+
+            try:
+                with request.urlopen(req, timeout=60.0) as response:
+                    return json.load(response)
+            except request.HTTPError as e:
+                if e.code < 500 or e.code > 599:
+                    raise
+                logging.warning(
+                    "server-side error during RPC request", exc_info=True
+                )
+                last_error = e
+
+            time.sleep(min(delay, remaining))
+            delay *= backoff
+
     def fetch_blocks(
         self,
         first: Uint,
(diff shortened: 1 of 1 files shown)


src/ethereum_spec_tools/sync.py
SamWilsn, line 1: Best to review with "hide whitespace" enabled:

image

gurukamath (approved):

danceratopz (commented):
Hi @SamWilsn! How do you feel about using [tenacity](https://github.com/jd/tenacity) to avoid custom boilerplate code? 

It's packages/testing dep already. For more context / examples, see:
- https://github.com/ethereum/execution-specs/pull/1930

Introduced here:
- https://github.com/ethereum/execution-spec-tests/pull/2205

SamWilsn:
Not opposed to it! I was considering reaching for urllib3, which also does retries, but if we're already doing tenacity, I'll look into it.

SamWilsn:
So, @danceratopz, it looks like the sync tool doesn't currently have any external dependencies. Would you recommend adding a new `tool` or `sync` dependency group?

src/ethereum_spec_tools/sync.py
```diff
@@ -188,6 +188,31 @@ def download(self) -> None:
...
 
+    def _make_request(self, req: request.Request) -> Any:
+        backoff = 1.75
+        start = time.monotonic()
+        last_error = None
+        delay = 1.0
+        while True:
+            now = time.monotonic()
+            elapsed = now - start
+            remaining = (60.0 * 60.0) - elapsed
```
danceratopz, line 199: Just to check, is it intentional to retry for an hour (3600 seconds)? If so, should we log intermittently (log a warning once, and then again on every N seconds) to inform the user? There's currently no user feedback (apart from failure) in this retry loop.

Also, this is quite an important value, how about defining it above as `max_retry_seconds`?

src/ethereum_spec_tools/sync.py
```diff
@@ -188,6 +188,31 @@ def download(self) -> None:
...
+
+            if 0.0 >= remaining:
+                assert last_error is not None
+                raise last_error
+
+            try:
+                with request.urlopen(req, timeout=60.0) as response:
+                    return json.load(response)
+            except request.HTTPError as e:
```
danceratopz, line 208: Perhaps `HTTPError` is sufficient, but should we consider expanding this to include other transient failure modes? E.g., `urllib.error.URLError` (checking `e.reason` for flaky, non-persistent modes)?

Perhaps this comment is moot, if you're checking a local full node? But it seems like  an PRC endpoint is typically used?

danceratopz (commented):
> So, @danceratopz, it looks like the sync tool doesn't currently have any external dependencies. Would you recommend adding a new `tool` or `sync` dependency group?

Ah I see, sorry for holding this up, @SamWilsn! I'm fine to leave it as-is if we'd like to keep dependencies minimal for `ethereum-execution`! As we have more boilerplate retry code in `execution-testing`, I prefer to use a library there.

If you'd like to keep as-is, I have a couple of comments/questions below!

src/ethereum_spec_tools/sync.py
SamWilsn: This process typically runs for months, so an hour retry before failing seemed reasonable to me. A warning is a great idea though!

src/ethereum_spec_tools/sync.py
SamWilsn: So far I've only hit 500 error codes. I'm not opposed to handling more of them, but for now, I'm happy with this.

I think I'm this tool's only user ;)

PR #2063, opened 2026-01-22
feat(specs): Introduce comprehensive state tracking objects

petertdavies:
## 🗒️ Description
This PR adds two new types for tracking state:

* `TxStateTracking`: Tracks reads and writes that happened during a transaction.
* `BlockStateTracking`: Tracks reads and writes that happened during a block.

`BlockStateTracking` should be convertible into a BAL. Most of the functions of the `State` object have superseded, it is still used to:

1. Answer queries about the pre-state.
2. Calculate the state root.
3. Communicate state changes back to the caller.

A future PR should be replace it with a suitable interface to do those things.

**Note:** This PR is an early draft. All pre-Amsterdam forks are broken (need back-porting) and code cleanup is required. EIP-7610 is unimplemented.

## Next steps

* Rebase EIP-7928 on this branch and port to use state tracking objects.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.- [ ] All: Set appropriate labels for the changes (only maintainers can apply labels).

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/state_tracking.py
+++ b/src/ethereum/forks/amsterdam/state_tracking.py
@@ -1,0 +1,520 @@
+"""
+State Tracker.
+"""
+
+from dataclasses import dataclass
+from typing import Callable, Dict, List, Optional, Set, Tuple
+
+from ethereum_types.bytes import Bytes, Bytes32
+from ethereum_types.frozen import modify
+from ethereum_types.numeric import U256, Uint
+
+from . import state as state_
+from .fork_types import EMPTY_ACCOUNT, Account, Address
+from .state import State
+
+BlockAccessIndex = Uint
+
+
+@dataclass
+class BlockStateTracking:
+    """
+    Tracking of state changes at the block level.
+    """
+
+    parent: State
+    current_index: BlockAccessIndex
+
+    account_reads: Set[Address]
+    storage_reads: Set[Tuple[Address, Bytes32]]
+
+    account_writes: Dict[
+        Address, List[Tuple[BlockAccessIndex, Optional[Account]]]
+    ]
+    storage_writes: Dict[
+        Address, Dict[Bytes32, List[Tuple[BlockAccessIndex, U256]]]
+    ]
+
+
+@dataclass
+class TxStateTracking:
+    """
+    Tracking of state changes within a transaction.
+    """
+
+    parent: BlockStateTracking
+
+    account_reads: Set[Address]
+    storage_reads: Set[Tuple[Address, Bytes32]]
+
+    account_writes: Dict[Address, Optional[Account]]
+    storage_writes: Dict[Address, Dict[Bytes32, U256]]
+
+    created_accounts: Set[Address]
+
+
+# get_account_optional
+# set_account
+# get_storage
+# set_storage
+# mark_account_created
+# destroy_storage
+
+
+def get_account_optional(
+    state: TxStateTracking | BlockStateTracking, address: Address
+) -> Optional[Account]:
+    """FIXME."""
+    if isinstance(state, TxStateTracking):
+        if address in state.account_writes:
+            return state.account_writes[address]
+        else:
+            return get_account_optional(state.parent, address)
+    elif isinstance(state, BlockStateTracking):
+        if address in state.account_writes:
+            return state.account_writes[address][-1][1]
+        else:
+            return state_.get_account_optional(state.parent, address)
+
+
+def set_account(
+    state: TxStateTracking | BlockStateTracking,
+    address: Address,
+    account: Optional[Account],
+) -> None:
+    """FIXME."""
+    if isinstance(state, TxStateTracking):
+        state.account_writes[address] = account
+    elif isinstance(state, BlockStateTracking):
+        writes = state.account_writes.get(address, [])
+        if writes == [] or writes[-1][0] < state.current_index:
+            writes.append((state.current_index, account))
+        else:
+            writes[-1] = (state.current_index, account)
+        state.account_writes[address] = writes
+
+
+def get_storage(
+    state: TxStateTracking | BlockStateTracking, address: Address, key: Bytes32
+) -> U256:
+    """FIXME."""
+    if isinstance(state, TxStateTracking):
+        if key in state.storage_writes.get(address, {}):
+            return state.storage_writes[address][key]
+        else:
+            return get_storage(state.parent, address, key)
+    elif isinstance(state, BlockStateTracking):
+        writes = state.storage_writes.get(address, {}).get(key, [])
+        if writes == []:
+            return state_.get_storage(state.parent, address, key)
+        else:
+            return writes[-1][1]
+
+
+def set_storage(
+    state: TxStateTracking | BlockStateTracking,
+    address: Address,
+    key: Bytes32,
+    value: U256,
+) -> None:
+    """FIXME."""
+    if isinstance(state, TxStateTracking):
+        if address not in state.storage_writes:
+            state.storage_writes[address] = {}
+        state.storage_writes[address][key] = value
+    elif isinstance(state, BlockStateTracking):
+        writes = state.storage_writes.get(address, {}).get(key, [])
+        if writes == [] or writes[-1][0] < state.current_index:
+            writes.append((state.current_index, value))
+        else:
+            writes[-1] = (state.current_index, value)
+        if address not in state.storage_writes:
+            state.storage_writes[address] = {}
+        state.storage_writes[address][key] = writes
+
+
+def destroy_storage(state: TxStateTracking, address: Address) -> None:
+    """FIXME."""
+    if address in state.storage_writes:
+        del state.storage_writes[address]
+
+
+def bump_index(state: BlockStateTracking) -> None:
+    """FIXME."""
+    state.current_index += Uint(1)
+
+
+def mark_account_created(state: TxStateTracking, address: Address) -> None:
+    """FIXME."""
+    state.created_accounts.add(address)
+
+
+def incorporate_tx_state_into_parent(tx_state: TxStateTracking) -> None:
+    """
+    After a transaction has finished, write the final state changes to the
+    parent.
+    """
+    parent = tx_state.parent
+    parent.account_reads |= tx_state.account_reads
+    parent.storage_reads |= tx_state.storage_reads
+    for address, account in tx_state.account_writes.items():
+        set_account(parent, address, account)
+    for address, items in tx_state.storage_writes.items():
+        for key, value in items.items():
+            set_storage(parent, address, key, value)
+
+
+def write_block_state_changes(block_state: BlockStateTracking) -> None:
+    """
+    At the end of the block. Write the final state changes to the state object.
+
+    This is a temporary arrangement, until a future refactor abolishes the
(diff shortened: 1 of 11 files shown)


src/ethereum/forks/amsterdam/utils/message.py
```diff
@@ -16,8 +16,8 @@
 from ethereum_types.numeric import Uint
 
 from ..fork_types import Address
-from ..state import get_account
 from ..state_tracker import create_child_frame
+from ..state_tracking import get_account
```
SamWilsn, line 20: This is probably part of the cleanup you're planning, so ignore this comment (and any others) in that case.

But, `state_tracker` and `state_tracking` are not distinguishable enough.

src/ethereum/forks/amsterdam/vm/__init__.py
```diff
@@ -40,6 +44,7 @@ class BlockEnvironment:
 
     chain_id: U64
     state: State
+    state_tracking: BlockStateTracking
```
SamWilsn, line 47: I assume it'll be possible to rename this to just `state` eventually? If there's only one state interface, no reason to keep the `_tracking` suffix.

src/ethereum/forks/amsterdam/vm/__init__.py
```diff
@@ -108,12 +113,13 @@ class TransactionEnvironment:
     gas: Uint
     access_list_addresses: Set[Address]
     access_list_storage_keys: Set[Tuple[Address, Bytes32]]
-    transient_storage: TransientStorage
```
SamWilsn, line 111: This feels odd. Isn't transient storage conceptually scoped to the tx env?

src/ethereum/forks/amsterdam/vm/__init__.py
SamWilsn, line 170: Blegh, this stuff has too many layers. You shouldn't have to go `Evm` -> `Message` -> `TransactionEnvironment`...

src/ethereum/forks/amsterdam/vm/__init__.py
```diff
@@ -184,6 +197,8 @@ def incorporate_child_on_success(evm: Evm, child_evm: Evm) -> None:
     evm.accounts_to_delete.update(child_evm.accounts_to_delete)
     evm.accessed_addresses.update(child_evm.accessed_addresses)
     evm.accessed_storage_keys.update(child_evm.accessed_storage_keys)
+    evm.state_tracking = child_evm.state_tracking
+    evm.transient_storage = child_evm.transient_storage
```
SamWilsn, line 201: Might be worth a comment in these two sibling functions (on_success and on_error) explaining what the implications are of assigning the entire object vs. unioning. I assume that the assignment is a shorthand for unioning all the fields, and the object identity is irrelevant?

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -115,12 +113,12 @@ def process_message_call(message: Message) -> MessageCallOutput:
...
     """
-    block_env = message.block_env
+    state_tracking = message.state_tracking
     refund_counter = U256(0)
     if message.target == Bytes0(b""):
         is_collision = account_has_code_or_nonce(
-            block_env.state, message.current_target
-        ) or account_has_storage(block_env.state, message.current_target)
+            state_tracking, message.current_target
+        )  # or account_has_storage(block_env.state, message.current_target)
```
SamWilsn, line 121: Intentional comment?

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -150,10 +149,12 @@ def process_message_call(message: Message) -> MessageCallOutput:
     if evm.error:
         logs: Tuple[Log, ...] = ()
         accounts_to_delete = set()
+        state_tracking = message.state_tracking
     else:
         logs = evm.logs
         accounts_to_delete = evm.accounts_to_delete
         refund_counter += U256(evm.refund_counter)
+        state_tracking = evm.state_tracking
```
SamWilsn, line 157: My spidey-sense is tingling a bit here. I don't really doubt your correctness, but this could probably do with some comments. Why do we discard the `evm`'s `state_tracking` on error?

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -331,6 +277,58 @@ def process_message(message: Message) -> Evm:
...
+        return_data=b"",
+        error=None,
+        accessed_addresses=message.accessed_addresses,
+        accessed_storage_keys=message.accessed_storage_keys,
+        state_changes=message.state_changes,
+        state_tracking=copy_tx_state_tracking(message.state_tracking),
+        transient_storage=message.transient_storage,
+    )
+
+    if create:
```
SamWilsn, line 304: Isn't there a `process_create_message`? If this check needs to be specifically where it is, we really should refactor `process_message` and `process_create_message` and hopefully avoid having to pass in a bool.

src/ethereum/forks/amsterdam/state_tracking.py
```diff
@@ -0,0 +1,520 @@
...
+        del state.storage_writes[address]
+
+
+def bump_index(state: BlockStateTracking) -> None:
+    """FIXME."""
+    state.current_index += Uint(1)
+
+
+def mark_account_created(state: TxStateTracking, address: Address) -> None:
+    """FIXME."""
```
SamWilsn, line 148: Best to omit these instead of adding a placeholder. It's annoying, but less chance of missing it before merging.

src/ethereum/forks/amsterdam/state_tracking.py
```diff
@@ -0,0 +1,520 @@
...
+    state.created_accounts.add(address)
+
+
+def incorporate_tx_state_into_parent(tx_state: TxStateTracking) -> None:
+    """
+    After a transaction has finished, write the final state changes to the
+    parent.
+    """
+    parent = tx_state.parent
```
SamWilsn, line 157: I'd really like to avoid a second stack if we can. If this is always two deep (block -> tx), can we omit this parent reference, and instead do the incorporating from the block's perspective? Instead of the transaction pushing itself up into the block, have the block pull the transaction into itself.

src/ethereum/forks/amsterdam/state_tracking.py
```diff
@@ -0,0 +1,520 @@
...
+    Tracking of state changes within a transaction.
+    """
+
+    parent: BlockStateTracking
+
+    account_reads: Set[Address]
+    storage_reads: Set[Tuple[Address, Bytes32]]
+
+    account_writes: Dict[Address, Optional[Account]]
+    storage_writes: Dict[Address, Dict[Bytes32, U256]]
```
SamWilsn, line 51: I remember talking about this, but I don't remember how we resolved it.

IMHO it might be better to have one state tracking object that always includes the block access index?

src/ethereum/forks/amsterdam/utils/message.py
petertdavies: Yeah. It's a mess. There are a number of places in this PR where new code has been added, but the code it is supposed to replace hasn't been removed yet.

src/ethereum/forks/amsterdam/vm/__init__.py
petertdavies: I think the final names should be:

`State` -> `PreState`
`BlockStateTracking` -> `BlockState`
`TxStateTracking` -> `TxState`

Everything should go in single module called `state`.

src/ethereum/forks/amsterdam/vm/__init__.py
petertdavies: The pattern is:
* When a new subframe is created, copy the object
* Make subframe changes to the copy
* If the subframe fails, discard the copy
* If the subframe succeeds, replace the parent object with the copy (thus preserving the changes)

Will add comments with an explanation.

src/ethereum/forks/amsterdam/vm/__init__.py
petertdavies: The `Message`/`Evm` distinction is a mess, `Message` is basically a proto-`Evm` object. Either we should create `Evm` earlier and not have `Message` or we should discard `Message` after `Evm` is created (copying the fields over).

src/ethereum/forks/amsterdam/vm/interpreter.py
petertdavies: Temporary hack until `account_has_storage` gets implemented

src/ethereum/forks/amsterdam/vm/interpreter.py
petertdavies: This create-specific code requires `Evm` to exist, which hasn't happened yet in `process_message`/`process_create_message`. I think `Evm` needs to be created earlier.

src/ethereum/forks/amsterdam/state_tracking.py
petertdavies: The stack is 3 deep (prestate -> block -> tx), it exists because functions like `get_account` work down the stack:
1. Has the account changed in `TxStateTracking`?
2. If not, has the account changed in `BlockStateTracking`?
3. If not, is it in the prestate (confusingly called `State` atm)?

We could get rid of the duplicate stack if all the state functions took `Evm | BlockEnviroment` rather than `TxStateTracking | BlockStateTracking`, but that can only happen if `Evm` is created before any state reads/writes.

src/ethereum/forks/amsterdam/state_tracking.py
petertdavies: I think they are just different enough to justify different objects, but it's arguable either way. I think we should leave as-is for now and merge them later if they end up being too similar.

src/ethereum/forks/amsterdam/vm/interpreter.py
gurukamath: In BALs, upon error, we convert state writes into reads and preserve them instead of fully discarding them. Might be something to keep in mind

src/ethereum/forks/amsterdam/vm/interpreter.py
gurukamath: I would say, we create the EVM earlier and not create the `Message` at all if possible. The line between those two has always been pretty blurry

src/ethereum/forks/amsterdam/state_tracking.py
```diff
@@ -0,0 +1,520 @@
...
+    """
+    Tracking of state changes within a transaction.
+    """
+
+    parent: BlockStateTracking
+
+    account_reads: Set[Address]
+    storage_reads: Set[Tuple[Address, Bytes32]]
+
+    account_writes: Dict[Address, Optional[Account]]
```
gurukamath, line 50: Would we also need an independent `storage_cleared: Set[Address]` attribute to track all the `destroy_storage` calls? In a non EIP-7610 world

petertdavies:
Closed in favour of #2207

Issue #1874, opened 2025-12-09
EIP-7778 Implementation Tracker: Block Gas Accounting without Refunds


marioevz:
### Target Fork

**Amsterdam**

### Instructions

- [ ] Assign issue to EIP specification and testing owner(s).

> [!IMPORTANT]
> A specifications specialist and a testing specialist should ideally share ownership of the EIP.

- [x] Add the issue to the target fork milestone if applicable (i.e., the EIP is at least in the [CFI stage](https://eips.ethereum.org/EIPS/eip-7723#considered-for-inclusion)).

#### Guidance for Marking Items Complete

An item should only be checked off once the EIP is considered *stable*. In this context, stable means:

- No major issues or ambiguities are still being uncovered in the specification or tests.
- There are no open discussion points awaiting resolution.
- Client implementations have been consistently passing the tests for at least a week.

It is ultimately up to the owners' discretion to decide when an item should be marked as complete, using this guidance as the basis for that decision.

In exceptional cases, an EIP may require changes after some items have been marked complete or even after the entire issue has been completed and closed. This can happen, for example, when significant design optimizations are identified and agreed upon in ACD, or when critical security issues surface and require updates to the specification or tests.

When this occurs, owners should either unmark the relevant checkboxes if the issue is still open, or create a new tracking issue for the modifications if the original issue had already been closed.

### Specification + Testing Status

- [x] Testing complexity assessed and documented: [EIP-7778.md](https://github.com/ethsteel/pm/blob/main/complexity_assessments/EIPs/EIP-7778.md) - 🟡 10.
- [x] Specification implementation merged to `eips/amsterdam/eip-7778` *(skip if the fork branch merge below is already complete)*.
- [x] Specification updates merged to the corresponding `forks/amsterdam` branch.
- [x] EIP updates proposed in case of architectural choices surfaced during implementation.
- [x] Required testing framework modifications implemented.
- [x] Test suite implemented.
- [x] Full code coverage for all changes.
- [x] No regressions or failures in tests from prior forks (including static tests).
- [x] [Testing checklist](https://github.com/ethereum/execution-specs/blob/HEAD/docs/writing_tests/checklist_templates/eip_testing_checklist_template.md) complete.
- [x] Hardening session completed.
- [x] Benchmarking tests written and results documented.
- [x] Ran tests using `execute` to ensure compatibility, and marked specific tests to be skipped when they cannot be executed on live networks.
[...]

PR #2074, opened 2026-01-26
feat(spec-tests): Add EIP-7778 tests with multiple refund types

gurukamath:
## 🗒️ Description
This PR does 2 things

1. Refactor the refund transaction building mechanism into its own helper function so each test doesn't have to build the `refund_tx` from scratch
2. Add a test that has multiple refund types within a single transaction

## 🔗 Related Issues or PRs
#1874 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Cute Animals - 7 of 12

### DIFF
--- a/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
+++ b/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
@@ -482,3 +482,49 @@ def test_varying_calldata_costs(
         ],
         post=post,
     )
+
+
+@pytest.mark.parametrize(
+    "refund_tx_reverts",
+    [
+        pytest.param(True, id="refund_tx_reverts"),
+        pytest.param(False, id=""),
+    ],
+)
+@pytest.mark.execute(pytest.mark.skip(reason="Requires specific gas price"))
+@pytest.mark.valid_from("Amsterdam")
+def test_multiple_refund_types_in_one_tx(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    refund_tx_reverts: bool,
+) -> None:
+    """Test gas accounting for all refund types available in the given fork."""
+    refunds_count = 10
+
+    post = Alloc()
+    refund_types = set(fork.refund_types())
+
+    (_, gas_used_pre_refund, call_data_floor_cost, refund_tx) = (
+        build_refund_tx(
+            fork=fork,
+            pre=pre,
+            post=post,
+            refund_types=refund_types,
+            refunds_count=refunds_count,
+            refund_tx_reverts=refund_tx_reverts,
+        )
+    )
+
+    refund_tx_block_gas_used = max(gas_used_pre_refund, call_data_floor_cost)
+
+    blockchain_test(
+        pre=pre,
+        blocks=[
+            Block(
+                txs=[refund_tx],
+                expected_gas_used=refund_tx_block_gas_used,
+            )
+        ],
+        post=post,
+    )
--- a/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
+++ b/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
@@ -4,6 +4,7 @@ Test cases for
 """
 
 from enum import Enum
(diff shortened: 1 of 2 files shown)


LouisTsai-Csie (approved):
I love this refactor! It is incredible. But it seems there is some issue with the `changelog` changes.

btw, I noticed some dotted lines in the codebase, but linting passes without issues. Not sure whether we’re missing some configuration or if this warning should simply be disabled, probably the latter.

Image

tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
```diff
@@ -28,144 +28,127 @@
...
+    refund_tx_extra_gas = 1 if refund_tx_has_extra_gas_limit else 0
 
-            if refund_tx_reverts:
-                post[contract_address] = Account(
-                    storage=dict.fromkeys(storage_slots, 1),
-                )
-            else:
-                post[contract_address] = Account(
-                    storage=dict.fromkeys(storage_slots, 0),
+    for refund_type in refund_types:
```
marioevz, line 62: I don't quite understand this `for` loop here because we only pass a single refund type each time.

tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
```diff
@@ -28,144 +28,127 @@
...
+    gas_used_pre_refund = intrinsic_cost_calc(
+        calldata=call_data,
+        return_cost_deducted_prior_execution=True,
+        authorization_list_or_count=authorization_list,
+    ) + code.gas_cost(fork)
+
+    # Calculate refund (still applied to user's balance)
+    refund_counter = gsc.R_AUTHORIZATION_EXISTING_AUTHORITY * count_auth_tuples
+    if not refund_tx_reverts:
+        refund_counter += code.refund(fork)
```
marioevz: Perhaps all of this logic could sandwich the match-case logic like so:
```python
authorization_list = None
refund_counter = 0
match refund_type:
    case RefundTypes.STORAGE_CLEAR:
        for slot in storage_slots:
            code += Op.SSTORE(
                slot,
                Op.PUSH0,
                # Gas accounting
                original_value=1,
                new_value=0,
            )
        empty_storage_on_success = True

    case RefundTypes.AUTHORIZATION_EXISTING_AUTHORITY:
        code += Op.PUSH0
        authorization_list = [
            AuthorizationTuple(
                address=contract_address,
                nonce=0,
                signer=pre.fund_eoa(amount=1),
            )
            for _ in range(refunds_count)
        ]
        refund_counter += gsc.R_AUTHORIZATION_EXISTING_AUTHORITY * refunds_count

    case _:
        raise ValueError(
            f"Unknown refund type: {refund_type} (Test needs update)"
        )
if not refund_tx_reverts:
  refund_counter += code.refund(fork)
```
Just to not have side-effects disjointed from the code that produces them I think.

marioevz (commented):
Just some small cosmetic comments, thanks!

tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
gurukamath: If one wants to build a refund transaction that has multiple type of refunds within a single transaction. I could pass something like `refund_types = {STORAGE_CLEAR, AUTHORIZATION_EXISTING_AUTHORITY}` to have both refund types in a single transaction. (or any combination of refund types once there are more than just those 2)

tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
gurukamath: Ah. You are right. I went with this approach because we won't have the `delegated_contract` for creating the auth_tuples until we have the full code and have pre_deployed the contract with that code.

However, as I think more about this, the delegated contract does not necessarily have to be the same as the contract called by the transaction. So I can just deploy a dummy delegated contract and use that for the auth tuples. I have updated with this approach

gurukamath:
@marioevz @LouisTsai-Csie Have rebased the PR on top of the changes introduced in #2073

PR #2078, opened 2026-01-26
fix(tests): fix EIP-7934 tests logic accuracy

fselmo:
## 🗒️ Description

We had some issues filling EIP-7934 tests for Amsterdam after introducing tests for EIP-7843, that we very quickly patched ([here](https://github.com/fselmo/execution-specs/commit/a64c42add7fd4855d3279b6bb2644b47931b9764#diff-5e81bf7f76169cc7ee4c5f8980679196b0c222055826aee1ae7cd3e8868d3f34R48)) so we could get the tests released. This "fix" was more of a bandaid fix that was likely going to leak issues out again eventually. This PR aims to fix this in a more proper way as this is still currently failing one test in `forks/amsterdam` since we haven't introduced this fix here.

- First few commits here Claude added +/- 1byte tolerance which helps us not have to find a perfect match (original problem). Then optimized the search by changing from linear to binary search.
- I then went down a path to play with increasing the starting `extraData` to half of what it can be. `extraData` is a 32-byte field, so if we start somewhere in the middle we can have even better tolerance. It turns out that our calldata estimate for the last tx is good enough where we're always within ~5 bytes of what we need. This actually removes the need for the binary search entirely as we have plenty of tolerance to adjust with block `extraData`.

This PR brings the `fill` time for these tests down significantly to where I don't believe these tests are a burden on `fill` anymore.

---
Bonus:

Fix typo in EIP-7251 BAL tests that was just merged to `forks/amsterdam` and slipped by PR reviews.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2026-01-26 at 16 14 53

### DIFF
--- a/tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py
+++ b/tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py
@@ -364,62 +364,62 @@ def _exact_size_transactions_impl(
             )
 
             if test_size == block_size_limit:
-                # if exact match, use the transaction
                 transactions.append(test_tx)
             else:
-                # search for the best adjustment
+                # Binary search for the calldata size that hits the target.
+                # Block size is monotonically non-decreasing with calldata
+                # length, so binary search converges in ~O(log N) iterations.
                 diff = block_size_limit - test_size
+                search_range = min(abs(diff) + 50, 1000)
+                low = max(0, estimated_calldata - search_range)
+                high = estimated_calldata + search_range
+
                 best_diff = abs(diff)
                 best_tx = test_tx
 
-                search_range = min(abs(diff) + 50, 1000)
+                while low <= high:
+                    mid = (low + high) // 2
+                    mid_calldata = b"\x00" * mid
+                    mid_gas = calculator(calldata=mid_calldata)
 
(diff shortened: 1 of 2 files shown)


tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py
```diff
@@ -123,13 +126,17 @@ def exact_size_transactions(
     emit_logs: bool = False,
     specific_transaction_to_include: Transaction | None = None,
     withdrawals: List[Withdrawal] | None = None,
-) -> Tuple[List[Transaction], int]:
+) -> Tuple[List[Transaction], int, int]:
```
gurukamath: I'm not sure if this function needs to return `gas_used` any more. I don't see it being used.

gurukamath (approved):
Just a minor nit about the values returned by the `exact_size_transactions`. Looks good to me otherwise

tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py
fselmo: Nice catch! ty!

PR #2079, opened 2026-01-27
feat(test-fill): speed up filling

fselmo:
## 🗒️ Description

- Build index incrementally on the fly, across runners, and merge at the end. Remove the bottleneck at the end of filling, re-reading all files in a single main thread.
- Defer index model validation until merge, append-only:
    - Claude summary is worth it here:
      ```
      Before: Each write_partial_index() call (which happens per module scope) had to:                                                                                                                                                                                                                                          
        1. Acquire FileLock                                                                                                                                                                                                                                                                                                       
        2. Read the entire existing JSON file                                                                                                                                                                                                                                                                                     
        3. Parse it (json.loads)                                                                                                                                                                                                                                                                                                  
        4. Append new entries to the list                                                                                                                                                                                                                                                                                         
        5. Serialize the entire thing back (json.dumps)                                                                                                                                                                                                                                                                           
        6. Write the whole file                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
        Every write re-reads and re-writes everything written so far. For a worker that processes 500 modules, the 500th write re-serializes all entries from the previous 499.                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                            
      After: Each call just:                                                                                                                                                                                                                                                                                                    
        1. Acquire FileLock                                                                                                                                                                                                                                                                                                       
        2. Open file in append mode ("a")                                                                                                                                                                                                                                                                                         
        3. Write new lines                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                            
        No reading, no parsing, no re-serialization of existing data. O(new entries) per write instead of O(all entries so far).```
- Use a trie-building approach for `HashableItem.from_index_entries()` and add tests to ensure same hashes are produced - from O(n^2) to O(n):
  - Every folder in the hierarchy used to scan all files twice. From Claude: `Each entry is now inserted once by walking its path components (bounded depth, typically 3-5 levels). The conversion visits each trie node exactly once. Total: O(n). No redundant scanning, no relative_to(), no ValueError try/except.`
- Use `--no-html` for releases since we don't publish the html in the artifacts, this is not needed (small win)
- Validate at `IndexFile` once, not at every single file (just one validation instead of potentially +100k times)
- Distribute slow-marked tests to workers first so we don't hang at the end of test runs (no bricked slow tail)
- Also build fixtures in parts via workers and merge at the end... this reduces test `teardown` phase significantly (seeing some improvements from ~80s to ~1` in some longer tests). 
- Turns on `--durations=50` for `py3` run so we can see the slowest 50 durations. Anything in the ~200s and above range was marked as a slow test in this PR.

## ✅ Checklist
<!-- Please check off all required items. For those that don't apply remove them accordingly. -->

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

<img width="477" height="429" alt="Screenshot 2026-01-27 at 10 46 09" src="https://github.com/user-attachments/assets/7f2dd341-0106-435e-ab92-6cad85184f0e" />

### DIFF
--- a/packages/testing/src/execution_testing/cli/gen_index.py
+++ b/packages/testing/src/execution_testing/cli/gen_index.py
@@ -257,7 +257,9 @@ def merge_partial_indexes(output_dir: Path, quiet_mode: bool = False) -> None:
         return
 
     # Merge all partial indexes (JSONL format: one entry per line)
-    all_entries: List[TestCaseIndexFile] = []
+    # Read as raw dicts — the data was already validated when collected
+    # from live Pydantic fixture objects in add_fixture().
+    all_raw_entries: list[dict] = []
     all_forks: set = set()
     all_formats: set = set()
 
@@ -268,27 +270,27 @@ def merge_partial_indexes(output_dir: Path, quiet_mode: bool = False) -> None:
                 if not line:
                     continue
                 entry_data = json.loads(line)
-                # Validate with Pydantic
-                # (deferred from add_fixture for performance)
-                entry = TestCaseIndexFile.model_validate(entry_data)
-                all_entries.append(entry)
-                # Collect forks and formats from validated entries
-                if entry.fork:
-                    all_forks.add(entry.fork.name())
-                if entry.format:
-                    all_formats.add(entry.format.format_name)
+                all_raw_entries.append(entry_data)
+                # Collect forks and formats from raw strings
+                if entry_data.get("fork"):
+                    all_forks.add(entry_data["fork"])
+                if entry_data.get("format"):
+                    all_formats.add(entry_data["format"])
 
-    # Compute root hash from collected fixture hashes
-    root_hash = HashableItem.from_index_entries(all_entries).hash()
+    # Compute root hash from raw dicts (no Pydantic needed)
+    root_hash = HashableItem.from_raw_entries(all_raw_entries).hash()
 
-    # Build final index
-    index = IndexFile(
-        test_cases=all_entries,
-        root_hash=HexNumber(root_hash),
-        created_at=datetime.datetime.now(),
-        test_count=len(all_entries),
-        forks=list(all_forks),
-        fixture_formats=list(all_formats),
+    # Build final index — Pydantic validates the entire structure once
+    # via model_validate(), not 96k individual model_validate() calls.
+    index = IndexFile.model_validate(
+        {
+            "test_cases": all_raw_entries,
+            "root_hash": HexNumber(root_hash),
+            "created_at": datetime.datetime.now(),
+            "test_count": len(all_raw_entries),
+            "forks": list(all_forks),
+            "fixture_formats": list(all_formats),
+        }
     )
 
     # Write final index
@@ -299,7 +301,7 @@ def merge_partial_indexes(output_dir: Path, quiet_mode: bool = False) -> None:
     if not quiet_mode:
         rich.print(
             f"[green]Merged {len(partial_files)} partial indexes "
-            f"({len(all_entries)} test cases) into {index_path}[/]"
+            f"({len(all_raw_entries)} test cases) into {index_path}[/]"
         )
 
     # Cleanup partial files
--- a/packages/testing/src/execution_testing/cli/gen_index.py
+++ b/packages/testing/src/execution_testing/cli/gen_index.py
@@ -226,5 +226,73 @@ def generate_fixtures_index(
         f.write(index.model_dump_json(exclude_none=False, indent=2))
 
 
+def merge_partial_indexes(output_dir: Path, quiet_mode: bool = False) -> None:
+    """
+    Merge partial index files from all workers into final index.json.
+
+    This is called by pytest_sessionfinish on the master process after all
+    workers have finished and written their partial indexes.
+
+    Args:
+        output_dir: The fixture output directory.
+        quiet_mode: If True, don't print status messages.
+
+    """
+    meta_dir = output_dir / ".meta"
+    partial_files = list(meta_dir.glob("partial_index*.json"))
+
+    if not partial_files:
+        # Fallback to old method if no partial indexes exist
+        if not quiet_mode:
+            rich.print(
+                "[yellow]No partial indexes found, "
+                "falling back to full scan[/]"
+            )
+        generate_fixtures_index(
+            output_dir, quiet_mode=quiet_mode, force_flag=True
+        )
+        return
+
+    # Merge all partial indexes
+    all_entries: List[TestCaseIndexFile] = []
+    all_forks: set = set()
+    all_formats: set = set()
+
+    for partial_file in partial_files:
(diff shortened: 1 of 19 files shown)


packages/testing/src/execution_testing/fixtures/collector.py
```diff
@@ -166,6 +179,24 @@ def add_fixture(self, info: TestInfo, fixture: BaseFixture) -> Path:
...
+                    "id": info.get_id(),
+                    "json_path": str(relative_path),
+                    "fixture_hash": str(fixture.hash)
+                    if fixture.hash
+                    else None,
+                    "fork": fixture_fork.name() if fixture_fork else None,
+                    "format": fixture.format_name,
+                    "pre_hash": getattr(fixture, "pre_hash", None),
+                }
+            )
```
marioevz:
```suggestion
            index_entry = {
                "id": info.get_id(),
                "json_path": str(relative_path),
                "fixture_hash": str(fixture.hash) if fixture.hash else None,
                "fork": fixture_fork.name() if fixture_fork else None,
                "format": fixture.format_name,
            }
            if (pre_hash := getattr(fixture, "pre_hash", None)) is not None:
                index_entry["pre_hash"] = pre_hash
            self.index_entries.append(index_entry)
```
Just so we can omit the pre_hash field from being added when unnecessary.

packages/testing/src/execution_testing/cli/gen_index.py
```diff
@@ -226,5 +226,88 @@ def generate_fixtures_index(
...
+        # Fallback to old method if no partial indexes exist
+        if not quiet_mode:
+            rich.print(
+                "[yellow]No partial indexes found, "
+                "falling back to full scan[/]"
+            )
+        generate_fixtures_index(
+            output_dir, quiet_mode=quiet_mode, force_flag=True
+        )
+        return
```
marioevz:
```suggestion
        raise Exception("No partial indexes found.")
```

I think this should be a direct error; if we are calling `merge_partial_indexes` it means that we should be sure the partial indexes should be there.

Perhaps in case no tests were filled, if this produces an error, we should skip calling `merge_partial_indexes`.

packages/testing/src/execution_testing/cli/gen_index.py
```diff
@@ -226,5 +226,88 @@ def generate_fixtures_index(
...
+            "test_count": len(all_raw_entries),
+            "forks": list(all_forks),
+            "fixture_formats": list(all_formats),
+        }
+    )
+
+    # Write final index
+    index_path = meta_dir / "index.json"
+    index_path.parent.mkdir(parents=True, exist_ok=True)
+    index_path.write_text(index.model_dump_json(exclude_none=False, indent=2))
```
marioevz:
```suggestion
    index_path.write_text(index.model_dump_json(exclude_none=True, indent=2))
```
Just so we can exclude `pre_hash` when most of the time is set to `None`.

We also would have to update `TestCaseBase` in `src/execution_testing/fixtures/consume.py` to set  the defaults to `None` for the ones that can be `None`, otherwise pydantic model validation will fail.

marioevz (commented):
Just a few comments but overall looks excellent, thanks!

fselmo:
Thanks @marioevz! I believe I added all your suggestions, if you can double check the last commit 👀

fselmo:
Still looking into some things after rebasing with #2080, will re-mark as ready when it's ready to go.

fselmo:
@marioevz ok this is ready for review again!

fselmo:
@marioevz or anyone reviewing... the last commit here essentially removes the smaller gains from two commits before it ([here](https://github.com/ethereum/execution-specs/pull/2079/changes/d4e3f5a3bc80edff32e5814839922f07a9e3a73f)) with pre-serialization (holding in memory).

I only did this to help our good friend, the `pypy3` CI run `::dies inside::`, as [this](https://github.com/ethereum/execution-specs/actions/runs/21451210455/job/61780369074?pr=2079) was happening again - which is a similar thing we ran into when attempting to turn on `--until=Amsterdam` recently. It seems the runners can't handle memory independently and workers start crashing and brick the run.

If we find another solution to help `pypy`, that commit took `80s` test teardowns to `60s`, which the following commit for partial fixture building across workers ([here](https://github.com/ethereum/execution-specs/pull/2079/changes/00a835cf3704d3f7cd529de21a76cd8b1111b7b6)) brought down to ~1s. Removing this small gain, these same tests are still at ~5s teardown so it's still a huge gain. It's quite unfortunate to have to do this just for `pypy` backend... but let's see if it even helps. I will wait until this CI runs and see what we can do here. Will put this back to `draft` for now.

----

Putting some `py3` CI performance comparisons here so we can compare cost / benefit:

- Before PR #2080 ---> [0:59:20](https://github.com/ethereum/execution-specs/actions/runs/21374446126/job/61527159243#step:5:5476)
- After PR #2080 was merged ---> ([0:54:15](https://github.com/ethereum/execution-specs/actions/runs/21410127576/job/61644855811#step:5:5476))
- This PR before this last commit that removed the pre-serialization ---> ([0:36:57](https://github.com/ethereum/execution-specs/actions/runs/21451210455/job/61780369075?pr=2079#step:5:2519))
- With the last commit (removes the small pre-serialization performance win) ---> [0:38:40](https://github.com/ethereum/execution-specs/actions/runs/21460114169/job/61810524445#step:5:2519)

fselmo:
> Putting some `py3` CI performance comparisons here so we can compare cost / benefit:
> 
>     * Before PR [feat(test-cli): add `hasher compare` subcommand #2080](https://github.com/ethereum/execution-specs/pull/2080) ---> [0:59:20](https://github.com/ethereum/execution-specs/actions/runs/21374446126/job/61527159243#step:5:5476)
> 
>     * After PR [feat(test-cli): add `hasher compare` subcommand #2080](https://github.com/ethereum/execution-specs/pull/2080) was merged ---> ([0:54:15](https://github.com/ethereum/execution-specs/actions/runs/21410127576/job/61644855811#step:5:5476))
> 
>     * This PR before this last commit that removed the pre-serialization ---> ([0:36:57](https://github.com/ethereum/execution-specs/actions/runs/21451210455/job/61780369075?pr=2079#step:5:2519))
> 
>     * With the last commit (removes the small pre-serialization performance win) ---> [0:38:40](https://github.com/ethereum/execution-specs/actions/runs/21460114169/job/61810524445#step:5:2519)

Seems like there's less than a 2min difference between dropping the pre-serialization and not and it seems to be better for `pypy3`. I feel this is a decent enough compromise to get this PR through but if we feel like keeping it since `pypy3` runs will improve, I'll leave this to the reviewer. This is ready for review again 😅.

marioevz (approved):
LGTM, testing this locally it really helps out to speed up the process!

PR #2080, opened 2026-01-27
feat(test-cli): add `hasher compare` subcommand

danceratopz:
## 🗒️ Description

This PR adds a `hasher compare` subcommand to diff two fixture directories and display hash differences with a clear hierarchical format.

The aim is to have a one-liner to diff and evaluate `fill` output that is useful for interactive use but that can later be used in an automated fixture diff flow.

- Add `compare` subcommand that exits 0 (identical), 1 (different), or 2 (error).
- Add `--depth N` flag to limit comparison depth.
- Display diff with full paths, hashes on separate lines, and visual hierarchy via indentation.
- Maintain backwards compatibility: `hasher FOLDER` still works.
- Share `-f`, `-t`, `-r` flags between `hash` and `compare` commands.
- Auto-detect CI/piped output for plain text (no rich output).

### Example Output

The happy case with no differences in fixtures produces no output; exit code 0.

Here is an example where all fixtures differ with `--depth=3`

image

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```
#### Cute Animal Picture
image

### DIFF
--- a/packages/testing/src/execution_testing/cli/hasher.py
+++ b/packages/testing/src/execution_testing/cli/hasher.py
@@ -170,80 +170,58 @@ def render_hash_report(
     )
 
 
-def parse_hash_lines(
-    lines: List[str], strip_root: bool = False
-) -> List[tuple[str, str]]:
-    """
-    Parse hash lines, reconstructing full paths from indentation hierarchy.
+def collect_hashes(
+    item: HashableItem,
+    *,
+    path: str = "",
+    print_type: Optional[HashableItemType] = None,
+    max_depth: Optional[int] = None,
+    depth: int = 0,
+) -> Dict[str, str]:
+    """Collect hashes from item tree as {path: hash_hex}."""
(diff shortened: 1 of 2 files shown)


packages/testing/src/execution_testing/cli/hasher.py
```diff
@@ -42,26 +45,43 @@ def hash(self) -> bytes:
             all_hash_bytes += item_hash_bytes
         return hashlib.sha256(all_hash_bytes).digest()
```
marioevz, line 46: Maybe a small win in terms of both fixture filling and this hasher command could be to replace this hashing function with something much faster; since we don't need a cryptographically secure hashing algo, we could replace it with something like [xxHash](https://xxhash.com/).

Perhaps for a future PR.

marioevz (approved):
Amazing, thanks!

PR #2082, opened 2026-01-27
fix(tests): EIP-7883 minor fixes

pdobacz:
## 🗒️ Description

Two unrelated fixes to the EIP-7883 testing:
- fixing the gas formula - I'm not sure why the `(len(modexp_expected) // 32)` factor was there, as the number of SSTOREs never depended on the length. At the same time there were several SSTOREs done throughout the test. Also the precompile gas part of the tx gas must take into account that the 63/64th rule is going to apply
- new gas should apply for forks coming after Osaka, so that forks that might not define its relation to Osaka use old gas

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/osaka/eip7883_modexp_gas_increase/conftest.py
+++ b/tests/osaka/eip7883_modexp_gas_increase/conftest.py
@@ -60,14 +60,18 @@ def total_tx_gas_needed(
         fork.transaction_intrinsic_cost_calculator()
     )
     memory_expansion_gas_calculator = fork.memory_expansion_gas_calculator()
-    sstore_gas = fork.gas_costs().G_STORAGE_SET * (len(modexp_expected) // 32)
+    sstore_gas = (
+        fork.gas_costs().G_STORAGE_SET + fork.gas_costs().G_COLD_SLOAD
+    ) * 5
+    # Ensures that the precompile call is not starved by the 63/64 rule.
+    precompile_gas_with_margin = precompile_gas * 64 // 63
     extra_gas = 100_000
 
     return (
         extra_gas
         + intrinsic_gas_cost_calculator(calldata=bytes(modexp_input))
         + memory_expansion_gas_calculator(new_bytes=len(bytes(modexp_input)))
-        + precompile_gas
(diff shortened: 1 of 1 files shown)


tests/osaka/eip7883_modexp_gas_increase/conftest.py
```diff
@@ -60,14 +60,18 @@ def total_tx_gas_needed(
         fork.transaction_intrinsic_cost_calculator()
     )
     memory_expansion_gas_calculator = fork.memory_expansion_gas_calculator()
-    sstore_gas = fork.gas_costs().G_STORAGE_SET * (len(modexp_expected) // 32)
+    sstore_gas = (
+        fork.gas_costs().G_STORAGE_SET + fork.gas_costs().G_COLD_SLOAD
+    ) * 5
```
fselmo: This should be 4 no? `gas_measure_contract` performs 4 cold SSTOREs in the success case: [1 and 2](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/tests/osaka/eip7883_modexp_gas_increase/conftest.py#L182-L191) then [3](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/tests/osaka/eip7883_modexp_gas_increase/conftest.py#L195-L198) and [4](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/tests/osaka/eip7883_modexp_gas_increase/conftest.py#L202-L207).

```suggestion
    ) * 4
```

fselmo (commented):
This lgtm, really nice find 👍🏼. Just one clarifying question for a bit more accuracy here before approving.

tests/osaka/eip7883_modexp_gas_increase/conftest.py
fselmo: I think we should add a succinct clarifying comment here for a reader to understand where this `4` is coming from quicker

fselmo (approved):
Thanks!

PR #2085, opened 2026-01-27
feat(src): add `getBlobsV3` support and test case for it

felix314159:
## 🗒️ Description
1. Fix: deterministic factory is now pre-deployed in the genesis, so u can do `execute hive` again because there is no need to allow unprotecte transactions anymore
2. Another fix that adds BAL support to `execute hive`
3. QoL fix that does not let blobs spam terminal with RLP data in case of failure (upper cap on printed chars)
4. Fixes flaw in blobs test where it tried to put 7 blobs for amsterdam when upper cap is 6
5. Framework improvements for testing different getBlobsVX behaviors, we can now explicitely tests getBlobsV1, V2 and V3
6. Adds warning when user tries to `fill` a blobs test (would just fill 0 without telling why). it is non-obvious that you are supposed to use `execute hive` for blobs tests, the printed warning clearly states that now
7. Improves logic flow of blobs execute function
8. Adds new tests (getblobsv3 behavior, validity of those blobs, and also edgecase for v3 where actually all blobs were indeed non-existing)

---

How to test:

1. Use this `clients.yaml` file for hive:

```
- client: go-ethereum
  nametag: default
  dockerfile: git
  build_args:
    github: ethereum/go-ethereum
    tag: bal-devnet-2
```

2. Ensure you have latest hive master and run:
`go run . --docker.nocache=go-ethereum --docker.buildoutput --docker.output --client=go-ethereum --client-file=clients.yaml --dev --loglevel=5 --sim.loglevel=5`

3. Cherry-pick all of these 5 commits from our SLOTNUM (7843) branch:
* `git cherry-pick 95794929f`
* `git cherry-pick fa411a3d2`
* `git cherry-pick ca529e272`
* `git cherry-pick 61c548aca`
* `git cherry-pick dba41c003`

4. Run:
`uv run execute hive -v -s ./tests/osaka/eip7594_peerdas/test_get_blobs.py --fork=amsterdam`

-> all 24 tests pass

## 🔗 Related Issues or PRs
[engine_getBlobsV3 PR](https://github.com/ethereum/execution-apis/pull/719)

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -1855,6 +1855,9 @@ def pytest_collection_modifyitems(
     These can't be handled in this plugins pytest_generate_tests() as the fork
     parametrization occurs in the forks plugin.
     """
+    # Track specs with no fixture formats for warning message
+    specs_without_fixture_formats: Dict[str, Set[str]] = {}
+
     items_for_removal = []
     for i, item in enumerate(items):
         item.name = item.name.strip().replace(" ", "-")
@@ -1870,6 +1873,15 @@ def pytest_collection_modifyitems(
         spec_type, fixture_format = get_spec_format_for_item(params)
         if isinstance(fixture_format, NotSetType):
             items_for_removal.append(i)
+            # Track this spec type and the test file
+            spec_name = spec_type.__name__
+            if spec_name not in specs_without_fixture_formats:
+                specs_without_fixture_formats[spec_name] = set()
+            # Get the test file path from the item
+            test_file = (
+                str(item.fspath) if hasattr(item, "fspath") else str(item.path)
+            )
+            specs_without_fixture_formats[spec_name].add(test_file)
             continue
         assert issubclass(fixture_format, BaseFixture)
         if not fixture_format.supports_fork(fork):
@@ -1912,6 +1924,39 @@ def pytest_collection_modifyitems(
 
     for i in reversed(items_for_removal):
         items.pop(i)
+
+    # Warn users if tests were removed because they have no fixture formats
+    if specs_without_fixture_formats:
+        reporter = config.pluginmanager.get_plugin("terminalreporter")
+        if reporter and isinstance(reporter, TerminalReporter):
+            reporter.write_line("")
+            reporter.write_sep(
+                "=",
+                "NOTICE: Tests skipped (no fixture formats)",
+                yellow=True,
+                bold=True,
+            )
+            for spec_name, test_files in specs_without_fixture_formats.items():
+                reporter.write_line(
+                    f"  {spec_name} has no supported_fixture_formats "
+                    f"for 'fill' command.",
+                    yellow=True,
+                )
+                for test_file in sorted(test_files):
+                    reporter.write_line(f"    - {test_file}", yellow=True)
+            reporter.write_line("")
+            reporter.write_line(
+                "  These tests are designed for the 'execute' command, "
+                "not 'fill'.",
+                yellow=True,
+            )
+            reporter.write_line(
+                "  Use 'uv run execute hive' or 'uv run execute remote' "
+                "instead.",
+                yellow=True,
+            )
+            reporter.write_sep("=", yellow=True, bold=True)
+            reporter.write_line("")
 
     # Build base_nodeid cache and identify slow groups.
     # If ANY fixture format variant is marked slow, treat ALL variants as slow
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -1932,7 +1932,7 @@ def pytest_collection_modifyitems(
             reporter.write_line("")
             reporter.write_sep(
                 "=",
-                "NOTICE: Tests skipped (no fixture formats)",
+                "NOTICE: Tests ignored (as expected; no fixture formats)",
                 yellow=True,
                 bold=True,
             )
--- a/packages/testing/src/execution_testing/execution/blob_transaction.py
+++ b/packages/testing/src/execution_testing/execution/blob_transaction.py
@@ -31,6 +31,81 @@ from .base import BaseExecute, ExecuteResult
 logger = get_logger(__name__)
 
 
+def _validate_blob_and_proof(
+    expected_blob: BlobAndProofV1 | BlobAndProofV2 | None,
+    received_blob: BlobAndProofV1 | BlobAndProofV2 | None,
+    index: int,
+) -> None:
+    """
+    Validate that a received blob and proof match the expected values.
+
+    When expected is None (non-existing blob hash), the received blob must
+    also be None. When expected is not None, the received blob must match.
+    Raise ValueError with a detailed message on mismatch.
+    """
+    if expected_blob is None:
+        if received_blob is None:
+            logger.info(
+                f"Blob at index {index} correctly returned null "
+                "(non-existing blob hash)"
+            )
+            return
+        raise ValueError(
+            f"Blob at index {index} should not exist but "
+            f"client returned: {received_blob}"
+        )
+    if received_blob is None:
+        raise ValueError(f"Received blob at index {index} is empty.")
+    if isinstance(expected_blob, BlobAndProofV1):
+        if not isinstance(received_blob, BlobAndProofV1):
+            raise ValueError(
+                f"Received blob at index {index} is not a BlobAndProofV1."
+            )
+        if expected_blob.blob != received_blob.blob:
+            raise ValueError(f"Blob mismatch at index {index}.")
+        if expected_blob.proof != received_blob.proof:
+            raise ValueError(f"Proof mismatch at index {index}.")
+    elif isinstance(expected_blob, BlobAndProofV2):
+        if not isinstance(received_blob, BlobAndProofV2):
+            raise ValueError(
+                f"Received blob at index {index} is not a BlobAndProofV2."
+            )
+        if expected_blob.blob != received_blob.blob:
+            raise ValueError(f"Blob mismatch at index {index}.")
+        if expected_blob.proofs != received_blob.proofs:
+            error_message = f"Proofs mismatch at index {index}."
+            expected_len = len(expected_blob.proofs)
+            received_len = len(received_blob.proofs)
+            error_message += f"len(expected_blob.proofs) = {expected_len}, "
+            error_message += f"len(received_blob.proofs) = {received_len}\n"
+            if expected_len == received_len:
+                for j, (expected_proof, received_proof) in enumerate(
+                    zip(
+                        expected_blob.proofs,
+                        received_blob.proofs,
+                        strict=False,
+                    )
+                ):
+                    if len(expected_proof) != len(received_proof):
+                        exp_len = len(expected_proof)
+                        rcv_len = len(received_proof)
+                        error_message += f"Proof length mismatch. index = {j},"
+                        error_message += f"expected_proof length = {exp_len}, "
+                        error_message += f"received_proof length = {rcv_len}\n"
+                        continue
+                    if expected_proof != received_proof:
+                        exp_hash = sha256(expected_proof).hexdigest()
+                        rcv_hash = sha256(received_proof).hexdigest()
+                        error_message += f"Proof mismatch. index = {j},"
+                        error_message += f"expected_proof hash = {exp_hash}, "
+                        error_message += f"received_proof hash = {rcv_hash}\n"
+            raise ValueError(error_message)
+    else:
+        raise ValueError(
+            f"Unexpected blob type at index {index}: {type(expected_blob)}"
+        )
+
+
 def versioned_hashes_with_blobs_and_proofs(
     tx: NetworkWrappedTransaction,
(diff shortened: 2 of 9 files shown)


felix314159:
After the [docs-fixes PR](https://github.com/ethereum/execution-specs/pull/2137) is merged I will rebase this getBlobsV3 PR again.

Edit: Ready for review now

felix314159:
Ready for review/merge

felix314159:
Still ready for review..

felix314159:
Ready for review

packages/testing/src/execution_testing/execution/blob_transaction.py
```diff
@@ -157,19 +163,66 @@ def execute(
...
+                        f"{blob_response.root}"
+                    )
+                else:
+                    logger.info(
+                        f"Test passed: getBlobsV{version} correctly returned "
+                        "'null' (partial responses not allowed in V1/V2)"
+                    )
+                    eth_rpc.wait_for_transactions(sent_txs)
+                    return
             else:
```
danceratopz: Wdyt about being more defensive here `version==3`? I.e., only validate V3 here and add an `else` block afterwards that raises an `UnexpectedGetBlobsVersionException` (for unexpected versions)?

This ensures we don't silently fail to validate V4 semantics should the protocol change in the future.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -1684,6 +1694,41 @@ def pytest_collection_modifyitems(
     for i in reversed(items_for_removal):
         items.pop(i)
 
+    # Warn users if tests were removed because they have no fixture formats
```
danceratopz, line 1928: This is really nice output, I have to say. But is too dangerous to silently remove them? I mean because that's the expected behavior. I'm happy either way, tbh. Perhaps we can say explicitly that it's expected; "nothing to see here"?

Image

Command:
```
uv run fill -v -s ./tests/osaka/eip7594_peerdas/ --fork=amsterdam --clean
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -1684,6 +1694,41 @@ def pytest_collection_modifyitems(
...
+    if specs_without_fixture_formats:
+        from _pytest.terminal import TerminalReporter
+
+        reporter = config.pluginmanager.get_plugin("terminalreporter")
+        if reporter and isinstance(reporter, TerminalReporter):
+            reporter.write_line("")
+            reporter.write_sep(
+                "=",
+                "NOTICE: Tests skipped (no fixture formats)",
```
danceratopz: Also technically, they're not skipped, they're initially collected and then removed from the test session.
```suggestion
                "NOTICE: Tests ignored (as expected; no fixture formats)",
```

packages/testing/src/execution_testing/execution/blob_transaction.py
```diff
@@ -157,19 +163,66 @@ def execute(
...
+                            )
                 logger.info(
-                    "Test was passed (partial responses are not allowed and "
-                    "the client correctly returned 'null')"
+                    f"Test passed: getBlobsV{version} correctly returned "
+                    f"partial response with {existing_count} existing blobs "
+                    f"and {nonexisting_count} null entries for missing blobs"
                 )
                 eth_rpc.wait_for_transactions(sent_txs)
                 return
```
danceratopz: Doesn't this mean that, in the case that we have some `nonexisting_blob_hashes` that we always return early and never perform the full byte comparison of blob data + proofs for the none-null blobs below? I.e., we skip the checks starting in L230.

I think we should to trim the null entries, fall-through and validate the non-null entries?

tests/osaka/eip7594_peerdas/test_get_blobs.py
```diff
@@ -352,18 +373,74 @@ def test_get_blobs(
 )
 @pytest.mark.exception_test
 @pytest.mark.valid_from("Cancun")
-def test_get_blobs_nonexisting(
+@pytest.mark.valid_until("Prague")
```
danceratopz, line 376: Should we be checking rejection behavior for V1 with Osaka, Amsterdam?

tests/osaka/eip7594_peerdas/test_get_blobs.py
danceratopz, line 367: When `blobs=` is omitted, it falls back to the default fork `getBlob` version. We need to add `engine_get_blobs_version` method to the `Amsterdam` fork class to get `getBlobsV3` behavior:
https://github.com/felix314159/execution-specs/blob/4a7a2c22ca5b32e3945dfbc46c6df8cedd6e4d1f/packages/testing/src/execution_testing/forks/forks/forks.py#L3023-L3029

danceratopz (commented):
Nice work! These are really nice tests and it'll be good to get them integrated in hive again. Few comments below.

Otherwise, do you think it'd be overkill to get a system test in with this PR? Thanks to the existing `.github/workflows/hive-consume.yaml`, I think it should be relatively simple to add it (the workflow/file should probably be renamed to hive-simulators.yaml to reflect this addition). 

See: https://github.com/felix314159/execution-specs/blob/4a7a2c22ca5b32e3945dfbc46c6df8cedd6e4d1f/.github/workflows/hive-consume.yaml#L63-L82

felix314159:
Thanks for the feedback, the commits above are just for rebasing + ensuring everything still works. Next I will look at your review, add requested changes and then push another commit

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
felix314159: I would prefer to keep this behavior, the print makes it very clear what is skipped and why.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
felix314159: Ok, I will modify this print

packages/testing/src/execution_testing/execution/blob_transaction.py
felix314159: The early returns are specifically for only the `nonexisting_blob_hashes` tests (where we request a combination of blobs clients have and blobs clients do not have), in those we only care about whether the response is `null` for all blobs or only for all missing blobs. We don't care about the blobs themselves or their validity here so it is intended behavior to skip the other checks.

tests/osaka/eip7594_peerdas/test_get_blobs.py
felix314159: We also run an equivalent of this test for Osaka and Amsterdam. The difference between V1 and V2 is the proof format (one proof vs cell proofs), and we run these tests for v2 against osaka and amsterdam, so there is no need to try to have v1 rejected (checking whether the old proof format is accepted in osaka or later is a different test altogether that i am pretty sure we already have, but this PR is about `getBlobs` specificially)

tests/osaka/eip7594_peerdas/test_get_blobs.py
felix314159: No, because:
* v3 is available already in osaka
* v3 is OPTIONAL, you MAY use v3 or just use v2

[Source](https://github.com/ethereum/execution-apis/blob/main/src/engine/osaka.md#engine_getblobsv3)

What we could do i rename our function to make it more clear that this is the required minimum (v2) for osaka and amsterdam, but in the docstring explain that v3 behavior MAY be used instead. But this would be a cosmetic change that we do not really need IMO

felix314159:
> Nice work! These are really nice tests and it'll be good to get them integrated in hive again. Few comments below.
> 
> Otherwise, do you think it'd be overkill to get a system test in with this PR? Thanks to the existing `.github/workflows/hive-consume.yaml`, I think it should be relatively simple to add it (the workflow/file should probably be renamed to hive-simulators.yaml to reflect this addition).
> 
> See: https://github.com/felix314159/execution-specs/blob/4a7a2c22ca5b32e3945dfbc46c6df8cedd6e4d1f/.github/workflows/hive-consume.yaml#L63-L82

We could do it, be it should be a seperate PR IMO. Thanks for all the feedback, let's see if CI passes and we can merge

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
danceratopz: Thanks.

tests/osaka/eip7594_peerdas/test_get_blobs.py
danceratopz: Should we parametrize for >= Osaka to test both variants v2 and v3?

felix314159: We already do that. The tests are valid from osaka and the test for v2 is seperate from the v3 test

danceratopz: I think we're talking past each other again. Yes, I agree that:
- V3 is available starting at Osaka.                                                                                               
- V3 is optional; clients MAY support it alongside V2 (however, all mainnet client releases do).

And that for the non-existing blob case:
- There are separate V2 and V3 nonexisting-blob tests (`test_get_blobs_nonexisting_getblobsv2` and `test_get_blobs_nonexisting_getblobsv3).  

On the other hand, I'm talking about the happy case. 

The test above in question tests a block with available blobs (and no non-existing blobs). As far I can tell, this is only one test for that: `test_get_blobs()` ([in your branch](https://github.com/felix314159/execution-specs/blob/getblobsv3-added/tests/osaka/eip7594_peerdas/test_get_blobs.py#L358)). This is parametrized by fork, not by `getBlobs` version.

As this test relies on the fork method `engine_get_blobs_version` (because it is not explicitly set), we're only testing with v2 and v3 is untested in the happy case.

tests/osaka/eip7594_peerdas/test_get_blobs.py
danceratopz: I think we're talking past each other.

I'm asking whether we have a test for post-Prague forks that tests the successful rejection of `getBlobsV1` calls. Perhaps this is overkill, but this behavior is [explicitly stated in the spec](https://github.com/ethereum/execution-apis/blob/a9e9095c453d7b3c95b89abfcf46afb2eff4e1b7/src/engine/osaka.md#cancun-api) and it's still a valid test (for `getBlobs`).

Follow-up issue?

packages/testing/src/execution_testing/execution/blob_transaction.py
danceratopz: V3 is the first version where the nonexisting test gets actual blob data back alongside nulls, and as [discussed in the comment below](https://github.com/ethereum/execution-specs/pull/2085#discussion_r2870940641), the happy-path test doesn't cover V3. 

This means we don't have any checks/tests that validate blob data from V3 calls.

danceratopz:
Thanks for implementing the suggestions @felix314159! However, I think we talked passed each other in other comments. I've further clarified what I meant in the original comment and replied in the relevant threads above.

packages/testing/src/execution_testing/execution/blob_transaction.py
felix314159: > We don't care about the blobs themselves or their validity here so it is intended behavior to skip the other checks.

sorry i was wrong about this. like u say it is important that we actually check blob validity of blobs received via v3. i added that change with my most recent push. thanks!

tests/osaka/eip7594_peerdas/test_get_blobs.py
felix314159: yes thanks, i get it now. i added what you want in my latest push

tests/osaka/eip7594_peerdas/test_get_blobs.py
```diff
@@ -352,18 +373,119 @@ def test_get_blobs(
...
+    txs_versioned_hashes: List[List[bytes]],
+) -> None:
+    """
+    Test that ensures clients respond with 'null' when at least one requested
+    blob is not available (getBlobsV2 behavior: all-or-nothing response).
+    """
+    nonexisting_blob_hashes = [
+        Hash(sha256(str(i).encode()).digest()) for i in range(5)
+    ]
+    print("Testing getBlobsV2 (all-or-nothing behavior)")
```
danceratopz, line 416:
```suggestion
    logger.debug("Testing getBlobsV2 (all-or-nothing behavior)")
```

danceratopz (commented):
Thanks for addressing the comments, @felix314159! Just one minor nit in `test_get_blobs.py`: `generate_valid_blob_tests()` uses `logger.debug()` whereas your recent commits use `print()`. Could you use `logger.debug()` or `logger.info()` as you see fit?

tests/osaka/eip7594_peerdas/test_get_blobs.py
felix314159: I tried logger.info() but it did not show up, so then I settled for a print (which can easily be hidden by removing `-s`) (i think i did try both `--eest-loglevel=INFO` and `--loglevel=INFO`)

danceratopz: Ok, let's get this merged and address this as a follow-up!

danceratopz (approved):
LGTM, thanks for getting these tests up-to-date for Osaka/Amsterdam!

I didn't manage to test this locally, but let's get devnets/bal/3 rebased on this and try there.

Issue #1875, opened 2025-12-09
EIP-7708 Implementation Tracker: ETH transfers emit a log


marioevz:
## [EIP-7708: ETH transfers emit a log](https://eips.ethereum.org/EIPS/eip-7708)

**Target Fork:** _Amsterdam_

**Complexity Assessment:** _[EIP-7708 Complexity Assessment](https://github.com/ethsteel/pm/blob/main/complexity_assessments/EIPs/EIP-7708.md)_

**Related PRs and Issues**
- #2062 
- #2023 
- #2059 
- https://github.com/Carsons-Eels/execution-specs/pull/2
- #2088 

### Instructions

- [x] Assign issue to EIP specification and testing owner(s).

> [!IMPORTANT]
> A specifications specialist and a testing specialist should ideally share ownership of the EIP.

- [x] Add the issue to the target fork milestone if applicable (i.e., the EIP is at least in the [CFI stage](https://eips.ethereum.org/EIPS/eip-7723#considered-for-inclusion)).

#### Guidance for Marking Items Complete

An item should only be checked off once the EIP is considered *stable*. In this context, stable means:

- No major issues or ambiguities are still being uncovered in the specification or tests.
- There are no open discussion points awaiting resolution.
- Client implementations have been consistently passing the tests for at least a week.

It is ultimately up to the owners' discretion to decide when an item should be marked as complete, using this guidance as the basis for that decision.

In exceptional cases, an EIP may require changes after some items have been marked complete or even after the entire issue has been completed and closed. This can happen, for example, when significant design optimizations are identified and agreed upon in ACD, or when critical security issues surface and require updates to the specification or tests.

When this occurs, owners should either unmark the relevant checkboxes if the issue is still open, or create a new tracking issue for the modifications if the original issue had already been closed.

### Specification + Testing Status

- [x] Testing complexity assessed and documented: [EIP-7708.md](https://github.com/ethsteel/pm/blob/main/complexity_assessments/EIPs/EIP-7708.md) - 🟢 9.
- [x] Specification implementation merged to `eips/amsterdam/eip-7708` *(skip if the fork branch merge below is already complete)*.
- [x] Specification updates merged to the corresponding `forks/amsterdam` branch.
- [x] EIP updates proposed in case of architectural choices surfaced during implementation.
- [x] Required testing framework modifications implemented -> https://github.com/Carsons-Eels/execution-specs/pull/2.
[...]

PR #2086, opened 2026-01-27
feat(spec,test): EIP-7708 spec updates for self as target

### LINKED ISSUE #2088: Small update to 7708 EIP requires minor spec update
### Metadata

* **Hardfork**: Amsterdam

### What was wrong?

Small spec update to exclude logs for transfers of value from an account to itself.

### Sources

[PR #11188](https://github.com/ethereum/EIPs/pull/11188)

### Additional Context

Should be a one-line implementation

fselmo:
### SELFDESTRUCT

- Tracking the discussion [here](https://discord.com/channels/595666850260713488/688075293562503241/1465722727481475174) that selfdestruct logs may only be emitted if created and destroyed in same tx ([EIP-6780](https://eips.ethereum.org/EIPS/eip-6780))
    - EIP update: https://github.com/ethereum/EIPs/pull/11190

### CALL and Transactions

- Exclude logs for transfer of value to self:
  - CALL EIP update: https://github.com/ethereum/EIPs/pull/11187
  - Transaction EIP update: https://github.com/ethereum/EIPs/pull/11188

Also added some more test cases for coverage emitting transfer logs when not in the same tx, self destruct logs when in same tx, combo of transfer and selfdestruct logs across fork transition, update tests to reflect above changes

closes #2088 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2026-01-28 at 11 06 45

### DIFF
--- a/tests/amsterdam/eip7708_eth_transfer_logs/test_selfdestruct_logs.py
+++ b/tests/amsterdam/eip7708_eth_transfer_logs/test_selfdestruct_logs.py
@@ -3,21 +3,27 @@ Tests for EIP-7708 Selfdestruct logs.
 
 Tests for the Selfdestruct(address,uint256) log emitted when:
 - SELFDESTRUCT to self with nonzero balance
-- Account closure after SELFDESTRUCT
+- Account created and destroyed in the same transaction
 """
 
 import pytest
 from execution_testing import (
     EOA,
+    Account,
     Alloc,
     Environment,
+    Initcode,
     Op,
     StateTestFiller,
     Transaction,
     TransactionReceipt,
+    compute_create_address,
+)
+from execution_testing import (
+    Macros as Om,
 )
 
-from .spec import ref_spec_7708, selfdestruct_log
+from .spec import ref_spec_7708, selfdestruct_log, transfer_log
 
 REFERENCE_SPEC_GIT_PATH = ref_spec_7708.git_path
 REFERENCE_SPEC_VERSION = ref_spec_7708.version
@@ -25,17 +31,18 @@ REFERENCE_SPEC_VERSION = ref_spec_7708.version
 pytestmark = pytest.mark.valid_from("Amsterdam")
 
 
-def test_selfdestruct_to_self_emits_log(
+def test_selfdestruct_to_self_pre_existing_no_log(
     state_test: StateTestFiller,
     env: Environment,
     pre: Alloc,
     sender: EOA,
 ) -> None:
     """
-    Test that selfdestruct-to-self emits a Selfdestruct log.
+    Test that selfdestruct-to-self emits NO log for pre-existing contracts.
 
-    Since the contract selfdestructs to itself, there is no transfer.
-    Instead, a Selfdestruct log is emitted with the contract's balance.
+    Per EIP-7708, SELFDESTRUCT to the same account does not emit a Transfer
+    log. The Selfdestruct log is only emitted when the account is created
+    and destroyed in the same transaction (actual ETH burn).
     """
     contract_balance = 2000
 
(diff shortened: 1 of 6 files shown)


tests/amsterdam/eip7708_eth_transfer_logs/test_selfdestruct_logs.py
```diff
@@ -47,9 +54,267 @@ def test_selfdestruct_to_self_emits_log(
...
+    contract_balance = 2000
+
+    initcode = Op.SELFDESTRUCT(Op.ADDRESS)
+    initcode_bytes = bytes(initcode)
+    initcode_len = len(initcode_bytes)
+
+    factory_code = Op.MSTORE(
+        0, Op.PUSH32(initcode_bytes.rjust(32, b"\x00"))
+    ) + Op.CREATE(
+        value=contract_balance, offset=32 - initcode_len, size=initcode_len
```
jochem-brouwer: Suggestion to collapse this test and the next test into one test: use `value=Op.CALLVALUE` and then send the value with the tx. You can now have this test and the next test (0 balance does not emit a log) in a single test 😄

tests/amsterdam/eip7708_eth_transfer_logs/test_selfdestruct_logs.py
```diff
@@ -47,9 +54,267 @@ def test_selfdestruct_to_self_emits_log(
...
+    c1_code = Op.SELFDESTRUCT(Op.CALLDATALOAD(0))
+    c2_code = Op.SELFDESTRUCT(Op.CALLDATALOAD(0))
+    c1 = pre.deploy_contract(c1_code, balance=100)
+    c2 = pre.deploy_contract(c2_code, balance=200)
+
+    # Factory: CREATE A, CREATE B, CALL A, CALL B, then CALL C1/C2 to send ETH
+    factory_code = (
+        Om.MSTORE(initcode, 0)
+        # CREATE A (nonce 1) and B (nonce 2)
+        + Op.SSTORE(0, Op.CREATE(value=1000, offset=0, size=initcode_len))
```
jochem-brouwer: Nit: you are using storage (expensive) here as an alternative to memory. For this purpose use the transient storage, so `Op.TSTORE` and `Op.TLOAD`. See also https://eips.ethereum.org/EIPS/eip-1153

tests/amsterdam/eip7708_eth_transfer_logs/test_selfdestruct_logs.py
```diff
@@ -47,9 +54,267 @@ def test_selfdestruct_to_self_emits_log(
...
+        transfer_log(factory, addr_a, 1000),
+        transfer_log(factory, addr_b, 2000),
+        transfer_log(addr_a, beneficiary, 1000),
+        transfer_log(addr_b, beneficiary, 2000),
+        transfer_log(c1, addr_a, 100),
+        transfer_log(c2, addr_b, 200),
+    ]
+    # Finalization logs in sorted address order
+    finalization_logs = [
+        selfdestruct_log(addr, amounts[addr]) for addr in sorted_addrs
```
jochem-brouwer: This is incorrect, we want to test here the SELFDESTRUCT logs which happen after executing, when accounts are deleted which have balance. In your current setup you create the selfdestruct contract, and then call into it, which invokes the selfdestruct opcode and then burns all eth. At the end of the transaction, we delete the accounts on the selfdestruct list and see they have no balance, so we emit no logs.
Here's how you could do it: if CALLDATASIZE is 0, do nothing, if CALLDATASIZE is not, then selfdestruct.
`Conditional(condition=Op.CALLDATASIZE, if_false=Op.STOP) + Op.SELFDESTRUCT(Op.CALLDATALOAD(0))`
Then in the test you could parametrize:
1. The amount sent via initcode (this is the amount which gets burned when we hit selfdestruct)
2. The amount we send after the selfdestruct

So you would first deploy the contract (with or without value), then call into it with an address (this selfdestructs and burns the eth) and then finally call it again (with no arguments) with value. This is what will get burned at the end of the tx.

Also, notice that you can run SELFDESTRUCT multiple times, even on already selfdestructed contracts. You could also send ETH to already-marked-for-selfdestruct contracts via SELFDESTRUCT of another contract (can think of many fancy constructions here)

tests/amsterdam/eip7708_eth_transfer_logs/test_transfer_logs.py
```diff
@@ -1203,6 +1246,77 @@ def test_selfdestruct_then_transfer_same_block(
...
+    Post: contract keeps balance (not deleted, not in created_accounts in Tx2)
+    """
+    contract_balance = 2000
+    sender = pre.fund_eoa()
+
+    runtime_code = Op.SELFDESTRUCT(Op.ADDRESS)
+    initcode = Initcode(deploy_code=runtime_code)
+    initcode_len = len(initcode)
+
+    factory_code = Om.MSTORE(initcode, 0) + Op.CREATE(
```
jochem-brouwer: Nit: you can also directly go to "initcode mode" by setting `to=None` in the transaction (and calculating the address created for the second tx)

jochem-brouwer (commented):
Checked tests, they look good but left a few nit comments and commented on one test where I think it is not tested what is intended.

Here are some extra cases to add:

- Ensure CALLCODE does not emit any logs
- Ensure SELFDESTRUCT to COINBASE emits the correct logs
This specific case will: 1. Invoke selfdestruct (address is coinbase), this could burn eth if coinbase has funds already (the contract created is thus sending to Op.ADDRESS, but the block.coinbase is set such that this matches the created address). 2. Pay the priority fees to coinbase and 3. Destroy coinbase.
- Ensure that in all 3 create situations (tx create, CREATE, CREATE2) the logs are emitted correctly
- Invoking selfdestruct in the same contract multiple times (parametrize sending to self (burn), other addresses, and then burn again). This should either burn or emit transfer logs.
- Send via selfdestruct funds to a contract already marked for selfdestruct. 
- Ensure that a transaction which 7702-delegates an account, which calls that account, which itself does SELFDESTRUCT, does not do the burn logs (as it does not burn since contract is not created)
- Ensure that if any action done in current tx, which reverts, that the logs are also deleted

Let me know if I can assist 😄 👍

tests/amsterdam/eip7708_eth_transfer_logs/test_selfdestruct_logs.py
fselmo: > which invokes the selfdestruct opcode and then burns all eth. 

There is no "burn ETH" here. This could probably be simplified but the selfdestructs triggered are not to self here. They are to beneficiary (`runtime = Op.SELFDESTRUCT(beneficiary))`. 

They selfdestruct to another address (transfer logs) -> are marked for deletion -> then receive a balance (transfer logs). At finalization they should emit selfdestruct logs (LOG2) in lexicographic order and this is what we want to check here.

I believe this test is actually correct. In fact, it fills. Am I missing something? 👀

I will address these comments tomorrow but wanted to make sure I'm not missing your point here.

fselmo:
> Here are some extra cases to add:

This is quite helpful, thank you 👍🏼. Imo, we should track these cases in the tracker for this EIP [here](https://github.com/ethereum/execution-specs/issues/1875). Or feel free to push here if you have permissions or PR to this branch. My goal here is to update the spec and provide enough tests to get a new release out and get clients to agree. Anything beyond this I think we can make subsequent PRs with more test cases. Wdyt?

>  Let me know if I can assist 😄 👍

The more the merrier 🙂! Thank you for the review! I will look into updating this tomorrow. Much appreciated 🙏🏼

tests/amsterdam/eip7708_eth_transfer_logs/test_selfdestruct_logs.py
```diff
@@ -47,9 +54,267 @@ def test_selfdestruct_to_self_emits_log(
...
+    """
+    beneficiary = pre.deploy_contract(Op.STOP)
+
+    runtime = Op.SELFDESTRUCT(beneficiary)
+    initcode = Initcode(deploy_code=runtime)
+    initcode_len = len(initcode)
+
+    # C1 and C2 selfdestruct to addresses in calldata
+    c1_code = Op.SELFDESTRUCT(Op.CALLDATALOAD(0))
+    c2_code = Op.SELFDESTRUCT(Op.CALLDATALOAD(0))
```
marioevz: We could create a couple more variants for these two contracts:
- Use `CALL` to pass value.
- Use `CALL` to pass value but then revert.

tests/amsterdam/eip7708_eth_transfer_logs/test_transfer_logs.py
```diff
@@ -776,6 +798,27 @@ def test_zero_value_operations_no_log(
...
 
 
+def test_call_to_self_no_log(
+    state_test: StateTestFiller,
+    env: Environment,
+    pre: Alloc,
+    sender: EOA,
+) -> None:
+    """Test that CALL with value to self emits no transfer log."""
+    contract_code = Op.CALL(gas=100_000, address=Op.ADDRESS, value=1)
```
marioevz: This is causing a recursive execution which eventually runs out of gas due to the 63/64 rule.

It seems to be working though, perhaps not as expected?

We could use `TSTORE` to stop the recursion at some level.

marioevz: Another comment is that we should also test `CALLCODE`.

marioevz (commented):
Some comments, thanks!

marioevz (approved):
Thanks for these tests! And many thanks to @jochem-brouwer for your review!

I think we can go ahead and merge as-is, and add any remaining tests to the tracker #1875, in order to (a) be able to release soon, and (b) keep this PR from growing out of proportion.

fselmo:
> add any remaining tests to the tracker

done https://github.com/ethereum/execution-specs/issues/1875#issuecomment-3820864077 👍🏼

PR #2099, opened 2026-01-29
feat(ci,tooling): improve static check developer experience

### LINKED ISSUE #2041: Refactor `static` into separate tox environments & dedicated `fast-checks` workflow
Split the monolithic `static` tox environment into granular, single-purpose environments and move them to a dedicated `fast-checks` workflow. This makes CI failures immediately obvious (the failing job name tells you what broke) and ensures these checks always run as a prerequisite for other workflows.

## Motivation

The current `static` tox environment bundles multiple unrelated checks into a single job:

- `codespell` (spelling).
- `ruff check` / `ruff format` (linting/formatting).
- `mypy` (type checking).
- `ethereum-spec-lint` (custom linting).
- `uv lock --check` (dependency lock validation).
- `actionlint` (GitHub Actions linting).

Splitting the static checks will:

1. **Reduce debugging friction.** When `static` fails, developers must dig through logs to find which check failed and why. Separate environments provide immediate visibility - the job name itself tells you what broke.
2. **Enable always-on checks.** With `paths-ignore` being introduced (#2038), doc-only PRs may skip the `test.yaml` workflow entirely. A dedicated `fast-checks` workflow ensures checks like spellcheck are never skipped.

### 1. Split tox environments

Create separate tox environments for each check category:

| New Environment | Command(s) |
|-----------------|------------|
| `lint` | `ruff check`, `ruff format --check` |
| `typecheck` | `mypy` |
| `spellcheck` | `codespell` |
| `spec-lint` | `ethereum-spec-lint` |
| `lockcheck` | `uv lock --check` |
| `actionlint` | `actionlint` |

This is similar to the [execution-spec-tests](https://github.com/ethereum/execution-spec-tests/blob/main/tox.ini) pattern.

Additionally, create a `fast-checks` alias environment that runs all split checks without code duplication. See [lean-ethereum's tox.ini](https://github.com/leanEthereum/leanSpec/blob/fbbacbea4545be870e25e3c00a90fc69e019c5bb/tox.ini#L17-L23) for this pattern:

```ini
[testenv:fast-checks]
description = Run all fast checks (lint, typecheck, spellcheck, spec-lint, lockcheck, actionlint)
commands =
    {[testenv:lint]commands}
    {[testenv:typecheck]commands}
    {[testenv:spellcheck]commands}
    {[testenv:spec-lint]commands}
    {[testenv:lockcheck]commands}
    {[testenv:actionlint]commands}
```

This enables both the aggregate command (`tox -e fast-checks`) and granular runs (`tox -e spellcheck`).

Consider aliasing a `static` tox env, too for backwards compatibility.

### 2. Create dedicated workflow

Create `.github/workflows/fast-checks.yaml`:

- Triggers on **all** pushes and PRs (no `paths-ignore`).
- Runs fast (~2 min total).
- Executes `tox -e fast-checks` (the alias) or runs individual envs in parallel jobs.

### 3. Maintain fast-checks as a prerequisite

The current `test.yaml` already uses `static` as a prerequisite for all other jobs (`needs: static`). This proposal maintains that behavior while extracting it into a reusable workflow:

```yaml
jobs:
  fast-checks:
    uses: ./.github/workflows/fast-checks.yaml

  py3:
    needs: [fast-checks]
    # ...
```

**Benefits of extraction:**

- **Decoupled triggers** - `fast-checks.yaml` has no `paths-ignore` and always runs. Other workflows (`test.yaml`, `benchmark.yaml`, `hive-consume.yaml`) keep their `paths-ignore` for expensive tests but call `fast-checks.yaml` as a prerequisite.
- **Single source of truth** - Fast checks are defined once, reused across all workflows via `uses:`.
- **Clearer CI view** - "fast-checks" appears as a separate workflow in GitHub Actions UI.

Alternatively, use `workflow_run` trigger or repository rulesets to enforce this.

## Related

- #2038: Optimizes CI by avoiding unnecessary checks, but introduces the side effect that doc-only PRs will not get spellchecked.
- Follow-up: GitHub Summary output with reproduce/fix commands (separate issue).

danceratopz:
## 🗒️ Description

This PR improves static check developer experience and introduces finer-grained tox environments for static checks as were configured in [ethereum/execution-spec-tests](https://github.com/ethereum/execution-spec-tests) pre-Weld.

It offers 4 main improvements to dev-ex:

1. **Flexible entry point** - User can choose their preferred "check" (aka recipe) runner (esp. helpful for non-Python devs):
    ```
    tox -e <env/check-name>  # or uvx tox -e <env/check-name>
    just <check-name>
    make <check-name>
    ```
    With this PR, I recommend replacing `uvx tox -e static` with:
    ```
    just check
    # or
    make check
    ```
    These run: `uvx tox --parallel -e lint,format,typecheck,markdownlint,spec-lint,spellcheck,actionlint,changelog,lockcheck`, NOT `uvx tox -e check` which is not parallelized.

2. **Parallelization** - Each one of the commands in the `tox` `static` env is now a separate `tox` env. This means they can be ran in parallel:

    i. **Faster** (~6s → ~3s warm, with mypy cache and `.tox/` present).
    ii. All checks run; they **don't stop upon first fail** (particularly annoying in CI).

3. **Tips upon fail** - Checks are wrapped in a helper `scripts/fast_checks.py` that:

    i. Provides auto-fix hints on failure (shown on command-line and in CI job summaries, screenshot below).
    ii. Uses a soft-fail mode for markdown-cli2 (locally, this was previously in `packages/testing/src/execution_testing/cli/tox_helpers.py`, now delted).
  
4. **Fix helper recipe** - Adds a "fix" `just`/`make` "recipe" for safe `ruff` (`ruff check --fix`  and `ruff format`):

    ```
    just fix
    make fix
    ``` 

This PR also:

- The `static` checks moved out of `test.yaml` to `check.yaml` and is configured to ensures no workflows get triggered if `check.yaml` fails. This involves some unavoidable duplication as each workflow must call `check.yaml` independently to preserve path filtering. **This needs discussion.**
- Maintains backwards compatibility with `tox -e static`; which also has `check` as alias.
- Updates documentation for the new commands and build tools.

#### Fix Hint Github Summary Screenshot

image

#### Fix Hint Console Output (also available locally)

image

From: https://github.com/danceratopz/execution-specs/actions/runs/21474528622
## 🔗 Related Issues or PRs

Closes #2041
- #2041

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

image


danceratopz:
Someone asked an excellent question via DM:

> i still don't really get what value tox brings locally

Essentially we could consider removing tox and rely on just or make as a check runner. However, as-is, I would keep tox as it still provides some useful functionality in both the old and new setups.
﻿
First some background, as I understand it:
﻿
- venv creation and management in Python was painful.
- one of the main value props of tox is/was helping with venvs.
- uv fixes this (coz venvs are free/fast), eroding the tox value prop.

I'd keep tox coz it still has some nice orchestration features:

1. run envs in parallel (this pr splits static to make this possible) this is now used by just/make (they call tox) to:
    - be fast,
    - get feedback on all checks in one command.
2. helps configure testing w/diff Python flavors, e.g., pypy (could be 3.{12,13,14}, but we don't).

Perhaps I'm missing something else?

.github/workflows/check.yaml
```diff
@@ -0,0 +1,58 @@
...
+            docs/**/*.md
+            *.md
+
+  sha-pinned-actions:
+    name: SHA Pinned Actions
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+      - name: Ensure SHA pinned actions
+        uses: zgosalvez/github-actions-ensure-sha-pinned-actions@6124774845927d14c601359ab8138699fa5b70c3
```
SamWilsn, line 58: This is available in the repository settings now, unless this serves a different purpose?

.github/workflows/benchmark.yaml
```diff
@@ -26,8 +26,12 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check:
+    uses: ./.github/workflows/check.yaml
+
```
SamWilsn, line 31: Hm, instead of duplicating check, would [`on: workflow_run`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run) work?

Justfile
SamWilsn, line 1: We should pick exactly one build system. I understand the desire to make our project as accessible as possible to external developers, but adding a `Makefile`/`Justfile`/`meson.build`/`CMakeLists.txt`/`package.json`/`pom.xml`/`ExecutionSpecs.sln`/`Cargo.toml` file just to support muscle memory for another language is silly.

Every build we add is more maintenance for us. If I want to change the build system, I don't want to have to test N different entry points.

Ideally we should have a nice short "Getting Started" section of our readme that includes installation instructions, common commands, and links to further documentation for advanced usage. If a developer can't be bothered to look in our readme for how to run our tooling, I don't really want their contributions.

.pre-commit-config.yaml
SamWilsn, line 1: I've done a bit more reading into this tool. It's not _as_ bad as I feared. You can't sneak a hook into `post-checkout` silently; the user needs to rerun `pre-commit install`.

It still makes me uncomfortable, and I would really rather not promote it, but I'll no longer die on this hill.

scripts/fast_checks.py
SamWilsn, line 1: If we're doing something like this anyway, why not make a `build.py` in the root directory and just use that instead of just/make/tox?

docs/getting_started/code_standards_details.md
```diff
@@ -93,7 +103,7 @@ For more information, see [Pre-commit Hooks Documentation](../dev/precommit.md).
 
 ## Formatting and Line Length
 
-The Python code in @ethereum/execution-spec-tests is formatted with `ruff` with a line length of 100 characters.
+The Python code in @ethereum/execution-specs is formatted with `ruff` with a line length of 100 characters.
```
SamWilsn, line 106:
```suggestion
The Python code in @ethereum/execution-specs is formatted with `ruff` with a line length of 79 characters.
```

.github/workflows/benchmark.yaml
SamWilsn: Hm, `workflow_run` has a few security considerations. Perhaps [`workflow_call`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_call) might be safer.

SamWilsn (commented):
Very old review I forgot to submit ^^;;

felix314159:
Will this PR get updated to remove tox? The PR was opened 2026-01-29, which is the day where we decided in the team meeting to get rid of tox (see gemini summary of 2026-01-29, although gemini phrased it differently), but it still includes tox. Issue #2041 is a bit older so that probably can be closed/reworked when tox is removed

danceratopz:
> Will this PR get updated to remove tox? The PR was opened 2026-01-29, which is the day where we decided in the team meeting to get rid of tox (see gemini summary of 2026-01-29, although gemini phrased it differently), but it still includes tox. Issue #2041 is a bit older so that probably can be closed/reworked when tox is removed

Yes, I'd like to update this PR to remove `tox`. 

In the same testing call, this was deemed low priority and added undue stress on the team, so it was put on hold. I can update if we'd like to prioritize it. Feel free to bring it up in the testing meeting.

We did agree to remove `tox`, but we didn't agree what to replace it with. Currently, my plan would be to update this PR to use `just`, https://just.systems/man/en/, in order to see how feasible a bare `uv` + `just` setup is. `just` has the simplicity of `make` but with cleaner syntax. Other build systems, e.g., `nox`, `waf` seem overkill for our purposes. You or anyone else should feel free to chime in if you/they have an opinion here!

felix314159:
I personally would prefer `uv` + `make`. I looked at some specific code examples of `make` vs `just` and the difference sometimes is just one line of code. No strong feelings though. `make` has the advantage of having 50 years of training data, so that could make things easier for 'us' even when it at first might seem otherwise

.github/workflows/check.yaml
danceratopz: Good point.

Justfile
danceratopz: Good point, got a bit carried away here! :smile:

.pre-commit-config.yaml
danceratopz: Thanks for checking it out, this was unintentionally committed, sorry about that!

scripts/fast_checks.py
danceratopz: I think this approach is now overkill.

danceratopz:
Thanks a lot @SamWilsn for your review and @felix314159 for your comments! 

> I personally would prefer uv + make. I looked at some specific code examples of make vs just and the difference sometimes is just one line of code. No strong feelings though. make has the advantage of having 50 years of training data, so that could make things easier for 'us' even when it at first might seem otherwise

Appreciate the feedback, I personally feel that just's UX improvements make it worth it (simple argument pass-through, tasks split by group in help output, no .PHONY, no tabs) and we don't need the additional complexity of a build system. 


I definitely got a bit carried away in this PR. And made sweeping changes across the build system and CI. I made a more focussed proposal (just a suggestion!) for replacing tox with just in:
- https://github.com/ethereum/execution-specs/pull/2555

PR #2105, opened 2026-01-29
refactor(spec-tool): remove unused fields

Galoretka:
Previously, the Load class stored _network and _fork_module fields that were never read anywhere in the codebase. One of them was even whitelisted in vulture_whitelist.py despite having no real dynamic usage. This made the loader API misleading and introduced unnecessary internal state.


spencer-tb (approved):
LGTM!

fselmo (approved):
Needed a nudge in the right direction to pass lint. Rebased and added a commit here. Waiting a bit for CI to pass then will merge 👍🏼

PR #2109, opened 2026-01-30
feat(specs): add max bal item check

nerolation:
🗒️ Description

Adds a gas limit constraint validation for EIP-7928 block access lists.

It bounds the block access list size based on available gas after accounting for transaction base costs.

Changes:
- Add BlockAccessListGasLimitExceededError exception
- Add validate_block_access_list_gas_limit() in block_access_lists/builder.py
- Call validation after building the BAL in apply_body()

🔗 Related Issues or PRs

https://github.com/ethereum/EIPs/pull/11223

### DIFF
--- a/packages/testing/src/execution_testing/forks/forks/forks.py
+++ b/packages/testing/src/execution_testing/forks/forks/forks.py
@@ -3449,7 +3449,7 @@ class Amsterdam(BPO2):
         in EIP-7928.
         """
         return replace(
-            super(Osaka, cls).gas_costs(
+            super(Amsterdam, cls).gas_costs(
                 block_number=block_number, timestamp=timestamp
             ),
             GAS_BLOCK_ACCESS_LIST_ITEM=2000,
--- a/packages/testing/src/execution_testing/forks/forks/forks.py
+++ b/packages/testing/src/execution_testing/forks/forks/forks.py
@@ -200,6 +200,7 @@ class Frontier(BaseFork, solc_name="homestead"):
             GAS_PRECOMPILE_BLS_PAIRING_BASE=0,
             GAS_PRECOMPILE_BLS_PAIRING_PER_PAIR=0,
             GAS_PRECOMPILE_P256VERIFY=0,
+            GAS_BLOCK_ACCESS_LIST_ITEM=0,
         )
 
     @classmethod
@@ -1010,6 +1011,14 @@ class Frontier(BaseFork, solc_name="homestead"):
         """At genesis, header must not contain block access list hash."""
         del block_number, timestamp
         return False
+
+    @classmethod
+    def empty_block_bal_item_count(
+        cls, *, block_number: int = 0, timestamp: int = 0
+    ) -> int:
+        """Pre-Amsterdam forks have no block access list."""
+        del block_number, timestamp
+        return 0
 
     @classmethod
     def engine_new_payload_version(
@@ -3430,6 +3439,37 @@ class Amsterdam(BPO2):
         """
         del block_number, timestamp
         return True
+
+    @classmethod
+    def gas_costs(
+        cls, *, block_number: int = 0, timestamp: int = 0
+    ) -> GasCosts:
+        """
+        On Amsterdam, the cost per block access list item is introduced
+        in EIP-7928.
+        """
+        return replace(
+            super(Osaka, cls).gas_costs(
+                block_number=block_number, timestamp=timestamp
+            ),
+            GAS_BLOCK_ACCESS_LIST_ITEM=2000,
+        )
+
+    @classmethod
+    def empty_block_bal_item_count(
+        cls, *, block_number: int = 0, timestamp: int = 0
+    ) -> int:
+        """
+        Return the BAL item count for an empty Amsterdam block.
+
+        Four system contracts produce 15 items:
+          EIP-4788 beacon roots:           1 address + 1 write + 1 read = 3
+          EIP-2935 history storage:        1 address + 1 write          = 2
+          EIP-7002 withdrawal requests:    1 address + 4 reads          = 5
+          EIP-7251 consolidation requests: 1 address + 4 reads          = 5
+        """
+        del block_number, timestamp
+        return 15
 
     @classmethod
     def is_deployed(cls) -> bool:
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -35,6 +35,7 @@ from .block_access_lists import (
     BlockAccessListBuilder,
     build_block_access_list,
     hash_block_access_list,
+    validate_block_access_list_gas_limit,
 )
 from .blocks import Block, Header, Log, Receipt, Withdrawal, encode_receipt
 from .bloom import logs_bloom
@@ -812,6 +813,12 @@ def apply_body(
 
     block_output.block_access_list = build_block_access_list(
         block_env.block_access_list_builder, block_env.state
+    )
+
+    # Validate block access list gas limit constraint (EIP-7928)
+    validate_block_access_list_gas_limit(
+        block_access_list=block_output.block_access_list,
+        block_gas_limit=block_env.block_gas_limit,
     )
 
     return block_output
--- a/src/ethereum/forks/amsterdam/vm/gas.py
+++ b/src/ethereum/forks/amsterdam/vm/gas.py
@@ -84,6 +84,8 @@ GAS_BLS_G2_ADD = Uint(600)
 GAS_BLS_G2_MUL = Uint(22500)
 GAS_BLS_G2_MAP = Uint(23800)
 
+GAS_BLOCK_ACCESS_LIST_ITEM = Uint(2000)
+
 
 @dataclass
 class ExtendMemory:
--- a/tests/frontier/validation/test_header.py
+++ b/tests/frontier/validation/test_header.py
@@ -1,6 +1,7 @@
 """Test the block header validations applied from Frontier."""
 
 import pytest
+from execution_testing import Fork
 from execution_testing.base_types.base_types import ZeroPaddedHexNumber
 from execution_testing.base_types.composite_types import Alloc
 from execution_testing.exceptions.exceptions import BlockException
@@ -18,7 +19,14 @@ from execution_testing.test_types.block_types import Environment
         pytest.param(0, marks=pytest.mark.exception_test),
         pytest.param(1, marks=pytest.mark.exception_test),
         pytest.param(4999, marks=pytest.mark.exception_test),
-        5000,
+        pytest.param(5000, marks=pytest.mark.valid_until("Osaka")),
+        pytest.param(
+            5000,
+            marks=[
+                pytest.mark.valid_from("Amsterdam"),
+                pytest.mark.exception_test,
+            ],
+        ),
     ],
 )
 def test_gas_limit_below_minimum(
@@ -26,6 +34,7 @@ def test_gas_limit_below_minimum(
     pre: Alloc,
     gas_limit: int,
     env: Environment,
+    fork: Fork,
 ) -> None:
     """
     Tests that a block with a gas limit below the minimum throws an error.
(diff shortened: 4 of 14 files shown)


src/ethereum/forks/amsterdam/fork.py
```diff
@@ -834,6 +839,13 @@ def apply_body(
         block_env.state_changes
     )
 
+    # Validate block access list gas limit constraint (EIP-7928)
+    validate_block_access_list_gas_limit(
```
gurukamath, line 819: Should the `t8n` tool be updated to reflect this as well?

nerolation:
Note that the two new constants are from EIP-7002 and EIP-7251.

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -517,3 +519,56 @@ def build_block_access_list(
...
+    system_allowance = (
+        Uint(15)
+        + Uint(3)
+        * (
+            MAX_WITHDRAWAL_REQUESTS_PER_BLOCK
+            + MAX_CONSOLIDATION_REQUESTS_PER_BLOCK
+        )
+    ) * item_cost
+
+    if bal_items * item_cost > available_gas + system_allowance:
```
jochem-brouwer: This assumes that all `bal_items` have a cost, but there are also free items, like the sender and receiver address. This will flag BALs which are valid as "too full"

jochem-brouwer: We should also add tests with BALs near max-size (I think we already have this) to check if clients don't early-reject those because they are too large. We cant create a BAL which has the max size as the hard cap calculation makes some assumptions which are not true in practice (SSTORE/SLOAD without pushing items to stack for instance)

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -517,3 +519,56 @@ def build_block_access_list(
...
+        MAX_CONSOLIDATION_REQUESTS_PER_BLOCK,
+        MAX_WITHDRAWAL_REQUESTS_PER_BLOCK,
+    )
+    from ..vm.gas import GAS_WARM_ACCESS
+
+    total_storage_reads = sum(
+        len(account.storage_reads) for account in block_access_list
+    )
+    total_addresses = len(block_access_list)
+    bal_items = Uint(total_storage_reads + total_addresses)
```
jochem-brouwer: total_storage_reads cost `TX_ACCESS_LIST_STORAGE_KEY_COST` per item, and for address the param respectively to that for addresses (which is higher). The cost is thus underestimated here (which is fine, because overestimating cost will invalidate blocks faster)

jochem-brouwer (commented):
Some comments 😄

src/ethereum/forks/amsterdam/block_access_lists/builder.py
nerolation: yeah good point!
I wouldn't say you get the sender and the receiver for free though - there's still the 21k gas costs which is above the 2000. The only truly "free" items are those from system contracts, if I'm not forgetting something here.

> We should also add tests with BALs near max-size (I think we already have this) to check if clients don't early-reject those because they are too large. We cant create a BAL which has the max size as the hard cap calculation makes some assumptions which are not true in practice (SSTORE/SLOAD without pushing items to stack for instance)

The goal here was to not create any new practical constrain for block building. This means, as long as there is gas available during builders, we must ensure that the BAL item size isn't hit yet -> thus the very conservate 2000 gas per item + not having the system contract items impact it.
Re tests, yeah, we should definitely have test for this, e.g. create a block that is valid (thus it must obey the max item limit too) but as well have tests with invalid BALs that are slightly over the max-items to make sure those are rejected early on.

src/ethereum/forks/amsterdam/block_access_lists/builder.py
nerolation: yeah right - it's a trade-off between simplicity and accuracy. Since this is mainly needed for early rejections of BALs that have way too many items (e.g. 2x what is realistic; those would cause 2x the work which we aren't accounting for when setting the gas limit), to counter DoS, I tend leaning towards simplicity first.

src/ethereum/forks/amsterdam/fork.py
fselmo: Yeah, we likely need to add this [here](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/src/ethereum_spec_tools/evm_tools/t8n/__init__.py#L463-L466) so we can test this appropriately.

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -517,3 +519,47 @@ def build_block_access_list(
...
+
+    storage_keys = sum(
+        len(
+            {sc.slot for sc in account.storage_changes}
+            | set(account.storage_reads)
+        )
+        for account in block_access_list
+    )
+    addresses = len({account.address for account in block_access_list})
+    bal_items = Uint(storage_keys + addresses)
```
gurukamath:
```suggestion
    count_bal_items = Uint(0)
    for account in block_access_list:
        # Count each address as one item
        count_bal_items += Uint(1)

        # Collect unique storage keys across both
        # reads and writes
        unique_slots: Set[U256] = set()
        for slot_change in account.storage_changes:
            unique_slots.add(slot_change.slot)
        for slot in account.storage_reads:
            unique_slots.add(slot)

        # Count each unique storage key as one item
        count_bal_items += Uint(len(unique_slots))
```

From a readability perspective, I think we should run an explicit loop instead of the nested generator.

fselmo:
Anything missing here @nerolation? I think we need to still add this validation to t8n as mentioned here and this needs a good rebase. Let me know when it's ready for eyes now that this change to the EIP has been merged in the EIPs repo 👀 .

nerolation:
I think this one is ready to be merged.

One question I had was if we also want the "unique items" check (introduced to the EIP [here](https://github.com/ethereum/EIPs/pull/11338)) in the specs or if this is something that happens before a block enters the EELS STF?

Curious what you think.

fselmo:
@nerolation we had some big changes to how BALs were organized upstream. I rebased this but copied your original state of things [here](https://github.com/fselmo/execution-specs/tree/copy/toni/max-items) so we can better diff it as a sanity check.

I also fixed an undefined var issue and added the validation to t8n as I had commented in the PR. This now correctly catches the same bug identified by nimbus where these 3 tests fail due to a low block gas limit in the tests:

- [FAILED ] (  0.01s) test_gas_limit_below_minimum.json
- [FAILED ] (  0.00s) InternalCallHittingGasLimit.json
- [FAILED ] (  0.00s) static_InternalCallHittingGasLimit.json 

so I'm going to work on a fix for these tests and push it here as well, then I think we're ready to merge this 👀

fselmo:
> so I'm going to work on a fix for these tests and push it here as well, then I think we're ready to merge this 👀

Ok, this was essentially fixed in the last commit by adding this validation check to t8n and inheriting the new exception from `InvalidBlock`. I think this looks good now.

> One question I had was if we also want the "unique items" check (introduced to the EIP [here](https://github.com/ethereum/EIPs/pull/11338)) in the specs or if this is something that happens before a block enters the EELS STF?

The builder was already built with this in mind, so we already generate unique items. If anything I think we can be a bit more explicit in the docstring to state this uniqueness requirement. I added this as a commit [here](https://github.com/ethereum/execution-specs/pull/2109/commits/fe171a38f33ee0a2a1a104c1cd1ac3d5662fed78), curious if you feel this deserves something more 👀. Lmk your thoughts.

fselmo (dismissed):
@nerolation I did a bit of rebasing here, added the validation to t8n so we can have this when filling tests (since `apply_body` is entirely missed via t8n path), and aligned the changes with the new `forks/amsterdam/block_access_lists.py` structure.

This looks good to me to merge so going to approve! Would love to get a last look and sign-off from you though before merging 👀.

gurukamath (dismissed):

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -3431,6 +3440,37 @@ def header_bal_hash_required(
...
+        """
+        return replace(
+            super(Osaka, cls).gas_costs(
+                block_number=block_number, timestamp=timestamp
+            ),
+            GAS_BLOCK_ACCESS_LIST_ITEM=2000,
+        )
+
+    @classmethod
+    def empty_block_bal_item_count(
```
fselmo, line 3459: This seemed like the best way to account for this boundary as it is something that is hard to define otherwise. Curious if there's a better way here?

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -84,6 +84,8 @@
 GAS_BLS_G2_MUL = Uint(22500)
 GAS_BLS_G2_MAP = Uint(23800)
 
+GAS_BLOCK_ACCESS_LIST_ITEM = Uint(2000)
```
fselmo, line 87: @nerolation see my comment [here](https://github.com/ethereum/EIPs/pull/11323#pullrequestreview-3863870579). This seems useful to define as a fork constant in case this ever changes. Would love your thoughts on defining this in the EIP itself.

tests/static/state_tests/stStaticCall/static_InternalCallHittingGasLimitFiller.json
```diff
@@ -3,7 +3,7 @@
         "env" : {
             "currentCoinbase" : "2adf5374fce5edbc8e2a8697c15331677e6ebf0b",
             "currentDifficulty" : "0x020000",
-            "currentGasLimit" : "22000",
+            "currentGasLimit" : "100000",
```
fselmo, line 6: It seems tests relate to testing the transaction gas limit as the tx gas limit is what's being hit here. I raised the block gas limit here so that we don't run into the BAL item check issues for Amsterdam. I could use a sanity check here that this isn't trying to test the block gas limit when the tx gas limit is the same as the block gas limit (this wouldn't make sense to me otherwise the tx gas wouldn't be equal to the block gas in the tests but doesn't hurt to check 😄).

tests/frontier/validation/test_header.py
```diff
@@ -18,27 +19,35 @@
...
-        exception=BlockException.INVALID_GASLIMIT
-        if gas_limit < 5000
-        else None,
-    )
+    block = Block(txs=[])
+
+    if gas_limit < 5000:
+        block.rlp_modifier = Header(**modified_fields)
+        block.exception = BlockException.INVALID_GASLIMIT
+    elif fork.header_bal_hash_required():
```
fselmo, line 50: `>=Amsterdam` this test will fail with the BAL item check as `15 * 2k = 30k` which is greater than the lower gas limit boundary of `5k`. I think perhaps it's useful to add this here instead of skipping this test with `mark.valid_until("Osaka")`. Thoughts?

fselmo:
@gurukamath, and anyone else that reviews, I had conflated the client recommendation check on the gas limit [here](https://github.com/ethereum/EIPs/pull/11223/changes) with the max item check [here](https://github.com/ethereum/EIPs/pull/11323/changes). The max item check _is_ a `MUST` check and does invalidate the BAL safely within the bounds of what the client recommendation does, to my understanding. With BOTH of these recent changes taken into account, this change here _is_ an implementation that affects consensus and so I think we're good to add this here and add proper tests for this.

I added a boundary check test here and I think this is good to go. I will refrain from approving since I added some assumptions here and left an open question on whether to properly define the gas for BAL items as a fork variable. I think this is a good call as this is based on other gas values for a fork but curious on thoughts there.

I've left some notes on relevant lines here, feel free to mark resolved if they are clear enough, or answer them inline.

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -3431,6 +3440,37 @@ def header_bal_hash_required(
...
+    @classmethod
+    def gas_costs(
+        cls, *, block_number: int = 0, timestamp: int = 0
+    ) -> GasCosts:
+        """
+        On Amsterdam, the cost per block access list item is introduced
+        in EIP-7928.
+        """
+        return replace(
+            super(Osaka, cls).gas_costs(
```
marioevz:
```suggestion
            super(Amsterdam, cls).gas_costs(
```
Was causing the CI issues.

marioevz (approved):
LGTM.

PR #2110, opened 2026-01-31
fix(test,benchmark): cap data tests to rlp block limit

fselmo:
## 🗒️ Description

- Consider RLP limit for max data test... this does cap the data gas at the RLP limit so takes the min between the benchmark gas value and the rlp limit... I think this is OK? But we'd have to consider the intent of the test. Not sure there's another way tbh?
- Use vars instead of hard-coded values for fixtures

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2026-01-30 at 17 09 31

### DIFF
--- a/tests/benchmark/compute/scenario/test_transaction_types.py
+++ b/tests/benchmark/compute/scenario/test_transaction_types.py
@@ -190,15 +190,15 @@ def test_block_full_of_ether_transfers(
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


fselmo:
meant to PR this [here](https://github.com/LouisTsai-Csie/execution-specs/pull/2)

PR #2113, opened 2026-02-02
feat(tests): EIP-7928 tests for EIP 2935

raxhvl:
## 🗒️ Description
Tests for the effects of EIP-2935 historical block hashes on EIP-7928.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```


raxhvl:
Note: The blockhash value set during pre-execution is internal to EVM and hence is not verified by BAL which in this test should be okay since clients will eventually compare values from fixtures.

The expression below is used to check the presence of the correct storage slot:

```py
storage_changes=[
             BalStorageSlot(
                     slot=get_history_storage_slot(0),
                     slot_changes=[],
            ),
]
```

It seems to work but I expect this to mean **Storage slot has NO changes** and NOT **storage slot has SOME changes**

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip2935.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+def test_bal_2935_selfdestruct_to_history_storage(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+    fork: Fork,
+) -> None:
+    """
+    Ensure BAL captures SELFDESTRUCT to history storage address alongside
+    system call storage writes.
```
fselmo, line 296: We don't seem to be checking any writes here (`BalStorageChanges`) as this states. Could we implement a similar approach as you did [here](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip4788.py#L50-L85) with beacon roots contract so that we have these expectations as well in the BAL?

I think we can make use of the new `validate_any_change` bool and validate these two things in the helper for each block:

- `HISTORY_STORAGE_ADDRESS` with `storage_changes=[BalStorageSlot(slot=..., validate_any_change=True)]`
- `SYSTEM_ADDRESS: None`

wdyt?

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip2935.py
```diff
@@ -0,0 +1,365 @@
...
+
+    tx = Transaction(
+        sender=alice,
+        to=oracle,
+        data=Hash(query_block_number),
+        value=value,
+        gas_limit=fork.transaction_gas_limit_cap(),
+    )
+
+    # A setup up block that writes genesis block-hash
```
fselmo, line 210:
```suggestion
    # A setup block that writes genesis block-hash
```

fselmo:
> Note: The blockhash value set during pre-execution is internal to EVM and hence is not verified by BAL which in this test should be okay since clients will eventually compare values from fixtures.
> 
> The expression below is used to check the presence of the correct storage slot:
> 
> ```python
> storage_changes=[
>              BalStorageSlot(
>                      slot=get_history_storage_slot(0),
>                      slot_changes=[],
>             ),
> ]
> ```
> 
> It seems to work but I expect this to mean **Storage slot has NO changes** and NOT **storage slot has SOME changes**

Nice catch... I don't think we've run into this yet as we have been able to be pretty deterministic with the expectations. I think this is a no-brainer, at least in consistency side. We need to keep consistency that `[]` means just that (empty set). I'm pushing a commit here that uses `validate_any_change: bool` as a new field and makes sure that you can't have this flag AND define `slot_changes`, you're either explicit or you're not.

Definitely happy to iterate on this (and we should discuss), but I think this is pretty explicit at least and would work well. Lmk your thoughts.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip2935.py
```diff
@@ -0,0 +1,365 @@
...
+    # to history storage contract so that it can be
+    # queried later.
+    block_1 = Block()
```
fselmo: This block could also have the BAL expectation from the helper method even though it's empty.

fselmo (commented):
Ty for this! Not sure if this was ready yet given it's failing CI but gave it a first pass. I added an improvement commit as I mentioned above and fixed the lint so we can run CI on this. Let me know your thoughts please.

... also some minor notes from Claude review:

```
- Missing test: Wrong calldata size queries (EIP-2935 reverts on calldata != 32 bytes)
- test_cases.md: Invalid query entry doesn't mention oracle balance_changes when value > 0
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip2935.py
raxhvl: Done.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip2935.py
raxhvl: Done.

raxhvl:
1. Refactor: Added a helper method for system expectation.
2. Fixes: Added missing write expectation to some of the blocks. Test case description added.
3. Additions: New test case for invalid calldata size.

Thanks for the API to validate any change. Should we a similar API for post too? I tried setting slot to `TRUE` but it doesnt work.

fselmo:
> Should we a similar API for post too? I tried setting slot to `TRUE` but it doesnt work.

I think we should. Going to create an issue to track this as it has bothered me for a little bit and has come up again somewhere else.

edit: created a tracking issue here: https://github.com/ethereum/execution-specs/issues/2227

fselmo (approved):
lgtm 👍🏼

PR #2117, opened 2026-02-02
fix(test-fill): improve memory buildup for fill; merge on `SIGINT` / `SIGTERM`

fselmo:
## 🗒️ Description

Recent speedups to `fill` command use JSONL partial fixtures on the fly and merge at the end. This helped `fill` speeds greatly but created some friction points addressed in this PR:

- Stream fixture and index writes to disk, instead of holding these in memory. This should allow for constant memory instead of ever-increasing.
- If a `KeyboardInterrupt` happens now, all of the fill that has happened until then is wasted because the merge is never called. This change makes sure to merge the partial fixtures on `SIGINT` and `SIGTERM` so you can stop in the middle of a fill and still get the current state of filling inside of `fixtures/`.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2026-02-02 at 10 00 12

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -6,10 +6,12 @@ and that modifies pytest hooks in order to fill test specs for all tests
 and writes the generated fixtures to file.
 """
 
+import atexit
 import configparser
 import datetime
 import json
 import os
+import signal
 import warnings
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/custom_logging/plugin_logging.py
```diff
@@ -70,6 +70,13 @@ def pytest_addoption(parser: pytest.Parser) -> None:  # noqa: D103
             "specified."
         ),
     )
+    logging_group.addoption(
```
danceratopz: I'd lean towards to disabling `fill`'s log by default, which would make this option unnecessary. Wdyt?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/custom_logging/plugin_logging.py
```diff
@@ -70,6 +70,13 @@ def pytest_addoption(parser: pytest.Parser) -> None:  # noqa: D103
...
+        "--no-log",
```
danceratopz: nit: pr description says `--no-logs`.

I think I prefer `--no-log` as is here.

danceratopz (commented):
Nice :) Also tested with `--generate-all-formats` - working nicely.

Just a quick question on what the default log behavior should be.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/custom_logging/plugin_logging.py
fselmo: Let's tackle this separately I suppose. This would be a breaking change and it feels like it's out of scope here. I'll drop this commit for now.

marioevz (approved):
LGTM, thanks!

PR #2116, opened 2026-02-02
chore(ci): only fill native test formats in tox's `pypy3` env

danceratopz:
## 🗒️ Description

This PR introduces a workaround for memory issues while running the tox `pypy3` env (with Amsterdam tests) on our CI runners:
1. Bump `--until=BPO4` to `--until=Amsterdam`.
2. Don't `fill` derived tests formats; only fill the native test format: `-m "not derived_test".
    - E.g., the `blockchain_test_engine`, `blockchain_test_from_state_test`, fixture formats get ignored.
3. Clean-up: It additionally removes references to the unused `zkevm` marker.

~~Not filling derived test formats (point 1.) was not enough to enable `--until=Amsterdam`; 2. was necessary as well.~~
**Update:** With the various optimizations this week, pypy3 now fills (with these changes in ~22 mins). I think we can merge this, but we will have to continue to be mindful that we will run into issues again as we add new tests, probably latest in Bogota.

## 🔗 Related Issues or PRs

Closes #2055
- #2055

## ✅ Checklist

- [x] All: Ran fast `tox` checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    uvx tox -e static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).

#### Cute Animal Picture
image

Issue #1393, opened 2025-08-28
Dynamically assign number of pytest-xdist workers via `pytest_xdist_auto_num_workers`


Carsons-Eels:
It should be possible with [`pytest_xdist_auto_num_workers`](https://github.com/pytest-dev/pytest-xdist/blob/8b60b1ef5d48974a1cb69bc1a9843564bdc06498/src/xdist/plugin.py#L16) to:

- Read numbers of cores
- Read total system memory
- Determine if CPython or PyPy

And come up with a reasonable number of cores.

_Originally posted by @SamWilsn in https://github.com/ethereum/execution-specs/issues/2120#issuecomment-3843476312_

*[pytest docs: CPU parallelization](https://pytest-xdist.readthedocs.io/en/stable/distribution.html)* 

Right now pytest uses the `-n auto` flag that directs pytest-xdist to use as many processes as the computer running the tests has **physical** CPU cores, up to the limit set by `--maxprocesses`. The maximum number of processes is currently set to specific values based on our experience of the CI machine's performance.

We could likely increase throughput by switching to the `-n logical` flag which would use the max number of **logical** CPU cores, however, before we do that we need to dynamically adjust the `--maxprocesses` cap based on the system that is running the tests. To do that, we will have to figure out how to determine the available memory on the system and properly adjust to ensure we don't run out.

We should also investigate using the `--dist worksteal` flag for impact on test times which could handle re-assigning jobs to workers who have completed their test queue more efficiently.

Cyber-Mitch:
Assign this to me @Carsons-Eels. I want to work on it

DeborahOlaboye:
Hi @Carsons-Eels , I’d love to work on this issue. Could you please assign it to me?

Carsons-Eels:
We don't generally assign issues to external contributors due to low rates of follow through and difficulty managing velocity with respect to that kind of dependency. If you see an issue you want to work on, just submit a PR or start asking us questions if there is a lack of clarity on what/how to implement.

danceratopz:
I just looked into whether `logical` helps us. The tldr is I don't think so, as it will report the number of logical units, so for a machine with 16 hyperthreaded cores, it will report 32. As our test generation is not only computationally but also memory intensive, I think we should avoid doubling the workers.

In general, 
- `-n auto` "should" be the number of physical cores **if `psutil is available`**, otherwise defaults to the number of logical cores,
- `-n logical` is physical × 2 (if the cores are hyperthreaded).

[...]

PR #2120, opened 2026-02-03
chore(tooling,ci): throttle py3 xdist workers locally, use all cores in CI

danceratopz:
## 🗒️ Description

This PR changes the xdist behavior of the `py3` and `tests_pytest_py3` tox environments such that:
- **All physical available cores are used in CI** on self-hosted runners (`-n auto`) instead of being capped via `--maxprocesses`.
- **Only 6 cores are used for local development, by default**; overridable via [`PYTEST_XDIST_AUTO_NUM_WORKERS`](https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus#:~:text=PYTEST_XDIST_AUTO_NUM_WORKERS). This lower default intends to avoid xdist utilizing all cores, which leaves the local workstation unresponsive until the end of the pytest run.

It additionally adds `psutil` to the `dev` dependency group in order to accurately detect the number of physical cores, cf https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus

### Results

py3 runs:
- 16 workers, but before #2127 (removes 23k tests) and other opts, e.g #2079, #2140:
    ```
    ======== 110675 passed, 1603 skipped, 17 warnings in 1917.08s (0:31:57) ========
    ```
- 16 workers, but after the above optimizations:
    ```
    ======== 87957 passed, 1603 skipped, 17 warnings in 1411.02s (0:23:31) =========
    ```
- 10 workers (before this PR), with the above optimizations:
    ```
    ======== 87957 passed, 1603 skipped, 11 warnings in 1853.90s (0:30:53) =========
    ```

### Approach

Replace `-n auto` (and `-n auto --maxprocesses N`) with:

```
-n {env:PYTEST_XDIST_AUTO_NUM_WORKERS:6}
```

Tox resolves `{env:VAR:default}` at config-parse time from the host shell. CI workflows set the env var to `auto`; locally it falls back to `6`.

| Scenario | Host env | Resolves to |
|---|---|---|
| Local (default) | *unset* | `-n 6` |
| Local (override) | `PYTEST_XDIST_AUTO_NUM_WORKERS=3` | `-n 3` |
| CI | `PYTEST_XDIST_AUTO_NUM_WORKERS=auto` | `-n auto` |

#### Why not `passenv` + `PYTEST_XDIST_AUTO_NUM_WORKERS` directly?

xdist reads `PYTEST_XDIST_AUTO_NUM_WORKERS` from the **child** process environment to resolve `-n auto` to a worker count and requires an integer. Passing `auto` through `passenv` would cause xdist to crash on `int("auto")`. By resolving the env var in tox's own substitution layer, the child process only ever sees `-n <int>` or `-n auto` (with the env var absent), so xdist never encounters a non-integer value.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```
#### Cute Animal Picture

:t-rex:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -76,6 +76,8 @@ jobs:
       - uses: ./.github/actions/setup-env
       - name: Run py3 tests
         run: tox -e py3
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: auto
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7
         with:
@@ -152,6 +154,8 @@ jobs:
       - uses: ./.github/actions/build-evmone
       - name: Run py3 tests
         run: tox -e tests_pytest_py3
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: auto
 
   tests_pytest_pypy3:
     runs-on: [self-hosted-ghr, size-xl-x64]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -227,6 +227,7 @@ dev = [
     { include-group = "doc" },
     { include-group = "mkdocs" },
     "ethereum-execution[optimized]",
+    "psutil>=7.2.2",
 ]
 
 [tool.setuptools.dynamic]
--- a/tox.ini
+++ b/tox.ini
@@ -92,6 +92,7 @@ description = Fill the tests using EELS (with Python)
 commands =
     fill \
         -m "not slow and not benchmark" \
+        # loadgroup: serialize xdist_group("bigmem") tests onto one worker to limit peak memory
         -n {env:PYTEST_XDIST_AUTO_NUM_WORKERS:6} --dist=loadgroup \
         --skip-index \
         --cov-config=pyproject.toml \
@@ -124,7 +125,9 @@ commands =
         --show-capture=no \
         --disable-warnings \
         -m "not slow and not benchmark and not derived_test" \
-        -n auto --maxprocesses 7 --dist=loadgroup \
+        -n auto --maxprocesses 7 \
+        # loadgroup: serialize xdist_group("bigmem") tests onto one worker to limit peak memory
+        --dist=loadgroup \
         --basetemp="{temp_dir}/pytest" \
         --log-to "{toxworkdir}/logs" \
         --clean \
--- a/tox.ini
+++ b/tox.ini
@@ -92,7 +92,7 @@ description = Fill the tests using EELS (with Python)
 commands =
     fill \
         -m "not slow and not benchmark" \
-        -n auto --maxprocesses 10 --dist=loadgroup \
+        -n auto --dist=loadgroup \
         --skip-index \
         --cov-config=pyproject.toml \
         --cov=ethereum \
--- a/tox.ini
+++ b/tox.ini
@@ -42,7 +42,7 @@ changedir = {toxinidir}/packages/testing
 package = editable
 commands =
     pytest \
-        -n auto --maxprocesses 6 \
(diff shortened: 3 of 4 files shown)


SamWilsn:
> avoid deadlocks when xdist spawns a worker per core

How does limiting the number of workers avoid a deadlock?

SamWilsn:
I'm thinking we maybe combine this with #1393? It should be possible with [`pytest_xdist_auto_num_workers`](https://github.com/pytest-dev/pytest-xdist/blob/8b60b1ef5d48974a1cb69bc1a9843564bdc06498/src/xdist/plugin.py#L16) to:

- Read numbers of cores
- Read total system memory
- Determine if CPython or PyPy

And come up with a reasonable number of cores.

danceratopz:
> > avoid deadlocks when xdist spawns a worker per core
> 
> How does limiting the number of workers avoid a deadlock?

Deadlock was a poor choice of wording! What I mean is: Running `-n auto` leads to full load, which makes your (local) machine unresponsive until pytest finishes.

Updated the description! :)

danceratopz:
> I'm thinking we maybe combine this with #1393?

 I took a look #1393 and commented there: https://github.com/ethereum/execution-specs/issues/1393#issuecomment-3853315256
 
 According to this, I've just updated this PR to:
 - Add `psutil` as a dev dep,

> It should be possible with [`pytest_xdist_auto_num_workers`](https://github.com/pytest-dev/pytest-xdist/blob/8b60b1ef5d48974a1cb69bc1a9843564bdc06498/src/xdist/plugin.py#L16) to:
> 
>     * Read numbers of cores
> 
>     * Read total system memory
> 
>     * Determine if CPython or PyPy
> 
> 
> And come up with a reasonable number of cores.

This is a great idea! This would def help future-proof our setup for CI runner changes. I'd prefer to defer this to a follow-up issue though (tbh, I need to move on in terms of priorities and patience), but happy if someone else picks it up! 

I'd just add for the follow-up issue, that this calculation probably won't be set in stone, as new optimizations will impact the memory footprint of generating tests.

For now,  this PR means:
- In CI: 
  - We accurately detect the number of physical cores and don't increase the count unintentionally to match the number of logical cores. 
  - We use the maximum number of physical cores with `-n auto`; this might be too memory hungry in the future.
- For local development: 
  - Defaults to 6.
  - Allows devs to override to their preferred value via the `PYTEST_XDIST_AUTO_NUM_WORKERS` env var.

spencer-tb (approved):
LGTM for now! I noticed `tests_pytest_py3` has `--dist=loadgroup` but py3 doesn't

danceratopz:
Rebased on forks/amsterdam to get the changes from #2116.

> LGTM for now! I noticed `tests_pytest_py3` has `--dist=loadgroup` but py3 doesn't

Thanks for flagging! It's the other way round (`py3` uses `loadgroup`). I added an inline comment explaining why for now:
https://github.com/danceratopz/execution-specs/blob/8cc9f9a1f5765f45ccab81accdfaf35cb0c20cee/tox.ini#L90-L95

We'll likely revisit this flag very soon, as we'll likely somehow also enable `--dist=worksteal` for better balance/distribution across workers.

danceratopz:
<del>
@SamWilsn I made a follow-up issue to not forget your idea [above](https://github.com/ethereum/execution-specs/pull/2120#issuecomment-3843476312): 
- https://github.com/ethereum/execution-specs/issues/2163
</del>

---

Ah, #1393 is already exactly this issue. I hadn't understood that from the description! I added your comment to #1393's description to make it clearer.

fselmo (approved):
It seems like this helps those of us who run these locally without an additional `-n` flag. I think this opens up some flexibility which is nice. Approving based on the other support here for this feature 👍🏼

PR #2130, opened 2026-02-04
feat(benchmark/tests): Extra sstore benchmark tests

marioevz:
## 🗒️ Description
Builds on top of changes done by #1774

Main differences:
- Osaka compatible
- Use `gas_benchmark_value` parameter instead of slot numbers
- Splits `test_storage_sload_benchmark` into  two separate test functions instead of using an `incrementer` (Variant 1 uses increasing slot number, variant 2 uses a constant key for SLOAD)
- Uses `bytecode.gas_cost` to calculate the exact gas cost of all bytecode used.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -10,11 +10,13 @@ abstract: BloatNet single-opcode benchmark cases for state-related operations.
 import json
 import math
 from pathlib import Path
+from typing import Tuple
 
 import pytest
 from execution_testing import (
     AccessList,
     Account,
+    Address,
     Alloc,
     BenchmarkTestFiller,
     Block,
@@ -400,7 +402,13 @@ def test_sstore_erc20_approve(
     )
 
 
-def sstore_helper_contract(sloads_before_sstore: bool) -> Bytecode:
+def sstore_helper_contract(
+    *,
+    sloads_before_sstore: bool,
+    key_warm: bool,
+    original_value: int,
+    new_value: int,
+) -> Tuple[Bytecode, Bytecode, Bytecode]:
     """
     Storage contract for benchmark slot access.
 
@@ -408,68 +416,72 @@ def sstore_helper_contract(sloads_before_sstore: bool) -> Bytecode:
     # - CALLDATA[0..31]: Number of slots to access
     # - CALLDATA[32..63]: Starting slot index
     # - CALLDATA[64..95]: Value to write
+
+    Returns:
+    - setup: Bytecode of the setup of the contract
+    - loop: Bytecode of the loop of the contract
+    - cleanup: Bytecode of the cleanup of the contract
+
     """
     setup = Bytecode()
     loop = Bytecode()
     cleanup = Bytecode()
 
-    start_marker = 10
-    end_marker = 30 + (3 if sloads_before_sstore else 0)
-
     setup += (
-        Op.CALLDATALOAD(0)  # num_slots
-        + Op.CALLDATALOAD(32)  # start_slot
+        Op.ADD(
+            Op.CALLDATALOAD(0), Op.CALLDATALOAD(32)
+        )  # num_slots + start_slot = end_slot
         + Op.CALLDATALOAD(64)  # value
+        + Op.CALLDATALOAD(32)  # start_slot = counter
     )
 
-    setup += Op.PUSH0  # Counter
-    setup += Op.JUMPDEST
-    # [counter, value, start_slot, num_slots]
+    # [counter, value, end_slot]
 
-    # Loop Condition: Counter < Num Slots
-    loop += Op.DUP4
-    loop += Op.DUP2
-    loop += Op.LT
-    loop += Op.ISZERO
-    loop += Op.PUSH1(end_marker)
-    loop += Op.JUMPI
-    # [counter, value, start_slot, num_slots]
+    start_marker = len(setup)
 
+    loop += Op.JUMPDEST
     # Loop Body: Store Value at Start Slot + Counter
-    loop += Op.DUP1
-    loop += Op.DUP4
-    loop += Op.ADD
-    loop += Op.DUP3
-    # [value, start_slot+counter, counter, value, start_slot, num_slots]
-
     if sloads_before_sstore:
-        loop += Op.DUP2
-        loop += Op.SLOAD
+        loop += Op.DUP1  # [counter, counter, value, end_slot]
+        loop += Op.SLOAD(key_warm=key_warm)
         loop += Op.POP
-        loop += Op.SWAP1
-        loop += Op.SSTORE
+        loop += Op.DUP2  # [value, counter, value, end_slot]
+        loop += Op.DUP2  # [counter, value, counter, value, end_slot]
+        loop += Op.SSTORE(
+            key_warm=False,
+            original_value=original_value,
+            new_value=new_value,
+        )
     else:
-        loop += Op.SWAP1
-        loop += Op.SSTORE  # STORAGE[start_slot + counter] = value
-    # [counter, value, start_slot, num_slots]
+        loop += Op.DUP2  # [value, counter, value, end_slot]
+        loop += Op.DUP2  # [counter, value, counter, value, end_slot]
+        loop += Op.SSTORE(  # STORAGE[counter, value] = value
+            key_warm=key_warm,
+            original_value=original_value,
+            new_value=new_value,
+        )
 
     # Loop Post: Increment Counter
     loop += Op.PUSH1(1)
     loop += Op.ADD
+    # [counter + 1, value, end_slot]
+
+    # Loop Condition: Counter < Num Slots
+    loop += Op.DUP3  # [end_slot, counter + 1, value, end_slot]
+    loop += Op.DUP2  # [counter + 1, end_slot, counter + 1, value, end_slot]
+    loop += Op.LT  # [counter + 1 < end_slot, counter + 1, value, end_slot]
+    loop += Op.ISZERO
+    loop += Op.ISZERO
     loop += Op.PUSH1(start_marker)
-    loop += Op.JUMP
-    # [counter + 1, value, start_slot, num_slots]
+    loop += Op.JUMPI
+    # [counter, value, end_slot]
 
     # Cleanup: Stop
-    cleanup += Op.JUMPDEST
     cleanup += Op.STOP
 
-    assert len(setup) - 1 == start_marker
-    assert len(setup) + len(loop) == end_marker
-    return setup + loop + cleanup
+    return setup, loop, cleanup
 
 
-@pytest.mark.parametrize("slot_count", [50, 100])
 @pytest.mark.parametrize("use_access_list", [True, False])
 @pytest.mark.parametrize("sloads_before_sstore", [True, False])
 @pytest.mark.parametrize("num_contracts", [1, 5, 10])
@@ -485,10 +497,10 @@ def sstore_helper_contract(sloads_before_sstore: bool) -> Bytecode:
 )
 def test_sstore_variants(
     benchmark_test: BenchmarkTestFiller,
+    fork: Fork,
     pre: Alloc,
     tx_gas_limit: int,
     gas_benchmark_value: int,
-    slot_count: int,
     use_access_list: bool,
     sloads_before_sstore: bool,
     num_contracts: int,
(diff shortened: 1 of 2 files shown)


tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,491 @@ def test_sstore_erc20_approve(
...
+@pytest.mark.parametrize("sloads_before_sstore", [True, False])
+@pytest.mark.parametrize("num_contracts", [1, 5, 10])
+@pytest.mark.parametrize(
+    "initial_value,write_value",
+    [
+        pytest.param(0, 0, id="zero_to_zero"),
+        pytest.param(0, 0xDEADBEEF, id="zero_to_nonzero"),
+        pytest.param(0xDEADBEEF, 0, id="nonzero_to_zero"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_diff"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_same"),
```
jochem-brouwer:
```suggestion
        pytest.param(0xDEADBEEF, 0xDEADBEEF, id="nonzero_to_same"),
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,491 @@ def test_sstore_erc20_approve(
...
+        for i in range(slots_per_contract):
+            expected_storage[i] = write_value
+
+        post[contract_addr] = Account(
+            code=contract,
+            storage=expected_storage,
+        )
+
+    benchmark_test(
+        blocks=[Block(txs=txs)],
```
jochem-brouwer, line 677: For refactor later on (new PR), we should make the `benchmark_test` accept a tx generator. In this tx generator we return the relevant fields for the newly generated tx (the gas limit is being handled by the benchmark test). In this case the calldata would differ per tx.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,491 @@ def test_sstore_erc20_approve(
...
+    ) -> int:
+        gsc = fork.gas_costs()
+        # SSTORE requires a minimum gas of G_CALL_STIPEND to operate.
+        # TODO: Correct fix is to introduce bytecode.gas_required.
+        return (
+            calc_gas_consumed(iteration_count, start_slot, contract_addr)
+            + gsc.G_CALL_STIPEND
+        )
+
+    # Calculate how many slots per contract per transaction are required
```
jochem-brouwer, line 584: Could this get a larger description? I'm not following what is being setup here. The slots per contract?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,491 @@ def test_sstore_erc20_approve(
...
+
+    start_slot = 0
+    txs: list[Transaction] = []
+    gas_used = 0
+    for iteration_count in iteration_counts:
+        calldata = get_calldata(iteration_count, start_slot)
+        access_list = get_access_list(
+            iteration_count, start_slot, contract_addr
+        )
+        tx_gas_limit = calc_gas_required(
```
jochem-brouwer, line 847: I'm actually not sure why this gas is calculated, if there is too many gas sent to the transaction but not all is spent, it will be refunded (and also does not count against  block gas limit). So not sure why tx gas limit is being set here.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,491 @@ def test_sstore_erc20_approve(
...
+) -> None:
+    """
+    Benchmark SLOAD instruction when loading the same key over and over.
+
+    Variants:
+    - storage_keys_pre_set: The key is pre-set to a non-zero value.
+    """
+    contract_storage = Storage()
+    if storage_keys_pre_set:
+        contract_storage[0] = 1
```
jochem-brouwer:
```suggestion
        contract_storage[1] = 1
```

Suggestion to change this test in an SLOAD loop (the value loaded will point to the next slot, which is the current slot which just got loaded)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,491 @@ def test_sstore_erc20_approve(
...
+
+    benchmark_test(
+        target_opcode=Op.SLOAD,
+        code_generator=JumpLoopGenerator(
+            attack_block=Op.POP(Op.SLOAD(Op.PUSH1[0])),
```
jochem-brouwer:
```suggestion
            attack_block=Op.SLOAD,
```

Note, initially a value should be pushed: if pre-set then push 1, else push 0. This initializes the stack to point to the storage slot to load.

jochem-brouwer (commented):
Some suggestions which can be tackled in a new PR. This should be fixed though: https://github.com/ethereum/execution-specs/pull/2130/changes#r2761584158

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: I think that would be a bit hard, the best approach would be using this [feature](https://github.com/ethereum/execution-specs/pull/2030) once merged.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: I will just use some random number as example:

Assume the total gas limit is 300M, and there are 5 contracts (see `num_contracts` parametrization) involved in the benchmark, each contract would be allocated 60M gas for execution.

Let's say, if 60M total gas limit would be split into 4 txs (16M * 3 + 12M), the `iteration_counts` number calculate the iteration for each transaction. So the original 4 txs (16M, 16M, 16M, 12M) might convert to (1k, 1k, 1k, 0.8K) iterations.

And we later dynamically construct the transaction based on the iteration sets.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: It is used for calculating the total gas used for sanity check, please see `gas_used` variable

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: This is a good idea!

LouisTsai-Csie (approved):
I apply some changes here. LGTM

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
+        )
+        iteration_cost = contract_loop.gas_cost(fork) * iteration_count
+        return overhead_gas + iteration_cost
+
+    def calc_gas_required(
+        iteration_count: int, start_slot: int, contract_addr: Address
+    ) -> int:
+        gsc = fork.gas_costs()
+        # SSTORE requires a minimum gas of G_CALL_STIPEND to operate.
+        # TODO: Correct fix is to introduce bytecode.gas_required.
```
CPerezz, line 578: Leftover?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
+                high = mid
+            else:
+                low = mid + 1
+
+        iteration_count = low - 1
+        iteration_counts.append(iteration_count)
+        start_slot += iteration_count
+        remaining_gas -= calc_gas_required(
+            iteration_count, start_slot, Address(0)
+        )
```
CPerezz, line 614: The binary search finds the iteration count using the old start_slot, but the gas is subtracted using the new
  start_slot. Since calc_gas_required depends on start_slot through get_calldata (different bytes = different intrinsic     gas costs), the gas accounting silently drifts. Over multiple loop iterations this accumulates and can cause
  remaining_gas to go negative (Python integers don't overflow — they just go negative and the while remaining_gas > 0      silently exits).

```suggestion
  iteration_count = low - 1
  iteration_counts.append(iteration_count)                                                                                  
  remaining_gas -= calc_gas_required(iteration_count, start_slot, Address(0))
  start_slot += iteration_count
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
```
CPerezz, line 823: ditto on the slot missmatch
```suggestion
  iteration_count = low - 1
  iteration_counts.append(iteration_count)                                                                                  
  remaining_gas -= calc_gas_required(iteration_count, start_slot, Address(0))
  start_slot += iteration_count       
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
+        pytest.param(0xDEADBEEF, 0, id="nonzero_to_zero"),
+        pytest.param(0xDEADBEEF, 0xBEEFBEEF, id="nonzero_to_diff"),
+        pytest.param(0xDEADBEEF, 0xDEADBEEF, id="nonzero_to_same"),
+    ],
+)
+def test_sstore_variants(
+    benchmark_test: BenchmarkTestFiller,
+    fork: Fork,
+    pre: Alloc,
+    tx_gas_limit: int,
```
CPerezz, line 498: This is shadowed later. So it's unused. Thus, can either be removed from here, or should not be shadowed below

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
+@pytest.mark.parametrize("warm_slots", [False, True])
+@pytest.mark.parametrize("storage_keys_pre_set", [False, True])
+def test_storage_sload_benchmark(
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+    warm_slots: bool,
+    storage_keys_pre_set: bool,
+    tx_gas_limit: int,
```
CPerezz, line 739: ditto on the shadowing

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
+        for i in range(slots_per_contract):
+            expected_storage[i] = write_value
+
+        post[contract_addr] = Account(
+            code=contract,
+            storage=expected_storage,
+        )
+
+    benchmark_test(
+        blocks=[Block(txs=txs)],
```
CPerezz, line 677: where did the `pre` go??


```suggestion
        pre=pre
        blocks=[Block(txs=txs)],
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
+            min(remaining_gas, gas_limit_cap)
+            if gas_limit_cap is not None
+            else remaining_gas
+        )
+        if calc_gas_required(0, start_slot, Address(0)) > gas_limit:
+            break
+
+        # Binary search the optimal number of iterations given the gas limit
+        low, high = 1, 2
+        while calc_gas_required(high, start_slot, Address(0)) <= gas_limit:
```
CPerezz, line 808: Won't computing with a 0's address imply different gas costs than when we have non-0 values? Maybe I'm missing something.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
+
+    # Loop Condition: Counter < Num Slots
+    loop += Op.DUP2  # [end_slot, counter + 1, end_slot]
+    loop += Op.DUP2  # [counter + 1, end_slot, counter + 1, end_slot]
+    loop += Op.LT  # [counter + 1 < end_slot, counter + 1, end_slot]
+    loop += Op.ISZERO
+    loop += Op.ISZERO
+    loop += Op.PUSH1(len(setup))
+    loop += Op.JUMPI
+    # [counter + 1, value, end_slot]
```
CPerezz, line 722: The SLOAD helper stack only has [counter, end_slot] — there is no value element. 
```suggestion
   # [counter + 1, end_slot].
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -394,3 +401,492 @@ def test_sstore_erc20_approve(
...
+        )
+        txs.append(tx)
+
+        start_slot += iteration_count
+
+    benchmark_test(
+        pre=pre,
+        blocks=[Block(txs=txs)],
+        expected_benchmark_gas_used=gas_used,
+    )
```
CPerezz, line 867: Why do we have no Post-state validation here? We do have it inside the other test. Is it too complex to add some kind of validity check?

CPerezz (commented):
A few things where I'm not sure about stuff

PR #2131, opened 2026-02-04
fix(test-fill): use k-way merge for fixture json to reduce memory footprint

fselmo:
## 🗒️ Description

- Ameliorate memory usage on merge, though this seems unrelated to current benchmark OOM issues.
- Lower number of workers for benchmark release and log `--durations=50` so we can see how long things are taking in the last ~300 tests or so when we get a SIGTERM on these feature releases.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

🐼

### DIFF
--- a/.github/configs/feature.yaml
+++ b/.github/configs/feature.yaml
@@ -9,7 +9,7 @@ develop:
 
 benchmark:
   evm-type: benchmark
-  fill-params: --no-html --fork=Osaka --gas-benchmark-values 1,5,10,30,60,100,150 -m benchmark ./tests/benchmark
+  fill-params: --no-html --fork=Osaka --gas-benchmark-values 1,5,10,30,60,100,150 -m benchmark ./tests/benchmark --maxprocesses=30 --durations=50
 
 benchmark_fast:
   evm-type: benchmark
--- a/packages/testing/src/execution_testing/fixtures/collector.py
+++ b/packages/testing/src/execution_testing/fixtures/collector.py
@@ -3,6 +3,7 @@ Fixture collector class used to collect, sort and combine the different types
 of generated fixtures.
 """
 
+import heapq
 import json
 import os
 import re
@@ -13,6 +14,7 @@ from typing import (
     IO,
     ClassVar,
     Dict,
+    Generator,
     List,
     Literal,
     Optional,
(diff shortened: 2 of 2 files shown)


.github/configs/feature.yaml
```diff
@@ -9,7 +9,7 @@ develop:
 
 benchmark:
   evm-type: benchmark
-  fill-params: --no-html --fork=Osaka --gas-benchmark-values 1,5,10,30,60,100,150 -m benchmark ./tests/benchmark
+  fill-params: --no-html --fork=Osaka --gas-benchmark-values 1,5,10,30,60,100,150 -m benchmark ./tests/benchmark -n 30 --durations=50
```
danceratopz: I would prefer to leave `-n 30` out and test the k-way merge without this addition, but we can always revert this after getting a release out.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -100,8 +100,7 @@ def _termination_handler(signum: int, frame: Any) -> None:
         if signum == signal.SIGTERM:
             raise SystemExit(128 + signum)
         raise KeyboardInterrupt
-    # Subsequent interrupts: ignore and print message
-    print("\nMerging fixtures, please wait...", flush=True)
+    # Subsequent interrupts: silently ignore
```
danceratopz: I found this message helpful in the CI log 😆 we should add more logging.

danceratopz (approved):
This looks like a very good candidate to fix the current issues. Keen to get this merged!

Two small comments, but feel free to leave as-is. About `-n 30`: Once we've confirmed this solves the issue, we can always try again without this reduction in workers afterwards.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
fselmo: put it back 👍🏼

.github/configs/feature.yaml
fselmo: I used `--maxprocesses=30` instead in case the worker ever changes we don't force `-n 30` for every machine. I think this is the only change in this PR that may help since I don't think we ever even get to the merge part at the moment.

@danceratopz let me know if you're good merging this.

PR #2138, opened 2026-02-04
feat(tests): add warm and cold account query benchmark

LouisTsai-Csie:
## 🗒️ Description

Add missing configurations for `CALL`, `CALLCODE`, `STATICCALL`, `DELEGATECALL`, `EXTCODEHASH`, `EXTCODESIZE`, `EXTCODECOPY`, `BALANCE`.

| Opcode         | Variants / Dimensions                                                                 |
|----------------|----------------------------------------------------------------------------------------|
| STATICCALL     | warm / cold; memory size                                                               |
| CALL           | warm / cold; memory size; value / no-value                                             |
| CALLCODE       | warm / cold; memory size                                                               |
| DELEGATECALL   | warm / cold; memory size                                                               |
| EXTCODEHASH   | warm / cold                                                                            |
| EXTCODECOPY   | warm / cold; code size; memory size                                                    |
| EXTCODESIZE   | warm / cold                                                                            |
| BALANCE       | warm / cold                                                                            |
| SLOAD          | warm / cold; no-slot / slot                                                            |
| SSTORE         | warm / cold; no-slot / slot; same-value / new-value                                   |

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/compute/instruction/test_account_query.py
+++ b/tests/benchmark/compute/instruction/test_account_query.py
@@ -17,21 +17,26 @@ from typing import Any
 
 import pytest
 from execution_testing import (
+    AccessList,
     Account,
     Address,
     Alloc,
     BenchmarkTestFiller,
     Block,
     Bytecode,
+    Create2PreimageLayout,
     ExtCallGenerator,
     Fork,
     Hash,
+    IteratingBytecode,
     JumpLoopGenerator,
     Op,
     TestPhaseManager,
     Transaction,
     While,
 )
+
+from tests.benchmark.compute.helpers import CustomSizedContractFactory
 
 
 @pytest.mark.repricing(contract_balance=1)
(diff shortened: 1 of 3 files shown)


tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+@pytest.mark.parametrize(
+    "code_size",
+    [0, 32, 256, 1024, 24576],
+)
+@pytest.mark.parametrize(
+    "mem_size",
+    [0, 32, 256, 1024],
+)
+@pytest.mark.parametrize(
+    "forward_value",
```
jochem-brouwer: Can this be made more descriptive? `value_sent` for instance

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+    ]:
+        kwargs["args_size"] = mem_size
+    if opcode == Op.CALL:
+        kwargs["value"] = forward_value
+
+    benchmark_test(
+        target_opcode=opcode,
+        code_generator=JumpLoopGenerator(
+            setup=Op.MSTORE(0, target_addr),
+            attack_block=Op.POP(opcode(**kwargs)),
```
jochem-brouwer: All opcodes push an item to stack, except EXTCODECOPY, so that should not Op.POP

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+    if opcode not in [Op.CALL] and forward_value > 0:
+        pytest.skip("no value parameter")
+
+    target_addr = (
+        pre.empty_account()
+        if code_size == 0
+        else pre.deploy_contract(code=Op.STOP + Op.JUMPDEST * (code_size - 1))
+    )
+
+    kwargs: dict[str, Any] = {"address": Op.MLOAD(0)}
```
jochem-brouwer: I think this should be `Op.PUSH20(target_addr)`

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+) -> None:
+    """
+    Test running a block with as many stateful opcodes doing warm access
+    for an account.
+    """
+    if (
+        opcode not in [Op.EXTCODECOPY, Op.EXTCODEHASH, Op.EXTCODESIZE]
+        and code_size > 0
+    ):
+        pytest.skip("code size not affected")
```
jochem-brouwer: This skips Op.CALL / Op.CALLCODE / Op.STATICCALL / Op.DELEGATECALL where code_size > 0. Why is this?

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+)
+@pytest.mark.parametrize(
+    "mem_size",
+    [0, 32, 256, 1024],
+)
+@pytest.mark.parametrize(
+    "forward_value",
+    [0, 1],
+)
+def test_account_query_warm(
```
jochem-brouwer: This test queries the same account multiple times (so after first query it is warm). We should also add a test where the accounts are marked as warm in the access list and query all these different accounts (like the cold test, but now we add them to the access list, so they are warm instead of cold). 
Can do in a new PR.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+    forward_value: int,
+    gas_benchmark_value: int,
+    tx_gas_limit: int,
+) -> None:
+    """
+    Test running a block with as many stateful opcodes doing cold access
+    for different accounts.
+    """
+    if opcode not in [Op.EXTCODECOPY, Op.EXTCODEHASH, Op.EXTCODESIZE]:
+        pytest.skip("code size not affected")
```
jochem-brouwer: This skips all tests with this opcode, I think we should test cold account queries here as well

jochem-brouwer: Ah, I see, this creates empty accounts and then queries those (which is fine). We also want to add a test to query accounts with code in a cold way (so we can also perform these code operations on top of it)

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+    )
+    max_creations_per_tx = int((tx_gas_limit * 0.9) // account_creation_gas)
+
+    factory_code = (
+        Op.CALLDATALOAD(0)  # addr_start
+        + Op.PUSH4(max_creations_per_tx)  # counter
+        + While(
+            body=Op.POP(
+                Op.CALL(
+                    address=Op.ADD(addr_offset, Op.ADD(Op.DUP7, Op.DUP7)),
```
jochem-brouwer: What does stack item 7 hold? (What's the layout of the created addresses?)

jochem-brouwer (commented):
Some feedback, have some extra test items we should add. I can pick up tomorrow to add CREATE2 factories to help with querying and adding the extra tests mentioned.

tests/benchmark/compute/instruction/test_account_query.py
CPerezz: I was going to say the exact same! Nice catch!

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+    )
+
+
+@pytest.mark.repricing
+@pytest.mark.parametrize(
+    "opcode",
+    [
+        Op.BALANCE,
+        Op.EXTCODESIZE,
+        Op.EXTCODEHASH,
```
CPerezz, line 416: Missing `EXTCODECOPY`. Any reason for it?

tests/benchmark/compute/instruction/test_account_query.py
CPerezz: Same question here

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +395,242 @@ def test_ext_account_query_cold(
...
+    for an account.
+    """
+    if (
+        opcode not in [Op.EXTCODECOPY, Op.EXTCODEHASH, Op.EXTCODESIZE]
+        and code_size > 0
+    ):
+        pytest.skip("code size not affected")
+
+    if opcode not in [Op.CALL] and forward_value > 0:
+        pytest.skip("no value parameter")
```
CPerezz: `CALLCODE` also accepts a value parameter (same signature as `CALL`). In the cold test, the combination of `codesize > 0` of earlier + this skip means the `forward_value` parameter doubles the test matrix but produces zero additional executed tests.

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: Nice idea!

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: Good catch! will handle it

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: This comes from Maria's request, please take a look at the table in the PR description.

There is no `code_size` configuration for `CALL*` opcodes, so here i skip the combination, where opcode is `CALL*` and the target account code size > 0.

does it make sense to you?

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: I will add this in the new commit.

tests/benchmark/compute/instruction/test_account_query.py
misilva73: The gas cost of `EXTCODECOPY` depends on the size of the code being copied and thus I need to see how the run time of these opcodes changes with varying code sizes. 

However, looking at the test implementation, I don't think code size is encoding the size of the code being copied by `EXTCODECOPY`. Can you confirm, @LouisTsai-Csie ?

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +400,205 @@ def test_ext_account_query_cold(
...
+    gas_benchmark_value: int,
+) -> None:
+    """Benchmark scenario of accessing max-code size bytecode."""
+    if (
+        opcode in (Op.EXTCODESIZE, Op.EXTCODEHASH, Op.BALANCE)
+        and mem_size != 0
+    ):
+        pytest.skip(f"No memory size configuration for {opcode}")
+
+    if opcode in (Op.CALL, Op.CALLCODE) and value_sent > 0:
```
spencer-tb: Should this be inverted here?
```suggestion
    if opcode not in (Op.CALL, Op.CALLCODE) and value_sent > 0:
```

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +400,205 @@ def test_ext_account_query_cold(
...
+        Op.EXTCODESIZE,
+        Op.EXTCODEHASH,
+        Op.EXTCODECOPY,
+    ],
+)
+@pytest.mark.parametrize("access_warm", [True, False])
+@pytest.mark.parametrize("mem_size", [0, 32, 256, 1024])
+@pytest.mark.parametrize("code_size", [0, 32, 256, 1024, 24576])
+@pytest.mark.parametrize("value_sent", [0, 1])
+def test_account_query(
```
spencer-tb, line 426: This produces around 310 tests with the skips (640 without). Just stating here if we wanted to reduce the parameterization a little here. If the latter is necessary then we obviously shouldn't reduce it.

spencer-tb: Do we want to add the repricing marker here too?

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +400,205 @@ def test_ext_account_query_cold(
...
+            AccessList(
+                address=custom_sized_contract_factory.created_contract_address(
+                    fork=fork, salt=i
+                ),
+                storage_keys=[],
+            )
+            for i in range(start_iteration, start_iteration + iteration_count)
+        ]
+
+    attack_address = pre.deploy_contract(code=attack_code, balance=10**21)
```
marioevz, line 557: Can we make this line a bit more precise? Perhaps `num_contracts * value_sent`.

marioevz (commented):
LGTM, just one comment.

marioevz:
@LouisTsai-Csie I've rebased and fixed all issues with the tests, but did _not_ address any of the comments in this PR, and I do agree with @spencer-tb that the amount of tests this adds is a bit concerning.
The tests now also run much faster due to the rebase to forks/amsterdam, which includes a speedup in the max-size contract factory.

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: @misilva73 does this cover your scenario?

```python
if opcode == Op.EXTCODECOPY:
    attack_call = Op.EXTCODECOPY(
        address=create2_preimage.address_op(),
        dest_offset=0,
        size=mem_size, # copy code size
        # Gas accounting
        data_size=mem_size,
        address_warm=access_warm,
    )
```

misilva73: Yes, it looks correct now

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: We refactored with a new helper, i believe this would be resolved.

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: Add `EXTCODECOPY` support in new commits!

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: I add the `value_sent` support for `CALLCODE`.

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: Asking Kamil for repricing marker. For the test combination, i think the updated version is the final configuration.

LouisTsai-Csie:
@marioevz @spencer-tb I review the latest change and do some updates to match the requested configurations, it is ready for review now.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -395,3 +400,212 @@ def test_ext_account_query_cold(
...
+    pre: Alloc,
+    fork: Fork,
+    opcode: Op,
+    access_warm: bool,
+    mem_size: int,
+    code_size: int,
+    value_sent: int,
+    gas_benchmark_value: int,
+) -> None:
+    """Benchmark scenario of accessing max-code size bytecode."""
```
spencer-tb, line 437: More skips?
```suggestion
    if opcode in (Op.EXTCODESIZE, Op.EXTCODEHASH, Op.BALANCE) and code_size != 0:                                  
      pytest.skip(f"No code size configuration for {opcode}") 
```
These opcodes only read account metadata, not bytecode. Should save 24 more tests. I think we drop to 94 variants after this.

spencer-tb (approved):
LGTM! Small suggestion for more skips if it makes sense.

tests/benchmark/compute/instruction/test_account_query.py
LouisTsai-Csie: Good catch!

LouisTsai-Csie: I combine other condition like this:

```python
if (
        opcode in (Op.EXTCODESIZE, Op.EXTCODEHASH, Op.BALANCE)
        and (mem_size != 0 or code_size!=0)
    ):
        pytest.skip(f"No memory size configuration for {opcode}")
```

PR #2139, opened 2026-02-04
feat(testing/fill): Account-hash-based deploy addresses

marioevz:
#### TL;DR

Reduce account duplication during pre-alloc grouping by making contract addresses a function of the account’s contents.

#### Details

Normally, account collisions are undesirable because once a test touches an account, its nonce increases, its storage may be modified, and its balance can change.

This is particularly relevant during test execution, where we deploy fresh contracts on a live network to start each test from the most deterministic state possible.

However, this constraint does not apply to tests that are filled and then consumed, since test execution always starts from a clean state.

In the case of Engine X tests, we always perform a forkchoice-update back to genesis. As a result, it does not matter if the same contract is used by multiple tests, because reverting to genesis rolls back all state changes. We can therefore safely reuse the same account across tests, provided it has the same initial state.

This PR modifies the behavior of `Pre` to calculate a contract’s address as a hash of the account’s contents, including its initial balance, nonce, storage, and code.

We also add a salt to the calculated hash to support cases where a single test requires multiple identical copies of the same account in the state.

#### Same contract, two different tests:

```python
def test_1(pre: Alloc) -> None:
    contract_test_1 = pre.deploy_contract(code=Op.STOP)

def test_2(pre: Alloc) -> None:
    contract_test_2 = pre.deploy_contract(code=Op.STOP)
```
##### Outcomes
- `contract_test_1 == contract_test_2`

#### Same contract, two different tests, multiple times:

```python
def test_1(pre: Alloc) -> None:
    contract_test_1_a = pre.deploy_contract(code=Op.STOP)
    contract_test_1_b = pre.deploy_contract(code=Op.STOP)

def test_2(pre: Alloc) -> None:
    contract_test_2_a = pre.deploy_contract(code=Op.STOP)
    contract_test_2_a = pre.deploy_contract(code=Op.STOP)
```
##### Outcomes
- `contract_test_1_a == contract_test_2_a`
- `contract_test_1_b == contract_test_2_b`
- `contract_test_1_a != contract_test_1_b`
- `contract_test_2_a != contract_test_2_b`

#### Same contract, two different tests, different properties:

```python
def test_1(pre: Alloc) -> None:
    contract_test_1 = pre.deploy_contract(code=Op.STOP, storage={0: 1})

def test_2(pre: Alloc) -> None:
    contract_test_2 = pre.deploy_contract(code=Op.STOP, storage={0: 2})
```
##### Outcomes
- `contract_test_1 != contract_test_2`

#### EOA, no arguments, two different tests:

```python
def test_1(pre: Alloc) -> None:
    sender_1 = pre.fund_eoa()

def test_2(pre: Alloc) -> None:
    sender_2 = pre.fund_eoa()
```
##### Outcomes
- `sender_1 == sender_2`

#### EOA, same fund amount, two different tests:

```python
def test_1(pre: Alloc) -> None:
    sender_1 = pre.fund_eoa(amount=0)

def test_2(pre: Alloc) -> None:
    sender_2 = pre.fund_eoa(amount=0)
```
##### Outcomes
- `sender_1 == sender_2`

#### EOA, different fund amount, two different tests:

```python
def test_1(pre: Alloc) -> None:
    sender_1 = pre.fund_eoa(amount=0)

def test_2(pre: Alloc) -> None:
    sender_2 = pre.fund_eoa(amount=1)
```
##### Outcomes
- `sender_1 != sender_2`

#### TL;DR

Automatically detect changes in the `pre` that would result in conflicts with other tests and put the test into a different group without having to rely on `pytest.mark.pre_alloc_group`.

#### Details

Keeps track of the addresses that were modified in the following ways:

- `pre[address] = Account(...)`
- `pre.fund_address(...)`
- `pre.deploy_contract(..., address=0x1234..., ...)`
- `del pre[address]`

All of these actions allow the test to directly modify an account at an specific address' contents, which would affect other tests.

We keep track of the addresses that were modified this way and hash the contents of each account to add it to the pre-alloc group hashing.

Reasoning behind `pre.fund_address(...)` in particular is that most tests expect other addresses (such as system addresses or precompiles, or even created contracts) to start with zero balance. This is different from specifying a non-zero balance in the `pre.deploy_contract` because this modifies the resulting address due to the hashing, which makes the collision impossible.

All usages of `pytest.mark.pre_alloc_group` have been removed from tests and pre-alloc grouping test execution of all Amsterdam test did not result in any issues. The marker itself was not removed just to retain the ability to group tests manually.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/base_types/composite_types.py
+++ b/packages/testing/src/execution_testing/base_types/composite_types.py
@@ -533,7 +533,11 @@ class Account(CamelModel):
     def hash(self) -> Hash:
         """Return the hash of the account given its properties."""
         data = self.model_dump(mode="json")
-        blob = json.dumps(data, sort_keys=True).encode("utf-8")
+        blob = json.dumps(
+            data,
+            sort_keys=True,
+            separators=(",", ":"),
+        ).encode("utf-8")
         return Hash(hashlib.sha256(blob).digest())
 
     @classmethod
--- a/packages/testing/src/execution_testing/base_types/composite_types.py
+++ b/packages/testing/src/execution_testing/base_types/composite_types.py
@@ -1,5 +1,7 @@
 """Base composite types for Ethereum test cases."""
 
+import hashlib
+import json
 from dataclasses import dataclass
 from typing import (
     Any,
@@ -346,6 +348,15 @@ class Storage(
         )
 
 
+class FrozenStorage(Storage):
+    """Frozen storage."""
+
+    model_config = {
+        **Storage.model_config,
+        "frozen": True,
+    }
+
+
 class Account(CamelModel):
     """State associated with an address."""
 
@@ -366,6 +377,11 @@ class Account(CamelModel):
     Sentinel object used to specify when an account should not exist in the
     state.
     """
+
+    model_config = {
+        **CamelModel.model_config,
+        "frozen": True,
+    }
 
     @dataclass(kw_only=True)
     class NonceMismatchError(Exception):
@@ -513,6 +529,12 @@ class Account(CamelModel):
     def __bool__(self: "Account") -> bool:
         """Return True on a non-empty account."""
         return any((self.nonce, self.balance, self.code, self.storage))
+
+    def hash(self) -> Hash:
+        """Return the hash of the account given its properties."""
+        data = self.model_dump(mode="json")
+        blob = json.dumps(data, sort_keys=True).encode("utf-8")
+        return Hash(hashlib.sha256(blob).digest())
 
     @classmethod
     def with_code(cls: Type, code: BytesConvertible) -> "Account":
--- a/packages/testing/src/execution_testing/base_types/composite_types.py
+++ b/packages/testing/src/execution_testing/base_types/composite_types.py
@@ -348,15 +348,6 @@ class Storage(
         )
 
 
-class FrozenStorage(Storage):
-    """Frozen storage."""
-
-    model_config = {
-        **Storage.model_config,
-        "frozen": True,
-    }
-
-
 class Account(CamelModel):
     """State associated with an address."""
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -23,7 +23,6 @@ from execution_testing.base_types import (
 )
 from execution_testing.base_types.conversions import (
     BytesConvertible,
-    FixedSizeBytesConvertible,
     NumberConvertible,
 )
 from execution_testing.forks import Fork
@@ -39,10 +38,11 @@ from execution_testing.test_types import (
     TransactionTestMetadata,
     compute_deterministic_create2_address,
 )
-from execution_testing.test_types import Alloc as BaseAlloc
 from execution_testing.tools import Initcode
 from execution_testing.vm import Bytecode, Op
 
+from ..shared.pre_alloc import Alloc as BaseAlloc
+from ..shared.pre_alloc import AllocFlags
 from .contracts import (
     check_deterministic_factory_deployment,
     deploy_deterministic_factory_contract,
@@ -228,7 +228,6 @@ class PendingTransaction(Transaction):
 class Alloc(BaseAlloc):
     """A custom class that inherits from the original Alloc class."""
 
-    _fork: Fork = PrivateAttr()
     _sender: EOA = PrivateAttr()
     _eth_rpc: EthRPC = PrivateAttr()
     _pending_txs: List[PendingTransaction] = PrivateAttr(default_factory=list)
@@ -243,7 +242,6 @@ class Alloc(BaseAlloc):
     def __init__(
         self,
         *args: Any,
-        fork: Fork,
         sender: EOA,
         eth_rpc: EthRPC,
         eoa_iterator: Iterator[EOA],
@@ -254,23 +252,12 @@ class Alloc(BaseAlloc):
     ) -> None:
         """Initialize the pre-alloc with the given parameters."""
         super().__init__(*args, **kwargs)
-        self._fork = fork
         self._sender = sender
         self._eth_rpc = eth_rpc
         self._eoa_iterator = eoa_iterator
         self._chain_id = chain_id
         self._node_id = node_id
         self._address_stubs = address_stubs or AddressStubs(root={})
-
-    def __setitem__(
-        self,
-        address: Address | FixedSizeBytesConvertible,
-        account: Account | None,
-    ) -> None:
-        """Set account associated with an address."""
-        raise ValueError(
-            "Tests are not allowed to set pre-alloc items in execute mode"
-        )
 
     def code_pre_processor(self, code: Bytecode) -> Bytecode:
         """Pre-processes the code before setting it."""
@@ -302,18 +289,18 @@ class Alloc(BaseAlloc):
         self._pending_txs.append(pending_tx)
         return pending_tx
 
-    def deterministic_deploy_contract(
+    def _deterministic_deploy_contract(
         self,
         *,
         deploy_code: BytesConvertible,
-        salt: Hash | int = 0,
-        initcode: BytesConvertible | None = None,
-        storage: Storage | StorageRootType | None = None,
-        label: str | None = None,
+        salt: Hash | int,
+        initcode: BytesConvertible | None,
+        storage: Storage | StorageRootType | None,
+        label: str | None,
     ) -> Address:
         """
-        Deploy a contract to the allocation at a deterministic location
-        using a deterministic deployment proxy.
+        Execute implementation of contract deployment to a deterministic
+        location.
         """
         del storage
         gas_costs = self._fork.gas_costs()
@@ -409,7 +396,7 @@ class Alloc(BaseAlloc):
 
         balance = self._eth_rpc.get_balance(contract_address)
         nonce = self._eth_rpc.get_transaction_count(contract_address)
-        super().__setitem__(
+        self.__internal_setitem__(
             contract_address,
             Account(
                 nonce=nonce,
@@ -422,18 +409,18 @@ class Alloc(BaseAlloc):
         contract_address.label = label
         return contract_address
 
-    def deploy_contract(
+    def _deploy_contract(
         self,
         code: BytesConvertible,
         *,
-        storage: Storage | StorageRootType | None = None,
-        balance: NumberConvertible = 0,
-        nonce: NumberConvertible = 1,
-        address: Address | None = None,
-        label: str | None = None,
-        stub: str | None = None,
+        storage: Storage | StorageRootType | None,
+        balance: NumberConvertible,
+        nonce: NumberConvertible,
+        address: Address | None,
+        label: str | None,
+        stub: str | None,
     ) -> Address:
-        """Deploy a contract to the allocation."""
+        """Execute implementation of contract deployment."""
         if storage is None:
             storage = {}
         assert address is None, "address parameter is not supported"
@@ -471,7 +458,7 @@ class Alloc(BaseAlloc):
                 f"Stub contract {contract_address}: balance={bal_eth:.18f} "
                 f"ETH, nonce={nonce}, code_size={len(code)} bytes"
             )
-            super().__setitem__(
+            self.__internal_setitem__(
                 contract_address,
                 Account(
                     nonce=nonce,
@@ -559,7 +546,7 @@ class Alloc(BaseAlloc):
             "impossible to deploy contract with nonce lower than one"
         )
 
-        super().__setitem__(
+        self.__internal_setitem__(
             contract_address,
             Account(
                 nonce=nonce,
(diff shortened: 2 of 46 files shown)


danceratopz:
> In the case of Engine X tests, we always perform a forkchoice-update back to genesis. As a result, it does not matter if the same contract is used by multiple tests, because reverting to genesis rolls back all state changes. We can therefore safely reuse the same account across tests, provided it has the same initial state.

This is not strictly necessary and is currently skipped in https://github.com/ethereum/execution-specs/pull/1964 see https://github.com/danceratopz/execution-specs/blob/6fe382c0370ce94a935f65c89eac1073bc6324b6/packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/simulator_logic/test_via_engine.py#L43-L76.

We can certainly benchmark to see how much time is saved by avoiding the initial FCU!

But the initial plan was to skip it unless it was considered necessary for intermittent garbage collection (e.g. benchmarking efforts hinted that this would be necessary for Besu). A follow-up PR was to add `--enginex-initial-fcu-freq` to enable this; as-is it is entirely skipped.

Edit: We don't FCU post-genesis, but the effect is the same (as we don't FCU after the test either) - we're always building on top of genesis in engine x :)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -352,20 +403,13 @@ def fund_address(
         If the address is already present in the pre-alloc the amount will be
         added to its existing balance.
```
spencer-tb: Docstring update suggestion
```suggestion
        Add a funded account to the pre-allocation.

        The address must not already exist in the pre-allocation. To set the
        balance of an account, use the `amount` parameter in `fund_eoa()` or
        the `balance` parameter in `deploy_contract()` at creation time.
```

packages/testing/src/execution_testing/base_types/composite_types.py
```diff
@@ -346,6 +348,15 @@ def canary(self) -> "Storage":
         )
 
 
+class FrozenStorage(Storage):
+    """Frozen storage."""
```
spencer-tb: Assuming will be used later.

spencer-tb (commented):
Smart optimization. Assuming this will save memory during phase 1 and intial stage of phase 2 filling. The salt mechanism is cool.

packages/testing/src/execution_testing/base_types/composite_types.py
marioevz: Removed as it ended up not being necessary 👍

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
marioevz: Added, thanks!

marioevz:
@danceratopz @spencer-tb I added another very nice enhancement that virtually makes the `pytest.mark.pre_alloc_group` marker obsolete (🎉 ), please take a look and let me know what you think!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -253,24 +250,13 @@ def __init__(
...
         self._address_stubs = address_stubs or AddressStubs(root={})
 
-    def __setitem__(
-        self,
-        address: Address | FixedSizeBytesConvertible,
-        account: Account | None,
-    ) -> None:
-        """Set account associated with an address."""
-        raise ValueError(
-            "Tests are not allowed to set pre-alloc items in execute mode"
```
marioevz, line 273: To reviewers: This check has been moved to `SharedAlloc` in the form of flags (see `alloc_flags` fixture for the skipping logic).

tests/berlin/eip2930_access_list/test_acl.py
```diff
@@ -233,7 +233,6 @@ def test_transaction_intrinsic_gas_cost(
     )
     sender = pre.fund_eoa()
     tx_value = 1
-    pre.fund_address(sender, tx_value)
```
marioevz, line 236: Note to reviewers: `execute` command handles this automatically since #1822 and for fill I think we should simply increment the default value if for some reason we come up with a test that requires more balance.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -50,97 +49,112 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+            buffer += b"\0"
+            for altered_account in sorted(altered_accounts):
+                buffer += altered_account
+                account = self[altered_account]
+                assert account is not None
+                buffer += account.hash()
+        if self._deleted_addresses:
+            buffer += b"\1"
+            for deleted_address in sorted(self._deleted_addresses):
+                buffer += deleted_address
```
marioevz, line 115: Note to reviewers: We could even completely ban this behavior and make this change a bit cleaner. In cases where we "remove" system contracts from the state we don't really remove them, we instead do `pre[SYSTEM_CONTRACT_ADDRESS] = Account(<empty account>)` to overwrite it.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -776,46 +760,33 @@ def fund_address(
         if address in self:
             account = self[address]
             if account is not None:
-                account.balance = ZeroPaddedHexNumber(new_balance)
+                self[address] = account.copy(balance=new_balance)
```
fselmo: Clarifying q... doesn't this also need to use `__internal_set_item__(address, ...)` in order to actually run through the RPC funding logic?

It doesn't seem easy, but do you think it's possible to write basic, sanity checking unit tests to make sure these execute mode expectations don't break? It would make reviews easier to have confidence in imo as it's not so straightforward / quick to test execute remote things as a reviewer.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -741,9 +725,9 @@ def fund_address(
                 if address in self:
                     account = self[address]
                     if account is not None:
-                        account.balance = ZeroPaddedHexNumber(current_balance)
+                        self[address] = account.copy(balance=current_balance)
```
fselmo: Same thought here about using `__internal_setitem__()`

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -3996,9 +3982,8 @@ def test_authorization_reusing_nonce(
     [True, False],
 )
 @pytest.mark.exception_test
-@pytest.mark.execute(
-    pytest.mark.skip(reason="Requires contract-eoa address collision")
-)
+@pytest.mark.pre_alloc_mutable
+@pytest.mark.pre_alloc_mutable
```
fselmo: Double marked 👀

```suggestion
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/pre_alloc.py
```diff
@@ -0,0 +1,329 @@
...
+        balance: NumberConvertible = 0,
+        nonce: NumberConvertible = 1,
+        address: Address | None = None,
+        label: str | None = None,
+        stub: str | None = None,
+    ) -> Address:
+        """
+        Deploy a contract to the allocation.
+
+        Warning: `address` parameter is a temporary solution to allow tests to
```
fselmo, line 184: Stray thought... I actually only now realized this is supposed to be temporary. It looks like only one test uses this and it makes sense. Would this be a good time to change this to something like `_address` (private attr... do not use unless you know what you're doing kind of thing?). I feel like it may be easy for a test that uses this to sneak by if it's a public property on `pre.deploy_contract()`.

packages/testing/src/execution_testing/base_types/composite_types.py
```diff
@@ -367,6 +369,11 @@ class Account(CamelModel):
     state.
     """
 
+    model_config = {
+        **CamelModel.model_config,
+        "frozen": True,
```
fselmo, line 374: 👌🏼

fselmo (commented):
This is a lot to review but I think it looks good. Really clever change that makes a lot of sense 🔥.

It's really difficult to know though, imo, if this will create any friction with `execute remote`. I think this is "less crucial" of a way to test (and more rnd) and we can certainly fix any quirks that bubble up _after_ this PR is in... but I do wonder if there aren't some basic sanity check unit tests we can write for some of these methods that guarantee we are not introducing bugs to this functionality. I think this would be pretty difficult though... curious if others have any ideas here.

I asked Claude to look at this and I think there's a potential bug with setting `self[address]` instead of using `__internal_set_item__(address, ...)`. I couldn't reproduce it in kurtosis bc I think it'd only be present in some very unique scenario... but asking here to make sure.

Everything else looks pretty good to me. This will be very nice to get in 👍🏼

packages/testing/src/execution_testing/base_types/composite_types.py
fselmo: Nothing to add here. Just a thought that since this is frozen I think we could probably cache the `hash` in case we ever end up calling it more than once we don't end up hashing it multiple times. I'm pretty sure the happy path in this PR only calls it once though so consider this just a nit.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/pre_alloc.py
marioevz: Agreed, I feel we need to wind that option down, as we move towards more control of where the contracts land in our tests.
I feel though that is out of scope because it's just too big now.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Fixed this in latest commit. I think we can expand our unit testing in execute, but perhaps out of scope for this PR.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Fixed, thanks!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/pre_alloc.py
fselmo: Sounds good 👍🏼

fselmo (approved):
This lgtm, great change 🚀

marioevz:
### Group Stats

#### Normal Tests

`--fork Amsterdam`

##### Before

```
Pre-Allocation Statistics Summary
Total groups: 238
Total tests: 13839
Total accounts: 61220
```

##### After

```
Pre-Allocation Statistics Summary
Total groups: 280
Total tests: 13839
Total accounts: 23768
```

#### Static Tests

`--fork Amsterdam --fill-static-tests ./tests/static`

##### Before

```
Pre-Allocation Statistics Summary
Total groups: 9394
Total tests: 13075
Total accounts: 169430
```

##### After

```
Pre-Allocation Statistics Summary
Total groups: 2181
Total tests: 13075
Total accounts: 22450
```

#### Benchmark Tests

`--fork Osaka --gas-benchmark-values 1,5,10,30 -m benchmark ./tests/benchmark`

##### Before

```
Pre-Allocation Statistics Summary
Total groups: 17
Total tests: 4171
Total accounts: 29767
```

##### After

```
Pre-Allocation Statistics Summary
Total groups: 7
Total tests: 4171
Total accounts: 9112
```

spencer-tb:
> ### Group Stats
> #### Normal Tests
> 
> `--fork Amsterdam`
> ##### Before
> 
> ```
> Pre-Allocation Statistics Summary
> Total groups: 238
> Total tests: 13839
> Total accounts: 61220
> ```
> 
> ##### After
> 
> ```
> Pre-Allocation Statistics Summary
> Total groups: 280
> Total tests: 13839
> Total accounts: 23768
> ```
> 
> #### Static Tests
> 
> `--fork Amsterdam --fill-static-tests ./tests/static`
> ##### Before
> 
> ```
> Pre-Allocation Statistics Summary
> Total groups: 9394
> Total tests: 13075
> Total accounts: 169430
> ```
> 
> ##### After
> 
> ```
> Pre-Allocation Statistics Summary
> Total groups: 2181
> Total tests: 13075
> Total accounts: 22450
> ```
> 
> #### Benchmark Tests
> 
> `--fork Osaka --gas-benchmark-values 1,5,10,30 -m benchmark ./tests/benchmark`
> ##### Before
> 
> ```
> Pre-Allocation Statistics Summary
> Total groups: 17
> Total tests: 4171
> Total accounts: 29767
> ```
> 
> ##### After
> 
> ```
> Pre-Allocation Statistics Summary
> Total groups: 7
> Total tests: 4171
> Total accounts: 9112
> ```

This is awesome! 🔥

PR #2141, opened 2026-02-05
fix(benchmark): reduce stateful stubs to 4 representative tokens

CPerezz:
## Summary
- Reduce `stubs.json` from 99 ERC20 tokens (297 entries) to 4 representative tokens (12 entries)
- Kept tokens: **30GB_ERC20** (largest), **XEN** (2nd), **USDC** (4th), **IMT** (smallest/99th)
- Original 99-token stubs preserved as `stubs_99.json` for reference


tests/benchmark/stateful/bloatnet/stubs.json
```diff
@@ -1,299 +1,14 @@
 {
   "test_sload_empty_erc20_balanceof_30GB_ERC20": "0x19fc17d87D946BBA47ca276f7b06Ee5737c4679C",
   "test_sload_empty_erc20_balanceof_XEN": "0x06450dEe7FD2Fb8E39061434BAbCFC05599a6Fb8",
```
jochem-brouwer, line 3: For Maria's analysis, the ERC20 target which is almost empty is for instance this one: 0x7c4d13a8e743b036f6ba71c5dcadfe4fc9aa7a17. It has 1 holder and I think 5-6 storage slots occupied (so near-empty). I think for analysis it would be great to add this address in these stubs.

jochem-brouwer: (I searched etherscan for "empty" tokens, it delivered 😆 )

jochem-brouwer: I will also use these contracts for the storage ERC20 targets 😄 👍

LouisTsai-Csie (approved):
LGTM

PR #2146, opened 2026-02-05
chore(ci): use official rustup installer instead of apt

spencer-tb:
## 🗒️ Description

Fix CI failures caused by apt proxy timeout when installing rustup.

### Problem

```
Err:1 http://archive.ubuntu.com/ubuntu noble/universe amd64 rustup amd64 1.26.0-5build1
  Could not connect to 10.128.2.134:3142 (10.128.2.134), connection timed out
```

### Solution

Use `dtolnay/rust-toolchain` GitHub Action (pinned SHA) instead of `apt-get install rustup`:
- More secure than `curl | sh`
- Avoids apt proxy issues
- Widely used (1.4k+ stars)

Also fixes:
- Missing `sudo` on tox dependencies install  
- Removes deprecated `--force-yes` flag

## 🔗 Related Issues or PRs

All PRs in EELS are currently blocked due to this CI failure.

### DIFF
--- a/.github/actions/setup-env/action.yaml
+++ b/.github/actions/setup-env/action.yaml
@@ -4,15 +4,18 @@ runs:
   using: "composite"
   steps:
     - name: Install Rust
+      uses: dtolnay/rust-toolchain@888c2e1ea69ab0d4330cbf0af1ecc7b68f368cc1
+      with:
+        toolchain: 1.89.0
+
+    - name: Install build-essential
       shell: bash
-      run: |
-        sudo DEBIAN_FRONTEND=noninteractive apt-get install --yes --force-yes build-essential rustup
-        rustup update --no-self-update 1.89.0 && rustup default 1.89.0
+      run: sudo DEBIAN_FRONTEND=noninteractive apt-get install --yes build-essential
 
     - name: Install Tox and any other packages
       shell: bash
       run: |
-        DEBIAN_FRONTEND=noninteractive apt-get install --yes --force-yes build-essential pkg-config
+        sudo DEBIAN_FRONTEND=noninteractive apt-get install --yes build-essential pkg-config
         pip install 'tox>=4.11,<5' requests
 
     - name: Download Geth and add to $PATH


.github/actions/setup-env/action.yaml
```diff
@@ -6,13 +6,14 @@ runs:
     - name: Install Rust
       shell: bash
       run: |
-        sudo DEBIAN_FRONTEND=noninteractive apt-get install --yes --force-yes build-essential rustup
-        rustup update --no-self-update 1.89.0 && rustup default 1.89.0
+        sudo DEBIAN_FRONTEND=noninteractive apt-get install --yes build-essential
+        curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain 1.89.0
```
marioevz: Can check in a copy of this script in the repo? Just so that we are not vulnerable to DNS hijack attacks.

spencer-tb: I updated to use this action: https://github.com/dtolnay/rust-toolchain

Prefer it?

spencer-tb: Needs to be added here if so but I don't have perms.
```
Error: The action dtolnay/rust-toolchain@888c2e1ea69ab0d4330cbf0af1ecc7b68f368cc1 is not allowed in ethereum/execution-specs because all actions must be from a repository owned by ethereum or match one of the patterns: JamesIves/github-pages-deploy-action@*, actions-rs/toolchain@*, actions/cache/restore@*, actions/cache@*, actions/checkout@*, actions/deploy-pages@*, actions/download-artifact@*, actions/github-script@*, actions/setup-go@*, actions/setup-python@*, actions/upload-artifact@*, actions/upload-pages-artifact@*, astral-sh/setup-uv@*, codecov/codecov-action@*, davidanson/markdownlint-cli2-action@*, jwlawson/actions-setup-cmake@*, softprops/action-gh-release@*, zgosalvez/github-actions-ensure-sha-pinned-actions@*. All actions must also be pinned to a full-length commit SHA.
```

SamWilsn (changes requested):
How does this solve the problem if we're still using apt at all?

This seems like solving an infrastructure problem at the wrong level.

.github/actions/setup-env/action.yaml
marioevz: Added to the repo allowed actions 👍

fselmo:
Yeah this feels like a bandaid. We're also experiencing issues in the release runs on any `apt` call.

marioevz:
Closing as the apt has been apparently resolved, we can re-open if necessary.

spencer-tb:
Nice! Yeah was a bandaid solution so we could merge PRs. Glad its fixed

PR #2150, opened 2026-02-05
feat(test): performance and fill updates; print -> debug log session finish

fselmo:
## 🗒️ Description

- Cache gas spent for recently problematic test
- Silence prints for fill as debug logs, turn this on for releases so we can debug issues better
- Increase `--durations` to `100` for releases

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2026-02-05 at 13 30 11

### DIFF
--- a/.github/actions/build-fixtures/action.yaml
+++ b/.github/actions/build-fixtures/action.yaml
@@ -39,9 +39,9 @@ runs:
       shell: bash
       run: |
         if [ "${{ steps.evm-builder.outputs.impl }}" = "eels" ]; then
-          uv run fill -n ${{ steps.evm-builder.outputs.x-dist }} ${{ steps.properties.outputs.fill-params }} --output=fixtures_${{ inputs.release_name }}.tar.gz --build-name ${{ inputs.release_name }}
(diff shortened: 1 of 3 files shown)


packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -592,23 +592,39 @@ def gas_cost_by_iteration_count(
...
+                int, tuple[int, int, int, int, int]
+            ] = {}
+        if cache_key not in self._gas_cost_cache:
+            self._gas_cost_cache[cache_key] = (
+                self.setup.gas_cost(fork=fork),
+                self.iterating.gas_cost(fork=fork),
+                self.warm_iterating.gas_cost(fork=fork),
+                self.iterating_subcall_gas_cost(fork=fork),
+                self.cleanup.gas_cost(fork=fork),
+            )
```
marioevz: I feel like an `@cache` from `functool` might fare better here than trying to create a new structure. My main criticism is that this tuple is unnamed and disjointed from the values it contains, it feels like it's easy to access one field instead of another by mistake.

We can also defer the changes from this file out of this PR since the main issue was the max-size-contract-creating class specifically, not `IteratingBytecode` in general.

marioevz (commented):
Thanks, just one comment.

packages/testing/src/execution_testing/tools/tools_code/generators.py
fselmo: I dropped the commit.

marioevz (approved):
LGTM, thanks!

Issue #2125, opened 2026-02-03
feat(testing): Stateful Engine X Format and Fill Transition Tool


marioevz:
Original idea from this comment: https://github.com/ethpandaops/gas-lighting-tracker/issues/34#issuecomment-3841169400

### New `StatefulBlockchainEngineFixture`

Similar to `BlockchainEngineXFixture` where all tests use the same genesis pre-allocation but with the main difference is that the pre-allocation is not included in a JSON file, rather it is a pointer to a stateful snapshot, which is the snapshot that should be used to initialize the client running the tests.

### New `StatefulTransitionTool`

Since the stateful snapshot is extremely big, it will not be possible to use common t8n tools, because these ingest the full pre-allocation via a JSON file, and since this file would be too big to process, it will be inefficient and maybe impossible.

Proposed solution is to create a new `TransitionTool` subclass that uses a fully instantiated client to perform the state transition.

#### Instantiation

The `start_server` method of this new class would reach out to the running client to verify it's running the correct snapshot, and that the client supports the [`testing_buildBlockV1` endpoint](https://github.com/ethereum/execution-apis/pull/747).

#### State transition evaluation

The `evaluate` method of this class would:
- Make sure the client is currently at the desired head for the current request
- Format the input and then perform a `testing_buildBlockV1` call to the client to build a block with the list of transactions it requires to be included.
- Record the output and transform it to the expected response.

#### Parallelism

Since we only have one instance of the client, and the client would need to rollback to a previous state via forkchoice update, it might not be possible to use xdist when filling these tests, but it's an open question.

### Modified Fill Process

Not exhaustive list of requirements:

#### New flag to activate mode

Perhaps we can add a new flag to `fill` to enable this mode, `--stateful`, which requires some other flags to be passed, e.g. the RPC address of the full client running the snapshot, and also a flag that contains the reference to the snapshot (the state root hash perhaps?)

### Other Considerations

Since we are relying on a pre-existing snapshot, we cannot modify the storage pre-allocation (perhaps we can? and produce a snapshot-delta?) the test filler must be modified slightly to do the `pre.deploy_contract`, `pre.fund_eoa`, etc., in blocks on top of the snapshot, and then run the test blocks on top of these "setup blocks". 

We also can take special care of using stubs during filling too in order to replace `pre.deploy_contract` with contracts we are sure exist in the snapshot.

marioevz:
cc @jochem-brouwer @CPerezz

[...]

PR #2169, opened 2026-02-09
feat(test-fixtures): add BlockchainEngineStatefulFixture format

danceratopz:
## 🗒️ Description

Add `BlockchainEngineStatefulFixture` format for snapshot-based stateful Engine API testing.

- Reference an external snapshot instead of embedding genesis state.
- Include setup payloads for deploying contracts on top of the snapshot.
- Support partial state verification via `post_state_diff`.
- Auto-register for deserialization without adding to fill.

## 🔗 Related Issues or PRs

Part of #2125.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture
image
Source: [commons.wikimedia.org](https://commons.wikimedia.org/wiki/File:Capivara(Hydrochoerus_hydrochaeris).jpg)


packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -833,6 +833,36 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
...
+    snapshot_id: str
+    snapshot_block_number: HexNumber
+    snapshot_block_hash: Hash
+
+    setup_payloads: List[FixtureEngineNewPayload] = Field(
+        ..., alias="setupEngineNewPayloads"
+    )
+    payloads: List[FixtureEngineNewPayload] = Field(
+        ..., alias="engineNewPayloads"
+    )
```
danceratopz, line 866: This matches the current two-level granularity of block type that test implementers can specify in the `BenchmarkTest` format. We could consider optional finer-grained/more descriptive labels with fallback to a reasonable default.

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -833,6 +833,36 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
...
+
+    format_name: ClassVar[str] = "blockchain_test_stateful_engine"
+    description: ClassVar[str] = (
+        "Tests that generate a Blockchain Test fixture for "
+        "snapshot-based stateful Engine API testing."
+    )
+
+    snapshot_id: str
+    snapshot_block_number: HexNumber
+    snapshot_block_hash: Hash
```
marioevz, line 859: Do we need a `snapshot_state_root: Hash` too?

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -833,6 +833,36 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
...
+
+    setup_payloads: List[FixtureEngineNewPayload] = Field(
+        ..., alias="setupEngineNewPayloads"
+    )
+    payloads: List[FixtureEngineNewPayload] = Field(
```
marioevz, line 864: Engine X also includes `post_state_diff: Alloc | None = None`, which might be worth considering, but not sure how useful in the context of benchmarking where we likely will only check the last state root and the timings, and then proceed to the next test.

marioevz (approved):
LGTM, just a couple of comments, thanks!

packages/testing/src/execution_testing/fixtures/blockchain.py
danceratopz: I had it originally, but then I also thought that it's enough to validate the payloads during consumption. I'm still implementing the new tooling, will think on it.

jochem-brouwer: Note that in the near future the block level access (BAL) lists will be part of the chain spec. The BALs could be provided as metadata (only the BAL hash is committed) to be used as post state diff.

Note that for benchmarking, the BAL is necessary, as the BAL unlocks optimizations like parallel transaction execution at the protocol level. Not providing the BAL where all transactions depend on the state of previous transactions are therefore forced to execute all transactions sequentially. So for benchmark tests, providing the BAL (as metadata, or as engine API input as specified per EIP, at least a way to provide the contents) is thus necessary to allow clients to fully use these optimizations and thus giving representative results.

(A somewhat general comment, since current fixture outputs which also have post state could benefit from this. However, one has to merge BALs if the test consist of multiple blocks to get the actual post state, so cannot directly use it as a quick lookup table to see what the final state of the chain should look like)

packages/testing/src/execution_testing/fixtures/blockchain.py
jochem-brouwer: The state root of the block is contained in the block header, and is therefore part of the blockhash. However I could see the benefit of also including the state root here. It's about thinking of the "bigger picture" and also taking some ideas of the current fixture format including how these are filled. 

The motivation for these stateful tests, at least from my perspective, comes from state-related research which focuses on questions like impact of the total state size on the chain, how to manage this (state rent? state expiry?) or other research like switching to a new trie structure (like binary tries). Obviously for a "full test" we care about the entire block(chain) but for research we mostly care about the impact of state as the main focus point.

Also want to raise another challenge which is to provide chains this state to run these tests on. One could imagine that stateful tests which are built on a snapshot of a devnet might not be easy to run mid-term, since the snapshot of this devnet might not be available (the devnet is offline). The same happens with specific snapshots of for instance the mainnet chain.

We are working on a deterministic tool to generate state: https://github.com/nerolation/state-actor and use this to bootstrap all clients (for now this only boots Geth at genesis) to startup the chain at this state, without the need of reading all state from genesis. To do so, the current draft idea is to boot Geth as a seeding snap sync node and then launching other clients to sync the chain (the state) to this state.

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -833,6 +833,36 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
...
+
+    model_config = CamelModel.model_config | {"extra": "ignore"}
+
+    format_name: ClassVar[str] = "blockchain_test_stateful_engine"
+    description: ClassVar[str] = (
+        "Tests that generate a Blockchain Test fixture for "
+        "snapshot-based stateful Engine API testing."
+    )
+
+    snapshot_id: str
```
jochem-brouwer: What is this ID? Is it a handle or identifier for this fixture/test?

packages/testing/src/execution_testing/fixtures/blockchain.py
jochem-brouwer: I see, so `setup_payloads` are payloads marked as the `TestPhase.SETUP` and the `payloads` those of `TestPhase.EXECUTION`?

Here is something what I believe should then change: currently these `TestPhase` are phases at the transaction level. This itself, at test level, does not make sense because it is thus possible to create a block (payload) which has both SETUP and EXECUTION transactions in it.

In practice, the EXECUTION transactions (blocks) is currently the final block of the test. It is obviously possible that in future benchmarks / tests the ones we find interesting could span multiple blocks. It could also be possible that the payloads we find interesting are not consecutive payloads. The current layout implies that, when executing this fixture, first the setup payloads and then the execution payload should be executed. There could also be benchmarks where reorgs are tested: root -> A - B - C -> rollback to A - D - E. Here A/B/D are setup blocks and C/E are the ones we are interested in (execution/testing phase).

I think that the payloads should be an ordered list, and we could provide this metadata (is it setup or execution) either as part of the payload metadata or as part of metadata of the fixture. This also allows us to annotate more phases in the future (for instance a "cleanup" phase)

jochem-brouwer (commented):
Great work! I left some initial thoughts here, it is somewhat lengthy at it is also a mini brain-dump. I'm mainly leaving these comments to think and frame this test format in the bigger picture and keeping it flexible.

Let me know if something is unclear 😄 👍

marioevz:
I don't think this affects anything in this PR but just FYI: https://github.com/ethereum/execution-specs/issues/2125#issuecomment-3879068274

marioevz:
> I don't think this affects anything in this PR but just FYI: [#2125 (comment)](https://github.com/ethereum/execution-specs/issues/2125#issuecomment-3879068274)

@danceratopz yes it does, we have to mark the fixture format as having two phases, like Engine X.

packages/testing/src/execution_testing/fixtures/blockchain.py
marioevz: Yeah, might be unnecessary, I think we originally thought of it like a way you could reference the snapshot somehow, but the block-hash/state-root should suffice. It's up to the test runner to set this up for us, both before filling or consuming the generated fixtures.

packages/testing/src/execution_testing/fixtures/blockchain.py
marioevz: Let's leave it IMO, we can remove it if it proves to be unnecessary.

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -833,6 +833,36 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
...
+    of the snapshot before the actual test payloads execute.
+    """
+
+    model_config = CamelModel.model_config | {"extra": "ignore"}
+
+    format_name: ClassVar[str] = "blockchain_test_stateful_engine"
+    description: ClassVar[str] = (
+        "Tests that generate a Blockchain Test fixture for "
+        "snapshot-based stateful Engine API testing."
+    )
```
marioevz, line 852:
```suggestion
    )
    format_phases: ClassVar[Set[FixtureFillingPhase]] = {
        FixtureFillingPhase.FILL,
        FixtureFillingPhase.PRE_ALLOC_GENERATION,
    }
```
Add the pre-alloc generation phase here because we intend for this to be a multi-stage filling. We can swap `PRE_ALLOC_GENERATION` for some new phase if we find it useful.

packages/testing/src/execution_testing/fixtures/blockchain.py
marioevz:
```suggestion
```
Removing to get merged.

packages/testing/src/execution_testing/fixtures/blockchain.py
marioevz: This will be added to other fixture formats too, so we can cross that bridge for all formats when we get there.

packages/testing/src/execution_testing/fixtures/tests/test_base.py
```diff
@@ -44,6 +60,88 @@ def test_json_dict() -> None:
             ),
             id="TransactionFixture",
         ),
+        pytest.param(
+            BlockchainEngineStatefulFixture(
+                fork=Prague,
+                last_block_hash=Hash(1),
+                post_state_hash=Hash(2),
+                config=FixtureConfig(fork=Prague),
+                snapshot_id="snapshot-1",
```
marioevz:
```suggestion
```

packages/testing/src/execution_testing/fixtures/tests/test_blockchain.py
```diff
@@ -985,6 +987,160 @@
...
             id="fixture_engine_new_payload_2",
         ),
+        pytest.param(
+            True,
+            BlockchainEngineStatefulFixture(
+                fork=Prague,
+                last_block_hash=Hash(1),
+                post_state_hash=Hash(2),
+                config=FixtureConfig(fork=Prague),
+                snapshot_id="snapshot-1",
```
marioevz:
```suggestion
```

packages/testing/src/execution_testing/fixtures/tests/test_blockchain.py
```diff
@@ -985,6 +987,160 @@
...
+            {
+                "_info": {},
+                "network": "Prague",
+                "postStateHash": Hash(2).hex(),
+                "lastblockhash": Hash(1).hex(),
+                "config": {
+                    "network": "Prague",
+                    "chainid": "0x01",
+                },
+                "snapshotId": "snapshot-1",
```
marioevz:
```suggestion
```

PR #2171, opened 2026-02-09
feat(test-benchmark): support eip-2780 benchmark

LouisTsai-Csie:
## 🗒️ Description

**ether transfer benchmark** (`test_ether_transfers`)
- receiver account is empty / existing account
- receiver is warm / cold access (via access list)
- the transfer amount is zero / non-zero
- the transfer is self transfer 

**ether transfer to precompile** (test_ether_transfers_to_precompile)
- zero / non-zero transfer

**max code contract creation transaction** (test_contract_creation)
- zero / non-zero ether transfer

## 🔗 Related Issues or PRs

None

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Patrick

### DIFF
--- a/tests/benchmark/compute/scenario/test_transaction_types.py
+++ b/tests/benchmark/compute/scenario/test_transaction_types.py
@@ -18,7 +18,10 @@ from execution_testing import (
     Hash,
     Op,
     Transaction,
+    compute_create_address,
 )
+
+from tests.benchmark.compute.helpers import CustomSizedContractInitcode
 
 
 def test_empty_block(
@@ -29,21 +32,6 @@ def test_empty_block(
         blocks=[Block(txs=[])],
         expected_benchmark_gas_used=0,
     )
-
-
-@pytest.fixture
-def iteration_count(intrinsic_cost: int, gas_benchmark_value: int) -> int:
-    """
-    Calculate the number of iterations based on the gas limit and intrinsic
-    cost.
-    """
-    return gas_benchmark_value // intrinsic_cost
-
-
-@pytest.fixture
-def transfer_amount() -> int:
-    """Ether to transfer in each transaction."""
-    return 1
 
 
 @pytest.fixture
@@ -85,9 +73,10 @@ def get_single_receiver_list(
 
 @pytest.fixture
 def ether_transfer_case(
-    case_id: str, pre: Alloc, balance: int
+    case_id: str, pre: Alloc, empty_account: bool
 ) -> Tuple[Generator[Address, None, None], Generator[Address, None, None]]:
     """Generate sender and receiver generators based on the test case."""
+    balance = 0 if empty_account else 1
     if case_id == "a_to_a":
         """Sending to self."""
         senders = get_single_sender_list(pre)
@@ -129,15 +118,18 @@ def ether_transfer_case(
         "diff_acc_to_diff_acc",
     ],
 )
-@pytest.mark.parametrize("balance", [0, 1])
-def test_block_full_of_ether_transfers(
+@pytest.mark.parametrize("transfer_amount", [0, 1])
+@pytest.mark.parametrize("empty_account", [True, False])
+@pytest.mark.parametrize("warm_access", [False, True])
+def test_ether_transfers(
     benchmark_test: BenchmarkTestFiller,
     pre: Alloc,
     case_id: str,
-    balance: int,
-    iteration_count: int,
+    empty_account: bool,
     transfer_amount: int,
-    intrinsic_cost: int,
+    fork: Fork,
+    gas_benchmark_value: int,
+    warm_access: bool,
     ether_transfer_case: Tuple[
         Generator[Address, None, None], Generator[Address, None, None]
     ],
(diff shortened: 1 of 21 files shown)


misilva73:
Here is a quick assessment of the covered cases for EIP-2780.

### Transfers to EOAs

✅ All cases covered

### Transfers to contracts

We also want to test eth transfers to contracts. This is important for the the code vs no-code distinction. We need to following configs:

- receiver is warm / cold access (via access list)
- the transfer amount is zero / non-zero

I am not seeing these test cases. Let me know if there is another file that contains it. 

###  EIP-7702 set code transactions

The current test sends value=0. We should also test:

- Auth tx with value > 0 to an existing account 
- Auth tx with value > 0 to a new account 

### Transfers to previously-delegated EIP-7702 EOAs 

We also want to test eth transfers to accounts that were previously delegated to. We need to following configs:

- receiver is warm / cold access (via access list)
- the transfer amount is zero / non-zero

I cannot find a test case for these.  Let me know if there is another file that contains it. 

An open question here is whether we care about the delegate target's bytecode? Something like:

| Scenario | Delegate target | Description |
|----------|----------------|----------------|
| Many EOAs → 1 small contract | Single, minimal code | Best case / baseline |
| Many EOAs → 1 max-size contract | Single, 24KB code | Tests caching benefit with large code |
| Many EOAs → N distinct max-size contracts | Each unique, 24KB | **Adversarial worst case** — no caching, max I/O per call |
| Many EOAs → N distinct small contracts | Each unique, minimal code | Isolates the per-account overhead from the code-loading overhead |

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -18,8 +18,11 @@
     Hash,
     Op,
     Transaction,
+    compute_create_address,
 )
 
+from tests.benchmark.compute.helpers import CustomSizedContractInitcode
```
marioevz:
```suggestion
from ..helpers import CustomSizedContractInitcode
```
Just so that folder restructuring does not break this test. We'd have to add `tests/benchmark/compute/scenario/__init__.py` and `tests/benchmark/compute/__init__.py` for this to work though.

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -511,3 +554,56 @@ def test_auth_transaction(
...
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    transfer_amount: int,
+    gas_benchmark_value: int,
+) -> None:
+    """Benchmark max code size contract creation."""
+    initcode = CustomSizedContractInitcode(
+        pre=pre, fork=fork, contract_size=fork.max_code_size()
+    )
```
marioevz: I feel this test should focus on the maximum amount of contract-creating transactions with a minimum sized initcode (`Op.STOP` or even empty bytes) instead of a minimum amount of contract creating transactions with a maximum sized contract.

Perhaps we can do multiple flavors:
- Small initcode, no bytecode deployed
- Small initcode, small bytecode deployed
- Small initcode, max bytecode deployed (all zeros perhaps)

I'd argue that initcode size does not play a role in here, because we are testing placing a contract into state from a transaction, and making a big initcode eats into the gas allowance unnecessarily IMO.

marioevz (commented):
Just two comments, thanks!

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -151,40 +143,91 @@ def test_block_full_of_ether_transfers(
...
+
     txs = []
     token_transfers: dict[Address, int] = {}
+
+    access_list = (
+        [AccessList(address=Address(0x100), storage_keys=[])]
+        if warm_access
+        else None
+    )
+    iteration_cost = fork.transaction_intrinsic_cost_calculator()(
```
jochem-brouwer: I think this access list is defined here to calculate the intrinsic gas costs, right? It reads a bit confusing. I'd suggest putting this access_list directly into the cost calculator:


```suggestion
    iteration_cost = fork.transaction_intrinsic_cost_calculator()(
        # Use 0x100 as dummy entry for the access list item
        access_list=[AccessList(address=Address(0x100), storage_keys=[])]
        if warm_access
        else [],
    )
```

tests/benchmark/compute/scenario/test_transaction_types.py
jochem-brouwer: Good point! We can subdivide this even more:

For the initcode: empty bytes (empty bytes or `Op.STOP` will execute the same, however `Op.STOP` costs one byte of initcode and is therefore paid for in the calldata, which is not strictly necessary). 
For the max bytecode deployed, use `Op.RETURN(Op.PUSH0, size_to_deploy)` which is the smallest size initcode to do so (cheapest in terms of calldata cost). Note that memory is initialized to zeros, so even if the memory has not yet been allocated (via MSTOREs for instance) it will default to zeros.

jochem-brouwer (commented):
Great work, some minor suggestions 😄

tests/benchmark/compute/scenario/test_transaction_types.py
LouisTsai-Csie: I add one commit that add 3 `__init__.py` file, and update the import for all the benchmarks.

tests/benchmark/compute/scenario/test_transaction_types.py
LouisTsai-Csie: Updatd!

tests/benchmark/compute/scenario/test_transaction_types.py
LouisTsai-Csie: Please correct me if this does not resolve the issue:

```python
@pytest.mark.parametrize("contract_size", [0, 1, None])
# if contract_size is None, set to max code size

initcode = Op.RETURN(
  Op.PUSH0,
  contract_size,
  # gas accounting
  old_memory_size=0,
  new_memory_size=contract_size,
  code_deposit_size=contract_size,
)
```

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -511,3 +551,66 @@ def test_auth_transaction(
         blocks=[Block(txs=txs)],
         expected_benchmark_gas_used=total_gas_used - total_refund,
     )
+
+
+@pytest.mark.parametrize("transfer_amount", [0, 1])
+@pytest.mark.parametrize("contract_size", [0, 1, None])
```
jochem-brouwer: These parameters are fine, but the test name will look weird (something like contract_size_None). In that case it should be something like contract_size_max

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -511,3 +551,66 @@ def test_auth_transaction(
...
+def test_contract_creation(
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    transfer_amount: int,
+    gas_benchmark_value: int,
+    contract_size: int | None,
+) -> None:
+    """Benchmark max code size contract creation."""
```
jochem-brouwer: This description is not correct anymore, something like "Benchmark contract creations via transactions" or something like that will do.

jochem-brouwer (commented):
Some minor comments. I also see you changed a lot of imports from `tests.benchmark.compute.helpers` to `..helpers`, is this intended? 😄

LouisTsai-Csie:
> EIP-7702 set code transactions
The current test sends value=0. We should also test:
Auth tx with value > 0 to an existing account
Auth tx with value > 0 to a new account


IIUC, in a type-4 transaction we have a `to` field and an authorization tuple. Does the “account” here refer to the `to` account?

misilva73:
> > EIP-7702 set code transactions
> > The current test sends value=0. We should also test:
> > Auth tx with value > 0 to an existing account
> > Auth tx with value > 0 to a new account
> 
> IIUC, in a type-4 transaction we have a `to` field and an authorization tuple. Does the “account” here refer to the `to` account?

@LouisTsai-Csie, yes, “account” here refers to the `to`

tests/benchmark/compute/scenario/test_transaction_types.py
LouisTsai-Csie: Nice catch, updated!

tests/benchmark/compute/scenario/test_transaction_types.py
LouisTsai-Csie: Done, thanks!

LouisTsai-Csie:
HiHi @jochem-brouwer , thanks for the feedback, the updated import is intended, which applies the changes based on this [comment](https://github.com/ethereum/execution-specs/pull/2171#discussion_r2783532416).

For the 7702 benchmark, i would create a new PR cc @misilva73

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -431,13 +507,17 @@ def test_block_full_access_list_and_data(
 
 @pytest.mark.parametrize("empty_authority", [True, False])
 @pytest.mark.parametrize("zero_delegation", [True, False])
+@pytest.mark.parametrize("empty_account", [True, False])
+@pytest.mark.parametrize("transfer_amount", [True, False])
 def test_auth_transaction(
```
marioevz, line 512: This test is currently breaking for me if we add the max-gas-limit check from #2203. I'll try to fix in this same PR and merge.

marioevz (approved):
LGTM, rebased and fixed `test_auth_transaction` so it never goes above the benchmark gas limit.

Issue #1940, opened 2025-12-18
EIP-2780 Implementation Tracker: Reduce intrinsic transaction gas


marioevz:
### Target Fork

**Amsterdam**

### Instructions

- [x] Assign issue to EIP specification and testing owner(s).

> [!IMPORTANT]
> A specifications specialist and a testing specialist should ideally share ownership of the EIP.

- [x] Add the issue to the target fork milestone if applicable (i.e., the EIP is at least in the [CFI stage](https://eips.ethereum.org/EIPS/eip-7723#considered-for-inclusion)).

#### Guidance for Marking Items Complete

An item should only be checked off once the EIP is considered *stable*. In this context, stable means:

- No major issues or ambiguities are still being uncovered in the specification or tests.
- There are no open discussion points awaiting resolution.
- Client implementations have been consistently passing the tests for at least a week.

It is ultimately up to the owners' discretion to decide when an item should be marked as complete, using this guidance as the basis for that decision.

In exceptional cases, an EIP may require changes after some items have been marked complete or even after the entire issue has been completed and closed. This can happen, for example, when significant design optimizations are identified and agreed upon in ACD, or when critical security issues surface and require updates to the specification or tests.

When this occurs, owners should either unmark the relevant checkboxes if the issue is still open, or create a new tracking issue for the modifications if the original issue had already been closed.

### Specification + Testing Status

- [x] Testing complexity assessed and documented: [EIP-2780.md](https://github.com/ethsteel/pm/blob/main/complexity_assessments/EIPs/EIP-2780.md) - 🟡 13.
- [x] Specification implementation merged to `eips/amsterdam/eip-2780` *(skip if the fork branch merge below is already complete)*.
- [x] Specification updates merged to the corresponding `forks/amsterdam` branch.
- [x] EIP updates proposed in case of architectural choices surfaced during implementation.
- [x] Required testing framework modifications implemented. Fork-aware intrinsic and top-frame calculators are available; see #3527 and #3515.
- [x] Test suite implemented. #2175 and subsequent suites; final coverage additions are in ready-for-review #3515.
- [x] Coverage audit completed in #3515: no missed statements in the seven named EELS functions; test code covers 1,556/1,561 statements, with five unsupported-helper fallback statements documented. This is statement coverage, not a claim of exhaustive branch coverage.
[...]

PR #2175, opened 2026-02-09
feat(specs,tests): Implement EIP-2780

gurukamath:
## 🗒️ Description
Implement EIP-2780 and some basic tests.

Note: This PR contains some testing framework changes which are not compatible with other Amsterdam tests. This PR however, does not focus on fixing those. Those should be fixed in a future PR.
As a result, the CI is not likely to pass for Amsterdam

## 🔗 Related Issues or PRs
issue #1940

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

TBD

### DIFF
--- a/packages/testing/src/execution_testing/forks/forks/forks.py
+++ b/packages/testing/src/execution_testing/forks/forks/forks.py
@@ -2,7 +2,15 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Callable, Dict, List, Mapping, Sized
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Dict,
+    List,
+    Mapping,
+    Optional,
+    Sized,
+)
 
 if TYPE_CHECKING:
     from execution_testing.fixtures.blockchain import FixtureHeader
@@ -21,6 +29,7 @@ from execution_testing.vm import (
     Opcodes,
 )
 
+from ...recipient_type import RecipientType
 from ..base_fork import (
     BaseFeeChangeCalculator,
     BaseFeePerGasCalculator,
@@ -835,8 +844,14 @@ class Frontier(
             access_list: List[AccessList] | None = None,
             authorization_list_or_count: Sized | int | None = None,
             return_cost_deducted_prior_execution: bool = False,
+            sends_value: bool = False,
+            recipient_type: RecipientType = RecipientType.CONTRACT,
+            recipient_is_warm: bool = False,
+            recipient_delegation_is_warm: Optional[bool] = None,
         ) -> int:
             del return_cost_deducted_prior_execution
+            del sends_value, recipient_type, recipient_is_warm
+            del recipient_delegation_is_warm
 
             assert access_list is None, (
                 f"Access list is not supported in {cls.name()}"
--- a/packages/testing/src/execution_testing/forks/gas_costs.py
+++ b/packages/testing/src/execution_testing/forks/gas_costs.py
@@ -146,3 +146,8 @@ class GasCosts:
     OPCODE_BLOBHASH: int = 0
     OPCODE_MCOPY_BASE: int = 0
     OPCODE_CLZ: int = 0
+
+    # EIP-2780: Reduced intrinsic transaction gas costs (Amsterdam)
+    COLD_ACCOUNT_COST_CODE: int = 0
+    COLD_ACCOUNT_COST_NO_CODE: int = 0
+    STATE_UPDATE: int = 0
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -12,10 +12,10 @@ Entry point for the Ethereum specification.
 """
 
 from dataclasses import dataclass
-from typing import List, Optional, Tuple
+from typing import List, Optional, Set, Tuple
 
 from ethereum_rlp import rlp
-from ethereum_types.bytes import Bytes
+from ethereum_types.bytes import Bytes, Bytes0
 from ethereum_types.frozen import slotted_freezable
 from ethereum_types.numeric import U64, U256, Uint
 
@@ -24,12 +24,19 @@ from ethereum.exceptions import (
     EthereumException,
     GasUsedExceedsLimitError,
     InsufficientBalanceError,
+    InsufficientTransactionGasError,
     InvalidBlock,
     InvalidSenderError,
     NonceMismatchError,
 )
 from ethereum.forks.bpo5.blocks import Header as PreviousHeader
-from ethereum.state import EMPTY_CODE_HASH, Address, BlockDiff, PreState
+from ethereum.state import (
+    EMPTY_ACCOUNT,
+    EMPTY_CODE_HASH,
+    Address,
+    BlockDiff,
+    PreState,
+)
 
 from . import vm
 from .block_access_lists import (
@@ -52,7 +59,10 @@ from .exceptions import (
     PriorityFeeGreaterThanMaxFeeError,
     TransactionTypeContractCreationError,
 )
-from .fork_types import Authorization, VersionedHash
+from .fork_types import (
+    Authorization,
+    VersionedHash,
+)
 from .requests import (
     CONSOLIDATION_REQUEST_TYPE,
     DEPOSIT_REQUEST_TYPE,
@@ -77,6 +87,7 @@ from .state_tracker import (
     set_account_balance,
 )
 from .transactions import (
+    AccessListTransaction,
     BlobTransaction,
     FeeMarketTransaction,
     LegacyTransaction,
@@ -93,7 +104,7 @@ from .trie import root, trie_set
 from .utils.hexadecimal import hex_to_address
 from .utils.message import prepare_message
 from .vm import Message
-from .vm.eoa_delegation import is_valid_delegation
+from .vm.eoa_delegation import get_delegated_code_address, is_valid_delegation
 from .vm.gas import (
     GasCosts,
     calculate_blob_gas_price,
@@ -479,7 +490,9 @@ def check_transaction(
     block_output: vm.BlockOutput,
     tx: Transaction,
     tx_state: TransactionState,
-) -> Tuple[Address, Uint, Tuple[VersionedHash, ...], U64]:
+) -> Tuple[
+    Address, Uint, Tuple[VersionedHash, ...], U64, Uint, Uint, Set, Set
+]:
     """
     Check if the transaction is includable in the block.
 
@@ -504,6 +517,14 @@ def check_transaction(
         The blob versioned hashes of the transaction.
     tx_blob_gas_used:
         The blob gas used by the transaction.
+    intrinsic_gas :
+        The total intrinsic gas cost including recipient cost.
+    calldata_floor_gas_cost :
+        The minimum gas cost based on calldata size.
+    access_list_addresses :
+        The set of addresses in the access list.
+    access_list_storage_keys :
+        The set of storage keys in the access list.
 
     Raises
     ------
@@ -538,8 +559,13 @@ def check_transaction(
     EmptyAuthorizationListError :
         If the transaction is a SetCodeTransaction and the authorization list
         is empty.
+    InsufficientTransactionGasError :
+        If the transaction gas is insufficient to cover intrinsic costs.
 
     """
+    # Validate transaction and get base intrinsic costs
+    intrinsic_gas, calldata_floor_gas_cost = validate_transaction(tx)
+
     gas_available = block_env.block_gas_limit - block_output.block_gas_used
     blob_gas_available = MAX_BLOB_GAS_PER_BLOCK - block_output.blob_gas_used
 
@@ -625,11 +651,43 @@ def check_transaction(
     ):
         raise InvalidSenderError("not EOA")
 
+    # Build access list for recipient cost calculation
+    access_list_addresses = set()
+    access_list_storage_keys = set()
+    access_list_addresses.add(block_env.coinbase)
+    if isinstance(
+        tx,
+        (
+            AccessListTransaction,
+            FeeMarketTransaction,
+            BlobTransaction,
+            SetCodeTransaction,
+        ),
+    ):
+        for access in tx.access_list:
+            access_list_addresses.add(access.account)
+            for slot in access.slots:
+                access_list_storage_keys.add((access.account, slot))
+
+    # Calculate recipient gas cost (EIP-2780)
+    recipient_gas_cost = calculate_recipient_gas_cost(
+        tx, tx_state, sender_address, intrinsic_gas, access_list_addresses
+    )
+    intrinsic_gas += recipient_gas_cost
+
+    # Check if transaction has sufficient gas for intrinsic costs
+    if intrinsic_gas > tx.gas or calldata_floor_gas_cost > tx.gas:
+        raise InsufficientTransactionGasError("Insufficient gas")
+
     return (
         sender_address,
         effective_gas_price,
         blob_versioned_hashes,
         tx_blob_gas_used,
+        intrinsic_gas,
+        calldata_floor_gas_cost,
+        access_list_addresses,
+        access_list_storage_keys,
     )
 
 
@@ -918,6 +976,90 @@ def process_general_purpose_requests(
         )
 
 
+def calculate_recipient_gas_cost(
+    tx: Transaction,
+    tx_state: TransactionState,
+    sender: Address,
+    intrinsic_gas: Uint,
+    access_list_addresses: Set[Address],
+) -> Uint:
+    """
+    Calculate the gas cost for accessing and updating the recipient account.
+
+    Post EIP-2780, the cost of accessing and updating the recipient involves
+    state access and is no longer included in TX_BASE. This function
+    computes the recipient-specific costs based on:
+    - Account type (EOA, contract, delegated EOA, or precompile)
+    - Whether the recipient is in the access list (warm vs cold access)
+    - Transaction value (zero vs non-zero transfers)
+    - Account existence (new account creation vs existing account update)
+
+    Parameters
+    ----------
+    tx :
+        The transaction being processed.
+    tx_state :
+        The current transaction state.
+    sender :
+        The address of the transaction sender.
+    intrinsic_gas:
+        The intrinsic gas of a transaction.
+    access_list_addresses :
+        Set of addresses that are pre-warmed via the access list.
+
+    Returns
+    -------
+    recipient_gas_cost : Uint
+        The total gas cost for accessing and updating the recipient account.
+        Returns 0 for contract creation or self-transfers.
+
+    """
+    from .vm.precompiled_contracts.mapping import PRE_COMPILED_CONTRACTS
+
+    tx_recipient_cost = Uint(0)
+    if isinstance(tx.to, Bytes0) or tx.to == sender:
+        return tx_recipient_cost
+
+    if tx.to in PRE_COMPILED_CONTRACTS:
+        if tx.value > U256(0):
+            tx_recipient_cost += GasCosts.STATE_UPDATE
+        return tx_recipient_cost
+
+    is_cold_access = tx.to not in access_list_addresses
+    access_gas_cost = GasCosts.WARM_ACCESS
+    if is_cold_access:
+        access_gas_cost = GasCosts.COLD_ACCOUNT_COST_NO_CODE
+
+    # Make sure the gas is enough to at least pay cold EOA
+    # before actually accessing the account
+    if intrinsic_gas + access_gas_cost > tx.gas:
+        raise InsufficientTransactionGasError("Insufficient gas")
+
+    recipient = get_account(tx_state, tx.to)
+    if recipient.code_hash != EMPTY_CODE_HASH:
+        if is_cold_access:
+            access_gas_cost = GasCosts.COLD_ACCOUNT_COST_CODE
+
+        recipient_code = get_code(tx_state, recipient.code_hash)
+        if is_valid_delegation(recipient_code):
+            delegated_address = get_delegated_code_address(recipient_code)
+            if delegated_address in access_list_addresses:
+                access_gas_cost += GasCosts.WARM_ACCESS
+            else:
+                access_gas_cost += GasCosts.COLD_ACCOUNT_COST_CODE
+
+    tx_recipient_cost += access_gas_cost
+
+    if tx.value > U256(0):
+        # Write cost for the recipient
+        if recipient == EMPTY_ACCOUNT:
+            tx_recipient_cost += GasCosts.NEW_ACCOUNT
+        else:
+            tx_recipient_cost += GasCosts.STATE_UPDATE
+
+    return tx_recipient_cost
+
+
 def process_transaction(
     block_env: vm.BlockEnvironment,
     block_output: vm.BlockOutput,
@@ -959,13 +1101,15 @@ def process_transaction(
         encode_transaction(tx),
     )
 
-    intrinsic_gas, calldata_floor_gas_cost = validate_transaction(tx)
-
     (
         sender,
         effective_gas_price,
         blob_versioned_hashes,
         tx_blob_gas_used,
+        intrinsic_gas,
+        calldata_floor_gas_cost,
+        access_list_addresses,
+        access_list_storage_keys,
     ) = check_transaction(
         block_env=block_env,
         block_output=block_output,
--- a/src/ethereum/forks/amsterdam/vm/gas.py
+++ b/src/ethereum/forks/amsterdam/vm/gas.py
@@ -43,6 +43,22 @@ class GasCosts:
     WARM_ACCESS = Uint(100)
     COLD_ACCOUNT_ACCESS = Uint(2600)
     COLD_STORAGE_ACCESS = Uint(2100)
+    COLD_ACCOUNT_COST_CODE = Uint(2600)
+    """
+    Gas cost for a cold access to an account that has code deployed.
+    [EIP-7702] delegated accounts are also charged this for cold access.
+    Defined in [EIP-2780].
+
+    [EIP-2780]: https://eips.ethereum.org/EIPS/eip-2780
+    [EIP-7702]: https://eips.ethereum.org/EIPS/eip-7702
+    """
+    COLD_ACCOUNT_COST_NO_CODE = Uint(500)
+    """
+    Gas cost for a cold access to an account without code.
+    Defined in [EIP-2780].
+
+    [EIP-2780]: https://eips.ethereum.org/EIPS/eip-2780
+    """
 
     # Storage
     STORAGE_SET = Uint(20000)
@@ -52,6 +68,13 @@ class GasCosts:
     CALL_VALUE = Uint(9000)
     CALL_STIPEND = Uint(2300)
     NEW_ACCOUNT = Uint(25000)
+    STATE_UPDATE = Uint(1000)
+    """
+    Gas cost for a state-changing operation (e.g. value transfer).
+    Defined in [EIP-2780].
+
+    [EIP-2780]: https://eips.ethereum.org/EIPS/eip-2780
+    """
 
     # Contract Creation
     CODE_DEPOSIT_PER_BYTE = Uint(200)
@@ -103,7 +126,12 @@ class GasCosts:
     BLOCK_ACCESS_LIST_ITEM = Uint(2000)
 
     # Transactions
-    TX_BASE = Uint(21000)
+    TX_BASE = Uint(4500)
+    """
+    Base transaction cost, reduced from 21000 by [EIP-2780].
+
+    [EIP-2780]: https://eips.ethereum.org/EIPS/eip-2780
+    """
     TX_CREATE = Uint(32000)
     TX_DATA_TOKEN_STANDARD = Uint(4)
     TX_DATA_TOKEN_FLOOR = Uint(10)
--- a/src/ethereum/forks/amsterdam/vm/instructions/environment.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/environment.py
@@ -14,7 +14,7 @@ Implementations of the EVM environment related instructions.
 from ethereum_types.bytes import Bytes32
 from ethereum_types.numeric import U256, Uint, ulen
 
-from ethereum.state import EMPTY_ACCOUNT
+from ethereum.state import EMPTY_ACCOUNT, EMPTY_CODE_HASH
 from ethereum.utils.numeric import ceil32
 
 from ...state_tracker import get_account, get_code
@@ -27,6 +27,7 @@ from ..gas import (
     calculate_blob_gas_price,
     calculate_gas_extend_memory,
     charge_gas,
+    check_gas,
 )
 from ..stack import pop, push
 
@@ -68,16 +69,28 @@ def balance(evm: Evm) -> None:
     address = to_address_masked(pop(evm.stack))
 
     # GAS
-    if address in evm.accessed_addresses:
-        charge_gas(evm, GasCosts.WARM_ACCESS)
-    else:
+    is_cold_access = address not in evm.accessed_addresses
+    gas_cost = GasCosts.WARM_ACCESS
+    if is_cold_access:
+        gas_cost = GasCosts.COLD_ACCOUNT_COST_NO_CODE
+
+    check_gas(evm, gas_cost)
+
+    # STATE ACCESS
+    tx_state = evm.message.tx_env.state
+    account = get_account(tx_state, address)
+
+    if is_cold_access:
+        if account.code_hash != EMPTY_CODE_HASH:
+            gas_cost = GasCosts.COLD_ACCOUNT_COST_CODE
+            check_gas(evm, gas_cost)
         evm.accessed_addresses.add(address)
-        charge_gas(evm, GasCosts.COLD_ACCOUNT_ACCESS)
+
+    charge_gas(evm, gas_cost)
 
     # OPERATION
     # Non-existent accounts default to EMPTY_ACCOUNT, which has balance 0.
-    tx_state = evm.message.tx_env.state
-    balance = get_account(tx_state, address).balance
+    balance = account.balance
 
     push(evm.stack, balance)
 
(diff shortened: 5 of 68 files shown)


src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -69,6 +69,10 @@
 GAS_BLOBHASH_OPCODE = Uint(3)
 GAS_POINT_EVALUATION = Uint(50000)
 
+GAS_COLD_ACCOUNT_COST_CODE = Uint(2600)
+GAS_COLD_ACCOUNT_COST_NOCODE = Uint(500)
```
SamWilsn: Just guessing at the meaning here, but how about:

```suggestion
GAS_COLD_ACCOUNT_WITH_CODE_COST = Uint(2600)
GAS_COLD_ACCOUNT_WITHOUT_CODE_COST = Uint(500)
```

Or more minimally:


```suggestion
GAS_COLD_ACCOUNT_COST_CODE = Uint(2600)
GAS_COLD_ACCOUNT_COST_NO_CODE = Uint(500)
```

Because nocode isn't a word ;3

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -69,6 +69,10 @@
...
+GAS_STATE_UPDATE = Uint(1000)
```
SamWilsn, line 85: These could use some docstrings before this comes out of draft.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -418,7 +425,9 @@ def check_transaction(
     block_env: vm.BlockEnvironment,
     block_output: vm.BlockOutput,
     tx: Transaction,
-) -> Tuple[Address, Uint, Tuple[VersionedHash, ...], U64]:
+) -> Tuple[
+    Address, Uint, Tuple[VersionedHash, ...], U64, Uint, Uint, Set, Set
+]:
```
SamWilsn, line 497: Probably getting to the size where we'd want named return parameters (i.e. a dataclass).

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -79,17 +80,29 @@ def balance(evm: Evm) -> None:
 
     # GAS
     is_cold_access = address not in evm.accessed_addresses
-    gas_cost = GAS_COLD_ACCOUNT_ACCESS if is_cold_access else GAS_WARM_ACCESS
     if is_cold_access:
+        gas_cost = GAS_COLD_ACCOUNT_COST_NOCODE
+    else:
+        gas_cost = GAS_WARM_ACCESS
+
```
SamWilsn: Should probably backport the one-line conditional -> multiline to old forks.

SamWilsn (commented):
Quick, incomplete review

marioevz:
There's a possibility this becomes a blocker for Devnet-3 since it's listed as requirement for https://eips.ethereum.org/EIPS/eip-8037 which is included.

Spencer checked with Maria, and it seems like it's not a hard requirement but it might be a good idea to get this ready to merge.

gurukamath:
> Spencer checked with Maria, and it seems like it's not a hard requirement but it might be a good idea to get this ready to merge.

To my mind, this needs further clarity, especially when it comes to the EIP-7702 interactions. @jochem-brouwer has nicely summarised some of the open questions in this [eth magicians thread](https://ethereum-magicians.org/t/eip-2780-reduce-intrinsic-cost-of-transactions/4413/10)

gurukamath:
Raised a PR on the EIPs repo to clarify some interactions with EIP-7702 and EIP-7928. See [here](https://github.com/ethereum/EIPs/pull/11332)

We can proceed to write more tests specific to these cases once we agree on these points.

gurukamath:
Converting back to draft so I can re-base this on the latest `forks/amsterdam`

gurukamath:
@danceratopz This got a bit out of hand in terms of PR size. If this is too much, I'd be happy to split this up into the following 2 PRs.

1. Implementation + Framework updates and EIP-specific tests (in this PR)
2. Fix broken legacy tests including BALs and other older fork tests

Do let me know what you think

src/ethereum/forks/amsterdam/vm/gas.py
SamWilsn:
```suggestion
GAS_COLD_ACCOUNT_COST_CODE = Uint(2600)
GAS_COLD_ACCOUNT_COST_EMPTY_CODE = Uint(500)
```

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -69,6 +69,10 @@
 GAS_BLOBHASH_OPCODE = Uint(3)
 GAS_POINT_EVALUATION = Uint(50000)
 
+GAS_COLD_ACCOUNT_COST_CODE = Uint(2600)
```
SamWilsn, line 72: There's kind of an implicit relationship between `GAS_COLD_ACCOUNT_COST_CODE` and `GAS_COLD_ACCOUNT_COST_NOCODE`: the former should always be greater than the latter, right?

Would it make sense to define this one as `GAS_COLD_ACCOUNT_COST_NOCODE + Uint(1100)`?

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -76,8 +78,21 @@ def balance(evm: Evm) -> None:
...
```
SamWilsn, line 84:
```suggestion
    gas_cost = GAS_WARM_ACCESS
    if is_cold_access:
        gas_cost = GAS_COLD_ACCOUNT_COST_NOCODE

```

Could probably be made in a few more places too.

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -343,10 +358,21 @@ def extcodesize(evm: Evm) -> None:
 
     # GAS
     is_cold_access = address not in evm.accessed_addresses
-    access_gas_cost = (
-        GAS_COLD_ACCOUNT_ACCESS if is_cold_access else GAS_WARM_ACCESS
-    )
     if is_cold_access:
+        access_gas_cost = GAS_COLD_ACCOUNT_COST_NOCODE
+    else:
+        access_gas_cost = GAS_WARM_ACCESS
```
SamWilsn:
```suggestion
    access_gas_cost = GAS_WARM_ACCESS
    if is_cold_access:
        access_gas_cost = GAS_COLD_ACCOUNT_COST_NOCODE
        
```

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -389,28 +391,34 @@ def call(evm: Evm) -> None:
...
 
-    create_gas_cost = GAS_NEW_ACCOUNT
-    if value == 0 or is_account_alive(tx_state, to):
-        create_gas_cost = Uint(0)
+    call_value_cost = Uint(0)
+    if value > U256(0):
+        if call_target == EMPTY_ACCOUNT:
+            call_value_cost += GAS_STATE_UPDATE + GAS_NEW_ACCOUNT
+        else:
+            call_value_cost += Uint(2) * GAS_STATE_UPDATE
```
SamWilsn, line 418: This diff is flipping the branching behaviour from the previous fork. Before we'd enter the branch when there are no changes, while now we enter the branch when we do change the account. Also, we're merging `transfer_gas_cost` and `create_gas_cost`.

Do you find that this muddies the changes between amsterdam and osaka, or is this necessary?

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -574,11 +593,43 @@ def check_transaction(
...
+
+    # Calculate recipient gas cost (EIP-2780)
+    recipient_gas_cost = calculate_recipient_gas_cost(
+        tx, tx_state, sender_address, intrinsic_gas, access_list_addresses
+    )
+    intrinsic_gas += recipient_gas_cost
+
+    # Check if transaction has sufficient gas for intrinsic costs
+    if max(intrinsic_gas, calldata_floor_gas_cost) > tx.gas:
+        raise InsufficientTransactionGasError("Insufficient gas")
```
SamWilsn: Is this a more natural way of writing it?

```suggestion
    if intrinsic_gas > tx.gas or calldata_floor_gas_cost > tx.gas:
        raise InsufficientTransactionGasError("Insufficient gas")
```

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -574,11 +593,43 @@ def check_transaction(
...
+    access_list_addresses.add(block_env.coinbase)
+    if isinstance(
+        tx,
+        (
+            AccessListTransaction,
+            FeeMarketTransaction,
+            BlobTransaction,
+            SetCodeTransaction,
+        ),
+    ):
```
SamWilsn, line 668: Should we make a `has_access_list` function?

SamWilsn (commented):
From the spec side, my review is complete.

src/ethereum/forks/amsterdam/fork.py
gurukamath: Perhaps this function should eventually be 

```
def check_transaction(
    block_env: vm.BlockEnvironment,
    block_output: vm.BlockOutput,
    tx: Transaction,
    tx_state: TransactionState,
) -> vm.TransactionEnvironment:
```

i.e build the transaction environment gets built here. What do you think? In any case, I'd like to reserve this change to a later re-factoring PR.

src/ethereum/forks/amsterdam/vm/gas.py
gurukamath: I'd be inclined to keep them independently defined (as suggested in the EIP). IMO, they can both be independently repriced to different value.

src/ethereum/forks/amsterdam/vm/instructions/system.py
gurukamath: The EIP does blur the line a little bit between transfer and create cost because they are not as independent of each other as in Osaka. In my opinion, the closest we can come to replication the Osaka style branching is something like

```create_gas_cost = GAS_NEW_ACCOUNT
    transfer_gas_cost = GAS_STATE_UPDATE
    if value == 0 or call_target != EMPTY_ACCOUNT:
        create_gas_cost = Uint(0)

    if value == 0:
        transfer_gas_cost = Uint(0)
    elif call_target != EMPTY_ACCOUNT:
        transfer_gas_cost += GAS_STATE_UPDATE```

File: src/ethereum/forks/amsterdam/vm/gas.py (hunk 1 above)
Comment (gurukamath) [reply src/ethereum/forks/amsterdam/vm/gas.py]: Updated to `GAS_COLD_ACCOUNT_COST_NO_CODE`

File: src/ethereum/forks/amsterdam/vm/instructions/environment.py (hunk 6 above)
Comment (gurukamath) [reply src/ethereum/forks/amsterdam/vm/instructions/environment.py:84]: Updated in all the relevant places.

File: src/ethereum/forks/amsterdam/vm/instructions/environment.py (hunk 7 above)
Comment (gurukamath) [reply src/ethereum/forks/amsterdam/vm/instructions/environment.py]: Updated

File: src/ethereum/forks/amsterdam/fork.py (hunk 9 above)
Comment (gurukamath) [reply src/ethereum/forks/amsterdam/fork.py]: Updated

File: src/ethereum/forks/amsterdam/fork.py (hunk 10 above)
Comment (gurukamath) [reply src/ethereum/forks/amsterdam/fork.py:668]: Possibly, but I'd still keep that outside the scope of this PR

File: src/ethereum/forks/amsterdam/vm/gas.py (hunk 2 above)
Comment (gurukamath) [reply src/ethereum/forks/amsterdam/vm/gas.py:85]: Added doc strings

File: src/ethereum/forks/amsterdam/vm/instructions/environment.py (hunk 4 above)
Comment (gurukamath) [reply src/ethereum/forks/amsterdam/vm/instructions/environment.py]: Switched to old fork style conditionals based on another review comment

File: src/ethereum/forks/amsterdam/vm/instructions/system.py (hunk 8 above)
Comment (gurukamath) [reply src/ethereum/forks/amsterdam/vm/instructions/system.py:418]: The EIP itself changes the gas model here to define a single `call_value_cost`. I'm ok with sticking to that even though the diff might get a touch messy

File: tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/__init__.py [hunk 11]
```diff
@@ -0,0 +1 @@
+"""Tests for [EIP-2780: Block-level Access Lists](https://eips.ethereum.org/EIPS/eip-2780)."""
```
danceratopz:
```suggestion
"""Tests for [EIP-2780: Reduce intrinsic transaction gas](https://eips.ethereum.org/EIPS/eip-2780)."""
```

packages/testing/src/execution_testing/forks/gas_costs.py
```diff
@@ -56,6 +56,11 @@ class GasCosts:
 
     GAS_AUTH_PER_EMPTY_ACCOUNT: int
 
+    # EIP-2780: Reduced intrinsic transaction gas costs (Amsterdam)
+    G_COLD_ACCOUNT_COST_CODE: int = 0
+    G_COLD_ACCOUNT_COST_NO_CODE: int = 0
+    G_STATE_UPDATE: int = 0
```
danceratopz: For consistency, shouldn't these be: 
```suggestion
    GAS_COLD_ACCOUNT_COST_CODE: int = 0
    GAS_COLD_ACCOUNT_COST_NO_CODE: int = 0
    GAS_STATE_UPDATE: int = 0
```
?

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/spec.py
```diff
@@ -0,0 +1,30 @@
...
+class Spec:
+    """Constants and parameters from EIP-2780."""
+
+    # GAS constants
+    GAS_COLD_ACCOUNT_COST_CODE: int = 2600
+    GAS_COLD_ACCOUNT_COST_NO_CODE: int = 500
+    GAS_STATE_UPDATE: int = 1000
+
+    # The base transaction cost
+    TX_BASE_COST: int = 4500
```
danceratopz: This class is unused by any tests as they use the values provided by `fork.gas_costs()`. I think that's the agreed pattern for handling gas cost changes, so I think we could remove this `Spec` class to avoid confusion.

src/ethereum/forks/amsterdam/vm/gas.py
danceratopz: I think we can remove this constant in Amsterdam, it's been replaced by `GAS_STATE_UPDATE` and `GAS_NEW_ACCOUNT` in 2780 mechanics, see:
https://github.com/ethereum/execution-specs/pull/2175/changes#diff-28a1b575fd7c3d82832c0826cf58a881101643543d35c123c78ca65202152c23R414

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -3500,3 +3516,202 @@ def engine_execution_payload_block_access_list(
...
+            G_STATE_UPDATE=1000,
+            GAS_BLOCK_ACCESS_LIST_ITEM=2000,
+        )
+
+    @classmethod
+    def transaction_intrinsic_cost_calculator(
+        cls, *, block_number: int = 0, timestamp: int = 0
+    ) -> TransactionIntrinsicCostCalculator:
+        """
+        At Prague, the transaction intrinsic cost needs to take the
```
danceratopz:
```suggestion
        At Amsterdam, the transaction intrinsic cost needs to take the
```

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -3500,3 +3516,202 @@ def engine_execution_payload_block_access_list(
...
+                else:
+                    value_cost = 2 * gas_costs.G_STATE_UPDATE
+
+        # Delegation cost (from Prague semantics)
+        delegation_cost = 0
+        if metadata["delegated_address"] or metadata["delegated_address_warm"]:
+            if metadata["delegated_address_warm"]:
+                delegation_cost = gas_costs.GAS_WARM_ACCOUNT_ACCESS
+            else:
+                delegation_cost = gas_costs.GAS_COLD_ACCOUNT_ACCESS
```
danceratopz: Shouldn't this be:
```suggestion
                delegation_cost = gas_costs.GAS_COLD_ACCOUNT_COST_CODE
```
to match:
https://github.com/gurukamath/execution-specs/blob/eip-2780/implement-eip-final/src/ethereum/forks/amsterdam/vm/eoa_delegation.py#L154 ?

Tricky code duplications!

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_move_to_precompiles.py
```diff
@@ -0,0 +1,145 @@
...
+
+    tx_data = _precompile_calldata(precompile)
+
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+    # Validate that the calculator accepts precompile parameters;
+    # the result is not used for balance checks because precompile
+    # execution gas varies.
+    intrinsic_gas_calculator(
+        calldata=tx_data,
+        sends_value=True if value else False,
```
danceratopz: I think this is more idiomatic/easier to read:
```suggestion
        sends_value=bool(value),
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
...
+            raise ValueError(f"Unknown recipient type {recipient_type}")
+
+    access_list = []
+    if warm_target:
+        access_list.append(AccessList(address=target, storage_keys=[]))
+
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+    total_gas_cost = intrinsic_gas_calculator(
+        access_list=access_list,
+        sends_value=True if value else False,
```
danceratopz:
```suggestion
        sends_value=bool(value),
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
...
+    access_list = []
+    if warm_target:
+        access_list.append(AccessList(address=target, storage_keys=[]))
+    if warm_delegation:
+        access_list.append(AccessList(address=delegated_to, storage_keys=[]))
+
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+    total_gas_cost = intrinsic_gas_calculator(
+        access_list=access_list,
+        sends_value=True if value else False,
```
danceratopz:
```suggestion
        sends_value=bool(value),
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
...
+    sender = pre.fund_eoa(initial_balance)
+    access_list = (
+        [AccessList(address=sender, storage_keys=[])] if warm_target else []
+    )
+
+    # Contract creation transaction
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+    total_gas_cost = intrinsic_gas_calculator(
+        access_list=access_list,
+        sends_value=True if value else False,
```
danceratopz:
```suggestion
        sends_value=bool(value),
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
...
+    execution_gas = call_data.execution_gas(
+        fork
+    ) + gsc.GAS_CODE_DEPOSIT_PER_BYTE * len(code_to_deploy)
+
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+    total_gas_cost = (
+        intrinsic_gas_calculator(
+            calldata=call_data,
+            contract_creation=True,
+            sends_value=True if value else False,
```
danceratopz:
```suggestion
            sends_value=bool(value),
```

src/ethereum/forks/amsterdam/vm/gas.py
danceratopz, line 66: I was thinking this constant becomes legacy and is no longer required. But it's correct to leave this, it's used [here in `selfdestruct()` in vm/instructions/system.py#L601](https://github.com/gurukamath/execution-specs/blob/eip-2780/implement-eip-final/src/ethereum/forks/amsterdam/vm/instructions/system.py#L601), please see this comment:
https://github.com/ethereum/execution-specs/pull/2175/changes#r2945517950

src/ethereum/forks/amsterdam/vm/instructions/system.py
danceratopz, line 601: This comment is a noop; it's just commentary, no action needed.

I tripped over this, as I assumed this "legacy" gas constant would be removed entirely, but the EIP explicitly states that:

> The semantics of SELFDESTRUCT remain unchanged. Following [EIP-6780](https://eip.directory/eips/eip-6780.md), only contracts created within the same transaction may be fully deleted. This EIP does not reprice or modify any SELFDESTRUCT side-effects.
https://github.com/ethereum/EIPs/blob/8634214221464da2fa9f47baab18b5252b2f5140/EIPS/eip-2780.md?plain=1#L421

This code is consistent with the EIP as-is.

danceratopz (commented):
Hey @gurukamath, here's some initial feedback. I must admit, I need to go deeper and improve my understanding to do this justice in a second round. I haven't looked at the tests updated for interactions yet. Will come back to this.

packages/testing/src/execution_testing/forks/gas_costs.py
gurukamath: Nice catch. I think these went unnoticed when I rebased on the latest `forks/amsterdam`. Updated

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -70,6 +70,22 @@
...
+"""
+Gas cost for a cold access to an account that has code deployed.
+7702 delegated accounts are also charged this for cold access.
```
SamWilsn:
```suggestion
[EIP-7702] delegated accounts are also charged this for cold access.

[EIP-7702]: https://eips.ethereum.org/EIPS/eip-7702
```

src/ethereum/forks/amsterdam/fork.py
SamWilsn: Works for me.

SamWilsn (approved):

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/__init__.py
gurukamath: Updated

packages/testing/src/execution_testing/forks/forks/forks.py
gurukamath: Updated the docsting

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/spec.py
gurukamath: Updated

src/ethereum/forks/amsterdam/fork.py
gurukamath: Created #2522

src/ethereum/forks/amsterdam/vm/gas.py
gurukamath: Updated

src/ethereum/forks/amsterdam/vm/gas.py
gurukamath: Updated

danceratopz:
> @danceratopz This got a bit out of hand in terms of PR size. If this is too much, I'd be happy to split this up into the following 2 PRs.
> 
>  1. Implementation + Framework updates and EIP-specific tests (in this PR)
> 
>  2. Fix broken legacy tests including BALs and other older fork tests

That's a good idea, @gurukamath, I think splitting the older fork tests out would be a great idea to give more targeted reviews.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
+"""
+Tests for EIP-2780 Reduce Transaction Intrinsic Cost.
+
+Tests that the value moving transactions charge gas appropriately.
```
danceratopz:
```suggestion
Test gas costs with EIP-2780 for value-moving transactions to:
- EOAs,
- contracts,
- empty accounts.
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
...
+    "value",
+    [
+        pytest.param(0, id="zero_value"),
+        pytest.param(1, id="non-zero_value"),
+    ],
+)
+def test_value_moving_transactions(
+    fork: Fork,
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
```
danceratopz: This should be a `state_test` if we're not touching block header fields. See
https://github.com/ethereum/execution-specs/blob/def3a0c214ee7eb9ce3aeeab467434e1d00477c6/docs/writing_tests/types_of_tests.md?plain=1#L78-L84

But please lmk if I missed something.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
...
+    "value",
+    [
+        pytest.param(0, id="zero_value"),
+        pytest.param(1, id="non-zero_value"),
+    ],
+)
+def test_value_moving_transaction_to_delegated_eoa(
+    fork: Fork,
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
```
danceratopz: Same here.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
...
+    "value",
+    [
+        pytest.param(0, id="zero_value"),
+        pytest.param(1, id="non-zero_value"),
+    ],
+)
+def test_value_transfer_to_self(
+    fork: Fork,
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
```
danceratopz: State test here, too.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
```diff
@@ -0,0 +1,337 @@
...
+    "value",
+    [
+        pytest.param(0, id="zero_value"),
+        pytest.param(1, id="non-zero_value"),
+    ],
+)
+def test_value_contract_creation_tx(
+    fork: Fork,
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
```
danceratopz: State test here, too.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_calls.py
danceratopz, line 1: I think all the tests here can be written as state tests.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_move_to_precompiles.py
danceratopz, line 1: I think all the tests here can also all be written as state tests.

danceratopz (commented):
Hey @gurukamath, here's something that stood out really quickly upon taking another look at these.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_calls.py
gurukamath: This is right. Will update

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
```diff
@@ -48,6 +49,7 @@ def test_bal_nonce_changes(
     tx = Transaction(
         sender=alice,
         to=bob,
+        gas_limit=10**6,
```
danceratopz: I personally prefer:
```suggestion
        gas_limit=1_000_000
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_invalid.py
danceratopz: _Missed a bit._ 😅 

This should also use the gas as calculated by the `intrinsic_gas_calculatooor`. But here the account will no longer be empty, so:
```
    intrinsic_gas_tx2 = fork.transaction_intrinsic_cost_calculator()(
        recipient_type=RecipientType.EOA,
        sends_value=True,		
    )
```

I think it only works because the actual intrinsic gas required is 6k (<21k!).

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
danceratopz, line 328: I think `"successful_extcodesize"` as-is deviates from the original test's intent; it should be renamed to `oog_phase2_at_extcodesize`, in which case, how about?
```suggestion
    ids=["oog_phase1_at_extcodesize", "oog_phase2_at_extcodesize"],
```
The question is do we need to add a new test case here `"successful_extcodesize"`, or is it covered elsewhere?

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -342,7 +344,9 @@ def test_bal_extcodesize_and_oog(
     # Create contract that checks target's code size
     codesize_checker_code = (
         Op.PUSH20(target_contract)  # Target contract address
-        + Op.EXTCODESIZE(address_warm=False)  # Check code size (cold access)
+        + Op.EXTCODESIZE(
+            address_warm=False, address_has_code=False
```
danceratopz, line 348: We only have an `Op.STOP`, but unless this changes too, we should rename `"successful_extcodesize"` to `"phase2_oog_at_extcodesize"`. We could consider adding a third successful case but I think the happy case is covered here https://github.com/gurukamath/execution-specs/blob/eip-2780/implement-eip-final/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py#L246.

```suggestion
            address_warm=False, address_has_code=True
```

danceratopz (commented):
Hey @gurukamath, here's a second round after reviewing `tests/amsterdam/eip7928_block_level_access_lists/`.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
gurukamath: I did not include a case here since it is covered specifically in `tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_account_access_opcodes.py::test_account_access_opcode`. The original idea was to keep the implementation (and tests) of EIP-7928 and EIP-2780 as independent as possible. Not sure how essential this separation is anymore after the debates we've been having about merging stuff to `forks/amsterdam`

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
gurukamath: I think the test ids are a bit misleading and might need to change. The test essentially looks at the target being in BAL vs target not being in BAL. Pre 2780 these two scenarios exactly coincide with Success vs OOG of the opcode. Post 2780, they don't. Which is also why there is a disconnect between the target having code and us using `address_has_code=False`.

danceratopz (commented):
I reviewed all other non-specific Amsterdam tests for hard-coded constants. Didn't add inline comments here, just Clauded these up.Each addresses a few missing hard-coded gas constants from a single fork (Cancun, Prague, Osaka).  I still need to self-review them:
- https://github.com/gurukamath/execution-specs/pull/12
- https://github.com/gurukamath/execution-specs/pull/11
- https://github.com/gurukamath/execution-specs/pull/10

gurukamath:
@danceratopz I have merged the changes you suggested in the 3 PRs as well as rebased this on the latest `forks/amsterdam` since the base was starting to get pretty stale. I made a call to squash all the 39 PR commits into one. Otherwise, the rebase would just be very painful. One more quick look and I think we are ready to merge this.

Next steps would be to write tests for more scenarios as well as fix the fails in the `ported_static` tests. (Separate PRs of course)

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -487,25 +492,28 @@ def callcode(evm: Evm) -> None:
...
-        access_gas_cost = GasCosts.COLD_ACCOUNT_ACCESS
-    else:
-        access_gas_cost = GasCosts.WARM_ACCESS
+        access_gas_cost = GasCosts.COLD_ACCOUNT_COST_NO_CODE
 
-    transfer_gas_cost = Uint(0) if value == 0 else GasCosts.CALL_VALUE
+    # Cost is simply dependent on value since the contract is always there
+    call_value_cost = (
+        Uint(0) if value == 0 else Uint(2) * GasCosts.STATE_UPDATE
+    )
```
danceratopz, line 502: Wouldn't this count as "self-call" as `caller == to` in CALLCODE? So there should only be one state update:
```suggestion
    call_value_cost = Uint(0) if value == 0 else GasCosts.STATE_UPDATE
```
If you agree, we should update the `transaction_intrinsic_cost_calculator()` and potentially tests this touches.

Also happy to do/discuss this as a follow-up!

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -380,29 +380,34 @@ def call(evm: Evm) -> None:
...
 
-    create_gas_cost = GasCosts.NEW_ACCOUNT
-    if value == 0 or is_account_alive(tx_state, to):
-        create_gas_cost = Uint(0)
+    call_value_cost = Uint(0)
+    if value > U256(0):
+        if call_target == EMPTY_ACCOUNT:
+            call_value_cost += GasCosts.STATE_UPDATE + GasCosts.NEW_ACCOUNT
+        else:
+            call_value_cost += Uint(2) * GasCosts.STATE_UPDATE
```
danceratopz, line 408: I think we need to check whether we're in a "self-call" context here and only charge `GasCosts.STATE_UPDATE` in that case.

danceratopz (approved):
Thanks @gurukamath for the rebase and merging the test updates in.

Two comments below. Even if they're valid, I don't want to block a merge, so giving an approve! We can tackle other fixes in follow-up PRs. But as you like!

gurukamath:
> Thanks @gurukamath for the rebase and merging the test updates in.
> 
> Two comments below. Even if they're valid, I don't want to block a merge, so giving an approve! We can tackle other fixes in follow-up PRs. But as you like!

Thanks for the review Dan.

Yes. The changes are relevant. The self-call semantics weren't part of he EIP when we implemented this PR and have only been added to the EIP recently, along with the transfer log costs associated with 7708. I am planning to add both of those in a follow-up PR. So I will go ahead and merge this for now.

PR #2176, opened 2026-02-09
refactor(tests): Add fork parameter to Initcode generator

pdobacz:
## 🗒️ Description

Initcode class now uses fork-specific gas costs and memory expansion formula instead of hardcoded values.

This should make tests leveraging this generator less sensitive to gas repricings. An EIP-3860 test needed to be adapted a bit, to not hardcode `Initcode` instances in parameters.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/tools/tools_code/generators.py
+++ b/packages/testing/src/execution_testing/tools/tools_code/generators.py
@@ -7,7 +7,7 @@ from pydantic import Field
 
 from execution_testing.base_types import Address, Bytes
 from execution_testing.forks import Fork, Frontier
-from execution_testing.test_types import EOA, Transaction, ceiling_division
+from execution_testing.test_types import EOA, Transaction
 from execution_testing.vm import Bytecode, ForkOpcodeInterface, Op
 
 
@@ -45,7 +45,6 @@ class Initcode(Bytecode):
         deploy_code: SupportsBytes | Bytes | None = None,
         initcode_length: int | None = None,
         initcode_prefix: Bytecode | None = None,
-        initcode_prefix_execution_gas: int = 0,
         padding_byte: int = 0x00,
         name: str = "",
         fork: Fork = Frontier,
@@ -62,24 +61,17 @@ class Initcode(Bytecode):
         if initcode_prefix is None:
             initcode_prefix = Bytecode()
 
-        gas_costs = fork.gas_costs()
-        memory_calculator = fork.memory_expansion_gas_calculator()
-
         initcode = initcode_prefix
         code_length = len(bytes(deploy_code))
-        execution_gas = initcode_prefix_execution_gas
 
         # PUSH2: length=<bytecode length>
         initcode += Op.PUSH2(code_length)
-        execution_gas = gas_costs.G_VERY_LOW
 
         # PUSH1: offset=0
         initcode += Op.PUSH1(0)
-        execution_gas += gas_costs.G_VERY_LOW
 
         # DUP2
         initcode += Op.DUP2
-        execution_gas += gas_costs.G_VERY_LOW
 
         # PUSH1: initcode_length=11 + len(initcode_prefix_bytes) (constant)
         no_prefix_length = 0x0B
@@ -87,21 +79,17 @@ class Initcode(Bytecode):
             "initcode prefix too long"
         )
         initcode += Op.PUSH1(no_prefix_length + len(initcode_prefix))
-        execution_gas += gas_costs.G_VERY_LOW
 
         # DUP3
         initcode += Op.DUP3
-        execution_gas += gas_costs.G_VERY_LOW
 
         # CODECOPY: destinationOffset=0, offset=0, length
-        initcode += Op.CODECOPY
-        word_copy_cost = gas_costs.G_COPY * ceiling_division(code_length, 32)
-        memory_cost = memory_calculator(new_bytes=code_length)
-        execution_gas += gas_costs.G_VERY_LOW + word_copy_cost + memory_cost
+        initcode += Op.CODECOPY(
+            data_size=code_length, new_memory_size=code_length
+        )
 
         # RETURN: offset=0, length
         initcode += Op.RETURN
-        execution_gas += 0
 
         initcode_plus_deploy_code = bytes(initcode) + bytes(deploy_code)
         padding_bytes = bytes()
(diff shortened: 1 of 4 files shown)


packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -126,7 +128,7 @@ def __new__(
         instance._name_ = name
         instance.deploy_code = deploy_code
         instance.execution_gas = execution_gas
```
marioevz: This is actually a great fit for the new `gas_cost` method of `Bytecode`, which we can use to get rid of the manual `execution_gas` calculation entirely:

```suggestion
        instance.execution_gas = initcode.gas_cost(fork)
```
works with the following caveats:
- `Op.CODECOPY` needs to know how much we are expanding the memory:
```python
initcode += Op.CODECOPY(
    data_size=code_length, new_memory_size=code_length
)
```
- `Op.RETURN` can return the cost of the code deposit when adding `code_deposit_size` parameter, but that would result in `instance.execution_gas` having the added deployment cost. So I think we can do:
```python
initcode += Op.RETURN
```
And then:
```python
instance.deployment_gas = Op.RETURN(
    code_deposit_size=len(bytes(instance.deploy_code))
).gas_cost(fork)
```

marioevz: This is actually a great fit for the new `gas_cost` method of `Bytecode`, which we can use to get rid of the manual `execution_gas` calculation entirely:

```suggestion
        instance.execution_gas = initcode.gas_cost(fork)
```

works with the following caveats:
- `Op.CODECOPY` needs to know how much we are expanding the memory:
```python
initcode += Op.CODECOPY(
    data_size=code_length, new_memory_size=code_length
)
```
- `Op.RETURN` can return the cost of the code deposit when adding `code_deposit_size` parameter, but that would result in `instance.execution_gas` having the added deployment cost. So I think we can do:
```python
initcode += Op.RETURN
```
And then:
```python
instance.deployment_gas = Op.RETURN(
    code_deposit_size=len(bytes(instance.deploy_code))
).gas_cost(fork)
```

marioevz: This is actually a great fit for the new `gas_cost` method of `Bytecode`, which we can use to replace the manual `execution_gas` calculation entirely:

```suggestion
        instance.execution_gas = initcode.gas_cost(fork)
```

works with the following caveats:
- `Op.CODECOPY` needs to know how much we are expanding the memory:
```python
initcode += Op.CODECOPY(
    data_size=code_length, new_memory_size=code_length
)
```
- `Op.RETURN` can return the cost of the contract deployment when adding `code_deposit_size` parameter, but that would result in `instance.execution_gas` having the added deployment cost. So I think we can do:
```python
initcode += Op.RETURN  # Zero gas cost here
```
And then:
```python
instance.deployment_gas = Op.RETURN(
    code_deposit_size=len(bytes(instance.deploy_code))
).gas_cost(fork)
```

tests/shanghai/eip3860_initcode/test_initcode.py
```diff
@@ -41,94 +40,101 @@
...
+            deploy_code=INITCODE_RESULTING_DEPLOYED_CODE,
+            initcode_length=33,
+            padding_byte=0x00,
+        )
+    elif initcode_name == "49120_bytes":
+        return Initcode(
+            name=initcode_name,
+            fork=fork,
+            deploy_code=INITCODE_RESULTING_DEPLOYED_CODE,
+            initcode_length=49120,
```
marioevz: I think this is the fork's max initcode size minus a 32-bytes.
```suggestion
            initcode_length=fork.max_initcode_size() - 32,
```

tests/shanghai/eip3860_initcode/test_initcode.py
```diff
@@ -41,94 +40,101 @@
...
-EMPTY_INITCODE.execution_gas = 0
+@pytest.fixture
+def initcode(fork: Fork, initcode_name: str) -> Initcode:
+    """Create an Initcode object with fork-specific gas calculations."""
+    if initcode_name == "max_size_ones":
+        return Initcode(
+            name=initcode_name,
+            fork=fork,
+            deploy_code=INITCODE_RESULTING_DEPLOYED_CODE,
+            initcode_length=Spec.MAX_INITCODE_SIZE,
```
marioevz: We can use the fork's variable since we now have access to the `fork` object.
```suggestion
            initcode_length=fork.max_initcode_size(),
```

tests/shanghai/eip3860_initcode/test_initcode.py
```diff
@@ -41,94 +40,101 @@
...
+            padding_byte=0x01,
+        )
+    elif initcode_name == "max_size_zeros":
+        return Initcode(
+            name=initcode_name,
+            fork=fork,
+            deploy_code=INITCODE_RESULTING_DEPLOYED_CODE,
+            initcode_length=Spec.MAX_INITCODE_SIZE,
```
marioevz: Same here.
```suggestion
            initcode_length=fork.max_initcode_size(),
```

tests/shanghai/eip3860_initcode/test_initcode.py
```diff
@@ -41,94 +40,101 @@
...
+            padding_byte=0x00,
+        )
+    elif initcode_name == "over_limit_ones":
+        return Initcode(
+            name=initcode_name,
+            fork=fork,
+            deploy_code=INITCODE_RESULTING_DEPLOYED_CODE,
+            initcode_length=Spec.MAX_INITCODE_SIZE + 1,
```
marioevz:
```suggestion
            initcode_length=fork.max_initcode_size() + 1,
```

tests/shanghai/eip3860_initcode/test_initcode.py
```diff
@@ -41,94 +40,101 @@
...
+            padding_byte=0x01,
+        )
+    elif initcode_name == "over_limit_zeros":
+        return Initcode(
+            name=initcode_name,
+            fork=fork,
+            deploy_code=INITCODE_RESULTING_DEPLOYED_CODE,
+            initcode_length=Spec.MAX_INITCODE_SIZE + 1,
```
marioevz:
```suggestion
            initcode_length=fork.max_initcode_size() + 1,
```

tests/shanghai/eip3860_initcode/test_initcode.py
```diff
@@ -41,94 +40,101 @@
...
+            deploy_code=INITCODE_RESULTING_DEPLOYED_CODE,
+            initcode_length=49120,
+            padding_byte=0x00,
+        )
+    elif initcode_name == "49121_bytes":
+        return Initcode(
+            name=initcode_name,
+            fork=fork,
+            deploy_code=INITCODE_RESULTING_DEPLOYED_CODE,
+            initcode_length=49121,
```
marioevz:
```suggestion
            initcode_length=fork.max_initcode_size() - 32 + 1,
```

marioevz (commented):
Thanks for this! Some comments I think might be good to get addressed before merging.

marioevz (approved):
Thanks!

PR #2182, opened 2026-02-10
refactor(bloatnet): split stubs into mainnet and bloatnet files

CPerezz:
## Summary
- Split `stubs.json` into two separate files: `stubs_mainnet.json` and `stubs_bloatnet.json`
- `stubs_mainnet.json` contains only real mainnet contract entries (XEN, USDC, IMT, STR), excluding the synthetic 30GB_ERC20 contract
- `stubs_bloatnet.json` is identical to the original `stubs.json` and includes all entries (including 30GB_ERC20)


kamilchodola (approved):

tests/benchmark/stateful/bloatnet/stubs_mainnet.json
```diff
@@ -0,0 +1,14 @@
+{
+  "test_sload_empty_erc20_balanceof_XEN": "0x06450dEe7FD2Fb8E39061434BAbCFC05599a6Fb8",
```
fselmo, line 2: nit: I think it would be good at some point to have a more central location for mainnet stubs so other contexts (e.g. `execute remote`) could draw from the same stubs file for mainnet use. We can do this here or we can refactor this later if this comes up.

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -594,7 +594,7 @@ def test_bloatnet_balance_extcodehash(
 APPROVE_SELECTOR = 0x095EA7B3  # approve(address,uint256)
 
 # Load token names from stubs.json for test parametrization
-_STUBS_FILE = Path(__file__).parent / "stubs.json"
+_STUBS_FILE = Path(__file__).parent / "stubs_bloatnet.json"
```
fselmo, line 597: Instead of redefining all mainnet stubs inside of `stubs_bloatnet.json` what if we load both and join them instead? That way we don't have repeated addresses / feels easier to keep track of things.

Suggestion:

- `stubs_mainnet.json` includes all the mainnet stubs of interest
- `stubs_bloatnet.json` only includes the additional stubs that are not on mainnet
- Offload this shared loading logic to something like `bloatnet/utils.py` as a single load of:

```py
import json                                                                                                                                                                                                    
from pathlib import Path

_STUBS_DIR = Path(__file__).parent

# if we move mainnet stubs to a more general place, can still point to that path here
with open(_STUBS_DIR / "stubs_mainnet.json") as _f:
    BLOATNET_STUBS = json.load(_f)
with open(_STUBS_DIR / "stubs_bloatnet.json") as _f:
    BLOATNET_STUBS.update(json.load(_f))
```

and in each of these test files use `from .utils import BLOATNET_STUBS`. Thoughts?

fselmo (approved):
I don't want to block process as my suggestions are mostly organizational, so I am approving. But I left some things for us to think about as I think we can share mainnet stubs across different testing contexts and I think we can DRY up duplication of mainnet stubs as well. But lgtm 👍🏼

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
fselmo: This could come as a later refactor as I think we would need to address how we pass `--address-stubs` json via CLI in case we would `execute remote` against bloatnet. There is no appending that can be done here so I think perhaps duplicating in its own file is better, at least for now.

Issue #2153, opened 2026-02-06
Add `docc` plugin that automatically links to an EIP when it is mentioned


SamWilsn:
We reference EIPs everywhere throughout the documentation, and every time we do, we need to link to it. For example:

```python
FLOOR_CALLDATA_COST = Uint(15)
"""
Minimum gas cost per byte of calldata as per [EIP-7976]. Used to calculate
the minimum gas cost for transactions that include calldata.

[EIP-7976]: https://eips.ethereum.org/EIPS/eip-7976
"""
```

It would be fairly easy to add a new `docc` plugin, like the ones in [here](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/src/ethereum_spec_tools/docc.py), that automatically finds text of the form `(EIP|ERC)-[0-9]+` and makes it a link to the relevant EIP.

PR #2195, opened 2026-02-11
chore: expand docstring guide

SamWilsn:
#### Cute Animal Picture

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,15 +95,15 @@ This specification aims to be:
 
 - Do not include the value of the constant in the docstring (not verbatim and not a decomposition). It's too easy to change the constant and forget to change the docstring.
 - Construct the constant's value from more basic components when doing so provides meaningful context.
-  - **Great:** `TARGET_BLOB_GAS_PER_BLOCK = GAS_PER_BLOB * BLOB_SCHEDULE_TARGET`
-  - **Good:** `TX_MAX_GAS = Uint(2 ** 24)`
-  - **Bad:** `TX_MAS_GAS = Uint(16_777_216)`
+    - **Great:** `TARGET_BLOB_GAS_PER_BLOCK = GAS_PER_BLOB * BLOB_SCHEDULE_TARGET`
+    - **Good:** `TX_MAX_GAS = Uint(2 ** 24)`
+    - **Bad:** `TX_MAS_GAS = Uint(16_777_216)`
 
 ##### Functions
 
 - The first paragraph should be a brief summary of what the function does, written in the imperative mood.
-  - **Good:** Build the house using the provided lumber.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -91,6 +91,20 @@ This specification aims to be:
 - Write in complete sentences, providing background and context for the associated code.
 - Link to relevant standards/EIPs.
```
danceratopz: I think it'd be good to re-order this and put the "don't"s together

danceratopz: We prefer reference-style links, I think it's worth mentioning.

We should update this when we have:
- #2153

CONTRIBUTING.md
```diff
@@ -91,6 +91,20 @@ This specification aims to be:
...
 
+##### Constants
+
+- Do not include the value of the constant in the docstring (not verbatim and not a decomposition). It's too easy to change the constant and forget to change the docstring.
```
danceratopz: "not verbatim and not a decomposition" feels a bit clunky.

CONTRIBUTING.md
```diff
@@ -91,6 +91,20 @@ This specification aims to be:
...
+- Construct the constant's value from more basic components when doing so provides meaningful context.
+    - **Great:** `TARGET_BLOB_GAS_PER_BLOCK = GAS_PER_BLOB * BLOB_SCHEDULE_TARGET`
+    - **Good:** `TX_MAX_GAS = Uint(2 ** 24)`
```
danceratopz: I would argue that this isn't really a "good" example, but rather "acceptable" as it still refers to a literal and goes against the above advice. But I think it's ok in this case!

CONTRIBUTING.md
```diff
@@ -91,6 +91,20 @@ This specification aims to be:
...
+    - **Bad:** `TX_MAS_GAS = Uint(16_777_216)`
```
danceratopz: I guess the "mas" typo in `TX_MAS_GAS` was unintentional even though it's a bad example.

CONTRIBUTING.md
```diff
@@ -91,6 +91,20 @@ This specification aims to be:
...
+
+##### Functions
+
+- The first paragraph should be a brief summary of what the function does, written in the imperative mood.
+    - **Good:** Build the house using the provided lumber.
```
danceratopz: Very clear, thanks!

danceratopz (commented):
Hey @SamWilsn, thanks for this! I made a few comments below, which I tried to address in:
- https://github.com/SamWilsn/eth1.0-specs/pull/5

This PR also revamps the top-level "Docstring" section bullet list. Please see the commit messages for more details.

danceratopz (approved):
Thanks @SamWilsn!

PR #2201, opened 2026-02-12
refactor(testing): Refactor `Initcode` to allign `gas_cost` behavior with `Bytecode`

marioevz:
## 🗒️ Description

Refactors on top of #2176 to align behavior of `Initcode.gas_cost` with what is normally expected of `Bytecode.gas_cost`.

Method `gas_cost` now correctly returns the full cost of executing the initcode including the code deposit.

`execution_gas` and `deployment_gas`, previously set during instantiation, are now public methods which take the `fork` as parameter and return the gas cost excluding the code deposit and the code deposit cost, respectively.

The `fork` parameter requirement was removed from the instantiation of `Initcode`.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/tools/tools_code/generators.py
+++ b/packages/testing/src/execution_testing/tools/tools_code/generators.py
@@ -25,7 +25,7 @@ class Initcode(Bytecode):
     EIP-3860 are *not* taken into account by any of these calculated costs.
     """
 
-    deploy_code: SupportsBytes | Bytes
+    deploy_code: Bytes | Bytecode
     """
     Bytecode to be deployed by the initcode.
     """
@@ -48,11 +48,13 @@ class Initcode(Bytecode):
         """
         if deploy_code is None:
             deploy_code = Bytecode()
+        elif not isinstance(deploy_code, Bytecode):
+            deploy_code = Bytes(deploy_code)
         if initcode_prefix is None:
             initcode_prefix = Bytecode()
 
         initcode = initcode_prefix
-        code_length = len(bytes(deploy_code))
+        code_length = len(deploy_code)
 
         # PUSH2: length=<bytecode length>
         initcode += Op.PUSH2(code_length)
@@ -79,7 +81,7 @@ class Initcode(Bytecode):
         )
 
         # RETURN: offset=0, length
-        initcode += Op.RETURN(code_deposit_size=len(bytes(deploy_code)))
+        initcode += Op.RETURN(code_deposit_size=len(deploy_code))
 
         initcode_plus_deploy_code = bytes(initcode) + bytes(deploy_code)
         padding_bytes = bytes()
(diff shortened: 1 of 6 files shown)


packages/testing/src/execution_testing/tools/tools_code/generators.py
fselmo: This docstring is stale now. We should update this as the `fork` is no longer part of the args here.

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -112,16 +104,48 @@ def __new__(
...
 
+    def execution_gas(
+        self,
+        fork: Type[ForkOpcodeInterface],
+        *,
+        block_number: int = 0,
+        timestamp: int = 0,
+    ) -> int:
+        """
+        Gas cost of executing the initcode, charged before the code
```
fselmo: Based on the previous comment, we could update that this is fork-aware here and in `deployment_gas()`. But this is very much a nit, take or leave 😄.

packages/testing/src/execution_testing/tools/tools_code/generators.py
fselmo: From Claude, there's a discrepancy here and in line `28` at the instance type definition. We should align these and it seems that `SupportsBytes | Bytes` is redundant since `Bytes` is also `SupportsBytes`.

Maybe we use something like `Bytecode | SupportsBytes | None` for both since it implies `Bytecode` is preferred... otherwise we can just use `SupportsBytes | None`. Thoughts?

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -112,16 +104,48 @@ def __new__(
...
+        fork: Type[ForkOpcodeInterface],
+        *,
+        block_number: int = 0,
+        timestamp: int = 0,
+    ) -> int:
+        """
+        Gas cost of deploying the contract.
+        """
+        return Op.RETURN(
+            code_deposit_size=len(bytes(self.deploy_code))
```
fselmo: nit: We removed the `bytes(...)` wrapping [here](https://github.com/ethereum/execution-specs/pull/2201/changes#diff-3f119b392d0b48df0dec9abc1952632641f35c7e78c8202ac3bbbb5930c57270L65-R57). I think we should be consistent with one or the other.

fselmo (approved):
lgtm! Left some nits but nothing blocking 👍🏼

packages/testing/src/execution_testing/tools/tools_code/generators.py
marioevz: Nice catch, yes, let's go with `Bytecode | SupportsBytes | None`, I'll update.

PR #2203, opened 2026-02-13
refactor(testing): Refactor specs to allow reusing logic for benchmark checks

marioevz:
## 🗒️ Description

This PR refactors the fill/execute processes to enable reuse of verification logic across both modes.

### Remove Mutable Fields from `BaseTest`

This PR removes all fields from `BaseTest` that were mutated as part of `generate` or `execute` and were (incorrectly) used as implicit outputs of those methods. These fields represented runtime-derived state rather than declarative test specification.

Removed fields:

- `_gas_optimization`: Previously used to store the result of the gas optimization process.
- `_opcode_count`: Output of opcode counting from the t8n tool (moved to the T8N class).
- `t8n_dump_dir`: Output directory for dumping state transition results (moved to the T8N class).
- `t8n_call_counter`: Counter tracking how many times the test invoked the t8n tool (moved to the T8N class).

Other removed fields:
- `_request`: Previously used for test introspection.
- `tag`: Unused leftover field from pre-pytest logic.

New input fields that supersede the removed ones:

- `operation_mode`: Set by the filler or executor and used by the test to determine whether benchmark or gas optimization modes are enabled.
- `is_tx_gas_heavy_test`/`is_exception_test`: Replace the previous introspection logic based on `_request`. These are now explicitly set by the filler or executor.

### Introduce `FillResult` and `ExecuteResult`

This PR introduces `FillResult` and `ExecuteResult`, which encapsulate the outputs of the `generate`/`execute` methods when used by fill and execute, respectively.

By returning structured result objects instead of mutating spec fields, the test specification becomes reusable across execution modes. This also allows benchmark validation logic to be extracted from the test spec classes and shared between `fill` and `execute`.

The change was motivated by this comment: https://github.com/ethereum/execution-specs/pull/2203#discussion_r2802420845
where we had different logic for the verification in `StateTest` and `BlockchainTest`.

### Abstract T8N Information Away from the Specs

This PR makes the distinction between `session_t8n` and `t8n`:
- `session_t8n`: A global instance shared across all tests executed within a process.
- `t8n`: A per-test configured instance derived from `session_t8n`.

At the start of each test, the t8n instance is prepared with the following setup:
- The opcode counter is reset
- The debug output path (if configured) is set for the specific test.
- The execution counter is reset to zero.

This removes t8n-related runtime state from the test specification and centralizes it within the T8N instance, and it's now responsibility of the filler or executor to properly extract and use the information.

 ### Fix Bug that Re-Wrapped Spec Class in Fill and Execute

The `BaseTest.__pydantic_init_subclass__` method now verifies that newly created subclasses are not wrapper classes generated by `base_test_parametrizer_func`.

Previously, dynamically created `BaseTestWrapper` subclasses would overwrite entries in `BaseTest.spec_types`, causing classes to be re-wrapped unintentionally.

This issue only manifested when the filler plugin was imported multiple times within the same process (e.g., during unit tests or pre-alloc grouping phase 1 execution). In those cases, tests could be executed twice because `__init__` was effectively re-entering due to repeated wrapping, and also caused multiple tests to reuse the same T8N instance that was configured for another test, causing unintended state-sharing.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
@@ -650,6 +650,8 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
             )
 
         class BaseTestWrapper(cls):  # type: ignore
+            __is_base_test_wrapper__ = True
+
             def __init__(self, *args: Any, **kwargs: Any) -> None:
                 if "pre" not in kwargs:
                     kwargs["pre"] = pre
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
@@ -625,6 +625,8 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
         max_fee_per_blob_gas: int,
         max_gas_limit_per_test: int | None,
         gas_limit_accumulator: GasInfoAccumulator,
+        is_tx_gas_heavy_test: bool,
+        is_exception_test: bool,
     ) -> Type[BaseTest]:
         """
         Fixture used to instantiate an auto-fillable BaseTest object from
@@ -655,10 +657,12 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
                     raise ValueError(
                         "The pre-alloc object was modified by the test."
                     )
-                # Set default for expected_benchmark_gas_used
                 if "expected_benchmark_gas_used" not in kwargs:
                     kwargs["expected_benchmark_gas_used"] = gas_benchmark_value
                 kwargs["fork"] = fork
+                kwargs["operation_mode"] = request.config.op_mode
+                kwargs["is_tx_gas_heavy_test"] = is_tx_gas_heavy_test
+                kwargs["is_exception_test"] = is_exception_test
                 kwargs |= {
                     p: request.getfixturevalue(p)
                     for p in cls_fixture_parameters
@@ -668,7 +672,6 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
                 request.node.config.sender_address = str(pre._sender)
 
                 super(BaseTestWrapper, self).__init__(*args, **kwargs)
-                self._request = request
                 execute = self.execute(execute_format=execute_format)
 
                 # get balances of required sender accounts
@@ -730,12 +733,17 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
                     [str(eoa) for eoa in pre._funded_eoa]
                 )
 
-                execute.execute(
+                execute_result = execute.execute(
                     fork=fork,
                     eth_rpc=eth_rpc,
                     engine_rpc=engine_rpc,
                     request=request,
                 )
+                self.validate_benchmark_gas(
+                    benchmark_gas_used=execute_result.benchmark_gas_used,
+                    gas_benchmark_value=gas_benchmark_value,
+                )
+
                 collector.collect(request.node.nodeid, execute)
 
         return BaseTestWrapper
@@ -744,7 +752,12 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
 
 
 # Dynamically generate a pytest fixture for each test spec type.
-for cls in BaseTest.spec_types.values():
+for name, cls in BaseTest.spec_types.items():
+    if getattr(cls, "__is_base_test_wrapper__", False):
+        raise RuntimeError(
+            f"Test spec type {name}: {cls.__name__} is already wrapped. "
+            f"{BaseTest.spec_types.items()}."
+        )
     # Fixture needs to be defined in the global scope so pytest can detect it.
     globals()[cls.pytest_parameter_name()] = base_test_parametrizer(cls)
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -42,6 +42,7 @@ from execution_testing.client_clis.clis.geth import FixtureConsumerTool
 from execution_testing.fixtures import (
     BaseFixture,
     BlockchainEngineFixture,
+    BlockchainEngineXFixture,
     BlockchainFixture,
     FixtureCollector,
     FixtureConsumer,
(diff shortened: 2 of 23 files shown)


packages/testing/src/execution_testing/specs/state.py
```diff
@@ -473,6 +477,11 @@ def make_state_test_fixture(
                     f"expected_benchmark_gas_used "
                     f"({expected_benchmark_gas_used}), difference: {diff}"
                 )
+                assert gas_used <= max_benchmark_gas_allowed, (
+                    f"gas_used ({gas_used}) exceeds maximum benchmark gas "
+                    "allowed for this configuration: "
+                    f"({max_benchmark_gas_allowed})"
+                )
```
LouisTsai-Csie: This check should live outside the `if not self.skip_gas_used_validation` check. Even when `skip_gas_used_validation` is enabled, we still need to ensure that gas used does not exceed the block gas limit.

Let’s align this with the behavior in `blockchain.py`.

LouisTsai-Csie (dismissed):
Could we add support to `execute.py` as well? Currently there is no check for stateful benchmark via `fill`, so we might encounter the same issue when doing execute remote.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -1550,6 +1564,12 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
                             self._gas_optimization
```
danceratopz: This now needs to take the value from `fill_result.gas_optimization`.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
```diff
@@ -649,17 +651,18 @@ def base_test_parametrizer_func(
 
         class BaseTestWrapper(cls):  # type: ignore
```
danceratopz, line 652: I think we need this to mirror `fill`'s `BaseTestWrapper`. It seems unlikely, but this should prevent wrapper class overwrites in the registry if execute is imported a second time.
```suggestion
        class BaseTestWrapper(cls):  # type: ignore
                __is_base_test_wrapper__ = True
```

packages/testing/src/execution_testing/specs/base.py
```diff
@@ -299,5 +250,41 @@ def get_genesis_environment(self) -> Environment:
...
 
+    def validate_benchmark_gas(
+        self, *, benchmark_gas_used: int | None, gas_benchmark_value: int
+    ) -> None:
+        """
+        Validates the total consumed gas of the last block in the test matches
+        the expectation of the benchmark test.
+
+        Requires the following fields to be set:
+        - _benchmark_gas_used
```
danceratopz:
```suggestion
        - benchmark_gas_used
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py
```diff
@@ -295,6 +295,27 @@ def alloc_flags(
...
+
+
+@pytest.fixture(scope="function")
+def is_exception_test(request: pytest.FixtureRequest) -> bool:
+    """
+    Check if the test is an exception test (invalid block, invalid
+    transaction).
+
+    `None` is returned if it's not possible to determine if the test is
+    negative or not. This is the case when the test is not run in pytest.
```
danceratopz: Nit: This is inconsistent with the return type (`bool`) in the function signature typehint.

danceratopz (commented):
Really nice addition. We've needed a better separation of concerns here for a while!

Few minor comments below, a now defunct field is still referenced and otherwise mainly nits.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
marioevz: Ah definitely, I had it in my head to do this but forgot it, thanks so much for noticing :D

packages/testing/src/execution_testing/specs/base.py
marioevz:
```suggestion
        - expected_benchmark_gas_used
```
Is more accurate because `self.expected_benchmark_gas_used` is used but `self.benchmark_gas_used` does not exist anymore. Thanks!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
marioevz: I just checked and I think it's ok to leave set as `None` because `fill_result` will most likely be undefined if there was an exception. Great catch.

packages/testing/src/execution_testing/execution/transaction_post.py
```diff
@@ -138,32 +133,21 @@ def execute(
...
-            total_gas_used = 0
-            # Fetch transaction receipts to get actual gas used
-            for tx_hash in all_tx_hashes:
+                current_block_tx_hashes = [tx.hash for tx in signed_txs]
+            all_tx_hashes.extend(current_block_tx_hashes)
+            last_block_tx_hashes = current_block_tx_hashes
+
+        # Fetch transaction receipts to get actual gas used
+        benchmark_gas_used: int | None = None
+        if self.benchmark_mode:
```
marioevz, line 142: I still don't really like that execute and fill have different switches (fill has `operation_mode`).

marioevz: Created an issue to follow this up: https://github.com/ethereum/execution-specs/issues/2258

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -1722,9 +1725,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
                         # Force adding something to the list, even if it's
                         # None, to keep track of failed tests in the output
                         # file.
-                        gas_optimized_tests[request.node.nodeid] = (
-                            self._gas_optimization
-                        )
+                        gas_optimized_tests[request.node.nodeid] = None
```
spencer-tb:
```suggestion
                        gas_optimized_tests[request.node.nodeid] = (
                            fill_result.gas_optimization if fill_result is not None else None
                        )
```
Following from https://github.com/ethereum/execution-specs/pull/2203#discussion_r2821829060:
I think this is needed for the success path

spencer-tb (approved):
LGTM! Really nice refactor :D

Only a small nit!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
marioevz: Nice catch!

Issue #2155, opened 2026-02-06
Switch to a `PreState` interface


petertdavies:
The prestate for the EELS state transition should provided via standard interface. This serves two purposes:
1. Clearly specifies what operations stateDBs/witnesses/etc need to support.
2. Allows EELS to use multiple different prestate implementations (e.g. python dict, on disk db, witness)

We should add a basic `DictPreState` for use by `t8n` and EEST. More sophisticated `PreState`s can be added later.

### Rough interface draft

```python
class PreState(Protocol):
    def get_account_optional(self, address: Address) -> Account:
        ...

    def get_storage(self, address: Address, key: Bytes32) -> U256:
        ...

    def account_has storage(self, address: Address) -> bool:
        """Only needed for EIP-7610"""
        ...

    def compute_state_root_and_trie_changes(
        self,
        account_changes: Dict[Address, Optional[Account]],
        storage_changes: Dict[Address, Dict[Bytes32, U256],
    ) -> Tuple[Bytes32, List[InternalNode]:
        ...
```

petertdavies:
Assigned to @gurukamath. @jsign has recently implemented something similar to `compute_state_root_and_trie_changes()` so we should be able to use his code.

kevaundray:
I believe this can now be closed? cc @gurukamath @jsign

jsign:
I think so, by https://github.com/ethereum/execution-specs/pull/2207.

I'll let @gurukamath double-check and close if feels that's correct.

gurukamath:
Close by #2207. The backporting will be tracked in #2245

PR #2207, opened 2026-02-13
feat(specs): Replace mutable `State` with `PreState` + `StateTracker`

gurukamath:
## 🗒️ Description

Note: This PR was originally aimed at #2155 which is in the context of the zkEVM project. However, this is a more elegant design of State for EELS (designed by Peter in #2063). Therefore, `forks/amsterdam` and the BALs implementation could also benefit from it

This PR is an implementation of the changes to state within EELS

1. Defines a PreState interface irrespective of implementation
2. Re-factors the current State as an implementation of the above interface
3. Creates state tracking classes which effectively track the state diffs as the block is processed
4. Incorporates the current BAL logic within the framework of this state tracker

### Next Steps (in separate PRs)

1. Code to be accessible only via codehash and codehash to be a part of the `Account`
2. For BALs tightly couple access and the corresponding access tracking (currently some access and its tracking are done separately which could lead to bugs)
3. Explore if the `Trie` implementation can be made fork agnostic
4. Make the testing framework and EELS share the `PreState` thus simplifying the `T8N` tool
5. Port changes to older forks

## 🔗 Related Issues or PRs
#2155
#2063

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

TBD

### DIFF
--- a/src/ethereum/forks/amsterdam/block_access_lists/builder.py
+++ b/src/ethereum/forks/amsterdam/block_access_lists/builder.py
@@ -14,12 +14,14 @@ The builder follows a two-phase approach:
 """  # noqa: E501
 
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Dict, List, Set
+from typing import Dict, List, Optional, Set
 
-from ethereum_types.bytes import Bytes
-from ethereum_types.numeric import U64, U256
+from ethereum_types.bytes import Bytes, Bytes32
+from ethereum_types.numeric import U64, U256, Uint
 
-from ..fork_types import Address
+from ethereum.state import Account, Address, PreState
+
+from ..state_tracker import BlockState, TransactionState
 from .rlp_types import (
     AccountChanges,
     BalanceChange,
@@ -30,9 +32,6 @@ from .rlp_types import (
     SlotChanges,
     StorageChange,
 )
-
-if TYPE_CHECKING:
-    from ..state_tracker import StateChanges
 
 
 @dataclass
@@ -89,6 +88,15 @@ class BlockAccessListBuilder:
     [`BlockAccessList`]: ref:ethereum.forks.amsterdam.block_access_lists.rlp_types.BlockAccessList
     """  # noqa: E501
 
+    block_access_index: BlockAccessIndex = BlockAccessIndex(0)
+    """
+    Current block access index.  Set by the caller before each
+    [`incorporate_tx_into_block`] call (0 for system txs, i+1 for the
+    i-th user tx, N+1 for post-execution operations).
+
+    [`incorporate_tx_into_block`]: ref:ethereum.forks.amsterdam.state_tracker.incorporate_tx_into_block
+    """  # noqa: E501
+
     accounts: Dict[Address, AccountData] = field(default_factory=dict)
     """
     Mapping from account address to its tracked changes during block execution.
@@ -104,7 +112,6 @@ def ensure_account(builder: BlockAccessListBuilder, address: Address) -> None:
     multiple times for the same address.
 
     [`AccountData`]: ref:ethereum.forks.amsterdam.block_access_lists.builder.AccountData
-
     """  # noqa: E501
     if address not in builder.accounts:
         builder.accounts[address] = AccountData()
@@ -156,8 +163,6 @@ def add_storage_read(
     Records that a storage slot was read during execution. Storage slots
     that are both read and written will only appear in the storage changes
     list, not in the storage reads list, as per [EIP-7928].
-
-    [EIP-7928]: https://eips.ethereum.org/EIPS/eip-7928
     """
     ensure_account(builder, address)
     builder.accounts[address].storage_reads.add(slot)
@@ -358,58 +363,111 @@ def _build_from_builder(
     return block_access_list
 
 
-def build_block_access_list(
-    state_changes: "StateChanges",
-) -> BlockAccessList:
+def _get_pre_tx_account(
+    pre_tx_accounts: Dict[Address, Optional[Account]],
+    pre_state: PreState,
+    address: Address,
+) -> Optional[Account]:
     """
-    Build a [`BlockAccessList`] from a StateChanges frame.
+    Look up an account in cumulative state, falling back to `pre_state`.
 
-    Converts the accumulated state changes from the frame-based architecture
-    into the final deterministic BlockAccessList format.
+    The cumulative account state (`pre_tx_accounts`) should contain state up
+    to (but not including) the current transaction.
 
-    [`BlockAccessList`]: ref:ethereum.forks.amsterdam.block_access_lists.rlp_types.BlockAccessList
-    [`StateChanges`]: ref:ethereum.forks.amsterdam.state_tracker.StateChanges
-    """  # noqa: E501
-    builder = BlockAccessListBuilder()
+    Returns `None` if the `address` does not exist.
+    """
+    if address in pre_tx_accounts:
+        return pre_tx_accounts[address]
+    return pre_state.get_account_optional(address)
 
-    # Add all touched addresses
-    for address in state_changes.touched_addresses:
-        add_touched_account(builder, address)
 
-    # Add all storage reads
-    for address, slot in state_changes.storage_reads:
-        add_storage_read(builder, address, U256(int.from_bytes(slot)))
+def _get_pre_tx_storage(
+    pre_tx_storage: Dict[Address, Dict[Bytes32, U256]],
+    pre_state: PreState,
+    address: Address,
+    key: Bytes32,
+) -> U256:
+    """
+    Look up a storage value in cumulative state, falling back to `pre_state`.
 
-    # Add all storage writes
-    # Net-zero filtering happens at transaction commit time, not here.
-    # At block level, we track ALL writes at their respective indices.
-    for (
-        address,
-        slot,
-        block_access_index,
-    ), value in state_changes.storage_writes.items():
-        u256_slot = U256(int.from_bytes(slot))
-        add_storage_write(
-            builder, address, u256_slot, block_access_index, value
+    Returns `0` if not set.
+    """
+    if address in pre_tx_storage and key in pre_tx_storage[address]:
+        return pre_tx_storage[address][key]
+    return pre_state.get_storage(address, key)
+
+
+def update_builder_from_tx(
+    builder: BlockAccessListBuilder,
+    tx_state: TransactionState,
+) -> None:
+    """
+    Update the BAL builder with changes from a single transaction.
+
+    Compare the transaction's writes against the block's cumulative
+    state (falling back to `pre_state`) to extract balance, nonce, code, and
+    storage changes.  Net-zero filtering is automatic: if the pre-tx value
+    equals the post-tx value, no change is recorded.
+
+    Must be called **before** the transaction's writes are merged into
+    the block state.
+    """
+    block_state = tx_state.parent
+    pre_state = block_state.pre_state
+    idx = builder.block_access_index
+
+    # Compare account writes against block cumulative state
+    for address, post_account in tx_state.account_writes.items():
+        pre_account = _get_pre_tx_account(
+            block_state.account_writes, pre_state, address
         )
 
-    # Add all balance changes (balance_changes is keyed by (address, index))
-    for (
-        address,
-        block_access_index,
-    ), new_balance in state_changes.balance_changes.items():
-        add_balance_change(builder, address, block_access_index, new_balance)
+        pre_balance = pre_account.balance if pre_account else U256(0)
+        post_balance = post_account.balance if post_account else U256(0)
+        if pre_balance != post_balance:
+            add_balance_change(builder, address, idx, post_balance)
 
-    # Add all nonce changes
-    for address, block_access_index, new_nonce in state_changes.nonce_changes:
-        add_nonce_change(builder, address, block_access_index, new_nonce)
+        pre_nonce = pre_account.nonce if pre_account else Uint(0)
+        post_nonce = post_account.nonce if post_account else Uint(0)
+        if pre_nonce != post_nonce:
+            add_nonce_change(builder, address, idx, U64(post_nonce))
 
-    # Add all code changes
-    # Filtering happens at transaction level in eoa_delegation.py
-    for (
-        address,
-        block_access_index,
-    ), new_code in state_changes.code_changes.items():
-        add_code_change(builder, address, block_access_index, new_code)
+        pre_code = pre_account.code if pre_account else b""
+        post_code = post_account.code if post_account else b""
+        if pre_code != post_code:
+            add_code_change(builder, address, idx, post_code)
+
+    # Compare storage writes against block cumulative state
+    for address, slots in tx_state.storage_writes.items():
+        for key, post_value in slots.items():
+            pre_value = _get_pre_tx_storage(
+                block_state.storage_writes, pre_state, address, key
+            )
+            if pre_value != post_value:
+                # Convert slot from internal Bytes32 format to U256 for BAL.
+                # EIP-7928 uses U256 as it's more space-efficient in RLP.
+                u256_slot = U256.from_be_bytes(key)
+                add_storage_write(builder, address, u256_slot, idx, post_value)
+
+
+def build_block_access_list(
+    builder: BlockAccessListBuilder,
+    block_state: BlockState,
+) -> BlockAccessList:
+    """
+    Build a [`BlockAccessList`] from the builder and block state.
+
+    Feed accumulated reads from the block state into the builder, then produce
+    the final sorted and encoded block access list.
+
+    [`BlockAccessList`]: ref:ethereum.forks.amsterdam.block_access_lists.rlp_types.BlockAccessList
+    """  # noqa: E501
+    # Add storage reads (convert Bytes32 to U256 for BAL encoding)
+    for address, slot in block_state.storage_reads:
+        add_storage_read(builder, address, U256.from_be_bytes(slot))
+
+    # Add touched addresses
+    for address in block_state.account_reads:
+        add_touched_account(builder, address)
 
     return _build_from_builder(builder)
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -27,9 +27,14 @@ from ethereum.exceptions import (
     InvalidSenderError,
     NonceMismatchError,
 )
+from ethereum.state import Address
 
 from . import vm
-from .block_access_lists.builder import build_block_access_list
+from .block_access_lists.builder import (
+    BlockAccessListBuilder,
+    build_block_access_list,
+)
+from .block_access_lists.rlp_types import BlockAccessIndex
 from .block_access_lists.rlp_utils import compute_block_access_list_hash
 from .blocks import Block, Header, Log, Receipt, Withdrawal, encode_receipt
 from .bloom import logs_bloom
@@ -44,7 +49,7 @@ from .exceptions import (
     PriorityFeeGreaterThanMaxFeeError,
     TransactionTypeContractCreationError,
 )
-from .fork_types import Account, Address, Authorization, VersionedHash
+from .fork_types import Authorization, VersionedHash
 from .requests import (
     CONSOLIDATION_REQUEST_TYPE,
     DEPOSIT_REQUEST_TYPE,
@@ -54,26 +59,19 @@ from .requests import (
 )
 from .state import (
     State,
-    TransientStorage,
-    account_exists_and_is_empty,
-    destroy_account,
-    get_account,
-    increment_nonce,
-    modify_state,
-    set_account_balance,
-    state_root,
+    apply_changes_to_state,
 )
 from .state_tracker import (
-    StateChanges,
-    capture_pre_balance,
-    commit_transaction_frame,
-    create_child_frame,
-    filter_net_zero_frame_changes,
-    increment_block_access_index,
+    BlockState,
+    TransactionState,
+    account_exists_and_is_empty,
+    destroy_account,
+    extract_block_diffs,
+    get_account,
+    incorporate_tx_into_block,
+    increment_nonce,
+    set_account_balance,
     track_address,
-    track_balance_change,
-    track_nonce_change,
-    track_selfdestruct,
 )
 from .transactions import (
     AccessListTransaction,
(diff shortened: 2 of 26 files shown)


kevaundray:
Does this supercede #2063 ?

petertdavies:
Had a quick look at a high level and this looks good. It's basically a better and more comprehensive version of my draft PR. I will do a full review on Tuesday (I don't work Mondays).

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -1,558 +1,781 @@
...
-    storage_writes: Dict[Tuple[Address, Bytes32, BlockAccessIndex], U256] = (
-        field(default_factory=dict)
+    storage_writes: Dict[Address, Dict[Bytes32, U256]] = field(
+        default_factory=dict
     )
 
-    balance_changes: Dict[Tuple[Address, BlockAccessIndex], U256] = field(
+
+@dataclass
+class TxStateTracker:
```
petertdavies: I think the `Tracker` suffix is redundant. We can just go with `BlockState` and `TxState`. Maybe we should decide the final name before merging.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -1,558 +1,781 @@
...
+        tracker.storage_writes[address] = {}
+    tracker.storage_writes[address][key] = value
 
 
-def track_balance_change(
-    state_changes: StateChanges,
-    address: Address,
-    new_balance: U256,
-) -> None:
+def destroy_account(tracker: TxStateTracker, address: Address) -> None:
```
petertdavies: `destroy_account()` and `destroy_storage()` don't support non-same-tx destruction. This isn't an issue in Amsterdam, but the docstring should mention this.

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -16,7 +16,16 @@
 from ethereum import trace
 from ethereum.exceptions import EthereumException, InvalidBlock
 from ethereum.fork_criteria import ByBlockNumber, ByTimestamp, Unscheduled
-from ethereum.forks.amsterdam.state_tracker import StateChanges
+
+# TODO: Make this not amsterdam specific before PR merge.
```
petertdavies: ^this

petertdavies (dismissed):
I found a few minor issues, but this looks really good. It's a much cleaner version of my design. Having all reads go through `TxStateTracker` makes the typing situation so much nicer.

src/ethereum/forks/amsterdam/state_tracker.py
gurukamath: Yup. Makes sense to me. I would propose `BlockState` and `TransactionState` which also nicely aligns with `PreState`. Also, we might want better names for the modules themselves. Right now, we have a ethereum level `state.py` that has the `PreState` protocol, a fork level `state.py` with the `PreState` implementation and `state_tracker.py`. If we decide to drop the `tracker` suffix from here as well, it might all get confusing.

petertdavies: I think the end goal is:
* `ethereum/state.py`: `PreState` protocol and `DictPreState` (basic implementation of PreState)
* `ethereum/<fork>/state.py`: `BlockState` and `TransactionState`

But that requires more refactoring that should go in a later PR.

Maybe we should leave the names as is and do renaming PR at the end.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -1,558 +1,781 @@
...
+
     """
-    Filter net-zero changes from transaction frame before commit.
+    tracker.account_reads.add(address)
 
-    Compares final values against pre-tx state for storage, balance, and code.
-    Net-zero storage writes are converted to reads. Net-zero balance/code
-    changes are removed entirely. Nonces are not filtered (only increment).
+
+def track_storage_read(
```
petertdavies, line 744: Reads should exactly match calls to `get_account()`/`get_storage()`. We shouldn't need independent tracking functions. I'm minded to try remove these in a later PR and just track calls to the getters.

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -451,69 +460,156 @@ def _build_from_builder(
     return block_access_list
 
 
+def _get_running_account(
```
SamWilsn: This name is easily misinterpreted as "get an account that's running" instead of "get account from running state".

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -451,69 +460,156 @@ def _build_from_builder(
...
+    running_accounts: Dict[Address, Optional[Account]],
+    pre_state: PreState,
+    address: Address,
+) -> Optional[Account]:
+    """
+    Look up an account in running state, falling back to pre_state.
```
SamWilsn:
```suggestion
    Look up an account in running state, falling back to `pre_state`.
```

src/ethereum/state.py
```diff
@@ -0,0 +1,144 @@
...
+    :local:
+
+Introduction
+------------
+
+Define shared types and the ``PreState`` protocol used by the state
+transition function.  The ``PreState`` protocol specifies the operations
+that any pre-execution state provider must support, allowing multiple
+backing implementations (in-memory dict, on-disk database, witness,
+etc.).
```
SamWilsn: Please use markdown when adding new files.

src/ethereum/state.py
```diff
@@ -0,0 +1,144 @@
...
+    """Branch node in the Merkle Trie."""
+
+    subnodes: BranchSubnodes
+    value: Extended
+
+
+InternalNode = LeafNode | ExtensionNode | BranchNode
+
+
+class PreState(Protocol):
```
SamWilsn, line 93: Using "pre" here feels off. That prefix implies some kind of temporal relationship, but there isn't anything inherently "pre-" about the interface. This protocol doesn't necessarily need to represent "prestate", but any read-only state, right? Some off-the-cuff suggestions: `Snapshot`, `ReadableState`, `StateRead`...

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -13,13 +13,18 @@
 [`BlockAccessList`]: ref:ethereum.forks.amsterdam.block_access_lists.rlp_types.BlockAccessList  # noqa: E501
 """
 
+from __future__ import annotations
+
```
SamWilsn: When should we be making this import? Do we need to do it in more files?

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -89,6 +91,13 @@ class BlockAccessListBuilder:
     [`BlockAccessList`]: ref:ethereum.forks.amsterdam.block_access_lists.rlp_types.BlockAccessList  # noqa: E501
     """
 
+    block_access_index: BlockAccessIndex = BlockAccessIndex(0)
+    """
+    Current block access index.  Set by the caller before each
+    ``incorporate_tx_into_block`` call (0 for system txs, i+1 for the
+    i-th user tx, N+1 for post-execution operations).
```
SamWilsn, line 95:
```suggestion
    Current block access index.  Set by the caller before each
    [`incorporate_tx_into_block`][0] call (0 for system txs, i+1
    for the i-th user tx, N+1 for post-execution operations).
    
    [0]: ref:...
```

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -451,69 +460,156 @@ def _build_from_builder(
     return block_access_list
 
 
+def _get_running_account(
+    running_accounts: Dict[Address, Optional[Account]],
```
SamWilsn: I'm not entirely sure that `running_accounts` is the best name for this. I'm guessing that this is "running" as in "running tally"?

Maybe... `working_state`? Or `scratch_accounts`?

I just want to avoid the common "running" as "executing" interpretation.

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -451,69 +460,156 @@ def _build_from_builder(
...
+
+
+def update_builder_from_tx(
+    builder: BlockAccessListBuilder,
+    tx_tracker: TxStateTracker,
+) -> None:
+    """
+    Update the BAL builder with changes from a single transaction.
+
+    Diff the transaction's writes against the block's cumulative
```
SamWilsn:
```suggestion
    Compare the transaction's writes against the block's cumulative
```

Or maybe:

```suggestion
    Compute the changes from the transaction's writes against the block's cumulative
```

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -451,69 +460,156 @@ def _build_from_builder(
...
+            add_code_change(builder, address, idx, post_code)
+
+    # Diff storage writes against block cumulative state
+    for address, slots in tx_tracker.storage_writes.items():
+        for key, post_value in slots.items():
+            pre_value = _get_running_storage(
+                block_tracker.storage_writes, pre_state, address, key
+            )
+            if pre_value != post_value:
+                u256_slot = U256(int.from_bytes(key))
```
SamWilsn: Does this not work:


```suggestion
                u256_slot = U256.from_be_bytes(key)
```

?

src/ethereum/forks/amsterdam/block_access_lists/builder.py
```diff
@@ -451,69 +460,156 @@ def _build_from_builder(
...
+                add_storage_write(builder, address, u256_slot, idx, post_value)
```
SamWilsn, line 450: Huh, why are we representing slots as `Bytes32` in some places, but `U256` in others? Shouldn't that be consistent everywhere?

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -95,7 +89,7 @@ def generic_create(
     if memory_size > U256(MAX_INIT_CODE_SIZE):
         raise OutOfGasError
 
-    state = evm.message.block_env.state
+    tx_tracker = evm.message.tx_env.tx_tracker
```
SamWilsn: More a comment for the `TransactionEnvironment` class, but I think including `tx_` is an inefficient use of space. We're already in the `TransactionEnvironment`, so unless there's another kind of tracker to differentiate, `tx_` is implied.

I think it'd be better to say what the tracker is tracking?

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -1123,20 +1076,12 @@ def increase_recipient_balance(recipient: Account) -> None:
...
-            block_env.state_changes,
-            wd.address,
-            new_balance,
-        )
-
-        if account_exists_and_is_empty(block_env.state, wd.address):
-            destroy_account(block_env.state, wd.address)
+        track_address(wd_tracker, wd.address)
+        current_balance = get_account(wd_tracker, wd.address).balance
+        new_balance = current_balance + wd.amount * U256(10**9)
```
SamWilsn: Named constant?

src/ethereum/forks/amsterdam/state.py
```diff
@@ -42,26 +40,76 @@ class State:
...
-    """
-    Contains all information that is preserved between message calls
-    within a transaction.
-    """
-
-    _tries: Dict[Address, Trie[Bytes32, U256]] = field(default_factory=dict)
-    _snapshots: List[Dict[Address, Trie[Bytes32, U256]]] = field(
-        default_factory=list
-    )
+    def get_account_optional(self, address: Address) -> Optional[Account]:
```
SamWilsn, line 43: You know my stance on `get_` prefixes 🤣 

They might make sense in this case though.

src/ethereum/forks/amsterdam/state.py
```diff
@@ -71,257 +119,63 @@ def close_state(state: State) -> None:
...
-    state._snapshots.pop()
-    if not state._snapshots:
-        state.created_accounts.clear()
-
-    transient_storage._snapshots.pop()
-
-
-def rollback_transaction(
-    state: State, transient_storage: TransientStorage
+def apply_diffs_to_state(
```
SamWilsn:
```suggestion
def apply_changes_to_state(
```

Since we use `_changes` in the parameter names below.

SamWilsn (commented):
Looks good for the most part. Just make sure to use markdown and not RST for new docstrings.

src/ethereum/forks/amsterdam/block_access_lists/builder.py
fselmo: There was an update to the EIP to use U256 to represent slots as this takes up less space in the BAL. So in the builder, when building the BAL, this is the expectation. See [EIP update](https://github.com/ethereum/EIPs/pull/10817).

src/ethereum/forks/amsterdam/state_tracker.py
gurukamath: Yes. The larger re-factor can be done in a different PR. However, I have gone ahead and dropped the `*_tracking` suffix from class, attribute and variable names in this PR, since it makes the names more concise. We now have `TransactionState` and `BlockState`.

As for the larger re-factor. I have created #2228

src/ethereum/forks/amsterdam/state_tracker.py
gurukamath: Updated the docstring. These functions will need to be updated to handle non-same tx destruction once we backport them to pre-Prague forks.

src/ethereum/forks/amsterdam/state_tracker.py
gurukamath: Yes. This one is on my radar as a next step. The current BAL implementation does it this way and I didn't want to introduce too many things in one PR

src/ethereum/forks/amsterdam/block_access_lists/builder.py
gurukamath: Good point. Did not see it that way myself but agree that this is could be confusing.

I think `_get_pre_tx_account` better describes what this is supposed to do. Have updated

src/ethereum/forks/amsterdam/block_access_lists/builder.py
gurukamath: Updated

src/ethereum/state.py
gurukamath: Updated

src/ethereum/state.py
gurukamath: On this one, I think `PreState` describes the dataclass. I am thinking of this as the state pre-block. It ties nicely with the state access chain `tx_state` -> `block_state` -> `pre_state` imo

src/ethereum/forks/amsterdam/block_access_lists/builder.py
gurukamath: Good catch. This is just left over from a former version of the PR that I was playing around with. Removed

src/ethereum/forks/amsterdam/block_access_lists/builder.py
gurukamath: Added the reference

src/ethereum/forks/amsterdam/block_access_lists/builder.py
gurukamath: Have used `pre_tx_accounts`

src/ethereum/forks/amsterdam/block_access_lists/builder.py
gurukamath: Updated

src/ethereum/forks/amsterdam/block_access_lists/builder.py
gurukamath: Have added comments to explain this

src/ethereum/forks/amsterdam/vm/instructions/system.py
gurukamath: Updated. This in combination with Peter's comment about dropping the `_tracker` suffix, now makes this look a lot cleaner. We now have

`block_state = block_env.state` and `tx_state = tx_env.state`

src/ethereum/forks/amsterdam/fork.py
gurukamath: Added `GWEI_TO_WEI` as a constant

src/ethereum/forks/amsterdam/state.py
gurukamath: Updated

src/ethereum/forks/amsterdam/block_access_lists/builder.py
gurukamath: Updated

kevaundray:
sudo failure seems to have been fixed by @SamWilsn in https://github.com/ethereum/execution-specs/pull/2232

fselmo:
> sudo failure seems to have been fixed

Yeah this was fixed. Re-started the CI on this now for a sanity check 👍🏼

kevaundray:
> > sudo failure seems to have been fixed
> 
> Yeah this was fixed. Re-started the CI on this now for a sanity check 👍🏼

Do we need to merge forks/amsterdam into this branch?

fselmo:
> Do we need to merge forks/amsterdam into this branch?

Ah yes of course we will need a rebase 😆

edit: Rebased 👍🏼. Was a clean rebase so I pushed here... but, just in case, [here](https://github.com/ethereum/execution-specs/pull/2207#event-22892415368) is the event with the `pre-rebase` -> `post-rebase` hashes.

src/ethereum/forks/amsterdam/fork_types.py
```diff
@@ -14,38 +14,18 @@
...
 
 from ethereum_rlp import rlp
-from ethereum_types.bytes import Bytes, Bytes20, Bytes256
+from ethereum_types.bytes import Bytes, Bytes256
 from ethereum_types.frozen import slotted_freezable
-from ethereum_types.numeric import U8, U64, U256, Uint
+from ethereum_types.numeric import U8, U64, U256
 
 from ethereum.crypto.hash import Hash32, keccak256
+from ethereum.state import Account, Address
```
kevaundray, line 22:
```suggestion
from ethereum.state import EMPTY_ACCOUNT, Account, Address
```

Could do this to fix the CI error, though its unused so would also likely need something like `__all__ = ("EMPTY_ACCOUNT",)` to indicate that its re-exported

kevaundray: Previously it was getting re-exported because it was defined in this file

kevaundray: Alternatively, it seems that we could change:

```python
return self._module("fork_types").EMPTY_ACCOUNT
```

for 

```python
 return self._module("state").EMPTY_ACCOUNT
```

Instead of re-exporting it from fork_types

gurukamath: Have just pushed a fix that should be able to fill `Amsterdam` as well as the older forks. This will no longer be required once we backport these changes to the older forks.

I'd still wait for the CI to be successful before merging. `json_infra` and `optimized` CI runs are not expected to be successful.

gurukamath:
As this PR gets ready to be merged, as an extra layer of check, I have run `uv run hasher compare` between the fixtures generated by commits `1ab92c600ac` and `e271298d390`. The fixtures generated match :)

gurukamath:
Based on the internal STEEL discussions, it was decided to remove the `optimized` run from CI with the aim of adding it back in once it is redesigned/refactored. I will create an issue to track progress on this.

marioevz (approved):
LGTM. I've only verified that the generated fixtures match exactly to the previous commit from this PR, so my approval comes from that.

If there's no outstanding comments from @SamWilsn or @petertdavies we should be good to merge 👍

src/ethereum/forks/amsterdam/vm/__init__.py
```diff
@@ -49,7 +50,7 @@ class BlockEnvironment:
     prev_randao: Bytes32
     excess_blob_gas: U64
     parent_beacon_block_root: Hash32
-    state_changes: StateChanges
+    bal_builder: BlockAccessListBuilder
```
SamWilsn: I don't think "bal" is a common enough abbreviation, but I'm happy to be overruled.

src/ethereum/forks/amsterdam/state.py
```diff
@@ -42,26 +39,76 @@ class State:
...
+                del storage_tries[address]
+
+        def get_storage_root(addr: Address) -> Root:
+            if addr in storage_tries:
+                return root(storage_tries[addr])
+            return EMPTY_TRIE_ROOT
+
+        state_root_value = root(main_trie, get_storage_root=get_storage_root)
+
+        # TODO: collect internal nodes for witness generation
```
SamWilsn: Should we remove this TODO and create an issue instead?

kevaundray: Can we create an issue without removing the TODO and thus restarting the CI?

src/ethereum/forks/amsterdam/vm/__init__.py
kevaundray: Can we create an issue and or change in a followup PR?

src/ethereum/forks/amsterdam/state.py
SamWilsn: Aw shit, #2254 got merged before this one.

kevaundray: haha I guess it makes sense to make these changes then, since we need to restart CI anyways

kevaundray: @SamWilsn if you're free, could you please rebase to restart the CI?

SamWilsn:
This should be good for re-review/hash comparison!

kevaundray:
`uv run hasher compare` between `a78af5e46`(latest on forks/amsterdam) and ea42efbcd show the fixtures match, though I welcome others to double check incase I made a mistake.

Noting that the `Build documentation`  job is currently failing

SamWilsn:
I missed the `Address` move; my bad!

fselmo (approved):
Confirmed one last time with the hasher, no changes across fixtures. Lgtm 👍🏼 

edit: will merge when `json_infra` finishes 🙃

PR #2220, opened 2026-02-16
feat(test-benchmark): add worst-case bloatnet benchmark tests

CPerezz:
## Summary

- Add 4 new worst-case stateful benchmark tests for bloatnet stress testing:
  - **CREATE2 + immediate access**: deploys via CREATE2 then immediately queries with EXTCODEHASH/BALANCE/EXTCODECOPY, stressing deploy-then-read trie paths (9 variants: 3 code sizes × 3 opcodes)
  - **TSTORE saturation**: writes unique transient storage keys each iteration with optional TLOAD readback (4 variants)
  - **SSTORE dirty transitions**: exercises 4 dirty-transition patterns (zero→nonzero→zero, nonzero→nonzero, etc.) × warm/cold via EIP-7702 delegation (8 variants)
  - **DELEGATECALL chain**: stub for future depth-based benchmark
- Fix gas accounting in all `test_multi_opcode.py` tests to use `fork.transaction_intrinsic_cost_calculator()` with actual calldata costs and `return_cost_deducted_prior_execution=True`
- Fix pre-existing kwarg bugs (`destOffset`→`dest_offset`, `access_warm`→`key_warm`)
- Restructure ERC-20 dispatch bytecodes for stack-checking compliance
- Add `FACTORY_STUBS` parametrization and multi-opcode BALANCE pairing to `test_bloatnet_balance_opcode`
- Add `skip_gas_used_validation=True` where precise gas prediction is infeasible (SSTORE refunds, CREATE2 subcalls, CALL stipend)

## Test plan

- [x] `fill --fork Prague` engine variants pass for all tests (test_multi_opcode: 92/92, test_create2_access: 9/9, test_single_opcode SSTORE dirty: 8/8, test_transient_storage: 4/4)
- [x] `execute remote` at 10M gas against geth dev node — all self-contained tests pass (Prague)
- [x] `execute remote` at 50M/100M gas with Osaka fork — tx splitting across multiple 16M-capped transactions works correctly
- [x] `execute remote` at 500M gas with Osaka — 61 transactions generated and executed successfully (TSTORE stress test)

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_create2_access.py
+++ b/tests/benchmark/stateful/bloatnet/test_create2_access.py
@@ -1,0 +1,236 @@
+"""
+abstract: CREATE2 deploy + immediate access benchmark cases.
+
+   These tests benchmark the deploy-then-access pattern: CREATE2 a
+   contract, then immediately query it with EXTCODEHASH, BALANCE, or
+   EXTCODECOPY in the same transaction. This tests whether clients
+   efficiently serve state that was just written to the trie.
+"""
+
+import pytest
+from execution_testing import (
+    Alloc,
+    BenchmarkTestFiller,
+    Block,
+    Bytecode,
+    Fork,
+    Hash,
+    Initcode,
+    Op,
+    Transaction,
+    While,
+)
+
+REFERENCE_SPEC_GIT_PATH = "DUMMY/bloatnet.md"
+REFERENCE_SPEC_VERSION = "1.0"
+
+
+# CREATE2 + ACCESS BENCHMARK ARCHITECTURE:
+#
+#   [Init Code Holder Contract] ──── Runtime code = init code bytes
+#           │
+#           │  EXTCODECOPY by attack contract during setup
+#           │
+#   [Attack Contract]
+#       │ Setup:
+#       │   1. EXTCODECOPY init code from holder into MEM[0..N]
+#       │   2. Store starting counter at MEM[N..N+32]
+#       │
+#       │ Loop(i=0 to M):
+#       │   1. CREATE2(value=0, offset=0, size=N, salt=counter)
+#       │      → deploys new contract, returns address
+#       │   2. EXTCODEHASH / BALANCE / EXTCODECOPY on address
+#       │   3. Increment counter
+#
+# WHY IT STRESSES CLIENTS:
+#   - Each CREATE2 inserts a new account + code into the trie
+#   - Immediate access tests if the just-written data is efficiently
+#     served from write caches vs requiring a trie re-read
+#   - Code deposit cost (200 gas/byte) dominates: larger code =
+#     fewer iterations but more trie data per cycle
+
+
+@pytest.mark.parametrize(
+    "code_size",
+    [32, 256, 1024],
+    ids=["32B", "256B", "1KB"],
+)
+@pytest.mark.parametrize(
+    "access_opcode",
+    ["EXTCODEHASH", "BALANCE", "EXTCODECOPY"],
+)
+def test_create2_immediate_access(
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+    tx_gas_limit: int,
+    code_size: int,
+    access_opcode: str,
+) -> None:
+    """
+    Benchmark CREATE2 followed by immediate opcode access.
+
+    Deploy a contract via CREATE2, then immediately query it with the
+    specified access opcode. Each iteration creates a new trie entry
+    and reads from it, stressing the deploy-then-access path.
+    """
+    # Build init code that deploys `code_size` bytes of zeros
+    deploy_code = bytes(code_size)
+    initcode = Initcode(deploy_code=deploy_code)
+    init_code_bytes = bytes(initcode)
+    init_code_size = len(init_code_bytes)
+
+    # Deploy holder contract whose runtime code IS the init code
+    init_holder = pre.deploy_contract(
+        code=Bytecode(
+            init_code_bytes,
+            popped_stack_items=0,
+            pushed_stack_items=0,
+        ),
+    )
+
+    # Memory layout:
+    #   MEM[0 .. init_code_size-1]     = init code (for CREATE2)
+    #   MEM[init_code_size .. +31]     = counter (salt)
+    counter_offset = init_code_size
+
+    # Setup: load init code + starting counter
+    setup = (
+        Op.EXTCODECOPY(
+            address=init_holder,
+            dest_offset=0,
+            offset=0,
+            size=init_code_size,
+            address_warm=False,
+            data_size=init_code_size,
+            old_memory_size=0,
+            new_memory_size=init_code_size,
+        )
+        + Op.MSTORE(
+            counter_offset,
+            Op.CALLDATALOAD(32),
+            old_memory_size=init_code_size,
+            new_memory_size=counter_offset + 32,
+        )
+        + Op.CALLDATALOAD(0)  # [num_iters]
+    )
+
+    # CREATE2 — deploys new contract each iteration
+    create2_op = Op.CREATE2(
+        value=0,
+        offset=0,
+        size=init_code_size,
+        salt=Op.MLOAD(counter_offset),
+        init_code_size=init_code_size,
+        old_memory_size=counter_offset + 32,
+        new_memory_size=counter_offset + 32,
+    )
+
+    # Access the just-deployed contract
+    if access_opcode == "EXTCODEHASH":
+        access_op = Op.POP(
+            Op.EXTCODEHASH(create2_op, address_warm=True)
+        )
+    elif access_opcode == "BALANCE":
+        access_op = Op.POP(
+            Op.BALANCE(create2_op, address_warm=True)
+        )
+    elif access_opcode == "EXTCODECOPY":
+        # Copy 1 byte from end of deployed code
+        access_op = Op.EXTCODECOPY(
+            address=create2_op,
+            dest_offset=counter_offset + 32,
+            offset=max(code_size - 1, 0),
+            size=1,
+            address_warm=True,
+            data_size=1,
+            old_memory_size=counter_offset + 32,
+            new_memory_size=counter_offset + 64,
+        )
+    else:
+        raise ValueError(f"Unsupported opcode: {access_opcode}")
+
+    # Increment counter
+    increment = Op.MSTORE(
+        counter_offset,
+        Op.ADD(Op.MLOAD(counter_offset), 1),
+    )
+
+    loop = While(
+        body=access_op + increment,
+        condition=(
+            Op.PUSH1(1)
+            + Op.SWAP1
+            + Op.SUB
+            + Op.DUP1
+            + Op.ISZERO
+            + Op.ISZERO
+        ),
+    )
+
+    code = setup + loop
+    attack_contract_address = pre.deploy_contract(code=code)
+
+    # Gas Accounting — CREATE2 subcall costs (init execution + code
+    # deposit) are not captured by loop.gas_cost(), so add them.
+    setup_cost = setup.gas_cost(fork)
+    loop_cost = loop.gas_cost(fork)
+
+    init_exec_gas = initcode.execution_gas(fork)
+    code_deposit_gas = initcode.deployment_gas(fork)
+    full_iteration_cost = (
+        loop_cost + init_exec_gas + code_deposit_gas
+    )
+
+    intrinsic_cost_calc = fork.transaction_intrinsic_cost_calculator()
+    max_intrinsic = intrinsic_cost_calc(calldata=b"\xff" * 64)
+
+    # Attack Loop
+    gas_remaining = gas_benchmark_value
+    txs = []
+    counter_start = 0
+    total_gas_consumed = 0
+
+    while gas_remaining > max_intrinsic + setup_cost + full_iteration_cost:
+        gas_available = min(gas_remaining, tx_gas_limit)
+
+        if gas_available < max_intrinsic + setup_cost:
+            break
+
+        num_iters = (
+            gas_available - max_intrinsic - setup_cost
+        ) // full_iteration_cost
+
+        if num_iters == 0:
+            break
+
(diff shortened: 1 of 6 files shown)


CPerezz:
cc: @LouisTsai-Csie @marioevz This is bloatnet related.


cc: @jochem-brouwer in case you can give some suggestions

tests/benchmark/stateful/bloatnet/test_create2_access.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Deploy a contract via CREATE2, then immediately query it with the
+    specified access opcode. Each iteration creates a new trie entry
+    and reads from it, stressing the deploy-then-access path.
+    """
+    # Build init code that deploys `code_size` bytes of zeros
+    deploy_code = bytes(code_size)
+    initcode = Initcode(deploy_code=deploy_code)
+    init_code_bytes = bytes(initcode)
+    init_code_size = len(init_code_bytes)
```
LouisTsai-Csie, line 85: I’m not fully sure about the intended logic here. For example, if `code_size` = 256, should the `initcode` be:

1. `Op.STOP` * 256 (i.e., `Op.STOP` * code_size), or
2. a literal 256 (i.e., `code_size`)?

iiuc, the deployed contract code produced by this `initcode` corresponds to option 2, but based on the comment, it seems like you want option 1?

tests/benchmark/stateful/bloatnet/test_create2_access.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+@pytest.mark.parametrize(
+    "code_size",
+    [32, 256, 1024],
+    ids=["32B", "256B", "1KB"],
+)
+@pytest.mark.parametrize(
+    "access_opcode",
+    ["EXTCODEHASH", "BALANCE", "EXTCODECOPY"],
```
LouisTsai-Csie: Just for style consistency, could we use this for all the parametrization? Thanks
```suggestion
    [Op.EXTCODEHASH, Op.BALANCE, Op.EXTCODECOPY],
```

tests/benchmark/stateful/bloatnet/test_create2_access.py
```diff
@@ -0,0 +1,190 @@
...
+        attack_contract_address=attack_contract_address,
+        setup_cost=setup.gas_cost(fork),
+        iteration_cost=full_iteration_cost,
+    )
+
+    benchmark_test(
+        pre=pre,
+        blocks=[Block(txs=txs)],
+        expected_benchmark_gas_used=total_gas_consumed,
+        skip_gas_used_validation=True,
```
LouisTsai-Csie: If `skip_gas_used_validation` is enabled, then the total gas consumed would not be checked at the end of the test run. I suggest remove it, and check the gas consumption as safety check:

```suggestion
        expected_benchmark_gas_used=total_gas_consumed,
```

LouisTsai-Csie (changes requested):
I like the new benchmark! And the refactor to the existing tests is good!

Leave some general suggestion for the benchmark implementation below instead of commenting on each of them:

**gas validation rule**
I notice `expected_benchmark_gas_used` and `skip_gas_used_validation` exist as parameter for `benchmark_test` at the same time, but if you configure `skip_gas_used_validation`, then the framework will skip the gas usage check.

I’d suggest removing `skip_gas_used_validation=True` whenever we can precisely calculate the expected gas usage, so we can verify that the benchmark’s gas accounting is working correctly. This should be feasible for most benchmarks that don’t rely on stub accounts. For benchmarks that do use stub addresses, feel free to keep `skip_gas_used_validation` enabled in cases where the gas calculation is non-trivial.

**parametrization style consistency**
This is not a big issue, but to align the rest of the test style. For the parametrization:

```python
@pytest.mark.parametrize(
    "second_opcode",
    ["EXTCODESIZE", "EXTCODECOPY", "EXTCODEHASH", "STATICCALL", "CALL"],
)
```

Instead of using string, could we use `Opcode` object instead? It would be like below:

```python
@pytest.mark.parametrize(
    "second_opcode",
    [Op.EXTCODESIZE, Op.EXTCODECOPY, Op.EXTCODEHASH, Op.STATICCALL, Op.CALL],
)
```

I am still reviewing `test_sstore_dirty_transitions` as i suspect the logic is not working as expected, but need more time on this one

tests/benchmark/stateful/bloatnet/test_create2_access.py
CPerezz: Notice my code will just do the same as placing Op.STOP * 256 (as it will all be empty 0's.)

If you prefer, I can refactor it to be more clear

tests/benchmark/stateful/bloatnet/test_create2_access.py
CPerezz: I tried removing this — the fill fails with ~550 gas discrepancy on several variants (e.g., gas_used=9996913 vs expected=9997463). The issue
   is the 63/64 rule for `CREATE2 `subcalls: the child execution doesn't receive exactly the gas we predict via `initcode.execution_gas() + initcode.deployment_gas()`.
   
   Any suggestions?

tests/benchmark/stateful/bloatnet/test_create2_access.py
CPerezz: Done

CPerezz:
> I like the new benchmark! And the refactor to the existing tests is good!
> 
> Leave some general suggestion for the benchmark implementation below instead of commenting on each of them:
> 
> **gas validation rule** I notice `expected_benchmark_gas_used` and `skip_gas_used_validation` exist as parameter for `benchmark_test` at the same time, but if you configure `skip_gas_used_validation`, then the framework will skip the gas usage check.
> 
> I’d suggest removing `skip_gas_used_validation=True` whenever we can precisely calculate the expected gas usage, so we can verify that the benchmark’s gas accounting is working correctly. This should be feasible for most benchmarks that don’t rely on stub accounts. For benchmarks that do use stub addresses, feel free to keep `skip_gas_used_validation` enabled in cases where the gas calculation is non-trivial.
> 
> **parametrization style consistency** This is not a big issue, but to align the rest of the test style. For the parametrization:
> 
> ```python
> @pytest.mark.parametrize(
>     "second_opcode",
>     ["EXTCODESIZE", "EXTCODECOPY", "EXTCODEHASH", "STATICCALL", "CALL"],
> )
> ```
> 
> Instead of using string, could we use `Opcode` object instead? It would be like below:
> 
> ```python
> @pytest.mark.parametrize(
>     "second_opcode",
>     [Op.EXTCODESIZE, Op.EXTCODECOPY, Op.EXTCODEHASH, Op.STATICCALL, Op.CALL],
> )
> ```
> 
> I am still reviewing `test_sstore_dirty_transitions` as i suspect the logic is not working as expected, but need more time on this one


Parametrization style: Done — converted all string-based opcode parametrizations to Op.* objects in both `test_create2_access.py` and `test_multi_opcode.py`.

`skip_gas_used_validation`: Awaiting to see what your thoughts are. LMK

tests/benchmark/stateful/bloatnet/test_create2_access.py
```diff
@@ -0,0 +1,190 @@
...
+        # Copy 1 byte from end of deployed code
+        access_op = Op.EXTCODECOPY(
+            address=create2_op,
+            dest_offset=counter_offset + 32,
+            offset=max(code_size - 1, 0),
+            size=1,
+            address_warm=True,
+            data_size=1,
+            old_memory_size=counter_offset + 32,
+            new_memory_size=counter_offset + 64,
```
LouisTsai-Csie:
```suggestion
            new_memory_size=counter_offset + 33,
```

This is the root cause for gas mismatch! You only copy 1 byte from the target

tests/benchmark/stateful/bloatnet/test_create2_access.py
LouisTsai-Csie: Ah sorry, after reviewing the implementation again, i found my previous understanding is wrong. It's all good now.

tests/benchmark/stateful/bloatnet/test_create2_access.py
```diff
@@ -0,0 +1,190 @@
...
+
+    txs, total_gas_consumed = build_benchmark_txs(
+        pre=pre,
+        fork=fork,
+        gas_benchmark_value=gas_benchmark_value,
+        tx_gas_limit=tx_gas_limit,
+        attack_contract_address=attack_contract_address,
+        setup_cost=setup.gas_cost(fork),
+        iteration_cost=full_iteration_cost,
+    )
```
LouisTsai-Csie: I suggest refactoring this section to use a feature we recently added and already use in `test_account_query`: `IteratingBytecode`.

`IteratingBytecode` is an wrapper for loop byte code that calculate gas costs across repeated iterations, including subcalls. In this case, it cleanly handles the `CREATE2` gas-accounting issue:

```python
code = IteratingBytecode(
    setup=setup,                      # one-time overhead
    iterating=loop,                   # the while loop (per-iteration gas via .gas_cost())
    iterating_subcall=subcall_cost,   # CREATE2 init execution + code deposit (not part of the loop bytecode)
)
```

The key feature here is `iterating_subcall`, which tells the gas model that each iteration performs a subcall costing `subcall_cost` gas. Based on that, `IteratingBytecode` automatically:

1. Computes the 63/64-rule reserve (`iterating_subcall_reserve`) so the final CREATE2 subcall does not run out of gas.
2. Generates transactions via `transactions_by_gas_limit`.

With this refactor, the code for this section would become:
```python
subcall_cost = (
    initcode.execution_gas(fork) + initcode.deployment_gas(fork)
)

code = IteratingBytecode(
    setup=setup,
    iterating=loop,
    iterating_subcall=subcall_cost,
)
attack_contract_address = pre.deploy_contract(code=code)

def calldata_builder(
    iteration_count: int, start_iteration: int
) -> bytes:
    return bytes(Hash(iteration_count) + Hash(start_iteration))

txs = list(
    code.transactions_by_gas_limit(
        fork=fork,
        gas_limit=gas_benchmark_value,
        sender=pre.fund_eoa(),
        to=attack_contract_address,
        calldata=calldata_builder,
    )
)
```

With this, we could enable `skip_gas_used_validation` to True, as the wrapper automatically ensure there is no OOG and maximize the loop.

I’ve attached the full file for reference.

https://gist.github.com/LouisTsai-Csie/0641b8a986948bceefc8bdfdc13cd42e

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -106,12 +121,54 @@ def test_bloatnet_balance_extcodesize(
...
+        other_op = Op.POP(
+            Op.STATICCALL(
+                gas=1,
+                address=create2_preimage.address_op(),
+                args_offset=0,
+                args_size=0,
+                ret_offset=0,
+                ret_size=0,
+            )
+        )
```
LouisTsai-Csie: We DUP the address in the loop, see below:

```python
loop = While(
        body=(
            create2_preimage.address_op()
            + Op.DUP1 # Address is duplicated here!
            + benchmark_ops
            + create2_preimage.increment_salt_op()
        ),
        condition=DECREMENT_COUNTER_CONDITION,
    )
```

We could just fetch this value from stack, but we need to do additional `POP` to make sure the stack is balanced.

```suggestion
        other_op = Op.POP(
            Op.STATICCALL(
                gas=1,
                address=Op.DUP5, # This could save gas
                args_offset=0,
                args_size=0,
                ret_offset=0,
                ret_size=0,
            )
        ) + Op.POP # This is needed as one value is DUP
```

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -106,12 +121,54 @@ def test_bloatnet_balance_extcodesize(
...
+            Op.CALL(
+                gas=1,
+                address=create2_preimage.address_op(),
+                value=0,
+                args_offset=0,
+                args_size=0,
+                ret_offset=0,
+                ret_size=0,
+            )
+        )
```
LouisTsai-Csie: Same as above

```suggestion
        other_op = Op.POP(
            Op.CALL(
                gas=1,
                address=Op.DUP6,
                value=0,
                args_offset=0,
                args_size=0,
                ret_offset=0,
                ret_size=0,
            )
        ) + POP
```

tests/benchmark/stateful/bloatnet/test_multi_opcode.py
```diff
@@ -121,81 +178,73 @@ def test_bloatnet_balance_extcodesize(
...
-            )
-        )
-
-        gas_remaining -= gas_available
-        salt_offset += num_contract
-
     benchmark_test(
         pre=pre,
         blocks=[Block(txs=txs)],
+        expected_benchmark_gas_used=total_gas_consumed,
```
LouisTsai-Csie, line 208:
```suggestion
       
```

I realize that it is impossible to calculate the gas cost precisely for this test, as we could not estimate the gas cost for the first `STATICCALL` in EELS framework. We could skip the validation here.

LouisTsai-Csie (changes requested):
Thanks! Resolve some issue regarding the gas accounting.

tests/benchmark/stateful/bloatnet/test_create2_access.py
LouisTsai-Csie: I add a suggested changes below, ptal!

tests/benchmark/stateful/helpers.py
```diff
@@ -1,7 +1,20 @@
...
+from execution_testing import (
+    Address,
+    Alloc,
+    Hash,
+    Op,
+    Transaction,
+)
+
+if TYPE_CHECKING:
+    from execution_testing import AccessList, Fork
```
LouisTsai-Csie:
```suggestion
from execution_testing import (
    AccessList,
    Address,
    Alloc,
    Fork,
    Hash,
    Op,
    Transaction,
)
```
Do we have circular import here? If not, we could remove the `TYPE_CHECKING` here.

LouisTsai-Csie (changes requested):
One more comment!

LouisTsai-Csie (approved):
LGTM

Issue #2223, opened 2026-02-17
Consider reducing fixture size for `account_query` (16 GB) and `unchunkified_bytecode` (13 GB)


danceratopz:
## Summary

Grab the latest artifact like this:
```
gh run download -R ethereum/execution-specs -n fixtures_benchmark_fast -D /mnt/eest_artifacts
```

Two benchmark fixture files are extreme size outliers — 100-1000× larger than all others. This issue documents the specific redundancies driving the size and suggests possible approaches to reduce it.

## Top-10 Benchmark Fixture Sizes

| Rank | Fixture | Size |
|------|---------|------|
| 1 | `account_query.json` | **16 GB** |
| 2 | `unchunkified_bytecode.json` | **13 GB** |
| 3 | `block_full_of_ether_transfers.json` | 123 MB |
| 4 | `block_full_data.json` | 42 MB |
| 5 | `auth_transaction.json` | 19 MB |
| 6 | `block_full_of_precompile.json` | 15 MB |
| 7 | `block_full_of_sha256_hashes.json` | 14 MB |
| 8 | `secp256k1.json` | 13 MB |
| 9 | `kzg_point_evaluation.json` | 13 MB |
| 10 | `modexp.json` | 9.3 MB |

The gap between #2 (13 GB) and #3 (123 MB) is >100×. The remaining fixtures (#3–#10) are proportionate to their workload.

---

## Outlier 1: `account_query.json` (16 GB)

**Test:** [`test_account_query`](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/tests/benchmark/compute/instruction/test_account_query.py#L428) at `tests/benchmark/compute/instruction/test_account_query.py:428`

**Parametrization:** 8 opcodes × 2 `access_warm` × 4 `mem_size` × 5 `code_size` × 2 `value_sent` = 640 theoretical combinations, reduced to ~94 by skip logic (lines 440–452).

**Root cause:** The 8 EXTCODECOPY + `max_code_size` cases (4 `mem_size` × 2 `access_warm`) each generate ~1.8 GB, totalling ~14.4 GB (roughly 96% of the file). Each case deploys ~44K contracts of 24,576 bytes via `CustomSizedContractFactory`.

#### 1. `mem_size=0` with EXTCODECOPY is a no-op copy

[...]

PR #2226, opened 2026-02-17
fix(test-specs): Exclude full post-state/setup-tx receipts on benchmarks

marioevz:
## 🗒️ Description

Adds defaults to `BenchmarkTest` to exclude the full post state and the setup tx receipts in the output by default, in contrast to the `BlockchainTest` defaults which is to include them both.

Drops the size of the top two biggest fixtures:
- `account_query.json`: `16 GB` -> `2.0G`
- `unchunkified_bytecode.json`: `13 GB` -> `350M`

But of course the rest of the fixtures are also affected.

## 🔗 Related Issues or PRs
#2223

## ✅ Checklist
    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/specs/blockchain.py
+++ b/packages/testing/src/execution_testing/specs/blockchain.py
@@ -378,7 +378,9 @@ class BuiltBlock(CamelModel):
     fork: Fork
     block_access_list: BlockAccessList | None
 
-    def get_fixture_block(self) -> FixtureBlock | InvalidFixtureBlock:
+    def get_fixture_block(
+        self, *, include_receipts: bool = True
+    ) -> FixtureBlock | InvalidFixtureBlock:
         """Get a FixtureBlockBase from the built block."""
         fixture_block = FixtureBlockBase(
             header=self.header,
@@ -398,7 +400,7 @@ class BuiltBlock(CamelModel):
                     )
                     for i, r in enumerate(self.result.receipts)
                 ]
-                if self.result.receipts
+                if self.result.receipts and include_receipts
                 else None
             ),
             block_access_list=self.block_access_list
@@ -499,10 +501,14 @@ class BlockchainTest(BaseTest):
     blocks: List[Block]
     genesis_environment: Environment = Field(default_factory=Environment)
     chain_id: int = 1
-    exclude_full_post_state_in_output: bool = False
+    include_full_post_state_in_output: bool = True
     """
-    Exclude the post state from the fixture output. In this case, the state
+    Include the post state in the fixture output. Otherwise, the state
     verification is only performed based on the state root.
+    """
+    include_intermediate_block_tx_receipts: bool = True
+    """
+    Include transaction receipts in blocks prior to the last block.
     """
 
     _benchmark_opcode_count: OpcodeCount | None = PrivateAttr(None)
@@ -875,14 +881,22 @@ class BlockchainTest(BaseTest):
             # This is the most common case, the RLP needs to be constructed
             # based on the transactions to be included in the block.
             # Set the environment according to the block to execute.
+            is_last_block = i == len(self.blocks) - 1
             built_block = self.generate_block_data(
                 t8n=t8n,
                 block=block,
                 previous_env=env,
                 previous_alloc=alloc,
-                last_block=i == len(self.blocks) - 1,
+                last_block=is_last_block,
             )
-            fixture_blocks.append(built_block.get_fixture_block())
+            include_receipts = (
+                is_last_block or self.include_intermediate_block_tx_receipts
+            )
+            fixture_blocks.append(
+                built_block.get_fixture_block(
+                    include_receipts=include_receipts
+                )
+            )
 
             # BAL verification already done in to_fixture_bal() if
             # expected_block_access_list set
@@ -918,10 +932,10 @@ class BlockchainTest(BaseTest):
             last_block_hash=head,
             pre=pre,
             post_state=alloc
-            if not self.exclude_full_post_state_in_output
+            if self.include_full_post_state_in_output
             else None,
             post_state_hash=state_root
-            if self.exclude_full_post_state_in_output
+            if not self.include_full_post_state_in_output
             else None,
             config=FixtureConfig(
                 fork=self.fork,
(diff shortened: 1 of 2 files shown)


packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -499,11 +501,15 @@ class BlockchainTest(BaseTest):
...
     genesis_environment: Environment = Field(default_factory=Environment)
     chain_id: int = 1
-    exclude_full_post_state_in_output: bool = False
+    include_full_post_state_in_output: bool = True
     """
-    Exclude the post state from the fixture output. In this case, the state
+    Include the post state in the fixture output. Otherwise, the state
     verification is only performed based on the state root.
     """
+    include_intermediate_block_tx_receipts: bool = True
```
jochem-brouwer: In benchmarks, I don't think we care about the tx receipts either in the output. The receipts are generated by the execution, so the receiptsRoot is calculated by this execution and is then checked against the provided block header. These should be correct (the tests are aimed for performance, not "weird" edge cases in consensus). For consensus if you get a receipsRoot mismatch you want the content of the receipts to quickly debug.

Regarding last block, if this is intended to target "the benchmark block" then this should be relaxed: it should include receipts for all blocks which have a test in them marked for the execution phase (these are benchmark blocks). We don't have multi-block benchmarks yet but this is definitely something which could happen in the future, I think this is something we should keep in mind.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -875,14 +881,22 @@ def make_fixture(
             # This is the most common case, the RLP needs to be constructed
             # based on the transactions to be included in the block.
             # Set the environment according to the block to execute.
+            is_last_block = i == len(self.blocks) - 1
```
jochem-brouwer: Nit: Is the comment above meant for the line below this one?

jochem-brouwer (commented):
One question + one nit

packages/testing/src/execution_testing/specs/blockchain.py
marioevz: Got it, I think then we have to make this a per-block property then, with a default that varies given the test type:
If `block.receipts_in_output is None`, use the `BlockchainTest.receipts_in_output` or `BenchmarkTest.receipts_in_output` default.
That way we can exclude them by default but in the future `BenchmarkTest` can apply different defaults depending on the block type (setup or execute).

packages/testing/src/execution_testing/specs/blockchain.py
marioevz: Yes, I've moved this line to before the comment 👍

marioevz:
@jochem-brouwer I've addressed all comments. As for including receipts for execution blocks, I think we have to defer that to a follow up PR because we currently do not mark blocks as execution/setup, we only do that for transactions and I think we should rethink how to approach this anyway, but it's perhaps out-of-scope for this PR.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -918,10 +939,10 @@ def make_fixture(
             last_block_hash=head,
             pre=pre,
             post_state=alloc
-            if not self.exclude_full_post_state_in_output
+            if self.include_full_post_state_in_output
             else None,
             post_state_hash=state_root
-            if self.exclude_full_post_state_in_output
+            if not self.include_full_post_state_in_output
```
jochem-brouwer, line 945: This does not look correct. We always want post state root

jochem-brouwer: Or is this a different hash than the block header state root? If so then dismiss this comment

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1004,7 +1025,7 @@ def make_hive_fixture(
             "payloads": fixture_payloads,
             "last_block_hash": head_hash,
             "post_state_hash": state_root
-            if self.exclude_full_post_state_in_output
+            if not self.include_full_post_state_in_output
```
jochem-brouwer, line 1028: Hmm no seems like this is the state root. Should always be reported as it is part of block header

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -880,9 +892,18 @@ def make_fixture(
                 block=block,
                 previous_env=env,
                 previous_alloc=alloc,
-                last_block=i == len(self.blocks) - 1,
+                last_block=is_last_block,
+            )
+            include_receipts = (
+                block.include_receipts_in_output
+                if block.include_receipts_in_output is not None
+                else self.include_tx_receipts_in_output
```
jochem-brouwer, line 900: To check: we use the (new?) property of block if it exists (override at block level?) and if it does not exist the setting of the test type runner (benchmark or fixture)?

jochem-brouwer (commented):
Agree with comment above that such change should be part of a future PR. Some more comments though

packages/testing/src/execution_testing/specs/blockchain.py
marioevz: That's right, the state root is included in the header regardless of the value of this field.

packages/testing/src/execution_testing/specs/blockchain.py
marioevz: This is at the test level in the final fixture, not at the block level. At the block level the state root hash is always included.

In a way this value is redundant because we always include the `last_block_hash`, which maps to a block inside of the blocks list, and you can retrieve the state root from there.

packages/testing/src/execution_testing/specs/blockchain.py
marioevz: That's correct. The idea is that `BenchmarkTest` can use either option to control if we want the receipts in the block or not. By default no receipts are included in any block for benchmark tests.

jochem-brouwer (approved):
With the extra clarifications it seems like all is correct, LGTM

felix314159 (approved):

PR #2241, opened 2026-02-19
feat(zkevm): Add IncrementalMPT

jsign:
## 🗒️ Description
This PR adds a new `IncrementalMPT` construction, which allows for a full in-memory MPT.

This is useful for the upcoming `ExecutionWitness` generation, since during STF we need to:
- Build pre-state MPT proofs for all state accessed during execution.
- During post-state root calculation, any branch compression might require including extra MPT nodes i.e. sibilings, when a deletion makes a `BranchNode` have a single child.

This code was pulled from the latest version of my branch, which:
- Fills all tests with a safety check that post-state root calculations matches `patricialize` results -- this gives confidence that workings are correct compared to `patricialize`.
- The ExecutionWitness nodes captured from `IncrementalMPT` were compared against an external production-EL (Reth), and (after many fixes) they both match -- so just an extra check that things look okay.

I pulled `IncrementalMPT` code into a new file, compared to what I did in my branch of including this in `trie.py`. My logic here is that adding all this to a new file can minimise rebasing conflicts. Happy to move around if you feel it's better.

### DIFF
--- a/src/ethereum/forks/amsterdam/incremental_mpt.py
+++ b/src/ethereum/forks/amsterdam/incremental_mpt.py
@@ -1,0 +1,802 @@
+"""
+Incremental Merkle Patricia Trie.
+
+.. contents:: Table of Contents
+    :backlinks: none
+    :local:
+
+Introduction
+------------
+
+Provide a mutable MPT that supports incremental updates and
+witness tracking. The tree structure is updated in-place rather
+than rebuilt from scratch on each root calculation.
+"""
+
+from dataclasses import dataclass, field
+from typing import (
+    Callable,
+    Dict,
+    Generic,
+    List,
+    Mapping,
+    MutableMapping,
+    Optional,
+    Tuple,
+    Union,
+)
+
+from ethereum_rlp import Extended, rlp
+from ethereum_types.bytes import Bytes
+from ethereum_types.numeric import Uint
+
+from ethereum.crypto.hash import keccak256
+
+from .fork_types import Account, Address, Root
+from .trie import (
+    EMPTY_TRIE_ROOT,
+    K,
+    V,
+    _prepare_data,
+    bytes_to_nibble_list,
+    common_prefix_length,
+    encode_node,
+    nibble_list_to_compact,
+)
+
+
+@dataclass
+class MutableLeafNode:
+    """Mutable leaf node in the Merkle Trie for in-place updates."""
+
+    rest_of_key: Bytes
+    value: Bytes
+    _hash: Optional[Bytes] = None
+    _rlp: Optional[Bytes] = None
+    _dirty: bool = False
+
+
+@dataclass
+class MutableExtensionNode:
+    """Mutable extension node in the Merkle Trie for in-place updates."""
+
+    key_segment: Bytes
+    child: "MutableNode"
+    _hash: Optional[Bytes] = None
+    _rlp: Optional[Bytes] = None
+    _dirty: bool = False
+
+
+@dataclass
+class MutableBranchNode:
+    """Mutable branch node in the Merkle Trie for in-place updates."""
+
+    children: List[Optional["MutableNode"]]
+    value: Bytes
+    _hash: Optional[Bytes] = None
+    _rlp: Optional[Bytes] = None
+    _dirty: bool = False
+
+
+MutableNode = Union[
+    MutableLeafNode, MutableExtensionNode, MutableBranchNode, None
+]
+
+
+@dataclass
+class Witness:
+    """Track nodes accessed during trie operations for witness generation."""
+
+    accessed_nodes: Dict[Bytes, Bytes] = field(default_factory=dict)
+
+
+@dataclass
+class IncrementalMPT(Generic[K, V]):
+    """
+    An MPT that supports incremental updates and witness tracking.
+
+    Maintain an actual tree structure that can be updated in-place,
+    rather than rebuilding the entire tree on each root calculation.
+    """
+
+    secured: bool
+    default: V
+    root_node: MutableNode = None
+    witness: Witness = field(default_factory=Witness)
+    _data: Dict[K, V] = field(default_factory=dict)
+
+
+def _build_mutable_tree(
+    obj: Mapping[Bytes, Bytes], level: Uint
+) -> MutableNode:
+    """
+    Build a mutable tree structure from a prepared key-value mapping.
+
+    Similar to ``patricialize()`` but create mutable nodes for
+    in-place updates.
+
+    Parameters
+    ----------
+    obj :
+        Underlying trie key-value pairs, with keys in
+        nibble-list format.
+    level :
+        Current trie level.
+
+    Returns
+    -------
+    node : `MutableNode`
+        Root node of the mutable tree.
+
+    """
+    if len(obj) == 0:
+        return None
+
+    arbitrary_key = next(iter(obj))
+
+    if len(obj) == 1:
+        return MutableLeafNode(
+            rest_of_key=arbitrary_key[level:],
+            value=obj[arbitrary_key],
+        )
+
+    substring = arbitrary_key[level:]
+    prefix_length = len(substring)
+    for key in obj:
+        prefix_length = min(
+            prefix_length,
+            common_prefix_length(substring, key[level:]),
+        )
+        if prefix_length == 0:
+            break
+
+    if prefix_length > 0:
+        prefix = arbitrary_key[int(level) : int(level) + prefix_length]
+        child = _build_mutable_tree(obj, level + Uint(prefix_length))
+        return MutableExtensionNode(key_segment=prefix, child=child)
+
+    branches: List[MutableMapping[Bytes, Bytes]] = [{} for _ in range(16)]
+    value = b""
+
+    for key in obj:
+        if len(key) == level:
+            value = obj[key]
+        else:
+            branches[key[level]][key] = obj[key]
+
+    children: List[Optional[MutableNode]] = [
+        _build_mutable_tree(branches[k], level + Uint(1)) for k in range(16)
+    ]
+
+    return MutableBranchNode(children=children, value=value)
+
+
+def build_mpt(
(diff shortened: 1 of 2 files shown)


src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+
+@dataclass
+class MutableLeafNode:
+    """Mutable leaf node in the Merkle Trie for in-place updates."""
+
+    rest_of_key: Bytes
+    value: Bytes
+    _hash: Optional[Bytes] = None
+    _rlp: Optional[Bytes] = None
+    _dirty: bool = False
```
jsign, line 56: `_dirty` is just a field that allows to identify all MPT node types if they were created or edited during the execution.

This feature is used whenever we are building the final `ExecutionWitness` such that we only put in it MPT nodes that `!_dirty`. As in, only nodes that existed in the pre-state MPT. e.g. including a sibiling during a branch compression if that sibiling was created during execution wouldn't be correct.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+MutableNode = Union[
+    MutableLeafNode, MutableExtensionNode, MutableBranchNode, None
+]
+
+
+@dataclass
+class Witness:
+    """Track nodes accessed during trie operations for witness generation."""
+
+    accessed_nodes: Dict[Bytes, Bytes] = field(default_factory=dict)
```
jsign, line 90: I use this `Witness` wrapper just to have a bit more signal that `accessed_nodes` is related for witness generation. If feels isn't useful, I can remove it and just put `accessed_nodes` directly as an `IncrementalMPT` field.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+
+    Maintain an actual tree structure that can be updated in-place,
+    rather than rebuilding the entire tree on each root calculation.
+    """
+
+    secured: bool
+    default: V
+    root_node: MutableNode = None
+    witness: Witness = field(default_factory=Witness)
+    _data: Dict[K, V] = field(default_factory=dict)
```
jsign, line 106: I think when we start working in Project 2, we can remove this `_data` since probably we want to re-use the same `IncrementalMPT` in a stateless STF.

For now I'm leaving it just to pull exact same code I've used in my other branch. The usage now isn't related to anything critical -- just mainly in `mpt_get` to get the value directly instead of pulling it from the LeafNode.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+
+
+def _invalidate_hash(node: MutableNode) -> None:
+    """Invalidate the cached hash of a node."""
+    if node is not None:
+        node._hash = None
+        node._rlp = None
+
+
+def _record_witness(
```
jsign, line 221: This is the method that is used whenever an action in the MPT should record an MPT node in the witness.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+    witness: Witness,
+    node: MutableNode,
+) -> None:
+    """Record a node access in the witness."""
+    if node is None:
```
jsign, line 226: Empty `BranchNode` children should be skipped. i.e. `0x80` RLP for the stateless client is enough signal that is an empty reference.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+        return
+
+    if node._dirty:
```
jsign, line 229: If we got a signal to track any MPT node that was created or updated, we skip it.
Doing this filtering here just simplify callers never having to identify these situations.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+        return
+
+    node_hash, node_rlp = _compute_node_hash_and_rlp(node)
+    if node_hash is not None and node_hash not in witness.accessed_nodes:
```
jsign, line 233: Reg `node_hash is not None`, worth mentioning as a refresher: in MPT whenever an RLP-encoded node is <32 bytes, then the node is embedded in its parent.

For these node scenarios, we don't need to include them in the witness since they technically live embedded in parents.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+    node._rlp = encoded
+
+    if len(encoded) >= 32:
+        node._hash = keccak256(encoded)
+        return node._hash, encoded
+    else:
+        return None, encoded
+
+
+def mpt_get(mpt: IncrementalMPT[K, V], key: K) -> V:
```
jsign, line 315: Whenever the state-tracker is in place, we will do `mpt_get` on all detected state. This method will do `_mpt_traverse_for_witness` which basically walks the branch and registers the MPT nodes to be included in the witness.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+        nibble_key = bytes_to_nibble_list(keccak256(key))
+    else:
+        nibble_key = bytes_to_nibble_list(key)
+
+    _mpt_traverse_for_witness(mpt, mpt.root_node, nibble_key, Uint(0))
+
+    return value
+
+
+def _mpt_traverse_for_witness(
```
jsign, line 347: Prefered a recursive style since it is very compact, and usually depths in tests are very shallow.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+@dataclass
+class IncrementalMPT(Generic[K, V]):
+    """
+    An MPT that supports incremental updates and witness tracking.
+
+    Maintain an actual tree structure that can be updated in-place,
+    rather than rebuilding the entire tree on each root calculation.
+    """
+
+    secured: bool
```
jsign, line 102: In theory we could remove `secured` for `ExecutionWitness`, since we only use this trie for the account and storage tries which are always secured.

But if we eventually compeletely remove the `patricialize`, we can use this trie for txs/withdrawals/etc root calcs. So, I think it makes sense to keep it.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,803 @@
...
+    mpt: IncrementalMPT,
+    node: MutableNode,
+    key: Bytes,
+    level: Uint,
+) -> None:
+    """Traverse the tree recording nodes in the witness."""
+    if node is None:
+        return
+
+    _record_witness(mpt.witness, node)
```
jsign, line 357: Track every node during the walk.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,802 @@
...
+
+def _collapse_branch(
+    mpt: IncrementalMPT, node: MutableBranchNode
+) -> MutableNode:
+    """Collapse a branch node if it has only one child."""
+    non_empty = [(i, c) for i, c in enumerate(node.children) if c is not None]
+
+    assert len(non_empty) > 0 or node.value != b""
+
+    if len(non_empty) == 1 and node.value == b"":
```
jsign, line 759: If the `BranchNode` is left with one child after the deletion, we must nuke it and compress the branch. In L745 we signal to record this remainig sibiling since it is needed to know how to compress the branch.

This is basically the only "obscure" need reg MPT node detection for `ExecutionWitness`.

Recall that `_record_witness` will automatically discard if this sibiling was created during execution.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,802 @@
...
+                rest_of_key=nibble + child.rest_of_key,
+                value=child.value,
+                _dirty=True,
+            )
+        elif isinstance(child, MutableExtensionNode):
+            return MutableExtensionNode(
+                key_segment=nibble + child.key_segment,
+                child=child.child,
+                _dirty=True,
+            )
```
jsign, line 775: Note that the caller of `_collapse_branch` will eventually return to a potential `_delete_from_extension`, that will merge this `MutableExtensionNode` to a potential parent that is `MutableExtensionNode`.

To explain more clearly: If we had `ExtensionNode -> BranchNode --> [A, ExtensionNode]`, and we delete `A`, then we would have `ExtensionNode -> ExtensionNode` -- but both are merged by union-ing their `key_segment` in the caller.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -0,0 +1,802 @@
...
+        else:
+            return MutableExtensionNode(
+                key_segment=nibble,
+                child=child,
+                _dirty=True,
+            )
+
+    if len(non_empty) == 0 and node.value != b"":
```
jsign, line 783: Not something technically possible in Ethereum, but MPT allows having values in BranchNodes...

src/ethereum/forks/amsterdam/trie.py
```diff
@@ -354,21 +354,26 @@ def bytes_to_nibble_list(bytes_: Bytes) -> Bytes:
     return Bytes(nibble_list)
```
jsign, line 313: This is just to allow `build_mpt` in the previous file reuse the code. Nothing fancy is going on here.

kevaundray:
Fixing the CI issues in #2269

kevaundray (approved):

gurukamath (approved):

PR #2243, opened 2026-02-19
feat(test-execute): Batch RPC requests

marioevz:
## 🗒️ Description

Allow using [JSON-RPC batch requests](https://json-rpc.dev/learn/examples/batch-requests) in `execute` so all transactions that are contained within a single block in a test, are always requested in the same batch at the same time.

We also now batch nonce, balance, code and storage requests, which are meant to verify the test at the end of execution, all in a single batch request always.

Also a nice low-hanging fruit fix is that `execute hive` now sets the deterministic contract factory directly in the pre-allocation, so it does not need to be deployed before the tests start.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
@@ -68,8 +68,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         type=int,
         default=120,
         help=(
-            "Timeout in seconds for workers waiting to be funded. "
-            "Default=120"
+            "Timeout in seconds for workers waiting to be funded. Default=120"
         ),
     )
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
@@ -61,6 +61,18 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         ),
     )
 
+    sender_group.addoption(
+        "--worker-funding-timeout",
+        action="store",
+        dest="worker_funding_timeout",
+        type=int,
+        default=120,
+        help=(
+            "Timeout in seconds for workers waiting to be funded. "
+            "Default=120"
+        ),
+    )
+
 
 @pytest.fixture(scope="session")
 def sender_funding_transactions_gas_price(
@@ -310,8 +322,18 @@ def session_worker_key(
             worker_funded_file.touch()
 
     if not is_last_worker:
-        logger.info("Waiting for all workers to be funded...")
+        funding_timeout = request.config.option.worker_funding_timeout
+        logger.info(
+            "Waiting for all workers to be funded "
+            f"(timeout={funding_timeout}s)..."
+        )
+        deadline = time.monotonic() + funding_timeout
         while not worker_funded_file.exists():
+            if time.monotonic() > deadline:
+                raise TimeoutError(
+                    f"Timed out after {funding_timeout}s waiting "
+                    "for all workers to be funded"
+                )
             time.sleep(0.1)
         logger.info("All workers funded, proceeding")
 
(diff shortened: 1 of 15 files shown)


kamilchodola:
Flawless improvement - on stateful scenarios speeding up test generation for 3 huge test cases from 13 minutes to 30 seconds.

On stateful generator side only nethermind config parameter needed to be adjusted to allow more batches in single request (default 1024)

kamilchodola (approved):

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
```diff
@@ -374,19 +390,18 @@ def worker_key(
 ) -> Generator[EOA, None, None]:
     """Prepare the worker key for the current test."""
     logger.debug(f"Preparing worker key {session_worker_key} for test")
-    rpc_nonce = Number(
-        eth_rpc.get_transaction_count(
-            session_worker_key, block_number="pending"
```
fselmo, line 379: nit: Curious if preserving this `pending` matters. I don't really think so since this runs before the test, but flagging in case it makes sense to keep this the same behavior. I believe the new code uses `latest` now.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
```diff
@@ -256,55 +262,58 @@ def session_worker_key(
...
+            if not dry_run:
+                eth_rpc.send_wait_transactions(fund_txs)
+                logger.info("All worker funding transactions confirmed")
+            else:
+                logger.info("Dry run: skipping funding transaction send")
+            worker_funded_file.touch()
+
+    if not is_last_worker:
+        logger.info("Waiting for all workers to be funded...")
+        while not worker_funded_file.exists():
```
fselmo, line 330: What happens in the case that a worker crashes? Do we ever hit the `if len(registered_keys) == worker_count` check above that sets the `is_last_worker`? And if not then do we just brick here?

I think we could use a reasonable timeout here just in case this scenario gets triggered?

fselmo (commented):
Still going through all the moving parts here but some initial comments on a quick first pass.

packages/testing/src/execution_testing/rpc/rpc.py
```diff
@@ -638,10 +753,32 @@ def send_transactions(
...
+                method="sendRawTransaction",
+                params=[tx.rlp().hex()],
+                request_id=tx.metadata_string(),
+            )
+            for tx in transactions
+        ]
+        responses = self.post_batch_request(calls=calls)
+
+        results: List[Hash] = []
+        for tx, response in zip(transactions, responses, strict=False):
```
spencer-tb: With `strict=False`, if the server returns less responses than transactions sent, the trailing transactions are silently dropped without error.
```suggestion
        for tx, response in zip(transactions, responses, strict=True):
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
marioevz: I added the pending back because otherwise we could get a `ReplacementNotAllowed` error at some point. I tested both and locally it did not make much of a difference but I agree we should not change what's not broken.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
marioevz: Yes, makes sense, added new `--worker-funding-timeout` parameter with a default of 120 seconds.

marioevz:
@fselmo @spencer-tb applied all comments, ready for another pass, thanks!

packages/testing/src/execution_testing/rpc/rpc.py
```diff
@@ -638,10 +753,32 @@ def send_transactions(
...
+            try:
+                result_hash = Hash(response.result_or_raise())
+                assert result_hash == tx.hash
+                assert result_hash is not None
+                results.append(tx.hash)
+            except Exception as e:
+                raise SendTransactionExceptionError(str(e), tx=tx) from e
+        return results
 
     def storage_at_keys(
```
fselmo: Flagged by PyCharm that this isn't used anywhere. I can't find any usage, is this dead code? 👀

fselmo (approved):
This lgtm, just one unrelated comment from a PyCharm warning. Will let you merge in case you want to look into it here 👍🏼

packages/testing/src/execution_testing/rpc/rpc.py
marioevz: Indeed, I'll remove it before merge, thanks! :D

Issue #2090, opened 2026-01-28
Speed-up JSON infra by filling a subset of test and executing it


marioevz:
## Description

Reduce the time JSON Infra CI takes and also improve its accuracy, in particular for `eips/` branches:

- Mark a small subset of tests (e.g., “`pytest.mark.json_infra`”) to cover each transaction type, format, and fork
- In the JSON infra job, first run `fill ... -m "json_infra" ...` to fill only those marked tests
- Immediately feed that freshly generated output into the JSON infra runner

#### Pre-requisites

#1537

PR #2251, opened 2026-02-20
chore(ci): `json_infra` Put each fork on a separate github runner

kevaundray:
## 🗒️ Description

Was initially curious to see what the workload distribution was across forks. If I haven't made a mistake then this PR:

- Removes the need for xl runners (switches back to the github free runners)
- Reduces CI time for json_infra from ~ 3 hours to ~20 minutes

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -110,30 +110,30 @@ jobs:
       fail-fast: false
       matrix:
         fork:
-          - Frontier
-          - Homestead
-          - "Dao Fork"
-          - EIP150
-          - EIP158
-          - Byzantium
-          - ConstantinopleFix
-          - Istanbul
-          - "Muir Glacier"
-          - Berlin
-          - London
-          - "Arrow Glacier"
-          - "Gray Glacier"
-          - Paris
-          - Shanghai
-          - Cancun
-          - Prague
-          - Osaka
-          - BPO1
-          - BPO2
-          - BPO3
-          - BPO4
-          - BPO5
-          - Amsterdam
+          - frontier
+          - homestead
+          - dao_fork
+          - tangerine_whistle
+          - spurious_dragon
+          - byzantium
+          - constantinople
+          - istanbul
+          - muir_glacier
+          - berlin
+          - london
+          - arrow_glacier
+          - gray_glacier
+          - paris
+          - shanghai
+          - cancun
+          - prague
+          - osaka
+          - bpo1
+          - bpo2
+          - bpo3
+          - bpo4
+          - bpo5
+          - amsterdam
     steps:
       - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955
         with:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -107,7 +107,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: static
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         fork:
           - frontier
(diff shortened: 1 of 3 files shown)


tests/json_infra/conftest.py
```diff
@@ -146,10 +146,25 @@ def pytest_configure(config: Config) -> None:
 
     desired_forks = []
     all_forks = list(FORKS.keys())
+    # Build a lookup from short_name to json_test_name so that
+    # --fork accepts either format (e.g. "gray_glacier" or
+    # "Gray Glacier").
+    short_to_json = {f.short_name: name for name, f in FORKS.items()}
```
kevaundray, line 152: As noted, it seems that there is no way to have tox accept parameters with spaces, so --fork= Gray Glacier wont work.

We could modify the JSON files to have `gray_glacier`, ie not have spaces then we can remove this

tests/json_infra/conftest.py
```diff
@@ -232,6 +247,9 @@ def __init__(self, root: Path) -> None:
 
     def fetch_http(self, url: str, location: str) -> None:
         path = self.root.joinpath(location)
+        if path.exists() and any(path.iterdir()):
+            print(f"{location} already available, skipping download.")
```
kevaundray: So we aren't redownloading `latest_fork_tests` and `amsterdam_tests` unconditionally

marioevz (commented):
A couple of things:
- The strategy matrix ends up running the fork even if there were no changes for that specific fork in the PR, we could improve that by adding a check there, and finish before reaching the `tox` command.
- We could remove the latest fork also, in this case `amsterdam` because these tests always fail because we always use an outdated set of fixtures for the development fork from the previous fill, so any change breaks this.

kevaundray:
> A couple of things:
> 
> * The strategy matrix ends up running the fork even if there were no changes for that specific fork in the PR, we could improve that by adding a check there, and finish before reaching the `tox` command.

will make this change -- I wasn't sure of a clean way to do this, but having a separate composite action might be best here. 

We would be able to remove `--file-list` once the `optimized` job is removed, since thats the only place it gets used after this change.

> * We could remove the latest fork also, in this case `amsterdam` because these tests always fail because we always use an outdated set of fixtures for the development fork from the previous fill, so any change breaks this.

I'm not familiar with the workflow for latest fork, but can also remove 🙏   

Does the general change make sense re the CI workflow?

.github/workflows/test.yaml
```diff
@@ -104,22 +104,66 @@ jobs:
...
+          - bpo3
+          - bpo4
+          - bpo5
+          - amsterdam
     steps:
       - uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955
         with:
-          submodules: recursive
-          fetch-depth: 0  # Fetch full history for commit comparison
+          fetch-depth: 0
```
kevaundray:
```suggestion
          fetch-depth: 0 # Fetch full history for commit comparison
```

kevaundray:
@danceratopz did you have an opinion on this change by chance?

gurukamath (commented):
There are two things we need to consider before we make the decision to go ahead with this strategy

1. In the scope of #2090, we have a plan to run a minimal number of tests under `json_infra` and that should significantly reduce the run time
2. One of the nice side-effects of the recent state tracker effort, is that it should lead to a much simpler `t8n` tool structure. This could lead to a lot of the json consumption code becoming redundant. We can delegate a lot of the json infra stuff to the EEST pydantic models  in that scenario

I think we should explore the above possibilities first, and fall back to this if they turn out to be infeasible.

kevaundray:
Closing -- there are issues for those two items and the worse case(3 hours) happens only if we modify the testing directory or all forks. Most people will only modify the latest fork, so they aren't hit with the 3 hour penalty.

PR #2253, opened 2026-02-20
feat(ci): auto-merge approved PRs via required workflow gates

spencer-tb:
### Description

Enables GitHub native auto-merge on approval. Replaces the merge-gate/polling approach with Sam's suggestion from the review discussion: job-level path filtering, so every PR workflow always reports a check run and can be a required status check.

- Each PR workflow gets a `changes` job (`dorny/paths-filter`) carrying the old `on.pull_request` path filters; root jobs skip when no relevant files changed. Skipped jobs satisfy required checks — skipped *workflows* never report, which is why workflow-level PR filters had to go. Push-side filters are unchanged.
- Each PR workflow ends in an always-run gate job that fails if any job failed: `specs-gate`, `benchmark-gate`, `docs-gate`, `render-gate`, `hive-consume-gate`, `hive-execute-gate`, `checklist-gate`. These are the required checks; GitHub natively waits for them, so the old race (gate evaluating before slow workflows register) is gone.
- `yolo-merge.yaml` arms auto-merge (squash, PR title as subject) when a non-draft PR targeting `forks/`, `devnets/`, or `eips/` is approved; other bases are never armed.
- PRs touching only ignored files (e.g. `whitelist.txt`) now auto-merge too: workflows run, jobs skip, gates pass.

Repo settings required (admin):

1. Enable "Allow auto-merge" (currently off).
2. Add the 7 gate jobs above as required status checks on a ruleset targeting `forks/**`, `devnets/**`, and `eips/**` (must cover every base yolo-merge arms; devnet/eips branches currently have no pull_request rules at all).
3. Enable "Dismiss stale pull request approvals when new commits are pushed" (per discussion with @Carsons-Eels).
4. Optional but recommended: add a `YOLO_MERGE_TOKEN` secret (PAT/App token with contents+PR write). Merges armed with `github.token` are attributed to github-actions[bot], whose push does not trigger the push workflows (docs deploy, fixture builds).

### Related Issues or PRs

N/A.

### Checklist

- [x] PR title has the form `<type>(<area>):`, where `<type>` and `<area>` come from an appropriate `C-<type>`, respectively `A-<area>`, label. The title should match the target squash commit message.

### Cute Animal Picture

yolo

### DIFF
--- a/.github/workflows/yolo-merge.yaml
+++ b/.github/workflows/yolo-merge.yaml
@@ -1,0 +1,37 @@
+name: Yolo Merge
+
+on:
+  pull_request_review:
+    types: [submitted]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  yolo-merge:
+    name: yolo-merge
+    runs-on: ubuntu-latest
+    # Base prefixes must match the branches whose ruleset requires the
+    # gate checks; arming auto-merge on an unprotected base merges
+    # without CI.
+    if: >-
+      github.event.review.state == 'approved' &&
+      !github.event.pull_request.draft &&
+      (startsWith(github.event.pull_request.base.ref, 'forks/') ||
+       startsWith(github.event.pull_request.base.ref, 'devnets/') ||
+       startsWith(github.event.pull_request.base.ref, 'eips/'))
+    steps:
+      - name: Enable auto-merge
+        env:
+          # Prefer a PAT/App token: merges performed with github.token do
+          # not trigger push workflows (docs deploy, fixture builds) on the
+          # target branch.
+          GH_TOKEN: ${{ secrets.YOLO_MERGE_TOKEN || github.token }}
+          GH_REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          gh pr merge "$PR_NUMBER" \
+            --squash --auto \
+            --subject "$PR_TITLE (#$PR_NUMBER)"
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -11,14 +11,6 @@ on:
       - "packages/testing/src/execution_testing/cli/pytest_commands/plugins/**"
       - ".github/workflows/benchmark.yaml"
   pull_request:
-    paths-ignore:
-      - "**.md"
-      - "LICENSE*"
-      - ".gitignore"
-      - ".vscode/**"
-      - "whitelist.txt"
-      - "docs/**"                                                                                                                     
-      - "mkdocs.yml"
   workflow_dispatch:
 
 concurrency:
@@ -26,9 +18,38 @@ concurrency:
   cancel-in-progress: ${{ github.ref_name != github.event.repository.default_branch }}
 
 jobs:
+  # Job-level path filter: the workflow must always run on PRs so its
+  # gate can be a required status check (skipped workflows never report).
+  changes:
+    name: Detect Changes
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      relevant: ${{ steps.filter.outputs.relevant || 'true' }}
+    steps:
+      - name: Filter relevant paths
+        id: filter
+        if: github.event_name == 'pull_request'
+        uses: dorny/paths-filter@7b450fff21473bca461d4b92ce414b9d0420d706 # v4.0.2
+        with:
+          predicate-quantifier: "every"
+          filters: |
+            relevant:
+              - "**"
+              - "!**/*.md"
+              - "!LICENSE*"
+              - "!.gitignore"
+              - "!.vscode/**"
+              - "!whitelist.txt"
+              - "!docs/**"
+              - "!mkdocs.yml"
+
   unit-tests:
     name: Benchmark Unit Tests
     runs-on: ubuntu-latest
+    needs: changes
+    if: needs.changes.outputs.relevant == 'true'
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
@@ -103,3 +124,20 @@ jobs:
       - uses: ./.github/actions/build-fixtures
         with:
           release_name: benchmark_fast
+
+  gate:
+    name: benchmark-gate
+    runs-on: ubuntu-latest
+    needs:
+      - changes
+      - unit-tests
+      - sanity-checks
+      - build-artifact
+    if: always()
+    steps:
+      - name: Check that all jobs passed or were skipped
+        env:
+          NEEDS: ${{ toJson(needs) }}
+        run: |
+          echo "$NEEDS" | jq -e \
+            '[.[].result] | all(. == "success" or . == "skipped")'
--- a/.github/workflows/docs-build.yaml
+++ b/.github/workflows/docs-build.yaml
@@ -25,7 +25,7 @@ on:
     branches:
       - mainnet
       - forks/amsterdam
-    paths: &docs_paths
+    paths:
       - "src/ethereum/**"
       - "src/ethereum_spec_tools/docc.py"
       - "src/ethereum_spec_tools/forks.py"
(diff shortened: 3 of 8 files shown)


.github/workflows/yolo-merge.yaml
```diff
@@ -0,0 +1,143 @@
...
+      (github.event_name == 'issue_comment' && github.event.issue.pull_request)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Detect magic word and enable auto-merge
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
+        env:
+          GH_TOKEN: ${{ github.token }}
+        with:
+          script: |
+            const MAGIC_WORDS = ['shipit', 'bless', 'lgtm', 'yolo'];
```
SamWilsn: Devil's advocate: what scenario exists where we wouldn't want to merge after approving? Could draft PRs solve this?

SamWilsn (commented):
If I'm reading this correctly, the magic word does not need to be in the approving review? That seems unexpected.

---

What would happen if we just blanket enabled auto-merge unconditionally? My understanding is that it would wait to merge until all of the pull request conditions are met. Have you considered this approach?

.github/workflows/yolo-merge.yaml
spencer-tb: Yeah, I figured magic words would be a nice way to decouple before we went all in. I prefer removing magic words all together now you mention it. But I will miss getting the change yolo!

spencer-tb: I think draft is the only option!

SamWilsn:
> What would happen if we just blanket enabled auto-merge unconditionally? My understanding is that it would wait to merge until all of the pull request conditions are met. Have you considered this approach?

What about this?

spencer-tb:
> > What would happen if we just blanket enabled auto-merge unconditionally? My understanding is that it would wait to merge until all of the pull request conditions are met. Have you considered this approach?
> 
> What about this?

Nice! Thanks much simpler :)

SamWilsn:
What checks do you want to make required? I'll add them.

spencer-tb:
> What checks do you want to make required? I'll add them.

Ideally all of them, but because `test.yaml` uses paths-ignore, skipped workflows don't satisfy the required check. What about a pass through job that aggregates at the end? Then we just require that.

cc: https://github.com/orgs/community/discussions/12395

SamWilsn:
Works for me!

spencer-tb:
I added a merge gate workflow that polls `gh pr checks` until all other checks pass or are skipped. This was the only route I could come to because our workflows use `paths-ignore`/`paths` filters. Github doesn't support "required if run" for status checks ([community discussion](https://github.com/orgs/community/discussions/12395)), so skipped required checks block merges. The alternative would be putting all jobs in a single workflow file, but that loses the per-workflow path filtering and concurrency controls.

Merge queues are also off the table for now, `merge_group` doesn't support path filtering ([discussion](https://github.com/orgs/community/discussions/26092)) :(

spencer-tb:
All possible PR flows (I think):
1. Open, push commits, approve, CI passes, merge gate passes, auto-merges.
2. Open, approve early, CI still running, merge gate polls, CI passes, auto-merges.
3. Open, approve, CI fails, merge gate fails, author fixes and pushes, merge gate re-triggers, CI passes, auto-merges. No re-approval needed.
4. Open, approve, push more commits, merge gate re-triggers, CI re-runs, auto-merges.
5. Open, approve, merge conflicts, author rebases and pushes, merge gate re-triggers, auto-merges.
6. Open, CI fails, fix and push, approve later, merge gate already polling, auto-merges.
7. Multiple approvals, yolo merge is a no-op after first, merge gate already running.
8. Close and reopen, merge gate re-triggers on reopen, approve, auto-merges.

spencer-tb:
I'm not sure if polling is wasteful but feels like the best solution, that works with the cases above!

SamWilsn:
I haven't really dug into the merge gate stuff yet (thanks a ton for working on it, it's a great idea), but would something like [`workflow_run`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run) be useful here instead of polling?

Carsons-Eels:
Re: Sam's `workflow_run` suggestion — I looked into this and unfortunately `workflow_run` doesn't support an "all workflows completed" trigger. When you list multiple workflows:

```yaml
on:
  workflow_run:
    workflows: [Test, Benchmark, Docs]
    types: [completed]
```

It fires when **any one** of them completes, not when all of them complete.

So to use `workflow_run` as a merge gate, you'd have to pick one workflow from amongst the 6 PR-triggered workflows, likely the benchmark or test workflow. That's a bit fragile though since there are some scenarios where the finishing order may not be as we expect.

Also, if a test fails due to a transient issue (network flakiness, runner problems) and then gets manually rerun, that workflow passes on retry but it would not trigger the merge gate. This one is a pretty minor concern though, I think, since the dev is already directly attending to it and can just manually trigger the merge after they fix it.

The polling approach is probably more straightforward, but it feels intensive and wasteful to be polling that much on a repo this busy, especially during a crunch week. Maybe drop polling interval to something like 5 min to cut down on runner usage?

Also, what about the scenario: PR -> CR -> Corrections -> CI failure -> Approval -> Fix CI failure (new code) -> Merge. This could allow code that has not been reviewed into the repo right? Actually, I think you can already do this as-is, but this makes it easy to accidentally do since the merge would be automatic. Not sure how much of a concern that is, but it does allow you to sneak around the rules heh.

SamWilsn:
> Re: Sam's `workflow_run` suggestion — I looked into this and unfortunately `workflow_run` doesn't support an "all workflows completed" trigger.

That's fine, isn't it? Whenever any workflow completes, you check if all workflows are completed.

SamWilsn:
> Also, what about the scenario: PR -> CR -> Corrections -> CI failure -> Approval -> Fix CI failure (new code) -> Merge. This could allow code that has not been reviewed into the repo right? Actually, I think you can already do this as-is, but this makes it easy to accidentally do since the merge would be automatic. Not sure how much of a concern that is, but it does allow you to sneak around the rules heh.

We can enable the GitHub option that clears approvals on code changes.

spencer-tb:
Adding this just now. Thanks for being patient with me here!

spencer-tb:
Okay added workflow run here! One edge case:

If a PR only touches files excluded by every workflow's path filter (e.g., whitelist.txt, .gitignore), no `workflow_run` fires and the "Merge Gate" status is never created. So for these rare case, we'd have to merge manually. Unless we have CI that we want to add for them.

spencer-tb:
Repo settings to check and add as far as I can tell @SamWilsn:

1. Require status checks to pass before merging, add "Merge Gate" (the commit status context) as the required check
2. Dismiss stale pull request approvals when new commits are pushed
3. Enable auto-merge, required for gh pr merge --auto to work
4. Set squash as the default/only merge method, matches the --squash flag in yolo-merge

SamWilsn:
I was going through this, and stumbled across this section of the [GitHub docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/about-protected-branches?utm_source=chatgpt.com#require-status-checks-before-merging):

image

Do we still need the merge gate job at all, or am I way off base?

SamWilsn:
Ahhh, it breaks when the whole workflow is skipped. Gotcha.

.github/workflows/merge-gate.yaml
```diff
@@ -0,0 +1,96 @@
...
+          # On initial PR event, set pending and wait for other checks
+          # to register before evaluating
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            gh api "repos/${REPO}/statuses/${SHA}" \
+              -f state=pending \
+              -f context="Merge Gate" \
+              -f description="Waiting for checks"
+            sleep 60
+          fi
+
```
SamWilsn, line 74: I'm not entirely clear on how this works 😅 

It looks like it waits 60 seconds after the PR is opened then looks to see if any checks are failed/pending? It also rechecks when every workflow finishes?

If that's correct, I see a race condition. If the fastest check returns before the slowest check gets marked as pending, this'll immediately merge?

SamWilsn:
Would it perhaps be cleaner if we dropped the filtering at the workflow level, and instead use something like https://github.com/dorny/paths-filter ?

spencer-tb:
Closing for now.

PR #2255, opened 2026-02-20
refactor(test-benchmarks): ensure `SSTORE` N->N writes nonzero key/value

jochem-brouwer:
## 🗒️ Description
Tweaks benchmarks to fix expectations in repricings. CC @misilva73
Draft, will push more tweaks if more are found after analysis

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/compute/instruction/test_storage.py
+++ b/tests/benchmark/compute/instruction/test_storage.py
@@ -501,8 +501,8 @@ def test_storage_access_warm_benchmark(
     """
     match storage_action:
         case StorageAction.WRITE_SAME_VALUE:
-            # Timestamp is nonzero (no txs run in Genesis block), so this always writes
-            # to the zero key a nonzero, constant value
+            # Timestamp is nonzero (no txs run in Genesis block),
+            # Always writes to the zero key a nonzero, constant value
             attack_block = Op.SSTORE(Op.PUSH0, Op.TIMESTAMP)
         case StorageAction.WRITE_NEW_VALUE:
             attack_block = Op.SSTORE(Op.PUSH0, Op.GAS)
--- a/tests/benchmark/compute/instruction/test_storage.py
+++ b/tests/benchmark/compute/instruction/test_storage.py
@@ -501,7 +501,9 @@ def test_storage_access_warm_benchmark(
     """
     match storage_action:
         case StorageAction.WRITE_SAME_VALUE:
-            attack_block = Op.SSTORE(Op.PUSH0, Op.PUSH0)
+            # Timestamp is nonzero (no txs run in Genesis block), so this always writes
+            # to the zero key a nonzero, constant value
+            attack_block = Op.SSTORE(Op.PUSH0, Op.TIMESTAMP)
         case StorageAction.WRITE_NEW_VALUE:
             attack_block = Op.SSTORE(Op.PUSH0, Op.GAS)
         case StorageAction.READ:


tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -503,7 +503,9 @@ def test_storage_access_warm_benchmark(
         case StorageAction.WRITE_SAME_VALUE:
             attack_block = Op.SSTORE(Op.PUSH0, Op.PUSH0)
```
misilva73: How about this one? The `Op.PUSH0` here is also a non-zero value, right?

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -503,7 +503,9 @@ def test_storage_access_warm_benchmark(
...
         case StorageAction.WRITE_NEW_VALUE:
-            attack_block = Op.SSTORE(Op.PUSH0, Op.GAS)
```
misilva73, line 506: I think this one was fine, no? It was writing a different value (Op.GAS) to the same slot (slot 0)

tests/benchmark/compute/instruction/test_storage.py
jochem-brouwer: Yes you are right, not sure why I changed the wrong action here 😅

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -501,7 +501,9 @@ def test_storage_access_warm_benchmark(
     """
     match storage_action:
         case StorageAction.WRITE_SAME_VALUE:
-            attack_block = Op.SSTORE(Op.PUSH0, Op.PUSH0)
+            # Timestamp is nonzero (no txs run in Genesis block), so this always writes
+            # to the zero key a nonzero, constant value
+            attack_block = Op.SSTORE(Op.PUSH0, Op.TIMESTAMP)
```
misilva73, line 506: Another dumb question: does  Op.TIMESTAMP change during the loop? Or is it fixed? I am thinking that if a block takes seconds to run, the Op.TIMESTAMP would change

jochem-brouwer: Not a dumb question! This is constant during the block! So even if the block takes multiple seconds to run the timestamp is equal. It is choose because it takes less gas (2) vs a push operation (3). For warm sstores to same value the sstore costs 100, so 2 vs 3 actually has some impact

misilva73:
To properly document this, `test_storage_access_cold_benchmark` also writes 0->0 in the `WRITE_SAME_VALUE` variant. However, in this case, since we are always writting to new accounts, I don't think we can easily fix it. Here is the code:

```python
def test_storage_access_cold_benchmark(
    storage_action: StorageAction,
) -> None:
    """
    Benchmark cold storage slot accesses using code generator.

    Each iteration accesses a different storage slot (incrementing key)
    to ensure cold access costs are measured.
    """
    if storage_action == StorageAction.READ:
        attack_block = Op.SLOAD(Op.GAS)
    elif storage_action == StorageAction.WRITE_SAME_VALUE:
        attack_block = Op.SSTORE(Op.GAS, Op.PUSH0)
    elif storage_action == StorageAction.WRITE_NEW_VALUE:
        attack_block = Op.SSTORE(Op.GAS, Op.GAS)
```

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: Good point!

LouisTsai-Csie (approved):
Hi @misilva73 @jochem-brouwer , what's the issue for these benchmark test, is there anything i could help? what benchmark combination do you need

The current changes looks good to me, i approved and wonder if you'd like to merge first

jochem-brouwer:
This one can be merged!

PR #2240, opened 2026-02-19
refactor(test-fill): remove plugin-based witness generation

jsign:
## 🗒️ Description
Doing some cleanup of the current plugin-based witness generation code. 
I think we can remove this to clean up some stuff (nothing like deleting code PRs), since we are actively working on spec-generated witnesses for zkVMs, so we won't be using this.

Also can avoid confusions in the branches on top of `forks/amsterdam` that we'll be working on.

## 🔗 Related Issues or PRs
https://github.com/ethereum/execution-spec-tests/pull/2066

PR #2259, opened 2026-02-20
feat(zkevm): wire ExecutionWitness into the testing framework

jsign:
## 🗒️ Description
This PR adds the required wiring to include the `executionWitness` as part of filled test fixtures.

Example (you can run it):
```
$ uv run fill --clean -m "blockchain_test" --fork Amsterdam ./tests/tangerine_whistle -k test_selfdestruct_to_account
...
$ cat fixtures/blockchain_tests/tangerine_whistle/eip150_operation_gas_costs/eip150_selfdestruct/selfdestruct_to_account.json
...
 "receipts": [...],
 "executionWitness": {
     "state": [],
     "codes": [],
     "headers": []
  },
  "blockAccessList": [...],
...
```

The change is quite similar to how it is usually done in general (e.g. BALs):
- We add `ExecutionWitness` as part of the block output of the STF.
- The Amsterdam fork has a property to signal it supports execution witness generation.
- t8n now asks the fork if it has the property, and properly pulls the block output result to be returned.
- The testing framework detects this new potential output of t8n and adds it to the fixture output.

The goal of this PR is to already do all the wiring with the testing framework, so future PRs can focus only on generating the witness during execution, and we can already see the results in the filled fixtures.

I'll add some extra comments to help review.

### DIFF
--- a/packages/testing/src/execution_testing/client_clis/cli_types.py
+++ b/packages/testing/src/execution_testing/client_clis/cli_types.py
@@ -25,10 +25,10 @@ from execution_testing.exceptions import (
     TransactionException,
     UndefinedException,
 )
+from execution_testing.fixtures.blockchain import ExecutionWitnessTemp
 from execution_testing.logging import (
     get_logger,
 )
-from execution_testing.fixtures.blockchain import ExecutionWitnessTemp
 from execution_testing.test_types import (
     Alloc,
     BlockAccessList,
--- a/packages/testing/src/execution_testing/client_clis/cli_types.py
+++ b/packages/testing/src/execution_testing/client_clis/cli_types.py
@@ -25,7 +25,6 @@ from execution_testing.exceptions import (
     TransactionException,
     UndefinedException,
 )
-from execution_testing.fixtures.blockchain import ExecutionWitnessTemp
 from execution_testing.logging import (
     get_logger,
 )
@@ -33,6 +32,7 @@ from execution_testing.test_types import (
     Alloc,
     BlockAccessList,
     Environment,
+    ExecutionWitness,
     Transaction,
     TransactionReceipt,
 )
@@ -290,7 +290,7 @@ class Result(CamelModel):
     requests: List[Bytes] | None = None
     block_access_list: BlockAccessList | None = None
     block_access_list_hash: Hash | None = None
-    execution_witness_temp: ExecutionWitnessTemp | None = None
+    execution_witness: ExecutionWitness | None = None
     block_exception: Annotated[
         BlockExceptionWithMessage | UndefinedException | None,
         ExceptionMapperValidator,
--- a/packages/testing/src/execution_testing/client_clis/cli_types.py
+++ b/packages/testing/src/execution_testing/client_clis/cli_types.py
@@ -28,6 +28,7 @@ from execution_testing.exceptions import (
 from execution_testing.logging import (
     get_logger,
 )
+from execution_testing.fixtures.blockchain import ExecutionWitnessTemp
 from execution_testing.test_types import (
     Alloc,
     BlockAccessList,
@@ -289,6 +290,7 @@ class Result(CamelModel):
     requests: List[Bytes] | None = None
     block_access_list: BlockAccessList | None = None
     block_access_list_hash: Hash | None = None
+    execution_witness_temp: ExecutionWitnessTemp | None = None
     block_exception: Annotated[
         BlockExceptionWithMessage | UndefinedException | None,
         ExceptionMapperValidator,
--- a/packages/testing/src/execution_testing/specs/blockchain.py
+++ b/packages/testing/src/execution_testing/specs/blockchain.py
@@ -61,7 +61,7 @@ from execution_testing.fixtures import (
     LabeledFixtureFormat,
 )
 from execution_testing.fixtures.blockchain import (
-    ExecutionWitnessTemp,
+    ExecutionWitness,
     FixtureBlock,
     FixtureBlockBase,
     FixtureConfig,
@@ -382,7 +382,7 @@ class BuiltBlock(CamelModel):
     engine_api_error_code: EngineAPIError | None = None
     fork: Fork
     block_access_list: BlockAccessList | None
-    execution_witness_temp: ExecutionWitnessTemp | None = None
+    execution_witness: ExecutionWitness | None = None
 
     def get_fixture_block(
         self, *, include_receipts: bool = True
@@ -412,8 +412,8 @@ class BuiltBlock(CamelModel):
             block_access_list=self.block_access_list
             if self.block_access_list
             else None,
-            execution_witness_temp=self.execution_witness_temp
-            if self.execution_witness_temp
+            execution_witness=self.execution_witness
+            if self.execution_witness
             else None,
             fork=self.fork,
         ).with_rlp(txs=self.txs)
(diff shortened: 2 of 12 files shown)


packages/testing/src/execution_testing/client_clis/cli_types.py
```diff
@@ -288,6 +289,7 @@ class Result(CamelModel):
     requests: List[Bytes] | None = None
     block_access_list: BlockAccessList | None = None
     block_access_list_hash: Hash | None = None
+    execution_witness_temp: ExecutionWitnessTemp | None = None
```
jsign: For t8n, I could already have used `execution_witness` instead of `execution_witness_temp` since https://github.com/ethereum/execution-specs/pull/2240 way of working was through a plugin outside t8n.

Honestly, I prefered to be coherent in this temporal naming to make the renaming easier later.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -781,6 +786,9 @@ def generate_block_data(
             engine_api_error_code=block.engine_api_error_code,
             fork=self.fork,
             block_access_list=bal,
+            execution_witness_temp=(
+                transition_tool_output.result.execution_witness_temp
+            ),
```
jsign: In the future, I think we might add capabilities in the testing framework to allow test creators have assertions about the generated execution witness. e.g. it should contain a bytecode with value XXX, etc.

When that happens, we would add some extra checks before as done with BALs. If those assertions aren't met, the filling would fail. 

For now since we don't have this capability, we simply accept whatever was filled.

src/ethereum/forks/amsterdam/vm/__init__.py
```diff
@@ -95,6 +99,7 @@ class BlockOutput:
     blob_gas_used: U64 = U64(0)
     requests: List[Bytes] = field(default_factory=list)
     block_access_list: BlockAccessList = field(default_factory=list)
+    execution_witness: Optional[ExecutionWitness] = None
```
jsign, line 103: New `BlockOutput` field as part of applying body.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -834,6 +836,10 @@ def apply_body(
         block_env.state_changes
     )
 
+    block_output.execution_witness = build_execution_witness(
```
jsign, line 838: Did `build_execution_witness` mirroring how BALs work, so can be used both for `state_transition` (i.e. here) and t8n (i.e. `src/ethereum_spec_tools/evm_tools/t8n/__init__.py`).

src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -45,31 +46,6 @@ class NewPayloadRequestHeader:
     execution_requests: ExecutionRequests
 
 
-@slotted_freezable
```
jsign, line 49: Moved `ExecutionWitness` from here to a new file `stateless_types.py`. This is mainly to avoid a circuilar dependency.

I think we can move more types in this file there, but for now doing only the minimal changes required.

src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -45,31 +46,6 @@ class NewPayloadRequestHeader:
...
-
-    codes: Tuple[Bytes, ...]
-    """
-    Contract-code preimages (created or accessed) needed during execution.
-    """
-
-    headers: Tuple[Bytes, ...]
-    """
-    RLP-encoded block headers used for pre-state and ``BLOCKHASH`` correctness
-    proofs. This may trend toward empty with EIP-2935 and EIP-7709.
```
jsign, line 70: BTW, EIP-2935 is already in mainnet 😅, I removed it when this code was moved.

src/ethereum/forks/amsterdam/stateless_types.py
```diff
@@ -0,0 +1,67 @@
...
+
+@dataclass
+class ExecutionWitnessBuilder:
+    """
+    Mutable accumulator for execution witness data during block execution.
+    """
+
+    state: List[Bytes] = field(default_factory=list)
+    codes: List[Bytes] = field(default_factory=list)
+    headers: List[Header] = field(default_factory=list)
```
jsign, line 48: This is the type that I added to the `BlockEnv` since `ExecutionWitness` is frozen, and also closer to the final serialization form.

src/ethereum/forks/amsterdam/stateless_types.py
```diff
@@ -0,0 +1,67 @@
...
+    ascending.
+    """
+    return ExecutionWitness(
+        state=tuple(sorted(builder.state)),
+        codes=tuple(sorted(builder.codes)),
+        headers=tuple(
+            rlp.encode(h)
+            for h in sorted(builder.headers, key=lambda h: h.number)
+        ),
+    )
```
jsign, line 67: Took advantage and put here the canonicalization rules for execution witness.

src/ethereum/forks/amsterdam/vm/__init__.py
```diff
@@ -50,6 +51,7 @@ class BlockEnvironment:
     excess_blob_gas: U64
     parent_beacon_block_root: Hash32
     state_changes: StateChanges
+    execution_witness: ExecutionWitnessBuilder
```
jsign, line 55: This is where we will collect mpt nodes, bytecodes, and ancestors during stf. 
With this PR wiring done, they will appear in the filled fixtures.

packages/testing/src/execution_testing/client_clis/cli_types.py
kevaundray: Is there a way to make it show up as executionWitness in the json instead of executionWitnessTemp? Or do we plan to change this to executionWitness before making a release?

src/ethereum/forks/amsterdam/stateless.py
kevaundray: Ah yeah I wrote this on the reth alloy docs too :D

packages/testing/src/execution_testing/client_clis/cli_types.py
jsign: PR description:
> Note: I named it executionWitnessTemp in all places to avoid upcoming conflicts when rebasing projects/zkevm into forks/amsterdam again after https://github.com/ethereum/execution-specs/pull/2240. This should make the rebase much easier, and then we can simply do a quick rename.

So I would say just before merging this PR hopefully. Or we can do it after quickly as soon as base branch is rebased.

kevaundray: Ah sorry, my eyes glossed over the italicised part -- makes sense

kevaundray (approved):

packages/testing/src/execution_testing/client_clis/cli_types.py
jsign: Rebased and renamed.

jsign:
The `static` CI failure is unrelated to this PR -- not getting there since there are many influx PRs. Happy to fix it here if is the next in the merging pipeline.

edit: @kevaundray, I think the lint failure comes from the EngineAPI PR, and to fix it asks to add a new field to `State` that not sure we want to do.

kevaundray:
> The `static` CI failure is unrelated to this PR -- not getting there since there are many influx PRs. Happy to fix it here if is the next in the merging pipeline.
> 
> edit: @kevaundray, I think the lint failure comes from the EngineAPI PR, and to fix it asks to add a new field to `State` that not sure we want to do.

Added a PR to fix here: https://github.com/ethereum/execution-specs/pull/2269 `created_accounts` was moved into the state_tracker, so we can remove it -- feel free to merge those changes into here if its blocking

gurukamath (approved):

PR #2264, opened 2026-02-21
chore(specs): Track account and storage reads automatically for BALs

kevaundray:
## 🗒️ Description

Taken from bulletpoint 2 in https://github.com/ethereum/execution-specs/pull/2207

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -71,7 +71,6 @@ from .state_tracker import (
     incorporate_tx_into_block,
     increment_nonce,
     set_account_balance,
-    track_address,
 )
 from .transactions import (
     AccessListTransaction,
@@ -916,8 +915,6 @@ def process_transaction(
     )
     tx_state = TransactionState(parent=block_env.state)
 
-    track_address(tx_state, block_env.coinbase)
-
     trie_set(
         block_output.transactions_trie,
         rlp.encode(index),
@@ -950,7 +947,6 @@ def process_transaction(
     gas = tx.gas - intrinsic_gas
 
     increment_nonce(tx_state, sender)
-    track_address(tx_state, sender)
 
     sender_balance_after_gas_fee = (
         Uint(sender_account.balance) - effective_gas_fee - blob_gas_fee
@@ -1080,7 +1076,6 @@ def process_withdrawals(
             rlp.encode(wd),
         )
 
-        track_address(wd_state, wd.address)
         current_balance = get_account(wd_state, wd.address).balance
         new_balance = current_balance + wd.amount * GWEI_TO_WEI
         set_account_balance(wd_state, wd.address, new_balance)
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -611,6 +611,54 @@ def make_receipt(
     return encode_receipt(tx, receipt)
 
 
+def process_checked_system_transaction(
+    block_env: vm.BlockEnvironment,
+    target_address: Address,
+    data: Bytes,
+) -> MessageCallOutput:
+    """
+    Process a system transaction and raise an error if the contract does not
+    contain code or if the transaction fails.
+
+    Parameters
+    ----------
+    block_env :
+        The block scoped environment.
+    target_address :
+        Address of the contract to call.
+    data :
+        Data to pass to the contract.
+
+    Returns
+    -------
+    system_tx_output : `MessageCallOutput`
+        Output of processing the system transaction.
+
+    """
+    pre_state_account = block_env.state.pre_state.get_account_optional(
+        target_address
+    )
+    if pre_state_account is None or len(pre_state_account.code) == 0:
+        raise InvalidBlock(
+            f"System contract address {target_address.hex()} does not "
+            "contain code"
+        )
+
+    system_tx_output = process_unchecked_system_transaction(
+        block_env,
+        target_address,
+        data,
+    )
+
+    if system_tx_output.error:
+        raise InvalidBlock(
+            f"System contract ({target_address.hex()}) call failed: "
+            f"{system_tx_output.error}"
+        )
+
+    return system_tx_output
+
+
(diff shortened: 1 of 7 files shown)


src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -259,6 +260,9 @@ def process_message(message: Message) -> Evm:
     # take snapshot of state before processing the message
     snapshot = copy_tx_state(tx_state)
 
+    # Entering an account is a state access for BAL tracking.
+    get_account_optional(tx_state, message.current_target)
+
```
kevaundray: calling this purely for the side effect, seems like a code smell -- the previous code was also doing this, checking the architecture

kevaundray: Okay the issue seemed to be that we were calling get_acount to get the code on a particular transaction state, but then we were not incorporating that transaction state into BlockState. So the account read was discarded and to compensate, we had this side effect only call much higher up in the stack. Modifying

kevaundray: So the old flow was this:

1. `process_checked_system_transaction` creates `TransactionState A`                                                                                      
2. It calls `get_account(A, target_address)` to fetch the system contract's code
3. It passes that code as a parameter into `process_system_transaction`
4. `process_system_transaction` creates `TransactionState B`
5. It builds a Message with the pre-fetched code and executes it on B
6. It calls `incorporate_tx_into_block(B, ...)`

So essentially the `get_account` read from `Transaction State A` to get the code was not being incorporated and therefore discarded.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -711,10 +708,10 @@ def process_checked_system_transaction(
         Output of processing the system transaction.
 
     """
-    system_tx_state = TransactionState(parent=block_env.state)
-    system_contract_code = get_account(system_tx_state, target_address).code
-
-    if len(system_contract_code) == 0:
+    pre_state_account = block_env.state.pre_state.get_account_optional(
+        target_address
+    )
```
kevaundray: Note this is not strictly equivalent to the previous code because it currently assumes that no previous transaction in the block somehow deployed code to that system address

kevaundray: (Previous code was checking block writes and then prestate)

kevaundray: Tests are passing, I'm guessing this is because there is no scenario that tests:

- A system contract was deployed within this block (like 4788)
- That contract is called with checks, ie its not allowed to silently fail (unlike 4788)

We don't have anything like this on mainnet right now fwiw. The consequences here would be that if a system contract like the above is deployed, then it would only be usable within the next block and not the current

src/ethereum/forks/amsterdam/vm/interpreter.py
kevaundray: New flow is this:

1. `process_checked_system_transaction` validates code exists via pre_state directly (no TransactionState involved)
2. It calls `process_system_transaction`
3. `process_system_transaction` creates one `TransactionState`
4. It calls `get_account` on that same state to fetch the code
5. It builds a Message with the code and executes it on that same state
6. It calls `incorporate_tx_into_block` on that same state

ie the TransactionState that is used to fetch the code is also the one that gets incorporated

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -711,10 +708,10 @@ def process_checked_system_transaction(
         Output of processing the system transaction.
 
     """
-    system_tx_state = TransactionState(parent=block_env.state)
-    system_contract_code = get_account(system_tx_state, target_address).code
-
-    if len(system_contract_code) == 0:
+    pre_state_account = block_env.state.pre_state.get_account_optional(
```
gurukamath: I am not super worried about this non-equivalence with the previous code. I don't think any of the system contract EIPs address this edge case behavior (same block contract deployment). However, if we want to maintain equivalence, one way would be to create the throwaway `TransactionState` instance like we were creating earlier. Just for verifying the existence of the contract. This tx_state does not incorporate into the block but that is not an issue since `process_system_transaction` does its own access and therefore, its own tracking.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -760,12 +756,9 @@ def process_unchecked_system_transaction(
         Output of processing the system transaction.
 
     """
-    system_tx_state = TransactionState(parent=block_env.state)
-    system_contract_code = get_account(system_tx_state, target_address).code
     return process_system_transaction(
```
gurukamath: At this point, we can simply rename the `process_system_transaction` to `process_unchecked_system_transaction`

src/ethereum/forks/amsterdam/fork.py
kevaundray: No strong opinion here, happy to leave it as is

gurukamath: Actually, I was looking into this in a bit more detail and it looks like both EIP-7002 and EIP-7251 mention this case. See [here](https://github.com/ethereum/EIPs/blob/2175b810f3fd7aa07f356e5e9799985ad47d04a5/EIPS/eip-7002.md?plain=1#L727) for example.

It looks like if the deployment happens in the same block, the system contract needs to be read. So we need to read the state from the tracker and not the pre state. We definitely need a test for this though. Will create one.

kevaundray: Perfect, will leave the test to you and use a throwaway state [282bca1](https://github.com/ethereum/execution-specs/pull/2264/commits/282bca123378eba220f936bc5f4d288ef840ad7f)

gurukamath (approved):
LGTM!

kevaundray (commented):
.

PR #2265, opened 2026-02-21
tests: reject leading-zero Uint RLP decodes

### LINKED ISSUE #2202: Add test for decoding RLP -> `Uint` with leading zeros
See https://github.com/ethereum/ethereum-rlp/issues/10

Yenya030:
## Summary
- add a regression test ensuring `rlp.decode_to(Uint, ...)` rejects non-canonical leading-zero encodings
- cover both `0x00` and `0x820001`

## Verification
- `uv run ruff check tests/common/test_rlp_uint_decoding.py`
- `uv run pytest tests/common/test_rlp_uint_decoding.py`

Closes #2202


SamWilsn:
Ah, this should be in the framework format, like, for example:

https://github.com/ethereum/execution-specs/blob/7ca499e04ef1c18c06f0421662ee05ceb4b998d9/tests/prague/eip7702_set_code_tx/test_invalid_tx.py#L639

@marioevz is there a better place for a pure RLP test?

marioevz:
> @marioevz is there a better place for a pure RLP test?

We don't have pure-rlp test fixtures at the moment, so the simplest way I can think of to include this test is to add another transaction test to that file where we modify one of the number fields to add zero padding to it.

Yenya030:
Thanks for the guidance — @SamWilsn @marioevz .

I moved this from a standalone pure-RLP test into the framework-format transaction tests in `tests/prague/eip7702_set_code_tx/test_invalid_tx.py`, and added coverage for the non-canonical zero integer encoding case (`0x00`) via `OversizedZeroInt` in `test_invalid_tx_invalid_nonce_encoding`.

I also removed the separate `tests/common/test_rlp_uint_decoding.py` test.

chfast:
Is this a part of official RLP spec?

tests/prague/eip7702_set_code_tx/test_invalid_tx.py
```diff
@@ -366,14 +373,15 @@ def test_invalid_tx_invalid_nonce_encoding(
...
+    oversized_type: type,
 ) -> None:
     """
     Test sending a transaction where the chain id field of an authorization has
     an incorrect encoding.
     """
 
     class ModifiedAuthorizationTuple(AuthorizationTuple):
-        nonce: OversizedInt  # type: ignore
+        nonce: oversized_type  # type: ignore
```
SamWilsn: Huh, so this did try a zero-prefixed integer prior to this pull request? How did we pass?

SamWilsn: Ah, because it works by _encoding_ the authorization tuple and comparing the hash, I think? This doesn't actually test decoding, as far as I am aware?

Yenya030: Good catch — you’re right, and your hypothesis is correct.

This passed because this test path does not deserialize transactions. It constructs typed objects, serializes them into fixture tx bytes (including signing/hash work), and records the expected error, so `rlp.decode_to(...)` for incoming tx bytes is never exercised here.

I’ll update this to be decode-focused in framework format by feeding tx bytes with a non-canonical `0x00` integer encoding and asserting rejection during tx deserialization.

SamWilsn:
> Is this a part of official RLP spec?

@chfast https://ethereum.org/developers/docs/data-structures-and-encoding/rlp includes it, but there might be a _more_ official spec somewhere.

SamWilsn:
@marioevz assigning this to you. I think the test makes sense now, but I'd like to make sure it has the proper form.

PR #2266, opened 2026-02-21
feat(zkevm): add ancestor header tracking

jsign:
## 🗒️ Description
This PR adds support for tracking ancestor headers in the `ExecutionWitness`.

The tracking is done mostly as one would expect -- the only quirk is that t8n should now also receive the block headers, not just the hashes. This is added as a new field in the input for all forks, but only Amsterdam uses it.

The block hashes input is still kept, since even if we could derive it from block headers, that would mean changing all forks' behaviour in their BLOCKHASH.

You can test this is working by running this example test that uses `BLOCKHASH`:
```
$ uv run fill --clean -m "blockchain_test" --fork Amsterdam ./tests/frontier/opcodes/test_blockhash.py -k "one_block_with_tx"
```

And if you see the output of the fixture:
```
...
                "receipts": [...],
                "executionWitness": {
                    "state": [],
                    "codes": [],
                    "headers": [
                        "0xf90278a00000000000000000000000000000000000000000000000000000000000000000a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347940000000000000000000000000000000000000000a014413270cd1509c3bc3194d1854cac72e031717b13f91dc9f14881fe2912e0efa056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421b901000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080808407270e00808000a0000000000000000000000000000000000000000000000000000000000000000088000000000000000007a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b4218080a00000000000000000000000000000000000000000000000000000000000000000a0e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
                        "0xf90278a0c0a29049136e8c01b1cc7869f95905bef6959430e8dbfb2052420b053c27f6aea01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347942adc25665018aa1fe0e6bc666dac8fc2697ff9baa0c37001fcb398e2b2c23fc42c2c662854405c77054faa96dbaff7984a90cfab93a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421b901000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080018407270e00800c80a0000000000000000000000000000000000000000000000000000000000000000088000000000000000007a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b4218080a00000000000000000000000000000000000000000000000000000000000000000a0e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855a0d19579646c765c8bd8958feabd54974d67214facc5b58ef1410a4b7fd50d1b5b"
                    ]
                },
                "blockAccessList": [...],
...
```

### DIFF
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -68,7 +68,6 @@ from .state_tracker import (
     destroy_account,
     extract_block_diffs,
     get_account,
-    get_witness_ancestors,
     incorporate_tx_into_block,
     increment_nonce,
     set_account_balance,
@@ -843,15 +842,8 @@ def apply_body(
     )
 
     block_output.execution_witness = build_execution_witness(
-        block_env.execution_witness
-    )
-    # TODO(zkevm): whenever PR #2259 is merged, rebase on
-    # top of it and make the following code add the ancestor
-    # in the block_output.execution_witness
-    ancestor_headers = get_witness_ancestors(  # noqa: F841
-        block_env.block_headers,
-        block_env.state.oldest_ancestor_offset,
-    )
+        block_env.execution_witness, block_env.state, block_env.block_headers
+    ) 
 
     return block_output
 
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -254,9 +254,7 @@ def state_transition(chain: BlockChain, block: Block) -> None:
         parent_beacon_block_root=block.header.parent_beacon_block_root,
         block_access_list_builder=BlockAccessListBuilder(),
         execution_witness=ExecutionWitnessBuilder(),
-        block_headers=[
-            rlp.encode(blk.header) for blk in chain.blocks
-        ],
+        block_headers=[rlp.encode(blk.header) for blk in chain.blocks],
     )
 
     block_output = apply_body(
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -68,10 +68,10 @@ from .state_tracker import (
     destroy_account,
     extract_block_diffs,
     get_account,
+    get_witness_ancestors,
     incorporate_tx_into_block,
     increment_nonce,
     set_account_balance,
-    get_witness_ancestors,
     track_address,
     track_ancestor_access,
 )
@@ -847,10 +847,10 @@ def apply_body(
     block_output.execution_witness = build_execution_witness(
         block_env.execution_witness
     )
-    # TODO(zkevm): whenever https://github.com/ethereum/execution-specs/pull/2259 is merged,
-    # rebase on top of it and make the following code add the ancestor in the
-    # block_output.execution_witness
-    ancestor_headers = get_witness_ancestors(
+    # TODO(zkevm): whenever PR #2259 is merged, rebase on
+    # top of it and make the following code add the ancestor
+    # in the block_output.execution_witness
+    ancestor_headers = get_witness_ancestors(  # noqa: F841
         block_env.block_headers,
         block_env.state.oldest_ancestor_offset,
     )
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -252,8 +252,9 @@ def state_transition(chain: BlockChain, block: Block) -> None:
         excess_blob_gas=block.header.excess_blob_gas,
         parent_beacon_block_root=block.header.parent_beacon_block_root,
         block_access_list_builder=BlockAccessListBuilder(),
-        execution_witness=ExecutionWitnessBuilder(),
-        block_headers=[rlp.encode(blk.header) for blk in chain.blocks],
+        execution_witness=ExecutionWitnessBuilder(
+            blockchain_headers=[rlp.encode(blk.header) for blk in chain.blocks]
+        ),
     )
 
     block_output = apply_body(
@@ -842,7 +843,7 @@ def apply_body(
     )
 
     block_output.execution_witness = build_execution_witness(
-        block_env.execution_witness, block_env.state, block_env.block_headers
+        block_env.execution_witness, block_env.state
     )
 
     return block_output
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -71,7 +71,9 @@ from .state_tracker import (
     incorporate_tx_into_block,
     increment_nonce,
     set_account_balance,
+    get_witness_ancestors,
     track_address,
+    track_ancestor_access,
 )
 from .stateless_types import ExecutionWitnessBuilder, build_execution_witness
 from .transactions import (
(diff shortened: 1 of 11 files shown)


src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -758,3 +759,51 @@ def track_storage_read(
 
     """
     tx_state.storage_reads.add((address, key))
+
+
+def get_witness_ancestors(
```
jsign: This method shouldn't live here -- is just a temporal place where both t8n and `state_transition` can share the implementation. This will be moved after #2259 is merged.

src/ethereum_spec_tools/evm_tools/loaders/fork_loader.py
```diff
@@ -290,6 +290,25 @@ def has_block_state(self) -> bool:
...
+
+    @property
+    def track_ancestor_access(self) -> Any:
+        """track_ancestor_access function of the fork."""
+        return self._module("state_tracker").track_ancestor_access
+
+    @property
+    def get_witness_ancestors(self) -> Any:
+        """get_witness_ancestors function of the fork."""
+        return self._module("state_tracker").get_witness_ancestors
```
jsign: This prob will also go away.

src/ethereum/forks/amsterdam/vm/instructions/block.py
```diff
@@ -57,6 +58,10 @@ def block_hash(evm: Evm) -> None:
         current_block_hash = evm.message.block_env.block_hashes[
             -(current_block_number - block_number)
         ]
+        track_ancestor_access(
```
jsign, line 61: `track_ancestor_access` basically tracks the furthest parent header offset that was accessed during execution.

This is the simplest and minimal information that we should record, since if the oldest block hash accessed was from, for example, 6 blocks ago, this means we should include all the parent headers up to that point.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -251,7 +252,9 @@ def state_transition(chain: BlockChain, block: Block) -> None:
         excess_blob_gas=block.header.excess_blob_gas,
         parent_beacon_block_root=block.header.parent_beacon_block_root,
         block_access_list_builder=BlockAccessListBuilder(),
-        execution_witness=ExecutionWitnessBuilder(),
+        execution_witness=ExecutionWitnessBuilder(
+            blockchain_headers=[rlp.encode(blk.header) for blk in chain.blocks]
+        ),
```
jsign, line 257: First real usage of the `ExecutionWitnessBuilder`. We provide all the (up to) 256-parent-headers as available information.

When the execution witness is finalised and built, this information is used to gather the required ancestors, considering the oldest parent accessed through BLOCKHASH and EIP-2935 system contract.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -836,7 +843,7 @@ def apply_body(
     )
 
     block_output.execution_witness = build_execution_witness(
-        block_env.execution_witness
+        block_env.execution_witness, block_env.state
```
jsign, line 846: So we have access to `oldest_ancestor_offset` accessed during STF execution.

src/ethereum/forks/amsterdam/stateless_types.py
```diff
@@ -43,25 +43,49 @@ class ExecutionWitnessBuilder:
     Mutable accumulator for execution witness data during block execution.
     """
 
+    blockchain_headers: List[Bytes] = field(default_factory=list)
     state: List[Bytes] = field(default_factory=list)
     codes: List[Bytes] = field(default_factory=list)
```
jsign, line 48: Separate note: as we keep integrating bytecode tracking and IncrementalMPT, prob these fields will be renamed or might mutate a bit in the same way this happened with the previous `header` transforming into `blockchain_headers`.

src/ethereum_spec_tools/evm_tools/t8n/env.py
```diff
@@ -301,6 +303,34 @@ def read_block_hashes(self, data: Any) -> None:
 
         self.block_hashes = block_hashes
 
+    def read_block_headers(self, data: Any, t8n: "T8N") -> None:
```
jsign, line 306: We read the new data provided by t8n from the testing framework.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -758,3 +759,27 @@ def track_storage_read(
 
     """
     tx_state.storage_reads.add((address, key))
+
+
+def track_ancestor_access(block_state: BlockState, offset: Uint) -> None:
+    """
+    Record that an ancestor block was accessed.
+
+    Update ``oldest_ancestor_offset`` if ``offset`` is further back (larger)
```
kevaundray, line 768: This is not done "automatically" like `get_account` and `get_storage` but I think its fine and nor really sure how we would do it automatically in a way that was not very invasive

jsign: Yeah, I think this ancestor access might not allow us to do the "automatic" style, since it is not part of the state thus might not have an indirect other use-case to hook into.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -758,3 +759,27 @@ def track_storage_read(
...
+    block_state :
+        The block state.
+    offset :
+        Offset from the current block to the ancestor that was
+        accessed.
+
+    """
+    if (
+        block_state.oldest_ancestor_offset is None
+        or offset > block_state.oldest_ancestor_offset
```
kevaundray, line 783: Nit: Not a blocker and should be done in a separate PR since this works -- I wonder if we can just pass in the blockindex and then compute the offset when its needed.

Current logic is essentially:

```python
if block_state.oldest_ancestor_offset is None:
    block_state.oldest_ancestor_offset = offset
else:
    block_state.oldest_ancestor_offset = max(offset, block_state.oldest_ancestor_offset)
```

I had to think twice re the `max` or the `>`

kevaundray: But yeah as I said not a blocker and if you think its clearer as is then lets keep :)

kevaundray (approved):

src/ethereum/forks/amsterdam/state_tracker.py
jsign: No strong opinion -- I felt the current way somewhat clear, but let's merge it as is and we can always polish later since isn't strictly required now. :)
Edit: waiting for CI to "stricly open for review"

Issue #2154, opened 2026-02-06
Use a common `Account` type across all forks


petertdavies:
There is a different `Account` type for each fork, despite the fact that `Account` has not and will not change between forks. It tends to bleed between forks and leave the fork specific code world causing various headaches that would be avoided if there was only one `Account` type.

jsign:
@petertdavies @gurukamath, want to raise a subtle point that will be fundamental for the stateless STF that will happen in the future. Sharing this since I discovered it while doing my e2e-PoC.

The current `Account` has the `code` field. This means that any `get_account` implies whatever is backing the pre-state abstraction must always be the account **with** the bytecode. This is not entirely correct for a stateless STF.

In a stateless world, you might call `BALANCE accountA` and that should only force the account data to be in the witness. The bytecode is not relevant i.e. would bloat the witness.

This also has implications to https://github.com/ethereum/execution-specs/issues/2155. Mostly because I believe we need `get_account_data` (instead of `get_account`) and also a separate `get_bytecode(code_hash)`.

LMK if this makes sense to you, since I can explain in more detail if needed!

petertdavies:
@jsign Good point. Created #2156.

gurukamath:
> LMK if this makes sense to you, since I can explain in more detail if needed!

Yes. This is a good point and we have run into some issues recently due to having the code directly accessible in the account. We should switch to `codehash` based code access.

SamWilsn:
Are there other types we should pull out as well? Like `Address`?

kevaundray:
Amsterdam doesn't have a fork-specific account type in `fork_types.py` like other forks, so I think this may fall under the [backport related](https://github.com/ethereum/execution-specs/issues/2245)

what do you think @gurukamath ?

kevaundray:
Created this issue to track Sam's question https://github.com/ethereum/execution-specs/issues/2293 in general, since this one was about the `Account` type

gurukamath:
This has been handled in #2750

Issue #2245, opened 2026-02-20
Changes to be back-ported from Amsterdam


gurukamath:
Sometimes, certain changes made to the Amsterdam fork need to be back-ported to older forks. However, doing this in one PR could blow up the size of the PR and make it unmanageable. This issues tracks such changes, that need a subsequent PR to back-port them to the older forks

- [ ] Implement state tracker in the older forks ( See ethereum/execution-specs#2207)
    - [ ] Remove fork independent constants/functions/classes from `fork_loader` (See [comment](https://github.com/ethereum/execution-specs/pull/2267#discussion_r2836412797))
    - [ ] Add `Amsterdam` specific imports from the t8n tool to `ForkLoader` (See [comment](https://github.com/ethereum/execution-specs/pull/2207#discussion_r2816866110))
- [x] `Account` dataclass should have `code_hash` instead of `code` (See ethereum/execution-specs#2244) (Closed in ethereum/execution-specs#2345)
- [ ] Fork agnostic `State` - ethereum/execution-specs#2381
- [ ] Update system contract access (See [comment](https://github.com/ethereum/execution-specs/pull/2264#discussion_r2835713125))
- [ ] Partially decouple state_transition from BlockChain (See ethereum/execution-specs#2307)
- [ ] Accept encoded `LegacyTransaction` (See ethereum/execution-specs#2365)
- [ ] Delay `get_code` call (See ethereum/execution-specs#2473)

fselmo:
Going to put this one here as well #1867. I have an older version of this locally, will make sure to get this in.

I assume the state tracking and gas accounting may bleed together. Perhaps it may be good to wait for the state tracker to be fully implemented in `forks/amsterdam` to understand how these changes work together.

PR #2267, opened 2026-02-21
chore(trie): Make `trie` Fork agnostic

kevaundray:
## 🗒️ Description

Bullet point 3 from #2207 

This makes the `Trie` data structure and its methods fork agnostic. 

There were some tox related errors that were surfacing due to `Account` not being fork agnostic for forks pre-Amsterdam. I've changed those imports for Account to point to the one defined in `ethereum.state` for simplicity (this is what I imagine the endgame solution tends towards, given the issue on making Account fork agnostic)  

## 🔗 Related Issues or PRs

https://github.com/ethereum/execution-specs/issues/2293

Related to https://github.com/ethereum/execution-specs/issues/2154 -- we could also pull out Account in a separate PR

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


kevaundray:
The issue with `EMPTY_ACCOUNT`, also applies to `Account` -- have removed those re-exports from fork_types since `Account` and `EMPTY_ACCOUNT` doesn't change across forks.

We now import from `ethereum.state` for previous forks

src/ethereum/forks/amsterdam/trie.py
```diff
@@ -12,456 +12,21 @@
 `.fork_types.Account` objects.
 """
```
kevaundray: Iteratively changing this, should be able to delete it soon

src/ethereum_spec_tools/evm_tools/loaders/fork_loader.py
```diff
@@ -204,9 +204,7 @@ def Bloom(self) -> Any:
     @property
     def EMPTY_ACCOUNT(self) -> Any:
         """EMPTY_ACCOUNT of the fork."""
-        if self.has_block_state:
-            return EMPTY_ACCOUNT
-        return self._module("fork_types").EMPTY_ACCOUNT
+        return EMPTY_ACCOUNT
```
kevaundray, line 207: See comment [here](https://github.com/ethereum/execution-specs/pull/2267/changes#r2836451460) on acknowledging that this is no longer fork dependent.

I didn't check what changes if we removed these methods though (ie what the additional diff would look like)

tests/json_infra/test_tools_new_fork.py
```diff
@@ -89,12 +89,6 @@ def test_end_to_end(template_fork: str) -> None:
         with (fork_dir / "fork.py").open("r") as f:
             assert "MAX_BLOB_GAS_PER_BLOCK = U64(99)" in f.read()
 
-        with (fork_dir / "trie.py").open("r") as f:
-            assert (
-                "from ethereum.forks.paris import trie as previous_trie"
-                in f.read()
-            )
```
kevaundray, line 96: trie.py is no longer fork dependent, so I removed this check in new-fork tool

src/ethereum_spec_tools/evm_tools/loaders/fork_loader.py
```diff
@@ -322,18 +322,18 @@ def create_ether(self) -> Any:
...
-        return self._module("trie").copy_trie
+        """copy_trie function."""
+        return copy_trie
 
     @property
     def trie_get(self) -> Any:
-        """trie_get function of the fork."""
-        return self._module("trie").trie_get
+        """trie_get function."""
+        return trie_get
```
kevaundray, line 343: These are no longer dependent on the fork, left them here to not increase the diff further in this PR.

kevaundray:
I think the rough shape of this is now visible, so would say the experiment is done. 

CC @SamWilsn and @gurukamath for input when you get back next week.

Kind of a PITA that it touches so many files, but this was somewhat inevitable

src/ethereum/trie.py
```diff
@@ -139,7 +132,7 @@ def encode_internal_node(node: Optional[InternalNode]) -> Extended:
         return keccak256(encoded)
 
 
-def encode_node(node: Node, storage_root: Optional[Bytes] = None) -> Bytes:
```
gurukamath:
```suggestion
def encode_node(node: Optional[Extended], storage_root: Optional[Bytes] = None) -> Bytes:
```
And then handle the `None` case explicitly in the function body to raise the relevant `AssertionError`. The `V` typevar also becomes `V = TypeVar("V", bound=Optional[Extended])` in that case.

For context, we do use the broad `Any` type in some tooling related code but try to avoid it in the specs code itself.

src/ethereum_spec_tools/evm_tools/loaders/fork_loader.py
gurukamath: This can be handled during the back-porting. Have added this to the back-porting tracker issue. #2245

src/ethereum/trie.py
```diff
@@ -70,28 +66,25 @@
...
-    Optional[LegacyTransaction | Bytes],
-    Optional[Receipt | Bytes],
-    Optional[Withdrawal | Bytes],
-    Uint,
-    U256,
-)
+V = TypeVar("V")
+
+
+def encode_account(raw_account_data: Account, storage_root: Bytes) -> Bytes:
```
gurukamath, line 73: Perhaps, this belongs in `ethereum.state` since that is also where we have defined `Account`

gurukamath (commented):
Largely looks good to me. Thanks for taking this on :). Just left a couple of smaller comments.

src/ethereum/trie.py
```diff
@@ -139,7 +132,7 @@ def encode_internal_node(node: Optional[InternalNode]) -> Extended:
         return keccak256(encoded)
 
 
-def encode_node(node: Node, storage_root: Optional[Bytes] = None) -> Bytes:
+def encode_node(node: Any, storage_root: Optional[Bytes] = None) -> Bytes:
```
SamWilsn, line 135: No `Any` please.

src/ethereum_optimized/state_db.py
SamWilsn, line 40: If there's only one `Account` now, I don't think we need this `Any` any longer.

src/ethereum_optimized/state_db.py
```diff
@@ -57,11 +58,8 @@ def get_optimized_state_patches(fork: str) -> Dict[str, Any]:
     """
     patches: Dict[str, Any] = {}
 
-    types_mod = cast(
-        Any, import_module("ethereum.forks." + fork + ".fork_types")
-    )
     state_mod = cast(Any, import_module("ethereum.forks." + fork + ".state"))
-    Account = types_mod.Account  # noqa N806
+    Account = Account_cls  # noqa: N806
```
SamWilsn, line 62: What is the purpose of this line?

SamWilsn (commented):
This makes sense. I guess maybe it might be reasonable to rename it to `mp_trie.py` eventually if we need to differentiate between trie implementations.

My internal reasoning for accepting this hoist is that the diffs between the MP trie and any other trie would be extremely messy and not particularly useful.

src/ethereum/trie.py
fselmo: This is more just a comment from a quick fly-by. @gurukamath any reason to not use `Extended | None` instead of importing `Optional` just to use `Optional[Extended]`? It seems to be the preferred usage for Python >= 3.10. Of course, we can make our own decisions on that, but just curious.

kevaundray:
I think this is semi-blocked on backporting `Account` to previous forks because Amsterdam has Account with code hash while other forks have Account with `code` and a part of making the Trie fork agnostic is that the `Account` is also fork agnostic.

We could have each fork define an `encode_account` method and `Account` class for now, but I think its probably better to backport first

kevaundray:
superceded by #2381

Issue #2212, opened 2026-02-14
Question: Which execution-spec modules are canonical for EIP-4844 blob gas pricing?


Fosurero:
I'm building PRSpec, a tool that compares client implementations against EIP specs and execution-specs. For EIP-4844, I'm pulling the reference implementation from this repo to use as the specification baseline.

Currently I reference the cancun fork modules. Are there specific modules or functions that are considered the canonical reference for:
- Blob gas price calculation (fake exponential)
- Max blobs per block validation
- Excess blob gas tracking

Want to make sure my tool is comparing against the right source of truth.

Tool: https://github.com/Fosurero/PRSpec

danceratopz:
Hey @Fosurero, thanks for reaching out!

Jicymi: The execution-specs follow the [WET principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#WET) and a new fork is created from the current fork by literally copy-pasting the current forks directory. This means that to see all the changes introduced in any fork version, you can diff it to its ancestor, e.g.,

```
diff src/ethereum/forks/shanghai/ src/ethereum/forks/cancun/
```

Also, of limited additional use in automated tooling, but the EELS documentation flow generates diffs from one fork to the next. You can browse the [differences between Shanghai and Cancun here](https://ethereum.github.io/execution-specs/diffs/shanghai/cancun/index.html).

Does that help answer your question?

Fosurero:
@danceratopz That's extremely helpful, thank you

The WET approach makes a lot of sense for PRSpec's file mapping diffing between fork directories gives me a clean way to isolate EIP-specific changes programmatically. I'll integrate this into PRSpec's spec fetching pipeline.

The EELS diff documentation is also a great reference I wasn't aware of it. This could serve as a validation layer to cross-check PRSpec's automated spec extraction against the canonical diffs.

I'll update PRSpec to use fork-to-fork diffing as the primary method for identifying EIP implementation boundaries in execution-specs. Thanks again for the guidance

spencer-tb:
Closing the issue for now. Please re-open if there is anything else!

PRSpec looks amazing btw - awesome job :D

PR #2271, opened 2026-02-21
chore(do not merge): Experiment with caching for "Build Doc" Job

kevaundray:
## 🗒️ Description

Was briefly looking into this to check why it takes so long.

Wrote some unorganized notes below for my own reference. (It could be wrong)

From my understanding, "Build Doc" is creating diffs across each fork so we can see how a file has changed over forks. 

If this is correct, then this jobs run time will increase if more forks are added.

I believe it does a pairwise diff, ie if we have the three forks:

- Frontier
- Homestead
- Amsterdam

Then it will create a diff for:

- Frontier -> Homestead
- Homestead -> Amsterdam

This is a file by file diff, so if frontier has 38 files and homestead has 38 files, then we would be diffing 38 file pairs.

-----

If we take the Frontier -> Homestead diff as an example. Lets say the code wants to check the diff between `frontier/vm/gas.py` and `homestead/vm/gas.py`.  It does the following:

- Parse both files into ASTs, so essentially converts the source code into a data structure
- Runs a TreeDiff algorithm, that compares the two ASTs to find differences.  (This is roughly O(n^2) in complexity
- Applies the result of the tree diff to generate the html rendering.(I'm a bit vague here as I didn't look too deep into this part)

Hypothesis: The tree diff algorithm is the expensive part. Noting that even if there is no change between Frontier and Homestead, I believe it still runs the algorithm, to then conclude that nothing has changed.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -39,6 +39,23 @@ jobs:
           version: ${{ vars.UV_VERSION }}
           python-version: ${{ vars.DEFAULT_PYTHON_VERSION }}
 
+      - uses: ./.github/actions/get-changed-files
+        if: github.event_name == 'pull_request'
+        id: get-changed-files
+
+      - name: Detect changed forks
+        if: github.event_name == 'pull_request'
+        run: |
+          FILE_LIST="${{ steps.get-changed-files.outputs.file_list }}"
+          FORKS=$(grep '^src/ethereum/forks/' "$FILE_LIST" | cut -d/ -f4 | sort -u | paste -sd, - || true)
+
+          if [ -n "$FORKS" ] && ! grep -qv '^src/ethereum/forks/' "$FILE_LIST"; then
+            echo "DOCC_DIFF_FORKS=$FORKS" >> "$GITHUB_ENV"
+            echo "Building diffs for changed forks only: $FORKS"
+          else
+            echo "Non-fork files changed, building all diffs"
+          fi
+
       - name: Build Documentation
         run: |
           uvx tox -e spec-docs
--- a/.github/workflows/gh-pages.yaml
+++ b/.github/workflows/gh-pages.yaml
@@ -31,17 +31,17 @@ jobs:
           fetch-depth: 0
           submodules: recursive
 
-      - name: Setup Python
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065
+      - name: Install uv ${{ vars.UV_VERSION }} and python ${{ vars.DEFAULT_PYTHON_VERSION }}
+        uses: astral-sh/setup-uv@0c5e2b8115b80b4c7c5ddf6ffdd634974642d182 # v5.4.1
         with:
-          python-version: "3.11"
-
-      - name: Install Tox
-        run: pip install tox
+          enable-cache: true
+          cache-dependency-glob: "uv.lock"
+          version: ${{ vars.UV_VERSION }}
+          python-version: ${{ vars.DEFAULT_PYTHON_VERSION }}
 
       - name: Build Documentation
         run: |
-          tox -e spec-docs
+          uvx tox -e spec-docs
           touch .tox/docs/.nojekyll
 
       - name: Upload Pages Artifact
--- a/src/ethereum_spec_tools/docc.py
+++ b/src/ethereum_spec_tools/docc.py
@@ -1116,11 +1116,9 @@ class MinimizeDiffsTransform(Transform):
     """
 
     _count: int
-    _total: int
 
-    def __init__(self, settings: PluginSettings) -> None:
+    def __init__(self, settings: PluginSettings) -> None:  # noqa: ARG002
         self._count = 0
-        self._total = 0
 
     def transform(self, context: Context) -> None:
         """
--- a/src/ethereum_spec_tools/docc.py
+++ b/src/ethereum_spec_tools/docc.py
@@ -1134,9 +1134,8 @@ class MinimizeDiffsTransform(Transform):
             kind = "diff"
 
         logging.info(
-            "MinimizeDiffs [%d]: %s (%s)",
+            "MinimizeDiffs [%d] (%s)",
             self._count,
-            context[Source].output_path,
             kind,
         )
 
--- a/src/ethereum_spec_tools/docc.py
+++ b/src/ethereum_spec_tools/docc.py
@@ -19,6 +19,7 @@ Plugins for docc specific to the Ethereum execution specification.
 
 import dataclasses
 import logging
+import os
 from collections import defaultdict
 from itertools import tee, zip_longest
 from pathlib import Path, PurePath
@@ -85,6 +86,25 @@ class EthereumDiscover(Discover):
         forks = base / "forks"
         self.forks = Hardfork.discover([str(forks)])
 
+    @staticmethod
+    def _diff_forks() -> Optional[FrozenSet[str]]:
+        """
+        Return fork short names to limit diffing to, or None for all.
+
+        Controlled by the DOCC_DIFF_FORKS environment variable (comma-
+        separated fork short names).  When set, only fork pairs where
+        at least one side is listed will be discovered.
+        """
+        env = os.environ.get("DOCC_DIFF_FORKS", "").strip()
+        if not env:
+            return None
+        names = frozenset(
+            name.strip() for name in env.split(",") if name.strip()
+        )
+        if names:
+            logging.info("Limiting diffs to forks: %s", ", ".join(names))
+        return names or None
+
     def discover(self, known: FrozenSet[T]) -> Iterator[Source]:
         """
         Find sources.
@@ -121,8 +141,17 @@ class EthereumDiscover(Discover):
 
             by_fork[fork][fork_relative_path] = source
 
+        diff_forks = self._diff_forks()
+
         diff_count = 0
         for before, after in pairwise(self.forks):
+            if diff_forks is not None:
+                if (
+                    before.short_name not in diff_forks
+                    and after.short_name not in diff_forks
+                ):
+                    continue
+
             paths = set(by_fork[before].keys()) | set(by_fork[after].keys())
 
             for path in paths:
--- a/src/ethereum_spec_tools/docc.py
+++ b/src/ethereum_spec_tools/docc.py
@@ -315,6 +315,17 @@ class DiffNode(Node):
         )
 
 
+class IdenticalDiffNode(DiffNode):
+    """
+    A DiffNode where both sides have identical source content.
+
+    FixIndexTransform still rewrites identifiers, but
+    MinimizeDiffsTransform skips tree diffing entirely.
+    """
+
+    pass
+
+
 class EthereumBuilder(PythonBuilder):
     """
     A `PythonBuilder` that additionally builds `Document`s from `DiffSource`s.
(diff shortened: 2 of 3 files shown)


kevaundray:
It skips around 70% of pairs and fails after 40 minutes. Barring the actual error, this makes sense since it was previously taking an hour and we shaved off 30% of the work.

This gives weight to the claim that the bottleneck is the tree diffing, but need to have better logs to confirm

kevaundray:
Seems the diffs rely on https://github.com/SamWilsn/fladrif

kevaundray:
Currently it seems to build docs for all forks unconditionally -- I think most people will modify the latest fork(currently amsterdam), so a lot of this is also repeated work that should never fail.

I think one reasonable optimization here is to have two settings:

- When its a PR, only build docs for fork pairs that have changed. So for people building ontop of amsterdam, build docs would only build for bpo5 -> amsterdam
- When it gets merged into forks/amsterdam, build for all forks and deploy the new docs

Rationale here is that docs are not released on the PRs anyways, so building everything is not needed

kevaundray:
Added the above suggestion, but it won't be noticeable in this PR because any change outside of a fork(like the github action fileS) will by default trigger a full rebuild -- should be runnable locally by setting `DOCC_DIFF_FORKS=amsterdam`

kevaundray:
Diffs are published [here](https://ethereum.github.io/execution-specs/diffs/bpo5/amsterdam/vm/exceptions.py.html) for reference.

@SamWilsn Do you know if folks are using this? (Asking because whenever I visit the specs, I usually just look at the code)

If so I think we could rewrite it in Rust to be faster, given that this will grow in time as we add more forks. I wrote a proof of concept here https://kevaundray.github.io/execution-specs-viewer/ that compiles in about a minute using Rust. It uses a different stack etc, but largely gives the features I think you originally intended?

kevaundray:
So if we just build docs for amsterdam, it takes around 40 minutes, which I think is still pretty high

danceratopz:
> Diffs are published [here](https://ethereum.github.io/execution-specs/diffs/bpo5/amsterdam/vm/exceptions.py.html) for reference.
> 
> @SamWilsn Do you know if folks are using this? (Asking because whenever I visit the specs, I usually just look at the code)
> 
> If so I think we could rewrite it in Rust to be faster, given that this will grow in time as we add more forks. I wrote a proof of concept here https://kevaundray.github.io/execution-specs-viewer/ that compiles in about a minute using Rust. It uses a different stack etc, but largely gives the features I think you originally intended?

Thanks for taking a look at this, @kevaundray. Yes, the docs build is too slow. I got it down to about ~4-5 minutes locally by tweaking the existing [build tool](https://github.com/SamWilsn/docc), but I never PR'd these changes / followed up with @SamWilsn. Perhaps it's time to do that. My changes didn't change the fork diffset, however, which is a also great idea!

kevaundray:
> > Diffs are published [here](https://ethereum.github.io/execution-specs/diffs/bpo5/amsterdam/vm/exceptions.py.html) for reference.
> 
> > 
> 
> > @SamWilsn Do you know if folks are using this? (Asking because whenever I visit the specs, I usually just look at the code)
> 
> > 
> 
> > If so I think we could rewrite it in Rust to be faster, given that this will grow in time as we add more forks. I wrote a proof of concept here https://kevaundray.github.io/execution-specs-viewer/ that compiles in about a minute using Rust. It uses a different stack etc, but largely gives the features I think you originally intended?
> 
> 
> 
> Thanks for taking a look at this, @kevaundray. Yes, the docs build is too slow. I got it down to about ~4-5 minutes locally by tweaking the existing [build tool](https://github.com/SamWilsn/docc), but I never PR'd these changes / followed up with @SamWilsn. Perhaps it's time to do that. My changes didn't change the fork diffset, however, which is a also great idea!

Ah nice, curious to know what  you changed!

I guess I'm still curious to know if this is actually used, given the work that needs to be done, grows with the number of hardforks

danceratopz:
> > > Diffs are published [here](https://ethereum.github.io/execution-specs/diffs/bpo5/amsterdam/vm/exceptions.py.html) for reference.
> > 
> > 
> > > 
> > 
> > 
> > > @SamWilsn Do you know if folks are using this? (Asking because whenever I visit the specs, I usually just look at the code)
> > 
> > 
> > > 
> > 
> > 
> > > If so I think we could rewrite it in Rust to be faster, given that this will grow in time as we add more forks. I wrote a proof of concept here https://kevaundray.github.io/execution-specs-viewer/ that compiles in about a minute using Rust. It uses a different stack etc, but largely gives the features I think you originally intended?
> > 
> > 
> > Thanks for taking a look at this, @kevaundray. Yes, the docs build is too slow. I got it down to about ~4-5 minutes locally by tweaking the existing [build tool](https://github.com/SamWilsn/docc), but I never PR'd these changes / followed up with @SamWilsn. Perhaps it's time to do that. My changes didn't change the fork diffset, however, which is a also great idea!
> 
> Ah nice, curious to know what you changed!

Will try and get some PRs up soon! But quite a few cache optimizations and replaced `libcst` (Concrete Syntax Tree) library with Python's built-in `ast` module for parsing Python source.

> I guess I'm still curious to know if this is actually used, given the work that needs to be done, grows with the number of hardforks

I think the diff is really nice feature, but I suspect it has a limited number of users, but there was recently this comment:

> The EELS diff documentation is also a great reference I wasn't aware of it. This could serve as a validation layer to cross-check PRSpec's automated spec extraction against the canonical diffs.

https://github.com/ethereum/execution-specs/issues/2212#issuecomment-3915677311

kevaundray:
> > > > Diffs are published [here](https://ethereum.github.io/execution-specs/diffs/bpo5/amsterdam/vm/exceptions.py.html) for reference.
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > @SamWilsn Do you know if folks are using this? (Asking because whenever I visit the specs, I usually just look at the code)
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > If so I think we could rewrite it in Rust to be faster, given that this will grow in time as we add more forks. I wrote a proof of concept here https://kevaundray.github.io/execution-specs-viewer/ that compiles in about a minute using Rust. It uses a different stack etc, but largely gives the features I think you originally intended?
> 
> > > 
> 
> > > 
> 
> > > Thanks for taking a look at this, @kevaundray. Yes, the docs build is too slow. I got it down to about ~4-5 minutes locally by tweaking the existing [build tool](https://github.com/SamWilsn/docc), but I never PR'd these changes / followed up with @SamWilsn. Perhaps it's time to do that. My changes didn't change the fork diffset, however, which is a also great idea!
> 
> > 
> 
> > Ah nice, curious to know what you changed!
> 
> 
> 
> Will try and get some PRs up soon! But quite a few cache optimizations and replaced `libcst` (Concrete Syntax Tree) library with Python's built-in `ast` module for parsing Python source.
> 
> 
> 
> > I guess I'm still curious to know if this is actually used, given the work that needs to be done, grows with the number of hardforks
> 
> 
> 
> I think the diff is really nice feature, but I suspect it has a limited number of users, but there was recently this comment:
> 
> 
> 
> > The EELS diff documentation is also a great reference I wasn't aware of it. This could serve as a validation layer to cross-check PRSpec's automated spec extraction against the canonical diffs.
> 
> 
> 
> https://github.com/ethereum/execution-specs/issues/2212#issuecomment-3915677311

Ah interesting, I didn't get to the parser, amazing! 

In that issue, it seemed like the user could benefit from just a normal diff between two forks but I could be mistaken as I didn't look carefully at their project. 

Though I think that if the time can be brought down to 4-5 minutes then my question no longer becomes less consequential since it is no longer the bottleneck

SamWilsn:
`docc` was started as a way to migrate away from our custom sphinx plugin, and I pretty much stopped working on it when we reached parity. There are a _ton_ of low-hanging fruits, for example:

- It's roughly architected to support parallelization, but doesn't do any. 
- Rendering to HTML has multiple passes from HTML -> syntax tree -> HTML.

> @SamWilsn Do you know if folks are using this? (Asking because whenever I visit the specs, I usually just look at the code)

I mean, I use it, especially whenever I'm trying to explain a change to someone else. One of the biggest selling points for creating EELS was the ability to show both snapshot (the source itself) and consensus-specs-style diffs.

That said, I don't think https://ethereum.github.io/execution-specs gets much attention. A lot of the fault there lies with me not continuing to improve `docc`. It's hardly usable in its current state.

> If so I think we could rewrite it in Rust to be faster, given that this will grow in time as we add more forks.

I'd love to write this in Rust! Would give an opportunity to fix a lot of the architectural issues in `docc`. I had originally used Python so the rest of the team would feel more comfortable, but I'm not sure that was actually a smart decision.

> I wrote a proof of concept here https://kevaundray.github.io/execution-specs-viewer/ that compiles in about a minute using Rust.

Holy shit that's impressive! Tree sitter was what I was going to use originally before I found libcst (which actually has [a partial implementation in rust](https://github.com/Instagram/LibCST/tree/main/native)).

Regardless of where we go, I'd love to get your help on the frontend side of things.

> It uses a different stack etc, but largely gives the features I think you originally intended?

Most of the complexity of `docc` isn't in the diff functionality, it's in everything else. For example, `docc` is aware of typing and definitions, so you can actually navigate through the source:

[Screencast_20260223_143534.webm](https://github.com/user-attachments/assets/a6a29686-7aac-4619-871a-e71bb206d02d)

kevaundray:
> `docc` was started as a way to migrate away from our custom sphinx plugin, and I pretty much stopped working on it when we reached parity. There are a _ton_ of low-hanging fruits, for example:
> 
> * It's roughly architected to support parallelization, but doesn't do any.
> * Rendering to HTML has multiple passes from HTML -> syntax tree -> HTML.
> 
> > @SamWilsn Do you know if folks are using this? (Asking because whenever I visit the specs, I usually just look at the code)
> 
> I mean, I use it, especially whenever I'm trying to explain a change to someone else. One of the biggest selling points for creating EELS was the ability to show both snapshot (the source itself) and consensus-specs-style diffs.
> 
> That said, I don't think https://ethereum.github.io/execution-specs gets much attention. A lot of the fault there lies with me not continuing to improve `docc`. It's hardly usable in its current state.
> 
> > If so I think we could rewrite it in Rust to be faster, given that this will grow in time as we add more forks.
> 
> I'd love to write this in Rust! Would give an opportunity to fix a lot of the architectural issues in `docc`. I had originally used Python so the rest of the team would feel more comfortable, but I'm not sure that was actually a smart decision.
> 
> > I wrote a proof of concept here https://kevaundray.github.io/execution-specs-viewer/ that compiles in about a minute using Rust.
> 
> Holy shit that's impressive! Tree sitter was what I was going to use originally before I found libcst (which actually has [a partial implementation in rust](https://github.com/Instagram/LibCST/tree/main/native)).
> 
> Regardless of where we go, I'd love to get your help on the frontend side of things.
> 
> > It uses a different stack etc, but largely gives the features I think you originally intended?
> 
> Most of the complexity of `docc` isn't in the diff functionality, it's in everything else. For example, `docc` is aware of typing and definitions, so you can actually navigate through the source:
> 
>  Screencast_20260223_143534.webm

Thanks for the clarification and context! This all makes sense and I think the go-to definition functionality is very useful :)

> continuing to improve docc

When you get the cycles, it may be interesting to scope out your vision of it in an issue, so others can maybe come and implement it.

> I'd love to write this in Rust! 

This was what I originally wanted to do! 

Intuitively, I would think an architecture that allows one to do something like `docc --old=forks/osaka --new=forks/amsterdam` would scale the best. So the tool itself would just create the diff and html between two forks and this can be ran in parallel in the CI.  The CI would then be in charge of not running docc for old forks that haven't changed because its cached.

Didn't dive too deep into this, so there might be something I was missing here.

> I'd love to get your help on the frontend side of things.

Ah I was using claude 🤣 it's pretty likely that you have more design skill points than me :)

> Regardless of where we go

Would leave this upto you and Dan, happy to follow with what you folks think makes sense. My initial investigation was aimed at reducing the CI times since this gets ran on every PR, but it seems Dan may have a way to reduce this down to less than 10 minutes, so its not that much of a concern

SamWilsn (commented):
Code-wise, I think this makes a lot of sense!

What about just disabling diffs when building on a branch? The artifacts don't get uploaded anywhere, and any missing references will get caught by the non-diff build anyway.

kevaundray:
> Code-wise, I think this makes a lot of sense!
> 
> What about just disabling diffs when building on a branch? The artifacts don't get uploaded anywhere, and any missing references will get caught by the non-diff build anyway.

I think this would be much better

kevaundray:
Closing due to #2296

PR #2299, opened 2026-02-24
feat(zkevm): fixe bytecode tracking and inclusion in the execution witness

jsign:
## 🗒️ Description
This PR fix a bug in the bytecode tracking and witness inclusion. Before we only tracked a set of `code_hash`, and included them in the witness if the `code_hash` existed in the pre-state.

This is not correct because if a block contains a single transaction that creates a contract with `code_hash_1`, and another already-deployed contract with `code_hash_1` exists (e.g., not accessed during the block), then the bytecode would be included in the execution witness. Since this code is generated at runtime, it isn't required to be provided.

The way this is solved is to also track which account was the target of the `code_read`. This allows later to check if the tracked bytecode was created during runtime or not, compared with the previous `code_hash` for that account in the pre-state.

Additionally, I added two tests that covers two important cases:
- The simple case: a block that creates some bytecode. This bytecode must not appear in the generated execution witness.
- A more nuanced case: tx1 in the block access an existing contract with `code_hash_1`, and later a tx2 creates a new contract also with `code_hash_1`. The bytecode in this case must appear, since despite tx2 generated this bytecode at runtime, tx1 accessed it from an existing contract in the pre-state.

This was discovered by comparing our Amsterdam witness generation against my e2e working on from Osaka (for some tests -- I need to do a deeper comparison!).

### DIFF
--- a/src/ethereum/forks/amsterdam/stateless_types.py
+++ b/src/ethereum/forks/amsterdam/stateless_types.py
@@ -10,7 +10,7 @@ from ethereum_types.frozen import slotted_freezable
 from ethereum_types.numeric import Uint
 
 from ethereum.crypto.hash import Hash32
-from ethereum.state import PreState
+from ethereum.state import Address, PreState
 
 from .state_tracker import BlockState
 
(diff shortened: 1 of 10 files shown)


src/ethereum/forks/amsterdam/stateless_types.py
```diff
@@ -74,26 +74,33 @@ def build_execution_witness(
...
-        Code hashes accessed during block execution.
+        Code reads as ``(address, code_hash)`` during block execution.
     pre_state :
         The pre-execution state.
 
     """
+    witness_code_hashes: Set[Hash32] = set()
+    for address, code_hash in code_reads:
+        pre_account = pre_state.get_account_optional(address)
+        if pre_account is None or pre_account.code_hash != code_hash:
```
jsign, line 98: Important note here: `pre_account.code_hash != code_hash` is not only useful for covering when a contract was created. But with 7702 you can keep changing the delegated contract for an existing account, thus technically speaking 7702 is the only way you can change the `code_hash` of existing accounts.

The more general rule (used here) is that if the tracked bytecode is different from the existing one, then it was runtime related (i.e. created, or came from an 7702 new delegation).

tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def test_witness_excludes_bytecode_created_in_same_block(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+) -> None:
+    """
+    Create a contract in a block without reading its code first.
+
+    Manual witness check until execution witness assertions are supported:
```
jsign, line 28: > until execution witness assertions are supported

I think we can work on this feature in the testing framework eventually, so we can add the assertion for expected `bytecodes` or similar as today exists for `post` state.

This to avoid manual checks ofc.

kevaundray (approved):

PR #2300, opened 2026-02-24
docs(specs, tests): fixed comments with values related to gas limits

SamueleA:
## 🗒️ Description
Updated the code comments and documentation related to gas limit which were out of date since the Osaka fork and the inclusion of EIP-7825 and EIP-7935.

1. In the tests, the comments related to the the gas limit per transaction showed an incorrect value of 30 million ( see  [EIP-7825](https://eips.ethereum.org/EIPS/eip-7825)) and were changed to a more general version that will work for all forks.

2. In the specs, the comments related to the maximum limit per block showed an incorrect value of 30 million (see  [EIP-7935](https://eips.ethereum.org/EIPS/eip-7935)). The comments were updated to not include a hard-coded number and the explanation was clarified in explaining how the gas limit relates to the base fee.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture
IMG_20250926_215245458_AE


src/ethereum/forks/amsterdam/blocks.py
```diff
@@ -164,7 +164,7 @@ class Header:
     [EIP-1559], this is still the maximum gas limit, but the base fee per gas
     is also considered when calculating the effective gas limit. This can be
     [adjusted by a factor of 1/1024] from the previous block's gas limit, up
-    until a maximum of 30 million gas.
+    until a maximum of 120 million gas.
```
SamWilsn: Stupid question: where does this number come from?

SamueleA: This is number from here: https://github.com/ethereum/execution-specs/blob/forks/amsterdam/packages/testing/src/execution_testing/test_types/block_types.py#L27

```
CURRENT_MAINNET_BLOCK_GAS_LIMIT = 60_000_000
DEFAULT_BLOCK_GAS_LIMIT = CURRENT_MAINNET_BLOCK_GAS_LIMIT * 2
```

`CURRENT_MAINNET_BLOCK_GAS_LIMIT` is the target gas limit of 60 million from [EIP-7935](https://eips.ethereum.org/EIPS/eip-7935), which is what the network will attempt to be at. But the hard cap limit that a block can have is 2 times the target limit which is 120 million.

The doubling of the target limit comes from [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) as the `ELASTICITY_MULTIPLIER`

packages/testing/src/execution_testing/exceptions/exceptions/transaction.py
```diff
@@ -175,7 +175,7 @@ class TransactionException(ExceptionBase):
     """Transaction causes block to go over blob gas limit."""
     GAS_LIMIT_EXCEEDS_MAXIMUM = auto()
     """
-    Transaction gas limit exceeds the maximum allowed limit of 30 million.
+    Transaction gas limit exceeds the maximum allowed limit of 16,777,216.
```
marioevz: One could argue that this comment is now incorrect for older forks.

What about:

```suggestion
    Transaction gas limit exceeds the maximum allowed limit for the given fork.
```

SamueleA: This is a good point! I have included this change in the PR!

src/ethereum/forks/amsterdam/blocks.py
SamueleA: Update on the above. I have spoken with Mario Havel from the Ethereum Protocol Study Group and he explained that the value from the code block above is just used for testing and is not a spec constant, so I made an oopsie by thinking the block target was 60 million and the block limit was 120 million.

The correct information is that the block limit is 60 million and thus the block target is 30 million.

I have updated the PR with the correct number!

SamWilsn: If we look at `check_gas_limit`:

https://github.com/ethereum/execution-specs/blob/03d69439661cdf67ddae83f919a5cbdc6501ba33/src/ethereum/forks/amsterdam/fork.py#L1068-L1105

I don't see where the 60 million comes from either. Is there somewhere else that the max theoretical block gas limit is capped?

SamueleA: The max gas limit is the `gas_limit` itself. The current 60 million proposed by EIP-7935 is not a hardcoded variable in the protocol, but it's a preference set in the validators.

```
A final important note is that blocks themselves are bounded in size. Each block has a target size of 30 million gas but the size of blocks will increase or decrease in accordance with network demands, up until the block limit of 60 million gas (2x target block size). The block gas limit can be adjusted upwards or downwards by a factor of 1/1024 from the previous block's gas limit. As a result, validators can change the block gas limit through consensus.
```
https://ethereum.org/developers/docs/blocks/

The block's `gas_limit` is voted by the validators by either keeping it the same or by changing it by a factor of 1/1024 between each block. The code for `check_gas_limit` above validates that the change between blocks is not beyond the 1/1024 allowed.

This can be seen in the etherscan block explorer which shows "jitter" of the gas limit value due to validators - probably running older software with a lower gas_limit preference - voting for the gas_limit to go down (and then quickly voted up to 60M by the majority).




Based on the above, I have updated the comments for the gas_limit to be clearer and to remove references to exact numbers (that aren't determined by the execution-specs, will change over time and aren't absolute). I also changed the wording since the previous version seemed to implied that the gas fee was used to calculate an effective gas limit, but that seems wrong since the gas fee is modified to push block utilization towards 50% of the gas limit.

felix314159:
Pls rebase so we can get this merged, also u can update your description to reflect the follow-up edits

SamueleA:
> Pls rebase so we can get this merged, also u can update your description to reflect the follow-up edits

I have rebased the PR and updated the description!

SamWilsn (approved):

PR #2305, opened 2026-02-24
fix: return False instead of raising AssertionError in FixedSizeBytes…

### LINKED ISSUE #2290: Change EEST's `Address` equality to account for typing
> Sadly our `Address` equality implementation automatically tries to transform the other object to address by doing, `Address(other)`, and when `other is None`, or some other type, it raises.
> 
> _Originally posted by @marioevz in https://github.com/ethereum/execution-specs/pull/2115#discussion_r2819613121_
            

`__eq__` and friends should return `False` or [`NotImplemented`](https://docs.python.org/3/library/constants.html#NotImplemented) instead of raising when the right hand side cannot be converted.

mandeepmourya007:
## 🗒️ Description

The FixedSizeBytes.__eq__ method was using assert to check type compatibility, which raises AssertionError when comparing with unsupported types like None or dict. This replaces the assert with a conditional that returns False for unsupported types, following Python's equality semantics.
## 🔗 Related Issues or PRs
Fixes #2290

## ✅ Checklist

- [x] All: Ran fast `tox` checks to avoid unnecessary CI fails
- [x] All: PR title adheres to the repo standard

#### Cute Animal Picture

Cute cat

### DIFF
--- a/packages/testing/src/execution_testing/base_types/base_types.py
+++ b/packages/testing/src/execution_testing/base_types/base_types.py
@@ -375,12 +375,13 @@ class FixedSizeBytes(Bytes):
         if other is None:
             return False
         if not isinstance(other, FixedSizeBytes):
-            assert (
+            if not (
                 isinstance(other, str)
                 or isinstance(other, int)
                 or isinstance(other, bytes)
                 or isinstance(other, SupportsBytes)
-            )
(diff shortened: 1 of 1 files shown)


marioevz (approved):
LGTM, thanks!

Issue #2183, opened 2026-02-10
zkEVM Projects Tracking Issue


marioevz:
## Project 1: State tracker

This project enables the execution specs to output an object that lets us know what state has been touched during block processing.

#### - Why

Creating the external API allows other components to start being built in parallel, since the internals are not important for those components (BALs and stateless)

#### - Why

Specifying how this object is built allows different clients to be used fungibly. Currently, we are in a situation where clients implement it differently, see [here](https://github.com/paradigmxyz/reth/issues/20365#issuecomment-3664196323) for example.

#### - Why

Having an early spec release allows ELs to start following the specification and raising possible discrepancies.

## Project 2: Stateless execution

Once the ExecutionWitness can be created, we should have enough information to validate blocks statelessly(without a database).

This project enables the execution specs to test stateless execution and add more test vectors solely for the stateless execution.

#### - Why

This doesn't change the behaviour of the specs, it solely prepares it for stateless execution by removing codepaths that were dependent on there being a database.

#### - Why

As Mario noted in our call, this closes the feedback loop from project 1, where the execution specs is also using the ExecutionWitness that it created to validate blocks.

#### - Why

This further closes the feedback loop by finding issues/discrepancies between the "normal" execution path and the stateless one. We expect existing clients will have a similar setup.

#### - Why

On top of being able to cross check between stateful and stateless, once zkEVM is enabled and we are relying on it, this would be the default way to do testing.

#### - Why

Currently the test vectors in the execution specs are geared towards stateful execution. There would be no adversarial test vectors concerning an invalid/malformed ExecutionWitness

kevaundray:
Added skeleton for Milestone 2 [here](https://github.com/ethereum/execution-specs/pull/2213)

Left in TODOs since we need the pre-state code and the ExecutionWitness generation code to close the feedback look

frisitano:
Have you considered using journaled state for the state tracker? On the surface it appears to fulfil your requirements. https://github.com/ethereum/execution-specs/blob/cdf9add28997a11367e969a92582a1898e3c555b/src/ethereum_optimized/state_db.py#L85
This is a first-class citizen in the specs and has well-defined behaviour in all execution clients. I have successfully built witnesses from journaled state derivative types (Journal->EvmState->Witness) in the past.

[revm journal](https://github.com/bluealloy/revm/blob/f4a6fb39f6a061cebcabfc99d71be18d130b0921/crates/context/src/journal.rs#L28-L43
):
```rust
[...]

PR #2307, opened 2026-02-24
feat(specs): Partially decouple `state_transition` from `BlockChain`

kevaundray:
## 🗒️ Description

**Goal**

The goal here is to decouple `BlockChain` from `state_transition` to make it easier for anyone with a given `PreState` to execute a block.

**What the PR does**

This PR creates a new method called `execute_block` which is effectively the code in `state_transition` that does not modify the `Blockchain` object.

The rough outline of what happens is:

```rust
fn state_transition(chain: BlockChain, block: Block) {
    // Convert `BlockChain` into `PreState` and `ChainContext`
    // This part will be different for stateful vs stateless.
   let pre_state = prestate_from_chain(chain);
   let chain_context = context_from(chain);
   
   // Call execute_block. This will be the same whether we are stateful or stateless.
   // No db changes are made here.
   let (account_changes, storage_changes, code_changes) = execute_block(block, pre_state, chain_context);
   
   // Apply all of the changes from block execution to the `BlockChain` object.
   // This is also specific to stateful, since stateless has no database.
   apply_state_changes(chain, account_changes, storage_changes, code_changes);
}
```

**Notable changes**

- `validate_header` previously was taking a `BlockChain` object in order to get the parent header. It now just takes the parent header as a parameter. This allows us to put it in `execute_block`
- `ChainContext` was introduced to group together the chain related items that don't fit into the `PreState` abstraction but are needed for block execution. This is the chain id, the last 256 block hashes and the parent header (for `validate_header`)

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -28,6 +28,7 @@ from ethereum.exceptions import (
     InvalidSenderError,
     NonceMismatchError,
 )
+from ethereum.forks.bpo5.blocks import Header as PreviousHeader
 from ethereum.state import EMPTY_CODE_HASH, Address, BlockDiff, PreState
 
 from . import vm
@@ -144,7 +145,7 @@ class ChainContext:
     block_hashes: List[Hash32]
     """Recent ancestor hashes (up to 256) for the ``BLOCKHASH`` opcode."""
 
-    parent_header: Header
+    parent_header: Header | PreviousHeader
     """Parent header used for header validation and system contracts."""
 
 
@@ -418,7 +419,9 @@ def calculate_base_fee_per_gas(
     return Uint(expected_base_fee_per_gas)
 
 
-def validate_header(parent_header: Header, header: Header) -> None:
+def validate_header(
+    parent_header: Header | PreviousHeader, header: Header
+) -> None:
     """
     Verify a block header against its parent.
 
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -134,11 +134,18 @@ BLOB_COUNT_LIMIT = 6
 @slotted_freezable
 @dataclass
 class ChainContext:
-    """Chain context needed for block execution."""
+    """
+    Chain context needed for block execution.
+    """
 
     chain_id: U64
+    """Identify the chain for transaction signature recovery."""
+
     block_hashes: List[Hash32]
+    """Recent ancestor hashes (up to 256) for the ``BLOCKHASH`` opcode."""
+
     parent_header: Header
+    """Parent header used for header validation and system contracts."""
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


kevaundray:
Related to project 2, milestone 1 in #2183

kevaundray:
Commit [e801c45](https://github.com/ethereum/execution-specs/pull/2307/commits/e801c452908bbce81538053691d0fcdb2f72941e) had tests_pytest_py3 failing, this was unrelated to the changes made in this PR.

Failiing CI job: https://github.com/ethereum/execution-specs/actions/runs/22374230637/job/64760791347

It is now passing after I merged master into the branch, however I wonder if it was a flaky test.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -130,6 +131,23 @@
...
+    Chain context needed for block execution.
+    """
+
+    chain_id: U64
+    """Identify the chain for transaction signature recovery."""
+
+    block_hashes: List[Hash32]
+    """Recent ancestor hashes (up to 256) for the ``BLOCKHASH`` opcode."""
+
+    parent_header: Header
```
SamWilsn: This'll need to be `Header | PreviousHeader` I think?

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -227,21 +245,76 @@ def state_transition(chain: BlockChain, block: Block) -> None:
...
+def execute_block(
+    block: Block,
+    pre_state: PreState,
+    chain_context: ChainContext,
+) -> Tuple[
+    Dict[Address, Optional[Account]],
+    Dict[Address, Dict[Bytes32, U256]],
+    Dict[Hash32, Bytes],
+]:
+    """
```
SamWilsn: Should we use an object with named fields?

SamWilsn (commented):
So far so good!

src/ethereum/forks/amsterdam/fork.py
kevaundray: Ah yes because the transition block will have a header from the previous fork

kevaundray: Ah this is a thing on the default branch, perhaps we should have an issue to change it in all of the places we need? see https://github.com/ethereum/execution-specs/blob/561ce7464ed757f71abb23a6028a7b43187b2e3f/src/ethereum/forks/amsterdam/vm/gas.py#L308 for an example

src/ethereum/forks/amsterdam/fork.py
kevaundray: Added a  `BlockDiff` object

src/ethereum/forks/amsterdam/fork.py
kevaundray: Ah actually I encountered it here: https://github.com/ethereum/execution-specs/pull/2362#issuecomment-3977827445

I would have expected it to fail on the default branch, but I think its likely okay as long as the fields being accessed across the forks are okay

gurukamath (approved):

PR #2309, opened 2026-02-25
chore(ci): Investigate possible flaky test

kevaundray:
## 🗒️ Description

See comment here: https://github.com/ethereum/execution-specs/pull/2307#issuecomment-3955373758

Expectedly, I haven't been able to reproduce this locally, but looking at the code, the only thing that is the same between cancun and prage is the blob filename, since it is `blob_0_cell_proofs_0.json`. 

If there is a race condition, then they would both use the same file. However, the file has the fork name in it, so if the Cancun run reads the Prague file because it was not deleted for some reason, then we get the error seen in the CI, since the file contents include the fork name.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/test_types/blob_types.py
+++ b/packages/testing/src/execution_testing/test_types/blob_types.py
@@ -86,6 +86,8 @@ class Blob(CamelModel):
         )
         return (
             "blob_"
+            + fork.name()
+            + "_"
             + str(seed)
             + "_cell_proofs_"
             + str(amount_cell_proofs)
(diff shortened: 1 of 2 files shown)


felix314159:
I wrote the code involved here and originally wanted to add the fork name to the blob filename, but there was [pushback](https://github.com/ethereum/execution-spec-tests/pull/1614#discussion_r2109805938). The lack of fork in the filename is also the direct reason for the [weird unit tests](https://github.com/ethereum/execution-specs/blob/e275279797a713872c3564886e9f604164a23549/packages/testing/src/execution_testing/test_types/tests/test_blob_types.py#L159-L161) where we use counters to ensure a specific test order that we need cuz otherwise another unit test might replace the fork within a blob files which would be undetected cuz the filename is still the same but which does occassionally lead to unit test failures. I am all for adding the fork name to the blob file name, even if it means being able to cache less. @marioevz wdyt

kevaundray:
No strong opinion from my side :)

felix314159:
if this starts making problems we can look into this again

PR #2310, opened 2026-02-25
chore(ci): add setup-uv action, bump default ci python to 3.14

### LINKED ISSUE #2315: chore(ci): replace actions/setup-python with setup-uv across all workflows
## Problem

`test.yaml` and `gh-pages.yaml` use `actions/setup-python` with a **hardcoded `python-version: "3.11"`**, while the repo defines GitHub Actions variables `DEFAULT_PYTHON_VERSION` (currently `3.12`) and `UV_VERSION` (currently `0.7.2`) that are intended to centralize version management.

Several other workflows (`test-checklist.yaml`, `test-docs.yaml`, `benchmark.yaml`, fixture release workflows) already use `setup-uv` with these variables correctly. But the two highest-traffic workflows still ignore them.

### Affected workflows

| Workflow | Jobs using `setup-python` | Python version |
|---|---|---|
| `test.yaml` | `static`, `py3`, `pypy3`, `json_infra`, `tests_pytest_py3`, `tests_pytest_pypy3` | Hardcoded `3.11` / `pypy3.11` |
| `gh-pages.yaml` | `build` | Hardcoded `3.11` |

### Why `setup-uv` doesn't override `setup-python`

In the `static` job, `setup-uv` runs **after** `setup-python` but without a `python-version` parameter, so it only installs the `uv` binary — it doesn't touch the Python installation. The Python 3.11 from `setup-python` remains the active interpreter.

### Additional inconsistency

These workflows also install tox via legacy `pip` and invoke it as `tox -e ...`, while the corrected workflows use `uvx tox -e ...` (no separate install step needed).

The `setup-env` composite action (`.github/actions/setup-env/action.yaml`) similarly uses legacy pip to install `tox` and `requests`.

## Proposed fix

1. **Replace `actions/setup-python`** with `astral-sh/setup-uv` in all jobs of `test.yaml` and `gh-pages.yaml`:
   - CPython jobs: `python-version: ${{ vars.DEFAULT_PYTHON_VERSION }}`
   - PyPy jobs: `python-version: "pypy3.11"`
   - All jobs: `version: ${{ vars.UV_VERSION }}`

2. **Switch tox invocations** from legacy pip-installed tox to `uvx tox -e ...` (consistent with other workflows).

3. **Update `.github/actions/setup-env/action.yaml`**: remove tox install (now handled by `uvx`), keep `requests` install for `download_geth_linux.py` (via `uv pip install --system requests`).

4. **Remove the separate `setup-uv` step** in the `static` job (now merged into the Python setup step).

After this, `actions/setup-python` is no longer used anywhere and can be dropped entirely. Python version management is fully centralized in `DEFAULT_PYTHON_VERSION`.

kevaundray:
## 🗒️ Description

- Updates the default Python version to 3.14 to improve performance, in particular for py3 which generates coverage:
    - Was looking at the overhead of using coverage and noticed that sysmon is now the default for python 3.14+: https://coverage.readthedocs.io/en/latest/changes.html#version-7-9-1-2025-06-13.
- Create `.github/actions/setup-uv/` composite action that centralizes the `astral-sh/setup-uv` SHA pin, installs Python via uv, and makes `tox` available as a plain command (`uv tool install tox`).
- Hardcodes defaults directly in the new action, eliminating the need for `vars.DEFAULT_PYTHON_VERSION` and `vars.UV_VERSION` repo settings.
- Adopt the new action across all applicable workflows, replacing repeated `astral-sh/setup-uv` boilerplate and `uvx tox` invocations with plain `tox`.
- Fix: Removes use of `setup-python` which was actually taking effect, although `astral-sh/setup-uv` was the intended action to install Python.
- Fix: Remove bare `pip install` from `setup-env` and instead adds the dependencies to `scripts/download_geth_linux.py` as PEP 723/uv-style inline deps.
- Bumps `coverage` package version used from `7.11.0` to `7.13.4` to get cp314 wheels (they're natively compiled extensions for Python 3.14).

### Not changed

`eip-rebase.yaml` and `update-devnet-branch.yaml` only need bare `uv` (no Python install, no tox), so they keep their direct `astral-sh/setup-uv` reference.

### Follow-up

At some point, after merge and all branches have been rebased to forks/amsterdam, the `DEFAULT_PYTHON_VERSION` and `UV_VERSION` repository variables can be deleted from repo settings — they are no longer referenced by any workflow using the composite action. Note: allow time for the change to propagate before removing them.

## 🔗 Related Issues or PRs
Closes #2315.
- #2315 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

:cat:


.github/workflows/test.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065
         with:
-          python-version: "3.11"
+          python-version: "3.14"
```
danceratopz: I was surprised that this worked, lol! And then surprised that we use `actions/setup-python` at all, as Python version management _should_ be managed by `astral-sh/setup-uv` in L53/L60. 

This is a pre-existing bug and would suggest we remove all use of `actions/setup-python` and `astral-sh/setup-uv` as intended, create an issue for this:
- https://github.com/ethereum/execution-specs/issues/2315

danceratopz (approved):
Really nice! Very down to bump these to 3.14 for this speed-up!

As it stands, I think the py3 job is our main verification that `ethereum-execution` stays compatible with 3.11? Either way, we should be careful that we don't drop this support, so I added #2316. 

In #2316, we might have to revert back to 3.11 for `static` (in order to allow `ruff` to check with `target-version = "py311"`), but I don't think that's a biggy.

Tbh, I'd say that we get this merged for the immediate benefit and then solve the following issues in follow-up PRs:
- #2315 
- https://github.com/ethereum/execution-specs/issues/2316

.github/workflows/test.yaml
```diff
@@ -114,7 +114,7 @@ jobs:
...
```
danceratopz: I'd keep this one as 3.11 in order to maintain a backwards compatibility check of ethereum-execution with 3.11 (although I don't like that it's hidden away here, we can try to improve this in #2316).
```suggestion
          python-version: "3.11"
```

danceratopz (changes requested):
As this PR doesn't bump uniformly to 3.14 to all CI steps jobs that could benefit from it and that's best handled by #2315, I'll bolt on that fix/improvement here, too.

kevaundray:
> As this PR doesn't bump uniformly to 3.14 to all CI steps jobs that could benefit from it and that's best handled by #2315, I'll bolt on that fix/improvement here, too.



> As this PR doesn't bump uniformly to 3.14 to all CI steps jobs that could benefit from it and that's best handled by #2315, I'll bolt on that fix/improvement here, too.

Makes perfect sense to me 🫡

danceratopz (approved):
`py3` at 19 minutes! And that's without fixing #2314.

The coverage warnings pre-date this PR and pytedt-cov version bump (for example https://github.com/ethereum/execution-specs/actions/runs/22300356643): 
```
/opt/actions-runner/_work/execution-specs/execution-specs/.tox/py3/lib/python3.11/site-packages/coverage/inorout.py:537: CoverageWarning: Module ethereum was previously imported, but not measured (module-not-measured); see https://coverage.readthedocs.io/en/7.11.0/messages.html#warning-module-not-measured
```

Carsons-Eels:
I'm generally positive about this PR, but a little hesitant about bumping the version like that, though I think keeping json_infra pinned at 3.11 helps that a bit.

I think before we bump the version to 3.14 we need to first verify if the py-ecc, ethereum-types, and ethereum-rlp libraries will function properly. If I recall correctly, py-ecc could be a sticking point.

danceratopz:
Thanks for the feedback @Carsons-Eels!

> I'm generally positive about this PR, but a little hesitant about bumping the version like that, though I think keeping json_infra pinned at 3.11 helps that a bit.

I understand, I think this depends on which versions of Python we'd like to support in `ethereum-execution` (I think we can do as we please with `./packages/testing/` and `./tests/`, so are currently only version bound by `ethereum-execution`). As Sam pointed out to me elsewhere, we don't make any promises about that. But here's the stats for downloads since ~August:
image

I'm not in any rush to bump the min required version and plan to test that 3.11 support is maintained (for now) in:
- #2316. 

> I think before we bump the version to 3.14 we need to first verify if the py-ecc, ethereum-types, and ethereum-rlp libraries will function properly. If I recall correctly, py-ecc could be a sticking point.

I took a look at this and all of these deps are pure Python packages. As far as I can tell, there are no removed stdlib modules or syntax changes that would affect these libraries.

I think a good check would be to do run `hasher` on the fixture output of `py3`?

spencer-tb (approved):
LGTM! Shit ip (I wrote this first lol but I meant ship it)!

Only catch is the uv.lock file.

danceratopz:
> LGTM! Shit ip (I wrote this first lol but I meant ship it)!

<3

> Only catch is the uv.lock file.

Oh yeah, thanks! That does indeed look like an orthogonal change in the context of this PR now! But it's relevant for the 3.14 bump.

Updated description: 
> - Bumps `coverage` package version used from `7.11.0` to `7.13.4` to get cp314 wheels (they're natively compiled extensions for Python 3.14).

We fixed up some other 3.14 packages here:
- https://github.com/ethereum/execution-specs/pull/2283

danceratopz:
py3 at ~17 minutes with #2332.

PR #2321, opened 2026-02-25
feat(tests): port all static tests to python

spencer-tb:
### Motivation
EIP-8037 requires complex logic changes for the static tests that is not possible within yaml fillers. Over 703 files contained errors on filling. Most of these fixes were gas increases. Some were not fixable due to the former. 

For the full version of 8037 we need the cost per state byte to be dependent on the block gas limit. This means having 2 ways of filling, static/non-static both needed an update in the framework.

Initially this just started as "lets port only the failing tests" but to me it just made sense to continue with all them given it was one script.

Note the first commit, is where we have hasher parity. That is the fixture output is indentical when filling from the python tests (ported) to the static yaml fillers with `--fill-static-tests`. With the parity the next step was to combine some files that made sense. 300+ tests with similar names and varying parameters did not need there own file. Finally the static filler components in the framework were removed. We no longer need them. This should hopefully speed up filling these tests too and the lllc is now pre-compiled into our opcode lang. LLLC is removed as a dependency for static tests now. 

### Remaining Todos
- Fix CI, mypy/ruff/codespell -> ruff overflows currently lol.
- Update the tests to use the dynamic account allocation instead of hardcoded addresses (required for hash matching initially). This is required for no duplicate accounts in the pre alloc.
- Replace the yul with python op code lang. Claude code can one shot this. Remove the yul code and remove the solc dependency.
- Update the docs to remove the static filler content.
- Confirm we don't have duplicates from PRs already ported in the past week.
- Move the tests to there appropriate fork folder (when the EIP was introduced). Still only fill them from Cancun.
- Add the correct ported from markers to all tests ported.

### 1st commit (hasher verify and contains script)
Add scripts/fixture_to_python.py that converts compiled static test
fixtures (from --fill-static-tests) to Python test files. The script:

- Reads compiled fixtures for hex values and source fillers for context
- Generates Op-language bytecode (readable, round-trip verified)
- Parses filler network fields for fork bounds (>=Cancun<Osaka)
- Emits valid_until markers when fillers exclude later forks
- Handles both range syntax and exact fork name lists

All 2411 generated tests pass (--until Osaka) and hash-match the
compiled static fillers (39,786 matching fixture hashes).

### 2nd commit
Remove all YAML/JSON filler files from tests/static/state_tests/ and
replace with auto-ported Python test files (moved from tests/ported/).
Merge related test files by common prefix to reduce file count
(2395 -> 2064 files). Remove converter and merge scripts.

All 40169 test cases pass and fixture hashes match the original
compiled static fillers (39,786 matching hashes).

### 3rd commit
Remove the static test filler framework now that all YAML/JSON fillers
have been replaced with Python tests:

- Delete static_filler.py plugin, base_static.py, static_state/ specs
- Remove --fill-static-tests CLI option and FixtureCollector.fill_static_tests
- Remove modify_static_test_gas_limits CLI tool
- Remove static_filler plugin registration from pytest-fill.ini
- Remove --fill-static-tests from CI configs (feature.yaml)

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


marioevz:
I looked at the produced tests and they look great in general. A few outstanding issues I could spot:

### `post` is not being populated

This is a breaking change, so we should resolve it before proceeding IMO, because even if the tests produce the exact same fixtures at the moment, if we plan on enhancing these tests, the post is required so we don't break the tests.

If case matching is the issue, we could perhaps add it to the parametrization, so in place of doing:

```python
@pytest.mark.parametrize(
    "tx_data_hex",
    [
        "264bb86a0000000000000000000000000000000000000000000000000000000000000006",
        "5114e2c8000000000000000000000000000000000000000000000000000000000000000a",
        ...,
    ],
    ids=['case0', 'case1', ...],
)
```
We could do:
```python
@pytest.mark.parametrize(
    "tx_data_hex,post",
    [
        pytest.param(
            "264bb86a0000000000000000000000000000000000000000000000000000000000000006",
            {"0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba": {"balance": 1}},
            id="case0",
        ),
        pytest.param(
            "5114e2c8000000000000000000000000000000000000000000000000000000000000000a",
            {"0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba": {"balance": 2}},
            id="case1",
        ),
    ],
)
```

### `evm_bytes` could be improved

Currently generated code is like this:
```python
pre[contract] = Account(
        balance=0xde0b6b3a7640000,
        nonce=0,
        code=(
        Op.PUSH1[0x40] + Op.PUSH1[0x0] + Op.PUSH1[0x40] + Op.PUSH1[0x0]
        + Op.PUSH1[0x0] + Op.PUSH20[0xd0735f094c16e509e8d76999d9ee2e4fd5166c2e]
        + Op.PUSH2[0x1770] + Op.CALL + Op.STOP
    ),
    )
```
But it could somewhat easily be improved to something like:
```python
pre[contract] = Account(
        balance=0xDE0B6B3A7640000,
        nonce=0,
        code=(
            Op.CALL(
                gas=Op.PUSH2[0x1770],
                address=Op.PUSH20[0xD0735F094C16E509E8D76999D9EE2E4FD5166C2E],
                value=Op.PUSH1[0x0],
                args_offset=Op.PUSH1[0x0],
                args_size=Op.PUSH1[0x40],
                ret_offset=Op.PUSH1[0x0],
                ret_size=Op.PUSH1[0x40],
            )
            + Op.STOP
        ),
    )
```

### `sender` as EOA
We currently do:
```python
sender = Address("0xebaf50debf10e08302fe4280c32df010463ca297")
...
pre[sender] = Account(balance=0xDE0B6B3A7640000, nonce=0)
...
tx = Transaction(
        secret_key=Hash(
            "0xe04d1ac7ddda0c98397d56a0b501e960d4cd325a39286919ac23c1a07009a869"
        ),
...
```
but one easy modification which could simplify later analysis of the test is:
```python
sender = EOA(key="0xe04d1ac7ddda0c98397d56a0b501e960d4cd325a39286919ac23c1a07009a869")
...
pre[sender] = Account(balance=0xDE0B6B3A7640000, nonce=0)
...
tx = Transaction(
        sender=sender,
...
```

spencer-tb:
Really nice catches! Post state especially. Will work on this.

leolara:
I am sorry if I am very wrong, because I am very new to this but I have a few ideas.

The script could create a map to `address -> variable name` and create the addresses with pre.fund_.. pre.deploy.. ?

I saw the ai agent command says to don't use pre[...] for stuff 

**https://github.com/ethereum/execution-specs/blob/forks/amsterdam/.claude/commands/write-test.md?plain=1#L14-L16**

chfast:
Doesn't it make sense to only covert the tests which are failing in the new fork?

spencer-tb:
> Doesn't it make sense to only covert the tests which are failing in the new fork?

That was the initial idea but to me it seems reasonable to do all of them. Since we have fixture matches with the hasher it means the fixture generation is exactly the same. The python is more readable in my opinion (for humans & LLMs).

We additionally get to remove the static tests filling code, which makes the framework more robust and concentrated, as well as LLLC/Yul requirements in our CI & release flow.

The main benefit for me is being able to add fill the tests for enginex format. This allows us to use all these tests within `consume enginex` which will give us an insane speed up when running the tests in hive locally and in CI.

Once this is completed, we should individually port each test in the long term, to a) see if we can add extra parameterization or more coverage and b) remove any duplication with pre-existing cases.

marioevz:
From our meeting, cc @leolara:

- Rebase the PR on top of the latest Amsterdam/static-tests changes so no recently added or cleaned-up tests are lost 
- Ensure conversion script is rerun with the updated EVM bytes converter so generated Python bytecode sections are more readable  
- Extend the script to also port post-state expectations from selectors into the Python tests, so postconditions are actually verified  
- Update the script to embed the original Solidity/Yul source as comments alongside the generated Python test code   
- After the above, remove the legacy static tests and their filling, once hash matches and coverage are confirmed

leolara:
Should we close this?

PR #2323, opened 2026-02-25
fix(ci): don't cancel in-progress runs on the default branch

danceratopz:
## 🗒️ Description

When multiple PRs get merged to the default branch in quick succession, `cancel-in-progress: true` causes earlier CI runs to be cancelled before completing. This means some merge commits never get a full CI run.

This is annoying as the majority of runs get cancelled. This strongly devalues this CI step:
- It's noisy, I never check fails.
- Binary search to detect a genuine fail doesn't work; there is incomplete information.
- Benchmark artifacts only get built if certain paths change; if a benchmark flow gets triggered, it can get killed by a merge commit to a non-benchmark PR, the new actions generated by this PR will not trigger the benchmark flow, so no benchmark artifact will get generated.

This will create more load on our runners, but it's in the right place imo.

What pushed me other the edge: 
- I manually re-triggered the Benchmark flow to check whether the test fail in the previous run was flaky.
- I merged PR while waiting; manually triggered flow got cancelled.
- I got frustrated.

---

This PR makes `cancel-in-progress` conditional: it remains `true` for PR and feature branch runs, but is set to `false` when the ref matches the default branch. Uses `github.event.repository.default_branch` so the expression adapts automatically if the default branch changes.

Also adds a `github.run_id` fallback to the benchmark workflow's concurrency group for robustness (consistent with the other workflows).

---

image

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

#### Cute Animal Picture

cute animal

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -23,7 +23,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
-  cancel-in-progress: ${{ github.ref != format('refs/heads/{0}', github.event.repository.default_branch) }}
+  cancel-in-progress: ${{ github.ref_name != github.event.repository.default_branch }}
 
 jobs:
   unit-tests:
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -22,8 +22,8 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: ${{ github.ref != format('refs/heads/{0}', github.event.repository.default_branch) }}
 
 jobs:
   unit-tests:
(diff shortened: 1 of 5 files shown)


SamWilsn (commented):
We use `github.ref_name == github.event.repository.default_branch` elsewhere. For consistency, should we use the same check everywhere we need the default branch, and if so, which is the "correct" way to write the condition?

danceratopz:
> We use `github.ref_name == github.event.repository.default_branch` elsewhere. For consistency, should we use the same check everywhere we need the default branch, and if so, which is the "correct" way to write the condition?

Thanks for spotting this @SamWilsn, the pre-existing `github.ref_name` is much cleaner. 

For other people stumbling on this, this PR introduced this check to detect the default branch:
```yaml
github.ref != format('refs/heads/{0}', github.event.repository.default_branch)
```
which clashed with the pre-existing style:
```yaml
github.ref_name == github.event.repository.default_branch
```

As you might expect:
- `github_ref_name`: is the short (branch) name `forks/amsterdam`. 
- `github.ref` is the full ref: `refs/heads/forks/amsterdam`.

I'll update the PR to `ref_name`.

spencer-tb (approved):
LGTM! Shitip!

PR #2325, opened 2026-02-25
fix: skip storage verification for empty storage dict

### LINKED ISSUE #2162: Account(storage={}) infinite loop-checks account storage over RPC
### What was wrong?

https://github.com/ethereum/execution-specs/blob/9959739d5cbfd90c1bd531e1be3510d373c3905e/packages/testing/src/execution_testing/execution/transaction_post.py#L197-L205

### How can it be fixed?

The tests written in the tooling often contain checks for storage which inspect specific fields. The other fields, by default, can have any value, and allows to write focused tests where it is clear what is actually being tested. This raises the question what `Account(storage={})` is supposed to be. If one key/value is added then only that key/value is checked to contain the correct value, the other storage keys do not matter. Therefore I think that `Account(storage={})` is therefore meaningless as it allows all key/values in the storage. If a test writer explicitly wants to test that the storage is empty, then I believe a specific flag for that should be added (just like `Account.NONEXISTENT` -> `Storage.EMPTY` for instance).

To test this over RPC (via `execute remote`) it is not feasible to verify all keys are empty. It is also not possible to directly get the state root of an account. However, via `eth_getProof` we can get the relevant fields. For empty storage we can check if the root of the account is indeed the empty storage root. Note that `getProof` is not available for all RPCs and the format also depends on the trie type, so it is not super steady, but I believe this is the only way to check for empty storage over the `eth_` RPC.

### Additional Context

Currently the empty storage account will fetch all storage keys (2^256 keys) over RPC to verify if they are empty, i.e. the test will never pass as this check stalls the process.

mandeepmourya007:
Closes #2162

Account(storage={}) caused infinite loop when verifying storage over RPC (attempting to check 2^256 keys). Skip storage check when no keys are specified - empty storage dict is now treated as 'allow any storage values'.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/base_types/composite_types.py
+++ b/packages/testing/src/execution_testing/base_types/composite_types.py
@@ -66,6 +66,13 @@ class Storage(
     Dictionary type to be used when defining an input to initialize a storage.
     """
 
+    EMPTY: ClassVar[None] = None
+    """
+    Sentinel object used to specify that the storage must be completely empty.
+    When used in a test's post state, this signals that the account's storage
+    should be verified as empty (e.g., via eth_getProof checking the storage root).
+    """
+
     @dataclass(kw_only=True)
     class InvalidTypeError(Exception):
         """
@@ -360,8 +367,11 @@ class Account(CamelModel):
     """The amount of Wei (10<sup>-18</sup> Eth) the account has."""
     code: Bytes = Bytes(b"")
     """Bytecode contained by the account."""
-    storage: Storage = Field(default_factory=Storage)
-    """Storage within a contract."""
+    storage: Storage | None = Field(default_factory=Storage)
+    """
+    Storage within a contract.
+    Use `Storage.EMPTY` (None) to explicitly verify that storage is completely empty.
+    """
 
     NONEXISTENT: ClassVar[None] = None
     """
@@ -515,7 +525,19 @@ class Account(CamelModel):
                 )
 
         if "storage" in self.model_fields_set:
-            self.storage.must_be_equal(address=address, other=account.storage)
+            if self.storage is None:
+                if account.storage is not None and account.storage:
+                    for key, value in account.storage.items():
+                        if value != 0:
+                            raise Storage.KeyValueMismatchError(
+                                address=address,
+                                key=key,
+                                want=0,
+                                got=value,
+                                hint="Storage.EMPTY expects completely empty storage",
+                            )
+            elif self.storage.root:
+                self.storage.must_be_equal(address=address, other=account.storage)
 
     def __bool__(self: "Account") -> bool:
         """Return True on a non-empty account."""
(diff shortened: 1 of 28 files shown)


marioevz (changes requested):
This change is modifying the behavior that a lot of tests expect.

We need to:
1) Define a new sentinel value `Storage.EMPTY` (perhaps simply equal to `None`) that is used to signal that the test expects the storage to be completely empty.
2) Go through all tests that are currently doing `Account(storage={})` and use the sentinel value instead, to make sure that the checks remain the same for these tests
3) Modify `TransactionPost` to handle the sentinel value in the way that the original issue describes (Using `eth_getProof` for example).

mandeepmourya007:
> This change is modifying the behavior that a lot of tests expect.
> 
> We need to:
> 
> 1. Define a new sentinel value `Storage.EMPTY` (perhaps simply equal to `None`) that is used to signal that the test expects the storage to be completely empty.
> 2. Go through all tests that are currently doing `Account(storage={})` and use the sentinel value instead, to make sure that the checks remain the same for these tests
> 3. Modify `TransactionPost` to handle the sentinel value in the way that the original issue describes (Using `eth_getProof` for example).

@marioevz please check i update PR but not sure why my old commit lost

marioevz:
> > This change is modifying the behavior that a lot of tests expect.
> > We need to:
> > 
> > 1. Define a new sentinel value `Storage.EMPTY` (perhaps simply equal to `None`) that is used to signal that the test expects the storage to be completely empty.
> > 2. Go through all tests that are currently doing `Account(storage={})` and use the sentinel value instead, to make sure that the checks remain the same for these tests
> > 3. Modify `TransactionPost` to handle the sentinel value in the way that the original issue describes (Using `eth_getProof` for example).
> 
> @marioevz please check i update PR but not sure why my old commit lost

Thanks, I see the changes and looks very good so far!

One missing thing though is that we need to search for tests doing `post = Account(..., storage={}, ...)` because this PR changes the meaning of this, and most of the tests should be updated to now do `post = Account(..., storage=Storage.EMPTY, ...)`, otherwise we are going to skip checks on those tests that currently expect the storage to be empty.

mandeepmourya007:
> > > This change is modifying the behavior that a lot of tests expect.
> > > We need to:
> > > 
> > > 1. Define a new sentinel value `Storage.EMPTY` (perhaps simply equal to `None`) that is used to signal that the test expects the storage to be completely empty.
> > > 2. Go through all tests that are currently doing `Account(storage={})` and use the sentinel value instead, to make sure that the checks remain the same for these tests
> > > 3. Modify `TransactionPost` to handle the sentinel value in the way that the original issue describes (Using `eth_getProof` for example).
> > 
> > 
> > @marioevz please check i update PR but not sure why my old commit lost
> 
> Thanks, I see the changes and looks very good so far!
> 
> One missing thing though is that we need to search for tests doing `post = Account(..., storage={}, ...)` because this PR changes the meaning of this, and most of the tests should be updated to now do `post = Account(..., storage=Storage.EMPTY, ...)`, otherwise we are going to skip checks on those tests that currently expect the storage to be empty.

please have a look I updated other test cases

LouisTsai-Csie:
Thanks for the contribution. However, this PR is out of date. Could you rebase it? Otherwise I'll close it in a few days.

PR #2094, opened 2026-01-28
refactor(test): rename gas cost constants

Carsons-Eels:
## 🗒️ Description

Changes:
- Refactored our Gas constants to have clearer naming (described in table below)
- Removed unnecessary `Uint` cast for REFUND_STORAGE_CLEAR

> [!NOTE]
> Cherry picked [PR#1870](https://github.com/ethereum/execution-specs/pull/1870) onto `forks/amsterdam`
> Order matches gas_costs.py

| old | new | Value | Description |
|-----|-----|-------|-------------|
| `G_JUMPDEST` | `GAS_JUMPDEST` | 1 | Jump destination opcode |
| `G_BASE` | `GAS_BASE` | 2 | Base operation cost |
| `G_VERY_LOW` | `GAS_VERY_LOW` | 3 | Very low cost operations |
| `G_LOW` | `GAS_LOW` | 5 | Low cost operations |
| `G_MID` | `GAS_MID` | 8 | Medium cost operations |
| `G_HIGH` | `GAS_HIGH` | 10 | High cost operations |
| `G_WARM_ACCOUNT_ACCESS` | `GAS_WARM_ACCOUNT_ACCESS` | 100 | Warm account access |
| `G_COLD_ACCOUNT_ACCESS` | `GAS_COLD_ACCOUNT_ACCESS` | 2600 | Cold account access |
| `G_ACCESS_LIST_ADDRESS` | `GAS_TX_ACCESS_LIST_ADDRESS` | 2400 | Access list address cost (EIP-2930) |
| `G_ACCESS_LIST_STORAGE` | `GAS_TX_ACCESS_LIST_STORAGE_KEY` | 1900 | Access list storage key cost (EIP-2930) |
| `G_WARM_SLOAD` | `GAS_WARM_SLOAD` | 100 | Warm storage load |
| `G_COLD_SLOAD` | `GAS_COLD_SLOAD` | 2100 | Cold storage load |
| `G_STORAGE_SET` | `GAS_STORAGE_SET` | 20000 | Storage slot set cost |
| `G_STORAGE_UPDATE` | `GAS_STORAGE_UPDATE` | 5000 | Storage slot update cost |
| `G_STORAGE_RESET` | `GAS_STORAGE_RESET` | 2900 | Storage slot reset cost |
| `G_SELF_DESTRUCT` | `GAS_SELF_DESTRUCT` | 5000 | Self-destruct base cost |
| `G_CREATE` | `GAS_CREATE` | 32000 | Contract creation cost |
[...]

PR #2331, opened 2026-02-26
feat(test): add fast gas repricing and gas-map CLI tool

### LINKED ISSUE #2200: feat(all): Repricing Tool
# Gas Repricing Support in Specs + Testing Framework

## Summary

We want a minimal, robust way to experiment with gas repricing across both the specs and the testing framework, without constantly touching/dirtying spec files or adding heavy configuration plumbing.

The core idea is:

- Define all baseline gas costs as usual in specs and the testing framework (no behaviour change by default).
- Introduce an optional, local **repricing configuration file** (e.g. `gas_repricings.json`) that, when present, is used to override gas costs at **import/load time**.
- Have both:
  - the testing framework (forks in `forks.py` / testing side), and
  - the spec side (per-fork `gas.py` / equivalent)

  consult this config on import and transparently update the gas cost data structures.

- Keep the mechanism as small and localized as possible (ideally one entry point per side, wired through base classes).

This should support:

- Fast local iterations on candidate gas schedules (e.g. for 7904 / Amsterdam and beyond).
- Running both `fill` and `execute` (including `execute-remote`) with modified gas schedules.
- Eventually feeding finalized repricing back into the canonical spec constants and tests.

## Goals

1. **Fast iteration loop for repricing**
   - Change gas costs in one JSON file.
   - Re-run tests/benchmarks against the new schedule.
   - No need to hand-edit spec/test constants repeatedly.

2. **Minimal intrusion into existing code**
   - Centralize repricing logic in a base layer (e.g. base fork class on testing side; single hook on spec side).
   - Default behaviour (no config file) is exactly current behaviour.

3. **Shared mechanism across specs and tests**
   - Single config format that can be consumed by:
     - Testing framework (forks)
     - EELS spec modules
   - No duplication of repricing logic.

4. **Safe and explicit**
   - Opt-in only: repricing happens only if the config is present (and possibly guarded by an env var).
   - Emit a clear warning when repricing is active to avoid “silent” surprises.
   - If we have a default repricing JSON file, it should be added to the .gitignore

## Non-goals

- This is **not** a permanent source of truth for gas costs.
  - Final gas schedules must still be reflected back into normal spec/test constants.
- This is **not** intended to replace Pitstop or client-side tooling.
  - It’s complementary: helps us generate/validate repricing test data and spec constants.

## Prerequisites

Before wiring in the repricing tool, we need:

1. **Benchmark refactor complete** ✔️
  - All benchmark tests use the gas cost calculator, not hard-coded values.
  - (Louis: this is already done.)

2. **Unified/renamed gas constants between specs and tests** ✔️
  - Carson’s existing [PR](https://github.com/ethereum/execution-specs/pull/2094) to align naming between `forks.py` and the spec constants needs:
    - Rebase on current `main` (`forks/amsterdam`).
    - Adjustments for recent refactors (new constructs in `forks.py`, etc.).
  - Any remaining spec-side changes (e.g., replacing hard-coded numeric gas charges with references to the unified constants) should be handled in/alongside that PR.

3. Source feedback on approach from stakeholders:
   - Maria
   - Kamil
   - Jochem

Once these are done, repricing logic can be implemented on top.

### 1. Repricing Config File

- **Location**: Current working directory where `uv run fill` / `pytest` / `execute` is invoked.
- **Name** (suggested): `gas_repricings.json`
- **Git**: Add to `.gitignore` so it’s treated as ephemeral/local by default.
- **Format** (example):

  ```json
  {
    "amsterdam": {
      "DUP": 1000,
      "SSTORE": 5000
    },
    "osaka": {
      "CALL": 800
    }
  }
  ```

Details to be finalized, but high-level:

- Top-level keys: fork identifiers (e.g. `"amsterdam"`, `"osaka"`).
- Values: `{ opcode_name: new_gas_value }` maps.
- Could optionally support more structured overrides later if needed (e.g. multiple cost categories per opcode).

#### Base idea

- Introduce a **single hook** in the base fork class that always passes gas maps through a repricing function.
- Fork-specific classes only define their “internal” / baseline gas maps.
- At import time, if the correct environment variable is set, `gas_repricings.json` exists and is non-empty, apply overrides to the gas map.

#### Sketch

- In `forks.py`:

      class BaseFork:
          @classmethod
          @abstractmethod
          def _internal_opcode_gas_map(
                cls, *, block_number: int = 0, timestamp: int = 0
            ) -> Dict[OpcodeBase, int | Callable[[OpcodeBase], int]]:
              # Definition meant to be overloaded in each subclass
              raise NotImplementedError

          @classmethod
          def opcode_gas_map(
                cls, *, block_number: int = 0, timestamp: int = 0
            ) -> Dict[OpcodeBase, int | Callable[[OpcodeBase], int]]:
              base_map = cls._internal_opcode_gas_map()
              return apply_gas_repricings(cls, base_map)

- For each fork (e.g. `Amsterdam`), rename all current `opcode_gas_map` to `_internal_opcode_gas_map`:

      class Amsterdam(BPO2):
          """Amsterdam fork."""
          @classmethod
          def _internal_opcode_gas_map(
                cls, *, block_number: int = 0, timestamp: int = 0
            ) -> Dict[OpcodeBase, int | Callable[[OpcodeBase], int]]:
              return {
                  Op.DUP1: 3,
                  Op.SWAP1: 3,
                  # ...
              }

- `base_fork.py`:
  - On import:
    - Resolve current working directory.
    - Look for `gas_repricings.json`.
    - Optionally check an env var gate (e.g. `EELS_GAS_REPRICINGS=1`) for extra safety (We can also set `EELS_GAS_REPRICINGS` to be a path and then look for that file, up to implementer)
    - If file exists, pydantic-load the file into a global variable `GAS_REPRICINGS`.
    - If file doesn’t exist or is empty, set the same global variable as an empty dictionary.
    - If it exists:
      - Load JSON.
      - Identify the current fork key (e.g. `"amsterdam"`).
      - For entries in that fork’s map, `base_map.update(overrides)`.
  - Should be designed so that:
    - No repricing occurs unless config is present (and optionally the env var is set).
    - A warning is emitted when repricings are applied:
- `apply_gas_repricings(cls, base_map)`:
  - On call:
    - Check if `GAS_REPRICINGS` global variable contains the current fork of `cls`
    - Modify each gas constant in the provided `base_map` and return the modified map.

### 3. Spec side (EELS fork gas definitions)

The spec side needs a similar mechanism, but with care for readability and “spec purity”.

#### Constraints

- Specs should remain readable and look like “normal” spec code.
- By default (no config), they must behave exactly as now.
- Repricing must not require every spec file to know about JSON/config.
- Everything must happen on import. This way we don't need to touch T8N interface.

#### Approach

** TBD **

### 4. Activation / safety

To reduce accidental use:

- **Config file presence**: repricing only happens if `gas_repricings.json` exists and contains relevant entries.
- **Optional env var**: e.g. `EELS_GAS_REPRICINGS=1`
  - If not set, even with a JSON file present, we can:
    - Either skip repricing entirely, or
    - Emit a warning that the config is ignored.

- **Warnings**:
  - Emit a `warnings.warn` (or logging) on **first use** per process:
    - Which file is being used.
    - For which fork(s) repricings are active.

#### a) Local repricing experiment

1.  Create `gas_repricings.json` in the repo root:

        {
          "amsterdam": {
            "DUP": 1000
          }
        }

    Note: We can do a trick here where instead of adding DUP1, DUP2, etc, we add a single DUP, but up to implementer.

2.  Set env var (if we decide to gate by env):

        export EELS_GAS_REPRICINGS=1

3.  Run benchmarks / tests:

        uv run fill <some-config>
        uv run execute --remote ...

4.  Observe breakage / results with `DUP` repriced for Amsterdam (Ideally, thanks to renamings and use of `gas_cost`, no breakage should occur).
5.  When done, delete or clear `gas_repricings.json` (or unset env var).

#### b) Finalizing repricing

1.  After converging on a final gas schedule via the above workflow:
    - Update:
      - The spec constants (per-fork `gas.py`).
      - The testing framework’s fork definitions.
    - Remove/ignore the JSON overrides.
2.  Regenerate fixtures/benchmarks (`fill`).
3.  Sync with client tooling (e.g. Pitstop) for updating client-side constants.

## Open Questions / Details to Decide

1.  **Exact JSON schema**
    - Just flat `{ fork: { opcode: gas } }`?
    - Do we need multiple “cost dimensions” (base, warm, cold, etc.) in the first version?

2.  **Shared vs duplicated repricing helper**
    - Single helper module shared across specs and tests?
    - Or two separate but identical helpers that both read `gas_repricings.json`? (Probably this)

3.  **Fork naming / mapping**
    - How do we map fork names used in JSON to actual classes/modules?
    - Should we enforce a canonical string (e.g. `"amsterdam"`, `"osaka"`)?

4.  **Env var gating**
    - Do we require `EELS_GAS_REPRICINGS=1` (or similar) or just rely on the presence of the JSON file?
    - Any additional safety checks we want?

5.  **Spec readability**
    - Do we prefer small helper wrappers (e.g. `OPCODE_GAS_COSTS = fork_gas("amsterdam", {...})`) over calling `apply_gas_repricings` directly?
    - How much “config awareness” are we okay with in spec modules?

## Next Steps

- [x] Finish benchmark refactoring (Louis: stateful + compute benchmarks already migrated to gas calculator).
- Rebase and update Carson’s **gas-constant renaming/alignment** PR:
  - [ ] Ensure spec and testing constants are in sync.
  - [ ] Replace remaining hardcoded usages with references to unified constants where appropriate.
- Get design feedback from stakeholders:
   - [ ] Maria
   - [ ] Kamil
   - [ ] Jochem
- Design and implement `apply_gas_repricings` helpers, specs and test since they will possibly be different:
  - [ ] Decide JSON schema and env var behavior.
  - [ ] Add warnings and tests.
- Wire repricing helper into:
  - [ ] Base fork class on testing side (`BaseFork.opcode_gas_map`).
  - [ ] Per-fork spec gas definitions.
- Document the workflow (README / docs):
  - [ ] How to create and use `gas_repricings.json`.
  - [ ] Example for repricing a few opcodes on Amsterdam.
  - [ ] Caveats and cleanup steps.

Carsons-Eels:
## 🗒️ Description

Implements gas repricing support for both the EEST test framework and the spec-side execution code, allowing users to override gas constants via a JSON config file without modifying source code. This enables fast iteration on alternative gas schedules, "what-if" analysis of proposed EIP gas changes, and comparative test runs under different cost models.

### Activation model

Repricing activates via a single env var, `EELS_GAS_REPRICING_CONFIG`, which must be a path to a valid JSON config. Unset means off, a valid path means on, and a missing path fails with `FileNotFoundError`. I chose this because the explicit-path approach is harder to trigger accidentally and lets the config be anywhere.

## 🔗 Related Issues or PRs

- Would close #2200
- Should merge #2383 first, which cleans up for things missed in #2094
- Should merge #2396 next, which adds opcode specific gas constants
- Will help with things like #1879

### Core Repricing Mechanism

- **Shared config reader `ethereum.utils.gas_repricing`**: Loads a JSON config from the path specified by the `EELS_GAS_REPRICING_CONFIG` environment variable. The config maps fork names to field overrides (e.g., `{"Osaka": {"GAS_VERY_LOW": 4}}`). Results are cached via `lru_cache`. Both spec-side and testing-side consumers import this shared reader.

- **Testing-side repricing**: `BaseFork.gas_costs()` is now concrete — it calls the new abstract `_base_gas_costs()` to get the fork's default costs, then passes them through `apply_repricing()`. Validation (field name + type checks against `GasCosts` dataclass) happens at apply time in the testing module. All fork subclasses have been renamed from `gas_costs()` to `_base_gas_costs()` accordingly.

- **Spec-side repricing**: `apply_spec_repricing()` is called at the bottom of all 24 `src/ethereum/forks/*/vm/gas.py` modules. It mutates module globals in place, preserving `Uint`/`U64` type wrappers via `type(original)(value)`. Validation against the module's actual globals ensures only real constants can be overridden. With no config set, this is a complete no-op.

- **`.gitignore`**: Added `gas_repricing.json` so local config files aren't accidentally committed.

### Gas Constant Name Unification

> [!important]
> 
> Covered in PR #2383 which should be merged _BEFORE_ this PR

Standardized 17 spec-side gas constant names to match the testing-side conventions (e.g., `GAS_ECRECOVER` → `GAS_PRECOMPILE_ECRECOVER`, `GAS_WARM_ACCESS` → `GAS_WARM_ACCOUNT_ACCESS`). This ensures users can use the same field names in their repricing config for both spec and test overrides. ~362 files affected, all within `src/ethereum/forks/`.

### :sparkles: CLI Tool `gas-map`

Invoke with:

```bash
uv run gas-map
# or
uv run eest gas-map
```

The mapping between opcodes (ADD, SLOAD, etc.) and `GasCosts` field names (GAS_LOW, GAS_WARM_SLOAD, etc.) is not obvious without reading source code. The `gas-map` CLI tool bridges this gap:

- Registered both as a standalone entry point and as an `eest` subcommand
- `--fork`/`-f` option (defaults to latest fork)
- `--opcode`/`-o` option for single-opcode detail view
- Uses source code analysis of the `opcode_gas_map()` inheritance chain and helper methods to accurately reverse-map opcodes to their `GasCosts` fields
- Groups static opcodes by tier (GAS_VERY_LOW, GAS_LOW, etc.)
- Details dynamic opcodes' relevant fields (including `GAS_MEMORY` for memory-expanding opcodes like REVERT), and shows constants
- Single-opcode view shows type, field values, and a ready-to-paste `gas_repricing.json` snippet

### :memo: Documentation

- **`docs/gas_repricing/repricing_guide.md`**: What repricing is, JSON config format, activation via env var, `gas-map` CLI usage, and an end-to-end example workflow.
- **`docs/gas_repricing/reference.md`**: Static convenience reference of all ~90 `GasCosts` fields grouped by category, with affected opcodes and typical values. `uv run gas-map` is the authoritative source of truth per fork.
- **`docs/navigation.md`**: New "Gas Repricing" nav section added after "Writing Tests".

### :microscope: Tests

- **`test_gas_repricing.py`**: Full test suite covering:
  - `load_repricing_config()` (shared reader):
    - no env var, empty env var, missing file
    - unknown fields pass through (no validation in shared reader)
    - valid config loading, warning emission
  - `apply_repricing()` (testing-side):
    - no config, no matching fork
    - field overrides applied correctly
    - invalid field names raise `ValueError`
    - non-int values raise `TypeError`
  - `apply_spec_repricing()` (spec-side):
    - no config (no-op), missing fork (no-op)
    - module globals mutated with correct `Uint`/`U64` type wrappers
    - unknown constant names raise `ValueError`
    - non-existent config file raises `FileNotFoundError`
  - Integration with fork classes:
    - `FORK.gas_costs()` returns repriced values
    - Earlier forks are unaffected by later fork repricing
    - Transition forks work correctly
  - Cache clearing between tests via `autouse` fixture

### :warning: Notes

> [!note]
> - **`ethereum-spec-lint` warnings**: The `apply_spec_repricing()` calls at the bottom of each `gas.py` produce 8 "The expression \<class 'ast.Expr'\> has been ignored." print messages from the `GlacierForksHygiene` lint. These are informational prints (not diagnostic errors) — the lint compares glacier forks against predecessors and doesn't have a visitor for `ast.Call` expressions. The warnings are harmless and do not cause lint failure. We could update the linter to ignore this, but I didn't want to go changing the linter in this PR.
> - **`reference.md` maintenance**: The static tables in `reference.md` are a convenience reference. `uv run gas-map` is the authoritative per-fork source. A CI freshness check could be added in the future.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```
- [x] Tests: Ran `mkdocs build --strict` locally and verified docs render correctly.

#### Cute Animal Picture

**Bobcat** by [Jean Beaufort](https://www.publicdomainpictures.net/en/browse-author.php?a=113726)
License: [CC0 Public Domain](http://creativecommons.org/publicdomain/zero/1.0/)


LouisTsai-Csie:
Amazing!

Link related issue #1879 .

LouisTsai-Csie:
There is a problem using this tool for repricing testing: eip-7904 reprices opcodes that currently share a `GasCosts` tier to different individual costs:

Example:

```
┌────────┬─────────┬────────────────┬───────────────────┐
│ Opcode │ Current │ GasCosts field │ EIP-7904 new cost │
├────────┼─────────┼────────────────┼───────────────────┤
│ DIV    │ 5       │ GAS_LOW        │ 15                │
├────────┼─────────┼────────────────┼───────────────────┤
│ SDIV   │ 5       │ GAS_LOW        │ 20                │
├────────┼─────────┼────────────────┼───────────────────┤
│ MOD    │ 5       │ GAS_LOW        │ 12                │
├────────┼─────────┼────────────────┼───────────────────┤
│ SMOD   │ 5       │ GAS_LOW        │ 5 (unchanged)     │
├────────┼─────────┼────────────────┼───────────────────┤
│ MUL    │ 5       │ GAS_LOW        │ 5 (unchanged)     │
├────────┼─────────┼────────────────┼───────────────────┤
│ ADDMOD │ 8       │ GAS_MID        │ 8 (unchanged)     │
├────────┼─────────┼────────────────┼───────────────────┤
│ MULMOD │ 8       │ GAS_MID        │ 11                │
└────────┴─────────┴────────────────┴───────────────────┘
```

The repricing config operates at the `GasCosts` field level, there's no way to express: DIV=15, SDIV=20, MOD=12, SMOD=5, changing `GAS_LOW` would move all five opcodes to the same value. Same issue with `GAS_MID`: `MULMOD` needs 11 but `ADDMOD` stays at 8.

I would suggest the `gas_repricings.json` follow the format defined in issue #2200 

```json
{
  "amsterdam": {
    "DUP": 1000,
    "SSTORE": 5000
  },
  "osaka": {
    "CALL": 800
  }
}
```

marioevz:
> The repricing config operates at the `GasCosts` field level, there's no way to express: DIV=15, SDIV=20, MOD=12, SMOD=5, changing `GAS_LOW` would move all five opcodes to the same value. Same issue with `GAS_MID`: `MULMOD` needs 11 but `ADDMOD` stays at 8.
> 
> I would suggest the `gas_repricings.json` follow the format defined in issue #2200
> 
> ```json
> {
>   "amsterdam": {
>     "DUP": 1000,
>     "SSTORE": 5000
>   },
>   "osaka": {
>     "CALL": 800
>   }
> }
> ```

We would need an intermediate mapping in each fork:
```python
GAS_LOW = 5
...
GAS_OPCODE_DIV = GAS_LOW
GAS_OPCODE_SDIV = GAS_LOW
GAS_OPCODE_MOD = GAS_LOW
```
Only then we would be able to target specific opcode prices with this tool.

It feels necessary too, I don't think we plan on affecting other Opcodes when repricing only because they use the same constant predecessor.

Should we open a precursor PR to create these new variables?

Carsons-Eels:
I ran into this issue headfirst while implementing, and my decision at the time was to implement a simpler version that dealt only with the constants just to get the thing up and running.

The problems with doing it per-opcode then were:
- Updating the costs for the constants took care of keeping the downstream opcode costs in line
- Doing it per-opcode pollutes the spec code by adding indirection that is harder to read and follow
- Easier to keep spec/test in sync with a single, simple config file

However, I think I misunderstood that we would not strictly need per-opcode repricing right from the outset, and by that time I was halfway done implementing and wanted to at least reach an island of sanity before planning a refactor. I also wasn't sure that *this* PR was the place to do that, and thought about kicking that to a different issue because, yeah, as you pointed out @marioevz, this requires a bunch of intermediate mapping that I wasn't sure was appropriate for this PR.

I think a precursor PR is the way to go so that this PR has the pieces it needs to implement it properly. I'm happy to do that tomorrow after I fix the 5 test failures in the CI.

Carsons-Eels:
Right now the instruction files from all 24 forks import gas constants using `from ..gas import CONSTANT`, which creates shallow copies at import time rather than live references to the gas module's namespace. The `apply_spec_repricing()` function (from #2331) mutates `gas.py`'s module globals at load time. If a gas constant has already been copied into the module's namespace with `from ..gas import`, the repriced value is silently lost — the instruction module would use the original value because only the copy changed.

Right now it works because `interpreter.py` happens to import `gas` (line 46) before importing instruction modules (line 55), so `gas.py` executes (including `apply_spec_repricing()`) before any instruction module copies those constants. This makes it pretty fragile; reordering those imports or introducing a circular dependency would silently break the repricing. Though, our ruff formatting keeps this kind of predictable at least.

PR #2396 has this fixed for the new `gas.GAS_OPCODE_*` module references. However, the remaining ~96 instruction files that would need modification still use the `from ..gas import` pattern.

I'm wondering if it's the right call to do this here in this PR, or in a separate PR. Mostly because it would bloat the PR another 96 files changed to 134 files which makes the other changes a pain to review. 

For now, I've made those changes in a [separate branch](https://github.com/Carsons-Eels/execution-specs/pull/4) as a PR to myself targeted at PR #2396

src/ethereum/utils/gas_repricing.py
```diff
@@ -0,0 +1,66 @@
...
+    Mutate the class attributes in place, preserving the
+    original type wrapper (Uint, U64, etc.).
+    """
+    config = load_repricing_config()
+    if config is None:
+        return
+
+    overrides = config.get(fork_name)
+    if overrides is None:
+        return
```
LouisTsai-Csie, line 56: Would be nice if we could add logging here, this could alert user that the file is loaded but there is no fork applied.

src/ethereum_spec_tools/lint/lints/glacier_forks_hygiene.py
```diff
@@ -231,6 +231,15 @@ def visit_Expr(self, expr: ast.Expr) -> None:
...
             return
 
+        # Ignore the module-level apply_spec_repricing() hook appended to
+        # each fork's gas.py; it is identical machinery across forks.
+        if (
+            isinstance(expr.value, ast.Call)
+            and isinstance(expr.value.func, ast.Name)
+            and expr.value.func.id == "apply_spec_repricing"
+        ):
+            return
```
LouisTsai-Csie, line 241: I am not very confident in reviewing this part, maybe @gurukamath could help review this change?

packages/testing/src/execution_testing/cli/eest/commands/gas_map.py
```diff
@@ -0,0 +1,293 @@
...
+
+OPCODE_TIER_FIELDS = (
+    "OPCODE_JUMPDEST",
+    "BASE",
+    "VERY_LOW",
+    "LOW",
+    "MID",
+    "HIGH",
+    "OPCODE_BLOCKHASH",
+    "WARM_SLOAD",
```
LouisTsai-Csie, line 23: Just a comment: this is manually maintained and could get outdated. Not sure of a better approach.

packages/testing/src/execution_testing/cli/eest/commands/gas_map.py
```diff
@@ -0,0 +1,293 @@
...
+                # the guard above ensures that opcode_gas_map exists before
+                # adding the source.
+                method = cls.opcode_gas_map  # type: ignore[attr-defined]
+                sources.append(inspect.getsource(method))
+            except (OSError, TypeError):
+                pass
+    return "\n".join(sources)
+
+
+def _get_helper_method_fields(
```
LouisTsai-Csie, line 71: there are some issue with this helper, here is one example:

```bash
$ gas-map --fork Osaka --opcode CREATE

CREATE — Osaka
══════════════════════════════
Type:     dynamic
GasCosts: BASE (2), MEMORY_PER_WORD (3), OPCODE_CREATE_BASE (32000)

To reprice in gas_repricing.json:
{
  "Osaka": {
    "BASE": 0, # incorrect
    "MEMORY_PER_WORD": 0, # correct
    "OPCODE_CREATE_BASE": 0 # correct
  }
}
# missing CODE_INIT_PER_WORD
```

It first detects which GasCosts field a helper uses with a plain substring check:

```python
for field_name in valid_fields:
  if field_name in src:
      found.add(field_name)
```

a shorter field name is falsely matched whenever it appears inside a longer one. For example, `BASE` is found inside `OPCODE_CREATE_BASE`, `LOW` inside `VERY_LOW`.

Second, the same helper method, take `_calculate_create_gas` as example, is written in three different classes that Osaka inherits from. Each version uses different gas fields:
  - EIP3860 version → uses CODE_INIT_PER_WORD
  - Frontier version → uses OPCODE_CREATE_BASE
  - BaseFork version → uses nothing

When Osaka actually runs, it should pick the latest one ->  EIP3860's version. That's the one that really counts.

But the function loops through all three and keeps overwriting the result each time it finds a version:

```python
helper_fields[name] = found   # this replaces, it 
```

The loop goes from newest class to oldest. So it sets the value to EIP3860's fields first, then replaces it with Frontier's, and BaseFork's is skipped because it's empty. Whatever the loop touches, the last wins, in this case, it is the oldest class, Frontier.

packages/testing/src/execution_testing/forks/tests/test_gas_repricing.py
```diff
@@ -0,0 +1,341 @@
+"""Tests for gas repricing override mechanism."""
```
LouisTsai-Csie, line 1: The test looks good, but the style here is inconsistent with the rest of `forks/tests/` folder, which uses flat, parametrized `test_*` functions rather than `Test*` classes. 

Non-blocking, but we could consider to align it with the convention just for consistency.

packages/testing/src/execution_testing/cli/eest/commands/gas_map.py
```diff
@@ -0,0 +1,293 @@
...
+
+        brace_depth += line.count("(") - line.count(")")
+        if current_opcode and brace_depth <= 0 and "," in line:
+            current_opcode = None
+            brace_depth = 0
+
+    return {k: sorted(v) for k, v in opcode_fields.items()}
+
+
+def _format_grouped_output(fork_class: type[BaseFork]) -> str:
```
LouisTsai-Csie, line 138: Currently the field shown is based on tier_reverse, so for example `ADD` is reported as `VERY_LOW`, but it should show `OPCODE_ADD`. Because `OPCODE_ADD` is an independent field, setting `VERY_LOW` to another value does not update the cost of `ADD`. So the guidance points users at a field that has no effect

Below is an example:

```
# The tool reports ADD under VERY_LOW:
$ uv run gas-map --fork Osaka --opcode ADD

opcode ADD
ADD — Osaka
══════════════════════════════
Type:     static
GasCosts: VERY_LOW (3)

To reprice in gas_repricing.json:
{
  "Osaka": {
    "VERY_LOW": 0
  }
}
```

The result:
```
# repricing VERY_LOW has no effect on ADD:
$ EELS_GAS_REPRICING_CONFIG=<{"Osaka":{"VERY_LOW":999}}> ... → ADD cost = 3   (unchanged)

# Repricing the actual field works:
$ EELS_GAS_REPRICING_CONFIG=<{"Osaka":{"OPCODE_ADD":999}}> ... → ADD cost = 999
```

Command to reproduce:
Using tier cost:
```
# repricing.json
{"Osaka":{"VERY_LOW":999}}

# the result is 3, not 999
$ EELS_GAS_REPRICING_CONFIG=./repricing.json uv run python -c "from execution_testing.forks.forks.forks import Osaka;print(Osaka.gas_costs().OPCODE_ADD)"
```
Using opcode based cost:
```
# repricing.json
{"Osaka":{"OPCODE_ADD":999}}

# the result is 999
$ EELS_GAS_REPRICING_CONFIG=./repricing.json uv run python -c "from execution_testing.forks.forks.forks import Osaka;print(Osaka.gas_costs().OPCODE_ADD)"
```
I think we should display the source instead of the tier. In future repricing updates, we may have different tier groups, and opcodes currently in the `LOW` tier could be split into different categories.

LouisTsai-Csie: We could find similar issue in `_format_single_opcode`

LouisTsai-Csie (changes requested):
Thank you for adding this feature! I leave some comment and question, please let me know if i missed anything

PR #2334, opened 2026-02-26
feat(bench): parametrize SLOAD before SSTORE in ERC20 approve

CPerezz:
## Summary
- Adds `sloads_before_sstore` parameter to `test_sstore_erc20_approve`, following the same pattern used by `test_sstore_variants`
- When `True` (existing behavior): SLOAD warms the slot, then SSTORE writes at warm cost (2,900 gas)
- When `False` (new variant): SSTORE hits a cold slot directly (5,000 gas), matching a vanilla Solidity `approve()` which does not SLOAD before writing

## Motivation
Standard ERC-20 `approve()` compiles to a bare SSTORE — there's no preceding SLOAD. The existing benchmark always included an SLOAD, which warmed the slot and changed the SSTORE gas profile. Parametrizing this lets us benchmark both patterns and decide which better represents the workload we want to measure.

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -32,6 +32,7 @@ from execution_testing import (
 )
 
 from tests.benchmark.stateful.helpers import (
+    ALLOWANCE_SELECTOR,
     APPROVE_SELECTOR,
     BALANCEOF_SELECTOR,
     CacheStrategy,
@@ -321,16 +322,20 @@ def test_sstore_erc20_approve(
     )
 
     if cache_strategy == CacheStrategy.CACHE_TX:
-        # Call balanceOf first to warm the storage slot, then
-        # restore the approve selector
+        # Call allowance(ADDRESS, spender) to warm the allowance
+        # storage slot that approve will later write to.
+        # Memory: save spender→[64], put ADDRESS→[32],
+        # set allowance selector, call, then restore.
         cache_warmup = (
-            Op.MSTORE(0, BALANCEOF_SELECTOR)
+            Op.MSTORE(64, Op.MLOAD(32))
+            + Op.MSTORE(32, Op.ADDRESS)
+            + Op.MSTORE(0, ALLOWANCE_SELECTOR)
             + Op.POP(
                 Op.CALL(
                     address=erc20_address,
                     value=0,
                     args_offset=28,
-                    args_size=36,
+                    args_size=68,
                     ret_offset=0,
                     ret_size=0,
                     # gas accounting
@@ -338,6 +343,7 @@ def test_sstore_erc20_approve(
                 )
             )
             + Op.MSTORE(0, APPROVE_SELECTOR)
+            + Op.MSTORE(32, Op.MLOAD(64))
         )
     else:
         cache_warmup = Bytecode()
@@ -407,6 +413,44 @@ def test_sstore_erc20_approve(
     )
 
     function_dispatch_cost = function_dispatch.gas_cost(fork)
+
+    if cache_strategy == CacheStrategy.CACHE_TX:
+        # Add allowance dispatch cost for the warmup call.
+        # allowance(owner, spender) computes the same double-
+        # keccak slot as approve but does SLOAD + RETURN.
+        function_dispatch_allowance = (
+            Op.PUSH4(ALLOWANCE_SELECTOR)
+            + Op.EQ
+            + Op.JUMPI
+            + Op.JUMPDEST
+            + Op.CALLDATALOAD(4)
+            + Op.CALLDATALOAD(36)
+            + Op.MSTORE(0, Op.CALLDATALOAD(4))
+            + Op.MSTORE(32, 1)
+            + Op.SHA3(
+                0,
+                64,
+                # gas accounting
+                data_size=64,
+                old_memory_size=0,
+                new_memory_size=64,
+            )
+            + Op.MSTORE(32)
+            + Op.MSTORE(0, Op.CALLDATALOAD(36))
+            + Op.SHA3(
+                0,
+                64,
+                # gas accounting
+                data_size=64,
+            )
+            + Op.SLOAD
+            + Op.PUSH0
+            + Op.MSTORE
+            + Op.RETURN(0, 32)
+        )
+        function_dispatch_cost += (
+            function_dispatch_allowance.gas_cost(fork)
+        )
 
     # Transaction Loops
     txs = []
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -34,12 +34,12 @@ from execution_testing import (
 from tests.benchmark.stateful.helpers import (
     APPROVE_SELECTOR,
     BALANCEOF_SELECTOR,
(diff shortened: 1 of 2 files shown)


LouisTsai-Csie (approved):
I am not very sure about the reason of modifying the function dispatching pattern here, will take a deeper look at the discussion. But this change is good format.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -330,10 +332,14 @@ def test_sstore_erc20_approve(
             # gas accounting
             data_size=64,
         )
-        + Op.DUP1
-        + Op.SLOAD.with_metadata(access_warm=False)
-        + Op.POP
-        + Op.SSTORE
+        + (
+            Op.DUP1
+            + Op.SLOAD.with_metadata(access_warm=False)
```
marioevz:
```suggestion
            + Op.SLOAD.with_metadata(key_warm=False)
```
Incorrect metadata, I think the test was not failing because we didn't use the gas cost calculator, but if we ever do this should fix calculations.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -330,10 +332,14 @@ def test_sstore_erc20_approve(
...
+            + Op.POP
+            + Op.SSTORE
```
marioevz:
```suggestion
            + Op.SSTORE.with_metadata(key_warm=True)
```
`False` is the default, so setting `True` here is necessary.

marioevz (changes requested):
Just a couple of comments, thanks!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -235,6 +236,7 @@ def test_sstore_erc20_approve(
     gas_benchmark_value: int,
     tx_gas_limit: int,
     token_name: str,
+    sloads_before_sstore: bool,
```
jochem-brouwer: Can this be changed to the new https://github.com/ethereum/execution-specs/blob/a7880f3b98e11563e3009045cbfea0812776f380/tests/benchmark/stateful/helpers.py#L41 ?  😄 👍 

The cache strategy here would be the tx-level cache. 

You can see how it is used in:

https://github.com/ethereum/execution-specs/blob/a7880f3b98e11563e3009045cbfea0812776f380/tests/benchmark/stateful/bloatnet/test_single_opcode.py#L95-L96

https://github.com/ethereum/execution-specs/blob/a7880f3b98e11563e3009045cbfea0812776f380/tests/benchmark/stateful/bloatnet/test_single_opcode.py#L433-L441

The reason why we have multiple cache strategies is that: the no cache should be *really* cold, the tx level one should be warm in both cache and EVM, and if you warm (you load it) in the previous block then in the "attack block" it is cold according to EVM, but in practice we will (likely) see that it is still warm, because we just loaded it from disk and likely clients use some kind of (LFU?) cache.

Also to keep things consistent.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -330,10 +332,14 @@ def test_sstore_erc20_approve(
...
-        + Op.SLOAD.with_metadata(access_warm=False)
-        + Op.POP
-        + Op.SSTORE
+        + (
+            Op.DUP1
+            + Op.SLOAD.with_metadata(key_warm=False)
+            + Op.POP
+            + Op.SSTORE.with_metadata(key_warm=True)
+            if sloads_before_sstore
+            else Op.SSTORE.with_metadata(key_warm=False)
```
jochem-brouwer, line 402: Yes very elegant 😄 👍

tests/benchmark/stateful/bloatnet/test_single_opcode.py
CPerezz: Thanks for this! Appreciate the explanation!!!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -263,23 +266,48 @@ def test_sstore_erc20_approve(
...
+                ret_offset=0,
+                ret_size=0,
+                # gas accounting
+                address_warm=True,
+            )
+        )
+    )
+
+    if cache_strategy == CacheStrategy.CACHE_TX:
+        # Call balanceOf first to warm the storage slot, then
```
jochem-brouwer: `balanceOf` reads the slot we write to for `transfer` and `transferFrom`. We are targeting `approve(address _spender, uint256 _value)` in this benchmark so this should call `allowance(address _owner, address _spender)`. Here `owner` is `ADDRESS` (because this address will be the msg.sender in the ERC20 target) and the spender is the target key (address) we write to later

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -263,23 +266,48 @@ def test_sstore_erc20_approve(
...
+        # restore the approve selector
+        cache_warmup = (
+            Op.MSTORE(0, BALANCEOF_SELECTOR)
             + Op.POP(
                 Op.CALL(
                     address=erc20_address,
                     value=0,
                     args_offset=28,
-                    args_size=68,
+                    args_size=36,
```
jochem-brouwer: Ah this is a nice trick to reuse the calldata of this call later for approve. However we can't do that because it thus writes to the wrong slot. So here we should call `allowance(Op.ADDRESS, spender)` and then for the write `approve(spender, value)`

jochem-brouwer: For this you can move the spender first to memory 64 and then write ADDRESS to 32. Then to restore the memory write the spender we just moved to slot 64 should go back in 32.

tests/benchmark/stateful/helpers.py
```diff
@@ -29,3 +30,15 @@
     for k in _STUBS.keys()
     if k.startswith("test_mixed_sload_sstore_")
 ]
+
+
+class CacheStrategy(str, Enum):
```
jochem-brouwer, line 73: I'm confused, why is this change here? This is already in `forks/amsterdam` target branch right?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -364,24 +397,40 @@ def test_sstore_erc20_approve(
...
         calldata = Hash(num_calls) + Hash(slot_offset)
 
-        txs.append(
-            Transaction(
-                gas_limit=gas_available,
-                data=calldata,
-                to=attack_contract_address,
-                sender=pre.fund_eoa(),
-                access_list=access_list,
+        if cache_strategy == CacheStrategy.CACHE_PREVIOUS_BLOCK:
```
jochem-brouwer, line 468: We likely want to make a helper for this pattern, because I think other tests integrating this will mostly do the same approach to essentially just run the same block twice.

jochem-brouwer (changes requested):
Looks good, we only have to change the cache to call balance to call allowance, which also means we have to move memory around slightly more 😄 👍

jochem-brouwer (approved):
LGTM! Thanks a lot @CPerezz 🙏 😄

marioevz (approved):
LGTM.

PR #2335, opened 2026-02-26
chore(ci,deps): remove `setup-env` from jobs that don't need it

danceratopz:
## Description

Remove the `setup-env` step from CI jobs that don't need Rust, `build-essential`, or Geth. Noticed while working on #2310 and #2283.

`setup-env` installs the native build toolchain and Geth. After #2310, `setup-uv` already handles Python, uv, and tox -- so `setup-env` is only needed by jobs that compile `ethereum-execution[optimized]` (Rust) or run Geth.

Four jobs don't need either:

- `py3` / `pypy3`: run `fill` using EELS (no Geth), don't use `optimized`.
- `tests_pytest_py3` / `tests_pytest_pypy3`: run testing package unit tests with evmone. `build-evmone` is self-contained. No Geth needed.

`py3` and `pypy3` are on the critical path -- removing `setup-env` saves ~2 min per CI run.

### tox.ini

Override `dependency_groups` in `py3` and `tests_pytest_py3` to exclude `ethereum-execution[optimized]`, matching `pypy3` and `tests_pytest_pypy3` which already do this. Also applied the same override to the three benchmark envs (`benchmark-gas-values`, `benchmark-fixed-opcode-cli`, `benchmark-fixed-opcode-config`) for consistency -- they run on `ubuntu-latest` (which has Rust pre-installed) so they happen to work today, but explicitly excluding `optimized` makes the intent clear and avoids a future footgun if runners change.

### test.yaml

Delete `- uses: ./.github/actions/setup-env` from `py3`, `pypy3`, `tests_pytest_py3`, `tests_pytest_pypy3`. `json_infra` is left as-is (may need Geth).

#### Before
https://github.com/ethereum/execution-specs/actions/runs/22436513783?pr=2310
image

#### After
https://github.com/ethereum/execution-specs/actions/runs/22441413436?pr=2335
image

## Related Issues or PRs

- #2310
- #2283

## Checklist

- [x] All: Ran fast `tox` checks to avoid unnecessary CI fails:
    ```console
    uvx tox -e static
    ```
- [x] All: PR title adheres to the repo standard.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -63,7 +63,6 @@ jobs:
         with:
           submodules: recursive
       - uses: ./.github/actions/setup-uv
-      - uses: ./.github/actions/setup-env
       - name: Run py3 tests
         run: tox -e py3
         env:
@@ -85,7 +84,6 @@ jobs:
       - uses: ./.github/actions/setup-uv
         with:
           python-version: "pypy3.11"
-      - uses: ./.github/actions/setup-env
       - name: Run pypy3 tests
         run: tox -e pypy3
         env:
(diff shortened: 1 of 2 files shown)


marioevz (approved):
LGTM, thanks!

PR #2353, opened 2026-02-27
feat(test-vm): implement create address calculation helper

LouisTsai-Csie:
## 🗒️ Description

- Add `CreatePreimageLayout` helper class for computing CREATE addresses on-chain, complementing the existing  
  `Create2PreimageLayout`                                                                                        
  - Supports both static nonces (RLP encoded at generation time) and dynamic nonces (CLZ-based branch-free RLP encoding at EVM runtime, nonces 1 to 2^64-1)
  - Provides `address_op()`, `set_nonce_op()`, and `increment_nonce_op()` methods for composing with loop-based contract creation patterns

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -489,6 +489,104 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
 
 
+def _rlp_encode_nonce(nonce: int) -> bytes:
+    """RLP-encode a nonce value for CREATE address computation."""
+    if nonce == 0:
+        return b"\x80"
+    if nonce <= 0x7F:
+        return bytes([nonce])
+    byte_len = (nonce.bit_length() + 7) // 8
+    return bytes([0x80 + byte_len]) + nonce.to_bytes(byte_len, "big")
```
jochem-brouwer: This method looks like the inner RLP encoding of an int (or actually: an uint, since we cannot encode negative values in RLP). I suspect this method is around the spec somewhere else and for cleani-ness / DRY (don't repeat yourself - or in this case have two copies of the same fucntionality) we should likely load this from the somewhere else (likely: the RLP package directly encoding the int)
Also see: 

https://github.com/ethereum/execution-specs/blob/47ae746a07ba797134dcf724a64fb61da78d63c1/src/ethereum/forks/frontier/utils/address.py#L41

https://github.com/ethereum/execution-specs/blob/47ae746a07ba797134dcf724a64fb61da78d63c1/src/ethereum/forks/frontier/utils/address.py#L53-L62

jochem-brouwer: Also see this quick interactive python:

```python
>>> from ethereum_rlp import rlp
>>> from ethereum_types.numeric import Uint
rlp.encode(Uint(1)).hex()
'01'
>>> rlp.encode(Uint(80)).hex()
'50'
>>> rlp.encode(Uint(128)).hex()
'8180'
>>> rlp.encode(Uint(127)).hex()
'7f'
>>> rlp.encode(Uint(0)).hex()
'80'
```

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -489,6 +489,104 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
+
+
+class CreatePreimageLayout(Bytecode):
```
jochem-brouwer, line 598: This works, but for most of our benchmarks we need a way to call this looper like CREATE2, where for CREATE2 we change salt and then calculate the address.
In here, the code generated targets a constant address. So if we want to calculate a new address we have to keep adding bytecode plus we have to know before what specific nonces we want so we hardcode the code to calculate it.
We should instead have a function to do this which accepts nonce (and address) as input.

jochem-brouwer (commented):
Hey Louis, great start! 😄 👍 I had a few comments, and also have a follow-up PR on top of this one: https://github.com/LouisTsai-Csie/execution-specs/pull/3

tests/osaka/create/test_create_address_computation.py
```diff
@@ -0,0 +1,165 @@
...
+    CreatePreimageLayout,
+    StateTestFiller,
+    Transaction,
+    While,
+)
+from execution_testing.vm import Op
+
+
+@pytest.mark.valid_from("Osaka")
+def test_create_address_dynamic_nonce(
```
jochem-brouwer: Ah right, this does introduce a normal test but this should actually "test the benchmark" and is not really intended to  be included in tests release I think

LouisTsai-Csie: I moved it under the `Frontier` fork, since that is the appropriate place for testing the `CREATE` opcode.

packages/testing/src/execution_testing/tools/tools_code/generators.py
LouisTsai-Csie: IIUC, this is done by your PR. thanks

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -489,6 +489,275 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
+        """
+        Assemble the bytecode that sets up the memory layout for
+        CREATE address computation.
+        """
+        if isinstance(nonce, bytes):
+            nonce = int.from_bytes(nonce, "big")
+
+        if isinstance(nonce, int):
+            # Static nonce: direct MSTORE8 approach
+            nonce_rlp = _rlp_encode_nonce(nonce)
```
jochem-brouwer: I think this might not get the correct results. If you run `test_craete_preimage_layout.py` ensuring that it does the "static nonce" then likely it will throw on some/all tests.

The reason is that for hashing you hash `RLP.encode([address, nonce])` where address is of type Bytes20 and nonce is of type Uint. We don't hash `RLP.encode([address, RLP(nonce)])`

I think we should do all with dynamic mode.

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -489,6 +489,275 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
+        return _dynamic_nonce_encode_bytecode(nonce, self.offset)
+
+    def increment_nonce_op(self, increment: int = 1) -> Bytecode:
+        """
+        Increment the nonce, re-encode, and update memory.
+
+        Read the raw nonce from scratch memory, add the
+        increment, and re-encode.
+        """
+        self._dynamic = True
```
jochem-brouwer: This also will change it to dynamic, but since we cannot change the bytecode it runs on (we have to change the contract' code) I don't think this works 🤔 

I think we should all do dynamic nonce and remove the static nonce. So also this "dynamic" flag can be removed

tests/frontier/create/test_create_preimage_layout.py
```diff
@@ -0,0 +1,276 @@
+"""State test for `CreatePreimageLayout` address computation."""
```
jochem-brouwer, line 1: This file is intended to be used as internal check that the CreatePreimageLayout works correctly. It is not a state test.

Should we move this somewhere else? Like can we run these kind of tests also within EELS inner tests?

jochem-brouwer (commented):
Some comments! Could you also check if the tests make sense, and if the files are in logical locations 😄 👍

packages/testing/src/execution_testing/tools/tools_code/generators.py
LouisTsai-Csie: Let me remove the dynamic mode entirely.

packages/testing/src/execution_testing/tools/tools_code/generators.py
LouisTsai-Csie: You are right, the code is hardcoded and there is no way to proceed with this approach.

tests/frontier/create/test_create_preimage_layout.py
LouisTsai-Csie: I do not have a good answer here - will ask the team for advice

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -489,6 +489,212 @@ def increment_salt_op(self, increment: int = 1) -> Bytecode:
...
+        offset: int = 0,
+        old_memory_size: int = 0,
+    ) -> Self:
+        """
+        Assemble the bytecode that sets up the memory layout for
+        CREATE address computation.
+        """
+        if isinstance(nonce, bytes):
+            nonce = int.from_bytes(nonce, "big")
+
```
marioevz:
```suggestion
```

This should be handled by the `Opcode.__call__` in all cases I think.

marioevz (approved):
LGTM, we should add a unit test (not spec test) where the result is actually executed in the EVM. I'll push the test and then merge.

PR #2359, opened 2026-02-27
feat(ci): add PR labeler for conventional commits (#1664)

### LINKED ISSUE #1664: Add labels to pull requests based on conventional commits type/scope
For example, if a pull request contains the commit messages:

- `fix(doc): flaz the floop`
- `fix(test-tools): outgrabe the jabberwocky`

Then the labels `A-doc`, `A-test-tools`, `C-bug` should be added to the pull request.

mandeepmourya007:
Add a GitHub Actions workflow that automatically labels pull requests based on conventional commit messages.

How it works:

Parses commit messages for conventional commit format: type(scope): description
Maps commit types to category labels (e.g., fix → C-bug, feat → C-feat)
Maps scopes to area labels (e.g., doc → A-doc, ci → A-ci)
Only adds labels that exist in the repository
Example:

fix(doc): update readme → adds C-bug, A-doc
feat(test-tools): add new helper → adds C-feat, A-test-tools
🔗 Related Issues or PRs
Fixes #1664

✅ Checklist
All: Ran fast tox checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):

All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start type(scope):.
All: Set appropriate labels for the changes (only maintainers can apply labels).

### DIFF
--- a/.github/workflows/pr-labeler.yaml
+++ b/.github/workflows/pr-labeler.yaml
@@ -1,0 +1,121 @@
+name: PR Labeler
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label PR based on conventional commits
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            const typeToLabel = {
+              'fix': 'C-bug',
+              'feat': 'C-feat',
+              'refactor': 'C-refactor',
+              'chore': 'C-chore',
+              'test': 'C-test',
+            };
+
+            const scopeToLabel = {
+              'doc': 'A-doc',
+              'docs': 'A-doc',
+              'ci': 'A-ci',
+              'deps': 'A-deps',
+              'packaging': 'A-packaging',
+              'tooling': 'A-tooling',
+              'tools': 'A-tooling',
+              'tests': 'A-tests',
+              'test': 'A-tests',
+              'test-tools': 'A-test-tools',
+              'test-benchmark': 'A-test-benchmark',
+              'test-cli': 'A-test-cli',
+              'test-commands': 'A-test-commands',
+              'test-config': 'A-test-config',
+              'test-consume': 'A-test-consume',
+              'test-eest': 'A-test-eest',
+              'test-exceptions': 'A-test-exceptions',
+              'test-execute': 'A-test-execute',
+              'test-execution': 'A-test-execution',
+              'test-fill': 'A-test-fill',
+              'test-fixtures': 'A-test-fixtures',
+              'test-forks': 'A-test-forks',
+              'test-gentest': 'A-test-gentest',
+              'test-make': 'A-test-make',
+              'test-rpc': 'A-test-rpc',
+              'test-specs': 'A-test-specs',
+              'test-tests': 'A-test-tests',
+              'test-types': 'A-test-types',
+              'test-vm': 'A-test-vm',
+              'spec-rpc': 'A-spec-rpc',
+              'spec-specs': 'A-spec-specs',
+              'spec-tests': 'A-spec-tests',
+              'spec-tool': 'A-spec-tool',
+            };
+
+            // Regex for conventional commit: type(scope): description or type: description
+            const conventionalCommitRegex = /^(\w+)(?:\(([^)]+)\))?!?:\s*.+/;
+
+            const { data: commits } = await github.rest.pulls.listCommits({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.payload.pull_request.number,
+            });
+
+            const labelsToAdd = new Set();
+
+            for (const commit of commits) {
+              const message = commit.commit.message.split('\n')[0]; // First line only
+              const match = message.match(conventionalCommitRegex);
+
+              if (match) {
+                const type = match[1].toLowerCase();
+                const scope = match[2] ? match[2].toLowerCase() : null;
+
+                // Add type label
+                if (typeToLabel[type]) {
+                  labelsToAdd.add(typeToLabel[type]);
+                }
+
+                // Add scope label
+                if (scope && scopeToLabel[scope]) {
+                  labelsToAdd.add(scopeToLabel[scope]);
+                }
+              }
+            }
+
+            if (labelsToAdd.size > 0) {
+              const labelsArray = Array.from(labelsToAdd);
+              console.log(`Adding labels: ${labelsArray.join(', ')}`);
+
+              // Get existing labels on the repo to avoid adding non-existent labels
+              const { data: repoLabels } = await github.rest.issues.listLabelsForRepo({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                per_page: 100,
+              });
+              const existingLabelNames = new Set(repoLabels.map(l => l.name));
+
+              // Filter to only labels that exist in the repo
+              const validLabels = labelsArray.filter(label => existingLabelNames.has(label));
+
+              if (validLabels.length > 0) {
+                await github.rest.issues.addLabels({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: context.payload.pull_request.number,
+                  labels: validLabels,
+                });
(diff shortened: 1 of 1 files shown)


danceratopz (changes requested):
Edit: lol, sorry about that, Claude went a big rogue :) I'll leave this here, but I'll add some personal comments below and delete what I don't agree with here. 

## Full Review: PR #2359 — PR Labeler for Conventional Commits

Good idea — automating label assignment from conventional commits is a clear maintainer ergonomics win. The implementation works for the simple case but has several issues that should be addressed before merging.

### Request Changes

#### 1. Parse the PR title, not (only) commits

This repo uses squash merging. The PR template explicitly says _"PR title adheres to the repo standard — it will be used as the squash commit message."_ The PR title is the canonical conventional commit, but the workflow only parses individual commit messages (line 65–74). For multi-commit PRs, intermediate commits are often WIP and don't follow conventions, while the PR title is the ground truth.

At minimum, add:
```js
const title = context.payload.pull_request.title;
const titleMatch = title.match(conventionalCommitRegex);
if (titleMatch) { /* extract labels */ }
```

Consider whether parsing individual commits is even needed — parsing just the title might be sufficient and simpler.

This shouldn't be required if we go off PR title:

Edit: deleted 2.

#### 3. Replace hard-coded `scopeToLabel` with dynamic derivation from repo labels

This is the biggest maintainability concern. The `scopeToLabel` map has ~30 entries that must be manually kept in sync with repo labels. It is **already out of date**:
- `A-test-tools` (line 36) does not exist in the repo — phantom mapping
- `A-test-client-clis` exists in the repo but has no scope entry
- Actively-used scopes like `clients`, `consume`, `stubs`, `specs`, `benchmarks`, `testing`, `spec` have no mapping at all

Since the workflow already fetches repo labels at runtime (line 98) to validate, it can go one step further and **derive the entire scope mapping dynamically** from the `A-` prefix convention:

```js
const repoLabels = await github.paginate(github.rest.issues.listLabelsForRepo, {
  owner: context.repo.owner,
  repo: context.repo.repo,
});

// Build scope→label map dynamically from A-* labels
const scopeToLabel = {};
for (const label of repoLabels) {
  if (label.name.startsWith('A-')) {
    scopeToLabel[label.name.slice(2)] = label.name;
  }
}
// Add common aliases
scopeToLabel['docs'] = 'A-doc';
scopeToLabel['tools'] = 'A-tooling';
scopeToLabel['test'] = 'A-tests';
```

This eliminates the hard-coded mapping entirely. New `A-*` labels added to the repo are automatically recognized. The only maintenance is the small alias map for non-obvious synonyms.

The `typeToLabel` map (5 entries) is small, changes rarely, and has non-obvious mappings (`fix`→`C-bug`), so keeping it hard-coded is fine.

#### 4. Paginate API calls

- `pulls.listCommits` (line 65): default page size is 30. PRs with >30 commits will silently miss some. Use `github.paginate()`.
- `issues.listLabelsForRepo` (line 98): `per_page: 100` works today (repo has ~53 labels) but won't scale. Also use `github.paginate()`.

#### 5. Missing `concurrency` group

Every other workflow in this repo uses:
```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
  cancel-in-progress: true
```
This workflow should follow the same pattern.

### Minor Notes

- **Missing type mappings:** `docs` (standard conventional commit type, used in this repo e.g. `docs: add Osaka fork...`) and `bug` (non-standard but used, e.g. `bug(tooling):...`) have no `typeToLabel` entry. Consider mapping `docs`→`C-chore` and `bug`→`C-bug`.
- **Add-only behavior:** Labels are never removed on force-push/reword. This is fine for v1 but worth a code comment documenting it as intentional.
- **`contents` permission:** The workflow only needs `pull-requests: write` (already set correctly). No `contents` read is needed since it doesn't check out code. Good.
- **Action pinning:** `actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1` — SHA-pinned with version comment. Follows repo convention. Good.

<del>
### Summary

The concept is right and the issue is worth solving. The three key changes needed are: (1) parse the PR title as the primary conventional commit source, (2) handle comma-separated multi-scope, and (3) derive scope mappings dynamically from `A-*` repo labels instead of hard-coding them. These are straightforward fixes (~20 lines changed) that will make the workflow both more correct and more maintainable from day one.
</del>

.github/workflows/pr-labeler.yaml
```diff
@@ -0,0 +1,121 @@
...
+          script: |
+            const typeToLabel = {
+              'fix': 'C-bug',
+              'feat': 'C-feat',
+              'refactor': 'C-refactor',
+              'chore': 'C-chore',
+              'test': 'C-test',
+            };
+
+            const scopeToLabel = {
```
danceratopz, line 26: I don't think we should hardcode these as they'll likely go out of date.

.github/workflows/pr-labeler.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label PR based on conventional commits
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            const typeToLabel = {
```
danceratopz, line 18: These will change less often than scopes, perhaps it's ok to hardcode them.

.github/workflows/pr-labeler.yaml
```diff
@@ -0,0 +1,121 @@
...
+    types: [opened, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Label PR based on conventional commits
```
danceratopz, line 14: We merge changes in this repository using a squash commit and don't enforce conventional commit messages on PR branches, only in the squash commit (which is the PR title).

.github/workflows/pr-labeler.yaml
```diff
@@ -0,0 +1,121 @@
...
+            const scopeToLabel = {
+              'doc': 'A-doc',
+              'docs': 'A-doc',
+              'ci': 'A-ci',
+              'deps': 'A-deps',
+              'packaging': 'A-packaging',
+              'tooling': 'A-tooling',
+              'tools': 'A-tooling',
+              'tests': 'A-tests',
+              'test': 'A-tests',
```
danceratopz, line 35: I appreciate the intention, but I think this only reinforces the use of an incorrect scope and would insist that (the PR title) uses the `tests` scope. `test` is ambiguous.

I'm not that keen on adding aliases here as I'd rather maintain a 1-to-1 mapping.

.github/workflows/pr-labeler.yaml
```diff
@@ -0,0 +1,121 @@
...
+              const match = message.match(conventionalCommitRegex);
+
+              if (match) {
+                const type = match[1].toLowerCase();
+                const scope = match[2] ? match[2].toLowerCase() : null;
+
+                // Add type label
+                if (typeToLabel[type]) {
+                  labelsToAdd.add(typeToLabel[type]);
+                }
```
danceratopz, line 84: This means that we will add multiple type labels, one for each unique type in the commit set, whereas a PR should only get one type. This will cause more work for maintainers, as they will have to remove the types that aren't applicable.

danceratopz (commented):
Hey @mandeepmourya007! Many thanks for your effort implementing this!

Unfortunately, #1664 is out-of-date. We no longer insist on PR commits following conventional commits, which means this is not a particularly helpful approach. Even if we did impose conventional commits, I'm not that optimistic that it would be very useful. See the comment below about applying type, for example.

We do insist that the PR title follows a `type(scope): ...` conventional commit style, as this will be used as the merge commit message. However, often this needs fixing up or changing, so again, will probably increase the label management work, not reduce it. Also, the title would need to be checked after creation and before merge. If you have a good idea how this could work, please feel free to comment, but I'm not very optimistic this can work well and/or reliably.

danceratopz:
Thanks again for this @mandeepmourya007, but due to the comments noted above in the review, the other priorities in this repo and the large number of PRs we have, I will close this one.

PR #2365, opened 2026-02-28
feat: Accept RLP encoded legacy transactions in `decode_transaction`

kevaundray:
## Problem

The ExecutionPayload is an object the the execution layer receives from the consensus layer. This object has a transactions field, which is a list of RLP encoded transactions. 

Currently, it is not possible to decode this list because `decode_transaction` assumes that all legacy transactions are python objects.

## Suggested Solution

Modify `decode_transaction` to also decode legacy transactions that have been rlp encoded.
 
## Underlying problem

Ideally we have something like `encode(tx : Transaction) -> Bytes` and `decode(bytes : Bytes) -> Transaction`. 

However, `LegacyTransaction` and `TypedTransaction` are encoded differently, and there is no way to tell this to the RLP library; legacy transactions are encoded are lists, while typed transactions are encoded as strings.

_List[Transaction]_ 

This wouldn't work because all transaction types would be encoded as RLP lists. This is fine for legacy transactions, however it breaks for typed transactions since they should be encoded as strings with a type prefix.

_List[Bytes]_

This also doesn't work because legacy transactions would be seen as strings as opposed to a list. `rlp.encode(block)` would give a different result in this case.

## Ideal solution

The ideal solution would be something like:

```python
transactions: Tuple[Annotated[Transaction, With(decode_tx, encode_tx)], ...]
```

ie we provide methods that specify exactly how to encode and decode the dataclass overriding the default method based on types.

## 🔗 Related Issues or PRs

N/A.

#2362 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


tests/json_infra/test_transaction_codec.py
```diff
@@ -0,0 +1,30 @@
...
+        value=U256(0),
+        data=Bytes(b""),
+        v=U256(27),
+        r=U256(1),
+        s=U256(2),
+    )
+    encoded = rlp.encode(tx)
+    assert encoded[0] >= 0xC0
+    assert decode_transaction(encoded) == tx
+    assert decode_transaction(tx) is tx
```
kevaundray, line 30: small test to check that whether we give the method Bytes or LegacyTransaction, we get the same result

src/ethereum/forks/amsterdam/transactions.py
```diff
@@ -520,6 +524,9 @@ def decode_transaction(tx: LegacyTransaction | Bytes) -> Transaction:
             return rlp.decode_to(BlobTransaction, tx[1:])
         elif tx[0] == 4:
             return rlp.decode_to(SetCodeTransaction, tx[1:])
+        elif tx[0] >= 0xC0:
+            assert tx[0] <= 0xFE
+            return rlp.decode_to(LegacyTransaction, tx)
```
kevaundray, line 529: can check the backwards compatibility section in 2718

tests/json_infra/test_transaction_codec.py
```diff
@@ -0,0 +1,30 @@
+"""Test that decode_transaction handles legacy transactions as bytes."""
```
kevaundray, line 1: Note: If this change gets backported, then this files only tests amsterdam

kevaundray:
CI errors seem unrelated since it mentions BlockAccessListGasLimitExceededError and this PR doesn't change BAL (it should be a loop in terms of previous tests)

gurukamath (approved):

PR #2369, opened 2026-03-01
refactor(json_loader): update json_loader tests to use the `fill-then-consume` workflow

### LINKED ISSUE #2090: Speed-up JSON infra by filling a subset of test and executing it

gurukamath:
## 🗒️ Description

Replace the download-only json_infra CI flow with a fill-then-consume pipeline.

This PR creates 3 modes for `json_infra` to run based on what the source of the json fixtures is (can be set using the `--fixture-source` cli parameter)
1. `fill` (default): fill the json_infra-marked tests and consume the generated fixtures. This is the default mode and will be run in CI
2. `download`: download the external fixture sets defined in `tests/json_infra/__init__.py` (current behaviour, very slow).
3. `local`: run against whatever fixtures are already present in `tests/json_infra/fixtures` without any setup (new functionality - could be useful for ad-hoc runs).

The minimal set of tests required to get the same coverage have been identified using the `minimize-tests` tool (See [here](https://github.com/SamWilsn/minimize-tests))

## 🔗 Related Issues or PRs

#2090

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Cute Animals - 9 of 12


danceratopz (commented):
Hey @gurukamath, thanks a lot for getting on this! 

However, I think we can create the desired flow in a more maintainable manner. What do you think about this:

1. Create a CI flow "Minimize Tests" that runs as a cron job weekly/monthly (or workflow dispatch). This flow:   i. `fill --cov...` → `.coverage`
   ii. `minimize-tests .coverage` → `minimized-tests.txt`  # is this how minize-tests works?
   iii. `fill @minimized-tests.txt → `minimized-fixtures.tar.gz`
   iv. Upload both as artifacts (`minimized-tests.txt` + `minimized-fixtures.tar.gz`)

2. Per-PR json_infra job:
   i. Download `minimized-fixtures.tar.gz` from latest (weekly/montly/scheduled) run.
   ii. Extract into `tests/json_infra/fixtures/`.
   iii. Update `tox -e json_infra` to run with this input (`--fixture-source=local`).

I would strongly favor the above approach, as:

1. Calling `fill` (pytest test suite) from another pytest test suite seems a little convoluted, and,
2. This avoids having to mark each test module source with the required marker (to indicate what the minimal test set is), which is not maintainable in the long term, imo.

Also I think the `json_infra` marker might be applied too liberally resulting in a larger test set than intended (perhaps intentionally?). I assume minimize-tests generates a list of pytest nodeids (each one being a parametrized test case), not of test functions. Currently, the whole test function is decorated with the `pytest.mark.json_infra` marker which, if I understand correctly,  would fill and run many more test cases than provided by the output of minimize-tests.

danceratopz:
> 1. iii. `fill @minimized-tests.txt → `minimized-fixtures.tar.gz` iv. Upload both as artifacts (`minimized-tests.txt`+`minimized-fixtures.tar.gz`)

We could also consider providing this is a release instead of an artifact to be readily downloaded by client teams to be as a very fast smoke test.

danceratopz:
Btw, @gurukamath, sorry, I'd missed #2090, or would have added this input there earlier.

Also, I only recently discovered that pytest natively supports running test cases from a file containing a list of nodeids via an filename prefixed by `@`, so we don't need to add this functionality to `fill` to run with the minimized test set, we can do:
```
fill @minimized-tests.txt
```
I.e., this means that we don't need #1807 (for `fill` or `pytest tests/json_infra`) and instead use this native pytest feature instead.

We can also add any required arguments to this file (e.g., `--until Amsterdam`).

This is new in 8.2, search for "Read arguments from file" here https://docs.pytest.org/en/stable/how-to/usage.html#specifying-which-tests-to-run.

danceratopz:
However, the suggestion above about using a "Minimize Tests" workflow for a `minimized-fixtures.tar.gz` might not be very robust if the "Minimize Tests" and the "Per-PR json_infra"  don't run on/target the same branch?

But either way, it's probably better to only generate the `minimized-fixtures.txt` in the "Minimize Tests" workflow for json_infra, as trying to run fixtures in a per-PR job that were generated from a different version is expected to break if the fixture formats change.

So perhaps it'd better to keep a split like:
1. Minimize Tests Workflow generates a list of tests (perhaps it fills, but not for json_infra, but rather to create a smoke test set for client teams).
2. Per-PR json_infra job runs `fill @minimized-tests.txt` and then runs json_infra.

marioevz:
> However, the suggestion above about using a "Minimize Tests" workflow for a `minimized-fixtures.tar.gz` might not be very robust if the "Minimize Tests" and the "Per-PR json_infra" don't run on/target the same branch?
> 
> But either way, it's probably better to only generate the `minimized-fixtures.txt` in the "Minimize Tests" workflow for json_infra, as trying to run fixtures in a per-PR job that were generated from a different version is expected to break if the fixture formats change.
> 
> So perhaps it'd better to keep a split like:
> 
>     1. Minimize Tests Workflow generates a list of tests (perhaps it fills, but not for json_infra, but rather to create a smoke test set for client teams).
> 
>     2. Per-PR json_infra job runs `fill @minimized-tests.txt` and then runs json_infra.

I have not reviewed the PR, but I wanted to chime in on this idea to mention that I don't like us relying again on a file that is out of sync in any way with the current PR, which is basically the same problem we are trying to solve with this PR in the first place. Why don't we just mark a subset of tests as `pytest.mark.json_infra` and run those tests?

gurukamath:
> 2. Per-PR json_infra job:
>    i. Download `minimized-fixtures.tar.gz` from latest (weekly/montly/scheduled) run.
>    ii. Extract into `tests/json_infra/fixtures/`.
>    iii. Update `tox -e json_infra` to run with this input (`--fixture-source=local`).
> 
Feels like this could introduce another source for things to go out of sync. We want to keep the filling of the json_infra tests as closely coupled with the json consumption CI as possible.

> 1. Calling `fill` (pytest test suite) from another pytest test suite seems a little convoluted, and,
> 2. This avoids having to mark each test module source with the required marker (to indicate what the minimal test set is), which is not maintainable in the long term, imo.
>

Personally, I don't see this test set changing drastically over time. In fact it should not change at all for older forks. For the newest fork, we can possibly just update the test set once the hardfork goes live on mainnet. So, I see a minimal maintenance overhead.

> Also I think the `json_infra` marker might be applied too liberally resulting in a larger test set than intended (perhaps intentionally?). I assume minimize-tests generates a list of pytest nodeids (each one being a parametrized test case), not of test functions. Currently, the whole test function is decorated with the `pytest.mark.json_infra` marker which, if I understand correctly, would fill and run many more test cases than provided by the output of minimize-tests.

I just decided to mark the test function instead of the individual node ID because of the simplicity of design. Plus we are not looking at that level of granularity anyway. With this set-up, the overall CI job takes about 10 mins which in my opinion is well within what we want.

gurukamath:
> 1. Minimize Tests Workflow generates a list of tests (perhaps it fills, but not for json_infra, but rather to create a smoke test set for client teams).

Generating the list of tests does take a significant amount of time, since the search space for the optimal combination of tests is large. I would recommend not making it part of the CI.

gurukamath:
> I have not reviewed the PR, but I wanted to chime in on this idea to mention that I don't like us relying again on a file that is out of sync in any way with the current PR, which is basically the same problem we are trying to solve with this PR in the first place. Why don't we just mark a subset of tests as `pytest.mark.json_infra` and run those tests?

That's essentially what the PR does

tox.ini
```diff
@@ -80,7 +80,9 @@ commands =
         packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_benchmarking.py
 
 [testenv:json_infra]
-description = Run the spec against released test fixtures
+description = Run the spec against test fixtures
+dependency_groups =
+    test
 commands =
     pytest \
```
danceratopz: I think I'd favor calling `fill` directly here instead of calling it as a subprocess from `pytest`. Keeps the json infra `pytest` simpler and keeps things transparent. 

This keep the fixtures output directory out of the (direct) user space.
```suggestion
    fill \
        -m json_infra \
        --until Amsterdam \
        # loadgroup: serialize xdist_group("bigmem") tests onto one worker to limit peak memory
        -n {env:PYTEST_XDIST_AUTO_NUM_WORKERS:6} --dist=loadgroup \
        --skip-index \
        --no-html \
        --clean \
        --output="{toxworkdir}/fixtures-json-infra/"
    pytest \
```

tests/json_infra/conftest.py
```diff
@@ -333,32 +349,70 @@ def pytest_sessionstart(session: Session) -> None:
...
+                "-m",
+                "json_infra and not slow and not benchmark"
+                " and not derived_test",
+                "-n",
+                str(os.cpu_count() or 6),
+                "--dist=loadgroup",
+                "--skip-index",
+                "--clean",
+                "--until",
+                "Amsterdam",
```
danceratopz: Hardcoding/hiding the fork name here is one example of why I'd just let the user (or generally `tox`) call `fill` directly.

But also the CPU count above.

tests/json_infra/conftest.py
```diff
@@ -333,32 +349,70 @@ def pytest_sessionstart(session: Session) -> None:
...
+    fixture_source = session.config.getoption("fixture_source")
+
+    if fixture_source == "fill":
+        shutil.rmtree(fixtures_dir, ignore_errors=True)
+        output_dir = fixtures_dir / "locally_filled"
+        subprocess.run(
+            [
+                "fill",
+                "-m",
+                "json_infra and not slow and not benchmark"
```
danceratopz: Not 💯 whether we should apply "not slow", if the priority is coverage (and we're fast enough)?

tests/json_infra/conftest.py
```diff
@@ -118,6 +119,21 @@ def pytest_addoption(parser: Parser) -> None:
...
+        dest="fixture_source",
+        default="fill",
+        choices=["fill", "download", "local"],
+        help=(
+            "Controls how test fixtures are obtained before running. "
+            "'fill' (default): fill the json_infra-marked tests and consume "
+            "the generated fixtures. "
+            "'download': download external fixture sets (legacy behaviour). "
+            "'local': run against whatever fixtures are already present in "
+            "tests/json_infra/fixtures without any setup."
```
danceratopz: Ok, either this would need a change here to grab the fixtures from `"{toxworkdir}/fixtures-json-infra/"` or if you prefer you can write the output of `fill` to `tests/json_infra/fixtures` in `tox.ini`. I would lean to an explicit input path here for the fixture directory to allow more flexible testing with json infra.

danceratopz (commented):
@gurukamath and @marioevz, thanks for your replies! 

I think I over-indexed on the maintability, thinking we'd have to update this list often. The priority is clearly a stable flow and I think marking the tests in source directly is the only way to achieve this ("single source").

As such, this looks good to me, just one comment below. See if you agree.

tests/json_infra/conftest.py
gurukamath: Good point. I can perhaps thread the inputs from the tox command down into this (have a single source)

tox.ini
gurukamath: I did consider this. The reason why I went with the sub-process is that it allowed me to better control what mode I wanted to run `json_infra` in (without introducing conditionals and cluttering up the `tox.ini` file). The two questions we need to answer here are

1. Do we still need `json_infra` to have all 3 modes available? `fill-then-run`, `download-then-run`, `local-run` or do we just have `fill-then-run`. I lean towards the former.
2. If we decide to go with just one mode, I'm happy to just introduce `fill` in `tox.ini`. However, if we decide that 3-modes makes sense, I'd rather keep those conditionals out of `tox.ini`.

@danceratopz Would love to hear your thoughts on this.

danceratopz: Hey Guru, thanks for the question. I think this is an important PR and we should try to get it in quickly. I’ll just add a bit of background on how I see this.

Post-Weld, we still have some duplication in our tooling. The goals of JSON Infra, [`consume cache`](https://github.com/ethereum/execution-specs/blob/a0f6c80f22e27670f1eeb5899bfec3c2d90fe04f/docs/running_tests/consume/cache.md)/`consume direct` are quite similar. For example, `consume cache` can take a fixture release tag, download it, and unpack it ready for use. It can run as a standalone command, or composed as a plugin in another `consume` command like `consume rlp`, e.g., `consume rlp --input=bal@v5.2.0`.

Longer term, I’d like to try and unify these interfaces and reuse as much of the same machinery as possible. Ideally, we would use the same infrastructure we use to test clients to also test the specs, or something very close to it.

Regarding modes, one possible model would be for JSON Infra to have a single mode that just consumes local fixtures. In that workflow you would run either `fill` or `consume cache` first to obtain the fixtures, and then JSON Infra would operate on those locally available inputs. But that would blow up the scope of this PR.

I don’t have particularly strong feelings about how many modes JSON Infra should expose now, because it's not particularly public facing and long term I think a good approach would be to test the specs via `consume direct` (but perhaps I'm missing something).

Regarding the subprocess approach: personally I think the one extra line in `tox.ini` isn’t much clutter, is more explicit and avoids additional complexity in json_infra. But if you prefer the subprocess approach, then I won't block this PR.

tests/json_infra/conftest.py
danceratopz: Possible, but adds more complexity. Other args are candidates for this too, (`-m`, -`n`), if not all args. 

Btw, we do lots of sneaky things and have this complexity in the click CLI here, https://github.com/ethereum/execution-specs/tree/a0f6c80f22e27670f1eeb5899bfec3c2d90fe04f/packages/testing/src/execution_testing/cli/pytest_commands.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
```diff
@@ -1985,6 +1985,7 @@ def test_bal_multiple_storage_writes_same_slot(
         pytest.param([2, 3, 4], id="depth_3"),
     ],
 )
+@pytest.mark.json_infra
```
SamWilsn: I'd like to drop the "infra" part, and this seems like a good place to do it. How about:

- `json_loader`
- `interesting_json`
- `eels_t8n` (not sure this is accurate really)
- `smoke`
- `minimal`
- `minimized`
- `coverage`

I guess it really depends on what these tests _are_. If they're the straight output of minimize-tests, the last four options above are accurate. If they're chosen more specifically for the JSON tooling, then one of the first three might be more applicable.

tests/json_infra/conftest.py
SamWilsn: Personally, I think we can drop everything other than "fill".

These are our internal tests, just to make sure EELS's tooling can read and execute JSON fixtures.

tests/json_infra/conftest.py
SamWilsn: Can we make more use of pytest here maybe? Kinda a dumb idea, but maybe implement a [`pytest_pyfunc_call`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_pyfunc_call) hook and call the json_infra "entry point" immediately after filling each fixture?

tests/json_infra/test_ethash.py
```diff
@@ -30,6 +31,13 @@
 
 ETHEREUM_TESTS_PATH = TEST_FIXTURES["ethereum_tests"]["fixture_path"]
 
+if not Path(ETHEREUM_TESTS_PATH).exists():
+    pytest.skip(
+        "ethereum_tests fixture set not available; "
+        "run with --fixture-source=download to populate it",
+        allow_module_level=True,
+    )
```
SamWilsn: Skipped tests don't really produce any output, do they? So most people won't know these tests aren't running.

We should make them work or remove them, IMO.

tox.ini
SamWilsn: Can we add `--cov-fail-under=...` and a comment linking somewhere that explains how to regenerate the minimal test set? Unless CI fails, it'll never get updated.

tests/json_infra/test_ethash.py
gurukamath: These depend on `ethereum/tests`. So if we are to only do `fill-and-run`, we won't be downloading `ethereum/tests`. So we can just get rid of these tests imo. An alternative is to include only the relevant json fixtures as static files in our `tests/json_infra` location. I'd prefer however, to simplify these by just getting rid of them

tox.ini
gurukamath: Based on Sam's comment, I have updated this to only have the `fill` mode and to include the `fill` command explicitly in `tox.ini`

tests/json_infra/conftest.py
gurukamath: Updated to drop everything other than `fill`

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
gurukamath: I think these are aimed at ensuring the json fixtures are loaded properly. The fact that the test set is minimized is imo, an implementation detail aimed to reduce CI time. From that perspective `json_loader` makes sense to me. Have updated accordingly.

tests/json_infra/test_ethash.py
gurukamath: Have updated to get rid of these tests

tox.ini
gurukamath: Done. And set the fail threshold to 70

Edit: Updated to set the threshold on the `fill` part of the command (for an apples-to-apples comparison with `py3`) and set the threshold at 85%

SamWilsn (approved):

PR #2370, opened 2026-03-01
refactor(spec-specs,test-types): replace `pycryptodome` with `hashlib` for keccak

kevaundray:
## 🗒️ Description

Both expose keccak, however I found hashlib to be around 60% faster on my laptop 

Extracted from #2368 

#### Benchmark of `fill tests/osaka`, hashlib vs pycryptodome 

Using ad1440bb89e.

Command (per row, only `--python` differs):

```
taskset -c 0,2,4,5,6,7 uv run --python <interp> \
    fill tests/osaka --output=<dir> --clean -x --no-html --skip-index -n=6
```

Host: Ubuntu (kernel `6.11.0-1017-oem`), x86_64. `n=10` runs per configuration.

| Python                  | Source         | OpenSSL (linked)     | Keccak backend          | Mean ± σ (s)     | Range (s)        |
| ----------------------- | -------------- | -------------------- | ----------------------- | ---------------- | ---------------- |
| CPython 3.12.12         | uv-managed     | 3.5.5 (27 Jan 2026)  | `hashlib`               | 69.27 ± 6.74     | 58.04 – 76.62    |
| CPython 3.12.3          | Ubuntu system  | 3.0.13 (30 Jan 2024) | `pycryptodome` fallback | 74.08 ± 6.47     | 65.95 – 81.54    |

**Delta:** 4.81 s (≈ 6.5 %) faster on the hashlib path. Welch t ≈ 1.63 (df ≈ 18), one-tailed p ≈ 0.06 — directionally consistent with prior runs, just shy of p < 0.05 two-tailed at this sample size.

The Ubuntu-system interpreter is linked against OpenSSL 3.0.13, which predates the OpenSSL 3.2.0 default-provider Keccak addition (Nov 2023), so `hashlib.new("keccak-256", ...)` raises `ValueError` and the module dispatches to `pycryptodome` instead. The uv-managed `cpython-3.12.12` from `python-build-standalone` bundles OpenSSL 3.5.5, which exposes Keccak via the default provider, so the fast `hashlib` path is taken.

image

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

:rabbit:

### DIFF
--- a/packages/testing/src/execution_testing/base_types/tests/test_keccak_dispatch.py
+++ b/packages/testing/src/execution_testing/base_types/tests/test_keccak_dispatch.py
@@ -124,7 +124,7 @@ def test_fallback_engages_when_hashlib_lacks_keccak(
     with patch.object(hashlib, "new", side_effect=mocked_new):
         h = _clean_reimport_hash()
 
-        assert hasattr(h, "_pycryptodome_keccak"), (
+        assert h._USE_HASHLIB is False, (
             "module did not engage pycryptodome fallback"
         )
         for buffer, expected_hex in KECCAK256_VECTORS:
@@ -148,7 +148,7 @@ def test_native_path_used_when_hashlib_has_keccak(
         pytest.skip("hashlib lacks keccak-256 on this OpenSSL build")
 
     h = _clean_reimport_hash()
-    assert not hasattr(h, "_pycryptodome_keccak"), (
+    assert h._USE_HASHLIB is True, (
         "module engaged pycryptodome fallback despite hashlib having keccak"
     )
 
--- a/packages/testing/src/execution_testing/base_types/tests/test_keccak_dispatch.py
+++ b/packages/testing/src/execution_testing/base_types/tests/test_keccak_dispatch.py
@@ -16,7 +16,6 @@ succeeds. These tests verify:
 
 import hashlib
 import importlib
-import sys
 from collections.abc import Iterator
 from typing import Any
 from unittest.mock import patch
@@ -61,8 +60,8 @@ def _hashlib_has_keccak() -> bool:
 
 def _clean_reimport_hash() -> Any:
     """Drop and reimport `ethereum.crypto.hash` for a fresh dispatch run."""
-    sys.modules.pop("ethereum.crypto.hash", None)
-    return importlib.import_module("ethereum.crypto.hash")
+    mod = importlib.import_module("ethereum.crypto.hash")
+    return importlib.reload(mod)
 
 
 @pytest.fixture
--- a/packages/testing/src/execution_testing/base_types/tests/test_keccak_dispatch.py
+++ b/packages/testing/src/execution_testing/base_types/tests/test_keccak_dispatch.py
@@ -1,0 +1,175 @@
+"""
+Tests for the Keccak backend dispatch in `ethereum.crypto.hash`.
+
+The module decides at import time whether to use hashlib (linked OpenSSL)
+or pycryptodome, depending on whether `hashlib.new("keccak-256", ...)`
+succeeds. These tests verify:
+
+* both backends produce byte-identical output (cross-backend equivalence);
+* the fallback engages cleanly when hashlib raises, simulated via
+  monkeypatch so we can exercise the path on a Python whose OpenSSL does
+  expose Keccak;
+* on a Python where hashlib supports Keccak, the fast path is selected
+  (guards against a regression where `algorithms_available` lies and the
+  module silently forces every user onto pycryptodome).
+"""
+
+import hashlib
+import importlib
+import sys
+from collections.abc import Iterator
+from typing import Any
+from unittest.mock import patch
+
+import pytest
+
+# Pre-NIST Keccak vectors. Empty-input digests are widely published; the
+# `hashme` vector was confirmed against a working hashlib build during
+# PR #2370 review.
+KECCAK256_VECTORS: list[tuple[bytes, str]] = [
+    (
+        b"",
+        "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470",
+    ),
+    (
+        b"abc",
+        "4e03657aea45a94fc7d47ba826c8d667c0d1e6e33a64a036ec44f58fa12d6c45",
+    ),
+    (
+        b"hashme",
+        "7f98885dc9cf152c0bb08eaf056668f99c47cabd8fe01b1276f9a305b1389646",
+    ),
+]
+
+KECCAK512_VECTORS: list[tuple[bytes, str]] = [
+    (
+        b"",
+        "0eab42de4c3ceb9235fc91acffe746b29c29a8c366b7c60e4e67c466f36a4304"
+        "c00fa9caf9d87976ba469bcbe06713b435f091ef2769fb160cdab33d3670680e",
+    ),
+]
+
+
+def _hashlib_has_keccak() -> bool:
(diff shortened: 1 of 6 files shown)


marioevz:
Seems like this is only compatible with python 3.12+. Tagging @danceratopz for follow up.

danceratopz (commented):
Thanks @kevaundray! More speed-up always good and as `hashlib` is in the standard library, this seems like a worthwhile change. 

I'm assuming we need to keep `pycrptodome` as a dep for other crypto functions? And that's why you didn't update pyproject.toml? 
https://github.com/ethereum/execution-specs/blob/4e768ab6987eb6fd772dd9b88383d11359b26dff/src/ethereum/crypto/elliptic_curve.py#L4

Happy to limit the change to replacing the keccak implementation though.

From @marioevz 
> Seems like this is only compatible with python 3.12+. Tagging @danceratopz for follow up.

I think it depends on the version of OPENSSL that your Python is linked against, or in the case of a `uv`-managed Python, bundled with (the [python-build-standalone] project, now also maintained by astral, (https://github.com/astral-sh/python-build-standalone) statically links these deps).

I did a quick test of `uv`-managed Pythons: Initially, this failed locally for me with 3.11, 3.12. Then I upgraded my uv-managed 3.11 and 3.12 versions and it worked across all of them.

E.g,
```
uv python upgrade 3.11
Installed Python 3.11.14 in 4.11s
 + cpython-3.11.14-linux-x86_64-gnu

uv run --python 3.11 python -c 'import ssl, sys, hashlib; print(sys.version, ssl.OPENSSL_VERSION); h=hashlib.new("keccak-256", b"hashme"); print(h.hexdigest())'
3.11.14 (main, Feb 12 2026, 00:42:50) [Clang 21.1.4 ] OpenSSL 3.5.5 27 Jan 2026
7f98885dc9cf152c0bb08eaf056668f99c47cabd8fe01b1276f9a305b1389646
```

kevaundray:
> Thanks @kevaundray! More speed-up always good and as `hashlib` is in the standard library, this seems like a worthwhile change.
> 
> I'm assuming we need to keep `pycrptodome` as a dep for other crypto functions? And that's why you didn't update pyproject.toml?
> 
> https://github.com/ethereum/execution-specs/blob/4e768ab6987eb6fd772dd9b88383d11359b26dff/src/ethereum/crypto/elliptic_curve.py#L4
> 
> Happy to limit the change to replacing the keccak implementation though.
> 
> From @marioevz
> 
> > Seems like this is only compatible with python 3.12+. Tagging @danceratopz for follow up.
> 
> I think it depends on the version of OPENSSL that your Python is linked against, or in the case of a `uv`-managed Python, bundled with (the [python-build-standalone] project, now also maintained by astral, (https://github.com/astral-sh/python-build-standalone) statically links these deps).
> 
> I did a quick test of `uv`-managed Pythons: Initially, this failed locally for me with 3.11, 3.12. Then I upgraded my uv-managed 3.11 and 3.12 versions and it worked across all of them.
> 
> E.g,
> 
> ```
> uv python upgrade 3.11
> Installed Python 3.11.14 in 4.11s
>  + cpython-3.11.14-linux-x86_64-gnu
> 
> uv run --python 3.11 python -c 'import ssl, sys, hashlib; print(sys.version, ssl.OPENSSL_VERSION); h=hashlib.new("keccak-256", b"hashme"); print(h.hexdigest())'
> 3.11.14 (main, Feb 12 2026, 00:42:50) [Clang 21.1.4 ] OpenSSL 3.5.5 27 Jan 2026
> 7f98885dc9cf152c0bb08eaf056668f99c47cabd8fe01b1276f9a305b1389646
> ```

Happy to keep it as is, if it may give compilation errors :) (I'm not that familiar with uv-managed)

> I'm assuming we need to keep pycrptodome as a dep for other crypto functions? And that's why you didn't update pyproject.toml?

Ah I was only focussed on keccak since that was where I was measuring (Title is misleading in that case)

src/ethereum/crypto/hash.py
SamWilsn, line 21: Is it possible for keccak-256 or keccak-512 to be disabled in OpenSSL? If so, should we do something like:

```python
if "keccak-256" in hashlib.algorithms_available:
    ...
else:
    ...
```

kevaundray: I guess I would lean towards not adding hashlib since it violates the "it just works" and KISS motto

kevaundray: Or atleast to my eyes, this part from Dan's message:

> depends on the version of OPENSSL that your Python is linked against

Is a quirk that I think is not worth it, given we are in the python ecosystem

SamWilsn: I think optimistically using OpenSSL would be fine, especially if we're keeping pycrptodome around for other reasons.

danceratopz: I agree, it's not much boilerplate code. But I wanted to see whether it was worth it, by checking how much of the performance gain would propagate to `fill`. TLDR: ~6.5% speed-up. Not huge but still worth it imo. More details in the updated PR description.

danceratopz: Unfortunately, `hashlib.algorithms_available` doesn't return `"keccak-256"` even when it is available, so I went for a `try-except`.

danceratopz (commented):
This looks good to me and think it's worth including. As we only use uv-managed Pythons in CI, this speed-up (~7%; see updated description for details) will be visible across all CI jobs that fill. I don't have a strong opinion though. I would just like to get resolution on this PR 🙂 

@SamWilsn, could you please re-review the recent changes which added the fallback to pycryptodome you requested?

packages/testing/src/execution_testing/base_types/tests/test_keccak_dispatch.py
```diff
@@ -0,0 +1,175 @@
...
+        hashlib.new("keccak-256")
+    except ValueError:
+        return False
+    return True
+
+
+def _clean_reimport_hash() -> Any:
+    """Drop and reimport `ethereum.crypto.hash` for a fresh dispatch run."""
+    sys.modules.pop("ethereum.crypto.hash", None)
+    return importlib.import_module("ethereum.crypto.hash")
```
SamWilsn:
```suggestion
    mod =  importlib.import_module("ethereum.crypto.hash")
    return importlib.reload(mod)
```

I think this would make me more comfortable?

src/ethereum/crypto/hash.py
```diff
@@ -11,14 +11,47 @@
...
+def _keccak256_digest(buffer: bytes | bytearray) -> bytes:
+    if _USE_HASHLIB:
+        return hashlib.new("keccak-256", buffer).digest()
+    return _pycryptodome_keccak.new(digest_bits=256).update(buffer).digest()
+
+
+def _keccak512_digest(buffer: bytes | bytearray) -> bytes:
+    if _USE_HASHLIB:
+        return hashlib.new("keccak-512", buffer).digest()
+    return _pycryptodome_keccak.new(digest_bits=512).update(buffer).digest()
```
SamWilsn: Should we invert these? For example:

```python
if _USE_HASHLIB:
    def _keccak256_digest(buffer: bytes | bytearray) -> bytes:
        return hashlib.new("keccak-256", buffer).digest()
else:
    def _keccak256_digest(buffer: bytes | bytearray) -> bytes:
        return _pycryptodome_keccak.new(digest_bits=256).update(buffer).digest()
```

That way there's no conditional on every function call.

SamWilsn (approved):

PR #2373, opened 2026-03-01
feat(crypto): Replace `py-ecc` with `py-arkworks-bls12381` for EIP-2537 precompiles and EIP-4844 precompile

kevaundray:
## Problem

py-ecc is a pure python implementation for elliptic curve cryptography, which means that its very slow. I benchmarked some of the diff with py-arkworks-bls12381 (a rust backed python implementation) and the its atleast 100x slow for some operations like pairings. 

## Solution

py-arkworks-bls12381 has been recently updated to have the interface needed for the bls12-381 precompiles. Before that it was being used for EIP-4844 in the consensus specs. So its possible to use it for the precompiles in tests.

Link: https://pypi.org/project/py_arkworks_bls12381/

Currently in the process of transferring it over to the EF github org

## Changes

- I added a bls12381.py file in crypto which means that one can switch out the implementation in the future and the rest of the code would not change. The previous code was mixing py-ecc types and operations with EIP specific semantics like padding.

- Updated kzg.py to use py-arkworks-bls12381

- Updated amsterdam and the other forks to use the byte interface.

## Note

- py-ecc is still being used for bls12-381 test generation, updating that is a bit more difficult because it uses an isomorphic curve and probably left to a future PR (if thats what folks want)

- py-ecc is also still being used for alt-bn128 precompiles. py-arkworks-bls12381 would not be able to replace that since its a different curve.

Extracted from #2368 

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,211 @@
...
+
+.. contents:: Table of Contents
+    :backlinks: none
+    :local:
+
+Introduction
+------------
+
+Byte-level interface to BLS12-381 curve operations. All inputs and
+outputs are raw (unpadded) bytes.
```
SamWilsn: Use markdown-formatted docstrings in new files.

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,211 @@
...
+"""
+
+from typing import Sequence
+
+from py_arkworks_bls12381 import GT, G1Point, G2Point, Scalar
+
+
+def g1_add(a: bytes, b: bytes) -> bytes:
```
SamWilsn:
```suggestion
def g1_add(a: Bytes96, b: Bytes96) -> Bytes96:
```

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,211 @@
...
+    result : bytes
+        96-byte G1 point.
+
+    """
+    p1 = G1Point.from_xy_bytes_unchecked_be(a)
+    p2 = G1Point.from_xy_bytes_unchecked_be(b)
+    return (p1 + p2).to_xy_bytes_be()
+
+
+def g2_add(a: bytes, b: bytes) -> bytes:
```
SamWilsn, line 26: Worth adding a `Bytes192` type?

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,211 @@
...
+
+    Raises
+    ------
+    ValueError
+        If a point is not on the curve or fails the subgroup check.
+
+    """
+    g1s = []
+    for p in points:
+        point = G1Point.from_xy_bytes_unchecked_be(p)
```
SamWilsn, line 54: On the topic of stricter typing, should we separate the bytes -> numeric conversion from the cryptography?

How the integers are encoded feels more like an "EVM thing" and less like a "crypto" thing.

In practice, I think this would mean changing the function parameters from `bytes` to an integer type.

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,211 @@
...
+        if not point.is_in_subgroup():
+            raise ValueError("Subgroup check failed for G1 point.")
+        g1s.append(point)
+    scs = [Scalar.from_be_bytes_mod_order(s) for s in scalars]
```
SamWilsn: `scs` is not a common enough abbreviation.

src/ethereum/crypto/bls12_381.py
SamWilsn: Though if this is a performance thing (because the library doesn't handle Python `int`), then I guess we can keep this.

src/ethereum/crypto/kzg.py
```diff
@@ -141,36 +62,20 @@ def verify_kzg_proof(
...
     assert len(proof_bytes) == BYTES_PER_PROOF
 
-    return verify_kzg_proof_impl(
-        bytes_to_kzg_commitment(commitment_bytes),
-        bytes_to_bls_field(z_bytes),
-        bytes_to_bls_field(y_bytes),
-        bytes_to_kzg_proof(proof_bytes),
-    )
+    # Validate and deserialize G1 points
+    # Note: Points must be in the prime-ordered subgroup
```
SamWilsn: How is this comment meant to be interpreted? Is it "the underlying library will raise an exception" or "the underlying library will misbehave"?

In either case, the specs are the only "user" of this file, right? So this is an instruction to ourselves. Instead of commenting this, we should implement it.

SamWilsn (commented):
Partial review I forgot to submit ^^;;

src/ethereum/crypto/bls12_381.py
kevaundray: Makes sense!

src/ethereum/crypto/bls12_381.py
kevaundray: This would require a new type in ethereum-types -- I think it would make sense though I don't know what the overhead is for ethereum-tyes

src/ethereum/crypto/bls12_381.py
kevaundray: I definitely agree, though yep as you mentioned, for performance the underlying library accepts bytes, so we would be going from bytes to int to bytes

src/ethereum/crypto/kzg.py
kevaundray: Ah yep, I made the underlying library raise here. I've updated comment

PR #2378, opened 2026-03-02
chore(specs, tests): restore spilled state gas to reservoir on revert/halt

spencer-tb:
## 🗒️ Description

When a child frame's state gas spills from the reservoir into `gas_left` and then reverts or halts, the spilled state gas was lost. Since state changes are rolled back, all state gas (reservoir + spill) should be restored to the parent's reservoir, not just the original reservoir.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/__init__.py
+++ b/src/ethereum/forks/amsterdam/vm/__init__.py
@@ -202,14 +202,14 @@ def incorporate_child_on_success(evm: Evm, child_evm: Evm) -> None:
 def incorporate_child_on_error(
     evm: Evm,
     child_evm: Evm,
-    child_state_gas_reservoir: Uint,
 ) -> None:
     """
     Incorporate the state of an unsuccessful `child_evm` into the parent `evm`.
 
     On failure (revert or exceptional halt) state changes are rolled back,
-    so no state was actually grown.  The full original reservoir is restored
-    to the parent and the child's state_gas_used is not accumulated.
+    so no state was actually grown.  All state gas, both reservoir and any
+    that spilled into `gas_left`, is restored to the parent's reservoir and
+    the child's `state_gas_used` is not accumulated.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


src/ethereum/forks/amsterdam/vm/__init__.py
```diff
@@ -202,25 +202,23 @@ def incorporate_child_on_success(evm: Evm, child_evm: Evm) -> None:
...
     ----------
     evm :
         The parent `EVM`.
     child_evm :
         The child evm to incorporate.
-    child_state_gas_reservoir :
-        The original state gas reservoir forwarded to the child frame.
 
     """
     evm.gas_left += child_evm.gas_left
```
gurukamath, line 222: Should `evm.gas_left` be renamed to something else just so we know how this differs from `evm.state_gas_left`?

spencer-tb: The reason I kept it was to prevent back porting but happy to change. `evm.compute_gas_left`?

gurukamath (approved):

jwasinger:
Isn't the state gas spent by a frame not refunded in the case of revert? From the EIP:
> State gas charged for account creation (CREATE, CALL to new account, and EOA delegation) is consumed even if the frame reverts — state changes are rolled back but gas is not refunded.

I'm guessing that this also applies to SSTORE as well:
> Refunds use refund_counter rather than direct gas accounting decrements, so that reverted frames do not benefit from the refund.

spencer-tb:
Thx @jwasinger. Comments below!

These two EIP quotes are about different items compared to what the PR changes. What this PR changes:
- When a child frame's state gas spills from the reservoir into `gas_left` and then reverts/halts, the spilled portion was previously lost. The update restores `child_evm.state_gas_used + child_evm.state_gas_left` (the full forwarded amount) instead of only the original reservoir.
- This now matches the updated EIP: "On child revert or exceptional halt, all state gas consumed by the child, both from the reservoir and any that spilled into gas_left, is restored to the parent's reservoir."

PR #2384, opened 2026-03-02
feat(spec): Abstract py-ecc implementation detail from EIP-2537

kevaundray:
## 🗒️ Description

Extracted from #2373 

This PR abstracts away the py-ecc implementation detail, making it easier to switch out the cryptography library in the future.

Can backport in a separate PR.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -299,324 +405,3 @@
...
-        The G1 point and the scalar.
-
-    Raises
-    ------
-    InvalidParameter
-        If the subgroup check failed.
-
-    """
-    if len(data) != 160:
-        InvalidParameter("Input should be 160 bytes long")
```
kevaundray, line 417: Note: I think this was a bug in the code since we are not raising this exception

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/bls12_381_g2.py
```diff
@@ -144,10 +142,11 @@ def bls12_map_fp2_to_g2(evm: Evm) -> None:
     charge_gas(evm, Uint(GAS_BLS_G2_MAP))
 
     # OPERATION
-    field_element = bytes_to_fq2(data)
-    assert isinstance(field_element, FQ2)
```
kevaundray, line 148: Note that `bytes_to_fq2` was being called twice, once here and another to assign to fp2. I also think this assert is always true

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/bls12_381_g1.py
```diff
@@ -63,12 +59,15 @@ def bls12_g1_add(evm: Evm) -> None:
     charge_gas(evm, Uint(GAS_BLS_G1_ADD))
 
     # OPERATION
```
kevaundray, line 61: All of the functions essentially have the following structure:

- unpad data (padding is applied because we wanted everything to be 32 byte aligned)
- call cryptography to do operation
- pad output (so that it is 32 byte aligned)

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,356 @@
...
+"""
+
+from typing import Sequence
+
+from py_ecc.bls.hash_to_curve import (
+    clear_cofactor_G1,
+    clear_cofactor_G2,
+    map_to_curve_G1,
+    map_to_curve_G2,
+)
```
felix314159: > This PR abstracts away the py-ecc implementation detail

What exactly does this mean and why do you still need to import `py_ecc`? Can you clarify whether the goal of the PR to get rid of the `py_ecc` dependency or not? And how can we verify that this implementation is correct / behaves like py-ecc version? I think you should add test cases in this PR. 

And if the goal is to completely get rid of `py_ecc` we need to make a follow-up PR because it currently is used in a bunch of places:
* `src/ethereum/crypto/kzg.py`
* `src/ethereum/forks/{arrow_glacier,berlin,byzantium,cancun,constantinople,gray_glacier,istanbul,london,muir_glacier,paris,shanghai,prague,amsterdam,osaka,bpo1,bpo2,bpo3,bpo4,bpo5}/vm/precompiled_contracts/alt_bn128.py`
* `src/ethereum/forks/{prague,amsterdam,osaka,bpo1,bpo2,bpo3,bpo4,bpo5}/vm/precompiled_contracts/bls12_381/`
* `tests/prague/eip2537_bls_12_381_precompiles/helpers.py`
* `tests/benchmark/compute/precompile/test_alt_bn128.py`

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,356 @@
...
+
+.. contents:: Table of Contents
+    :backlinks: none
+    :local:
+
+Introduction
+------------
+
+Byte-level interface to BLS12-381 curve operations. All inputs and
+outputs are raw (unpadded) bytes.
```
SamWilsn: Use markdown formatted docstrings in new files please!

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,356 @@
...
+    curve_order,
+    is_inf,
+    is_on_curve,
+    normalize,
+)
+from py_ecc.optimized_bls12_381.optimized_curve import add as bls12_add
+from py_ecc.typing import Optimized_Point3D as Point3D
+
+
+def _bytes_to_g1(raw: bytes) -> Point3D[FQ]:
```
SamWilsn:
```suggestion
def _bytes_to_g1(raw: Bytes96) -> Point3D[FQ]:
```

Then we can remove the length check, unless there's a reason to keep it `bytes`.

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -11,29 +11,135 @@
...
-
-from ....vm.memory import buffer_read
+from ethereum_types.numeric import Uint
+
 from ...exceptions import InvalidParameter
 
+ZERO_PAD = b"\x00" * 16
+
+
+def _unpad_fp(data: Bytes) -> Bytes:
```
SamWilsn: We don't really use the `_` prefix.

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -11,29 +11,135 @@
...
+
+    Raises
+    ------
+    InvalidParameter
+        If the padding is invalid.
+
+    """
+    x = _unpad_fp(data[:64])
+    y = _unpad_fp(data[64:])
+    return bytes(x + y)
```
SamWilsn: Don't these just get split later on, or is there a reason to re-join them into `bytes`?

SamWilsn (commented):
As far as I understand this stuff, it seems to have the right shape. I believe @petertdavies might be a better choice for reviewer, however.

General comments are that we use markdown for new docstrings,

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
kevaundray: Ah makes sense, wasn't sure how to say "this thing is private and shouldn't be used outside of this module"

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
kevaundray: Yep, this was mostly the abstraction layer applied -- we can return the unconcatenated items

src/ethereum/crypto/bls12_381.py
kevaundray: Ah good point

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
SamWilsn: I'm indifferent, I think. If it's cleaner/clearer/nicer to concatenate then split, I'm fine with that.

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
```diff
@@ -11,29 +11,135 @@
...
+    raw : Bytes
+        The 48-byte unpadded field element.
+
+    Raises
+    ------
+    InvalidParameter
+        If the leading 16 bytes are not zero.
+
+    """
+    if data[:16] != ZERO_PAD:
```
SamWilsn, line 42:
```suggestion
    if not data.startswith(ZERO_PAD):
```

This is arguably safer (accounts for the length of `ZERO_PAD`), and easier to read.

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/bls12_381_g1.py
```diff
@@ -63,12 +59,15 @@ def bls12_g1_add(evm: Evm) -> None:
...
-    p1 = bytes_to_g1(buffer_read(data, U256(0), U256(128)))
-    p2 = bytes_to_g1(buffer_read(data, U256(128), U256(128)))
+    p1 = unpad_g1(data[:128])
+    p2 = unpad_g1(data[128:256])
```
SamWilsn, line 63: Is dropping `buffer_read` here intentional, and if so, why?

src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/bls12_381_g1.py
```diff
@@ -63,12 +59,15 @@ def bls12_g1_add(evm: Evm) -> None:
...
 
-    result = bls12_add(p1, p2)
+    try:
+        raw = g1_add(p1, p2)
+    except ValueError as e:
+        raise InvalidParameter(str(e)) from e
```
SamWilsn, line 68: What case raises `ValueError`, and why wasn't it being triggered before? Is this a good candidate for a missing test or is this an impossible situation we should leave as a plain `ValueError`?

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,264 @@
...
+
+    `a` and `b_point` are 192-byte G2 points.
+    Return the 192-byte sum as a G2 point.
+    """
+    p1 = bytes_to_g2(a)
+    p2 = bytes_to_g2(b_point)
+    return g2_to_bytes(bls12_add(p1, p2))
+
+
+def g1_msm(
```
SamWilsn, line 147: "msm" isn't really a well-known abbreviation.

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,264 @@
...
+    Return the 96-byte result as a G1 point.
+
+    Raise `ValueError` if a point is not on the curve or fails the
+    subgroup check.
+    """
+    result: Point3D[FQ] = Z1
+    for raw_point, raw_scalar in zip(points, scalars, strict=True):
+        point = bytes_to_g1(raw_point)
+        if not is_inf(bls12_multiply(point, curve_order)):
+            raise ValueError("Subgroup check failed for G1 point")
```
SamWilsn, line 165: Should we move the subgroup check into a dedicated function instead of repeating it in both `g1_msm` and `g2_msm`?

src/ethereum/crypto/bls12_381.py
SamWilsn: Though I guess the original code used it, so it isn't your responsibility to fix 🤣

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,264 @@
...
+    G1 multi-scalar multiplication with subgroup checks.
+
+    `points` is a sequence of 96-byte G1 points and `scalars` is a
+    sequence of 32-byte big-endian scalars.
+    Return the 96-byte result as a G1 point.
+
+    Raise `ValueError` if a point is not on the curve or fails the
+    subgroup check.
+    """
+    result: Point3D[FQ] = Z1
```
SamWilsn, line 161: The old code special cased the first iteration of the loop. What's happening here?

src/ethereum/crypto/bls12_381.py
```diff
@@ -0,0 +1,264 @@
...
+        raise ValueError("y coordinate >= field modulus")
+
+    z = 1
+    if x == 0 and y == 0:
+        z = 0
+
+    point: Point3D[FQ] = (FQ(x), FQ(y), FQ(z))
+    if not is_on_curve(point, b):
+        raise ValueError("G1 point is not on curve")
+    return point
```
SamWilsn, line 58: I see we're dropping the subgroup check from this function (and the caching).

The subgroup checks have been moved directly into the msm functions, so that makes sense I think.

How about the caching?

SamWilsn (commented):
I'm still not qualified to review crypto code, but here are some more thoughts.

PR #2387, opened 2026-03-03
feat(test-all): Use xxhash instead of sha256

### LINKED ISSUE #2087: chore(testing): Replace `hashllib.sha256` with faster algorithm
Maybe a small win in terms of both fixture filling and this hasher command could be to replace this hashing function with something much faster; since we don't need a cryptographically secure hashing algo, we could replace it with something like [xxHash](https://xxhash.com/).

Perhaps for a future PR.

_Originally posted by @marioevz in https://github.com/ethereum/execution-specs/pull/2080#discussion_r2733408033_

marioevz:
## 🗒️ Description

Use a fast, non-crytographic hash for certain use cases that don't require any cryptographic guarantees:
- Pre-alloc group hashes
- Fixture checksums
- Test account checksums (used for pre-allocation grouping de-duplication).

In particular this PR proposes to use `XXH128` from the [xxhash library](https://xxhash.com/).

The new class `Hash128` is also introduced to homogenize usages of this shorter hash throughout all the framework.

## 🔗 Related Issues or PRs
Fixes #2087

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


packages/testing/src/execution_testing/base_types/composite_types.py
```diff
@@ -521,15 +522,15 @@ def __bool__(self: "Account") -> bool:
...
         """Return the hash of the account given its properties."""
         data = self.model_dump(mode="json")
         blob = json.dumps(
             data,
             sort_keys=True,
             separators=(",", ":"),
-        ).encode("utf-8")
-        return Hash(hashlib.sha256(blob).digest())
+        )
+        return Hash128(xxhash.xxh128(blob).digest())
```
marioevz:
```suggestion
        return Hash128(xxhash.xxh128_digest(blob))
```

felix314159:
Have you identified sha256 to be some kind of bottleneck? I had a small `xxhash` vs `sha256` benchmark script written for me and the result was:

```
xxh3_128.digest() 102.5 ns/op (stdev 1.6) x 1.00 
sha256.digest() 293.1 ns/op (stdev 1.3) x 2.86
``` 

So for like 1 million hashes we would save only around ~0.19 seconds. Any modern CPU has hardware-acceleration for sha256 so I am not surprised that it is fairly close.

marioevz:
> Have you identified sha256 to be some kind of bottleneck? I had a small `xxhash` vs `sha256` benchmark script written for me and the result was:
> 
> ```
> xxh3_128.digest() 102.5 ns/op (stdev 1.6) x 1.00 
> sha256.digest() 293.1 ns/op (stdev 1.3) x 2.86
> ```
> 
> So for like 1 million hashes we would save only around ~0.19 seconds. Any modern CPU has hardware-acceleration for sha256 so I am not surprised that it is fairly close.

Not sure but I'm running the experiment to measure the same fill command with the entire set of tests twice and see how much it affects the time.

felix314159:
I am not even sure if re-running the exact fill command would result in consistent times. Even without caching, what if stuff like the worker distribution (im assuming u use `-n auto` or sth like that) is not deterministic, that alone could have a greater effect than the hash function, especially with small sample sizes its impact would be hard to detect

marioevz:
Times before and after the change:
```
uv run fill --clean --fork Amsterdam --generate-all-formats -m "not slow" -n   30602.10s user 246.41s system 1549% cpu 33:10.38 total
```
```
uv run fill --clean --fork Amsterdam --generate-all-formats -m "not slow" -n   29627.44s user 242.34s system 1553% cpu 32:03.03 total
```

marioevz:
Closing this for now since the account hashing that is critical to the ported static tests is affected by this. We can revisit once the static tests are successfully ported.

PR #2389, opened 2026-03-03
fix(clients): update ethrex BAL exception mapper for bal-devnet-3

edg-l:
- Map INVALID_BAL_EXTRA_ACCOUNT to hash mismatch message as ethrex  does not distinguish extra vs missing accounts.
- Add BLOCK_ACCESS_LIST_GAS_LIMIT_EXCEEDED mapping for BAL size cap validation.


felix314159 (approved):
Thanks!

danceratopz (approved):
LGTM!

PR #2396, opened 2026-03-03
refactor(spec-specs): add opcode gas constants

Carsons-Eels:
## 🗒️ Description

Currently, multiple opcodes share the same gas tier constant (e.g. DIV, SDIV, MOD, SMOD, MUL all reference `GAS_LOW = Uint(5)`). This makes it impossible to reprice individual opcodes because changing `GAS_LOW` would affect all of them at once. [EIP-7904 - General Repricing](https://eips.ethereum.org/EIPS/eip-7904) requires per-opcode repricing (e.g. DIV=15, SDIV=20, MOD=12 while MUL stays at 5).

This PR adds per-opcode gas constants (`OPCODE_*`) to all 24 fork `gas.py` modules and updates instruction files to use them instead of shared tier constants (`VERY_LOW`, `LOW`, `MID`, `HIGH`), and likewise makes the same changes to the test code by adding counterpart gas constants to the GasCosts dataclass.

It introduces intermediate variables like `OPCODE_DIV = LOW` in each fork's `gas.py` and updates instruction files to charge against the opcode-specific constant. The existing `apply_spec_repricing()` mechanism (from #2331) can then target `OPCODE_DIV` directly in the JSON config, enabling per-opcode repricing without changing spec code.

**What changed:**

Spec side:
- added 30–38 `OPCODE_*` variable definitions (varying by fork based on available opcodes)
- added ~20 `OPCODE_*` constants for BASE-tier opcodes (`OPCODE_POP`, `OPCODE_MSIZE`, `OPCODE_ADDRESS`, `OPCODE_TIMESTAMP`, `OPCODE_GASLIMIT`, etc.) plus fork-specific constants (`OPCODE_DIFFICULTY` pre-Paris, `OPCODE_PREVRANDAO` Paris+, `OPCODE_RETURNDATASIZE` Byzantium+, `OPCODE_CHAINID` Istanbul+, `OPCODE_BASEFEE` London+, `OPCODE_BLOBBASEFEE` Cancun+, `OPCODE_PUSH0` Shanghai+)
- 168 instruction files (7 per fork): updated imports and `charge_gas()` calls to use opcode-specific constants
- 24 `fork.py` files: moved `GAS_LIMIT_ADJUSTMENT_FACTOR` and `GAS_LIMIT_MINIMUM` into `GasCosts` (minus `GAS_` prefix); updated `check_gas_limit()` references
- 24 `transactions.py` files: moved `GAS_TX_BASE`, `GAS_TX_CREATE`, `GAS_TX_DATA_PER_*`, `GAS_TX_ACCESS_LIST_*` into `GasCosts` (minus `GAS_` prefix); removed module-level definitions and RST docstrings; used lazy imports where circular dependencies exist
- 8 `vm/eoa_delegation.py` files (Prague+): moved `GAS_AUTH_PER_EMPTY_ACCOUNT` into `GasCosts` as `AUTH_PER_EMPTY_ACCOUNT`
- 120 instruction files (5 per fork): replaced all `charge_gas(evm, GasCosts.BASE)` calls with named `GasCosts.OPCODE_*` constants

Testing side:
- `gas_costs.py`: added 39 `OPCODE_*` fields to `GasCosts` dataclass
- `forks.py`: updated `Frontier.gas_costs()` initialization and `opcode_gas_map()` in Frontier + 4 fork overrides (Byzantium, Constantinople, Cancun, Osaka) to reference per-opcode fields

No behavioural changes — all variables default to their original tier constant values

**Fork groups by opcode availability:**
| Forks | Variables |
|-------|-----------|
| frontier → spurious_dragon | 30 (baseline) |
| byzantium | +RETURNDATACOPY |
| constantinople → shanghai | +SHL, SHR, SAR |
| cancun, prague | +MCOPY |
| osaka → amsterdam | +CLZ |

| Forks | Tx Constants |
|-------|-------------|
| frontier → amsterdam | TX_BASE, LIMIT_ADJUSTMENT_FACTOR, LIMIT_MINIMUM |
| frontier | No TX_CREATE |
| homestead → amsterdam | + TX_CREATE |
| frontier → constantinople | TX_DATA_PER_NON_ZERO = 68 |
| istanbul → amsterdam | TX_DATA_PER_NON_ZERO = 16 |
| berlin → amsterdam | + TX_ACCESS_LIST_ADDRESS, TX_ACCESS_LIST_STORAGE_KEY |
| prague → amsterdam | TX_DATA_TOKEN_FLOOR, TX_DATA_TOKEN_STANDARD (replace PER_ZERO/PER_NON_ZERO), AUTH_PER_EMPTY_ACCOUNT |

## 🔗 Related Issues or PRs

- Precursor to [#2331](https://github.com/ethereum/execution-specs/pull/2331) (gas repricing tool)
- Enables per-opcode repricing needed by [#2200](https://github.com/ethereum/execution-specs/issues/2200)
- Required for EIP-7904 general repricing implementation

## :question: Further Questions

- With constants like `OPCODE_BALANCE`, post-Berlin they go from a static cost to a dynamic cost. I've left them in the dynamic section for now, but with the static naming. Treating them differently depending on the fork will cause ordering issues that `ethereum-spec-lint` will complain about. Which to pick, or is there a better option?
- Should the transaction max gas limit also be re-pricable? Currently let it in the module code.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles)
  - it will be used as the squash commit message and should start
  `type(scope):`.
- [x] All: Set appropriate labels for the changes (only maintainers can
  apply labels).
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles)
  - it will be used as the squash commit message and should start
  `type(scope):`.


marioevz:
Awesome!

cc @raxhvl, since this was directly mentioned in the gas-lighters call today.

marioevz:
All failing tests are due to https://github.com/ethereum/execution-specs/pull/2109, and unrelated to the changes in this PR:

```
2026-03-03T22:40:39.1715358Z FAILED tests/json_infra/fixtures/amsterdam_tests/fixtures/blockchain_tests/frontier/validation/test_gas_limit_below_minimum.json::tests/frontier/validation/test_header.py::test_gas_limit_below_minimum[fork_Amsterdam-blockchain_test-gas_limit_5000]
2026-03-03T22:40:39.1719512Z FAILED tests/json_infra/fixtures/amsterdam_tests/fixtures/blockchain_tests/static/state_tests/stStaticCall/static_InternalCallHittingGasLimit.json::tests/static/state_tests/stStaticCall/static_InternalCallHittingGasLimitFiller.json::static_InternalCallHittingGasLimit[fork_Amsterdam-blockchain_test_from_state_test-]
2026-03-03T22:40:39.1723591Z FAILED tests/json_infra/fixtures/amsterdam_tests/fixtures/blockchain_tests/static/state_tests/stTransactionTest/InternalCallHittingGasLimit.json::tests/static/state_tests/stTransactionTest/InternalCallHittingGasLimitFiller.json::InternalCallHittingGasLimit[fork_Amsterdam-blockchain_test_from_state_test-]
```

It should be ok for us to proceed to review -> merge.

packages/testing/src/execution_testing/forks/gas_costs.py
```diff
@@ -91,3 +91,44 @@ class GasCosts:
...
+    GAS_OPCODE_CALLDATACOPY: int
+    GAS_OPCODE_CODECOPY: int
+    GAS_OPCODE_RETURNDATACOPY: int
+    GAS_OPCODE_MLOAD: int
+    GAS_OPCODE_MSTORE: int
+    GAS_OPCODE_MSTORE8: int
+    GAS_OPCODE_MCOPY: int
+    GAS_OPCODE_PUSH_N: int
+    GAS_OPCODE_DUP_N: int
+    GAS_OPCODE_SWAP_N: int
```
marioevz: These two in particular might be confusing since EIP-8024 introduces DUPN and SWAPN opcodes. Maybe we can leave them as `GAS_OPCODE_DUP` and `GAS_OPCODE_SWAP` respectively.

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -87,6 +87,46 @@
...
+GAS_OPCODE_CALLDATACOPY = GAS_VERY_LOW
+GAS_OPCODE_CODECOPY = GAS_VERY_LOW
+GAS_OPCODE_RETURNDATACOPY = GAS_VERY_LOW
+GAS_OPCODE_MLOAD = GAS_VERY_LOW
+GAS_OPCODE_MSTORE = GAS_VERY_LOW
+GAS_OPCODE_MSTORE8 = GAS_VERY_LOW
+GAS_OPCODE_MCOPY = GAS_VERY_LOW
+GAS_OPCODE_PUSH_N = GAS_VERY_LOW
+GAS_OPCODE_DUP_N = GAS_VERY_LOW
+GAS_OPCODE_SWAP_N = GAS_VERY_LOW
```
marioevz: Same comment as in the test file regarding `GAS_OPCODE_DUP` and `GAS_OPCODE_SWAP` due to EIP-8024.Same comment as in the test file regarding `GAS_OPCODE_DUP` and `GAS_OPCODE_SWAP` due to EIP-8024.

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -201,6 +201,46 @@ def gas_costs(
             GAS_PRECOMPILE_BLS_PAIRING_PER_PAIR=0,
             GAS_PRECOMPILE_P256VERIFY=0,
             GAS_BLOCK_ACCESS_LIST_ITEM=0,
+            # Opcode specific gas constants for repricing
+            GAS_OPCODE_ADD=3,
```
marioevz: I wish we could do `GAS_OPCODE_ADD=GAS_VERY_LOW` here, but I don't see how.

marioevz (commented):
Looks good to me, considering we are not touching opcodes that use GAS_BASE, which I'm guessing is because the amount of changes would grow even more, and it's not necessary for the gas repricings at the moment, we could leave it for a future PR if it becomes necessary IMO.

I just left a couple of comments and leave the approval to Sam when he returns.

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -201,6 +201,46 @@ def gas_costs(
...
+            GAS_OPCODE_ISZERO=3,
+            GAS_OPCODE_AND=3,
+            GAS_OPCODE_OR=3,
+            GAS_OPCODE_XOR=3,
+            GAS_OPCODE_NOT=3,
+            GAS_OPCODE_BYTE=3,
+            GAS_OPCODE_SHL=3,
+            GAS_OPCODE_SHR=3,
+            GAS_OPCODE_SAR=3,
+            GAS_OPCODE_CLZ=5,
```
LouisTsai-Csie:
```suggestion
```

Should we remove `CLZ` since this is the `Frontier` fork? I noticed you've already added an entry in the Osaka fork. 

The same issue found in `SHL`, `BLOBHASH`, `MCOPY`, and `RETURNDATACOPY`. Should these only be included in the forks where they were introduced?

LouisTsai-Csie (dismissed):

packages/testing/src/execution_testing/forks/forks/forks.py
LouisTsai-Csie: what about defining `GasCostTier` in the beginning of the file, and reference them here?

```python
class GasCostTier:
  G_BASE = ...
  G_LOW = ...
  G_VERY_LOW = ...

...
GAS_OPCODE_ADD=GasCostTier.G_LOW
```

Irrelevant note: i think we should split this file, it is very large now.

packages/testing/src/execution_testing/forks/gas_costs.py
Carsons-Eels: Nice catch! :blue_heart: I'll let others weigh in on this, but that's def an option.

packages/testing/src/execution_testing/forks/forks/forks.py
Carsons-Eels: Yeah, it does look a tad bloated now... :grimacing:

Carsons-Eels: I thought about this; the tiers come from the yellowpaper so it makes sense to have them since it makes the code easy to work with and verify at a glance.

I think we can use tier constants as the defaults, and let `apply_repricing()` in #2331 override individual opcodes. Something like:

```python
# Tier constants from yellowpaper
GAS_TIER_VERY_LOW = 3
GAS_TIER_LOW = 5
GAS_TIER_MID = 8
GAS_TIER_HIGH = 10

# In gas_costs()
GasCosts(
    ...
    GAS_OPCODE_ADD=GAS_TIER_VERY_LOW,
    GAS_OPCODE_MUL=GAS_TIER_LOW,
    GAS_OPCODE_ADDMOD=GAS_TIER_MID,
    # etc
)
```

The repricing config in #2331 handles the per-opcode overrides via `replace(base_costs, **overrides)` which is already implemented. This PR just needs to reference the tier constants instead of raw values.

SamWilsn (approved):
Looks good to me! Is the intent to make these constants modifiable at runtime? If so, there's some additional nonsense you need to do because Python.

packages/testing/src/execution_testing/forks/forks/forks.py
Carsons-Eels: Okay I've adjusted the code to fix this. I set defaults in the GasCosts class, and added some calls to `replace()` in each fork (or updated the existing ones) to make sure those get added into where they should be in the forks they are defined.

Carsons-Eels (commented):
Regarding Louis' comment about constants in Frontier; I pushed a change that moves the 8 `GAS_OPCODE_*` fields that don't exist until after Frontier (`SHL`, `SHR`, `SAR`, `RETURNDATACOPY`, `BLOBHASH`, `MCOPY`, `CLZ`) to dataclass defaults and then initialize them via `replace()` where the fork introduces new terms. Frontier no longer references them.

I think we **could** do the same thing for the other 23 non-opcode constants that are zero-initialised in Frontier but aren't used until later forks:

**Byzantium:** `GAS_PRECOMPILE_ECADD`, `GAS_PRECOMPILE_ECMUL`, `GAS_PRECOMPILE_ECPAIRING_BASE`, `GAS_PRECOMPILE_ECPAIRING_PER_POINT`
**Istanbul:** `GAS_PRECOMPILE_BLAKE2F_BASE`, `GAS_PRECOMPILE_BLAKE2F_PER_ROUND`
**Berlin:** `GAS_TX_ACCESS_LIST_ADDRESS`, `GAS_TX_ACCESS_LIST_STORAGE_KEY`
**Cancun:** `GAS_PRECOMPILE_POINT_EVALUATION`
**Prague:** `GAS_TX_DATA_TOKEN_STANDARD`, `GAS_TX_DATA_TOKEN_FLOOR`, `GAS_AUTH_PER_EMPTY_ACCOUNT`, `REFUND_AUTH_PER_EXISTING_ACCOUNT`, `GAS_PRECOMPILE_BLS_G1ADD`, `GAS_PRECOMPILE_BLS_G1MUL`, `GAS_PRECOMPILE_BLS_G1MAP`, `GAS_PRECOMPILE_BLS_G2ADD`, `GAS_PRECOMPILE_BLS_G2MUL`, `GAS_PRECOMPILE_BLS_G2MAP`, `GAS_PRECOMPILE_BLS_PAIRING_BASE`, `GAS_PRECOMPILE_BLS_PAIRING_PER_PAIR`
**Osaka:** `GAS_PRECOMPILE_P256VERIFY`
**Amsterdam:** `GAS_BLOCK_ACCESS_LIST_ITEM`

None of these are referenced in Frontier, and they're already set by `replace()` by the fork that introduces them; they'd just also get `= 0` defaults on the dataclass. Frontier doesn't need to mention them at all. This would follow along the same lines as the adjustment I made for Louis' comment, but that predates this PR, so I'd rather get feedback first. We're kind of already here though, so might as well if it feels right.

Thoughts? @LouisTsai-Csie @marioevz

packages/testing/src/execution_testing/forks/forks/forks.py
Carsons-Eels: Okay, I put together an implementation. See how that looks.

Carsons-Eels:
> Looks good to me! Is the intent to make these constants modifiable at runtime? If so, there's some additional nonsense you need to do because Python.

~~I'll address that in #2331~~ scratch that, already modifying those files here. Might as well make the changes here and keep the diff of #2331 smaller.

packages/testing/src/execution_testing/forks/gas_costs.py
LouisTsai-Csie: I support Mario for this, we have `DUPN` and `SWAPN` opcode, renaming `GAS_OPCODE_DUP_N` to `GAS_OPCODE_DUP` seems reasonable.

Carsons-Eels: I'll refactor this and the [other one](https://github.com/ethereum/execution-specs/pull/2396#discussion_r2886324555) today

raxhvl:
opitonal nit: rename `GAS_OPCODE_PUSH_N` to `GAS_OPCODE_PUSH`

To ensure consistency with `GAS_OPCODE_DUP/SWAP` which are now named after their opcodes

src/ethereum/forks/arrow_glacier/vm/gas.py
```diff
@@ -22,46 +22,92 @@
...
+    GAS_PRECOMPILE_IDENTITY_BASE = Uint(15)
+    GAS_PRECOMPILE_IDENTITY_PER_WORD = Uint(3)
+    GAS_RETURN_DATA_COPY = Uint(3)
+    GAS_FAST_STEP = Uint(5)
+    GAS_PRECOMPILE_BLAKE2F_PER_ROUND = Uint(1)
+    GAS_COLD_STORAGE_ACCESS = Uint(2100)
+    GAS_COLD_ACCOUNT_ACCESS = Uint(2600)
+    GAS_WARM_ACCESS = Uint(100)
+
+    # Opcode specific vars used for repricing
```
SamWilsn: Some opcodes (eg. `COINBASE`) don't have their own constant. Is there a rule or pattern for determining when an opcode gets a constant?

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -114,7 +113,7 @@
     "0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02"
 )
 SYSTEM_TRANSACTION_GAS = Uint(30000000)
-MAX_BLOB_GAS_PER_BLOCK = BLOB_SCHEDULE_MAX * GAS_PER_BLOB
+MAX_BLOB_GAS_PER_BLOCK = GasCosts.BLOB_SCHEDULE_MAX * GasCosts.GAS_PER_BLOB
```
SamWilsn: This reminds me, does the `ethereum-spec-new-fork` tool still work?

src/ethereum/forks/arrow_glacier/vm/gas.py
```diff
@@ -82,8 +128,8 @@ class ExtendMemory:
 @dataclass
 class MessageCallGas:
     """
-    Define the gas cost and gas given to the sub-call for
-    executing the call opcodes.
+    Define the gas cost and gas given to the sub-call for executing the call
+    opcodes.
```
SamWilsn, line 183: This change doesn't seem to be mirrored in other forks, was it intentional?

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -76,7 +69,11 @@ def balance(evm: Evm) -> None:
...
     # GAS
     is_cold_access = address not in evm.accessed_addresses
-    gas_cost = GAS_COLD_ACCOUNT_ACCESS if is_cold_access else GAS_WARM_ACCESS
+    gas_cost = (
+        GasCosts.GAS_COLD_ACCOUNT_ACCESS
+        if is_cold_access
+        else GasCosts.GAS_WARM_ACCESS
+    )
     if is_cold_access:
         evm.accessed_addresses.add(address)
```
SamWilsn: Not a requirement, but perhaps this is cleaner:


```suggestion
    gas_cost = GasCosts.GAS_WARM_ACCESS
    if is_cold_access:
        gas_cost = GasCosts.GAS_COLD_ACCOUNT_ACCESS
        evm.accessed_addresses.add(address)
```

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -344,7 +347,9 @@ def extcodesize(evm: Evm) -> None:
     # GAS
     is_cold_access = address not in evm.accessed_addresses
     access_gas_cost = (
-        GAS_COLD_ACCOUNT_ACCESS if is_cold_access else GAS_WARM_ACCESS
+        GasCosts.GAS_COLD_ACCOUNT_ACCESS
+        if is_cold_access
+        else GasCosts.GAS_WARM_ACCESS
     )
```
SamWilsn: Similar kind of comment.

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -484,7 +496,9 @@ def extcodehash(evm: Evm) -> None:
     # GAS
     is_cold_access = address not in evm.accessed_addresses
     access_gas_cost = (
-        GAS_COLD_ACCOUNT_ACCESS if is_cold_access else GAS_WARM_ACCESS
+        GasCosts.GAS_COLD_ACCOUNT_ACCESS
+        if is_cold_access
+        else GasCosts.GAS_WARM_ACCESS
```
SamWilsn: I'll stop mentioning these.

Carsons-Eels:
I updated `GAS_OPCODE_PUSH_N` to `GAS_OPCODE_PUSH` and also added `GAS_OPCODE_COINBASE` too. I'm validating that no more have been missed for any of the forks, and will add those opcodes as well.

packages/testing/src/execution_testing/forks/gas_costs.py
```diff
@@ -91,3 +97,46 @@ class GasCosts:
     REFUND_AUTH_PER_EXISTING_ACCOUNT: int
 
     GAS_BLOCK_ACCESS_LIST_ITEM: int
+
+    # Opcode specific gas constants for repricing
+    GAS_OPCODE_ADD: int
```
SamWilsn: This ship has likely sailed, so feel free to disregard, but if we use `GasCosts.` as the prefix every time, we can probably drop `GAS_` from the constants.

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -25,67 +25,115 @@
...
-GAS_PRECOMPILE_BLS_G1MAP = Uint(5500)
-GAS_PRECOMPILE_BLS_G2ADD = Uint(600)
-GAS_PRECOMPILE_BLS_G2MUL = Uint(22500)
-GAS_PRECOMPILE_BLS_G2MAP = Uint(23800)
-
-GAS_BLOCK_ACCESS_LIST_ITEM = Uint(2000)
+
+class GasCosts:
+    """
+    Constant gas values for the Amsterdam fork of the EVM.
```
SamWilsn: I'm not sure it's necessary to specify the fork in the docstring. It'll just end up being another meaningless diff.

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -201,7 +201,7 @@ def process_create_message(message: Message) -> Evm:
     if not evm.error:
         contract_code = evm.output
         contract_code_gas = (
-            Uint(len(contract_code)) * GAS_CODE_DEPOSIT_PER_BYTE
+            Uint(len(contract_code)) * GasCosts.GAS_CODE_DEPOSIT_PER_BYTE
```
SamWilsn: Not super important, but we have [`ulen`](https://ethereum.github.io/ethereum-types/src/ethereum_types/numeric.py.html#ethereum_types.numeric.ulen:0) should you want to use it.

SamWilsn (commented):
This is `packages/**` and `src/ethereum/forks/amsterdam/**`, and the occasional comment elsewhere.

packages/testing/src/execution_testing/forks/gas_costs.py
Carsons-Eels: That's a good point, the prefix definitely came from before I refactored.

src/ethereum/forks/arrow_glacier/vm/gas.py
Carsons-Eels: No, I think that was at the end of a round of work and I just forgot to put it in my TODOs to update the other forks. I'll align it

src/ethereum/forks/arrow_glacier/vm/gas.py
Carsons-Eels: Not as such. I was thinking to refactor so that every opcode would have a corresponding constant, and the ones that are not opcodes themselves remain as is. Does that sound reasonable?

src/ethereum/forks/arrow_glacier/vm/gas.py
Carsons-Eels: Looks like the text is the same, the formatting just ended up different. I aligned it.

src/ethereum/forks/arrow_glacier/vm/instructions/block.py
```diff
@@ -102,7 +102,7 @@ def timestamp(evm: Evm) -> None:
     pass
 
     # GAS
-    charge_gas(evm, GAS_BASE)
+    charge_gas(evm, GasCosts.GAS_BASE)
```
SamWilsn: Missing opcode constant again perhaps?

src/ethereum/forks/arrow_glacier/vm/gas.py
SamWilsn: Yep, that's perfect.

SamWilsn (commented):
That's all of it, quickly read.

src/ethereum/forks/amsterdam/vm/gas.py
Carsons-Eels: Fixed.

src/ethereum/forks/amsterdam/fork.py
Carsons-Eels: I'll validate this

Carsons-Eels: As far as I can verify, yes this works as expected. Though perhaps we should have it add a docstring to the `src/ethereum/forks/new_fork/__init__.py`? Maybe something like:
```
"""
The TODO fork ([EIP-XXXX]) includes ...

### Changes

TODO

### Releases

TODO
"""

Comment (Carsons-Eels) [reply src/ethereum/forks/amsterdam/fork.py]: Actually, I just noticed some minor weirdness with the way some of the imports are propagated. I'll DM you about it. We can file an issue. Nothing that needs fixing right now

**Comment by SamWilsn** [comment]
To fix the new fork tool tests, you'll need to update the constants that it replaces. For example:

https://github.com/Carsons-Eels/execution-specs/blob/84105b4aa3a92ed051daa9b972a75f3a03c580c7/src/ethereum_spec_tools/new_fork/builder.py#L541

File: packages/testing/src/execution_testing/forks/gas_costs.py [hunk 15]
```diff
@@ -2,92 +2,147 @@
 
 from dataclasses import dataclass
 
+# Common Gas Cost Tiers
+BASE = 2
+VERY_LOW = 3
+LOW = 5
+MID = 8
+HIGH = 10
```
SamWilsn, line 10: Do these names provide enough context, since they are accessed without a leading `GasCosts.`?

```python
from execution_testing.forks.gas_costs import VERY_LOW
```

This would be the standard way to import the constant, and now `VERY_LOW` doesn't indicate anything about gas.

I'm happy either way, but wanted to ask.

packages/testing/src/execution_testing/forks/gas_costs.py
```diff
@@ -2,92 +2,147 @@
...
+    OPCODE_SELFDESTRUCT: int
+
+    # Defined post-Frontier
+    OPCODE_SHL: int = 0
+    OPCODE_SHR: int = 0
+    OPCODE_SAR: int = 0
+    OPCODE_RETURNDATACOPY: int = 0
+    OPCODE_BLOBHASH: int = 0
+    OPCODE_MCOPY: int = 0
+    OPCODE_CLZ: int = 0
```
SamWilsn, line 148: Using zero as a sentinel for "not present" could cause issues, because zero is also (obviously) a valid integer. I can foresee a scenario where we accidentally assume `OPCODE_SHL` is present where it was set to zero, mypy wouldn't be able to detect it, and we end up with a zero cost operation. Using `int | None` would force developers to assert at the time of use.

That said, if this is already how the code works, changing it would be out of scope for this PR and this should be addressed in a separate issue.

src/ethereum/forks/amsterdam/vm/instructions/block.py
```diff
@@ -89,7 +89,7 @@ def coinbase(evm: Evm) -> None:
     pass
 
     # GAS
-    charge_gas(evm, GAS_BASE)
+    charge_gas(evm, GasCosts.BASE)
```
SamWilsn: Why isn't this `GasCosts.OPCODE_COINBASE`?

src/ethereum/forks/amsterdam/vm/instructions/block.py
```diff
@@ -123,7 +123,7 @@ def timestamp(evm: Evm) -> None:
...
```
SamWilsn:
```suggestion
    charge_gas(evm, GasCosts.OPCODE_TIMESTAMP)
```

Similarly the rest of these.

Or did we discuss that but I've forgotten?

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -75,12 +68,11 @@ def balance(evm: Evm) -> None:
...
-    is_cold_access = address not in evm.accessed_addresses
-    gas_cost = GAS_COLD_ACCOUNT_ACCESS if is_cold_access else GAS_WARM_ACCESS
-    if is_cold_access:
+    if address in evm.accessed_addresses:
+        charge_gas(evm, GasCosts.WARM_ACCESS)
+    else:
         evm.accessed_addresses.add(address)
-
-    charge_gas(evm, gas_cost)
+        charge_gas(evm, GasCosts.COLD_ACCOUNT_ACCESS)
```
SamWilsn, line 75: This is _much_ nicer looking.

src/ethereum/forks/amsterdam/vm/instructions/environment.py
```diff
@@ -450,11 +444,14 @@ def returndatacopy(evm: Evm) -> None:
 
     # GAS
     words = ceil32(Uint(size)) // Uint(32)
-    copy_gas_cost = GAS_RETURN_DATA_COPY * words
+    copy_gas_cost = GasCosts.RETURN_DATA_COPY * words
```
SamWilsn: When does a constant get `_PER_WORD`?

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -25,67 +25,141 @@
...
-GAS_PRECOMPILE_BLS_G2MAP = Uint(23800)
-
-GAS_BLOCK_ACCESS_LIST_ITEM = Uint(2000)
+
+class GasCosts:
+    """
+    Constant gas values for the EVM.
+
+    These values may be patched at runtime by a future gas repricing utility
+    """
```
SamWilsn, line 33:
```suggestion
# These constants are in a class so they can be patched at runtime.
class GasCosts:
    """
    Gas value constants for the EVM.
    """
```

I'd put the note about patching into a comment, since it isn't relevant to someone reading the specification; it's more of a note to other developers.

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -25,67 +25,141 @@
...
+    CALL_STIPEND = Uint(2300)
+    NEW_ACCOUNT = Uint(25000)
+
+    # Contract Creation Costs
+    CODE_DEPOSIT_PER_BYTE = Uint(200)
+    CODE_INIT_PER_WORD = Uint(2)
+
+    # Utility
+    ZERO = Uint(0)
+    RETURN_DATA_COPY = Uint(3)  # TODO refactor to OPCODE_RETURNDATACOPY?
```
SamWilsn: Put longer lived TODOs in issues, not in code.

src/ethereum/forks/amsterdam/vm/gas.py
```diff
@@ -25,67 +25,141 @@
...
+    OPCODE_COPY_BASE = Uint(3)
+    OPCODE_CALLDATACOPY = VERY_LOW
+    OPCODE_CODECOPY = VERY_LOW
+    OPCODE_MLOAD = VERY_LOW
+    OPCODE_MSTORE = VERY_LOW
+    OPCODE_MSTORE8 = VERY_LOW
+    OPCODE_SELFDESTRUCT_BASE = Uint(5000)
+
+    # TODO
+    MEMORY = Uint(3)
```
SamWilsn: Should this have a `_PER_WORD` suffix?

src/ethereum/forks/amsterdam/vm/instructions/block.py
Carsons-Eels: Just was lagging behind on commits put through a rebase today, it's pushed up now and should be fixed.

src/ethereum/forks/amsterdam/vm/instructions/environment.py
Carsons-Eels: That has been changed, and should now be per-word. Basically it should be `_BASE` any time it's a base cost for a dynamic calculation, and then either `_PER_` or `_WHATEVER` (eg. `_TOPIC`) for anything else that is a component cost of a dynamic calculation.

packages/testing/src/execution_testing/forks/gas_costs.py
Carsons-Eels: Filed an issue for later https://github.com/ethereum/execution-specs/issues/2659

src/ethereum/forks/amsterdam/vm/instructions/block.py
Carsons-Eels: Nope, they should be fixed now.

packages/testing/src/execution_testing/forks/gas_costs.py
Carsons-Eels: Since we will always be referencing these in the rest of the code base as `GasCosts.X` for repricing safety, it will have the context off the module reference. Plus, we'll be aiming to only ever use the tiers to assign value to the other constants, not to be used in the rest of the code base's functions.

src/ethereum/forks/amsterdam/vm/eoa_delegation.py
```diff
@@ -21,14 +21,13 @@
...
-from ..vm.gas import GAS_COLD_ACCOUNT_ACCESS, GAS_WARM_ACCESS
+from ..vm.gas import GasCosts
 from . import Evm, Message
 
 SET_CODE_TX_MAGIC = b"\x05"
 EOA_DELEGATION_MARKER = b"\xef\x01\x00"
 EOA_DELEGATION_MARKER_LENGTH = len(EOA_DELEGATION_MARKER)
 EOA_DELEGATED_CODE_LENGTH = 23
-GAS_AUTH_PER_EMPTY_ACCOUNT = 25000
 REFUND_AUTH_PER_EXISTING_ACCOUNT = 12500
```
SamWilsn, line 31: This is a gas thing. Should it be moved as well?

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -1122,14 +1119,14 @@ def check_gas_limit(gas_limit: Uint, parent_gas_limit: Uint) -> bool:
...
-    check fails because the gas limit doesn't allow for a sufficient or
-    reasonable amount of gas to be used on a block.
+    ``LIMIT_ADJUSTMENT_FACTOR``. Therefore, if the gas limit that is passed
+    through as a parameter is greater than or equal to the *sum* of the
+    parent's gas and the adjustment delta then the limit for gas is too high
+    and fails this function's check. Similarly, if the limit is less than or
+    equal to the *difference* of the parent's gas and the adjustment delta *or*
+    the predefined ``LIMIT_MINIMUM`` then this function's check fails because
+    the gas limit doesn't allow for a sufficient or reasonable amount of gas to
+    be used on a block.
```
SamWilsn, line 1129: I quickly compared this to arrow glacier, and it looks like the word wrapping wasn't applied there. Is that intentional?

src/ethereum/forks/amsterdam/transactions.py
```diff
@@ -28,43 +28,6 @@
...
-"""
-Gas cost for including an address in the access list of a transaction.
-"""
-
-GAS_TX_ACCESS_LIST_STORAGE_KEY = Uint(1900)
-"""
-Gas cost for including a storage key in the access list of a transaction.
-"""
-
 TX_MAX_GAS_LIMIT = Uint(16_777_216)
```
SamWilsn, line 31: Why not this one?

SamWilsn (approved):

PR #2400, opened 2026-03-04
feat(benchmarks): switch mint test to gas test to prevent OOGs

jochem-brouwer:
## 🗒️ Description
This PR changes the scenario in the ERC20 mint test to do gas checks at the EVM level to ensure it does not OOG. 
NOTE: erc20_approve is also susceptible to this. We can target that one in a new PR.

We have to run this test once to see how it behaves, I have not tested this yet. This can be done on an empty state and the result should be that this keeps looping in all variants, and that at some point it will exit the loop, will do the final SSTORE and the transactions should not OOG.

## 🔗 Related Issues or PRs
I'll inline the idea to do more correct accounting which would work for any unknown contract, provided that at least 3 runs can be done:

```
The tests with these kind of loops look like this in the new version:

setup --> attack_1 --> (call Op.GAS) --> attack_2 --> (call Op.GAS)

This is the setup uses the attack block which is currently used in the While loop. At the end of this setup phase you have two items on stack. If you SWAP1 SUB now this is the gas used of attack_2. 
Why in attack 2? Because attack 1 will EVM-warm up contracts/slots which are shared. Most notably the contract we are targeting (the ERC20) is warm in attack_2 but not in attack_1.

We now Op.ADD(threshold) to this item to leave room for error. To not get involved in too much stack magic we can use Transient storage to use this as lookup table! 😄 Op.TSTORE(x, gas_necessary) where x is a free slot which is never written to again, only read.

Now we do the while loop and cleanup:

setup --> attack_1 --> (call Op.GAS) --> attack_2 --> (call Op.GAS) --> While loop (condition=Op.GT(Op.GASLEFT, Op.TLOAD(x))) --> cleanup

In setup we would load the start value of the loop from storage. In cleanup we write to this slot. In the contract creation we should set the start value of this slot to nonzero to avoid paying for slot creation in cleanup. Notice tha for the cleanup loop we can use the gas calculation. We should still add extra to the threshold because we dit not account for the condition and for any extra opcodes the While adds
```

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -496,6 +496,59 @@ def test_sstore_erc20_approve(
     benchmark_test(pre=pre, blocks=blocks)
 
 
+def build_call_memory_setup(
+    selector: int,
+    *args: Bytecode | int,
+) -> Bytecode:
+    """
+    Build ABI-encoded memory layout for a contract call.
+
+    MEM[0]  = selector (4 bytes, right-aligned in 32-byte word)
+    MEM[32] = args[0]
+    MEM[64] = args[1]  ...
+    """
+    bytecode = Op.MSTORE(
+        0,
+        selector,
+        old_memory_size=0,
+        new_memory_size=32,
+    )
+    for i, arg in enumerate(args):
+        offset = 32 * (i + 1)
+        bytecode += Op.MSTORE(
+            offset,
+            arg,
+            old_memory_size=offset,
+            new_memory_size=offset + 32,
+        )
+    return bytecode
+
+
+def build_external_call(
+    address: Address,
+    num_args: int,
+    *,
+    address_warm: bool = True,
+) -> Bytecode:
+    """
+    Build POP(CALL(...)) using standard ABI memory layout at offset 0.
+
+    args_offset = 28 (selector at byte 28 of the 32-byte word)
+    args_size   = 4 + 32 * num_args
+    """
+    return Op.POP(
+        Op.CALL(
+            address=address,
+            value=0,
+            args_offset=32 - 4,
+            args_size=4 + 32 * num_args,
+            ret_offset=0,
+            ret_size=0,
+            address_warm=address_warm,
+        )
+    )
+
+
 @pytest.mark.parametrize("token_name", SSTORE_MINT_TOKENS)
 @pytest.mark.parametrize("existing_slots", [False, True])
 @pytest.mark.parametrize("cache_strategy", list(CacheStrategy))
@@ -529,11 +582,11 @@ def test_sstore_erc20_mint(
     gas_threshold = 100_000
 
     # Storage key to read and write address pointer to
-    address_pointer_slot = 0
+    slot_offset = 0
 
     # Start at 1 for existing balance slots,
     # or at keccak256("random") for non-existing slots
-    start_address = (
+    start_slot = (
         1
         if existing_slots
         else 0xA4896A3F93BF4BF58378E579F3CF193BB4AF1022AF7D2089F37D8BAE7157B85F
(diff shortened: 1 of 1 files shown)


tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -567,161 +601,104 @@ def test_sstore_erc20_mint(
...
-    # or at keccak256("random") for non-existing slots
-    slot_offset = (
-        1
-        if existing_slots
-        else 0xA4896A3F93BF4BF58378E579F3CF193BB4AF1022AF7D2089F37D8BAE7157B85F
-    )
 
-    while gas_remaining > intrinsic_gas_with_access_list:
-        gas_available = min(gas_remaining, tx_gas_limit)
+    intrinsic_gas_cost_calc = fork.transaction_intrinsic_cost_calculator()
```
LouisTsai-Csie:
```suggestion
    intrinsic_gas_cost_calc = fork.transaction_intrinsic_cost_calculator()()
```

The original `intrinsic_gas_cost_calc` is a callable object, here we should convert it into an integer.

LouisTsai-Csie (dismissed):
Thanks for the update! I think there is not much to change in the code logic, but i would like to refactor the code structure for readability, below are some suggestion:

**Variable rename**
suggested change, not sure if this more clear
- address_pointer_slot -> slot_offset
- start_address -> start_slot (since we are targeting slot)

**Reuse helper function**
I noticed that the memory setup for the (function selector + calldata), and external call argument preparation is repeated multiple times. We could extract this into a helper function.

```python
def build_call_memory_setup(
    selector: int,
    *args: Bytecode | int,
) -> Bytecode:
    """
    Build ABI-encoded memory layout for a contract call.

    MEM[0]  = selector (4 bytes, right-aligned in 32-byte word)
    MEM[32] = args[0]
    MEM[64] = args[1]  ...
    """
    bytecode = Op.MSTORE(
        0,
        selector,
        old_memory_size=0,
        new_memory_size=32,
    )
    for i, arg in enumerate(args):
        offset = 32 * (i + 1)
        bytecode += Op.MSTORE(
            offset,
            arg,
            old_memory_size=offset,
            new_memory_size=offset + 32,
        )
    return bytecode


def build_external_call(
    address: Address,
    num_args: int,
    *,
    address_warm: bool = True,
) -> Bytecode:
    """
    Build POP(CALL(...)) using standard ABI memory layout at offset 0.

    args_offset = 28 (selector at byte 28 of the 32-byte word)
    args_size   = 4 + 32 * num_args
    """
    return Op.POP(
        Op.CALL(
            address=address,
            value=0,
            args_offset=32 - 4,
            args_size=4 + 32 * num_args,
            ret_offset=0,
            ret_size=0,
            address_warm=address_warm,
        )
    )
```

**Code ordering**
I suggest grouping related logic together. For example, `setup_prev_block_cache_warmer` is currently placed near setup_attack, even though it is only used when `cache_strategy == CacheStrategy.CACHE_PREVIOUS_BLOCK` inside that specific `if` block. I would recommend keeping all of the related logic for that path together in one place, including the setup, loop, and cleanup, to make the flow easier to follow.

I’ve put together a refactored version as a reference, hope it helps clarify the idea.
https://gist.github.com/LouisTsai-Csie/fd69e491ce969dd615433a88460b5480

marioevz:
Getting following error when running locally:
```
tests/benchmark/stateful/bloatnet/test_single_opcode.py::test_sstore_erc20_mint[fork_Osaka-blockchain_test_engine-no_change_True-cache_strategy_CacheStrategy.NO_CACHE-existing_slots_True-token_name_30GB_ERC20-benchmark-gas-value_1M] - UnboundLocalError: cannot access local variable 'cache_txs' where it is not associated with a value
```

This is why we have to make these tests fill-compatible, we could catch these issues earlier. We can't fill any tests in this file because we don't have the code of the stubs in EELS, we should just add it.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -529,199 +603,95 @@ def test_sstore_erc20_mint(
...
-        with TestPhaseManager.execution():
-            # Same here, does this create tx is execution mode?
-            # And above setup mode?
-            txs.append(
+    intrinsic_gas_cost = fork.transaction_intrinsic_cost_calculator()()
+    while gas_remaining >= intrinsic_gas_cost + gas_threshold:
+        gas_limits.append(min(gas_remaining, tx_gas_limit))
+        gas_remaining -= gas_limits[-1]
+
+    if cache_strategy == CacheStrategy.CACHE_PREVIOUS_BLOCK:
```
marioevz, line 679:
```suggestion
    cache_txs: List[Transaction] = []
    if cache_strategy == CacheStrategy.CACHE_PREVIOUS_BLOCK:
```
Fix to the "UnboundLocalError: cannot access local variable 'cache_txs' where it is not associated with a value" issue.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -529,199 +603,95 @@ def test_sstore_erc20_mint(
...
-                )
-
-        with TestPhaseManager.execution():
-            # Same here, does this create tx is execution mode?
-            # And above setup mode?
-            txs.append(
+    intrinsic_gas_cost = fork.transaction_intrinsic_cost_calculator()()
+    while gas_remaining >= intrinsic_gas_cost + gas_threshold:
+        gas_limits.append(min(gas_remaining, tx_gas_limit))
+        gas_remaining -= gas_limits[-1]
```
marioevz:
```suggestion
        gas_limit = min(gas_remaining, tx_gas_limit)
        gas_limits.append(gas_limit)
        gas_remaining -= gas_limit
```

marioevz (approved):
I don't like the tests in this file, they cannot be filled because we are treating the ERC20 as blobs, we should just copy the code into this file (use `evm_bytes` command) and compute the correct gas costs, otherwise this test is very brittle to any gas cost changes.

jochem-brouwer:
@marioevz RE:

> I don't like the tests in this file, they cannot be filled because we are treating the ERC20 as blobs, we should just copy the code into this file (use evm_bytes command) and compute the correct gas costs, otherwise this test is very brittle to any gas cost changes.

This test is indeed targeted at a very specific bloated contract. I should maybe have mentioned that I only see this test as temporary to be used with gas repricings. I would want a more robust test which does not target a Solidity generated contract, or an unknown contract (like the ERC20-targeted bloated contracts).
It should target an account with filled storage, delegated to code with EIP-7702, and use that account to make these tests more controllable.
I do not like submitting temporary tests to master. We can/I can open an issue tracking these tests which should be removed and by what these should be replaced with.

jochem-brouwer:
> Getting following error when running locally:
> 
> ```
> tests/benchmark/stateful/bloatnet/test_single_opcode.py::test_sstore_erc20_mint[fork_Osaka-blockchain_test_engine-no_change_True-cache_strategy_CacheStrategy.NO_CACHE-existing_slots_True-token_name_30GB_ERC20-benchmark-gas-value_1M] - UnboundLocalError: cannot access local variable 'cache_txs' where it is not associated with a value
> ```
> 
> This is why we have to make these tests fill-compatible, we could catch these issues earlier. We can't fill any tests in this file because we don't have the code of the stubs in EELS, we should just add it.

Yes we should make them fill compatible, there's also another test generating an error which should be caught by fill. Maybe add a fill mode which bypasses the stub check and only does the static analysis for these tests, and run this on CI?

jochem-brouwer:
Commenting issue here to link this PR to issue, to replace these temp benchs with more general ones: https://github.com/ethereum/execution-specs/issues/2411

PR #2415, opened 2026-03-05
refactor(test-benchmark): ensure no precompile cache issue

### LINKED ISSUE #1603: Implement missing uncachable benchmark test for precompiles
We should add support for generating uncachable test cases for precompiles, similar to the Nethermind Modexp [implementation](https://github.com/NethermindEth/nethermind/blob/94592614342bd26e34912844c183bf918f6eba37/tools/EngineRequestsGenerator/TestCases/Modexp.cs), this could help the benchmark test avoid the cache mechanism in clients.

The core idea is to slightly vary parameters (such as the Base value) for each precompile parameter, ensuring that each invocation produces a unique set of inputs and thus bypasses the precompile cache. This can be done either by incrementing bytes in a flat loop or by implementing jump-based loops due to contract size limits.

LouisTsai-Csie:
## 🗒️ Description

For `test_ecrecover` and `test_alt_bn128_benchmark`, they are using the same input for every call. Some clients (e.g., `Nethermind` and `Reth`) might cache the precompile result, thus skewing the result. This PR changes the precompile input for every call.

## 🔗 Related Issues or PRs

close #1603 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/compute/precompile/test_alt_bn128.py
+++ b/tests/benchmark/compute/precompile/test_alt_bn128.py
@@ -637,3 +637,73 @@ def test_ec_pairing(
         skip_gas_used_validation=True,
         blocks=[Block(txs=txs)],
     )
+
+
+@pytest.mark.repricing
+@pytest.mark.parametrize(
+    "precompile_address,calldata",
+    [
+        pytest.param(
+            0x06,
+            concatenate_parameters(
+                [
+                    "18B18ACFB4C2C30276DB5411368E7185B311DD124691610C5D3B74034E093DC9",
+                    "063C909C4720840CB5134CB9F59FA749755796819658D32EFC0D288198F37266",
+                    "07C2B7F58A84BD6145F00C9C2BC0BB1A187F20FF2C92963A88019E7C6A014EED",
+                    "06614E20C147E940F2D70DA3F74C9A17DF361706A4485C742BD6788478FA17D7",
+                ]
+            ),
+            id="ec_add",
+        ),
+        pytest.param(
+            0x07,
+            concatenate_parameters(
+                [
+                    "1A87B0584CE92F4593D161480614F2989035225609F08058CCFA3D0F940FEBE3",
+                    "1A2F3C951F6DADCC7EE9007DFF81504B0FCD6D7CF59996EFDC33D92BF7F9F8F6",
+                    "0000000000000000000000000000000000000000000000000000000000000002",
+                ]
+            ),
+            id="ec_mul_small_scalar",
+        ),
+        pytest.param(
+            0x07,
+            concatenate_parameters(
+                [
+                    "1A87B0584CE92F4593D161480614F2989035225609F08058CCFA3D0F940FEBE3",
+                    "1A2F3C951F6DADCC7EE9007DFF81504B0FCD6D7CF59996EFDC33D92BF7F9F8F6",
+                    "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF",
+                ]
+            ),
+            id="ec_mul_max_scalar",
+        ),
+    ],
+)
+def test_alt_bn128_uncachable(
+    benchmark_test: BenchmarkTestFiller,
+    precompile_address: Address,
+    calldata: bytes,
+) -> None:
+    """
+    Benchmark ecAdd/ecMul with chained output-to-input feedback.
+
+    Result overwrites the input point each iteration, so every call
+    has unique input and cannot benefit from precompile caching.
+    """
+    attack_block = Op.POP(
+        Op.STATICCALL(
+            gas=Op.GAS,
+            address=precompile_address,
+            args_size=Op.CALLDATASIZE,
+            ret_size=64,  # One G1 point (2 * 32 bytes), overwrites input.
+        ),
+    )
+
+    benchmark_test(
+        target_opcode=Op.STATICCALL,
+        code_generator=JumpLoopGenerator(
+            setup=Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE),
+            attack_block=attack_block,
+            tx_kwargs={"data": calldata},
+        ),
+    )
--- a/tests/benchmark/compute/precompile/test_alt_bn128.py
+++ b/tests/benchmark/compute/precompile/test_alt_bn128.py
@@ -501,6 +501,9 @@ def test_bn128_pairings_amortized(
     )
 
 
+NUM_INPUT_VARIANTS = 4
+
+
 @pytest.mark.repricing
 @pytest.mark.parametrize("num_pairs", [1, 3, 6, 12, 24])
 def test_alt_bn128_benchmark(
@@ -513,6 +516,48 @@ def test_alt_bn128_benchmark(
     attack_block = Op.POP(
         Op.STATICCALL(gas=Op.GAS, address=0x08, args_size=Op.CALLDATASIZE),
     )
+
+    benchmark_test(
+        target_opcode=Op.STATICCALL,
+        code_generator=JumpLoopGenerator(
+            setup=Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE),
+            attack_block=attack_block,
+            tx_kwargs={"data": calldata},
+        ),
+    )
+
+
+@pytest.mark.repricing
+@pytest.mark.parametrize("num_pairs", [1, 12, 24])
+def test_ec_pairing(
+    benchmark_test: BenchmarkTestFiller,
+    num_pairs: int,
+) -> None:
+    """Benchmark ecpairing precompile with rotating inputs to avoid caching."""
+    pair_size = num_pairs * 192
+    counter_offset = NUM_INPUT_VARIANTS * pair_size
+
+    calldata = Bytes(
+        b"".join(
+            _generate_bn128_pairs(num_pairs, seed=42 + i)
+            for i in range(NUM_INPUT_VARIANTS)
+        )
+    )
+
+    attack_block = Op.POP(
+        Op.STATICCALL(
+            gas=Op.GAS,
+            address=0x08,
+            args_offset=Op.MUL(
+                Op.AND(
+                    Op.MLOAD(counter_offset),
+                    NUM_INPUT_VARIANTS - 1,
+                ),
+                pair_size,
+            ),
+            args_size=pair_size,
+        ),
+    ) + Op.MSTORE(counter_offset, Op.ADD(Op.MLOAD(counter_offset), 1))
 
     benchmark_test(
         target_opcode=Op.STATICCALL,
--- a/tests/benchmark/compute/precompile/test_alt_bn128.py
+++ b/tests/benchmark/compute/precompile/test_alt_bn128.py
@@ -549,26 +549,20 @@ def test_ec_pairing(
         + gsc.GAS_PRECOMPILE_ECPAIRING_PER_POINT * num_pairs
     )
 
-    # Each iteration: STATICCALL ecpairing at a rotating calldata offset,
-    # then increment counter at memory[CALLDATASIZE].
+    # Each iteration: STATICCALL ecpairing at advancing calldata offset,
+    # then advance offset by pair_size at memory[CALLDATASIZE].
     # The loop condition checks remaining gas against one body execution.
     attack_block = Op.POP(
         Op.STATICCALL(
             gas=Op.GAS,
             address=0x08,
-            args_offset=Op.MUL(
-                Op.MOD(
-                    Op.MLOAD(Op.CALLDATASIZE),
-                    Op.DIV(Op.CALLDATASIZE, pair_size),
-                ),
-                pair_size,
-            ),
+            args_offset=Op.MLOAD(Op.CALLDATASIZE),
             args_size=pair_size,
             address_warm=True,
         ),
     ) + Op.MSTORE(
         Op.CALLDATASIZE,
-        Op.ADD(Op.MLOAD(Op.CALLDATASIZE), 1),
+        Op.ADD(Op.MLOAD(Op.CALLDATASIZE), pair_size),
     )
 
     setup = Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE)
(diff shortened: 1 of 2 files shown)


tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -501,19 +501,38 @@ def test_bn128_pairings_amortized(
     )
 
 
+NUM_INPUT_VARIANTS = 5
```
misilva73: Are we sure that 5 variants is enough? This seems a bit low to ensure clients are cycling through their cache. We should double check with the folks from nethermind - I think they had a good set of precompile tests before we moved to EELS

tests/benchmark/compute/precompile/test_ecrecover.py
```diff
@@ -44,7 +44,11 @@ def test_ecrecover(
 
     attack_block = Op.POP(
         Op.STATICCALL(
-            gas=Op.GAS, address=precompile_address, args_size=Op.CALLDATASIZE
+            gas=Op.GAS,
```
misilva73: I am not following how this solves the caching issue. For ecRecover, we would need to source multiple valid (hash, v, r, s) tuples to feed into the STATICCALL. I am not sure the best way to do this. I thinkwe should reach out the Lukazs from nethermind to see how they initially implemented it.

jochem-brouwer (changes requested):
I would like to go a bit in-depth about the precompile caching or the general cache strategies used by clients to align on how we should address those in the benchmark suite. It's for instance also the tx splits happening since Osaka, which possibly mess up Prague measurements because of these splits (these actually speed up (!) the calc)
Not a blocker, but for completeness mentioning here.

tests/benchmark/compute/precompile/test_alt_bn128.py
LouisTsai-Csie: After checking with Marcin, there appears to be no good general solution for making the `ecPairing` benchmark uncachable, Nethermind still suggest hardcode values and switch between iterations, please see discussion:

https://discord.com/channels/1359927674746835211/1478793120442355742/1479054414214991963

For this refactor, why choosing `NUM_INPUT_VARIANTS` = 4? I break down the gas cost usage, not sure if this is reasonable to you:

- Precompile base: 45,000
- Precompile per point: 34,000 × num_pairs
- STATICCALL warm access: 100
- Stack/memory ops overhead: ~35

**Calldata cost**
num_pairs | calldata size (4 variants) | intrinsic gas cost | % of 16M
-- | -- | -- | --
1 | 768 bytes | ~12K | 0.08%
12 | 9,216 bytes | ~147K | 0.9%
24 | 18,432 bytes | ~295K | 1.8%

**Iteration & costs**

num_pairs | gas/iteration | iterations (16M tx cap) | cycles through 4 variants
-- | -- | -- | --
1 | ~79,135 | ~202 | ~50
12 | ~453,135 | ~35 | ~9
24 | ~861,135 | ~18 | ~4

Question 1: What `num_pairs` value should we target for benchmark use case? If a high value like 24 isn't required, we could afford to increase `NUM_INPUT_VARIANTS` further?

Question 2: How does the client precompile cache operation? Should i make sure the input are entirely different within the same transaction, or even the same block. 

I could adjust the benchmark accordingly, please let me know what best fits in this scenario

tests/benchmark/compute/precompile/test_ecrecover.py
LouisTsai-Csie: I updated the benchmark to follow Nethermind's approach: incrementing the `hash` field of the `ecrecover` input by 1 on each call to prevent caching.

I also updated `ret_offset` and `ret_size` to non-zero values. I'm not entirely sure whether this affects performance, but there is a concern that setting both to zero could allow clients to skip the precompile call entirely, since the return value would never be used.

LouisTsai-Csie:
@jochem-brouwer I'm also not entirely sure how the caching works here. Should we ensure that all precompile calls within a block have distinct inputs, or is the cache cleared between transactions? And is there bounded capacity for the client cache?

marcindsobczak:
> Should we ensure that all precompile calls within a block have distinct inputs

Yes
> is the cache cleared between transactions?

No. In Nethermind cache is per block

tests/benchmark/compute/precompile/test_alt_bn128.py
misilva73: Ideally, we would have enough variants to fill the max block (300M gas units). EcPairing receives inputs of size 192 and each additional pair (i.e., the 192 sized input) costs 34k. So, if we have 300'000'000/34'000 = 8'824 different pairs, we are certain to be on the safe side. We will not be able to use all these pairs of course (there is still call data costs, intrinsic tx costs, other opcodes, etc.), but this is a upper bound. 

My question now is: it is a problem to have a test with such a large call data size? Can we forsee any issues with having so many variants?

tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -522,3 +525,45 @@ def test_alt_bn128_benchmark(
             tx_kwargs={"data": calldata},
         ),
     )
+
+
+@pytest.mark.repricing
+@pytest.mark.parametrize("num_pairs", [1, 12, 24])
```
misilva73: Can we keep same values as the previous test?


```suggestion
@pytest.mark.parametrize("num_pairs", [1, 3, 6, 12, 24])
```

tests/benchmark/compute/precompile/test_alt_bn128.py
LouisTsai-Csie: To ensure there are no caching issues at all, and based on Marcin’s [feedback](https://github.com/ethereum/execution-specs/pull/2415#issuecomment-4022355054)￼, I added a new test, `test_ec_pairing`, where each `staticcall` invokes the precompile with distinct input 👍🏻

tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -522,3 +527,122 @@ def test_alt_bn128_benchmark(
...
+            args_size=pair_size,
+            address_warm=True,
+        ),
+    ) + Op.MSTORE(
+        Op.CALLDATASIZE,
+        Op.ADD(Op.MLOAD(Op.CALLDATASIZE), pair_size),
+    )
+
+    setup = Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE)
+    loop = While(
```
misilva73, line 569: Why are we using an infinite loop here? Given that `per_tx_variants` (the number of unique inputs generated) is sized using a conservative overestimate of cost-per-variant, I wonder if we could have a case where the loop always runs more times than there are unique pairs.

To make sure we don't do this, is it possible to add a loop-exit condition after we do per_tx_variants number of calls? Or make a underestimation on the cost-per-variant?

tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -522,3 +527,122 @@ def test_alt_bn128_benchmark(
...
+            - execution_intrinsic
+            - setup_cost
+            - math.ceil(len(calldata) / 32) * gsc.GAS_COPY
+            - mem_exp(new_bytes=len(calldata) + 32)
+        )
+
+        if gas_for_loop < iteration_cost:
+            break
+
+        num_iterations = gas_for_loop // iteration_cost
```
misilva73: These two varibles don't seem to be used in the code. What is their purpose?

misilva73 (commented):
Clever tricks on on both EcRecover and EcPairings! well done 👏 

The logic looks correct to me. I only added two small comments/questions.

On a different note, while reviewing this PR, I noticed that EcAdd and EcMul from `test_alt_bn128` may also suffer from these caching issue. @marcindsobczak, does netherming also cache the results from these two precompiles?

tests/benchmark/compute/precompile/test_alt_bn128.py
LouisTsai-Csie: I add a condition to the while-loop: `condition=Op.GT(Op.CALLDATASIZE, Op.MLOAD(Op.CALLDATASIZE))`

`MEM[CALLDATASIZE]` is used as a counter, it stored the current maximum calldata counter being used, we keep increasing `MEM[CALLDATASIZE]` until it reaches the `CALLDATASIZE`, meaning that there is no more inputs could be used.

tests/benchmark/compute/precompile/test_alt_bn128.py
LouisTsai-Csie: I was doing some gas calculation based on these variable, will remove them as they are redundant now.

LouisTsai-Csie:
@misilva73 Add uncachable `ecAdd` and `ecMul` benchmarks. Please take a look.

Both precompiles take a G1 point (x, y) as the first 64 bytes of input and return a G1 point (x, y) as output. The strategy (discussed with Marcin) is to overwrite the first 64 bytes of input with the return value of each precompile call, so every iteration within a transaction has unique precompile input. Does this approach make sense to you?

tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -522,3 +527,195 @@ def test_alt_bn128_benchmark(
...
+    remaining_gas = gas_benchmark_value
+
+    seed = 0
+    while remaining_gas > 0:
+        gas_available = min(tx_gas_limit, remaining_gas)
+
+        calldata = Bytes(
+            _generate_g1_point(seed) + _generate_g1_point(seed + 1000)
+            if scalar is None
+            else scalar.to_bytes(32, "big")
```
misilva73: For ecMul, shouldn't the calldata be a G1 point plus the scalar? Something like:

```suggestion
            else _generate_g1_point(seed) + scalar.to_bytes(32, "big")
```

misilva73 (commented):
Strategy for ecMul and ecAdd seems solid. Great job 👏

tests/benchmark/compute/precompile/test_alt_bn128.py
LouisTsai-Csie: Good catch 💯  I thought Python was doing:

```python
Bytes(                                                                                                       
      _generate_g1_point(seed) + 
      (_generate_g1_point(seed + 1000))
      if scalar is None                                                                                        
      else (scalar.to_bytes(32, "big"))
     )
```

But it is doing this instead:
```python
Bytes(                                                                                                       
      (_generate_g1_point(seed) + _generate_g1_point(seed + 1000))  
      if scalar is None                                                                                        
      else (scalar.to_bytes(32, "big")) 
)
```

tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -522,3 +527,195 @@ def test_alt_bn128_benchmark(
...
+                gas_limit=per_tx_gas,
+                data=calldata,
+            )
+        )
+        remaining_gas -= per_tx_gas
+        seed_offset += per_tx_variants
+
+    benchmark_test(
+        target_opcode=Op.STATICCALL,
+        skip_gas_used_validation=True,
```
danceratopz, line 637: Any reason why we still `skip_gas_used_validation` here? Performance? It's not skipped below in `test_alt_bn128_uncachable`. I didn't get through all the tests, but they were pass in `fill`.

tests/benchmark/compute/precompile/test_ecrecover.py
danceratopz:
```suggestion
    """
    Benchmark ECRECOVER precompile with unique input per call.

    Each loop iteration increments the hash at memory[0] by
    the STATICCALL success flag (1) so every call receives a
    distinct input, avoiding precompile result caching in
    clients.
    """
```

tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -522,3 +527,195 @@ def test_alt_bn128_benchmark(
...
+@pytest.mark.parametrize("num_pairs", [1, 3, 6, 12, 24])
+def test_ec_pairing(
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_benchmark_value: int,
+    tx_gas_limit: int,
+    num_pairs: int,
+) -> None:
+    """Benchmark ecpairing precompile with unique inputs per call."""
```
danceratopz, line 542:
```suggestion
    """
    Benchmark ecPairing precompile with unique inputs per call.

    Pack multiple uniquely-seeded pairing sets into calldata
    and advance a memory-stored offset each iteration so every
    STATICCALL reads a distinct slice, avoiding precompile
    result caching in clients.
    """
```

tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -522,3 +527,195 @@ def test_alt_bn128_benchmark(
...
+
+    benchmark_test(
+        target_opcode=Op.STATICCALL,
+        skip_gas_used_validation=True,
+        blocks=[Block(txs=txs)],
+    )
+
+
+def _generate_g1_point(seed: int) -> Bytes:
+    """Generate a valid random G1 point from a deterministic seed."""
```
danceratopz, line 643:
```suggestion
    """
    Benchmark ecAdd/ecMul with unique input per call.

    Write the precompile's G1 output (64 bytes) back over the
    input point so each loop iteration receives a distinct
    input, avoiding precompile result caching in clients.
    """
```

danceratopz (approved):
Nice cache-busting! One question and some suggested docstring elaborations to assist future benchmarkoors. Otherwise, an approve!

tests/benchmark/compute/precompile/test_alt_bn128.py
LouisTsai-Csie: I do this because there was some concern that the infinite loop condition might trigger unexpected call with cache precompile input. Please see this comment: https://github.com/ethereum/execution-specs/pull/2415#discussion_r2932725718

That said, this could be refactored in the future, use the same approach as `test_alt_bn128_uncachable` that does not skip the gas usage validation. But i would suggest a separate PR.

danceratopz: Is this worth a follow-up issue or is it so low-prio it's not worth it?

danceratopz (approved):
Thanks @LouisTsai-Csie!

PR #2425, opened 2026-03-05
feat(tests-static): Port Remaining ECMUL Tests

marioevz:
## 🗒️ Description
Port and remove the remaining ECMUL (EIP-196) static tests.

### Static Test Filename Structure

All static tests for ecmul follow this structure:

```
ecmul_<PointX>-<PointY>_<Scalar>_<Gas>_<PrecompileInputLength>Filler.json
```

E.g. test with filename:
```
tests/static/state_tests/stZeroKnowledge/ecmul_1-2_340282366920938463463374607431768211456_21000_128Filler.json
```
References:
- `1-2`: the generator point (1, 2)
- `340282366920938463463374607431768211456`: 2**128 scalar
- `21000`: 21K evm gas
- `128`: Input size of 128 (32 bytes more than the expected size).

### Porting Process

1. Locate the relevant test's input: grep for `:raw 0x30c8d1da` across both state test directories (`stZeroKnowledge` and `stZeroKnowledge2`) to identify all ECMUL test vectors' input data.
2. Strip the ABI wrapper: Remove the next 64 bytes following the selector, which are the ABI-encoded bytes argument header (32 bytes offset + 32 bytes length), leaving the raw precompile input.
3. Decode the precompile input: The remaining bytes can be divided in 32-byte chunks for easier analysis, and normally map directly to the ecMul input layout per EIP-196:
- bytes 0x00..0x01 - point x coordinate
- bytes 0x00..0x02 - point y coordinate
- bytes 0x00..0x03 - scalar
- bytes 0x00..0x04 - (used to test extra data that is supposed to be ignored by precompile)
4. Match against known constants: From the obtained byte words, decode the well-known values in the decoded fields, such as points:
- G1 (1, 2): Referenced as `1-2` in filenames.
- Point Not in Curve (1, 3): Referenced as `1-3` in filenames.
- Point Not in Curve (0, 3): Referenced as `0-3` in filenames.
- Infinity Point (0, 0): Referenced as `0-0` in filenames.
- T1: Point in curve used in some of the tests, referenced as `7827-6598` in filenames.
And scalars:
- R / R - 1: Group order for the BN254 curve, referenced as `5617` (or `5616` for R - 1) in filenames.
- 2**128: Referenced as `340282366920938463463374607431768211456` in test filenames.

### Skipped Variants

1. The original static tests use 96 and 128 input length variants for every single test, which seemed excessive. The ported tests use 96 input size for most tests and 128 for a small subset to keep the amount of tests low.
2. The 21K/28K variants are reduced to a single variant that uses the exact gas required by the precompile.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


felix314159 (approved):
thanks. we might be missing a test though: eip-196 states that both ECADD and ECMUL fail on invalid input and consume all gas provided. the current `test_invalid` cases verify that the call returns failure (return code 0) and empty return data, but they don't verify that the gas forwarded to the precompile was actually fully consumed.

maybe we could add a test that:

* passes extra gas to the precompile (e.g. via `precompile_gas_modifier`)
* records `GAS` before and after the `CALL`
* asserts the consumed gas equals the total gas forwarded (i.e. all gas was consumed, not just the precompile cost)

maybe even do this for both `ECMUL` and `ECADD` invalid input tests. can be a follow-up pr tho wdyt

marioevz:
> thanks. we might be missing a test though: eip-196 states that both ECADD and ECMUL fail on invalid input and consume all gas provided. the current `test_invalid` cases verify that the call returns failure (return code 0) and empty return data, but they don't verify that the gas forwarded to the precompile was actually fully consumed.
> 
> maybe we could add a test that:
> 
>     * passes extra gas to the precompile (e.g. via `precompile_gas_modifier`)
> 
>     * records `GAS` before and after the `CALL`
> 
>     * asserts the consumed gas equals the total gas forwarded (i.e. all gas was consumed, not just the precompile cost)
> 
> 
> maybe even do this for both `ECMUL` and `ECADD` invalid input tests. can be a follow-up pr tho wdyt

Yes, agreed. I've created an issue to follow this up, since I think the purpose of this PR is to port only the existing static tests.

PR #2447, opened 2026-03-09
feat(tests): add missing legacy modexp cases with oversized lengths

### LINKED ISSUE #1564: feat(tests): additional test cases for eip-7883
#### ModExp Gas Cost Increase

- [ ] Cc-ing from ethereum/execution-spec-tests#1929

> I didn’t create a separate PR for https://github.com/ethereum/execution-specs/issues/1576, as it depends on the infrastructure introduced in this PR.
> 
> There are 37 test cases in the legacy modexpFiller.json test. Most of them have been ported, but the following cases are still pending and require further analysis:
> 
>     Case 2: Would parse a base length of 0, a modulus length of 32, and an exponent length of 2256 - 1, where the base is empty, the modulus is 2256 - 2 and the exponent is (2256 - 3) * 256(2**256 - 33) (yes, that's a really big number). It would then immediately fail, as it's not possible to provide enough gas to make that computation.
>     Case 28: base length 4TiB
>     Case 29: exp length 4TiB; returns 0 because mod is zero
>     Case 30: base and mod have zero-length. exp's length is 2^255. Since mod is zero, the result should be zero.
> 
> Case 36: the input found on 10 Oct. 2017 that overflows the gas calculation
> 
>     Case 37: input found in July 2022, overflows the gas calculation
> 
>     Based on the description of case 2, 28, 29 and 30, they will fail as eip7623 introduces the upper bound for each field, we already have similar test in test_modexp_invalid_inputs.
>     For case 36, 37, I am checking if it is valid after Fusaka.

- [ ] Cases that hit the max tx gas limit cc ethereum/execution-spec-tests#1929 

> I initially created test cases for two scenarios:
> 
>     base/exponent is invalid while modulus is valid.
>     All components are invalid.
> 
> However, I later realized that these tests fail not due to the input upper bound, but because they exceed the transaction gas limit cap (see explanation https://github.com/ethereum/execution-spec-tests/pull/1929#discussion_r2241517616).
> 
> While filling each component with 0xFF bytes makes such scenarios impossible in practice, I believe it may still be possible to construct cases that trigger this condition. I explored how to generate these cases but have not much idea now.

- [ ] Update spec: https://github.com/ethereum/execution-spec-tests/pull/1993#issuecomment-3164035922

Please comment any additional test cases or edit the issue and add more.

spencer-tb:
## 🗒️ Description

Add 6 legacy modexpFiller.json test cases (2, 28, 29, 30, 36, 37) that were excluded due to oversized declared lengths. These cases fail under EIP-7823 upper bounds (Osaka+). A future port can handle older forks.

## 🔗 Related Issues or PRs

Closes https://github.com/ethereum/execution-specs/issues/1564

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
+++ b/tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
@@ -118,6 +118,94 @@ def test_modexp_invalid_inputs(
 
 
 @pytest.mark.parametrize(
+    "modexp_input,",
+    [
+        pytest.param(
+            # base_len=0, exp_len=32, mod_len=2^256-1
+            bytes.fromhex(
+                "0000000000000000000000000000000000000000000000000000000000000000"
+                "0000000000000000000000000000000000000000000000000000000000000020"
+                "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
+                "fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe"
+                "fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffd"
+            ),
+            id="legacy-case-2",
+        ),
+        pytest.param(
+            # base_len=4TiB, exp_len=0, mod_len=0
+            bytes.fromhex(
+                "0000000000000000000000000000000000000000000000000000040000000000"
+                "0000000000000000000000000000000000000000000000000000000000000000"
+            ),
+            id="legacy-case-28",
+        ),
+        pytest.param(
+            # base_len=0, exp_len=4TiB, mod_len=0
+            bytes.fromhex(
+                "0000000000000000000000000000000000000000000000000000000000000000"
+                "0000000000000000000000000000000000000000000000000000040000000000"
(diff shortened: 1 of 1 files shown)


tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py
```diff
@@ -117,6 +117,94 @@ def test_modexp_invalid_inputs(
     )
 
 
+@pytest.mark.parametrize(
+    "modexp_input,",
+    [
+        pytest.param(
+            # base_len=0, exp_len=32, mod_len=2^256-1
```
LouisTsai-Csie, line 124: From original issue, it is

>  base length of 0, a modulus length of 32, and an exponent length of 2^256 - 1

The `exp_len` and `mod_len` here are swapped, that said, i think both cases would just fail since the input size limit

LouisTsai-Csie (approved):
Leave one comment

spencer-tb:
Thanks @LouisTsai-Csie! Added!

PR #2455, opened 2026-03-09
feat(tests): add script to port static state test fixtures to Python

leolara:
## 🗒️ Description

Add `scripts/fixture_to_python.py` to convert compiled static state test
fixtures (JSON) into Python test files under `tests/ported_static/`. This
generates 5,518 test files (1,848 unique test modules) that produce 40,587
test cases across forks — all passing with zero failures and zero changes
to `pyproject.toml`.

The script reads compiled fixture JSON (produced by `--fill-static-tests`)
and the original filler source (for docstrings), then generates a Python
test file with:

- Op-language bytecode (readable, self-documenting)
- Proper pytest markers (`@pytest.mark.valid_from`, `@pytest.mark.ported_from`)
- Multi-case parametrization when the original filler varies data/gas/value
- Post-state assertions derived from the compiled fixture's actual execution
  results
- Post-generation `ruff format` pass with automatic `# noqa: E501` patching
  for remaining long lines
- Type annotations on all parametrized function signatures (passes mypy strict)

### Excluded categories

Three test categories are excluded from compilation because they are
prohibitively slow to fill:

- `stQuadraticComplexityTest`
- `stStaticCall`
- `stTimeConsuming`

### Key design decisions

**Post assertions from compiled fixture state, not filler expect results.**
The original filler `expect` blocks use labels (`<eoa:0xADDR>`), reference
addresses not in the final allocation, and may disagree with actual execution.
Instead, post assertions are taken from the compiled fixture's `post.state`,
which is the actual EVM execution result.

**Only assert storage and code, not balance or nonce.** The compiled fixtures
are fork-specific, but the generated Python tests run across multiple forks
via `valid_from`. Balance depends on gas costs which vary across forks, so
asserting exact balances would cause false failures.

**Label regex handles both 2-part and 3-part formats.** The filler YAML uses
`<eoa:0xADDR>` (2-part) and `<contract:name:0xADDR>` (3-part). The regex
`<[^>]*?(0x[0-9a-fA-F]+)>` handles both.

**Addresses are always padded to 20 bytes.** Short addresses like `0x512` are
zero-padded to the full 40 hex chars before emitting `Address("0x...")`.

### Fixture comparison script

`scripts/compare_fixtures.py` compares two fixture directories that have
different internal layouts. The compiled fixtures use
`for_{fork}/static/state_tests/{category}/{name}.json` while the
Python-generated ones use `for_{fork}/ported_static/{category}/{name}/{name}.json`.
The script matches by `(category, normalized_name)` across all fork
directories and compares post-state hashes. Names are normalized via the same
camelCase-to-snake_case transforms as the generator so that e.g.
`addNonConst` matches `add_non_const`.

### CI changes

The ported tests are excluded from `tox -e py3` and `tox -e pypy3` via
`--ignore=tests/ported_static` to avoid OOM on CI runners (the extra ~40K
items would push the total from ~100K to ~160K). The ported tests are
intended to be filled separately or included once CI runners are sized
appropriately.

### Verification

```bash
# Generate
uv run python scripts/fixture_to_python.py \
    --fixtures /tmp/compiled_static/state_tests/ \
    --fillers tests/static/state_tests/ \
    --output tests/ported_static/
# 5518 generated, 0 failed

# Lint
uvx tox -e static
# All checks pass (codespell, ruff check, ruff format, mypy, ethereum-spec-lint)

# Fill
uv run fill tests/ported_static/ --output /tmp/fixtures_python/ --clean -n 10
# 40587 passed, 0 failed, 0 errors

# Compare
python scripts/compare_fixtures.py /tmp/compiled_static/ /tmp/fixtures_python/
# 1846/1846 matched, 0 mismatches
```

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Screenshot 2026-03-09 at 21 02 49


marioevz (commented):
Looking pretty good.

Some comments (I'm taking `tests/ported_static/stCreate2/test_create2_first_byte_loop.py` as example here):

- The contract bytecode should be run through our `evm_bytes` converter. You can import it from `execution_testing.cli.evm_bytes`.
- For the sender, we can use `EOA` and the resulting code should be exactly equivalent: instead of `sender = Address("0xc4a2ca1058df329e5da4755f9921ddaf05cbaa06")`, do `sender = EOA(key=0xf79127a3004abde26a4cbd80c428cb10f829fa11b54d36e7b326f4f4a5927acf)`, and then use that as if it was an Address.
- For the contracts we should be able to, instead of:
```python
contract = Address("0x09fdd11d68be787a4c43f692a0778befc011cd35")
pre[contract] = Account(
    balance=0,
    nonce=0,
    code=...,
)
```
do instead:
```python
contract = pre.deploy_contract(
    code=...,
    nonce=0,
    balance=0,
    address=Address("0x09fdd11d68be787a4c43f692a0778befc011cd35"),
)
```
This way, the removal of the hard-coded address in some test is only a single-line removal.
- Default values from `Transaction` can be removed (`value=0`, `data=b""`, `nonce=0`, etc.) and the test should still be equivalent.

I've tried to take one of the resulting python test files and do a manual conversion and it's pretty straight-forward, so this should automatic conversion seems like a step in the right direction!

marioevz:
Suggestion for commit order to merge this PR: first add the conversion script in one commit, then add the ported Python tests in a second commit, then remove the old YAML/static tests in a third commit, this keeps everything in git history, and then we can Rebase and merge.

marioevz:
After a quick chat with Claude about this, I think we could tag the commit that is the base of this PR's commit, e.g. call the tag `static-tests`, and then reference it in the docs as a way to recover these files temporarily for reference when someone wants to improve the ported tests:

```bash
git show static-tests:tests/static/state_tests/stExtCodeHash/codeCopyZero_ParisFiller.yml
# or
git checkout static-tests -- tests/static/state_tests/stExtCodeHash/codeCopyZero_ParisFiller.yml
```

With this command you can easily access the static test source file even though it's no longer in the main branch's head.

This can be done after this PR is merged but wanted to point it out so we don't forget.

leolara:
@marioevz what should we do about the "excluded categories" ?

> Excluded categories
> Three test categories are excluded from compilation because they are
> prohibitively slow to fill:

> stQuadraticComplexityTest
> stStaticCall
> stTimeConsuming

spencer-tb:
> @marioevz what should we do about the "excluded categories" ?
> 
> > Excluded categories
> > Three test categories are excluded from compilation because they are
> > prohibitively slow to fill:
> 
> > stQuadraticComplexityTest
> > stStaticCall
> > stTimeConsuming

We should still have them, but add a `@pytest.mark.slow` to the test function in the ported test:
```python
 @pytest.mark.slow
  def test_something_expensive():
```

leolara:
> The contract bytecode should be run through our evm_bytes converter. You can import it from execution_testing.cli.evm_bytes.

I think from already spencer's code, it was using  execution_testing.cli.evm_bytes.process_evm_bytes_string

leolara:
> > @marioevz what should we do about the "excluded categories" ?
> > > Excluded categories
> > > Three test categories are excluded from compilation because they are
> > > prohibitively slow to fill:
> > 
> > 
> > > stQuadraticComplexityTest
> > > stStaticCall
> > > stTimeConsuming
> 
> We should still have them, but add a `@pytest.mark.slow` to the test function in the ported test:
> 
> ```python
>  @pytest.mark.slow
>   def test_something_expensive():
> ```

Do I manually add to the script that those 3 directories should have that mark?

spencer-tb:
Here is what I can see now, we have 2,180 test files that generate 149,490 test cases across all forks (Cancun to Amsterdam)!

The script now has the following features:
- Op-language bytecode via `evm_bytes` (not raw hex)
- Yul/LLL source as comments (2,088 files)
- `deploy_contract()` for contract deployment
- Transaction defaults, omits `gas_price`, `nonce`, `value`, `to` when they match defaults
- `@pytest.mark.valid_from` / `valid_until`, 396 tests with upper bounds from filler network specs
- Fork-divergent post-states, 106 `_from_<fork>` functions across 88 tests where expected results differ between forks
- `@pytest.mark.exception_test`, 48 tests expecting tx failure
- `@pytest.mark.slow`, 315 tests in the 3 heavy categories
- `@pytest.mark.ported_from`, all tests link back to original filler
- Passes ruff + mypy strict with auto # noqa: E501 patching

marioevz (approved):
Excellent job on this port, we can continue improving the tests based on this issue: #2506

PR #2457, opened 2026-03-09
feat(tests): port goevmlab regression tests for EELS consensus bugs 

### LINKED ISSUE #1482: tests(feat): port testcases for EELS related bugs from geth
1) Port over the state tests highlighted from https://github.com/holiman/goevmlab/pull/140 to python.
2) Assert that the filled fixtures actually catch the bugs on EELS.

spencer-tb:
## 🗒️ Description

Adds tests for bugs in EELS found by goevmlab a while back. 

## 🔗 Related Issues or PRs

Closes https://github.com/ethereum/execution-specs/issues/1482

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
+++ b/tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
@@ -11,7 +11,6 @@ from execution_testing import (
     Address,
     Alloc,
     Bytecode,
-    Environment,
     Initcode,
     Op,
     StateTestFiller,
@@ -259,12 +258,7 @@ class TestTransientStorageInContractCreation:
             ),
         }
 
-        state_test(
-            env=Environment(),
-            pre=pre,
-            post=post,
-            tx=tx,
-        )
+        state_test(pre=pre, post=post, tx=tx)
 
 
 @pytest.mark.ported_from(
@@ -346,4 +340,4 @@ def test_tstore_rollback_on_failed_create(
         caller_address: Account(storage={0: 0, 1: 0}),
     }
 
-    state_test(env=Environment(), pre=pre, post=post, tx=tx)
+    state_test(pre=pre, post=post, tx=tx)
--- a/tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
+++ b/tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
@@ -272,11 +272,11 @@ class TestTransientStorageInContractCreation:
         "https://github.com/holiman/goevmlab/blob/master/examples/tstore_bug-2/main.go",
     ],
 )
-@CreateOpcodeParams.parametrize()
+@pytest.mark.parametrize("create_opcode", [Op.CREATE, Op.CREATE2])
 def test_tstore_rollback_on_failed_create(
     state_test: StateTestFiller,
     pre: Alloc,
-    opcode: Op,
+    create_opcode: Op,
 ) -> None:
     """
     Test TSTORE is rolled back after failed CREATE/CREATE2 initcode.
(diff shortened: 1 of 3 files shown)


tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
```diff
@@ -263,3 +264,69 @@ def test_contract_creation(
...
+        to=caller_address,
+        gas_limit=10_000_000,
+    )
+
+    post = {
+        # Both CREATE2s fail because TSTORE is rolled back
+        caller_address: Account(storage={0: 0, 1: 0}),
+    }
+
+    state_test(env=Environment(), pre=pre, post=post, tx=tx)
```
felix314159: I remember doing this `env=Environment()` myself and being [corrected](https://github.com/ethereum/execution-specs/pull/2021#discussion_r2695379768). Probably would be a good idea to apply that fix throughout all tests here. Edit: Maybe it does not apply here tho, but worth looking into

tests/shanghai/eip3860_initcode/test_initcode.py
```diff
@@ -585,3 +585,56 @@ def test_create_opcode_initcode(
...
+    Regression test for
+    https://github.com/ethereum/execution-specs/issues/914
+
+    CREATE2 is called with initcode size 0x20000 (exceeds
+    MAX_INITCODE_SIZE) and an endowment of 1123123123 (exceeds the
+    contract's zero balance). The initcode size check must take
+    priority over the balance check:
+
+    - Initcode too large: consumes all gas, exits scope (correct).
+    - Insufficient balance: pushes 0, continues execution (wrong).
```
LouisTsai-Csie: my notes:

In the insufficient balance scenario, if there is no initcode-too-large issue, `CREATE` is still triggered but returns 0 since failure. However, it does not consume all remaining gas, so the subsequent `SSTORE` can still be executed.

However, if the initcode is too large, much more gas is consumed, which lead to OOG condition. In that case, the following `SSTORE` becomes unreachable. If the checking order is incorrect, the balance check is executed first, the SSTORE would be unexpectedly executed.

---

Interesting case, could we maybe add a parametrization `inicode_oversize` (`True`, `False`).
- If `inicode_oversize` is `True`, then `STORAGE[1] = 0`
- If not, then `STORAGE[1] = 1`

This helps me better understand the intention of this case. We now know that the initcode size is checked first, wdyt?

tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
```diff
@@ -263,3 +264,69 @@ def test_contract_creation(
...
+    )
+    initcode_bytes = bytes(initcode)
+    initcode_len = len(initcode_bytes)
+
+    caller_code = (
+        Om.MSTORE(initcode_bytes, 0)
+        + Op.SSTORE(0, Op.CREATE2(0, 0, initcode_len, 0))
+        + Op.SSTORE(1, Op.CREATE2(0, 0, initcode_len, 0))
+    )
+    caller_address = pre.deploy_contract(caller_code)
```
LouisTsai-Csie:
```suggestion
    caller_address = pre.deploy_contract(caller_code, storage={0: 1, 1:1})
```

The test intends to verify that transient storage written during `CREATE2` initcode is rolled back when the creation fails. However, it doesn't actually test this behavior.                                                                                                                                     
                                                                                                                                                          
When the first `CREATE2` fails because the returned code exceeds the max code size (`0x600a > 0x6000`), it consumes all remaining gas. This means the second `CREATE2`, which is supposed to prove that the `TSTORE` was rolled back, never executes due to OOG. We can never observe the transient storage behavior of the second `CREATE2` because execution halts before reaching it.
                                                                                                                                                          
This is the evidence: adding non-zero initial storage (`storage={0: 1, 1: 1}`) causes the test to fail, not overwriting the storage to zeros.

LouisTsai-Csie (changes requested):
One question, plz let me know if i understand it wrong!

tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
LouisTsai-Csie: Plus, could we test out both `CREATE` / `CREATE2`:

> I believe the case is of the following
> 1. A does a tstore a value and and tries to create B using CREATE/CREATE2

https://github.com/ethereum/execution-specs/pull/918#issuecomment-2042167064

spencer-tb:
Thanks @LouisTsai-Csie! Addressed the comments.

felix314159 (approved):
tried both
`uv run fill -v -s --clean ./tests/cancun/eip1153_tstore/`
and
`uv run fill -v -s --clean ./tests/shanghai/eip3860_initcode/`
and there are no issues. lgtm, but also give louis another chance to review

tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
spencer-tb: Thanks will address this.

LouisTsai-Csie (approved):
Thanks, looks good from my side

tests/cancun/eip1153_tstore/test_tstorage_create_contexts.py
spencer-tb: Added now! Thx!

PR #2460, opened 2026-03-09
chore(zkevm): Document HashedNode collapse behaviour with test

kevaundray:
## 🗒️ Description

Adds a test to document what happens when a branch has two children and we delete one of them (meaning the branch collapses to its remaining child). If the remaining child is a HashedNode, we raise an error because HashedNodes cannot be witnessed (with the current `_record_witness` code)

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/json_infra/test_incremental_mpt.py
+++ b/tests/json_infra/test_incremental_mpt.py
@@ -444,6 +444,47 @@ class TestPartialWitness:
         with pytest.raises(AssertionError, match="cannot be invalidated"):
             mpt_set(decoded_mpt, b"ba", b"new_c")
 
+    def test_partial_witness_delete_collapses_to_hashed_node(self) -> None:
+        """
+        Delete a key whose sibling is a HashedNode.
+
+        When a branch has two children and one is deleted, the branch
+        must collapse. If the remaining child is a HashedNode (from a
+        partial witness), _collapse_branch cannot merge the nibble into
+        it because its structure is unknown. This currently raises
+        AssertionError.
+        """
+        fake_hash = keccak256(b"some large subtree")
+        hashed_child = HashedNode(_hash=fake_hash)
+
+        # Branch with a leaf at nibble 0 and a HashedNode at nibble 1.
+        # Key b"\x01" -> nibbles [0, 1]: child index 0, rest_of_key [1].
+        branch = MutableBranchNode(
+            children=[None] * 16,
+            value=b"",
+        )
+        branch.children[0] = MutableLeafNode(
+            rest_of_key=Bytes(b"\x01"),
+            value=b"hello",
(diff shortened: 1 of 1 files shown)


tests/json_infra/test_incremental_mpt.py
```diff
@@ -444,6 +444,47 @@ def test_partial_witness_insert_into_hashed_node_fails(self) -> None:
...
+        )
+
+        # Deleting the leaf at nibble 0 leaves only the HashedNode,
+        # triggering a branch collapse. _collapse_branch calls
+        # _record_witness on the remaining child, which fails
+        # because HashedNode cannot be witnessed.
+        with pytest.raises(
+            AssertionError, match="HashedNode cannot be witnessed"
+        ):
+            mpt_set(mpt, Bytes(b"\x01"), b"")
```
kevaundray, line 486: Perhaps we want to add a `mpt_delete` method to make it easier to read in the future

kevaundray:
cc @jsign good ser, mainly adding a test to document an edge case -- doesn't change any behaviour

tests/json_infra/test_incremental_mpt.py
jsign: I think given that today doing `mpt_set` with the default-value is interpreted as deletion, it should be fine. But no strong opinions here.

tests/json_infra/test_incremental_mpt.py
```diff
@@ -444,6 +444,47 @@ def test_partial_witness_insert_into_hashed_node_fails(self) -> None:
         with pytest.raises(AssertionError, match="cannot be invalidated"):
             mpt_set(decoded_mpt, b"ba", b"new_c")
 
+    def test_partial_witness_delete_collapses_to_hashed_node(self) -> None:
```
jsign, line 447: Cool, makes sense. I think there is value in these unit-tests today, mostly to have a more direct/quicker way to cover these particular cases.

Eventually, we'll have the same coverage in the testing framework with EEST tests, so we can trigger the same situation whenever the filling triggers the stateless run indirectly.

At that point, we can re-eval if keeping these tests or not.
Feels more like a question for the STEEL team if the "spec testing" should be thought in isolation with unit-tests depite EEST might be able to also be though as integration tests in a way.

NBD right now -- just thinking out loud.

jsign (approved):

tests/json_infra/test_incremental_mpt.py
kevaundray: Yeah I agree -- I noticed that there were not a lot of unit tests

PR #2461, opened 2026-03-09
fix(testing): stop inserting empty Account in execute empty_account()

chfast:
## 🗒️ Description

The execute backend's `_empty_account()` was inserting `Account(nonce=0, balance=0)` into the pre-state. Since `Account(nonce=0, balance=0)` is falsy (per `Account.__bool__`), the `Alloc.merge()` step during fixture generation silently discarded it — so the insertion had no observable effect, but was conceptually wrong and inconsistent with the filler backend which returns a bare address without touching the pre-state.

This PR removes the unnecessary insertion so both backends behave identically, adds a unit test asserting the address is not in the pre-state after the call, and updates docstrings to clarify that the returned address refers to a nonexistent account.

In a follow-up PR the method will be renamed to `nonexistent_account()` to better reflect its semantics.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

🦔

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -741,18 +741,13 @@ class Alloc(SharedAlloc):
 
     def _empty_account(self) -> Address:
         """
-        Execute implementation of empty account creation.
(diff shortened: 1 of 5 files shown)


marioevz (approved):
Excellent, thanks for this!

PR #2469, opened 2026-03-10
feat(zkevm): Add execution witness bytecode assertions and mutation support

jsign:
## 🗒️ Description
This PR adds a new feature to the testing framework to support `ExecutionWitness` bytecode assertions and mutations.

Before creating more tests about `ExecutionWitness.bytecode` border-cases, we need to provide the test creator a way to check that whatever situation is trying to check does indeed result in the correct `ExecutionWitness.bytecode` that is intended i.e. here we are talking about the execution witness generation in the spec is doing what we want.

The way `ExecutionWitness.bytecode` assertions work is by providing the test writer three main flags:
- `codes_present`: whatever the test author adds in this list must appear in the generated execution witness bytecodes field.
- `codes_absent`: whatever the test author adds in this list must not appear in the generated execution witness bytecodes field.
- `allow_unexpected`: this bool allows being strict in the comparison and only allows bytecode present in the `codes_present`. For example, if the test wants to know that EXACTLY a certain list of bytecodes is present, and nothing else.

The last flag is useful to avoid being dense; it always adds `codes_present` bytecodes that are expected, like system contracts, which always appear.

Apart from adding assertions, we also support arbitrary mutations for the test author. This allows us to create test cases that corrupt an execution witness bytecode field, so in the future, we can check that running these invalid blocks makes the stateless guest fail. So we are also getting a bit ahead on testing framework features for this front.

Both features reg assertions and mutations, follow the same style on how BALs work, so it should feel pretty natural.

In the same way, this PR adds assertions & mutations for bytecodes; we can extend it in the near future to ancestors and MPT nodes. I started with bytecodes since we already have two custom tests that didn't have assertions, so we can use them now (which is also done in this PR). I also have a list of more bytecodes border cases tests that I want to create, so I want to have the assertions ready before jumping into adding more tests.

## 🔗 Related Issues or PRs
https://github.com/ethereum/execution-specs/issues/2183

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -88,6 +88,9 @@
     BlockAccessList,
```
jsign, line 88: The main logic flow mirrors the assertion feature from BALs -- recommend looking some lines above line changes to make sense of it.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -776,6 +786,20 @@ def generate_block_data(
...
+        execution_witness = t8n_witness
+        if (
+            block.expected_execution_witness_codes is not None
+            and t8n_witness is not None
+        ):
+            block.expected_execution_witness_codes.verify_against(t8n_witness)
+            execution_witness = (
+                block.expected_execution_witness_codes.modify_if_invalid_test(
+                    t8n_witness
+                )
```
jsign, line 800: As with BALs, this is what allows the test author to influence what is finally put in the fixture to create invalid cases.

packages/testing/src/execution_testing/test_types/execution_witness/expectations.py
```diff
@@ -0,0 +1,172 @@
...
+                    dupes.append(code)
+                seen.add(code)
+            raise ExecutionWitnessValidationError(
+                f"Witness codes contain duplicates: {[c.hex() for c in dupes]}"
+            )
+
+        if actual_codes != sorted(actual_codes):
+            raise ExecutionWitnessValidationError(
+                "Witness codes are not sorted in lexicographic order"
+            )
```
jsign, line 123: There are two "default assertions" backed here: no duplicates and proper sorting.

These assertions are fixed and independent of what the test author decides, since should always make sense. The rest below are checking what the test author asked for.

packages/testing/src/execution_testing/test_types/execution_witness/modifiers.py
```diff
@@ -0,0 +1,52 @@
+"""
```
jsign, line 1: These are the mutators you can chain to modify the correct generated witness for a case, to an invalid one to test that the block validation should fail.

tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes.py
```diff
@@ -6,6 +6,8 @@
     Alloc,
```
jsign, line 6: Improving existing tests to now use assertions -- now we can start being confident that the spec is correct, and also any refactors we do doesn't break stuff.

jsign:
@kevaundray ready for some eyes whenever is a good moment to you.

kevaundray (approved):

PR #2473, opened 2026-03-10
refactor(specs): delay get_code calls in CALL-like opcodes after gas charging and stack-depth checks

jsign:
## 🗒️ Description
Move the `get_code` call from each `CALL*` opcode (`call`, `callcode`,`delegatecall`, `staticcall`) into `generic_call`, removing the `code`parameter from its signature. 

All four callers were doing the identical`get_account(tx_state, code_address).code_hash` →`get_code(tx_state, code_hash)` sequence before passing the result to `generic_call`, which already receives `code_address`.

The motivation for this is two-fold:
- Do the `get_code` call after the `charge_gas` call for CALL-like opcodes -- this avoids pulling bytecode that isn't used if that OOGs.
- Also avoid pulling the bytecode if we could `charge_gas` but we would also fail due to a stack depth overflow, which would also not use pulled bytecodes.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -297,7 +297,6 @@ def generic_call(
     memory_input_size: U256,
     memory_output_start_position: U256,
     memory_output_size: U256,
-    code: Bytes,
     disable_precompiles: bool,
 ) -> None:
     """
@@ -311,6 +310,10 @@ def generic_call(
         evm.gas_left += gas
         push(evm.stack, U256(0))
         return
+
+    tx_state = evm.message.tx_env.state
+    code_hash = get_account(tx_state, code_address).code_hash
+    code = get_code(tx_state, code_hash)
 
     call_data = memory_read_bytes(
         evm.memory, memory_input_start_position, memory_input_size
@@ -424,9 +427,6 @@ def call(evm: Evm) -> None:
         if code_address not in evm.accessed_addresses:
             evm.accessed_addresses.add(code_address)
 
-    code_hash = get_account(tx_state, code_address).code_hash
-    code = get_code(tx_state, code_hash)
-
     message_call_gas = calculate_message_call_gas(
         value,
         gas,
@@ -456,7 +456,6 @@ def call(evm: Evm) -> None:
             memory_input_size,
             memory_output_start_position,
             memory_output_size,
-            code,
             is_delegated,
         )
 
@@ -527,9 +526,6 @@ def callcode(evm: Evm) -> None:
         if code_address not in evm.accessed_addresses:
             evm.accessed_addresses.add(code_address)
 
-    code_hash = get_account(tx_state, code_address).code_hash
-    code = get_code(tx_state, code_hash)
-
     message_call_gas = calculate_message_call_gas(
         value,
         gas,
@@ -561,7 +557,6 @@ def callcode(evm: Evm) -> None:
             memory_input_size,
             memory_output_start_position,
             memory_output_size,
-            code,
             is_delegated,
         )
 
@@ -666,7 +661,6 @@ def delegatecall(evm: Evm) -> None:
     check_gas(evm, access_gas_cost + extend_memory.cost)
 
     # STATE ACCESS
-    tx_state = evm.message.tx_env.state
     if is_cold_access:
         evm.accessed_addresses.add(code_address)
 
@@ -683,9 +677,6 @@ def delegatecall(evm: Evm) -> None:
         check_gas(evm, extra_gas + extend_memory.cost)
         if code_address not in evm.accessed_addresses:
             evm.accessed_addresses.add(code_address)
-
-    code_hash = get_account(tx_state, code_address).code_hash
-    code = get_code(tx_state, code_hash)
 
     message_call_gas = calculate_message_call_gas(
         U256(0),
@@ -711,7 +702,6 @@ def delegatecall(evm: Evm) -> None:
         memory_input_size,
         memory_output_start_position,
         memory_output_size,
-        code,
         is_delegated,
     )
 
@@ -756,7 +746,6 @@ def staticcall(evm: Evm) -> None:
     check_gas(evm, access_gas_cost + extend_memory.cost)
 
     # STATE ACCESS
-    tx_state = evm.message.tx_env.state
     if is_cold_access:
         evm.accessed_addresses.add(to)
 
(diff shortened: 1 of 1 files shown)


jsign:
cc @kevaundray

kevaundray:
This makes sense to me CC @gurukamath (for thoughts)

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -312,6 +311,10 @@ def generic_call(
         push(evm.stack, U256(0))
         return
 
+    tx_state = evm.message.tx_env.state
```
gurukamath, line 314: @fselmo This will be significant from a BAL perspective. This change means the account is no longer accessed if the stack depth limit is reached. The fact that this passes all the tests tells me that there are no tests for this right now. Would love to get your thoughts

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -312,6 +311,10 @@ def generic_call(
...
+    code_hash = get_account(tx_state, code_address).code_hash
```
gurukamath, line 315: Another thing to keep in mind is a potential interaction with EIP-2780 which is in CFI status for amsterdam. In that EIP, the gas charge is closely coupled with whether an account has code or not. So we will need to access the account at the gas charging stage itself.

IMO, this point is not a huge concern right now but if 2780 gets SFI'd, we'll have to move the `get_account` call, back to where it currently is.

gurukamath (commented):
Is the concern here related to when the account access happens? Or is it when the code access happens?

If it is the latter, maybe we could just delay the code access and pass the `code_hash` to the `generic_call` instead of the `code`? That will enable us to keep the account access closely coupled with its gas charging and warming.

src/ethereum/forks/amsterdam/vm/instructions/system.py
fselmo: Sorry I missed this. How so? The account is accessed after `check_gas` for both `is_account_alive()` as well as a check on whether it is delegated or not within `calculate_delegation_cost` in all CALL-like opcodes. Lmk if I missed something here 👀.

This delegation check is also a `get_code` check so are you accounting for that in these changes @jsign? We perform `get_code` there. In fact we have to in order to charge for delegation costs.

fselmo: Ah wait no sorry this is the delegated address then at this point which is the `code_address`. Hmm yeah I will take a look into this

fselmo: Yeah this only affects the BAL for a delegated address being touched or not. This means the original address is _always_ in the BAL (touched at `is_account_alive` and delegation check) but the _delegated_ address is conditionally in the BAL depending on stack too deep.

I don't believe this has been specified in EIP-7928 tbh but I wouldn't be against this. We could clarify this delegated code path in the EIP. I wonder, adding a test for this now, how many clients would disagree on it. I would've imagined we had stack overflow tests but since this is specific for EIP-7702, maybe this is a missing case indeed.

@nerolation any strong opinions here?

jochem-brouwer:
Which stack overflow does this refer to? Is this the operand stack or is this the call stack?

fselmo:
> Which stack overflow does this refer to? Is this the operand stack or is this the call stack?

This is the call stack (message depth) check

src/ethereum/forks/amsterdam/vm/instructions/system.py
jochem-brouwer: Note that this is about the call depth maximum of 1024, which since the [63/64 max gas forward rule](https://eips.ethereum.org/EIPS/eip-150) is unreachable for any practical gas limit, and since EIP-7825 it is not reachable in a transaction anymore. To go a call level deeper, always have to pay the base costs of a CALL opcode which is 100, which makes the call depth impossible to reach with EIP-7825 active, and for networks without it active the gas limit has to be at unpractical limits (I should actually figure out the precise value of this at some point).
I don´t think it influences 7702-delegated accounts specifically as this does not increase the call depth (in the EVM) by two, it creates a new call frame so it "just" increases the call depth by one (except it now accesses two accounts). 

Here is some small thing to think about though: the scenario with BALs on top of Fusaka makes it impossible to reach this call depth so we cannot test this scenario. However if at some point in the future we raise the max transaction gas or we remove the tx gas limit (doubt) then it is reachable again. But I think this is the analysis which should then be taken if such EIP is introduced to see what interactions it has.

fselmo: Yeah agreed, good points. This is a safe change imo and we similarly don't cover these test cases in create tests as outlined [here](https://github.com/fselmo/execution-specs/blob/633a8ab2610f91602f6c5de6709f62fa801d50a1/tests/berlin/eip2929_gas_cost_increases/test_create.py#L9-L14).

> I don´t think it influences 7702-delegated accounts specifically as this does not increase the call depth (in the EVM) by two, it creates a new call frame so it "just" increases the call depth by one (except it now accesses two accounts).

This wasn't my worry. My worry is that the touching of `code_address` was done after the stack depth call and at this point only if `code_address` was delegated was it not already touched previously in the code path (included in the BAL). If it was not a delegated address it is touched beforehand so it's always in the BAL, etc.

But as you mentioned this is a safe change regardless due to tx gas limit cap 👍🏼. Thanks for the thoughts @jochem-brouwer :)

fselmo (dismissed):
lgtm but I'll defer to @gurukamath for a final review since he's assigned himself a review here.

gurukamath:
> lgtm but I'll defer to @gurukamath for a final review since he's assigned himself a review here.

Yeah. Looks good to me too but I'd still want to wait for what might happen with EIP-2780 just because that EIP tightly couples account access gas accounting with actual account access.

gurukamath:
Update: A recent proposed [update](https://github.com/ethereum/EIPs/pull/11645) to EIP-2780 gets rid of the two level gas charges for accounts with and without code. If it is decided that we go with this, we can go ahead and merge this PR. However, I am just going to go wait for that decision.

jangko:
I would love to see this PR merged. In nimbus EL, we have to create an ugly workaround specifically for this case to generate correct BAL output where the delegated target still appears in BAL when the sender have insufficient fund to let the call happen. If the get_code is called after those checks, not only reduce BAL output but also avoid unnecessary state access.

https://github.com/status-im/nimbus-eth1/pull/4240

gurukamath (approved):
LGTM!

Issue #2436, opened 2026-03-06
Existing contracts deployed via deterministic CREATE2 factory are still being deployed


jochem-brouwer:
When filling the deep storage benchmark we noticed that even when the deterministic contracts were deployed already,
https://github.com/ethereum/execution-specs/blob/446eabfb62595a06e5f292751ffe9f082f434d12/tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py#L308, 
EEST was still deploying them. This is because there is no check 

https://github.com/ethereum/execution-specs/blob/forks/amsterdam/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py#L339-L351

to pre-check if the target contract already exists. This check is being done for the deterministic factory contract. 

This is a blocker because filling these deep contracts takes a very long time and this bug therefore makes filling very slow, we have now disabled this test for the benchmarks.

marioevz:
This should be the check:
https://github.com/ethereum/execution-specs/blob/525810670fd06f1ff228f8514e41a76c340def78/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py#L324-L325

It's not working then. Could you share some logs for this?

marioevz:
I have some theories for the culprit:
- This function generates deterministic addresses based on the assumption that the contracts were deployed with the factory at `0x4E59B44847B379578588920CA78FBF26C0B4956C`, if that's not the case for the contracts in bloatnet, that's the reason.
- Maybe initcode does not match

Could you verify that the contracts in bloatnet do match the addresses in the `contract_address` fields in the mined assets file, e.g. https://raw.githubusercontent.com/CPerezz/worst_case_miner/refs/heads/master/mined_assets/s10_acc3.json

jochem-brouwer:
Let me summon @CPerezz here 😄 👍 

Was not aware that the check is already there - so should indeed be something else going wrong here.

CPerezz:
@marioevz I made sure that the initcode matched the factories. And the factories were deployed by the scenario.

The config for spamoor was:
```
seed: storage-trie-brancher_v2-460466 # seed for the wallet
refill_amount: 5000000000000000000 # refill 5 ETH when
refill_balance: 1000000000000000000 # balance drops below 1 ETH
refill_interval: 600 # check every 10 minutes
[...]

PR #2475, opened 2026-03-11
feat(test-execute): Defer pre-alloc methods remote checks to batch them

marioevz:
## 🗒️ Description

Adds on-chain checks to lists to defer them in order to be able to batch them instead of trying to immediately resolve.

Affected methods:
- Deterministic contract code checks and deployment in `_deterministic_deploy_contract`
- Balance checks in `_fund_address`
- Stub code checks in `_deploy_contract`

Reduces execution time significantly in tests that do heavy use of these functions.

## 🔗 Related Issues or PRs
PR #2436 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -1,5 +1,6 @@
 """Pre-allocation fixtures used for test filling."""
 
+from dataclasses import dataclass
 from itertools import count
 from pathlib import Path
 from random import randint
@@ -20,6 +21,9 @@ from execution_testing.base_types import (
     Number,
     Storage,
     StorageRootType,
+)
+from execution_testing.base_types import (
+    Alloc as BaseAlloc,
 )
 from execution_testing.base_types.conversions import (
     BytesConvertible,
@@ -225,6 +229,36 @@ class PendingTransaction(Transaction):
     value: HexNumber | None = None  # type: ignore
 
 
+@dataclass
+class _DeferredDeterministicDeploy:
+    """Descriptor for a deferred deterministic contract deployment."""
+
+    contract_address: Address
+    deploy_code: Bytes
+    salt: Hash
+    initcode: Bytes | Initcode
+    label: str | None
+    deploy_gas_limit: int
+
+
+@dataclass
+class _DeferredStubCheck:
+    """Descriptor for a deferred stub contract account fetch."""
+
+    contract_address: Address
+    stub: str
+    label: str | None
+
+
+@dataclass
+class _DeferredFundAddress:
+    """Descriptor for a deferred address funding balance check."""
+
+    address: Address
+    amount: int
+    minimum_balance: bool
+
+
 class Alloc(SharedAlloc):
     """A custom class that inherits from the original Alloc class."""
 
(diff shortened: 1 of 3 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -435,20 +455,16 @@ def _deploy_contract(
...
-            balance = account.balance
-            nonce = account.nonce
-            bal_eth = balance / 10**18
-            logger.debug(
-                f"Stub contract {contract_address}: balance={bal_eth:.18f} "
-                f"ETH, nonce={nonce}, code_size={len(code)} bytes"
             )
-            self.__internal_setitem__(contract_address, account)
+            contract_address.label = label
             return contract_address
```
LouisTsai-Csie, line 474: Do we need to add this here? As it does in `_fund_address` and `_deploy_contract_deterministic`

```python
self.__internal_setitem__(contract_address, Account())
```

My understanding is: without this, the address would not be added to the `Alloc`, the existence guard, like the one in `fund_address` would not block the stub address.

LouisTsai-Csie (approved):
Thanks! Leave one comment, and Kamil reports running the test generation successfully, i think we are good to merge.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
LouisTsai-Csie: This check under `fund_address`:

```python
if address in self:
  raise Exception(
      "Cannot fund an account already in state. "
      "Use the appropriate `amount`, `balance` arguments "
      "when creating the account."
  )
```

felix314159:
it would be good to add a few unit tests here, they should verify that deferred stub and funding operations still reserve addresses immediately, resolve placeholders into the correct final alloc state and pending transactions, and that the new batch RPC helpers handle empty inputs, preserve request order, and parse responses correctly

felix314159:
i think the main concern is that with the current PR we would delay the remote checks correctly, but in doing so we also delay marking some addresses as 'already used' in the local alloc. so checks that previously used to fail immediately like trying to fund an address that was already introduced through a stub or a previous `fund_address()` call, may now silently pass until much later and that might be hard to debug (code that should be rejected immediately may now be accepted and produce the wrong setup state or extra funding transactions)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
marioevz: Added in the latest commit, good catch!

marioevz:
@felix314159 agreed here, thanks for this, I'll time box trying to introduce execute unit tests in this PR, and otherwise will create a tracking issue for adding these.

marioevz:
Created a follow up PR to introduce unit tests to execute remote: https://github.com/ethereum/execution-specs/pull/2485

It's in draft because is currently failing.

marioevz:
@felix314159 addressed your concern in the first comment and execute now adds the account to the alloc when using `fund_address`, which was the previous behavior. Also added a test to #2485 to verify this behavior.

felix314159 (approved):
thanks!

PR #2472, opened 2026-03-10
feat(zkevm): add test for execution witness bytecodes for contract creation, CALL-variants and empty block

jsign:
## 🗒️ Description
This PR adds more test coverage for `ExecutionWitness.codes` cases, in three main groups:
- Empty block: in this case we do a strict check that only system contracts bytecodes are present (and nothing else).
- CALL-like opcodes: for multiple CALL-variants check that target bytecodes work as expected.
- Contract creation: test contract creation via tx creation, CREATE and CREATE2.

This PR isn't targeting full coverage of codes in witness, more PRs will come soon. Just opening PRs in parts to avoid them being too big and be easy to review.

```
> uv run fill --clean -m "blockchain_test" --fork Amsterdam ./tests/amsterdam/eip8025_optional_proofs/
========================================================== test session starts ==========================================================
platform linux -- Python 3.12.3, pytest-8.4.2, pluggy-1.6.0
Generating fixtures for: Amsterdam, BPO2ToAmsterdamAtTime15k
2.19.0
Log file: logs/fill-20260310-222215-main.log
rootdir: /home/ignacio/code/execution-specs
configfile: packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-fill.ini
plugins: cov-4.1.0, metadata-3.1.1, json-report-1.5.0, xdist-3.8.0, regex-0.2.0, custom-report-1.0.1, html-4.1.1
collecting ...  pytest-regex selected 50 tests to run for regex: .*
collected 75 items / 25 deselected / 50 selected                                                                                        

tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_call_variants.py .............                                    [13/25]
tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_contract_creation.py ...........                                  [24/25]
tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_system_contracts.py .                                             [25/25]

=========================================================== warnings summary ============================================================
.venv/lib/python3.12/site-packages/_pytest/config/__init__.py:812
  /home/ignacio/code/execution-specs/.venv/lib/python3.12/site-packages/_pytest/config/__init__.py:812: PytestAssertRewriteWarning: Module already imported so cannot be rewritten; execution_testing.cli.pytest_commands.plugins.filler.pre_alloc
    self.import_plugin(arg, consider_entry_points=True)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
[...]

PR #2479, opened 2026-03-11
fix(zkevm): optimize further bytecode in witness

jsign:
## 🗒️ Description
This PR changes the specs on how created code is observed by the state tracker thus influencing when they appear in the execution witness. For more context, please see this [PR comment](https://github.com/ethereum/execution-specs/pull/2472#discussion_r2914808882).

The spec now only adds bytecode to `code_reads` if it is read from the pre-state, leveraging the fact that the bytecode might have been observed during execution by previous transactions in the block.

More concretely, imagine the case that tx1 CREATE some bytecode with `code_hash_1`, and a later tx2 accesses a separate account with `code_hash_1`:
- Before this PR: the bytecode for `code_hash_1` was part of the witness, since the tx2 code read was unconditionally added.
- After this PR: we leverage `code_writes` which happened earlier in the block, thus in this corner-case we would avoid adding it to the witness since it is already part of the "in memory" guest program bytecode dataset.

This still leaves one even more border case with the "inefficiency" which is that the tx1 contract creation reverts. The way `code_write` works is that it rolls back too, so tx2 can't leverage it. I think this could also be solved, but it will probably make the specs more complex, which I'm not sure is worth it. (An extra test was added to be sure clients behave identically in that case).

I'm opening this PR mainly to tighten up a bit more bytecode optimality in the witness since this change in the spec is quite simple -- I think going the last mile is probably not worth the complexity. 

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_contract_creation.py
+++ b/tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_contract_creation.py
@@ -293,16 +293,11 @@ def test_witness_codes_create_same_hash_then_read(
     """
     Tx1 deploys a contract, tx2 calls a pre-state contract with same code.
 
-    The pre-state contract's bytecode should appear in
-    executionWitness.codes because get_witness_codes checks per-address
-    pre-state match. The pre-state contract qualifies regardless of the
-    earlier CREATE writing the same hash.
-
-    TODO(zkevm): it might be worth reconsidering how the execution witness
-    generation handles this case, since technically this same bytecode was
-    observed in tx1 thus can in theory be exploited to avoid including it
-    for tx2. Feels inconsistent with the other test
-    test_witness_codes_create_then_call_same_block(...).
+    The pre-state contract's bytecode must not appear in
+    executionWitness.codes because the same code hash was already
+    written by tx1's CREATE.  A stateless verifier observed
+    the bytecode from the CREATE transaction data, so including it
+    in the witness is redundant.
     """
     runtime_code = bytes(Op.STOP)
 
@@ -330,7 +325,7 @@ def test_witness_codes_create_same_hash_then_read(
                 txs=[tx1_create, tx2_read],
                 expected_execution_witness_codes=(
                     ExecutionWitnessCodesExpectation(
-                        codes_present=[Bytes(runtime_code)],
+                        codes_absent=[Bytes(runtime_code)],
                     )
                 ),
             )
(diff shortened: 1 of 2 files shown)


tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_contract_creation.py
```diff
@@ -330,7 +325,7 @@ def test_witness_codes_create_same_hash_then_read(
                 txs=[tx1_create, tx2_read],
                 expected_execution_witness_codes=(
                     ExecutionWitnessCodesExpectation(
-                        codes_present=[Bytes(runtime_code)],
+                        codes_absent=[Bytes(runtime_code)],
```
jsign, line 328: This PR now changes the expectation of this existing test as expected.

tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_contract_creation.py
```diff
@@ -493,3 +488,72 @@ def test_witness_codes_failed_create_after_initcode_read(
             creator: Account(nonce=1),
         },
     )
+
+
+def test_witness_codes_reverted_create_same_hash_then_read(
```
jsign, line 493: This is a new test to surface an even more twisted border case that I mentioned in the PR description.

jsign:
cc @kevaundray to get your feedback whenever time allows
I would say to see this PR as a potential proposal -- just sharing a bit how this change would look like.

kevaundray:
> This still leaves one even more border case with the "inefficiency" which is that the tx1 contract creation reverts. The way code_write works is that it rolls back too, so tx2 can't leverage it

Ah yeah I think not adding complexity to address this make sense -- the rule that if its in code_writes, then we don't include it in the witness seems pretty simple already imo. 

I don't have a strong opinion on whether we add this in though, mainly think its good to document it and be consistent, so would lean on your judgement here

jsign:
> > This still leaves one even more border case with the "inefficiency" which is that the tx1 contract creation reverts. The way code_write works is that it rolls back too, so tx2 can't leverage it
> 
> Ah yeah I think not adding complexity to address this make sense -- the rule that if its in code_writes, then we don't include it in the witness seems pretty simple already imo.
> 
> I don't have a strong opinion on whether we add this in though, mainly think its good to document it and be consistent, so would lean on your judgement here

Yeah, I think this PR should be okay to merge. Feels like a low-hanging fruit that makes things more optimal. The extra step of solving the other case I think requires someone else with a much stronger argument to justify that complexity, which I don't think is worth it (i.e. in favor of keeping the specs simple).

Let's merge this, and the current tests will surface this situation to ELs and we might get feedback indirectly by them running into these cases.

PR #2487, opened 2026-03-12
fix(test-benchmark): update `approve` function for `SSTORE` benchmark

LouisTsai-Csie:
## Background

After the first round of `SSTORE` benchmark data collection, we found that `test_sstore_erc20_mint` is unreliable — it cannot produce the non-cached `SSTORE` scenario because there is always one `SLOAD` before `SSTORE` is executed. We switched to `test_sstore_erc20_approve`, which we verified has no `SLOAD` along the execution path by inspecting the [[execution trace](https://evmole.xyz/#0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/eth/control_flow)](https://evmole.xyz/#0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2/eth/control_flow) (see function selector `0x095ea7b3`).

## Changes

This PR ensures the benchmark meets the requirements by addressing two issues:

1. Correct slot interaction via account delegation

In the bloated ERC20 contract, the bloating logic updates allowances as `allowance[msg.sender][i] = i`:

```python
function bloatStorage(uint256 startSlot, uint256 numAddresses) external {
    uint256 endSlot = startSlot + numAddresses;
    unchecked {
        for (uint256 i = startSlot; i < endSlot; i++) {
            address targetAddr = address(uint160(i));
            balanceOf[msg.sender] -= i;
            balanceOf[targetAddr] += i;

            // This is equivalent to allowance[sender][i] = i
            allowance[msg.sender][targetAddr] = i;
        }
    }
    nextStorageSlot = endSlot;
    emit StorageBloated(startSlot, endSlot, numAddresses * 2); // 2 slots per address
}
```

Here, `msg.sender` is the account that initiated the bloating transaction (denoted as `sender`), so the bloated slots are of the form `allowance[sender][i]`.

To touch these same slots via `approve`, in the `test_sstore_erc20_approve` benchmark, we must use the private key of this `sender` and delegate the account to the contract that batch-calls the `approve` function.

2. Missing `SSTORE` benchmark configurations

Two new configurations are added:

- **`write_new_value`**: Controls whether the slot value is overwritten with a new value.
- **`existing_slot`**: Controls whether the slot is initialized before the benchmark runs.

By inspecting the bloatnet account that initiated the bloating process, we found that account `0x6c2624662dC514A69955E9A6a195ECe23AcED5a0` bloats slots from 371 to 45,139 (~44,770 slots). Therefore, we set the starting slot to 371 when touching existing slots.

For the "write same value" scenario: since the bloating pattern is `allowance[sender][i] = i`, the current value stored in each slot equals its slot offset. In the benchmark, we fetch the value via `MLOAD(32)`, where `MEM[32]` originates from `CALLDATA[32]` (the slot offset), which increments with each iteration — thus writing the same value back.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -34,6 +34,7 @@ from execution_testing import (
     While,
     keccak256,
 )
+from execution_testing.base_types.base_types import Number
 
 from tests.benchmark.stateful.helpers import (
     ALLOWANCE_SELECTOR,
@@ -273,6 +274,8 @@ def test_sload_erc20_balanceof(
 
 @pytest.mark.parametrize("cache_strategy", list(CacheStrategy))
 @pytest.mark.parametrize("token_name", SSTORE_TOKENS)
+@pytest.mark.parametrize("write_new_value", [False, True])
+@pytest.mark.parametrize("existing_slot", [True, False])
 def test_sstore_erc20_approve(
     benchmark_test: BenchmarkTestFiller,
     pre: Alloc,
@@ -280,9 +283,21 @@ def test_sstore_erc20_approve(
     gas_benchmark_value: int,
     tx_gas_limit: int,
     token_name: str,
+    write_new_value: bool,
+    existing_slot: bool,
     cache_strategy: CacheStrategy,
 ) -> None:
     """Benchmark SSTORE using ERC20 approve on bloatnet."""
+    sender = EOA(
+        key=0x90655B60A56E01389173510C3DDF1B969803A9F41844BF344CBF3D1AC3A80845,
+        # nonce fetched from the pre-deployed state on the network
+        nonce=45344,
+    )
+    pre.fund_address(address=sender, amount=10**9)
+    assert Address(sender) == Address(
+        0x6C2624662DC514A69955E9A6A195ECE23ACED5A0
+    ), "EOA address mismatch"
+
     # Stub Account
     erc20_address = pre.deploy_contract(
         code=Bytecode(),
@@ -309,7 +324,14 @@ def test_sstore_erc20_approve(
         + Op.CALLDATALOAD(0)  # [num_calls]
     )
 
-    call_approve = Op.MSTORE(64, Op.MLOAD(32)) + Op.POP(
+    value_to_write: Bytecode | int
+    if write_new_value:
+        # non-zero, avoids refund and always differs from existing values
+        value_to_write = 1
+    else:
+        value_to_write = Op.MLOAD(32) if existing_slot else 0
+
+    call_approve = Op.MSTORE(64, value_to_write) + Op.POP(
         Op.CALL(
             address=erc20_address,
             value=0,
@@ -322,6 +344,7 @@ def test_sstore_erc20_approve(
         )
     )
 
+    cache_warmup = Bytecode()
     if cache_strategy == CacheStrategy.CACHE_TX:
         # Call allowance(ADDRESS, spender) to warm the allowance
         # storage slot that approve will later write to.
@@ -346,8 +369,6 @@ def test_sstore_erc20_approve(
             + Op.MSTORE(0, APPROVE_SELECTOR)
             + Op.MSTORE(32, Op.MLOAD(64))
         )
-    else:
-        cache_warmup = Bytecode()
 
     loop = While(
         body=cache_warmup
@@ -364,6 +385,22 @@ def test_sstore_erc20_approve(
     setup_cost = setup.gas_cost(fork)
     loop_cost = loop.gas_cost(fork)
     access_list = [AccessList(address=erc20_address, storage_keys=[])]
+    # Delegate sender to attack_contract_address once in setup
+    delegation_tx = Transaction(
+        gas_limit=100_000,
+        to=sender,
+        value=0,
+        sender=pre.fund_eoa(),
+        authorization_list=[
+            AuthorizationTuple(
+                chain_id=0,
+                address=attack_contract_address,
+                nonce=sender.nonce,
+                signer=sender,
+            ),
+        ],
+    )
+    sender.nonce = Number(sender.nonce + 1)
     intrinsic_gas_with_access_list = (
         fork.transaction_intrinsic_cost_calculator()(
             access_list=access_list,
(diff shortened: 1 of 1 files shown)


tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -309,7 +319,9 @@ def test_sstore_erc20_approve(
         + Op.CALLDATALOAD(0)  # [num_calls]
     )
 
-    call_approve = Op.MSTORE(64, Op.MLOAD(32)) + Op.POP(
+    value_to_write = 0 if not write_new_value else Op.MLOAD(32)
+
+    call_approve = Op.MSTORE(64, value_to_write) + Op.POP(
```
jochem-brouwer, line 335: If `write_new_value` is `False` it will write 0, i.e. it will delete the key, is this intended?

LouisTsai-Csie: The allowance is bloated as `allowance[sender][i] = i`, if we want to write the same value, we set it to `MLOAD(32)`, if not, we could reset it to 0 or something that is not `MLOAD(32)`

Not sure what would be the best approach here, as writing to zero would lead to refund.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -462,7 +491,7 @@ def test_sstore_erc20_approve(
     txs = []
     cache_txs = []
     gas_remaining = gas_benchmark_value
-    slot_offset = 0
+    slot_offset = 371 if existing_slot else START_SLOT
```
LouisTsai-Csie, line 507: We are using this account (0x6c2624662dC514A69955E9A6a195ECe23AcED5a0) for testing, and the smallest nonzero slots is 371.

Image

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -346,8 +359,6 @@ def test_sstore_erc20_approve(
             + Op.MSTORE(0, APPROVE_SELECTOR)
             + Op.MSTORE(32, Op.MLOAD(64))
         )
```
LouisTsai-Csie, line 372: Please take a closer look at this caching warm up, i did review it again to make sure nothing breaks due to 7702 authorization, but an extra eye on this would be helpful.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
spencer-tb: I think its also inverted, based on the labels, currently we have:
- write_new_value=True, Op.MLOAD(32), writes the value that's already there, not a new value
- write_new_value=False, 0, writes something different. is a new value

What about:
```python
if write_new_value:
      value_to_write = 1 # non-zero, avoids refund and always differs from existing values
  else:
      value_to_write = Op.MLOAD(32) if existing_slot else 0
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
spencer-tb, line 507: Adding a comment here or a named constant for 371 would be nice

spencer-tb (approved):
Just chiming in! LGTM otherwise

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -273,16 +274,26 @@ def test_sload_erc20_balanceof(
...
+    existing_slot: bool,
     cache_strategy: CacheStrategy,
 ) -> None:
     """Benchmark SSTORE using ERC20 approve on bloatnet."""
+    pkey = 0x90655B60A56E01389173510C3DDF1B969803A9F41844BF344CBF3D1AC3A80845
+    sender = pre.fund_eoa(key=pkey)
+    assert Address(sender) == Address(
+        0x6C2624662DC514A69955E9A6A195ECE23ACED5A0
+    ), "EOA address mismatch"
+
```
marioevz: Just do:
```suggestion
    sender = EOA(key=0x90655B60A56E01389173510C3DDF1B969803A9F41844BF344CBF3D1AC3A80845)
    pre.fund_address(address=sender, amount=<some reasonable amount>)
    assert Address(sender) == Address(
        0x6C2624662DC514A69955E9A6A195ECE23ACED5A0
    ), "EOA address mismatch"
```
to avoid the framework changes in this PR. I think the `pre.fund_eoa` modification might be worth it eventually but it's too deep of a change IMO.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
fselmo: +1 to @spencer-tb comment for clarifying this directly in the code for a reader

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -350,9 +352,12 @@ def _fund_eoa(
         else:
             account = Account()
 
-        account_hash = account.hash()
-        salt = self.get_next_account_salt(account_hash)
-        eoa = eoa_from_hash(account_hash, salt)
+        if key is not None:
+            eoa = EOA(key=key)
```
marioevz: This line breaks assumptions in `modified_accounts_salt`, and can break pre-allocation grouping.

Search for `_hardcoded_addresses_deployed_to` to see how it's handled for contracts deployed to a fixed location.

marioevz (dismissed):
I think the framework modification is not worth the risk for this isolated case where is needed.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: Nice, I would follow this

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: When we use this stub account, configuring it as an EOA object, which sets the nonce to 0 by default. However, the actual nonce for this account is already around 45,345. I need to make sure the nonce is updated automatically based on the live network state. I’ll verify this by running the gas benchmarks.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
LouisTsai-Csie: Updated! Please take a look

spencer-tb (approved):
LGTM!

PR #2491, opened 2026-03-12
feat(zkevm): add execution witness bytecode tests for SELFDESTRUCT and extra CALL-like opcode scenarios

jsign:
## 🗒️ Description
This PR adds tests for SELFDESTRUCT opcodes relevant cases, plus some extra situations in CALL-like opcodes.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_call_variants.py
+++ b/tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_call_variants.py
@@ -111,3 +111,125 @@ def test_witness_codes_nested_calls(
             sender: Account(nonce=1),
         },
     )
+
+
+def test_witness_codes_dedup_identical_bytecode(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+) -> None:
+    """
+    Two pre-state contracts with identical bytecode, both CALLed.
+
+    Only one copy of the bytecode should appear in executionWitness.codes
+    because get_witness_codes() deduplicates by code hash.
+    """
+    shared_code = Op.PUSH1(0x42) + Op.POP + Op.STOP
+    contract_a = pre.deploy_contract(code=shared_code)
+    contract_b = pre.deploy_contract(code=shared_code)
+
+    caller_code = (
+        Op.CALL(Op.GAS, contract_a, 0, 0, 0, 0, 0)
+        + Op.POP
+        + Op.CALL(Op.GAS, contract_b, 0, 0, 0, 0, 0)
+        + Op.POP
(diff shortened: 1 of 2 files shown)


tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_call_variants.py
```diff
@@ -111,3 +111,133 @@ def test_witness_codes_nested_calls(
...
+    sender = pre.fund_eoa()
+    tx = Transaction(sender=sender, to=caller, gas_limit=500_000)
+
+    blockchain_test(
+        pre=pre,
+        blocks=[
+            Block(
+                txs=[tx],
+                expected_execution_witness_codes=(
+                    ExecutionWitnessCodesExpectation(
```
jsign, line 149: The "deup" assertion is backed in in the `ExecutionWitnessCodesExpectation` -- so this test is more about forcing the situation in the block so this default-assertion can do the check (i.e. not an explicit assertion that can be done by the test writer).

tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_selfdestruct.py
```diff
@@ -0,0 +1,370 @@
...
+                txs=[tx],
+                expected_execution_witness_codes=(
+                    ExecutionWitnessCodesExpectation(
+                        codes_present=system_codes
+                        + [
+                            Bytes(bytes(caller_code)),
+                            Bytes(bytes(target_code)),
+                        ],
+                        codes_absent=[Bytes(bytes(beneficiary_code))],
+                        allow_unexpected=False,
```
jsign, line 65: I'm thinking of removing `allow_unexpected` and by default always making it behave as `False`.

Mainly because now I'm feeling for many tests `allow_unexpectd=False` is the safest thing to do; and in that case `codes_present` is the only thing you need to provide, since is strict expected set thus `codes_absent` doesn't make sense.

So what I'm thinking is:
- Removing `allow_unexpected`, and default behavior is current `False`.
- By default always add the system contracts in the expected output -- so this removes the "burden" that we wanted to avoid with the current style of `codes_present` and `code_absent`.

Still thinking a bit more about it -- but if I get convinced, I'll do it in a separate PR.

jsign:
cc @kevaundray for some quick eyes whenever time allows

kevaundray:
Seems ruff format is not failing ser

kevaundray (approved):

PR #2496, opened 2026-03-13
refactor(test-benchmark): collect benchmark tests using testpath not marker

### LINKED ISSUE #1616: feat(fill,execute,benchmark): test selection via pytest's `testpaths` instead of markers via `--mode`

LouisTsai-Csie:
## Summary
The current benchmark marker logic is overly complicated. Although there were prior [efforts](https://github.com/ethereum/execution-specs/pulls?page=2&q=is:pr+author:LouisTsai-Csie+is:closed) to improve it, this PR provides another solution, and try to entirely eliminating the markers.

## Background
Currently, all benchmark tests live under the `tests/benchmark` folder. Stateless benchmarks are under the `compute/` subfolder and require the `-m benchmark` flag to run, while stateful benchmarks are under the `stateful/` subfolder and require the `-m stateful` flag. That is:

- Stateless benchmarks: `uv run -v tests/benchmark/compute/ -m benchmark`
- Stateful benchmarks: `uv run -v tests/benchmark/stateful/ -m stateful`

This is redundant, since the tests can already be differentiated by their path, the additional marker flags are unnecessary.

This redundancy is a result of prev technical debt (see more discussion in issue #1616 ). Before the folder restructuring, the layout looked like this:

```
tests/
└── benchmark/
    ├── test_benchmark1.py
    ├── test_benchmark2.py
    ├── test_benchmark3.py
    └── bloatnet/
```

The `bloatnet/` benchmarks are a type of stateful benchmark that require pre-deployed contracts on the bloatnet and interact with those accounts via stub addresses. The `-m stateful` marker was introduced at this time to separate filtering logic:

- All tests under `tests/benchmark/`, **excluding** `tests/benchmark/bloatnet/`, were labeled with the `benchmark` marker.
- All tests under `tests/benchmark/bloatnet/` were labeled with the `stateful` marker.
- Running with the `benchmark` marker would automatically exclude `stateful`-labeled tests, ensuring benchmarks ran successfully without requiring network interaction.

After the folder restructuring, the layout changed to:

```
tests/
└── benchmark/
    ├── compute/
    │   ├── test_benchmark1.py
    │   ├── test_benchmark2.py
    │   └── test_benchmark3.py
    └── stateful/
```

With `bloatnet/` changes to `stateful`, now separated from `compute/`, the test path alone is sufficient to identify which tests to run. However, the old markers were left in place, and the logic grew increasingly complicated, especially after the `repricing` logic was introduced.

## Solution
This PR proposes removing all marker flags from benchmark test invocations, reducing friction for contributors and other teams. A `benchmark` marker is still needed to distinguish benchmark tests from consensus tests, but its scope is simplified.

The changes are as follows:

- Redefine the `benchmark` marker: all tests under `tests/benchmark/`, including both `compute/` and `stateful/`, are now labeled with the `benchmark` marker.
  - In CI (and anywhere consensus tests are run), the `-m not benchmark` flag can still be used to exclude all benchmark tests. This behavior is unchanged.
- Interaction with the `repricing` marker: the `repricing` marker is used to restrict test collection to repricing-analysis tests only. When `-m repricing` is provided, only tests explicitly tagged with `repricing` are collected. This behavior is also unchanged.

In issue #1616 , we explored using distinct `mode` to differentiate benchmark types, but as noted, the test path alone is already sufficiently descriptive. A second [approach](https://github.com/ethereum/execution-specs/issues/1616#issuecomment-3588049927) considered was adding a `pytest ignore` setting so that benchmark tests are excluded by default, requiring the benchmark path to be specified explicitly when needed. However, this introduced problems with documentation generation, for which we could not find a clean way to specify the benchmark tests.
Learning from both experiences, the conclusion is that we need at least one marker to separate benchmark tests from consensus tests, but no additional marker is needed to distinguish between benchmark subtypes (e.g., compute vs. stateful). This is the core motivation behind these changes.

## 🔗 Related Issues or PRs

Closes #1616
- #1616  

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


spencer-tb (commented):
Thanks for adding this, overall a great refactor!

I think its one step too deep on the removals as now when we only want to fill the consensus tests (non benchmark tests) I believe we fill both benchmark and consensus now. We should add back the exclusion logic for benchmark tests within: `tests/benchmark/conftest.py` in my opinion.

Claude pilled dump to add back to the conftest:
```python
# 1. If user explicitly requested benchmarks via -m, keep them                                                                                            
marker_expr = config.getoption("-m", default="")                                                                                                          
if "benchmark" in marker_expr or "repricing" in marker_expr:
    return                                                                                                                                                
                                                                                                                                                        
# 2. If user targeted benchmark dir directly (all items are                                                                                               
#    benchmarks), keep them
if items and all(                                                                                                                                         
    item.get_closest_marker("benchmark") for item in items                                                                                              
):
    return

# 3. Mixed collection (e.g. fill tests/) — exclude benchmarks
items[:] = [
    item                                                                                                                                                  
    for item in items
    if not item.get_closest_marker("benchmark")                                                                                                           
] 
```

The documentation here: `docs/writing_tests/benchmarks.md` still references `-m stateful`, we should remove that too!

LouisTsai-Csie:
@spencer-tb Good catch, this is a better approach. I've updated accordinly.

spencer-tb (approved):
LGTM! Ahh we still have the docs with -m stateful I think?

Could @danceratopz or @fselmo take a gaze for final approve. No merge yet please.

tests/benchmark/conftest.py
```diff
@@ -36,55 +36,27 @@ def pytest_collection_modifyitems(config: Any, items: Any) -> None:
...
 
+    for item in items:
+        if benchmark_dir in Path(
+            item.fspath
+        ).parents and not item.get_closest_marker("benchmark"):
+            item.add_marker(benchmark_marker)
+
+    # If user explicitly requested benchmarks via -m, keep them
     marker_expr = config.getoption("-m", default="")
+    if "benchmark" in marker_expr or "repricing" in marker_expr:
```
fselmo: It seems like this would still trigger if `-m "not benchmark"` was used (same with repricing). I think we'd need to be quite explicit here to make sure we avoid collecting these items if `not` is used. `pytest` would still process the `not` correctly, and so the tests wouldn't _run_... but they would still be collected iiuc.

So I think we can do something like:

```py
benchmark_requested = "benchmark" in marker_expr and "not benchmark" not in marker_expr
repricing_requested = "repricing" in marker_expr and "not repricing" not in marker_expr
if benchmark_requested or repricing_requested:
    return
```

fselmo:
Just a single comment. Any reason this went back to draft? @LouisTsai-Csie 👀

LouisTsai-Csie:
@fselmo It was a draft because i want to collect some feedback on gas lighting call before merging!

tests/benchmark/conftest.py
LouisTsai-Csie: Nice catch, i will check this and add unit test if needed.

packages/testing/src/execution_testing/tools/tests/test_create_preimage_layout.py
```diff
@@ -140,7 +140,7 @@ def test_create_preimage(
     tx = Transaction(
         sender=sender,
         to=contract_address,
-        gas_limit=500_000,
+        gas_limit=1_000_000,
```
danceratopz: Seems out of scope for this PR, is it really required here?

packages/testing/src/execution_testing/tools/tests/test_create_preimage_layout.py
```diff
@@ -228,7 +228,7 @@ def test_create2_preimage(
...
```
danceratopz: Seems out of scope for this PR, is it really required here?

danceratopz (commented):
Niice! Thanks for tackling this logic, @LouisTsai-Csie!!!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
```diff
@@ -904,3 +911,552 @@ def test_fill_variables(
...
+    test_module = paris_tests_dir.join("test_module_paris.py")
+    test_module.write(test_module_paris)
+
+    benchmark_tests_dir = tests_dir.mkdir("benchmark")
+    benchmark_tests_dir.join("conftest.py").write(
+        BENCHMARK_CONFTEST_PATH.read_text()
+    )
+    benchmark_module = benchmark_tests_dir.join("test_module_benchmark.py")
+    benchmark_module.write(test_module_benchmark)
+
```
danceratopz: This is repeated in-test for many tests, could we refactor this out to a shared pytest fixture?

danceratopz (commented):
Niice! Thanks for tackling this logic, @LouisTsai-Csie!!!

This is a fly-by ai-assisted review, will take a deeper look in due course!

Could you also please remove:
1. The `stateful` marker registration in `execute_fill.py`? https://github.com/LouisTsai-Csie/execution-specs/blob/1bfe4c1ab26dbd2686728a5dbdcaacc8fb5f92ee/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py#L117-L120
2. The Stale `stateful` marker in this docstring: https://github.com/LouisTsai-Csie/execution-specs/blob/1bfe4c1ab26dbd2686728a5dbdcaacc8fb5f92ee/tests/benchmark/stateful/bloatnet/test_extcodesize_bytecode_sizes.py#L46 ?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
LouisTsai-Csie: Good idea, let me refactor the repetitive code into fixture.

danceratopz (changes requested):
I think the PR is great, but it's missing docs changes which I think would be really helpful for people to understand how to run these tests :) The PR description is great btw, thanks for that.

In particular here, but please check all the commands in the benchmarking docs:
https://github.com/LouisTsai-Csie/execution-specs/blob/0cb107d2f53664c09cb9bc05e583e98df64c2e22/docs/writing_tests/benchmarks.md?plain=1#L20-L35

I think we should make sure to also mention the `repricing` marker there.

LouisTsai-Csie:
@danceratopz Thanks for the advice, but I'd like to handle the repricing part in a separate PR. It's a big topic that worth an individual PR imo.

danceratopz:
Some further suggestions regarding the marker logic here, @LouisTsai-Csie! 
- https://github.com/LouisTsai-Csie/execution-specs/pull/5

danceratopz:
Btw, I believe this will close #1616, finally!

LouisTsai-Csie:
What a clean approach, this looks really nice, will merge after CI runs.

docs/writing_tests/benchmarks.md
```diff
@@ -23,15 +23,13 @@ There are multiple files under `instruction/`, users can check each file's docst
...
+These tests are located under `./tests/benchmark/stateful`.
 
 ### Compute Benchmarks
 
 Other benchmark tests do not require any pre-state configuration. These benchmarks could be run even without pre-deployed contracts or initialized storage.
 
-These tests are located under `./tests/benchmark/compute.` When running these cases, users should specify the `benchmark` marker like `-m benchmark`, or the test would be ignored, even the path is specified correctly.
-
-**Note:** Using `-m benchmark` under `tests/benchmark/stateful`, or `-m stateful` under `tests/benchmark/compute`, will cause the tests to be ignored. Make sure the user-provided flag matches the directory of the test being executed.
+These tests are located under `./tests/benchmark/compute`.
```
danceratopz, line 39:
```suggestion
These tests are located under `./tests/benchmark/compute`. They are not collected by `fill` by default, in order to fill them, their path must explicitly be provided to `fill` :

```bash
fill tests/benchmark/compute --evm-bin=..
```
```

File: tests/benchmark/conftest.py [hunk 6]
```diff
@@ -32,59 +34,23 @@ def pytest_generate_tests(metafunc: Any) -> None:
...
-        ):
-            items_for_removal.append(i)
-
-    for i in reversed(items_for_removal):
-        items.pop(i)
+    if benchmark_dir not in collection_path.parents and (
+        collection_path != benchmark_dir
+    ):
+        return None
+
```
danceratopz: Just saw that this check is superfluous. `collection_path` is the path that pytest is currently considering collecting and as this `conftest.py` lives in `tests/benchmark/`, `collection_path` will always be a path within `tests/benchmark/`, so the check is redundant.
```suggestion
```

danceratopz (approved):
@LouisTsai-Csie LGTM! Two small nits below, but approving to unblock a merge!

Could the CI fail be due to a change in geth behavior?

docs/writing_tests/benchmarks.md
LouisTsai-Csie: This is a good enhancement. I'd suggest moving it to the beginning of this "benchmark" section though, since the filtering logic isn't limited to compute benchmarks, the same rule applies to stateful benchmarks as well. 

Also, the filtering rule applies to both `fill` and `execute remote`.

```
Benchmark tests are not collected by `fill` and `execute remote`  by default, in order to run them, their path must explicitly be provided to these command:

fill -v tests/benchmark/
execute remote -v tests/benchmark

```

tests/benchmark/conftest.py
```diff
@@ -8,6 +8,8 @@
 
 DEFAULT_BENCHMARK_FORK = "Prague"
 
+pytestmark = pytest.mark.benchmark
```
LouisTsai-Csie: We have logic to set the block gas limit to a relatively high value (1 trillion) under benchmarking mode:
https://github.com/ethereum/execution-specs/blob/c9e90a24afc92799327d66b4e886cabbf4e44f93/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py#L519-L527
However, this setting is not visible to `genesis_environment` when it configures the block gas limit, so the value falls back to 120M, the same as it is under consensus test. As a result, heavy setup benchmarks like `test_unchunkified_bytecode` fail, because the deployment phase constructs blocks with gas consumption much more than 120M, exceeding the limit.

`pytestmark` in `conftest.py` affects `-m` filtering during collection (`pytest` knows about it for selecting/deselecting tests), but it does not attach the marker to individual test nodes in a way that `get_closest_marker()` can find at fixture resolution time. Only `pytestmark` in the test module itself, or
  `item.add_marker()`, makes it visible on the node.

On `forks/amsterdam`, the old `pytest_collection_modifyitems` called `item.add_marker(pytest.mark.benchmark)` on each test item directly. This made the marker visible to `get_closest_marker()`.

LouisTsai-Csie: One possible solution: In `benchmarking.py`, we remove the logic of setting the block gas limit to 1 trillion for benchmark mode, instead, we override the value in `tests/benchmark/conftest.py`

LouisTsai-Csie: Update: this is handled by comparing the test path, instead of relying on the marker approach!

danceratopz (approved):
One small clean-up left: Remove the benchmark marker from the unit tests and then I think we can go all the way and remove its registration: Added here: 
- https://github.com/LouisTsai-Csie/execution-specs/pull/6

tests/benchmark/conftest.py
danceratopz: This is actually really nice! I was concerned it wouldn't be as clean. But I think I even prefer it, thanks!

docs/writing_tests/benchmarks.md
danceratopz: Thanks!

PR #2508, opened 2026-03-14
chore(tests): Add test for exponent padding in modexp precompile

kevaundray:
## 🗒️ Description

Was looking at `0x7b62d5284ea063743e4c5b1817d667ddd601a914dc8b3f547b74127bf453a9e4` on mainnet and was wondering why the modexp operation cost 7M gas for 3072 bits. 

The reason is because the gas cost formula is based on the number of bytes declared and not the number of signifcant bytes. This means that although 2^5 and 2^05 both equal 32, the second is charged more since its two bytes and not 1 byte.

A client that first, strips the padded zeroes would compute the wrong gas formula. I checked that none of the new clients like ethrex do this. If any of the popular clients did this, then the above tx would have caused a chain split.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/byzantium/eip198_modexp_precompile/test_modexp.py
+++ b/tests/byzantium/eip198_modexp_precompile/test_modexp.py
@@ -449,32 +449,6 @@ REFERENCE_SPEC_VERSION = "5c8f066acb210c704ef80c1033a941aa5374aac5"
             ),
             id="truncated_input_4",
         ),
-        # Packed vs padded exponent
-        # The padded form has a higher gas cost due to the larger declared
-        # exponent length in the iteration count formula. Clients should
-        # charges the higher gas for the padded exponent encoding.
-        pytest.param(
-            ModExpInput(
-                base="02",
-                exponent="010001",
-                modulus="ff" * 32,
-            ),
-            ModExpOutput(
-                returned_data="00" * 31 + "02",
-            ),
-            id="exponent-packed-3-bytes",
-        ),
-        pytest.param(
-            ModExpInput(
-                base="02",
-                exponent="00" * 125 + "010001",
-                modulus="ff" * 32,
-            ),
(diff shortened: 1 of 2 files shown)


kevaundray:
Its likely that some test implicitly tests this, I didn't look very hard. If this is the case, then feel free to close this as a duplicate. This mainly adds an explicit test case for it.

LouisTsai-Csie (commented):
These cases are valuable! but I'd suggest moving them into `test_modexp_thresholds.py`. Although that file sits under the `Osaka` fork, it starts from the `Berlin` fork, so it covers both before and after EIP-7883.

https://github.com/ethereum/execution-specs/blob/365433e1ba3d5cc2716db0b08a3cdd6a2f24210a/tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py#L776

The existing `test_modexp` tests do not verify gas usage, they only check whether the call succeeded and whether the output is correct.

kevaundray:
> These cases are valuable! but I'd suggest moving them into `test_modexp_thresholds.py`. Although that file sits under the `Osaka` fork, it starts from the `Berlin` fork, so it covers both before and after EIP-7883.
> 
> https://github.com/ethereum/execution-specs/blob/365433e1ba3d5cc2716db0b08a3cdd6a2f24210a/tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py#L776
> 
> The existing `test_modexp` tests do not verify gas usage, they only check whether the call succeeded and whether the output is correct.

Changed it, I was assuming that those would be put into a Block and so if the gas cost differed then it would be caught by the block hash

LouisTsai-Csie (approved):
LGTM, thanks for this case

PR #2511, opened 2026-03-16
✨ feat(fw,fill): Hive genesis format

raxhvl:
## 🗒️ Description

Tools like benchmarkoor depend on eest for client specific genesis files besides the tests.  This is done via `extract_config` cli involving hive (7 network calls!) and  docker  (8 shell commands!) all to run what amounts to one line: `jq -f mapper.jq genesis.json` with HIVE_* env vars set. 

This made the benchmark workflow slow to iterate on, and coupled to a CI artifact that had to be rebuilt whenever tests changed.

The PR introduces a "generic" genesis file which can be processed by tools like benchmarkoor to produce client specific genesis files.

**Why create a new file?** The env values required by hive's mapper is stored in eest's ruleset.py. This information is never written to disk. Which creates a dependency on eest to generate genesis files. Hive has access to eest via eest simulators. To keep benchmarkoor independent of eest, we simply write the required env information to disk, which is read by jq. I have not looked into it, but this could also decopule eest when running hive.

`hive.json`  = (generic gensis + hive ENV vars) i.e the payload used by `extract_config`

The file is named after hive because its tightly coupled with it and uses its mappers to generate client specific genesis.


felix314159:
The fix is merged, you can rebase

danceratopz (commented):
I really like this approach! Using hive's mapper files which requires a hive clone instead of running hive itself to extract the genesis files is a good solution!

I also think it's cleaner to let the consumer of these files generate the genesis; it's a better separation of concerns and gives a consumer more freedom to select:
1. test content version - which EEST release you consume,
2. transform version - which hive mapper vintage renders the configs,
3. client version - which client image/binary you actually boot.

Wih these changes, this is what a consumer, e.g., benchmarkoor, needs:

1. `jq`
2. hive's `clients/<client>/mapper.jq` files (checked out or vendored, pinned)
3. The ~15 lines of `consume.sh`: split the ingredients file, export the `HIVE_*` env, one `jq -f mapper.jq` per client.

I played around with it here, the required code footprint is minimal:
- https://github.com/danceratopz/execution-specs/tree/hive-genesis-demo/tmp

Will get feedback from @skylenet to ensure this is fine on the benchmarkooor side before proper review.

Issue #2423, opened 2026-03-05
Stateful Tests CI Verification


marioevz:
- **Phase 1 – Basic filling in CI (minimum bar)**
  - Add CI jobs that fill stateful benchmark tests and fail if filling breaks  
  - Focus first on benchmarks without stub accounts to reduce complexity  
  - For stub accounts, use `eth_getCode` RPC + `evm_bytes` to get the stub bytecode into Python

- **Phase 2 – Hive execution integration**
  - Use `execute hive` in CI so we can actually execute them, not just fill

- **Phase 3 – Full workflow automation with Kamil's tooling**  
  - Integrate Kamil's gas-benchmarks payload generation tool into CI  
  - Reproduce his full manual workflow: generate payloads, run benchmarks, validate results  
  - Aim for end-to-end automated verification (generation → execution → validation) to catch bugs early and avoid regressions

Note: Phase 3 seems complex, therefore Phase 2 could be our aim for the following weeks.

PR #2512, opened 2026-03-16
feat(test-fill,test-benchmark): support stub accounts in `fill` via RPC

LouisTsai-Csie:
## 🗒️ Description

One pain point for stateful benchmark debugging has been the requirement to run `execute remote` against bloatnet. Because EELS has no built-in rollback mechanism, this could only be handled through the `gas-benchmarks` tooling, thus leading to the following workflow:

Write tests in EELS → Push to GitHub → Run in gas-benchmarks → Check console logs → Debug

This made it hard to add custom logging or inspect opcode count number during execution.

With this PR, the `fill` command now supports stub accounts, which achieves essentially the same result as the execute-remote stub address approach:

Given a stub address and an RPC endpoint, it fetches the contract code from the network and injects it under the stub address.

This means stateful benchmarks can now be run locally with the contract logic embedded, no more depending on gas-benchmarks logging for debugging.

## 🔗 Related Issues or PRs

This PR achieves "Phase 1" of the verification steps outlined in:
- #2423 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

:spider:

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/remote.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/remote.py
@@ -11,7 +11,6 @@ from execution_testing.test_types.chain_config_types import (
     ChainConfigDefaults,
 )
 
-from ..pre_alloc import AddressStubs
 from .chain_builder_eth_rpc import ChainBuilderEthRPC, TestingRPC
 
 
@@ -25,7 +24,8 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         required=False,
         action="store",
         dest="rpc_endpoint",
-        help="RPC endpoint to an execution client",
+        default=None,
+        help="RPC endpoint to an execution client.",
     )
     remote_rpc_group.addoption(
         "--tx-wait-timeout",
@@ -36,18 +36,6 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         help=(
             "Maximum time in seconds to wait for a transaction to be "
             "included in a block"
-        ),
-    )
-    remote_rpc_group.addoption(
-        "--address-stubs",
-        action="store",
-        dest="address_stubs",
-        default=AddressStubs(root={}),
-        type=AddressStubs.model_validate_json_or_file,
-        help=(
-            "The address stubs for contracts that have already been placed "
-            "in the chain and to use for the test. Can be a JSON formatted "
-            "string or a path to a YAML or JSON file."
         ),
     )
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/remote.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/remote.py
@@ -1,6 +1,5 @@
 """Pytest plugin to run the execute in remote-rpc-mode."""
 
-import os
 from pathlib import Path
 
 import pytest
@@ -11,6 +10,7 @@ from execution_testing.test_types.chain_config_types import (
     ChainConfigDefaults,
 )
 
+from ...shared.helpers import get_rpc_endpoint
 from .chain_builder_eth_rpc import ChainBuilderEthRPC, TestingRPC
 
 
@@ -95,9 +95,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
 def pytest_configure(config: pytest.Config) -> None:
     """Check if a chain ID configuration is provided."""
     # Verify chain ID config is consistent with the remote RPC endpoint
-    rpc_endpoint = config.getoption("rpc_endpoint") or os.environ.get(
-        "RPC_ENDPOINT"
-    )
+    rpc_endpoint = get_rpc_endpoint(config)
     if rpc_endpoint is None:
         pytest.fail(
             "RPC endpoint must be provided with the --rpc-endpoint flag or "
@@ -162,9 +160,7 @@ def rpc_endpoint(request: pytest.FixtureRequest) -> str:
     Return remote RPC endpoint to be used to make requests to the execution
     client.
     """
-    rpc_endpoint = request.config.getoption("rpc_endpoint") or os.environ.get(
-        "RPC_ENDPOINT"
-    )
+    rpc_endpoint = get_rpc_endpoint(request.config)
     assert rpc_endpoint is not None
     return rpc_endpoint
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
@@ -48,9 +48,13 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         "pre_alloc",
         "Arguments defining pre-allocation behavior during test filling.",
     )
-
-    # No options for now
-    del pre_alloc_group
+    pre_alloc_group.addoption(
+        "--rpc-endpoint",
+        action="store",
+        dest="rpc_endpoint",
+        default=None,
+        help="RPC endpoint to an execution client.",
+    )
 
 
 DELEGATION_DESIGNATION = b"\xef\x01\x00"
@@ -62,12 +66,20 @@ class Alloc(SharedAlloc):
 
     _eoa_fund_amount_default: int = PrivateAttr(10**21)
     _account_salt: Dict[Hash, int] = PrivateAttr(default_factory=dict)
+    _stub_accounts: Dict[str, Account] = PrivateAttr(default_factory=dict)
 
     def __init__(
-        self, *args: Any, fork: Fork, flags: AllocFlags, **kwargs: Any
+        self,
+        *args: Any,
+        fork: Fork,
+        flags: AllocFlags,
+        stub_accounts: Dict[str, Account] | None = None,
+        **kwargs: Any,
     ) -> None:
         """Initialize the pre-alloc."""
         super().__init__(*args, fork=fork, flags=flags, **kwargs)
+        if stub_accounts is not None:
+            self._stub_accounts = stub_accounts
 
     def get_next_account_salt(self, account_hash: Hash) -> int:
         """Retrieve the next salt for this account."""
@@ -232,28 +244,32 @@ class Alloc(SharedAlloc):
         label: str | None,
         stub: str | None,
     ) -> Address:
-        """
-        Filler implementation of contract deployment.
-        """
-        del stub
+        """Filler implementation of contract deployment."""
+        if stub is not None:
+            if stub not in self._stub_accounts:
+                raise ValueError(
+                    f"Stub '{stub}' not found in address stubs. "
+                    "Provide --address-stubs with a mapping file."
+                )
+            account = self._stub_accounts[stub]
+        else:
+            if storage is None:
+                storage = {}
+            code = self.code_pre_processor(code)
+            code_bytes = (
+                bytes(code) if not isinstance(code, (bytes, str)) else code
+            )
+            max_code_size = self._fork.transitions_from().max_code_size()
+            assert len(code_bytes) <= max_code_size, (
+                f"code too large: {len(code_bytes)} > {max_code_size}"
+            )
 
-        if storage is None:
-            storage = {}
-        code = self.code_pre_processor(code)
-        code_bytes = (
-            bytes(code) if not isinstance(code, (bytes, str)) else code
-        )
-        max_code_size = self._fork.transitions_from().max_code_size()
-        assert len(code_bytes) <= max_code_size, (
-            f"code too large: {len(code_bytes)} > {max_code_size}"
-        )
-
-        account = Account(
-            nonce=nonce,
-            balance=balance,
-            code=code,
-            storage=storage,
-        )
+            account = Account(
+                nonce=nonce,
+                balance=balance,
+                code=code,
+                storage=storage,
+            )
 
         if address is not None:
             assert address not in self, (
@@ -451,11 +467,20 @@ def eoa_by_index(i: int) -> EOA:
     return EOA(key=TestPrivateKey + i if i != 1 else TestPrivateKey2, nonce=0)
 
 
+@pytest.fixture(scope="session")
+def stub_accounts(
+    request: pytest.FixtureRequest,
+) -> Dict[str, Account]:
+    """Return stub accounts pre-populated during configuration."""
+    return getattr(request.config, "stub_accounts", {})  # type: ignore
+
+
 @pytest.fixture(scope="function")
 def pre(
     alloc_flags: AllocFlags,
     fork: Fork | None,
     request: pytest.FixtureRequest,
+    stub_accounts: Dict[str, Account],
 ) -> Alloc:
     """Return default pre allocation for all tests (Empty alloc)."""
     # FIXME: Static tests don't have a fork so we need to get it from the node.
@@ -467,4 +492,5 @@ def pre(
     return Alloc(
         flags=alloc_flags,
         fork=actual_fork,
+        stub_accounts=stub_accounts,
     )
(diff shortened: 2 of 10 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/address_stubs.py
```diff
@@ -0,0 +1,54 @@
+"""Address stubs for mapping stub labels to on-chain contract addresses."""
```
danceratopz: Just to better indicate that this is a helper module, not a plugin in itself.

```suggestion
"""
Address stubs model shared by the filler and execute plugins.

This model maps stub labels to on-chain contract addresses.
"""
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -38,19 +40,40 @@
...
 def pytest_addoption(parser: pytest.Parser) -> None:
     """Add command-line options to pytest."""
     pre_alloc_group = parser.getgroup(
         "pre_alloc",
         "Arguments defining pre-allocation behavior during test filling.",
     )
-
-    # No options for now
-    del pre_alloc_group
+    pre_alloc_group.addoption(
```
danceratopz, line 53: Just a comment. This dupliates these cli flags from `execute` and `--rpc-endpoint` also exists in `execute/eth_config/eth_config.py`. I don't have a clear strategy to remove this redundancy. Let's skip it for now, unless you have a clear idea.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/address_stubs.py
```diff
@@ -0,0 +1,54 @@
...
+    @classmethod
+    def model_validate_json_or_file(cls, json_data_or_path: str) -> Self:
+        """
+        Try to load from file if the value resembles a path that ends with
+        .json/.yml and the file exists.
+        """
+        lower_json_data_or_path = json_data_or_path.lower()
+        if (
+            lower_json_data_or_path.endswith(".json")
+            or lower_json_data_or_path.endswith(".yml")
```
danceratopz: Looks like we only have stubs in JSON format? I only get these hits, no YAML:
```
tests/benchmark/stateful/bloatnet/stubs_mainnet.json
tests/benchmark/stateful/bloatnet/stubs_bloatnet.json
tests/benchmark/stateful/bloatnet/stubs_99.json
```

In which case, I'd suggest cleaning out this code in a follow-up PR, wdyt?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/address_stubs.py
```diff
@@ -0,0 +1,54 @@
...
+class AddressStubs(EthereumTestRootModel[Dict[str, Address]]):
+    """
+    Address stubs class.
+
+    The key represents the label that is used in the test to tag the contract,
+    and the value is the address where the contract is already located at in
+    the current network.
+    """
+
+    root: Dict[str, Address]
```
danceratopz, line 21: Another comment on pre-existing code.

This gives the impression that we correctly validate addresses, but as the `Address` model is intentionally lenient for test code and left-pads addresses (e.g. `0x100` is ok), it doesn't even catch an address of an incorrect length (so it may as well be `str`).

If there's a typo, all tests using this address fail with:
```
Stub 'test_sload_empty_erc20_balanceof_50GB_ERC20' at 0x0398324972fce0e89e048c2104f1298031d1931f has no code on chain.
```

We can easily improve the test validation of a  here, but perhaps stubs should be one-off verified before we execute/fill tests (not here, as a follow-up)?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -230,10 +267,38 @@ def _deploy_contract(
...
+            stub_address = self._address_stubs[stub]
+            if stub_address in self._stub_code_cache:
+                fetched_code = self._stub_code_cache[stub_address]
+                logger.info(
+                    f"Using cached code for stub '{stub}' at "
+                    f"{stub_address} ({len(fetched_code)} bytes)"
+                )
+            else:
+                fetched_code = self._eth_rpc.get_code(stub_address)
+                if fetched_code == b"":
```
danceratopz: More idiomatic:
```suggestion
                if not fetched_code:
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -230,10 +267,38 @@ def _deploy_contract(
...
+                )
+            stub_address = self._address_stubs[stub]
+            if stub_address in self._stub_code_cache:
+                fetched_code = self._stub_code_cache[stub_address]
+                logger.info(
+                    f"Using cached code for stub '{stub}' at "
+                    f"{stub_address} ({len(fetched_code)} bytes)"
+                )
+            else:
+                fetched_code = self._eth_rpc.get_code(stub_address)
```
danceratopz: Should we also retrieve the contract's nonce, balance and strorage here and fill with that? Or do we lean on hard-coded values in the address? Follow-up issue?

danceratopz (commented):
Great idea, @LouisTsai-Csie! I think this PR pretty much good to go, just one nit to resolve here I think. Otherwise, as I haven't checked this code in a while most of my comments are about potential issues in pre-existing code, perhaps it's worthwhile to address these in follow-ups?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/address_stubs.py
danceratopz: I.e., a hard fail for unavailable addresses before starting the test session could avoid wasting time.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -38,19 +40,40 @@
...
+        "--address-stubs",
+        action="store",
+        dest="address_stubs",
+        default=None,
+        type=AddressStubs.model_validate_json_or_file,
+        help=(
+            "Address stubs mapping stub names to on-chain addresses. "
+            "Can be a JSON string or path to a JSON/YAML file."
+        ),
+    )
```
spencer-tb, line 60: Should we fail fast when we add address stubs without the rpc endpoint? Not required but could be a nice addition.
```python
 def pytest_configure(config: pytest.Config) -> None:
      """Validate stub-related options are consistent."""
      address_stubs = config.getoption("address_stubs", None)
      rpc_endpoint = config.getoption("rpc_endpoint", None)
      if address_stubs is not None and rpc_endpoint is None:
          pytest.exit(
              "--address-stubs requires --rpc-endpoint to fetch "
              "contract code from the network."
          )
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/address_stubs.py
```diff
@@ -0,0 +1,54 @@
...
+            path = Path(json_data_or_path)
+            if path.is_file():
+                path_suffix = path.suffix.lower()
+                if path_suffix == ".json":
+                    return cls.model_validate_json(path.read_text())
+                elif path_suffix in [".yml", ".yaml"]:
+                    loaded_yaml = yaml.safe_load(path.read_text())
+                    if loaded_yaml is None:
+                        return cls(root={})
+                    return cls.model_validate(loaded_yaml)
```
spencer-tb: Should we fail for file not found here (following the `if path.is_file()`):
```python
else:
    raise FileNotFoundError(
        f"Address stubs file not found: {path}"
    )
```

spencer-tb (commented):
Nice review from Dan. I mentioned a couple more things to make it more robust but nothing essential.

fselmo:
Is there a benefit to this approach over mapping stubbed account bytecode as a static map in EELS or elsewhere to be loaded? I feel like if this bytecode will not change (it shouldn't change for stubbed contracts) then we should remove the network call part of this and just run the `get_code()` once and map this internally as static bytecode values. Am I missing something here?

fselmo:
> Is there a benefit to this approach over mapping stubbed account bytecode as a static map in EELS or elsewhere to be loaded? I feel like if this bytecode will not change (it shouldn't change for stubbed contracts) then we should remove the network call part of this and just run the `get_code()` once and map this internally as static bytecode values. Am I missing something here?

I think either way we would _probably_ want to make a call to verify that it matches the expectation but this would be a validation then in that sense. We would then have an _expectation_ of the bytecode to sanity check against the RPC call though - instead of just accepting whatever is there blindly (which is also probably fine but not ideal).

In other words, right now we just assume whatever bytecode `eth_getCode` returns is correct. But if we had an internal map of the expected bytecode that we know is correct, we could then use an optional `eth_getCode` call in fill to verify that it is the same and if it's not we could raise.

This is how I understood we would implement it, so don't mind me if the approach in this PR was discussed elsewhere as the way to go. Just trying to sanity check here :)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
LouisTsai-Csie: I found some duplication for both "--rpc-endpoint" (3 duplicates) and "--address-stubs" (2 duplicates), let me try to put them into somewhere the module could share.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/address_stubs.py
LouisTsai-Csie: We only use JSON file, i've checked with CPerezz, it would be good to restrict the validation here.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
LouisTsai-Csie: I move the duplicate rpc endpoint and address stub variables from separate filler and execute module to `execute_filler.py`. But i do not have a good idea for the `--rpc-endpoint` duplication in `eth_config`

LouisTsai-Csie:
@fselmo Your approach is more standard and better validates the stub account and stub address. I believe Mario also suggested moving forward with this path in issue #2423.

That said, I have a few concerns:
- The stubs might be large, for example, only [stubs_99.json](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/tests/benchmark/stateful/bloatnet/stubs_99.json) contains nearly 200 ERC20 contracts, most of them have different bytecode. This means we need to fetch and store the contract code for each of them.
- Given the large amount of contract bytecode, it's not really feasible for us to verify correctness directly. We might end up just verifying that the stub address returns a non-empty code field. If this is the case, this could be handled by checking the return value of `eth_getCode` as part of the stub fetching integration as well.
- We currently use Spamoor to deploy these contracts. I think the private key for deployer account for Spamoor are the same across networks, deploying contracts with the same nonce on different networks will yield the same contract address, regardless of bytecode differences. This is because contract address calculation depends only on the sender address and nonce, but not the chain ID. This means, we need to maintain [network][contract address] for each stub account, this raises the same question in 1 and 2.

What do you think, happy to discuss more on this topic to make the validation more robust.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
LouisTsai-Csie: I think nonce and balance should be enough here. Although i would like to include storage, fetching the entire storage would be difficult. afaik, we can only rely on the `getStorageAt` RPC method, which means: (1) we would need to pre-calculate all storage slots that may be touched, and (2) we would need to call this RPC for every single slot. That becomes hard when the storage range is large (potentially more than 4K slots for a single contract in some benchmark).

Because of this, I do not think the stub account solution can easily resolve the storage issue. We may need another approach, such as interacting with perfnet or the snapshot more directly.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
LouisTsai-Csie: No luck, this requires larger refactor (--rpc-endpoiont), as it is also being used by `recover` and some other module, i would prefer to do it in a separate PR for this.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/address_stubs.py
LouisTsai-Csie: I first query the contract accounts and cache the results, so no need to query again during pre-allocation.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py
```diff
@@ -2,22 +2,32 @@
...
 from execution_testing.specs.base import OpMode
 from execution_testing.test_types import EOA, Alloc, ChainConfig
 
+from ..shared.address_stubs import AddressStubs
 from ..shared.pre_alloc import AllocFlags
 from ..spec_version_checker.spec_version_checker import EIPSpecTestItem
 
+logger = get_logger(__name__)
+
+stub_accounts_key: StashKey[Dict[str, Account]] = StashKey()
```
LouisTsai-Csie, line 31: I'm using a `StashKey` object here to explicitly define the type of `stub_accounts_key`. Without it, a `# type: ignore` comment below would be needed to silence the `mypy` error:

```python
config.stub_accounts.get(stub_accounts_key, {}) # type: ignore
```

By defining `StashKey[T]` with a concrete type, `mypy` can verify all reads and writes to config.stash without any ignores.

Does this approach make sense, or is there a better alternative?

spencer-tb (commented):
.

spencer-tb (commented):
LGTM! Small comment! Approved!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -48,9 +49,13 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         "pre_alloc",
         "Arguments defining pre-allocation behavior during test filling.",
     )
-
-    # No options for now
-    del pre_alloc_group
+    pre_alloc_group.addoption(
+        "--rpc-endpoint",
+        action="store",
```
spencer-tb, line 55: The `--rpc-endpoint` duplication is fine imo (as discussed), but you could leave a brief `# TODO: consolidate with execute/rpc/remote.py` so it's discoverable in a future cleanup and/or a new issue?

spencer-tb (approved):

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py
spencer-tb: Feels like the right approach from my side. I didn't know about it until now :P

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
danceratopz: Let me handle this one! On it now :)

danceratopz: Handled in https://github.com/LouisTsai-Csie/execution-specs/pull/7

danceratopz (changes requested):
Hey @LouisTsai-Csie, please check:
- https://github.com/LouisTsai-Csie/execution-specs/pull/7

My main remaining nit was not updating the `help` plugin to show the new CLI flags. This is stupid I know, it should be auto-generated, but hey.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/remote.py
danceratopz: One last question: This logic is no longer respected when we use `--address-stubs`. I think we should fix this up in this PR.

danceratopz (commented):
Hey @LouisTsai-Csie, nearly theeeeere. Just one last nit.

I had the context, so I addressed it here:
https://github.com/LouisTsai-Csie/execution-specs/pull/8

danceratopz (approved):
Nice! Thanks @LouisTsai-Csie! We did it 😆

PR #2514, opened 2026-03-16
fix(tests): use filler expect entries for ported static test post assertions

### LINKED ISSUE #2507: Open issues with ported static tests
### Post is not fully ported

Example test [tests/static/state_tests/stBugs/staticcall_createfailsFiller.json](https://github.com/ethereum/execution-specs/blob/static-tests/tests/static/state_tests/stBugs/staticcall_createfailsFiller.json) contains the following post:
https://github.com/ethereum/execution-specs/blob/130951253d53f26fd6c1c3218aa69eeaed694d48/tests/static/state_tests/stBugs/staticcall_createfailsFiller.json#L18-L27
But in the ported test the post is empty:
https://github.com/ethereum/execution-specs/blob/365433e1ba3d5cc2716db0b08a3cdd6a2f24210a/tests/ported_static/stBugs/test_staticcall_createfails.py#L108

leolara:
## 🗒️ Description

Switch `fixture_to_python.py` post-state generation from compiled fixture state to filler `expect` entries.

The compiled fixture path only had access to final account state, so it silently dropped `shouldnotexist`, zero-valued storage, nonce, balance, and code assertions. This caused 22% of ported tests to generate `post: dict = {}`. The filler `expect` entries already contain the original test author's intended assertions — and the functions to parse them (`load_filler_expect_results`, `resolve_expect_for_case`, `generate_post_dict`, `generate_post_value_string`) already existed as dead code.

This PR wires them up, with the compiled fixture path kept as a fallback for fillers that lack expect entries.

### What changed

- **`extract_case_indices`**: Support dash-separated fixture keys (`-d0-g0-v0]`) used by compiled fixtures, in addition to the existing `d0g0v0-` format.
- **`generate_test_file`**: Accept `expect_entries`, resolve each case's filler result via `extract_case_indices` + `resolve_expect_for_case`, prefer filler results over compiled fixture state with fallback.
- **`generate_post_dict` / `generate_post_value_string`**: Handle `code` field — string hex, empty (`""`, `"0x"` → `b""`), YAML-parsed integers, and retesteth label syntax.
- **`_resolve_code_labels`**: New helper to replace `<contract:0xADDR>` / `<eoa:0xADDR>` tags embedded in bytecode with raw hex addresses, reusing the existing `_LABEL_RE` regex.
- **`process_single_fixture`**: Load filler expect entries and pass to `generate_test_file`.

### Validation

- 2149/2149 generated, 0 failures
- 446/446 paired fixtures match
- Empty `post: dict = {}`: 483 → 18

## 🔗 Related Issues or PRs

Fixes #2507

## ✅ Checklist

    ```console
    uvx tox -e static
    ```


spencer-tb:
Thanks Leo! Some comments:

It looks like we are missing the fork specific post checks:
- `resolve_expect_for_case` doesn't consider the network (fork) field from filler expect entries, `load_filler_expect_results` drops it. When a filler has different expected results per fork (e.g different sender balance for Cancun vs Prague), the first matching entry is always used regardless of the target fork.
- Only 4 fillers are affected by the latter (OverflowGasRequire2, creationTxInitCodeSizeLimit, transactionCosts,
  precompsEIP2929Cancun), their `_from_prague` test variants end up with Cancun post-state assertions. For example, OverflowGasRequire2 expects `balance=0x...E6357F` on Cancun and `0x...E5F97F` on Prague, but both generated functions use the Cancun value.
- We should store the network/fork in the expect entries and filter by fork when resolving.
- Can we make sure to verify for Amsterdam as well?

Minor nit:
- The code field handling is duplicated between `generate_post_dict` and `generate_post_value_string` could be a small helper.

Otherwise the PR looks great!!

leolara:
So, I have been trying to vibe code a solution for post with differences between forks. I think I am almost there. But I have the impression that the more we keep adding stuff treating the script as a blackbox, the more difficult it becames and the more confused the agent gets.

I think perhaps I could refactor the script until it is now into a nice architecture, and make sure that it produces the same result as until now. And then from there continue.

For example, there is a new thing that the multi-fork post triggers, something call :label substitution that was being done in the static tests. Like in tests/static/state_tests/stPreCompiledContracts/precompsEIP2929CancunFiller.yml 

```
  expect:

  - indexes:
      data: 
      - :label yes
    network: 
    - ">=Cancun"
    result:
```

Also, there is the issue about what Felipe said about the tags in the original static tests, that perhaps considering them would help with the `contract_deploy` thing.

I am afraid that it will get more confused the vibe coded script, as we add more. Although it might work as well.

Also, the parametrization that the script is generating seems a bit crazy.

leolara:
I am getting this type of message from the agent:

```
This is a fundamental issue: the compiled fixture uses label-based test IDs (not d0g0v0), and extract_case_indices can't extract d=1 from -invalid]. The d,g,v values we get are both (0,0,0).

  The fix needs to resolve the label to a d-index from the fixture key. But this brings us back to the label resolution problem. For this specific case, the fixture key has valid and invalid as labels, and the filler's data array maps valid → d=0, invalid → d=1.

  This is getting deep. The creationTxInitCodeSizeLimit issue requires extending extract_case_indices to resolve labels from fixture keys using the filler's label map. That's beyond just the fork filtering fix.
  ```
  
 The smartest at the highest effort. I am more convinced that this needs a refatoring to add more stuff. I think we should actually had to mimic most of what static-fill was doing, but produce a test instead of a fixture.

leolara:
I sent some code that I think it fixes the multi-fork. But perhaps there are many tests failing now, related to the new post.

leolara:
There are still errors when runing ported_static. I am fixing them

leolara:
@spencer-tb @marioevz now all tests pass with the multi-fork post

marioevz:
The core issue at https://github.com/ethereum/execution-specs/issues/2507 still stands.

The problem in that particular example is that `0x1d0384eb7c2b1a9d9862c8e180f9e4d1696a2a8e` is still not appearing in the converted `tests/ported_static/stBugs/test_staticcall_createfails.py` test, and it might be because the address is not the sender, nor the coinbase, nor any of the contracts. Rather it is the computed address of the created contract from contract `0xc94f5374fce5edbc8e2a8697c15331677e6ebf0b` at nonce 63.

It's an important address because is the main check of the test: the contract was not created when the `CREATE` opcode was used in an static context. So it's decisive that we keep that check to the python ported test.

I don't think we should brute-force check if the unidentified addresses in the expect section are the result of a contract creation, but we still should copy these addresses over to the python as hard-coded strings to preserve the check.

leolara:
@marioevz it gets worse, it seems the contact bytecode is converted anymore to opcodes like before.

It seems it is following pareto law, the vibecoding did very well the easy 80%, but when you ask it to fix some of the rest 20%, it breaks something else somewhere.

I think it is better to ask the agent to code something from a plan based on static-fill code.

spencer-tb:
Just added the idea for post state verifcation checks from the call to this PR: https://github.com/ethereum/execution-specs/pull/2552

Was fairly easy to add! I think this will help with verification. We might need to consider where its worth keeping in the future, as I think its redundant once all the python ports are complete.

leolara:
> Just added the idea for post state verifcation checks from the call to this PR: #2552
> 
> Was fairly easy to add! I think this will help with verification. We might need to consider where its worth keeping in the future, as I think its redundant once all the python ports are complete.

this is great!!

leolara:
Closing this, I will send another one

PR #2520, opened 2026-03-17
feat(test-fixture): add `metadata` subfield to `_info` with `target_opcode` 

### LINKED ISSUE #2208: Add `target_opcode` field in fixture for benchmark
Although the benchmark tests define a `target_opcode` field, it is not included in the fixture. As a result, users currently have to parse the target opcode from the test name.
We should instead embed the target opcode directly into the fixture to make this information explicit and readable

LouisTsai-Csie:
## 🗒️ Description

Add `target_opcode` field in the fixture, so that user could parse the benchmark target more easily from the fixture, instead of the test name.

## 🔗 Related Issues or PRs

Closes #2208 

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -1811,14 +1811,21 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
                             fixture.post_state, group.pre
                         )
 
+                fill_metadata: Dict[str, Any] = {}
+                if t8n.opcode_count is not None:
+                    fill_metadata["opcode_count"] = (
+                        t8n.opcode_count.model_dump()
+                    )
+                if fill_result.metadata:
+                    fill_metadata.update(fill_result.metadata)
+
                 fixture.fill_info(
                     t8n.version(),
                     test_case_description,
                     fixture_source_url=fixture_source_url,
-                    opcode_count=t8n.opcode_count,
                     ref_spec=reference_spec,
                     _info_metadata=t8n._info_metadata,
-                    fill_metadata=fill_result.metadata or None,
+                    metadata=fill_metadata,
                 )
 
                 output_subdir = resolve_fixture_subfolder(
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -1818,6 +1818,7 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
                     opcode_count=t8n.opcode_count,
                     ref_spec=reference_spec,
                     _info_metadata=t8n._info_metadata,
+                    fill_metadata=fill_result.metadata or None,
                 )
 
                 output_subdir = resolve_fixture_subfolder(
--- a/packages/testing/src/execution_testing/specs/benchmark.py
+++ b/packages/testing/src/execution_testing/specs/benchmark.py
@@ -554,6 +554,11 @@ class BenchmarkTest(BaseTest):
                 self._verify_target_opcode_count(
                     fill_result.benchmark_opcode_count
                 )
+
+            # Embed target opcode in the fill result for fixture metadata
+            if self.target_opcode is not None:
+                fill_result.metadata["target_opcode"] = str(self.target_opcode)
+
             return fill_result
         else:
             raise Exception(f"Unsupported fixture format: {fixture_format}")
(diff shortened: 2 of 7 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -1793,6 +1793,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
                     opcode_count=t8n.opcode_count,
                     ref_spec=reference_spec,
                     _info_metadata=t8n._info_metadata,
+                    target_opcode=fill_result.target_opcode,
```
raxhvl: suggestion: collect into a metadata dict

fn params has grown quite a bit, adding a metadata dict:

- allows for arbitrary metadata 
- avoids benchmark specific metadata from polluting param name space for generic tests (where it will be none)
- maybe club with _info_metadata?
```
 def fill_info(self, ..., metadata: dict[str, Any])
```

jsign (approved):
LGTM!

No strong opinions on @raxhvl point about where to place the field!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
danceratopz: Great suggestion, thanks.

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -554,6 +554,11 @@ def generate(
                 self._verify_target_opcode_count(
                     fill_result.benchmark_opcode_count
                 )
+
+            # Embed target opcode in the fill result for fixture metadata
```
danceratopz: Unnecessary comment.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -1818,6 +1818,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
                     opcode_count=t8n.opcode_count,
                     ref_spec=reference_spec,
                     _info_metadata=t8n._info_metadata,
+                    fill_metadata=fill_result.metadata or None,
```
danceratopz: I think `or None` is redundant here.

danceratopz (changes requested):
Thanks for adding this @LouisTsai-Csie!

@raxhvl made a [great suggestion above](https://github.com/ethereum/execution-specs/pull/2520#discussion_r2944676199) and the internals were implemented for the new field. However, the existing field was left as-is. I think it'd be worth the extra effort to clean that up now and move `BaseFixture.info["opcode_count"]` to 
`BaseFixture.fill_medata["opcode_count"]`. I'm referring to this:
https://github.com/LouisTsai-Csie/execution-specs/blob/72b40e928911c222cea8e74bb8f73746ecc4c73c/packages/testing/src/execution_testing/fixtures/base.py#L184-L185

Also, I think we should start grouping this data in the fixture JSON itself into a `"metadata"` subfield of `"_info"`. This will be a breaking change for @jsign and other users, but worth it longer term, imo. 
```json
        "_info": { "hash":
            "0xbd4e46ed66434c9a0c01b4368b7f7f73b692ca6169bc567b3e7f500283bf601a",
            "fixture-format": "blockchain_test",
            "comment":
            "`execution-specs` generated test",
            "filling-transition-tool": "evmone-t8 0.18.0-65+commit.e70cfc5c", 
            "description": "Benchmark BLS12_381 precompile.",
            "url": "https://github.com/ethereum/execution-specs/blob/0cc455b1cf225059cdfb2567ff78f2fa09af5cd9/tests/benchmark/compute/precompile/test_bls12_381.py#L18",
            "metadata" {
                "opcode_count": {
                    "MOD": 2,
                    "GT": 4,
                    "EQ": 10,
                    "ISZERO": 14
                },
                "target_opcode": "GT"
            },    
```

I do wonder if, additionally, the test spec type should be included in metadata? Or we provide some other way to know which fields to expect in `metadata`. But this can be a follow-up.

I think it's important to bring `opcode_count` into the metadata in the scope of this PR. And am not a fan of flag-merging more fields to `_info`.

spencer-tb (approved):
Approved, nice addition! Won't block from my side but maybe we want to consider only adding this to benchmark tests, we'd add it here:
 > packages/testing/src/execution_testing/specs/benchmark.py, in the generate method where it already sets target_opcode

Can be a follow up, do we have any reason to include it for non-benchmark tests?

danceratopz (approved):
Thanks so much for going the extra mile @LouisTsai-Csie. It was worth it imo, this looks great!

Regarding @spencer-tb's comment:
> Won't block from my side but maybe we want to consider only adding this to benchmark tests

I 💯 agree and I hit a wall last week trying to fill consensus tests with older geth's as it now fails with an unsupported flag `--opcode.count` error. I created a follow-up issue for this, not high prio!
- https://github.com/ethereum/execution-specs/issues/2585

PR #2526, opened 2026-03-18
feat(specs,tests): EIP-8037 state gas ordering and clarifications

spencer-tb:
## 🗒️ Description

Note: these changes are already added to `devnets/bal/3`, and this PR is to highlight key changes to the EIP.

Fix EIP-8037 state gas charging order and add detection/clarification tests. Includes `forks/amsterdam` rebase changes.

**Spec fix**: When an opcode charges both regular gas and state gas, regular gas must be charged first, EIPs PR: https://github.com/ethereum/EIPs/pull/11421. If regular gas OOGs, state gas is not consumed. The previous ordering (state before regular) allowed state gas spilled into `gas_left` to be lifted into the parent's reservoir via `incorporate_child_on_error`, inflating it and changing the transaction's effective gas consumption.

This was identified by @rakita (reth/revm) who found a failing test fixture on the `tests-bal@v5.3.0` tag. The revm implementation already charges regular gas before state gas across all opcodes.

**Opcodes fixed** (state gas moved after regular gas):
- `SSTORE` -- storage set (zero to non-zero)
- `CALL` -- value transfer to non-existent account
- `SELFDESTRUCT` -- dead beneficiary with non-zero balance

`CREATE`, `CREATE2`, and code deposit already had the correct order.

**CALL memory expansion double-charge fix**: `CALL` with value transfer to a new account was double-charging memory expansion costs — once in regular gas and again in state gas. Fixed to only charge memory costs in regular gas.

**1D to 2D gas accounting fix** (`test_sstore_oog_no_reservoir_inflation` in `test_state_gas_create.py`): The original `test_max_initcode_size_gas_metering_via_create` from `tests-bal@v5.3.0` used 1D gas accounting (all gas through CALL, zero reservoir). This test uses the same pattern with `reservoir=0`, verifying the gas accounting is correct in the fixture for both `exact_gas` and `short_one_gas` cases.

**Detection tests** (`test_state_gas_ordering.py`): Each test gives a child frame 1 gas less than needed, then probes the parent's reservoir with SSTOREs. Clients with wrong ordering produce different storage values (not just different gas accounting), causing fixture assertion failures.

**Edge case tests** for [ethereum/EIPs#11414](https://github.com/ethereum/EIPs/pull/11414) (EIP-8037 text clarifications by @nerolation ):
- SSTORE stipend check uses `gas_left` only, excluding reservoir
- CREATE does not double-charge `GAS_NEW_ACCOUNT`

**Edge case test** from @jwasinger (via @MariusVanDerWijden):
- Calldata floor exceeding `TX_MAX_GAS_LIMIT` rejects the transaction at validation

**Fixture format divergence fix:** The `blocks` fixture in `tests/prague/eip7002_el_triggerable_withdrawals/conftest.py` mutated `WithdrawalRequestContract.tx_gas_limit` in-place (`+=`) to add EIP-8037 state gas costs. Since pytest shares parameter objects across fixture format runs, the gas cost compounded, added once for `blockchain_test`, then again for `blockchain_test_engine` on the already-modified object, producing different transactions and block hashes between formats. Fixed by deep copying the parameters before mutation. Reported by @jangko (nimbus).

## 🔗 Related Issues or PRs

- [ethereum/EIPs#11421](https://github.com/ethereum/EIPs/pull/11421) -- EIP-8037 state gas charging ordering clarifications
- [ethereum/EIPs#11414](https://github.com/ethereum/EIPs/pull/11414) -- EIP-8037 text clarifications (stipend, CREATE formula)

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -436,15 +436,6 @@ def call(evm: Evm) -> None:
     if is_cold_access:
         evm.accessed_addresses.add(to)
 
-    # Charge state gas for new account creation (replaces GAS_NEW_ACCOUNT)
-    if value != 0 and not is_account_alive(tx_state, to):
-        cost_per_state_byte = state_gas_per_byte(
-            evm.message.block_env.block_gas_limit
-        )
-        charge_state_gas(
-            evm, STATE_BYTES_PER_NEW_ACCOUNT * cost_per_state_byte
-        )
-
     extra_gas = access_gas_cost + transfer_gas_cost
     (
         is_delegated,
@@ -462,14 +453,28 @@ def call(evm: Evm) -> None:
     code_hash = get_account(tx_state, code_address).code_hash
     code = get_code(tx_state, code_hash)
 
+    # Charge regular overhead first, then state gas, before the
+    # child gas calculation. This ensures:
+    # 1. Regular-gas OOG does not consume state gas (no inflation).
+    # 2. State gas spill into gas_left reduces the child's gas
+    #    (matching reth's ordering).
+    charge_gas(evm, extra_gas + extend_memory.cost)
+    if value != 0 and not is_account_alive(tx_state, to):
+        cost_per_state_byte = state_gas_per_byte(
+            evm.message.block_env.block_gas_limit
+        )
+        charge_state_gas(
+            evm, STATE_BYTES_PER_NEW_ACCOUNT * cost_per_state_byte
+        )
+
     message_call_gas = calculate_message_call_gas(
         value,
         gas,
         Uint(evm.gas_left),
         extend_memory.cost,
-        extra_gas,
+        extra_gas=Uint(0),
     )
-    charge_gas(evm, message_call_gas.cost + extend_memory.cost)
+    charge_gas(evm, message_call_gas.cost)
     escrow_subcall_regular_gas(evm, message_call_gas.sub_call)
 
     evm.memory += b"\x00" * extend_memory.expand_by
@@ -651,18 +656,22 @@ def selfdestruct(evm: Evm) -> None:
     if is_cold_access:
         evm.accessed_addresses.add(beneficiary)
 
-    if (
+    needs_state_gas = (
         not is_account_alive(tx_state, beneficiary)
         and get_account(tx_state, evm.message.current_target).balance != 0
-    ):
+    )
+
+    # Charge regular gas before state gas so that a regular-gas OOG
+    # does not consume state gas that would inflate the parent's
+    # reservoir on frame failure.
+    charge_gas(evm, gas_cost)
+    if needs_state_gas:
         cost_per_state_byte = state_gas_per_byte(
             evm.message.block_env.block_gas_limit
         )
         charge_state_gas(
             evm, STATE_BYTES_PER_NEW_ACCOUNT * cost_per_state_byte
         )
-
-    charge_gas(evm, gas_cost)
 
     originator = evm.message.current_target
     originator_balance = get_account(tx_state, originator).balance
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -453,11 +453,6 @@ def call(evm: Evm) -> None:
     code_hash = get_account(tx_state, code_address).code_hash
     code = get_code(tx_state, code_hash)
 
-    # Charge regular overhead first, then state gas, before the
-    # child gas calculation. This ensures:
-    # 1. Regular-gas OOG does not consume state gas (no inflation).
-    # 2. State gas spill into gas_left reduces the child's gas
-    #    (matching reth's ordering).
     charge_gas(evm, extra_gas + extend_memory.cost)
     if value != 0 and not is_account_alive(tx_state, to):
         cost_per_state_byte = state_gas_per_byte(
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -453,6 +453,9 @@ def call(evm: Evm) -> None:
     code_hash = get_account(tx_state, code_address).code_hash
     code = get_code(tx_state, code_hash)
 
+    # TODO: Consider consolidating charge_gas + charge_state_gas into
+    # a single gas charge to avoid duplicate EVM trace entries.
+    # Applies here and in create, create2, selfdestruct. See #2526.
     charge_gas(evm, extra_gas + extend_memory.cost)
     if value != 0 and not is_account_alive(tx_state, to):
         cost_per_state_byte = state_gas_per_byte(
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -466,7 +466,7 @@ def call(evm: Evm) -> None:
         value,
         gas,
         Uint(evm.gas_left),
-        extend_memory.cost,
+        memory_cost=Uint(0),
         extra_gas=Uint(0),
     )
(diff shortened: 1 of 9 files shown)


spencer-tb:
Tagging @fselmo re-STEEL call. This will be added to `devnets/bal/3` branch nonetheless for now (unless anything major caught).

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -462,14 +453,23 @@ def call(evm: Evm) -> None:
...
+        )
+        charge_state_gas(
+            evm, STATE_BYTES_PER_NEW_ACCOUNT * cost_per_state_byte
+        )
+
     message_call_gas = calculate_message_call_gas(
         value,
         gas,
         Uint(evm.gas_left),
         extend_memory.cost,
```
fselmo: We charge the mem cost now on line 456 so we should zero this out, as we did with `extra_gas` here, right?

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_calldata_floor.py
```diff
@@ -128,3 +129,61 @@ def test_calldata_floor_higher_than_execution_with_state_ops(
...
+    exceeds_cap: bool,
+) -> None:
+    """
+    Verify calldata floor > TX_MAX_GAS_LIMIT rejects the transaction.
+
+    When the EIP-7623 calldata floor cost exceeds the EIP-7825 transaction
+    gas limit cap, the transaction must be rejected at validation —
+    even though the regular intrinsic gas may be within the cap.
+
+    at_cap: calldata floor exactly equals the cap — transaction accepted.
```
fselmo: Hmm this is sort of a nit but we need to be more clear here regardless so also not really a nit. This isn't exactly at the cap because it's not divisible enough to _be_ exact. I asked Claude to see what the tightest gap here can be and I got this:

```
The tightest gaps achievable:                                                                                                                                                                                  
                                                                       
  ┌──────────────────────────────┬───────────┬────────────┬─────┐                                                                                                                                                
  │           Approach           │  Tokens   │   Floor    │ Gap │      
  ├──────────────────────────────┼───────────┼────────────┼─────┤                                                                                                                                                
  │ All nonzero (current test)   │ 1,675,620 │ 16,777,200 │ -16 │      
  ├──────────────────────────────┼───────────┼────────────┼─────┤                                                                                                                                                
  │ Mixed (nonzero + zero bytes) │ 1,675,621 │ 16,777,210 │ -6  │                                                                                                                                                
  ├──────────────────────────────┼───────────┼────────────┼─────┤                                                                                                                                                
  │ Next token up                │ 1,675,622 │ 16,777,220 │ +4  │                                                                                                                                                
  └──────────────────────────────┴───────────┴────────────┴─────┘                                                                                                                                                
                                                                       
  So the current all-nonzero approach gives a gap of 16. Using mixed bytes could tighten it to 6. But exact boundary is unreachable — this is the tightest we can get with only nonzero bytes.                   
                                                                       
  The docstring should acknowledge this: the at_cap case is "largest calldata floor that fits within the cap" and exceeds_cap is "smallest calldata floor that exceeds the cap," not "exactly equals" and        
  "exceeds by 1."
```

It would be nice to tighten this gap to be as close as possible to the cap and still pass.

fselmo: Actually this ended up being trivial to tighten up  so I'm putting it in a commit here. Feel free to drop any of these commits too or build on top. It's just a bit easier to do for these smaller changes.

src/ethereum/forks/amsterdam/vm/instructions/system.py
fselmo, line 428: One thing I noticed is that these messages are inconsistent atm across the opcodes that introduced this concept with BALs. I think it would be good not only to align on the design pattern for these similar opcodes, but also this is maybe a chance to make things more clear now that we have 2D gas.

A Claude sweep analysis on this thought:

- Current — inconsistent wording (selfdestruct), no mention of "regular" gas:
   - call() L428: `# check static gas before state access`
   - callcode() L548: `# check static gas before state access`
   - selfdestruct() L646: `# check access gas cost before state access`
   - delegatecall() L725: `# check static gas before state access`
   - staticcall() L822: `# check static gas before state access`

Maybe all these comments can align on "regular" gas wording? I'm not sure what we're calling this... "regular" seems weird so maybe we can say something like:

`# ensure enough non-state gas before accessing state`

Still not sure if that's super clear... but this is just a nit to make specs a bit more readable with 2D gas.

---

Another catch that Claude caught that was unrelated to my ask but fair enough since it aligns with consistent messaging across opcodes:

```
Current — exists in sstore (L148-150) and selfdestruct (L659-661):
 # Charge regular gas before state gas so that a regular-gas OOG
 # does not consume state gas that would inflate the parent's
 # reservoir on frame failure.

 Missing from — call() (L456), create() (L189), create2() (L241).
```

This is for the `charge_gas()` call in each opcode.

Thoughts on this? 👀. Simple one-liner comments but helps to align the design on the specs side since they are bits of logic that are trying to achieve the same goal.

fselmo (commented):
Found one bug and one possible improvement, narrowing the calldata floor gap. I put them as commits here but feel free to drop them if I missed something, add on to them, etc. It just felt easier to sanity check the bug with a test, pairing with Claude, see that it failed and then the change itself was a one-liner... so I feel like we can use this test 👀.

Also a nit on aligning comments / logic design across similar opcodes for consistency.

Lmk your thoughts :)

src/ethereum/forks/amsterdam/vm/instructions/system.py
fselmo: Follow-up comment got lost here I think. I put this as two separate commits here. The first is a test I worked with Claude on to prove this was a bug and debug it - this did indeed fail to fill. The next commit was the fix to zero this out and it fills after this change. Lmk your thoughts.

spencer-tb: Thx so much for this!

src/ethereum/forks/amsterdam/vm/instructions/system.py
spencer-tb: Very fair! We should add these. I think the whole 8037 EIP needs a full review before we merge into `forks/amsterdam` in the future, there will be more things like this I'm sure! :D

fselmo: Yeah makes sense to get the logic working first. Can skip this for this PR, take or leave :)

fselmo (approved):
Going to approve here @spencer-tb. Since I added my proposed fixes, let me know if you agree or not, etc. If you want another review with any more changes / dropped commits, feel free to re-request and I'll try to prioritize this and quickly get another review in but lgtm!

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -462,14 +453,23 @@ def call(evm: Evm) -> None:
     code_hash = get_account(tx_state, code_address).code_hash
     code = get_code(tx_state, code_hash)
 
+    charge_gas(evm, extra_gas + extend_memory.cost)
```
fselmo, line 459: I think there is an issue with the design to `charge_gas` multiple times. There's a bit of context [here](https://discord.com/channels/595666850260713488/1435353174532952074/1435356835598045330) from implementing EIP-7928. Some of this pertains to EVM trace specification where we should only charge it once. Perhaps there is more nuance there too.

I believe we can account for this gas, and pass it to `calculate_message_call_gas`, without calling `charge_gas` more than once. We can use `check_gas` and we can do proper accounting within the logic and pass this accurate value to the `calculate_message_call_gas()`.

Curious on your thoughts here.

cc: @gurukamath

spencer-tb: Added a todo.

PR #2527, opened 2026-03-19
feat(test-benchmark): verify transaction status

LouisTsai-Csie:
## 🗒️ Description

We use `expected_benchmark_gas_used` to verify that benchmarks behave as expected. For more complex scenarios, such as complex gas accounting cases or stub account interactions, we may apply `skip_gas_used_validation` to bypass validation. However, this makes it harder to confirm correct benchmark behavior. That said, for certain scenarios, particularly storage-related benchmarks, we still need to know whether a transaction runs out of gas (OOG). If a transaction hits OOG, we cannot accurately measure storage write performance.

In this PR, we add a new verification approach, such that we ensure each transaction is not OOG.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/specs/base.py
+++ b/packages/testing/src/execution_testing/specs/base.py
@@ -36,6 +36,9 @@ from execution_testing.fixtures import (
 from execution_testing.forks import Fork
 from execution_testing.forks.base_fork import BaseFork
 from execution_testing.test_types import Environment, Withdrawal
+from execution_testing.test_types.receipt_types import (
+    TransactionReceipt,
+)
 
 
 class HashMismatchExceptionError(Exception):
@@ -109,6 +112,7 @@ class BaseTest(BaseModel):
     gas_optimization_max_gas_limit: int | None = None
     expected_benchmark_gas_used: int | None = None
     skip_gas_used_validation: bool = False
+    expected_receipt_status: int | None = None
     is_tx_gas_heavy_test: bool = False
     is_exception_test: bool = False
 
@@ -287,6 +291,33 @@ class BaseTest(BaseModel):
             "benchmark gas allowed for this configuration: "
             f"{gas_benchmark_value}"
         )
+
+    def validate_receipt_status(
+        self,
+        *,
+        receipts: List[TransactionReceipt],
+        block_number: int,
+    ) -> None:
+        """
+        Validate receipt status for every transaction in a block.
+
+        When expected_receipt_status is set, verify that all
+        receipts match. Catches silent OOG failures that roll
+        back state and invalidate benchmarks.
+        """
+        if self.expected_receipt_status is None:
+            return
+        for i, receipt in enumerate(receipts):
+            if receipt.status is not None and (
+                int(receipt.status) != self.expected_receipt_status
+            ):
+                raise Exception(
+                    f"Transaction {i} in block "
+                    f"{block_number} has receipt "
+                    f"status {int(receipt.status)}, "
+                    f"expected "
+                    f"{self.expected_receipt_status}."
+                )
 
 
 TestSpec = Callable[[Fork], Generator[BaseTest, None, None]]
--- a/packages/testing/src/execution_testing/specs/base.py
+++ b/packages/testing/src/execution_testing/specs/base.py
@@ -305,7 +305,7 @@ class BaseTest(BaseModel):
         receipts match. Catches silent OOG failures that roll
         back state and invalidate benchmarks.
         """
-        if self.expected_receipt_status is None:
+        if "expected_receipt_status" not in self.model_fields_set:
             return
         for i, receipt in enumerate(receipts):
             if receipt.status is not None and (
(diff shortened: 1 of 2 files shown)


packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -872,6 +870,11 @@ def make_fixture(
             if is_last_block and self.operation_mode == OpMode.BENCHMARKING:
                 benchmark_gas_used = int(built_block.result.gas_used)
                 benchmark_opcode_count = built_block.result.opcode_count
+            if built_block.result.receipts:
+                self.validate_receipt_status(
+                    receipts=built_block.result.receipts,
+                    block_number=i,
```
fselmo: Hmm we enumerate here so this isn't correct, is it? `i` would start at `0` whereas I imagine blocks would start at `1`? Can you confirm? I think we can / should eliminate this `enumerate(...)` and skip using the `i` altogether -- instead we should set a var immediately in this code block and use this. All references to `i` in this block are for a block num anyhow so this works well and won't confuse us.

packages/testing/src/execution_testing/specs/blockchain.py
fselmo: See last comment but we can do something like

```diff
-        for i, block in enumerate(self.blocks):
-           is_last_block = i == len(self.blocks) - 1
+        last_block = self.blocks[-1]
+        for block in self.blocks:
             built_block = self.generate_block_data(
                 t8n=t8n,
                 block=block,
                 previous_env=env,
                 previous_alloc=alloc,
             )
+            block_number = int(built_block.header.number)
-            if is_last_block and self.operation_mode == OpMode.BENCHMARKING:
+            if block is last_block and self.operation_mode == OpMode.BENCHMARKING:
                 benchmark_gas_used = int(built_block.result.gas_used)
                 benchmark_opcode_count = built_block.result.opcode_count
             if built_block.result.receipts:
                 self.validate_receipt_status(
                     receipts=built_block.result.receipts,
-                    block_number=i,
+                    block_number=block_number,
                 )                                                                                                                                            
```

This skips the potential off-by-one `i` issue altogether and makes it a bit less "guesswork" to a reader that we are actually using the correct number. I think we indeed were off by one though since our last block check was using "- 1" in `is_last_block = i == len(self.blocks) - 1`.

packages/testing/src/execution_testing/specs/base.py
```diff
@@ -288,5 +292,32 @@ def validate_benchmark_gas(
...
+        block_number: int,
+    ) -> None:
+        """
+        Validate receipt status for every transaction in a block.
+
+        When expected_receipt_status is set, verify that all
+        receipts match. Catches silent OOG failures that roll
+        back state and invalidate benchmarks.
+        """
+        if self.expected_receipt_status is None:
```
fselmo: nit: We generally use this pydantic property to check whether fields were set or not instead of the `None` check.

```suggestion
        if "expected_receipt_status" not in self.model_fields_set:
```

fselmo (commented):
One apparent silent bug in implementation with indexing blocks via `enumerate` and one small nit... but I think this looks good otherwise. Lmk your thoughts.

packages/testing/src/execution_testing/specs/blockchain.py
LouisTsai-Csie: Ah i see, we should not start with 0 as it is genesis block, right? I do not notice this, thanks!

packages/testing/src/execution_testing/specs/blockchain.py
LouisTsai-Csie: > block_number = int(built_block.header.number)

Like this approach, especially this is clear^

packages/testing/src/execution_testing/specs/base.py
LouisTsai-Csie: > if "expected_receipt_status" not in self.model_fields_set:

The rationale behind this check is to see whether `expected_receipt_status` was explicitly set by the user. 

However, in `BaseTest.from_test()`, all `BaseTest` fields are copied into the constructor kwargs, including those that were never explicitly set. This caused `expected_receipt_status=None` to appear in `model_fields_set` on the new `BlockchainTest`, bypassing the check and causing a failure with `1 != None`.

I add a fix in `base.py` — `from_test()` now only copies fields that are present in `base_test.model_fields_set`, preventing unset defaults from leaking as "explicitly set" on the target test.

```python
for k in BaseTest.model_fields.keys():
  if k not in kwargs and k in base_test.model_fields_set:
      kwargs[k] = getattr(base_test, k)
  return cls(**kwargs)
```

fselmo: > edit: this seems to break some other logic now.

Oh I see... pay no mind then. This is a nit... as long as we make sure it works as expected, all good from my end 👍🏼

fselmo: oh the actual changes look different from this comment and look like they don't break anything. This wasn't marked as resolved so still unsure if this is breaking? Is this resolved (it looks to be)? 👀

fselmo (approved):
This lgtm!

PR #2528, opened 2026-03-19
chore(ci): [Experiment] split py3 job into 6 pytest-split groups

danceratopz:
## 🗒️ Description

Experimental: split the `py3` CI job into 6 parallel matrix groups using `pytest-split` with `least_duration` balancing. Codecov merges the per-group coverage reports automatically.

It enables `./tests/static_ported/`.

Note: this approach won't work for enginex fixture fills since `pytest-split` doesn't support per-runner grouping (unlike xdist's per-worker grouping).

I don't think the pytest-split approach is it tbh, it should provide very good grouping per worker, see https://jerry-git.github.io/pytest-split/#splitting-algorithms. But I think a simple split by fork would be best as this:
1. Plays well with our new fixture dir layout `for_cancun` (we can now trivially merge the fixture output for releases).
2. Per-fork split works for EngineXFixtures as tests from different groups will never get split across runners; each fork gets at least one new pre-alloc group anyway (you can't group across forks).

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

cute animal

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -56,8 +56,13 @@ jobs:
           uvx --from actionlint-py actionlint
 
   py3:
+    name: py3 (${{ matrix.group }}/6)
     runs-on: [self-hosted-ghr, size-xl-x64]
     needs: static
+    strategy:
+      fail-fast: false
+      matrix:
+        group: [1, 2, 3, 4, 5, 6]
     steps:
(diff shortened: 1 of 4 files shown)


danceratopz:
Looks like pytest-split doesn't play well with loadgroup (which is a prerequisite for caching and EngineXFixtures and bigmem), so a manual fork split it is.

PR #2530, opened 2026-03-19
fix(test-benchmarks): gas accounting for `test_account_access` when nonzero value

jochem-brouwer:
## 🗒️ Description
Fixes the gas accounting for account_query, which before this fix made the account_query go OOG if value > 0, defeating the purpose of the stateful test.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_account_query.py
+++ b/tests/benchmark/stateful/bloatnet/test_account_query.py
@@ -164,6 +164,7 @@ def test_account_query(
                 # Gas accounting
                 address_warm=access_warm,
                 new_memory_size=max(mem_size, 96),
+                value_transfer=value_sent > 0,
             )
         )
     elif opcode in (Op.STATICCALL, Op.DELEGATECALL):


LouisTsai-Csie (approved):
LGTM

tests/benchmark/stateful/bloatnet/test_account_query.py
```diff
@@ -164,6 +164,7 @@ def test_account_query(
                 # Gas accounting
                 address_warm=access_warm,
                 new_memory_size=max(mem_size, 96),
+                value_sent=value_sent > 0,
```
danceratopz: If `value_sent` can't be negative, I find this more idiomatic. But if it can be negative it'll be wrong, so perhaps I prefer yours?
```suggestion
                value_sent=bool(value_sent),
```

danceratopz (commented):
Thanks @jochem-brouwer, one comment.

tests/benchmark/stateful/bloatnet/test_account_query.py
jochem-brouwer: Hmm, this is a style thing, I personally would prefer the value_sent > 0 thing but this might come from my background in JS. If I see these bool type conversions I would first have to look up what type value_sent is and how python converts is (a negative number would be something I think I would convert to False, but it actually converts to True). 
The number cannot be negative.

Small note, we merged this a week ago (also my code :sweat_smile: ):
https://github.com/ethereum/execution-specs/blob/df37d252b1e5af2e99144f6b92810d5332c7f68a/tests/benchmark/stateful/bloatnet/test_single_opcode.py#L1664-L1665

I am fine with the bool() approach but I think for consistency this rule should then be enforced by the linter :smile:

jochem-brouwer: And note, the value_sent is 0 or 1 here but it could also be 0/1/5 for instance, or 0/1000 (so the 0 or 1 is not meant to be a 0 or 1 as to be interpreted as boolean)

LouisTsai-Csie: I personally think we could move forward with current approach, and revisit necessary refactor later, when the state benchmark becomes more stable.

danceratopz: No strong feelings, I've also seen something like this recently, too:
```python
value_transfer=True if value_sent else False
```
in which case the boolean is nicer. 

I don't think we can enforce this stylistic choice by the linter (it is also not equivalent if , but I am concerned that you initially had `value_sent=` and not `value_transfer=`, i.e., this required fix was not caught by linter:
https://github.com/ethereum/execution-specs/compare/392f784182d0dcf9c72ac15b279297aa38284d19..df37d252b1e5af2e99144f6b92810d5332c7f68a

danceratopz: I looked into this. It would only be validated at runtime and these tests aren't ran in CI. This kind of mistake will be caught, once https://github.com/ethereum/execution-specs/pull/2512 gets merged.

PR #2532, opened 2026-03-19
fix(tests): prevent `tx_gas_limit` double-accumulation across fixture format runs in withdrawal request contract tests

felix314159:
jangko from nimbus notified us that in our bal 5.5.1 fixtures there is a block hash mismatch between `blockchain_test ` and `blockchain_test_engine` for `test_withdrawal_requests`.

## Cause of `block_hash` mismatches

Since pytest runs the blocks fixture separately for each fixture format (once for `blockchain_test`, once for `blockchain_test_engine`), and the `WithdrawalRequestContract` objects are shared across those runs (they're pytest parameter objects), the gas limit was accumulated twice — adding `121404` each time. This made the transaction RLP differ between formats, which changed the `transactions_trie`, which changed the `block_hash`.

Why only contract tests were affected: Only `WithdrawalRequestContract` had this gas adjustment. EOA-based tests (`WithdrawalRequestTransaction`) didn't get the adjustment, so their hashes were consistent.

## Fix
Save the original gas limit before the first adjustment and always compute from that base value.

## How to reproduce 

Before applying this PR's fix run:
`uv run fill --clean -k "test_withdrawal_requests" tests/prague/eip7002_el_triggerable_withdrawals/test_withdrawal_requests.py --fork=Amsterdam -s`

and then verify issue via:
	
```
import json

bt = json.load(
    open(
        "fixtures/blockchain_tests/for_amsterdam/prague/eip7002_el_triggerable_withdrawals/withdrawal_requests/withdrawal_requests.json"
    )
)

be = json.load(
    open(
        "fixtures/blockchain_tests_engine/for_amsterdam/prague/eip7002_el_triggerable_withdrawals/withdrawal_requests/withdrawal_requests.json"
    )
)


def n(k):
    return k.replace("blockchain_test_engine", "X").replace(
        "blockchain_test", "X"
    )


bm = {n(k): k for k in bt}
em = {n(k): k for k in be}

d = [
    (
        nk,
        bt[bm[nk]]["blocks"][-1]["blockHeader"]["hash"],
        be[em[nk]]["lastblockhash"],
    )
    for nk in sorted(bm.keys() & em.keys())
    if bt[bm[nk]]["blocks"][-1]["blockHeader"]["hash"]
    != be[em[nk]]["lastblockhash"]
]

print(
    f"{len(d)}/{len(bm.keys() & em.keys())} tests have mismatched block hashes"
)

for nk, bh, eh in d:
    print(f"  {nk.split('X-')[-1][:70]}")

```

then apply the fix from this PR and re-run the script above. You will see that pre-fix shows `14/26` tests have mismatched block hashes, after-fix `0/26` mismatches.

#### Cute Animal Picture

Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -1148,6 +1148,44 @@ def pytest_html_results_table_row(report: Any, cells: Any) -> None:
 
 
 @pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_setup(item: Any) -> Generator[None, None, None]:
+    """
+    Snapshot parametrize values before fixture setup to detect unintended
+    mutations of shared pytest parameter objects across fixture format runs.
+    """
+    if hasattr(item, "callspec"):
+        item._param_repr_snapshot = {
+            key: repr(value) for key, value in item.callspec.params.items()
+        }
+    yield
+
+
+def pytest_runtest_teardown(item: Any) -> None:
+    """
+    Compare parametrize values after test teardown to the pre-setup snapshot.
+
+    Warn if any fixture mutated shared parameter objects — these mutations
+    persist across fixture format runs and can cause subtle bugs (e.g.
+    block hash mismatches between blockchain_test and blockchain_engine_test).
+    """
+    snapshot = getattr(item, "_param_repr_snapshot", None)
+    if snapshot is None:
+        return
+    for key, original_repr in snapshot.items():
+        current_repr = repr(item.callspec.params[key])
+        if current_repr != original_repr:
+            warnings.warn(
+                f"Shared pytest parameter '{key}' was mutated during "
+                f"test '{item.nodeid}'. Mutations on parametrize values "
+                f"persist across fixture format runs and can cause "
+                f"divergent test results. Avoid mutating these objects "
+                f"in fixtures; compute derived values locally instead.",
+                stacklevel=1,
+            )
+    del item._param_repr_snapshot
+
+
+@pytest.hookimpl(hookwrapper=True)
 def pytest_runtest_makereport(
     item: Any, call: Any
 ) -> Generator[None, None, None]:
--- a/packages/testing/src/execution_testing/forks/forks/forks.py
+++ b/packages/testing/src/execution_testing/forks/forks/forks.py
@@ -3691,17 +3691,21 @@ class Amsterdam(BPO2):
             current_value = original_value
         new_value = metadata["new_value"]
 
-        gas_cost = 0 if metadata["key_warm"] else gas_costs.GAS_COLD_STORAGE_ACCESS
+        gas_cost = (
+            0 if metadata["key_warm"] else gas_costs.GAS_COLD_STORAGE_ACCESS
+        )
 
         if original_value == current_value and current_value != new_value:
             if original_value == 0:
                 # EIP-8037: regular portion + state gas
                 gas_cost += (
-                    gas_costs.GAS_COLD_STORAGE_WRITE - gas_costs.GAS_COLD_STORAGE_ACCESS
+                    gas_costs.GAS_COLD_STORAGE_WRITE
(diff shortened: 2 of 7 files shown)


felix314159:
We should also add a script similar to above to CI so that this can not again happen undetected

fselmo:
Hey @felix314159, thanks for sniffing this out... this is a really important finding. I agree with you, we need to have some guardrails here so these things don't happen again.

I think we can simplify the approach a bit but I'm happy to discuss it. Instead of stashing the base gas limit to make the mutation idempotent, we could pass `fork` into `transactions()` and compute the state gas adjustment there as a pure function. That way the shared `pytest.param` object(s) are never mutated at all... so the problem can't really happen regardless of how many format runs share the same object.

I pushed two commits here with this approach that I'm happy to drop after discussion if this isn't agreed on. The second commit snapshots `repr` (since copy / pickle isn't possible on certain objects) of params before a test and compares it after to issue a warning for tests that currently mutate params in place. It flags some tests but these currently don't accumulate (`+=`), they mutate via assignment (`self.var = val`) so in this sense they are OK. But imo, we should create an issue to fix any tests that trigger this warning to return a new value rather than mutate and then we can flip this warning to a `raise` or something like this.

What do you think?

Again, happy to drop these and get your idempotent fix in but I think this might help steer us to the right framework changes.

cc @danceratopz

fselmo:
> But imo, we should create an issue to fix any tests that trigger this warning to return a new value rather than mutate and then we can flip this warning to a `raise` or something like this.

I asked Claude to summarize the goal for the follow-up issue:

```
Context

PR #2532 fixed a bug where WithdrawalRequestContract.tx_gas_limit was double-accumulated across fixture 
format runs because pytest parameter objects are shared by reference. The fix moved the gas computation 
into transactions() as a pure function. However, the underlying anti-pattern — fixtures mutating shared 
pytest.param objects — remains in all three EL request test modules via update_pre().

A new repr-based snapshot hook in filler.py (pytest_runtest_setup/pytest_runtest_teardown) now warns 
when parametrize values are mutated during a test run. Currently, every EL request test triggers this 
warning due to update_pre() setting sender_account, contract_address, and entry_address on shared 
helper instances.

These mutations happen to be idempotent today (assignments, not accumulations), so they don't cause 
incorrect fixtures. But the pattern is fragile — any future += or conditional mutation would silently produce 
divergent block hashes across blockchain_test vs blockchain_engine_test, exactly like the bug #2532 fixed.

- Goal

  - Refactor the EL request test helpers so that update_pre() returns values rather than mutating self, and 
    downstream methods (transactions(), valid_requests()) accept those values as parameters.
  - Once all warnings are eliminated, flip the filler hook from warning to hard failure.
  ```

felix314159:
thanks for your commits, making it fork-aware and using a pure function is cleaner. 

however, i now worry about what happens if no fork is passed, because that still seems to be optional: e.g if we look at [this line](https://github.com/ethereum/execution-specs/blob/7bc95e113ecabda5363ed69a235418a20bc6eb44/tests/prague/eip7002_el_triggerable_withdrawals/test_withdrawal_requests.py#L868) we can see `WithdrawalRequestInteractionBase.transactions()` being called without a fork explicitly passed to it (still optional today..). so it seems that correctness now depends on callers remembering to pass `fork`..

i fully support us removing mutations for el helpers in `update_pre()` and just have it return a value. also +1 on the start with warnings and then pivot to hard failure.

however even with the proposed warning/failure hook, i still would like to also have an explicit hash comparison script (compares certain hashes in `blockchain_test` vs `blockchain_test_engine` fixtures) to run. it should be simple to write and it would detect even cases where no `fork` is passed, that seems like a footgun that now exists (separate problem). the proposed warning hook wouldn't catch a missed fork because nothing mutates.

fselmo:
Ah bummer I meant to make one last comment that we should push this to `forks/amsterdam` :). Looks like it was sitting as "pending" on a new review or something. Anyhow, I think we should get this in the framework not the EIP branch. But I guess it will get there eventually and the only issue was within EIP-8037... so it's ok I think 👍🏼

PR #2535, opened 2026-03-20
feat(test-fill): Add --verify-traces flag for trace comparison against baselines

### LINKED ISSUE #2506: feat(test-fill): `--verify-traces` flag for automatic trace comparison across test runs
## Summary

Add a `--verify-traces <folder>` parameter that points to a folder containing traces from a previous test run, automatically compares them against the current run's traces, and warns when execution behavior has changed in unexpected ways.

The goal is to give developers a fast feedback loop when modifying tests: confirm that the underlying EVM execution hasn't shifted in ways that indicate a broken or semantically different test, rather than just checking that it still passes.

## Motivation

When refactoring or modifying tests, a passing result alone doesn't guarantee the test is still exercising the same code paths. Subtle changes — like different storage slots being touched, altered call depths, or shifted gas profiles can slip through unnoticed. Comparing execution traces between the original and modified test runs surfaces these regressions.

Particularly ported tests in #2455 will need to be improved to better match the behavior of the rest of the python tests (remove hard-coded addresses, hard-coded environment, verify that test still behaves exactly as before).

### Step 1: Capture baseline traces

Run the test suite in its original (known-good) state and dump traces to a stable location:

```bash
uv run fill --clean --fork Osaka ./tests/ported_static/  -m static_test --evm-dump-dir ./baseline-traces --traces
```

This produces a folder of per-test trace files that serve as the reference baseline.

### Step 2: Modify tests, then verify

After making changes to the test or tests, re-run with the new flag:

```bash
uv run fill --clean --fork Osaka ./tests/ported_static/path/to/modified/test.py  -m static_test --verify-traces ./baseline-traces
```

Note that we don't need to pass `--evm-dump-dir` or `--traces` flag since these are handled by the `--verify-traces`.

This will:

1. Execute the modified tests and generate fresh traces (internally / in a temp directory).
2. Match each new trace to its corresponding baseline trace by test name/identifier.
3. Diff the two traces and classify each difference as **expected** or **suspicious**.
4. Print a summary report with warnings for any unexpected changes.

## Trace Diffing — Acceptable vs. Suspicious Changes

The core complexity is in deciding what constitutes a meaningful execution change. A starting heuristic:

### Acceptable differences

- Remaining execution gas difference: ideally we want to be able to adjust the amount of gas that is passed with the transaction of the test without this triggering a warning.
- Addresses in the stack should not trigger a difference because the contract addresses and transaction senders are what we primarily want to update (indirectly, created contract addresses are also susceptible to changes).

### Error differences

- Different opcodes executed.
- Opcode raises execution error (out of gas, invalid opcode, etc.) when it previously did not, or vice versa.
- Call that execution aborts or reverts in the baseline now succeeding, or vice versa.
- Opcode gas cost changes.

## Output Format

The test filling should fail if the trace comparison trace fails, to signal that there was an unexpected change in the traces, while also printing a comprehensive explanation of the difference found.

leolara:
## 🗒️ Description

Add a `--verify-traces <folder>` flag to the `fill` command that compares EVM execution traces from a baseline run against the current run. This enables developers to confirm that test modifications haven't changed underlying EVM execution behavior.

**Comparators:**
- `exact` — compares all fields including gas and gas_cost. Detects any difference, including opcode gas cost changes.
- `exact-no-gas` (default) — excludes `gas`, `gas_cost`, and `gas_used` fields, and applies GAS opcode stack pollution cleanup via `remove_gas()`. This tolerates remaining gas differences when transaction gas amounts change, but also tolerates opcode gas cost changes since those fields are excluded. Use the `exact` comparator when gas cost changes need to be detected.

Multiple comparators can be run simultaneously via `--verify-traces-comparator exact,exact-no-gas` to see how each test differs under different comparison strategies.

**Design decisions:**

- **Diagnostic, not failing:** Unlike the original issue which proposed failing tests on mismatch, this implementation reports all results across all comparators without failing. This lets developers see the full picture — a test might show DIFFERENT under `exact` but EQUIVALENT under `exact-no-gas`, which immediately tells them the only changes are gas-related. Failing on the first mismatch would hide this information.

- **Pluggable comparator architecture:** The ABC + factory pattern supports future comparators beyond exact matching:
  - **Isomorphic comparator** (follow-up): will handle address bijection so that address changes in the stack don't trigger differences — the primary use case for ported tests from #2455, as described in the issue.
  - **Out-of-gas tolerant comparator** (follow-up): for tests that intentionally run out of gas, where the execution may diverge at different points but the overall behavior is equivalent.

- **In-memory traces:** `--verify-traces` enables trace collection without writing files to disk (unless `--evm-dump-dir` is also passed). Baseline traces are loaded from disk; current traces are compared from memory.

- **Cache-aware:** The t8n output cache now restores traces on cache hits, so verification works correctly even for cached test formats.

## 🔗 Related Issues or PRs

Fixes #2506.

**Follow-up work:**
- Isomorphic comparator for address-aware comparison (addresses the issue's requirement that address changes should not trigger differences)
- Out-of-gas tolerant comparator
- JSON/HTML report formatters

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/client_clis/trace_comparators.py
+++ b/packages/testing/src/execution_testing/client_clis/trace_comparators.py
@@ -17,6 +17,7 @@ class TraceComparatorType(StrEnum):
     EXACT = "exact"
     EXACT_NO_GAS = "exact-no-gas"
     EXACT_NO_STACK = "exact-no-stack"
+    EXACT_NO_STACK_NO_GAS = "exact-no-stack-no-gas"
     GAS_EXHAUSTION = "gas-exhaustion"
 
 
@@ -285,8 +286,9 @@ _FIELD_EXCLUSION_CONFIGS: dict[
     tuple[set[str] | None, bool],
 ] = {
     TraceComparatorType.EXACT: (None, False),
-    TraceComparatorType.EXACT_NO_GAS: ({"gas", "gas_cost"}, True),
+    TraceComparatorType.EXACT_NO_GAS: ({"gas"}, True),
     TraceComparatorType.EXACT_NO_STACK: ({"stack"}, False),
+    TraceComparatorType.EXACT_NO_STACK_NO_GAS: ({"gas", "stack"}, False),
 }
 
 
--- a/packages/testing/src/execution_testing/client_clis/trace_comparators.py
+++ b/packages/testing/src/execution_testing/client_clis/trace_comparators.py
@@ -1,0 +1,227 @@
+"""Trace comparators for verifying EVM execution traces against a baseline."""
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from enum import StrEnum
+
+from execution_testing.client_clis.cli_types import (
+    TraceLine,
+    Traces,
+    TransactionTraces,
+)
+
+
+class TraceComparatorType(StrEnum):
+    """Supported trace comparator strategies."""
+
+    EXACT = "exact"
+    EXACT_NO_GAS = "exact-no-gas"
+
+
+def _format_trace_line_diff(
+    trace_line: TraceLine,
+    differing_fields: dict[str, str],
+) -> str:
+    """Format a trace line as an assembly-like string with diffs.
+
+    Return the opcode name, with differing field values in brackets
+    if any non-opcode fields differ.
+    Example: "PUSH1 (pc=0x3, stack=['0x4'])"
+    """
+    if not differing_fields:
+        return trace_line.op_name
+    fields_str = ", ".join(
+        f"{k}={v}" for k, v in differing_fields.items()
+    )
+    return f"{trace_line.op_name} ({fields_str})"
+
+
+@dataclass
+class TraceDifference:
+    """A difference between baseline and current trace at a specific line."""
+
+    transaction_index: int
+    trace_line_index: int
+    baseline: str
+    current: str
+
+
+@dataclass
+class TransactionCountMismatch(TraceDifference):
+    """Structural mismatch: different number of transactions."""
+
+    transaction_index: int = 0
+    trace_line_index: int = -1
+    baseline: str = ""
+    current: str = ""
+    baseline_count: int = 0
+    current_count: int = 0
+
+
+@dataclass
+class TraceComparisonResult:
+    """Result of comparing two Traces objects."""
+
+    equivalent: bool
+    differences: list[TraceDifference] = field(default_factory=list)
+
+
+class TraceComparator(ABC):
+    """Abstract base class for trace comparison strategies."""
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """Return the comparator's name."""
+        ...
+
+    @abstractmethod
+    def compare_transaction_traces(
+        self,
+        baseline: TransactionTraces,
+        current: TransactionTraces,
+        transaction_index: int,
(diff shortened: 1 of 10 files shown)


leolara:
I added two new comparators. So now we have:


 - exact — **general refactoring** Compares all fields of every trace line. Any difference at all (gas, stack, pc, etc.) is reported. Strictest
  comparator. 
  - exact-no-gas — **should match in EIP 8037** Compares all fields except gas and gas_cost. Also applies post-processing to nullify GAS opcode stack pollution  and ignores gas_used differences. Useful when gas schedules changed but execution logic is the same.
  - exact-no-stack — **should match if we refactor static tests to don't use hardcoded addresses** Compares all fields except stack. Gas and gas_cost must still match. Useful when contract or EOA addresses changed (which affects stack values) but the execution path and gas consumption should be identical.
  - gas-exhaustion — **should mismatch and help to fix OOG test in EIP 8037** Only checks whether both sides run out of gas at the same point(s). Equivalent if neither has out-of-gas errors, or both have them at the same trace line indices. Reports the opcode where each side's out-of-gas diverges.

leolara:
I have done some refactoring and improved the output report. It can print the report in human and json (useful for automatisation and AI agent).

packages/testing/src/execution_testing/client_clis/trace_comparators.py
```diff
@@ -0,0 +1,308 @@
...
+        )
+
+
+_FIELD_EXCLUSION_CONFIGS: dict[
+    TraceComparatorType,
+    tuple[set[str] | None, bool],
+] = {
+    TraceComparatorType.EXACT: (None, False),
+    TraceComparatorType.EXACT_NO_GAS: ({"gas", "gas_cost"}, True),
+    TraceComparatorType.EXACT_NO_STACK: ({"stack"}, False),
```
marioevz: I'd do something like this because `gas_cost` is almost always going to mean there's an issue with the new test:
```suggestion
    TraceComparatorType.EXACT_NO_GAS: ({"gas"}, True),
    TraceComparatorType.EXACT_NO_STACK: ({"stack"}, False),
    TraceComparatorType.EXACT_NO_STACK_NO_GAS: ({"gas", "stack"}, False),
```
`EXACT_NO_STACK_NO_GAS` would be the one that seems the most useful to me.

packages/testing/src/execution_testing/client_clis/trace_comparators.py
```diff
@@ -0,0 +1,308 @@
...
+    TransactionTraces,
+)
+
+
+class TraceComparatorType(StrEnum):
+    """Supported trace comparator strategies."""
+
+    EXACT = "exact"
+    EXACT_NO_GAS = "exact-no-gas"
+    EXACT_NO_STACK = "exact-no-stack"
```
marioevz, line 19: Following the previous comment.
```suggestion
    EXACT_NO_GAS = "exact-no-gas"
    EXACT_NO_STACK = "exact-no-stack"
    EXACT_NO_STACK_NO_GAS = "exact-no-stack-no-gas"
```

marioevz (approved):
Looks and works great, thanks for this! I

left two comments I think can be immediately applied and we can merge.

marioevz:
Broke an unit test with the suggestion, I'll try to fix it and push a fix, and add more tests for the new mode.

PR #2537, opened 2026-03-21
feat(utils): add proactive external temporal attestation for fork transitions

Heime-Jorgen:
## 🗒️ Description

This PR adds an optional `attestation_hook` to a new `validate_transition_timestamp()` function in `ethereum.utils.timestamp_attestation`. The hook is called **before** a fork transition is applied, allowing callers to validate that the block timestamp corresponds to real physical time — not just that it is numerically ≥ the configured threshold.

This is a backwards-compatible, opt-in extension. No existing behavior changes unless a caller supplies the hook.

## 🔗 Related Issues or PRs

Extends #2449 (spencer-tb — parametrize `TIMESTAMP_AT_YEAR_2106`).

### The timestamp manipulation surface

Fork transitions gated on `block.timestamp` create a narrow but real attack surface: a validator who controls block production can submit a timestamp that is technically valid (≥ `transition_timestamp`) yet meaningfully ahead of physical wall-clock time. On PoS Ethereum, this manifests as a validator advancing their timestamp by a few seconds to dozens of seconds within the protocol's tolerance window.

The consequences are fork-specific but include:

- Triggering EIP activation (or its fee/gas regime change) one or more slots early
- Influencing MEV extraction windows tied to fork boundaries
- Skewing governance vote windows that reference fork timestamps

### What existing tooling does — and does not do

Tools like Google's [Roughtime](https://roughtime.googlesource.com/roughtime) and similar timestamp-anomaly detectors are **retrospective**: they observe finalized block data, correlate it with external time signals, and flag anomalies after the fact. By the time the anomaly is reported, the block is finalized and its effects are committed to chain state.

Roughtime proves time fraud after the fact. This hook rejects the block before the transition is committed.

### What this PR proposes: proactive attestation

The `attestation_hook` operates at a different point in the lifecycle:

```
block received
    │
    ▼
transition_timestamp() called  ──→  numeric threshold (int)
    │
    ▼
validate_transition_timestamp(block.timestamp, threshold, hook?)
    │
    ├─── [hook present] ──→  attestation_hook(block.timestamp) ──→  True / False
    │                                                                     │
    │                                                           True: proceed to fork
    │                                                          False: reject block
    │
    └─── [hook absent] ──→  numeric check only (existing behavior, no change)
```

The timestamp is checked **before** the fork transition is committed. If the hook returns `False`, the block is invalid under the configured node's transition logic — the same as if the numeric threshold were not met.

### Interface

```python
from typing import Callable, Optional

AttestationHook = Callable[[int], bool]
# Receives: block.timestamp (Unix seconds)
# Returns:  True  → timestamp is externally verified, proceed
#           False → timestamp rejected, block invalid

def validate_transition_timestamp(
    block_timestamp: int,
    transition_ts: int,
    attestation_hook: Optional[AttestationHook] = None,
) -> bool:
    if block_timestamp < transition_ts:
        return False
    if attestation_hook is not None:
        return attestation_hook(block_timestamp)
    return True
```

A minimal reference NTP hook is documented in the module docstring — it is illustrative only and not imported into the spec.

## Game-theoretic property

With this hook active, a validator submitting a timestamp ahead of physical time by more than the operator's configured tolerance will have their block rejected by nodes running the hook. The economic incentive to manipulate timestamps is directly proportional to how far ahead they can push them; proactive attestation compresses that range toward zero.

Validators who consistently produce well-attested timestamps suffer no penalty. Validators who rely on drift for MEV or governance advantage face increasing rejection rates as adoption grows. This is economic self-eviction without a slashing condition.

## Backwards compatibility

- `attestation_hook` defaults to `None`
- All existing call sites continue to work without modification
- Clients that do not supply a hook see identical behavior to the current spec
- This PR does not change the numeric semantics of `transition_timestamp`

## Relationship to #2449

#2449 refactors the `15_000`-constant into a classmethod (`TIMESTAMP_AT_YEAR_2106`), making the numeric threshold configurable and testable. Once the threshold is parametric, it is natural to ask "parametric with respect to what else?" This PR answers: with respect to an optional external attestation step.

| | #2449 (spencer-tb) | This PR |
|---|---|---|
| Change | Hardcoded → classmethod | Adds optional attestation hook |
| Timestamp ground truth | `block.timestamp` (as-is) | External physical time (opt-in) |
| Enforcement point | Numeric gate (existing) | Pre-fork validation (new layer) |
| Backwards compat | Yes | Yes |

## Related work

**Reactive/retrospective approaches**

- Google Roughtime: cryptographically signed time protocol for detecting server clock skew. Operates as an audit layer over finalized data. No enforcement at block ingestion.
- Ethereum EIP-1559 base fee smoothing: uses timestamp deltas between consecutive blocks to bound fee volatility. This is a downstream signal, not a pre-ingestion gate.
- arXiv:2505.05328 (SUUM — Secure Unified UTXO Model): demonstrates empirically that PoW miners manipulated timestamps to ease difficulty adjustment — an analogous incentive structure exists for PoS validators seeking to drift timestamps within the protocol tolerance window. Note: SUUM's empirical findings are specific to the PoW context; the incentive analogy to PoS is structural, not quantitative.

**Proactive external time sources**

- IETF draft-helmprotocol-tttps: a proposed HTTPS-based time attestation protocol designed for smart contract environments. Referenced as an example of the proactive attestation model applied to a different layer of the stack.

## Open questions for reviewers

1. **Placement**: Should `validate_transition_timestamp` live in `ethereum/transitions.py`, or in a new `ethereum/time_validation.py` module to keep the concerns separated?
2. **Hook signature**: The current proposal passes only `block_timestamp: int`. Should it also receive `fork_name: str` and `transition_ts: int` to allow hooks to implement fork-specific tolerance windows?
3. **Fail-open vs. fail-closed default**: The reference implementation fails open on NTP query failure. Is this the right default for a spec-level interface, or should the spec leave this policy entirely to the hook implementor?
4. **Testing**: Should this PR add mock-hook tests to `tests/transitions/`, or is that out of scope for the spec layer?

## Files changed

```
src/ethereum/utils/timestamp_attestation.py  — AttestationHook type alias, validate_transition_timestamp()
tests/utils/__init__.py                       — new package init
tests/utils/test_timestamp_attestation.py     — 9 unit tests
```

No changes to fork-specific modules. No changes to existing `transition_timestamp()` classmethod signatures.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

A very patient tortoise, the original timestamp keeper

### DIFF
--- a/src/ethereum/utils/timestamp_attestation.py
+++ b/src/ethereum/utils/timestamp_attestation.py
@@ -1,0 +1,70 @@
+"""
+Proactive timestamp attestation for fork transition validation.
+
+Provides an optional hook mechanism that allows callers to validate that a
+block timestamp corresponds to real physical time — not just that it is
+numerically >= the configured transition threshold — before a fork transition
+is applied.
+
+This is the enforcement-layer complement to after-the-fact timestamp anomaly
+detection tools (e.g. Roughtime). Where those tools identify drift after
+finalization, the hook here operates before the transition is committed.
+
+See: https://eips.ethereum.org/EIPS/eip-1482 (15-second drift tolerance)
+"""
+
+from __future__ import annotations
+
+from typing import Callable, Optional
+
+from ethereum_types.numeric import U256
+
+AttestationHook = Callable[[int], bool]
+"""
+A callable that receives a block timestamp (Unix seconds, as int) and
+returns True if the timestamp is considered valid by an external source,
+or False if it should be rejected.
+
+The hook is called **before** the fork transition is applied. A return
+value of False causes the block to be treated as invalid under the
+transition logic of any caller that supplies the hook.
+
+Implementations should:
+- Be deterministic for the same input timestamp.
+- Not introduce blocking network I/O in the hot path (use cached values).
+- Fail open (return True) if the external source is unreachable, unless
+  the caller explicitly requires fail-closed behaviour.
+"""
+
+
+def validate_transition_timestamp(
+    block_timestamp: int,
+    transition_ts: int,
+    attestation_hook: Optional[AttestationHook] = None,
+) -> bool:
+    """
+    Return True if block_timestamp clears both the numeric threshold and,
+    when provided, the external attestation check.
+
+    Parameters
+    ----------
+    block_timestamp:
+        The Unix timestamp (seconds) from the block header.
+    transition_ts:
+        The minimum timestamp at which the fork activates.
+    attestation_hook:
+        Optional callable conforming to AttestationHook. When None, only
+        the numeric comparison is performed (existing behaviour unchanged).
+
+    Returns
+    -------
+    bool
+        False if block_timestamp < transition_ts.
+        False if attestation_hook is provided and returns False.
+        True otherwise.
+    """
+    if block_timestamp < transition_ts:
+        return False
+    if attestation_hook is not None:
+        return attestation_hook(block_timestamp)
+    return True
--- a/tests/utils/test_timestamp_attestation.py
+++ b/tests/utils/test_timestamp_attestation.py
@@ -1,0 +1,89 @@
+"""
+Tests for proactive timestamp attestation utilities.
+"""
+from __future__ import annotations
+
+import pytest
+
+from ethereum.utils.timestamp_attestation import (
+    AttestationHook,
+    validate_transition_timestamp,
+)
+
+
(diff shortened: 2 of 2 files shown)


spencer-tb:
Same as #2536, closing. Please stop submitting promotional PRs for OpenTTT to this repo.

PR #2538, opened 2026-03-21
fix(t8n): correct `Txs.transactions` type annotation

jihoonsong:
This PR corrects `Txs.transactions` type annotation.


danceratopz (approved):
Thanks, LGTM! Just pushed a fix to lint.

PR #2548, opened 2026-03-24
chores(consume): rename `--bin` flag of consume direct to `--evm-bin` to be consistent with fill's `--evm-bin`

felix314159:
## 🗒️ Description
Renames `consume direct`'s `--bin` flag to `--evm-bin` so the binary flag is consistent with `fill --evm-bin`, and it removes stale references to `ethereum-spec-evm-resolver` now that the weld is completed.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
+++ b/packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
@@ -21,7 +21,7 @@ from execution_testing.test_types import Alloc, Environment, Transaction
 
 CURRENT_FOLDER = Path(realpath(__file__)).parent
 FIXTURES_ROOT = CURRENT_FOLDER / "fixtures"
-DEFAULT_EVM_T8N_BINARY_NAME = "ethereum-spec-evm-resolver"
+DEFAULT_EVM_T8N_BINARY_NAME = "ethereum-spec-evm"
 
 
 @pytest.fixture(autouse=True)
@@ -35,8 +35,8 @@ def monkeypatch_path_for_entry_points(
     This would typically be in the venv in which pytest is running these tests
     and fill, which, with uv, is `./.venv/bin`.
 
-    This is required in order for fill to locate the ethereum-spec-evm-resolver
-    "binary" (entrypoint) when being executed using pytester.
+    This is required so pytester-based invocations can locate the
+    execution-specs transition tool entrypoint while running in tests.
     """
     bin_dir = sysconfig.get_path("scripts")
     monkeypatch.setenv("PATH", f"{bin_dir}:{os.environ['PATH']}")
--- a/packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
+++ b/packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
@@ -5,8 +5,7 @@ import os
 import sysconfig
 from os.path import realpath
 from pathlib import Path
-from shutil import which
-from typing import Dict, List, Type
+from typing import Dict, List
 
 import pytest
 from pydantic import TypeAdapter
--- a/packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
+++ b/packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
@@ -21,7 +21,6 @@ from execution_testing.test_types import Alloc, Environment, Transaction
 
 CURRENT_FOLDER = Path(realpath(__file__)).parent
 FIXTURES_ROOT = CURRENT_FOLDER / "fixtures"
-DEFAULT_EVM_T8N_BINARY_NAME = "ethereum-spec-evm"
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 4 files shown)


danceratopz (changes requested):
@felix314159 could you pull the cli flag change out into a dedicated PR (but see comment below), these changes are too orthogonal respectively contentious for a single PR imo.

About the flag change, I know the difference is mildly annoying, but this was a deliberate choice in `consume` as it can run tests that are don't strictly require an EVM binary. I must admit that the only example I can think of right now is `eoftest` which validated EOF bytecode without EVM execution. But I think I'd still rather keep it more general at `--bin`. If anything, `fill`'s `--evm-bin` flag is slightly too verbose (it was too tightly coupled to passing geth's `evm` binary). I think would prefer to leave consume as-is, but you could add an additional a `--bin` alias for `--evm-bin` to `fill` if helps things feel more consistent?

packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
```diff
@@ -21,7 +21,7 @@
 
 CURRENT_FOLDER = Path(realpath(__file__)).parent
 FIXTURES_ROOT = CURRENT_FOLDER / "fixtures"
-DEFAULT_EVM_T8N_BINARY_NAME = "ethereum-spec-evm-resolver"
```
danceratopz, line 24: Def worth cleaning up here, but I think we could completely remove this global and the test that uses it. The test is marked as skip, so essentially deadcode atm. The other option is to parametrize the test with other detected t8n tools? But I consider that overkill as we're not missing this test right now and it's implicitly system-tested via the benchchmark tests which use `evmone-t8n`/`geth`.

packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
```diff
@@ -35,8 +35,8 @@ def monkeypatch_path_for_entry_points(
     This would typically be in the venv in which pytest is running these tests
     and fill, which, with uv, is `./.venv/bin`.
 
-    This is required in order for fill to locate the ethereum-spec-evm-resolver
-    "binary" (entrypoint) when being executed using pytester.
+    This is required so pytester-based invocations can locate the
+    execution-specs transition tool entrypoint while running in tests.
```
danceratopz: turbo nit sorry
```suggestion
    `ethereum-spec-evm t8n` entrypoint while running in tests.
```

packages/testing/src/execution_testing/client_clis/tests/test_execution_specs.py
danceratopz: This is the skip I mean. I think we can entirely remove this test.

danceratopz (commented):
Here's a review on the `ethereum-spec-evm-resolver` to `ethereum-spec-evm` fix-ups! Good catch!

danceratopz:
Just to add that I didn't manage to finish reviewing the main part, but it looks to make makes sense. I'd like to check whether geth's behavior changed here (intentionally)? Will come back to this. Regarding this, assuming this is the main feature of this PR, wouldn't the title be: `fix(test-fill): fix filling pre-Byzantium forks with geth`?

felix314159:
the PR started with me wanting to change the `consume direct` flag and then evolved to a `ethereum-spec-evm-resolver` cleanup and then evolved to a `fix filling pre-Byzantium forks with geth` PR haha. i can split it

danceratopz:
> I'd like to check whether geth's behavior changed here (intentionally)?

Nothing changed on the geth side. I'm just surprised that we didn't spot this earlier. Perhaps we had a regression on the testing framework side. Thanks for finding and fixing it!

felix314159:
I moved the geth fix to a different PR, so this review should be straight-forward (just ignore my branch name)

spencer-tb (approved):
I like the change. LGTM! Approved the geth tx reciepts PR too.

felix314159:
@danceratopz can you 'resolve the threads' and make the 'requested change' thing disappear if you agree with how i implemented them?

danceratopz:
> @danceratopz can you 'resolve the threads' and make the 'requested change' thing disappear if you agree with how i implemented them?

As per https://github.com/ethereum/execution-specs/pull/2548#pullrequestreview-4019940031 above, I would still prefer to:
- add `--bin` as an alias to `fill` (in addition to `--evm-bin`) and additionally change all doc refs for `fill` from `--evm-bin` to `--bin`. 
- leave `consume` as-is.

@felix314159 wdyt?

spencer-tb:
> > @danceratopz can you 'resolve the threads' and make the 'requested change' thing disappear if you agree with how i implemented them?
> 
> As per [#2548 (review)](https://github.com/ethereum/execution-specs/pull/2548#pullrequestreview-4019940031) above, I would still prefer to:
> 
>     * add `--bin` as an alias to `fill` (in addition to `--evm-bin`) and additionally change all doc refs for `fill` from `--evm-bin` to `--bin`.
> 
>     * leave `consume` as-is.
> 
> 
> @felix314159 wdyt?

In hindsight, I think this is the way. `--evm-bin` I feel we got from geth terminology when EEST started, and is maybe too verbose. Having `--bin` feels more general, and better suited to client statetest/blocktest runner names

felix314159:
I would like there to be one flag, and I don't care whether it is `--bin` for both or `--evm-bin` for both, but I don't want to different flags that are the same (no alias)

spencer-tb:
Happy to use `--bin` if we add a deprecation message for people when they use `--evm-bin` with fill

danceratopz:
> Happy to use `--bin` if we add a deprecation message for people when they use `--evm-bin` with fill

Yup, I echo this. I don't see the need to make a sudden breaking change for everyone. So alias, with deprecation message if you consider important (for me leaving an alias indefinitely is also ok).

Alongside the doc changes, we should also fix-up CI.

PR #2551, opened 2026-03-24
feat(test-types): improve error messages for invalid tx fields

felix314159:
## 🗒️ Description
while writing tests i sometimes mess up when defining a tx and i think that `Transaction.InvalidFeePaymentError()` currently gives u a fairly unhelpful error message when u accidentally define a legacy tx field + 1559 field at the same time. this PR improves the logging (it tells u which fields that u defined are conflicting with regards to which tx type they belong to). also adds unit tests to ensure we don't break this accidentally.

## Command to showcase new logged error
`uv run pytest -v -s packages/testing/src/execution_testing/test_types/tests/test_types.py -k test_invalid_fee_payment_error_message`

and

`uv run pytest -v -s packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_error_messages.py`

now show you e.g. `cannot mix legacy/type-1 fee field 'gas_price' with type-2+ fee field 'max_fee_per_gas'; 'gas_price' is for legacy/type-1 txs, while 'max_priority_fee_per_gas' and 'max_fee_per_gas' are for type-2+ txs` which is much more helpful than the error u saw before that (`only one type of fee can be set`).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_error_messages.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_error_messages.py
@@ -1,7 +1,5 @@
 """Regression tests for fill error messages."""
 
-import subprocess
-import sys
 import textwrap
 from typing import Any
 
@@ -23,7 +21,9 @@ invalid_fee_payment_test_module = textwrap.dedent(
 )
 
 
-def test_fill_reports_conflicting_fee_fields(pytester: Any) -> None:
+def test_fill_reports_conflicting_fee_fields(
+    pytester: Any, capsys: Any, pytestconfig: Any
+) -> None:
     """Test that fill surfaces the conflicting fee fields in failures."""
     tests_dir = pytester.mkdir("tests")
     berlin_tests_dir = tests_dir / "berlin"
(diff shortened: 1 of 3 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_error_messages.py
```diff
@@ -0,0 +1,73 @@
...
+    fee_test_dir = berlin_tests_dir / "invalid_fee_payment_module"
+    fee_test_dir.mkdir()
+    test_module = fee_test_dir / "test_invalid_fee_payment_error.py"
+    test_module.write_text(invalid_fee_payment_test_module)
+
+    pytester.copy_example(
+        name="src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-fill.ini"
+    )
+
+    result = subprocess.run(
```
danceratopz: I think these tests should be ran using `pytester`, not just a plain `subprocess`. I.e., one of these (we have both, but I'm not sure what the difference is tbh:
```suggestion
    result = pytester.runpytest_subprocess(
    # or
    result = pytester.runpytest(
```
Or was there a particular reason for using `subprocess.run()`?

danceratopz (commented):
Nice! Thanks for tackling this; definitely a good dx gain!

I wondered if we could simplify the message formatting logic as it was quite verbose and included several conditionals. I made a suggestion here for your review:
- https://github.com/felix314159/execution-specs/pull/3

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_error_messages.py
felix314159: Yes there was a particular reason:

`pytester.runpytest` is fundamentally different because it is not isolated (so we usually want to avoid this one for avoiding bugs). `pytester.runpytest_subprocess` is very similar to running `subprocess.run` directly, but it gives u access to some pytest features (that we do not need here). it does have the big disadvantage that is will print the entire scary looking pytest failure, do u see all of that and then a green passed, which is a bit confusing. you can work around this (and i did just for you) by adding stuff like `capsys.readouterr()` but then you have the problem that it will show the print no matter whether u put `-s` or not. but that then can also be fixed with sth like:

```
    if pytestconfig.getoption("capture") == "no":
        with capsys.disabled():
            print(error_line)
```
so yes, you can get it working and idc in this example, but i do think it negatively affects readability in cases where we don't need specific pytest features

felix314159: i added a commit to do what you request, feel free to decide which version we use

danceratopz: That makes a lot of sense, thanks for the explanation. I think both approaches are totally valid, the raw `subprocess` just stood out to me in the review. I took another look at the different methods and think we should just pick one to be consistent. I went with `pytester` and the approach you applied in your last commit. I added a skill here to help clarity in the future: #2573

danceratopz (approved):
Thanks @felix314159! Let's go.

Issue #1431, opened 2025-09-19
Look into replacing `mypy` with `ty` once it is released


fselmo:
Astral (builders of `uv` and `ruff`) are working on a type-checker, `ty`, that is currently in preview / alpha. Look into replacing `mypy` with `ty` once this is released as we could get significant performance gains from rust here.

DeborahOlaboye:
Hi @fselmo , I’d love to work on this issue. Could you please assign it to me?

SamWilsn:
A production-ready version of ty has not yet been released. I suppose you could do some research on what is/isn't working so far (and take some performance measurements), but I don't think it's ready for us to make the switch.

DeborahOlaboye:
Alright, I'd do my research and revert. Thank you.

fselmo:
It is worth noting here that `ty` does not currently have, nor seem to plan to add, support for pydantic type coercion. `mypy` has this via the `pydantic.mypy` plugin. Astral seems to want to align `ty` more with `pyright` than `mypy` in this sense and I believe that's their goal. This means we would have to be very explicit and instantiate every type that is passed in to every pydantic model rather than the repo allowing us to let pydantic do the conversion.

This sort of defeats a lot of the "devex" / ease-of-use parts of pydantic. This would be quite warranted for our specs side of things but it does burden quite a bit the testing package where we tend to benefit from the "pydantic magic".

For this reason, this may not end up being a desirable dependency for this repo, but it's still up in the air and should be discussed / researched further.

danceratopz:
Thanks for the update @fselmo!!

I was curious so asked a friend to experiment with `ty` defaults on `src/ethereum` (which almost has no pydantic usage). It definitely is fast! Report summary below; more details: [ty-execution-specs-experiment-src-ethereum.md](https://github.com/user-attachments/files/24469320/ty-execution-specs-experiment-src-ethereum.md).

**Date:** 2026-01-07
**Git Ref:** `d618e5259a06179d7556af1961166cb7bc5da1e6`
**Versions:** ty 0.0.9, mypy 1.17.0
**Target:** 1062 source files in `src/ethereum`

#### Key Findings

- **ty finds 128 errors, mypy finds 0** — all ty errors are false positives or external library issues.
- **ty is ~16x faster cold, ~3.5x slower cached** than mypy.
- **Conclusion** All errors are false positives or stub issues. ty (v0.0.9) is too immature for this codebase's type patterns. Re-evaluate at v1.0+.
[...]

PR #2555, opened 2026-03-25
feat(ci,tooling): replace `tox` with `just`

danceratopz:
## 🗒️ Description

Replaces `tox` with `just`. Try this branch by installing `just` ([docs](https://just.systems/man/en/packages.html)):
- `brew install just`
- `apt install just`
- `emerge -av dev-build/just`
- `uv tool install just-bin`
- curl-install: https://just.systems/man/en/pre-built-binaries.html#pre-built-binaries

Then run `just` (without arguments) to see which tasks (="recipes") are available.

### Changes:

- `tox.ini` gets replaced by `Justfile`.
- tox envs get replaced by just "recipes".
- many tox envs got renamed, e.g., 
  - `tox -e py3` -> `just fill`, 
  - `tox -e tests_benchmark_pytest_py3` -> `just test-tests-bench`.
- Separate recipes for each "static" check additionally added.
- `setup-uv` action also install `just` so its available everywhere.

#### Some nice features, not all just-specific

1. Fill all EIP-7928 tests (via path for fast collection), exit on first fail, then generate an HTML coverage report:
    ```
    just fill tests/amsterdam/eip7928_block_level_access_lists/ -x  # easy arg pass-through is just-specific
    just coverage  # new recipe
    ```
2. Running `just` (without args) shows you all recipes, grouped by area (this is a native just-specific feature; see screenshot below):
    ```
    Available recipes:
        [benchmark tests]
        bench-gas *args           # Fill benchmark tests with --gas-benchmark-values
        bench-opcode *args        # Fill benchmark tests with --fixed-opcode-count 1
        bench-opcode-config *args # Run benchmark_parser, then fill benchmark tests using its config
        test-tests-bench *args    # Run benchmark framework unit tests (with Python)
    
        [consensus tests]
        coverage                  # Generate HTML coverage report from last just fill run
        fill *args                # Fill the consensus tests using EELS (with Python)
    
        [docs]
        changelog                 # Validate docs/CHANGELOG.md entries
        docs                      # Build HTML site documentation with mkdocs
        docs-fast                 # Build HTML site documentation with mkdocs (skip test case reference)
        docs-spec                 # Generate documentation for EELS using docc
        lint-md                   # Lint markdown files (markdownlint)
        ...
    ```

#### Other changes

- **Per-recipe output isolation.** All output goes to `.just/<recipe>/` (fixtures, tmp, logs, coverage). No clobbering of user-generated fixture directories.
- **Failure hints.** `spellcheck` and `lock-check` print fix commands on failure (e.g. `just whitelist <word>`).

#### `just` help output

image

#### `just` w/tab completion

image

### `tox` env-to-recipe migration helper

image

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

image


.github/actions/setup-uv/action.yaml
```diff
@@ -23,6 +23,7 @@ runs:
     - name: Prefer uv-managed Python over system Python
       shell: bash
       run: echo "UV_PYTHON_PREFERENCE=only-managed" >> "$GITHUB_ENV"
-    - name: Install tox
-      shell: bash
-      run: uv tool install tox
+    - name: Install just
+      uses: taiki-e/install-action@e24b8b7a939c6a537188f34a4163cb153dd85cf6  # v2.69.1
+      with:
+        tool: just@1.46.0
```
SamWilsn: Should this be pinned?

Justfile
```diff
@@ -0,0 +1,291 @@
+set positional-arguments := true
+
+# List available recipes
+default:
+    @just --list
```
SamWilsn:
```suggestion
# List available recipes
[default]
default:
    @just --list
```

Just to make it order independent?

Justfile
```diff
@@ -0,0 +1,291 @@
...
+
+# Check EELS import isolation
+[group('static analysis')]
+ethereum-spec-lint:
+    uv run ethereum-spec-lint
+
+# Verify uv.lock is up to date
+[group('static analysis')]
+lock-check:
+    #!/usr/bin/env bash
```
SamWilsn, line 72: Dumb question, but do we ever want non-unix compatibility?

Justfile
```diff
@@ -0,0 +1,291 @@
...
+        -ra \
+        --show-capture=no \
+        --disable-warnings \
+        -m "eels_base_coverage and not derived_test" \
+        -n auto --maxprocesses 7 \
+        --dist=loadgroup \
+        --basetemp="{{ output_dir }}/pypy3/tmp" \
+        --log-to "{{ output_dir }}/pypy3/logs" \
+        --clean \
+        --until {{ latest_fork }} \
```
SamWilsn: Can this ever have a space?

Justfile
```diff
@@ -0,0 +1,291 @@
...
+        --ignore=tests/ported_static \
+        "$@" \
+        tests
+
+# --- Integration Tests ---
+
+# Fill the base coverage consensus tests and run EELS against the fixtures
+[group('integration tests')]
+json_loader *args:
+    @mkdir -p "{{ output_dir }}/json_loader/tmp"
```
SamWilsn: I guess we lose the per-interpreter temp files directory. Is that an issue?

pyproject.toml
SamWilsn: This should probably match the path in the Justfile

SamWilsn (commented):
Looks reasonable. What's the _just_ification for the change?

Justfile
danceratopz: Just to clarify? Do you mean per Python interpreter temp files directories (i.e., separate subdirs for python3.12/3.13/...)?

Justfile
danceratopz: Buenos Aires? 🫠

Justfile
SamWilsn: Yeah, I think tox clears the temp dir if you change interpreters.

.github/actions/setup-uv/action.yaml
danceratopz: Ah, thanks for pointing this out. I think so yes, to avoid surprises in CI. But we can relax the constraint, to only `@major` or `@major.minor`. I'll update it to the latter so we get patch updates automatically.

Justfile
danceratopz: Good catch! I decided to rename the recipe as `list` and make it private, to keep the `just` (=`just list`) output cleaner. There's no functional impact of making it private; it's just not listed in the output.

Justfile
danceratopz: I don't think so tbh :sweat_smile: I think WSL on Windows should suffice! But we could potentially lean on this if becomes a blocker: https://github.com/casey/just/blob/4e9d198268a092ac59ca553580f405afac337ab3/examples/cross-platform.just

Justfile
danceratopz: Added quotes.

Justfile
danceratopz: The `tox` docs are pretty vague https://tox.wiki/en/4.50.3/reference/config.html#temp_dir. It certainly doesn't seem to be per-invocation, but could well be upon interpreter switch.

Switching interpreters is something that we don't do per env in `tox` (or recipe in `just`). Unless the user decides to suddenly pin a different python version via, e.g., `uv python pin 3.13`. In CI this will not be cached. If we decided we need to run the same command with different interpreters, then we would probably have different recipes (depending on a base recipe for the underlying command) for which we can specify different tmp dirs (above, for example, it's scoped to the `json_loader` via `{{ output_dir }}/json_loader/tmp`).

You made me wonder if we should always clear this directory between invocations, but I suppose it's fine without. Happy to be convinced otherwise. For fixture generation we lean on `fill --clean` to ensure that the output fixture directory doesn't get polluted across runs.

pyproject.toml
danceratopz: This isn't used by tox at the mo, which runs:
```
docc --output /path/to/execution-specs/.tox/docs
```

I think it would be cleaner if it matched, but they'll be prone to go out of sync. Updated to:
`.just/spec-docs/`

Justfile
danceratopz: Or we could add inline Python scripts via https://just.systems/man/en/python-recipes-with-uv.html#python-recipes-with-uv

danceratopz:
> Looks reasonable. What's the _just_ification for the change?

@SamWilsn The tldr is that tox is a heavier tool than we need since the introduction of `uv` and that switching to `just` gives us the slickest config and cli ergonomics that I could find. 

Also, the team unanimously claimed to strongly dislike tox when we discussed in a Jan team meeting :smile: 

I put together a doc and [posted it as a discussion](https://github.com/ethereum/execution-specs/discussions/2559) that aims to provide a neutral comparison of tox/make/just, please check the tables for an overview of dev ergonomics! Happy to discuss there!

Justfile
danceratopz: Also made `help` an alias of `list`.

spencer-tb (commented):
In favor of just! Nice addition! When ty

felix314159:
lgtm

danceratopz:
If we do migrate to just, I'd propose that we additionally clean up the new recipe names to something along the lines of (open to suggestions!):

Current | Proposed
-- | --
typecheck | type
ethereum-spec-lint | spec-lint
format-check | fmt
tests_pytest_py3 | test
tests_pytest_pypy3 | test-pypy
tests_benchmark_pytest_py3 | test-bench
pypy3 | fill-pypy
spellcheck | spell
lock-check | lock
markdownlint | mdlint
benchmark-gas-values | bench-gas
benchmark-fixed-opcode-cli | bench-opcode-cli
benchmark-fixed-opcode-config | bench-opcode-config

I'd also consider including a minimal `tox.ini` that still has the current envs, but instead of running the env as-is today, it simply links just installation help and prints the command to run the env as a just recipe.
E.g.
```
$ tox - e py3
execution-specs has migrated to just: https://just.systems/man/en/packages.html
this tox env can be ran using just via: 
just fill
```

danceratopz:
> In favor of just! Nice addition! When ty

@spencer-tb tnx! When `ty`? I guess when it gets less strict and accepts pydantic magic! That's the current blocker, see @fselmo's https://github.com/ethereum/execution-specs/issues/1431#issuecomment-3713265090. Iirc the tldr is that we'd have to provide the correct types whenever you instantiate pydantic models (which is everywhere in our tests). This breaks the useful devx around pydantic (you can be relaxed with inputs as it coerces the correct type for you).

.github/actions/setup-uv/action.yaml
SamWilsn: I'd prefer going in the opposite direction, and using a full commit hash.

Justfile
SamWilsn: We probably shouldn't clear it every run, there's probably some incremental compilation stuff in there, though I've definitely seen tox clearing the cache when I change interpreters.

SamWilsn (approved):

.github/actions/setup-uv/action.yaml
danceratopz: Ah, I see. Good point and I appreciate this level of paranoia. This action does not support commit hashes in `@version` syntax. It does support "artifact attestations" (`gh-attestation`) but unfortunately "casey/just" does not (`uv` does btw). More info here in its readme: https://github.com/taiki-e/install-action/blob/e9e8e031bcd90cdbe8ac6bb1d376f8596e587fbf/README.md?plain=1#L101

I think a better security model would be to either:
1. Build and install from source at a pinned version (which we then cache).
2. `apt install` from Ubuntu, it should be available on Github's and our custom runners.

I tried 2. and yes `just` is in the PPA, but it installs `1.21` which doesn't have some of the features we use 😭 I will revert. 

Made a follow-up issue for 1. if we think it's worthwhile:
- https://github.com/ethereum/execution-specs/issues/2584

danceratopz:
@SamWilsn the last 3 commits addressed minor existing issues in this PR. There were two other, larger changes I wanted to add in the scope of this PR: I've put them here in https://github.com/danceratopz/execution-specs/pull/57, so they can be considered in isolation. Could you please review them?

marioevz (approved):
LGTM. Thanks!

PR #2562, opened 2026-03-26
feat(test-benchmark): parametrize stateful benchmark stubs from `--address-stubs` 

LouisTsai-Csie:
## 🗒️ Description

This PR replaces hardcoded token lists with `stub_parametrize` marker. 

In the past, we only had two stub files: `stubs_mainnet.json` and `stubs_bloatnet.json`. In `tests/benchmark/stateful/helpers.py`, the stub selection logic was hardcoded based on the network being run:

```python
# Load token names from stubs_bloatnet.json for test parametrization
_STUBS_FILE = Path(__file__).parent / "bloatnet" / "stubs_bloatnet.json"
with open(_STUBS_FILE) as f:
    _STUBS = json.load(f)
```

This made it difficult to support additional stub files. For example, we now need a separate repricing stub file (see `tests/benchmark/stateful/stubs/stubs_repricing.json` in this PR) that restricts the ERC20 tokens used during execution.

Required tokens:
- 50GB_ERC20
- 30GB_ERC20
- 9_39GB_ERC20
- 1GB_ERC20
- 4_23MB_ERC20

The goal of this PR is to introduce a more flexible way of handling stubs. Instead of relying on hardcoded logic, we now use a Pydantic model to load and validate stub files.

For example, when fetching stub tokens for tests named test_sstore_erc20_approve_<token>, we previously handled this as follows:

```python
# inside the helper function, we parse the stub account like this

SSTORE_TOKENS = [
    k.replace("test_sstore_erc20_approve_", "")
    for k in _STUBS.keys()
    if k.startswith("test_sstore_erc20_approve_")
]

# test parametrization
@pytest.mark.parametrize("token_name", SSTORE_TOKENS)
def test_sstore_erc20_approve(
...
    token_name: str
) -> None:
    # Stub Account
    erc20_address = pre.deploy_contract(
        code=Bytecode(),
        stub=f"test_sstore_erc20_approve_{token_name}",
    )
```

This approach depends on manually parsing stub keys, which was not very clean or extensible. In this PR, that logic is replaced with a more structured approach.

```python
@pytest.mark.stub_parametrize("erc20_stub", "test_mixed_sload_sstore_")
@pytest.mark.parametrize(
    "sload_percent,sstore_percent",
    [
        pytest.param(10, 90, id="10-90"),
        ...
    ],
)
def test_mixed_sload_sstore(
    ...
    erc20_stub: str,
    ...
) -> None:
```

This makes the stub configuration explicit at the test level, removes the need for manual key parsing in helper code, and provides a more maintainable foundation for supporting additional stub dimensions such as repricing.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


spencer-tb (approved):
LGTM!

The old code produced test IDs like 1KB, 2KB, but the new `parametrize_args` doesn't uppercase, so the IDs become 1kb, 2kb now. Just mentioning as the file names will change. Maybe you want to update here to be consistent etc

danceratopz:
I don't really have the context, but about: 
```
-k "4_23MB_ERC20 or 1GB_ERC20 or 9_39GB_ERC20 or 30GB_ERC20 or 50GB_ERC20"
```
Do we need a marker that would help to select these tests here? Or is it just to help test?

LouisTsai-Csie:
Hi @danceratopz i am sorry for the confusion, the `-k` command here is just a note for me. I was trying to see if this PR has the same effect of running with the filtering command. 

Update more context in the PR description also 👍🏻 Hope this could help better understand the intention of this PR.

packages/testing/src/execution_testing/__init__.py
```diff
@@ -135,6 +136,7 @@
     "BenchmarkCodeGenerator",
     "BenchmarkTest",
     "BenchmarkTestFiller",
+    "StubConfig",
```
danceratopz: Meganit: Could you move this so that the list is ordered alphabetically?

tests/benchmark/conftest.py
```diff
@@ -4,15 +4,26 @@
...
+from execution_testing import Fork, StubConfig
 
 DEFAULT_BENCHMARK_FORK = "Prague"
 
+_stub_config_key = pytest.StashKey[StubConfig]()
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    """Build StubConfig from ``--address-stubs``."""
+    address_stubs = config.getoption("address_stubs", default=None)
```
danceratopz, line 16: Previously, tests would have always been parametrized with some stubs, right. But this has changed, now no test cases will be generated, unless address stubs are provided.

I.e., this is not a pure refactor.

danceratopz (commented):
Hey @LouisTsai-Csie, the `--stub-addresses` flag from #2512 for `fill` doesn't exist here yet, right? If you rebased, then I think I could test the logic using `fill`? 

Initial sweep below, would love to get a feel for the behavior via `fill`, if possible.

tests/benchmark/conftest.py
LouisTsai-Csie: I didn’t realize this changes the parametrization behavior. What I want is:
- If a test does not use stubs, it should always be collected, regardless of whether a stub file is provided.
- If a test uses stubs, all available stubs should be included during parametrization at collection time.

For example:
- All `test_account_access` tests should be collected, even if no stub file is provided.
- If a mainnet stub file is provided (let's assume there is ERC20 A, B, C, ...) for the `test_sstore_erc20_approve` benchmark, ERC20 A, B, C, .... should be used for parametrization during test collection -> this depends on the stub file provided.

LouisTsai-Csie: Example Command:

```bash
uv run fill \
  -v \
  --clean \
  --evm-bin <path-to-evm-bin> \
  --gas-benchmark-values 120 \
  --fork Osaka \
  tests/benchmark/stateful/bloatnet/test_single_opcode.py::test_sload_erc20_generic \
  --address-stubs tests/benchmark/stateful/stubs/stubs_bloatnet.json \
  --rpc-endpoint <rpc-endpoint>
```

danceratopz: This is working nicely! About the change in behavior. If no address stubs are provided, not tests are collected, but I think that's the expected behavior in that case. But perhaps it should be a required command-line arg if a test requiring a stub is required.

If you like I can look into this for you?

For me it's "ok" to live it as-is. It's just not great dx, because you get a warning that some tests are skipped: 
```
tests/benchmark/conftest.py:51
  /home/dtopz/code/github/execution-specs.review-2562/tests/benchmark/conftest.py:51: UserWarning: stub_parametrize: no stubs matched prefix 'test_sload_empty_erc20_balanceof_' for test_sload_erc20_generic; test will be skipped
    values, ids = stub_config.parametrize_args(
```
... but you're not told how to alleviate the problem. 

But perhaps our tight-knit stateful testing community can handle that :-D

danceratopz: About the change, before, these tests were generated because the stub parametrization values were hard-coded. No big deal.

danceratopz (approved):
Thanks @LouisTsai-Csie!

Issue #2534, opened 2026-03-20
Refactor EL request helpers to avoid shared param mutation and add cross-format fixture regression checks


felix314159:
Felipe wrote the [blueprint](https://github.com/ethereum/execution-specs/pull/2532#issuecomment-4093291174) for this issue. I added wishes of my own and had an LLM summarize the situation so we get:

## Context

#2532 fixed a bug in branch `eips/amsterdam/eip-8037` where `WithdrawalRequestContract.tx_gas_limit` was double-accumulated across fixture format runs because pytest parameter objects are shared by reference. The immediate fix moved the gas adjustment into `transactions()` as a pure function.

That addressed the concrete bug, but the broader anti-pattern still exists: the EL request helper objects used in the request tests are still mutated in `update_pre()`. Today these mutations are idempotent assignments, so they do not currently produce incorrect fixtures, but the pattern is fragile. A future `+=`, conditional mutation, or other stateful update could silently reintroduce divergence between `blockchain_test` and `blockchain_test_engine`.

There is also a separate footgun now: some behavior depends on callers remembering to pass `fork`, and a missed `fork` would not be caught by the mutation warning hook because nothing is being mutated in that case.

## Scope of what to do

Refactor the EL request helper flows in:

* `tests/prague/eip6110_deposits`
* `tests/prague/eip7002_el_triggerable_withdrawals`
* `tests/prague/eip7251_consolidations`

so that `update_pre()` no longer mutates shared parametrized helper instances. Instead, it should return derived/prepared values, and downstream methods such as `transactions()` and `valid_requests()` should consume those values explicitly.

## Guardrails

1. Add a warning-based guard first:
* Keep the filler hook that detects mutation of parametrized values during a test run.
* Use it to identify and clean up all remaining EL request helper mutation sites.

2. Add an explicit end-to-end regression check:
* Add a CI workflow script that compares relevant hashes between generated `blockchain_test` and `blockchain_test_engine` fixtures.

This should catch cross-format divergence directly, including cases that would not be detected by the mutation hook alone (for example, if a required fork is not passed somewhere).

3. Tighten the guard once the tree is clean:
* After the remaining mutation warnings are eliminated, flip the filler hook from warning to hard failure.

## Why both checks (mutation hook + explicit hash comparisons) are needed

* The mutation hook is a good guardrail against shared-object mutation bugs.
[...]

PR #2566, opened 2026-03-26
fix(test-tests,ci): fix flakiness due to mutated `EnvironmentDefaults.gas_limit` with xdist

felix314159:
## Summary
Fixes unit tests that 'sometimes' would be failing, due to a flaky `pytest -n auto` failure caused by worker-local leakage of `EnvironmentDefaults.gas_limit`.

You can reproduce the problem (before applying the PR) by navigating to:
`cd packages/testing`

and then repeatedly running:
	
`uv run pytest -n auto -p no:cacheprovider --ignore=src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_benchmarking.py src -v -s`

e.g:
* 1. try: `29 failed, 1416 passed`
* 2. try: `14 failed, 1431 passed`
* 3. try: `22 failed, 1423 passed`
* 4. try: `29 failed, 1416 passed`

after applying the PR you consistently get:
	
* `1447 passed` (the PR also adds two new unit tests so that this is detected before it can happen again)

## Problem
Some pytest plugins were mutating the global `EnvironmentDefaults.gas_limit` during `pytest_configure`, even when the corresponding CLI option had not been explicitly passed by the user.

That was mostly harmless in a fresh process, but it became a problem in nested in-process pytest runs and under xdist:

* a nested pytest session could rewrite the global gas limit
* later tests on the same worker would inherit the mutated value
* depending on worker scheduling, this produced different numbers of failures under `-n auto`

In practice this showed up as occasional unit test CI failures with bogus tiny gas limits and errors like `gas used exceeds limit`.

## Root Cause
Both `fill` and `execute` use config.getoption(...) to decide whether to override the global gas limit.

That meant the override happened not only for explicit flags like `--transaction-gas-limit` or `--block-gas-limit`, but also when pytest was just returning the option default value.

In `execute` mode this was especially bad because the default transaction gas limit is derived from the current global block gas limit, so repeated nested sessions could keep shrinking the global value.

## Fix
Make it explicit-only:

* add a shared helper to detect whether a long CLI option was actually passed on the command line
* only override `EnvironmentDefaults.gas_limit` in `execute` when `--transaction-gas-limit` was explicitly provided
* only override `EnvironmentDefaults.gas_limit` in `fill` when `--block-gas-limit` was explicitly provided
* reset `EnvironmentDefaults.gas_limit` to the canonical default before and after each filler test to avoid cross-test worker contamination
* add regression tests covering the explicit-vs-default execute behavior

Note: The user-facing feature is preserved:

* `--transaction-gas-limit` still overrides the default when explicitly requested
* `--block-gas-limit` still overrides the default when explicitly requested

What changes is only the accidental behavior where merely loading a plugin or using default option values could mutate global state.

## 🔗 Related Issues or PRs
I originally started this work when looking into why a [harmless exeception mapper update PR](https://github.com/ethereum/execution-specs/pull/2550) had py3 failures. The problem with silencing those failures with 'rebasing' is that it does not fix the underlying issue, it re-runs CI and when you get a lucky (or rather unlucky) worker distribution you might pass the tests without noticing the sleeper bug.

The fix is also related to ongoing work started with [#2534](https://github.com/ethereum/execution-specs/issues/2534) which recognized that there is a problem in our codebase where we modify shared objects and end up with subtle / sometimes hard-to-reproduce failures.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py
@@ -26,6 +26,7 @@ from ..shared.helpers import (
     get_spec_format_for_item,
     is_help_or_collectonly_mode,
     labeled_format_parameter_set,
+    option_was_explicitly_set,
 )
 from ..spec_version_checker.spec_version_checker import EIPSpecTestItem
 from .pre_alloc import Alloc
@@ -200,8 +201,9 @@ def pytest_configure(config: pytest.Config) -> None:
        called before the pytest-html plugin's pytest_configure to ensure that
        it uses the modified `htmlpath` option.
     """
-    # Modify the block gas limit if specified.
-    if config.getoption("transaction_gas_limit"):
+    # Keep execute-mode overrides working, but avoid rewriting the global
+    # default when this plugin is merely imported into nested pytest sessions.
+    if option_was_explicitly_set(config, "--transaction-gas-limit"):
         EnvironmentDefaults.gas_limit = config.getoption(
             "transaction_gas_limit"
         )
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/tests/test_execute.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/tests/test_execute.py
@@ -1,0 +1,57 @@
+"""Regression tests for execute plugin configuration."""
+
+from types import SimpleNamespace
+from typing import Any
+
+from execution_testing.test_types.block_types import EnvironmentDefaults
+
+from ..execute import pytest_configure
+
+
+class FakeConfig:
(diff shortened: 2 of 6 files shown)


spencer-tb (approved):
LGTM!

I had a similar issue with `test_filler.py:801` so added an explicit `@pytest.mark.usefixtures("restore_environment_defaults")` which is now redundant since the fixture is `autouse=True`.

Could we remove that line for that test?

felix314159:
ok, i'll let CI complete to ensure everything passes, and then add another commit that removes the line
edit: all passed

spencer-tb:
How'd you get that commit hash @felix314159... did you do some magic

felix314159:
must be a coincidence

PR #2572, opened 2026-03-26
refactor(test-vm): optimize `Bytecode.__mul__` and `__add__` performance 

kevaundray:
## 🗒️ Description

From my understanding, we had this loop:

```python
  output = self                                                                                                            
  for _ in range(other - 1):                                                                                               
      output += self 
```

Each `output += self` calls `__add__`, which creates a new bytes object via bytes(self) + bytes(other). 

Since python bytes are  immutable, each concatenation allocates a new buffer and copies all previous bytes into it.

So:

  For `self * n` where `len(self) = k`:                                                                                        
  - Iteration 1: copies k + k = 2k bytes
  - Iteration 2: copies 2k + k = 3k bytes                                                                                  
  - ...                                                           
  - Iteration n-1: copies nk bytes  

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

:cat2:

### DIFF
--- a/packages/testing/src/execution_testing/vm/bytecode.py
+++ b/packages/testing/src/execution_testing/vm/bytecode.py
@@ -227,7 +227,7 @@ class Bytecode:
         )
 
         return Bytecode(
-            bytes(self) + bytes(other),
+            self._bytes_ + other._bytes_,
             popped_stack_items=c_pop,
             pushed_stack_items=c_push,
             min_stack_height=c_min,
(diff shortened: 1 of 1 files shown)


kevaundray:
Looking mainly at  `Benchmarking / Benchmark Gas Values (pull_request)` this seems to shave off about 2 minutes -- checking the previous PR

packages/testing/src/execution_testing/vm/bytecode.py
danceratopz: Let's also fix-up this pre-existing copy-paste error:
```suggestion
        Repeat the bytecode a given number of times.
```

danceratopz (approved):
Good find @kevaundray! Thanks for this. 

Looks like we have adequate unit tests for this, but I'm paranoid so I `hasher` diff'd these fixtures:
```
fill --no-html --fork=Osaka --gas-benchmark-values 1 --output=fixtures -n 8 --clean --evm-bin=evmone-t8n -m benchmark
```
against 365433e and it was clean.

Only a single run datapoint, but there's consistent improvement across jobs! 🚀 
Image

PR #2588, opened 2026-03-30
feat(benchmarks): add BAL contract chase

jochem-brouwer:
## 🗒️ Description
This test re-introduces the contract chase benchmark in #2033 which looks wrong there. The current status is that this appears to consume about 47% of the gas it has available. Since we want to bench using most of the available gas this is something to look into to find out if we can spend this on something else in this benchmark.

Note: builds on #2033, merge that prior to merging this. 

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/eip7928_block_level_access_lists/test_block_access_lists_cross_contract_chase.py
+++ b/tests/benchmark/stateful/eip7928_block_level_access_lists/test_block_access_lists_cross_contract_chase.py
@@ -1,0 +1,325 @@
+"""
+Tests for EIP-7928 BAL with cross-contract pointer chasing.
+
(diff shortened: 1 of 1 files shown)


jochem-brouwer:
CCing [this](https://github.com/ethereum/execution-specs/pull/2033#issuecomment-4155467821) comment from @fselmo here:

> I think this is 63/64 rule compounding with each nested call. We probably need to figure out the actual gas limit used since it doesn't use the per-tx-gas-cap. Then, we can squeeze more transactions in the block.

I tested this on a gas limit of 1M and this executed ~470k gas, so this might be part of the problem (txs not using full cap)

Issue #2474, opened 2026-03-10
Guest program: should it enforce execution witness canonicality?


jsign:
Today, the execution witness generation in the specs builds a minimal and canonical execution witness. However, the stateless guest only uses the witness to execute the block — it does not verify that:
- No more information than strictly needed is provided (e.g. no extra or duplicated entries in fields)
- Field ordering complies with the canonical witness definition (i.e. lexicographic or block-number ordering, depending on the field)

Strictly speaking, these checks aren't required for correct stateless execution. One could also argue that being strict increases the risk of consensus failure, since it opens the door to attacks that exploit witness-generation bugs in EL clients. For example: if some EL client generates a witness with extra (but harmless) data for an edge case not covered by tests, enforcing canonicality in the guest would reject that witness — causing a consensus split even though the block itself is valid.

There are also performance implications: canonicality checks (sorted order, no duplicates, minimality) must be proven inside the zkVM, adding overhead beyond what's needed for execution correctness.

Finally, this determines whether we should create test cases for witness canonicality validation.

cc @kevaundray for visibility/discussion

PR #2589, opened 2026-03-30
feat(zkevm): harden witness state validation for malformed trie proofs

jsign:
## 🗒️ Description
This PR:
- Added some extra execution witness completeness tests -- not main intention of this PR, but useful to add more coverage.
- Tighten Amsterdam `WitnessState` and incremental MPT witness decoding so stateless execution rejects malformed or incomplete trie proofs, including unresolved hashed nodes, and empty extension paths.
- Add execution-witness tests covering missing account and storage proof nodes, absent-slot and absent-account proofs, delete-collapse auxiliary nodes, and reverted or failed-call read scenarios.

The last bullet basically means testing invalid `ExecutionWitness.state` regarding missing information for correctly verifying the witness. 

I also believe eventually we need to also cover invalid `ExecutionWitness.state` test cases for "structurally invalid MPTs". I already added the checks for this in the spec in this PR. For an idea of which cases to add, I wrote a document [here](https://hackmd.io/@jsign/invalid-decoded-mpt-test-cases) -- I think this probably needs more helpers in the testing framework to construct these tests nicely.

```
> uv run fill --clean -m "blockchain_test" --fork Amsterdam ./tests/amsterdam/eip8025_optional_proofs     
===================================================================== test session starts ======================================================================
platform linux -- Python 3.12.3, pytest-8.4.2, pluggy-1.6.0
Generating fixtures for: BPO2ToAmsterdamAtTime15k, Amsterdam
2.19.0
Log file: logs/fill-20260330-164614-main.log
rootdir: /home/ignacio/code/execution-specs
configfile: packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-fill.ini
plugins: cov-4.1.0, metadata-3.1.1, json-report-1.5.0, xdist-3.8.0, regex-0.2.0, custom-report-1.0.1, html-4.1.1
collecting ...  pytest-regex selected 328 tests to run for regex: .*
collected 492 items / 164 deselected / 328 selected                                                                                                            

tests/amsterdam/eip8025_optional_proofs/test_witness_headers.py ..                                                                                    [  2/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_7702.py .................                                                                        [ 19/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_call_variants.py ........                                                              [ 27/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_contract_creation.py ..........                                                        [ 37/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_eoa_precompiles.py ...................                                                 [ 56/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_extcode.py ............................................                                [100/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_selfdestruct.py .......                                                                [107/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_bytecodes_system_contracts.py .                                                                  [108/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_headers.py ..............                                                                        [122/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_state_deletes.py .....                                                                           [127/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_state_invariants.py .                                                                            [128/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_state_reads.py .....                                                                             [133/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_state_replay_order.py ..                                                                         [135/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_state_writes.py ....                                                                             [139/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_validation_codes.py ...........                                                                  [150/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_validation_headers.py .....                                                                      [155/164]
tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py .........                                                                    [164/164]

======================================================================= warnings summary =======================================================================
.venv/lib/python3.12/site-packages/_pytest/config/__init__.py:812
  /home/ignacio/code/execution-specs/.venv/lib/python3.12/site-packages/_pytest/config/__init__.py:812: PytestAssertRewriteWarning: Module already imported so cannot be rewritten; execution_testing.cli.pytest_commands.plugins.filler.pre_alloc
    self.import_plugin(arg, consider_entry_points=True)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
--------------------------------------- Log file: /home/ignacio/code/execution-specs/logs/fill-20260330-164614-main.log ----------------------------------------
------------------------------- Generated html report: file:///home/ignacio/code/execution-specs/fixtures/.meta/report_fill.html -------------------------------
================================================  T8n cache: 164 unique test groups, no cache sharing possible =================================================
=================================  No tests executed - the test fixtures in "fixtures/" may now be executed against a client  ==================================
======================================================= 164 passed, 164 deselected, 1 warning in 35.36s ========================================================
```

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/incremental_mpt.py
+++ b/src/ethereum/forks/amsterdam/incremental_mpt.py
@@ -903,9 +903,7 @@ def _resolve_child_ref(
             f"Unexpected child ref length: {len(ref_bytes)}"
         )
         if ref_bytes in node_db:
-            return _decode_witness_node(
-                node_db, node_db[ref_bytes], secured
-            )
+            return _decode_witness_node(node_db, node_db[ref_bytes], secured)
         return HashedNode(_hash=ref_bytes)
     else:
         return _decode_witness_node(node_db, rlp.encode(child_ref), secured)
@@ -925,6 +923,8 @@ def _decode_witness_node(
         Mapping from node hash to RLP-encoded node data.
     rlp_bytes :
         The RLP-encoded node to decode.
+    secured :
+        Whether the trie uses secured (hashed) keys.
 
     """
     node_hash: Optional[Bytes] = None
--- a/src/ethereum/forks/amsterdam/incremental_mpt.py
+++ b/src/ethereum/forks/amsterdam/incremental_mpt.py
@@ -886,6 +886,7 @@ def compact_to_nibbles(compact: Bytes) -> Tuple[Bytes, bool]:
 def _resolve_child_ref(
     node_db: Dict[Bytes, Bytes],
     child_ref: Extended,
+    secured: bool,
 ) -> MutableNode:
     """
     Resolve a child reference from an RLP-decoded trie node.
@@ -902,15 +903,18 @@ def _resolve_child_ref(
             f"Unexpected child ref length: {len(ref_bytes)}"
         )
         if ref_bytes in node_db:
-            return _decode_witness_node(node_db, node_db[ref_bytes])
+            return _decode_witness_node(
+                node_db, node_db[ref_bytes], secured
+            )
         return HashedNode(_hash=ref_bytes)
     else:
-        return _decode_witness_node(node_db, rlp.encode(child_ref))
+        return _decode_witness_node(node_db, rlp.encode(child_ref), secured)
 
 
 def _decode_witness_node(
     node_db: Dict[Bytes, Bytes],
     rlp_bytes: Bytes,
+    secured: bool,
 ) -> MutableNode:
     """
     Decode an RLP-encoded trie node into a ``MutableNode``.
@@ -950,7 +954,11 @@ def _decode_witness_node(
                 _rlp=rlp_bytes,
             )
         else:
-            child = _resolve_child_ref(node_db, decoded[1])
+            assert len(nibbles) > 0 or not secured, (
+                "ExtensionNode must have a non-empty path "
+                "in state/storage trie"
+            )
+            child = _resolve_child_ref(node_db, decoded[1], secured)
             assert isinstance(child, (MutableBranchNode, HashedNode)), (
                 "ExtensionNode child must be a BranchNode"
             )
@@ -964,13 +972,15 @@ def _decode_witness_node(
     elif len(decoded) == 17:
         children: List[Optional[MutableNode]] = []
         for i in range(16):
-            children.append(_resolve_child_ref(node_db, decoded[i]))
+            children.append(_resolve_child_ref(node_db, decoded[i], secured))
         value_raw = decoded[16]
         if isinstance(value_raw, (bytes, bytearray)):
             value = Bytes(value_raw)
         else:
             value = b""
-        # TODO: value is always empty in practice; refactor
+        assert value == b"" or not secured, (
+            "BranchNode value must be empty in state/storage trie"
+        )
         occupied = 16 - children.count(None) + (value != b"")
         assert occupied >= 2, "BranchNode must have at least 2 children"
         return MutableBranchNode(
@@ -1022,7 +1032,7 @@ def decode_witness_to_mpt(
         )
 
     root_rlp = node_db[root_hash]
-    root_node = _decode_witness_node(node_db, root_rlp)
+    root_node = _decode_witness_node(node_db, root_rlp, secured)
 
     return IncrementalMPT(
         secured=secured,
--- a/src/ethereum/forks/amsterdam/incremental_mpt.py
+++ b/src/ethereum/forks/amsterdam/incremental_mpt.py
@@ -973,7 +973,9 @@ def _decode_witness_node(
         else:
             value = b""
         occupied = 16 - children.count(None) + (value != b"")
-        assert occupied >= 2, "BranchNode must have at least 2 children"
+        assert occupied >= 2, (
+            "BranchNode must have at least 2 occupied entries"
+        )
         return MutableBranchNode(
             children=children,
             value=value,
--- a/src/ethereum/forks/amsterdam/incremental_mpt.py
+++ b/src/ethereum/forks/amsterdam/incremental_mpt.py
@@ -886,7 +886,6 @@ def compact_to_nibbles(compact: Bytes) -> Tuple[Bytes, bool]:
 def _resolve_child_ref(
     node_db: Dict[Bytes, Bytes],
     child_ref: Extended,
-    secured: bool,
 ) -> MutableNode:
     """
     Resolve a child reference from an RLP-decoded trie node.
@@ -903,16 +902,15 @@ def _resolve_child_ref(
             f"Unexpected child ref length: {len(ref_bytes)}"
         )
         if ref_bytes in node_db:
-            return _decode_witness_node(node_db, node_db[ref_bytes], secured)
+            return _decode_witness_node(node_db, node_db[ref_bytes])
         return HashedNode(_hash=ref_bytes)
     else:
-        return _decode_witness_node(node_db, rlp.encode(child_ref), secured)
+        return _decode_witness_node(node_db, rlp.encode(child_ref))
 
 
 def _decode_witness_node(
     node_db: Dict[Bytes, Bytes],
     rlp_bytes: Bytes,
-    secured: bool,
 ) -> MutableNode:
     """
     Decode an RLP-encoded trie node into a ``MutableNode``.
@@ -923,9 +921,6 @@ def _decode_witness_node(
         Mapping from node hash to RLP-encoded node data.
     rlp_bytes :
         The RLP-encoded node to decode.
-    secured :
-        Whether the trie uses secured (hashed) keys.
-
     """
     node_hash: Optional[Bytes] = None
     if len(rlp_bytes) >= 32:
@@ -954,11 +949,10 @@ def _decode_witness_node(
                 _rlp=rlp_bytes,
             )
         else:
-            assert len(nibbles) > 0 or not secured, (
-                "ExtensionNode must have a non-empty path "
-                "in state/storage trie"
+            assert len(nibbles) > 0, (
+                "ExtensionNode must have a non-empty path"
             )
-            child = _resolve_child_ref(node_db, decoded[1], secured)
+            child = _resolve_child_ref(node_db, decoded[1])
             assert isinstance(child, (MutableBranchNode, HashedNode)), (
                 "ExtensionNode child must be a BranchNode"
             )
@@ -972,15 +966,12 @@ def _decode_witness_node(
     elif len(decoded) == 17:
         children: List[Optional[MutableNode]] = []
         for i in range(16):
-            children.append(_resolve_child_ref(node_db, decoded[i], secured))
+            children.append(_resolve_child_ref(node_db, decoded[i]))
         value_raw = decoded[16]
         if isinstance(value_raw, (bytes, bytearray)):
             value = Bytes(value_raw)
         else:
             value = b""
-        assert value == b"" or not secured, (
-            "BranchNode value must be empty in state/storage trie"
-        )
         occupied = 16 - children.count(None) + (value != b"")
         assert occupied >= 2, "BranchNode must have at least 2 children"
         return MutableBranchNode(
@@ -1032,7 +1023,7 @@ def decode_witness_to_mpt(
         )
 
     root_rlp = node_db[root_hash]
-    root_node = _decode_witness_node(node_db, root_rlp, secured)
+    root_node = _decode_witness_node(node_db, root_rlp)
 
     return IncrementalMPT(
         secured=secured,
--- a/src/ethereum/forks/amsterdam/incremental_mpt.py
+++ b/src/ethereum/forks/amsterdam/incremental_mpt.py
@@ -921,6 +921,7 @@ def _decode_witness_node(
         Mapping from node hash to RLP-encoded node data.
     rlp_bytes :
         The RLP-encoded node to decode.
+
     """
     node_hash: Optional[Bytes] = None
     if len(rlp_bytes) >= 32:
@@ -949,9 +950,7 @@ def _decode_witness_node(
                 _rlp=rlp_bytes,
             )
         else:
-            assert len(nibbles) > 0, (
-                "ExtensionNode must have a non-empty path"
-            )
+            assert len(nibbles) > 0, "ExtensionNode must have a non-empty path"
             child = _resolve_child_ref(node_db, decoded[1])
             assert isinstance(child, (MutableBranchNode, HashedNode)), (
                 "ExtensionNode child must be a BranchNode"
--- a/src/ethereum/forks/amsterdam/witness_state.py
+++ b/src/ethereum/forks/amsterdam/witness_state.py
@@ -21,8 +21,12 @@ from ethereum.state import (
 )
 
 from .incremental_mpt import (
+    HashedNode,
     IncrementalMPT,
-    compact_to_nibbles,
+    MutableBranchNode,
+    MutableExtensionNode,
+    MutableLeafNode,
+    MutableNode,
     decode_witness_to_mpt,
     mpt_root,
     mpt_set,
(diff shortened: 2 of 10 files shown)


src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -970,9 +972,10 @@ def _decode_witness_node(
             value = Bytes(value_raw)
         else:
             value = b""
-        # TODO: value is always empty in practice; refactor
```
jsign, line 973: After some thinking, I think it is fine to remove this TODO. 

First, I thought that would be better for a stateless validator to reject if it deserializes a `BranchNode` with a value, since this must never happen in practice given how Ethereum uses the MPT.

But doing this feels like we are indirectly re-defining the rules of MPT as given today in the yellow paper. 

I ended up convicing myself that we don't need to add this rejection rule, since in theory if we get this kind of `BranchNode` it should never really verify against a `state_root` for an Ethereum block -- so in practice this rule is enforced anyway in practical terms.

Also, the spec already have tests which allow for a value in the `BranchNode` -- so being as strict as I wanted originally also means removing tests that exists which starts to feel quite hairy if can't be properly justified.

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -970,9 +972,10 @@ def _decode_witness_node(
...
         occupied = 16 - children.count(None) + (value != b"")
-        assert occupied >= 2, "BranchNode must have at least 2 children"
+        assert occupied >= 2, (
+            "BranchNode must have at least 2 occupied entries"
```
jsign, line 976: Nit: Given the `value != b""`, I think this is a better message.

src/ethereum/forks/amsterdam/witness_state.py
```diff
@@ -46,96 +50,56 @@ def build_code_db(code_entries: Tuple[Bytes, ...]) -> Dict[Hash32, Bytes]:
     return db
 
 
-def _resolve_node_ref(
```
jsign, line 49: This method was used before since on every trie lookup, as we walked the trie we decoded stuff.

Now that I wanted to do tighter checks in decoded nodes, this meant I needed to duplicate the checks here and in another place (i.e. `decode_witness_to_mpt`).

So basically removed this method, and concentrate the logic using that to remove duplication. 

This also means that we now deserialize all the provided MPT in the witness first, so we can cache it and avoid this re-decoding of stuff on every trie lookup, which should make runs faster (although I'm not strictly interested in that -- just another benefit).

src/ethereum/forks/amsterdam/witness_state.py
```diff
@@ -46,96 +50,56 @@ def build_code_db(code_entries: Tuple[Bytes, ...]) -> Dict[Hash32, Bytes]:
...
-            return None
-        if len(ref_bytes) == 32:
-            if ref_bytes not in node_db:
-                return None
-            return rlp.decode(node_db[ref_bytes])
-        return rlp.decode(ref_bytes)
-    return None
-
-
 def _trie_lookup(
```
jsign, line 53: Now `_trie_lookup` already receives a fully deserialized MPT instead of raw bytes.

src/ethereum/forks/amsterdam/witness_state.py
```diff
@@ -46,96 +50,56 @@ def build_code_db(code_entries: Tuple[Bytes, ...]) -> Dict[Hash32, Bytes]:
...
-    node = _resolve_node_ref(node_db, root_hash)
+    node = root_node
     pos = 0
 
     while node is not None:
-        if not isinstance(node, list):
+        if isinstance(node, HashedNode):
+            raise AssertionError(
+                "Encountered unresolved HashedNode during witness lookup"
+            )
```
jsign, line 75: This is an important bug fix -- the previous code did not fail if found a `HashedNode`!

To explain more in detail:
1. `_resolve_node_ref(..)` (method removed above) returned `None` if a reference wasn't in `node_db`.
2. In the previous impl of this `_trie_lookup(...)`, if `_resolve_node_ref(...)` returned `None`, we exit the main while-loop (L71) and returned `None`.

This was not correct.

Now that `_trie_lookup(...)` receives a deserialized MPT, we have a much clear way to fail on this case of a `_trie_lookup(...)` walking a `HashedNode` which clearly means the witness is incorrect.

src/ethereum/forks/amsterdam/witness_state.py
```diff
@@ -46,96 +50,56 @@ def build_code_db(code_entries: Tuple[Bytes, ...]) -> Dict[Hash32, Bytes]:
...
-            node = _resolve_node_ref(node_db, node[idx])
-        else:
-            return None
+            pos += len(segment)
+            node = node.child
+            continue
+
+        assert isinstance(node, MutableBranchNode), (
+            f"Unexpected node type {type(node)}"
+        )
```
jsign, line 92: We already checked for cases `HashedNode`, `MutableExtensionNode` and `MutableLeafNode` so there isn't other option -- still, let's be safe and add this assertion here.

src/ethereum/forks/amsterdam/witness_state.py
```diff
@@ -46,96 +50,56 @@ def build_code_db(code_entries: Tuple[Bytes, ...]) -> Dict[Hash32, Bytes]:
...
+
+        assert pos <= len(nibbles), (
+            "BranchNode cannot appear past the end of a key "
+            "in state/storage trie"
+        )
+        if pos == len(nibbles):
+            return node.value or None
```
jsign, line 95: Can't happen in practice in Ethereum, but still valid case.

src/ethereum/forks/amsterdam/witness_state.py
```diff
@@ -180,6 +140,23 @@ class WitnessState:
     _state_root: Root
     _code_db: Dict[Hash32, Bytes]
     _storage_root_cache: Dict[Address, Root] = field(default_factory=dict)
+    _decoded_secure_roots: Dict[Root, MutableNode] = field(
+        default_factory=dict
+    )
+
+    def _get_decoded_secure_root(self, root_hash: Root) -> MutableNode:
```
jsign, line 147: This is basically a new method that we just below in L168 and L191.

Whenever we want to access a trie (account or storage(s)), we deserialize the whole MPT trie and cache it.

This MPT deserialization has many MPT correctness checks (plus the new ones mentioned in my comment in file `src/ethereum/forks/amsterdam/incremental_mpt.py` at the start of this PR (i.e. extension nodes can't have empty `key`)

Just think about concentrating all MPT-structural-correctness checks in a single place.

src/ethereum/forks/amsterdam/witness_state.py
```diff
@@ -188,7 +165,9 @@ def get_account_optional(self, address: Address) -> Optional[Account]:
         Return ``None`` if there is no account at the address.
         """
         key_hash = keccak256(address)
-        leaf = _trie_lookup(self._node_db, self._state_root, key_hash)
+        leaf = _trie_lookup(
```
jsign, line 168: Now that `_trie_lookup` uses the cached-deserialized-MPT, on second/third/etc accesses we won't be re-RLP-deserializing as we originally did when using the (now removed in this file) `_resolve_node_ref(...)`.

tests/amsterdam/eip8025_optional_proofs/state_helpers.py
```diff
@@ -102,3 +116,101 @@ def collect_storage_post_state_only_nodes(
         pre_storage, pre_state_reference_slots
     )
     return _nodes(post_state_nodes - pre_state_nodes)
+
+
+def find_account_with_shared_secured_nibble(
```
jsign, line 121: For many invalid `state` tests, we usually want to remove a node that is required.

This helper simply helps to find a correct account address that forces a branch node with two leaves, so we can specifically remove one we need.

Instead of fixing magical addresses that create branch nodes, this helper makes this more understandable and maintainable by simply finding an address that creates this situation.

tests/amsterdam/eip8025_optional_proofs/state_helpers.py
```diff
@@ -102,3 +116,101 @@ def collect_storage_post_state_only_nodes(
...
+    target: Address,
+    excluded: set[Address],
+) -> Address:
+    """Return an occupied address that deepens one account absence proof."""
+    target_nibble = keccak256(bytes(target))[0] >> 4
+    for value in range(0x100, 0x1000):
+        address = Address(value)
+        if address in excluded:
+            continue
+        if (keccak256(bytes(address))[0] >> 4) == target_nibble:
```
jsign, line 131: We found one that shares the same first nibble, so we're done. 
This means the branch node creation will happen which is what he caller wants.

tests/amsterdam/eip8025_optional_proofs/state_helpers.py
```diff
@@ -102,3 +116,103 @@ def collect_storage_post_state_only_nodes(
...
+
+def collect_account_proof_nodes(
+    alloc: Alloc,
+    addresses: Sequence[StateAddress | bytes],
+) -> list[Bytes]:
+    """Collect account-trie proof nodes for the given addresses."""
+    return _nodes(_collect_account_proof_node_rlps(alloc, addresses))
+
+
+def collect_account_path_only_nodes(
```
jsign, line 208: The above are mainly sub-helpers for this "public helper" for test writers.

Basically allows the test writer to know which MPT proofs for an address are non-overlapping with the MPT proof nodes of other set of addresses. Said differently, which MPT nodes are unique for a particular address.

This is useful to let the test writer only remove from `ExecutionWitness.state` very specific nodes to trigger an invalid witness validation for a particular account branch (e.g. absent address, or similar cases).

src/ethereum/forks/amsterdam/incremental_mpt.py
```diff
@@ -950,6 +950,7 @@ def _decode_witness_node(
                 _rlp=rlp_bytes,
             )
         else:
+            assert len(nibbles) > 0, "ExtensionNode must have a non-empty path"
```
jsign, line 953: A bit tighter check here. This definitely must never happen.

tests/amsterdam/eip8025_optional_proofs/test_witness_state_reads.py
```diff
@@ -94,6 +96,100 @@ def test_witness_state_reverted_sload_still_contains_storage_proof(
     )
 
 
+def test_witness_state_reverted_inner_sload_still_contains_storage_proof(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+) -> None:
+    """
+    A reverted inner-call SLOAD should still leave its proof nodes in witness.
```
jsign, line 104: Just an extra witness completeness check.

Still pending to do a more throghout pass on these revert/stack-overflow/oog cases, but doesn't hurt to keep adding as being identified doing other things.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
jsign, line 1: Apart from spec improvements & fixes, these are the bulk of test cases intention of the PR.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -26,6 +40,29 @@
...
+        if not isinstance(decoded, list) or len(decoded) != 2:
+            continue
+        path_bytes = decoded[0]
+        assert isinstance(path_bytes, (bytes, bytearray))
+        _, is_leaf = compact_to_nibbles(TrieBytes(path_bytes))
+        if is_leaf:
+            leaves.append(node)
+
+    assert len(leaves) == 1
+    return leaves[0]
```
jsign, line 63: Given some MPT proof, these are two main helpers used to pick a paritcular MPT node to be removed.

Depending on the case, we can pick any (i.e. first node) or be more specific and pick hte leaf node.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -26,6 +40,29 @@
...
+
+
 def test_validation_state_missing_storage_proof_node(
```
jsign, line 66: This was just the simple test of removing a node and seeing a failure.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,371 @@ def test_validation_state_missing_storage_proof_node(
             contract: Account(storage=post_storage),
         },
     )
+
+
+def test_validation_state_missing_absent_slot_proof_leaf_node(
```
jsign, line 105: This is a bit nuanced case of a missing node that justifies a proof of absence failure.

I hope the comment in the test makes clear what are we trying to do here.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,371 @@ def test_validation_state_missing_storage_proof_node(
...
+    #   absent slot 1 follows this edge, then diverges inside the leaf
+    #
+    # The proof for slot 1 is therefore `extension -> branch -> leaf`.
+    # Removing that leaf makes the absence proof invalid due to missing
+    # data.
+    pre_storage = build_large_storage([14, 24])
+    insert_slot = 1
+    insert_value = large_storage_value(insert_slot)
+    proof_nodes = collect_storage_proof_nodes(pre_storage, [insert_slot])
+    assert len(proof_nodes) == 3
```
jsign, line 137: Just be sure what I explained in the comment hold

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,371 @@ def test_validation_state_missing_storage_proof_node(
...
+            ),
+        },
+    )
+
+
+def test_validation_state_missing_delete_auxiliary_node(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+) -> None:
+    """Removing the delete-collapse auxiliary node should fail."""
```
jsign, line 173: The other main interesting case to cover for our interests.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,371 @@ def test_validation_state_missing_storage_proof_node(
...
+    storage = build_large_storage([1, 2])
+    proof_nodes = collect_storage_proof_nodes(storage, [1])
+    auxiliary_nodes = collect_storage_delete_auxiliary_nodes(storage, 1)
```
jsign, line 176: Worth checking how `collect_storage_delete_auxiliary_nodes(...)` work.

Basically, it takes the `storage` trie and a storage slot target (`1`) and:
1. Does an `mpt_get` of `1` to get the pre-state MPT proof.
2. Does the deletion of `1` and gets the MPT proof again.

It takes the difference of 2. minus 1., which should basically return the sibiling node (since hte branch was compressed).

Now that we have it, we know which node to remove from the witness for this test.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,371 @@ def test_validation_state_missing_storage_proof_node(
...
+    assert len(auxiliary_nodes) == 1
```
jsign, line 177: Just check we found the sibiling.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,371 @@ def test_validation_state_missing_storage_proof_node(
...
+    # the absence proof extends below the shared root node.
+    sibling = find_account_with_shared_secured_nibble(
+        recipient,
+        {
+            sender,
+            recipient,
+            *Amsterdam.execution_witness_implicit_code_addresses(),
+        },
+    )
+    pre.fund_address(sibling, 1)
```
jsign, line 263: The idea of this test is to also have an invalid witness nodes for proving an absent account(i.e. a non-existent recipient of an eth transfer).

What we do here is leverage the `find_account_with_shared_secured_nibble(...)` to find an address `sibiling` that will share the first nibble with `recipient`.

This mean that the MPT proof (of absence) of `recipient` will have this `sibiling` leaf as the terminal node that will prove the absence. 

This means we need to remove `sibiling` in the witness to generate this "insufficient information to prove `recipient` absence" situation.

tests/json_infra/test_incremental_mpt.py
```diff
@@ -219,6 +236,124 @@ def test_single_entry(self) -> None:
...
+            _decode_root_from_rlp(Bytes(rlp.encode(b"\x01")))
+
+    def test_invalid_rlp_node_length(self) -> None:
+        """Only 2-item and 17-item node lists are valid."""
+        with pytest.raises(AssertionError, match="Invalid RLP node length: 3"):
+            _decode_root_from_rlp(
+                Bytes(rlp.encode([b"\x01", b"\x02", b"\x03"]))
+            )
+
+    def test_extension_with_empty_child_ref(self) -> None:
```
jsign, line 260: This and other unit test below is what I mentioned in hte PR description that might be good to do as "normal tests" eventually. (e.g. extension node with empty child, branch node with 1 or 0 children, etc)

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,377 @@ def test_validation_state_missing_storage_proof_node(
...
+            contract: Account(storage={2: storage[2]}),
+        },
+    )
+
+
+def test_validation_state_missing_sender_account_proof_leaf_node(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+) -> None:
+    """Removing the sender account proof leaf should fail a transfer."""
```
jsign, line 211: An easy-ish test for a simple case required mpt node removal in an eth transfer.

This is similar to other test above but for the account trie instead of storage trie. I'm sometimes doubting if it is worth targeting these same scenarios for each kind of trie since both are MPT tries anyway and they should behave the same.

But we never know if an implemnetation have the same witness check rules for account and storage tries. While I can't imagine why anybody would code an EL to make a distinction, I guess the safest thing is to include the test always for both cases.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,377 @@ def test_validation_state_missing_storage_proof_node(
...
+            recipient: Account(balance=1),
+        },
+    )
+
+
+def test_validation_state_missing_failed_call_target_account_proof_node(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+) -> None:
+    """Removing the failed CALL target proof should fail witness replay."""
```
jsign, line 302: Just a corner case of a CALL that fails because of insufficient balance. 

The way the spec works now is that the target account is accessed even if the CALL-fails, thus if we remove a required mpt node for that target the stateless execution should also signal al invalid witness (despite the CALL is never made).

A bit of a tongue twister, but nothing weird here. Just checking nobody thought was a good idea to do witness checks only if the CALL didn't OOG or similar. Hard to imagine, but we never know.

tests/amsterdam/eip8025_optional_proofs/test_witness_validation_state.py
```diff
@@ -64,3 +100,377 @@ def test_validation_state_missing_storage_proof_node(
...
+            )
+        ],
+        post={
+            sender: Account(nonce=1),
+            contract: Account(storage=storage),
+        },
+    )
+
+
+def test_validation_state_extra_unused_trie_node(
```
jsign, line 354: This and next test are cases related to https://github.com/ethereum/execution-specs/issues/2474 (note compared to all the rest of test in this file, these two have `expected_stateless_validation_success=True`).

jsign:
@kevaundray, this is ready for some eyes. I think it might be one of the last big-ish PRs, although I added many comments to help ease the review.

I know you're at the event this week, so not rush and take a look whenever time allows.

src/ethereum/forks/amsterdam/witness_state.py
kevaundray: TODO: check this python syntax `node.value or none`

kevaundray (approved):

PR #2529, opened 2026-03-19
chore(ci): enable ported static in py3 and split into 5 parallel fork-range runners

danceratopz:
## 🗒️ Description

The `py3` CI job would run ~419 minutes of `fill` test time (if ran sequentially with `./tests/ported_static`) on a single runner. This PR splits it into 5 parallel matrix runners grouped by fork duration, using the existing `--from`/`--until` CLI flags:

| Group | Forks | Est. duration |
|-------|-------|---------------|
| `pre-cancun` | Frontier → Shanghai | ~61m |
| `cancun` | Cancun | ~72m |
| `prague` | Prague | ~104m |
| `osaka` | Osaka | ~86m |
| `amsterdam` | Amsterdam | ~97m |

Durations were derived from `.test_durations_py3` (this was generated with the help of the pytest-switch plugin which I was experimenting with cf #2528, it's not required for this PR). The pre-Cancun forks (Frontier through Shanghai) total ~61m combined — light enough for a single runner. Each post-Shanghai fork needs its own runner due to the volume of tests and transition fork coverage.

Prague is the bottleneck at ~104m despite having fewer tests than Amsterdam (39k vs 44k), because ported static tests marked `valid_until("Prague")` — particularly `test_create_oo_gafter_max_codesize` (~14.3m across 4 variants) — don't run in later forks.

Coverage reports are auto-merged by Codecov via the shared `unittests` flag. Running `tox -e py3` locally without posargs still fills all forks (existing behavior unchanged).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

- [x] All: Ran fast `tox` checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    uvx tox -e static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).

#### Cute Animal Picture
image

PR #2592, opened 2026-03-30
feat(ci): split release builds across multiple runners; delete develop; stable becomes mainnet

danceratopz:
## 🗒️ Description

Split multi-fork fixture releases across parallel runners by fork range, then simplify the release strategy by collapsing the two release workflows into one.

### Summary of changes

- Multi-fork releases (e.g., `mainnet`) are split across parallel self-hosted runners by fork range (pre-cancun, cancun, prague, osaka, ...), then combined into a single tarball.
- The `develop` feature is removed; `stable` is renamed to `mainnet`.
- The full release workflow (`release_fixture_full.yaml`) is deleted. All releases now use `release_fixture_feature.yaml` with `tests-<feature>@v*` tags.
- Mainnet releases are marked as regular releases (not pre-releases). Other features (bal, benchmark) remain pre-releases.

### How we got here

The initial goal was to reduce release CI wall-clock time by splitting expensive fill jobs across parallel runners, following the pattern from #2529.

While implementing the split, we noticed that `stable` and `develop` would share identical fork-range builds (same EELS implementation, same fill-params minus `--until`). The deduplication logic to handle this was fragile: it silently assumed both evm-types produce the same output, and would break if they ever diverged.

This led to a simpler realization: the `develop` release is redundant. The `bal` feature already fills Amsterdam for devnets, and the full release workflow exists only to bundle `stable` + `develop`. Removing `develop` eliminates the cross-feature deduplication problem entirely and lets us collapse the full and feature release workflows into one. A mainnet release is just another feature release (`tests-mainnet@v6.0.0`).

### Workflow structure

Before:

```
features -> build (matrix: feature) -> release
```

After:

```
setup -> build (matrix: fork-range) -> combine -> release
```

The `setup` job reads `feature.yaml` and emits a flat build matrix of fork ranges. Each runner fills its range to a directory. The `combine` job merges the directories and creates the final tarball. Features without fork ranges (e.g., `benchmark`) skip the combine step and produce the tarball directly.

### Config changes

- `feature.yaml`: top-level `fork-ranges` list defines how multi-fork features are split. Features with `--until` in their `fill-params` are automatically split across applicable ranges.
- `evm.yaml`: `stable`/`develop` entries replaced with single `eels` entry.
- New `feature_only: true` flag excludes features (benchmark, bal) from `--all` mode (now removed, but the flag remains useful as documentation).

### New scripts

- `generate_build_matrix.py`: reads `feature.yaml`, emits build matrix and combine metadata for GitHub Actions.
- `create_release_tarball.py`: creates a release tarball from a merged fixture directory, using `pigz` for parallel compression when available.
- Both scripts have integration tests in `.github/scripts/tests/` (runnable via `just test-ci-scripts`).

## PRs and Actions

- #2529: same pattern applied to `py3` tox env.
- Successful `tests-mainnet@v6.0.0` build here: https://github.com/ethereum/execution-specs/actions/runs/23894376018

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

image


danceratopz:
Ah @spencer-tb, this is probably missing an `index.json`!!!

.github/configs/feature.yaml
```diff
@@ -1,22 +1,43 @@
 # Unless filling for special features, all features should fill for previous forks (starting from Frontier) too
-stable:
-  evm-type: stable
-  fill-params: --until=Prague --fill-static-tests --ignore=tests/static/state_tests/stQuadraticComplexityTest
+#
+# Shared fork ranges for splitting multi-fork releases across parallel runners.
+# Features using --until are automatically split using applicable ranges.
+# Features using --fork (single fork) are never split.
+fork-ranges:
```
spencer-tb: I think this should be in its own yaml file, as its not a feature: `configs/forks.yaml`?

.github/configs/feature.yaml
```diff
@@ -1,22 +1,43 @@
...
+  - label: amsterdam
+    from: Amsterdam
+    until: Amsterdam
 
-develop:
-  evm-type: develop
-  fill-params: --until=BPO4 --fill-static-tests --ignore=tests/static/state_tests/stQuadraticComplexityTest
+mainnet:
+  evm-type: eels
+  fill-params: --until=BPO2 --fill-static-tests --ignore=tests/static/state_tests/stQuadraticComplexityTest
```
spencer-tb:
```suggestion
  fill-params: --until=BPO2
```
Can remove this now!

.github/configs/feature.yaml
```diff
@@ -1,22 +1,43 @@
...
 
 benchmark_fast:
   evm-type: benchmark
   fill-params: --fork=Osaka --gas-benchmark-values 100 ./tests/benchmark/compute
   feature_only: true
 
 bal:
-  evm-type: develop
+  evm-type: eels
   fill-params: --fork=Amsterdam --fill-static-tests
```
spencer-tb:
```suggestion
  fill-params: --fork=Amsterdam
```
Same here!

.github/workflows/release_fixture_feature.yaml
```diff
@@ -61,11 +64,67 @@ jobs:
...
+          submodules: false
+      - uses: ./.github/actions/setup-uv
+      - name: Install pigz
+        run: sudo apt-get install -y pigz
+      - name: Download and merge split artifacts
+        shell: bash
+        run: |
+          mkdir -p combined
+          for label in ${{ needs.setup.outputs.combine_labels }}; do
+            echo "Downloading: fixtures__${label}"
```
spencer-tb, line 88: __ intentional or just _, its in multiple spots
```suggestion
            echo "Downloading: fixtures_${label}"
```

packages/testing/src/execution_testing/fixtures/consume.py
```diff
@@ -91,6 +91,36 @@ class IndexFile(BaseModel):
...
+            all_forks.update(idx.forks or [])
+            all_formats.update(idx.fixture_formats or [])
+
+        root_hash = HashableItem.from_index_entries(all_cases).hash()
+
+        return cls(
+            root_hash=HexNumber(root_hash),
+            created_at=datetime.datetime.now(datetime.timezone.utc),
+            test_count=len(all_cases),
+            forks=sorted(all_forks, key=lambda f: f.name()),
```
spencer-tb, line 119: Should we make this by fork order, just a nit

spencer-tb (commented):
Some small comments!

.github/workflows/release_fixture_feature.yaml
danceratopz: It was intentional, but maybe no longer required. Let me check.

packages/testing/src/execution_testing/fixtures/consume.py
danceratopz: I'm not sure it matters.

.github/scripts/generate_build_matrix.py
```diff
@@ -0,0 +1,226 @@
...
+    for r in fork_ranges:
+        if FORK_INDEX[r["from"]] <= limit:
+            entry = dict(r)
+            if FORK_INDEX[r["until"]] > limit:
+                entry["until"] = until_fork
+            result.append(entry)
+    return result
+
+
+def build_matrices(
```
spencer-tb: From claude:

> Now that the full release workflow is deleted and main() always passes a single feature, the multi-feature deduplication logic (the groups dict / effective_key grouping) is unreachable. Could this be simplified to just handle one feature?

packages/testing/src/execution_testing/fixtures/consume.py
danceratopz: Ideally, we'd have the same result as without a merge.

.github/configs/feature.yaml
danceratopz: Went with `configs/fork-ranges.yaml`.

.github/scripts/generate_build_matrix.py
danceratopz: Good catch Claudine!

danceratopz: Done.

.github/workflows/release_fixture_feature.yaml
danceratopz: Still required, tentative to change this now in a rush.

danceratopz:
Applied most suggestions. Tagged a new full build to check nothing broke. Was in a bit of a hurry.
https://github.com/ethereum/execution-specs/actions/runs/23908626680

spencer-tb (approved):
LGTM!

spencer-tb:
Going to merge. Anything else can be a follow up! :D

PR #2593, opened 2026-03-30
feat(cli): Add --until option for checklist generation

kclowes:
## 🗒️ Description
The checklist CLI command wraps `fill`, and so checklist generation wasn't picking up EIPs in a future fork. Added the `--until` option.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
@@ -22,7 +22,16 @@ from .fill import FillCommand
     multiple=True,
     help="Generate checklist only for specific EIP(s)",
 )
-def checklist(output: str, eip: tuple[int, ...], **kwargs: Any) -> None:
+@click.option(
+    "--until",
+    "-u",
+    type=str,
+    default=None,
+    help="Include upcoming forks up to and including this fork",
+)
+def checklist(
+    output: str, eip: tuple[int, ...], until: str | None, **kwargs: Any
+) -> None:
     """
     Generate EIP test checklists based on pytest.mark.eip_checklist markers.
 
@@ -39,6 +48,9 @@ def checklist(output: str, eip: tuple[int, ...], **kwargs: Any) -> None:
         # Generate checklists for specific test path
         uv run checklist tests/prague/eip7702*
 
+        # Include upcoming forks
+        uv run checklist --eip 8037 --until Amsterdam
+
         # Specify output directory
         uv run checklist --output ./my-checklists
 
@@ -51,6 +63,10 @@ def checklist(output: str, eip: tuple[int, ...], **kwargs: Any) -> None:
     # Add EIP filter if specified
     for eip_num in eip:
         args.extend(["--checklist-eip", str(eip_num)])
+
+    # Add --until fork if specified
+    if until:
+        args.extend(["--until", until])
 
     command = FillCommand(
         plugins=[
--- a/packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
@@ -4,7 +4,14 @@ from typing import Any
 
 import click
 
+from ...forks import get_development_forks
 from .fill import FillCommand
+
+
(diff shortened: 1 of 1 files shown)


spencer-tb (commented):
Thanks! I'm wondering if there is a nice way we can make this the default, instead of passing a flag. It should ideally resolve for all forks imo without requiring any user input.

Quick dump below, but I think there is a better solution.
```python
from execution_testing.forks import get_forks

last_fork = get_forks()[-1]                 
args.extend(["--until", last_fork.name()])
```
Not super precious here though!

kclowes:
That makes sense to me, will update!

spencer-tb (approved):
LGTM! Thanks :D

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
```diff
@@ -4,9 +4,16 @@
 
 import click
 
+from ...forks import get_development_forks
 from .fill import FillCommand
 
 
+def _last_development_fork() -> str | None:
+    """Return the name of the last development fork, if any."""
+    dev_forks = get_development_forks()
```
danceratopz, line 13: Interesting, perhaps we could make `get_development_forks` return a single fork now. In EEST, we had multiple development forks (and this could be unreliable), but in execution-specs, we should _probably_ only have one dev fork in the default branch at a time? 

If so, this is ok (but then we should probably change the underlying helper), if not, then there's not guarantee this will return the fork you want.

danceratopz (commented):
Haha rugpulled during review. Thanks for the merge, @spencer-tb!

> Thanks! I'm wondering if there is a nice way we can make this the default, instead of passing a flag. It should ideally resolve for all forks imo without requiring any user input.

We could lean on `just` here to guarantee we get checklists for the dev fork we're interested in (see comment below).

```just
# Generate EIP checklists for consensus tests
[group('consensus tests')]
checklist *args:
    uv run checklist --until "{{ latest_fork }}"
```

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
kclowes: hmm, I suspect that we will have a Bogota branch before we close out the Amsterdam branch. But I think this is okay because we really just want whatever the latest branch is. Let me know if I'm misunderstanding though!

danceratopz: As the the active fork under development should be set per branch in execution-specs, I think we'll be fine! But it does point to a small code smell in our fork helper functions - I don't think `get_development_forks()` should exist anymore at all. I _think_ that we only ever need one dev fork per execution-specs branch because that's how we structure fork development in execution-specs. In that case, I think adding a global helper somewhere that returns a single fork as a string would be really helpful. This was added to the Justfile as `latest_fork` (it was defined multiple times in each fill command previously), but we should make this available everywhere.

danceratopz: Apparently,  `get_development_forks()` currently returns 4 forks: BPO3, BPO4, BPO5, Amsterdam.

fselmo: > I _think_ that we only ever need one dev fork per execution-specs branch because that's how we structure fork development in execution-specs.

When we were recreating `amsterdam` from `osaka` while working on that, the `forks/amsterdam` branch would've had two dev forks - `osaka` and `amsterdam`, no?

kclowes: > When we were recreating amsterdam from osaka while working on that, the forks/amsterdam branch would've had two dev forks - osaka and amsterdam, no?

Yeah, I think there is a good chance that we'll have two development forks at a time, but I think what @danceratopz is saying is that on each branch, the latest fork that gets populated will correspond to the fork branch you're on, even if there is a more "recent" development fork on a different branch. This is just the default though, you can always pass along any cli args you want. 

> I think adding a global helper somewhere that returns a single fork as a string would be really helpful.

Can do!

PR #2596, opened 2026-03-30
fix(specs): remove None from trie Node type

### LINKED ISSUE #1044: trie.Node type is not compatible with rlp.Extended though used in encode_node
### Metadata

* **Hardfork**: cancun

### What was wrong?

The `encode_node` function signature is `def encode_node(node: Node, storage_root: Optional[Bytes] = None) -> Bytes:` with

```
Node = Union[
    Account, Bytes, LegacyTransaction, Receipt, Uint, U256, Withdrawal, None
]
```

The `None` in the `Node` definition makes it incompatible with the `rlp.Extended` type used in the body in `rlp.encode(node)`.

Either `Extended` should have `Optional` or `Node` not have `None`.

See https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/trie.py#L154-L168

SamWilsn:
Closes: #1044

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/trie.py
+++ b/src/ethereum/forks/amsterdam/trie.py
@@ -70,16 +70,7 @@ EMPTY_TRIE_ROOT = Root(
     )
 )
 
-Node = (
-    Account
-    | Bytes
-    | LegacyTransaction
-    | Receipt
-    | Uint
-    | U256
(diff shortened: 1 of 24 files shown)


fselmo (approved):
lgtm 👍🏼

PR #2598, opened 2026-03-30
feat(benchmarks): add runtime delegated EIP-7702 benchmarks

jochem-brouwer:
## 🗒️ Description
This adds benchmarks plus a template for future use. On perf-devnet-3, there are (so far) three EOAs being targeted with their bytecode being delegated to: 

`0x5f541515600b5760015f555b5f54805b818155600101906001019061ffff5a11600f575f55`

This (22866e06568ba74ddbdd7dec3811de0a5d0d5948):

> This is example code to delegate to (runtime code):
    0x5f541515600b5760015f555b5f54805b818155600101906001019061ffff5a11600f575f55
    This works even on empty accounts: if the first slot is 0, set it to 1,
    then: initialize stack to SLOAD(0) and proceed to write stack key as value
    to that slot. It will keep looping until the gasleft is lower than or equal
    to 0xffff (65535), i.e. currently enough to write to an empty slot (about
    22100 gas as of Fusaka)
    The cleanup phase at the end of the loop writes next key to slot 0.
    Note that this key is either already dirty (initial call on empty storage)
    or is already set. So the max gas cost of
    cleanup is 5000 + 2100 as of Fusaka.
    
We now use the fact that we have pkey of these accounts to overwrite runtime bytecode via EIP-7702 to target these for our needs. The bloatnet/perf-devnet-3 accounts are thus bloated with a range from 1-N (1/10/20Gb currently) which we can now target for storage related benchmarks.
So far this PR accounts for the NO-CACHE variant with existing/non-existing keys in the case we run the setup bytecode as provided and then start the attack block.

To add:

- The SLOAD variant which does not do SSTORE, only SLOAD
- The CACHE_TX variant which does SLOAD before either SSTORE/SLOAD of the target
- The CACHE_PREVIOUS_BLOCK variant which does **ONLY** SLOADs before the attack block. This also resets the start slot to the same key it started before doing the cache initialization.

Draft, but want to finish this ASAP as we are super interested in the benchmark results.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -35,6 +35,7 @@ from execution_testing import (
     keccak256,
 )
 from execution_testing.base_types.base_types import Number
+from execution_testing.rpc import EthRPC
 
 from tests.benchmark.stateful.helpers import (
     ALLOWANCE_SELECTOR,
@@ -42,8 +43,10 @@ from tests.benchmark.stateful.helpers import (
     BALANCEOF_SELECTOR,
     DECREMENT_COUNTER_CONDITION,
     MINT_SELECTOR,
+    STORAGE_BLOATED_EOAS,
     CacheStrategy,
     build_cache_strategy_blocks,
+    get_storage_bloated_eoa,
 )
 
 REFERENCE_SPEC_GIT_PATH = "DUMMY/bloatnet.md"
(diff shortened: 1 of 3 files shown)


misilva73:
I am wondering if we could merge this with only the NO_CACHE variant for SLOAD and SSTORE to see results faster. What do you think, @jochem-brouwer ? Will this save us significant time?

jochem-brouwer:
No cache variants added, open for review! I checked locally against a client and it *looks* fine.

misilva73 (commented):
After a first-level look at the code, I cannot find any issues. However, I do have two comments:

1. A big part of the code in `test_sload_bloated` and `test_sstore_bloated` is duplicated (Same setup, same gas-loop to build transactions and same benchmark_test() call). I wonder is logic should be moved to a helper function to help with code maintainability. This is a nitpick and should definitely not block merging this.
2. The EIP-7702 delegation trick is quire clever 👏 However, I worry it could lead to different measurements. From a specification perspective, it shouldn't - every SLOAD/SSTORE should execute against the EOA's storage trie exactly the same way it would for a normal contract. The trie structure and lookup path should be identical regardless of how the code got there. However, individual client implementations could have different code paths or caching behavior for delegated accounts vs. native contracts. Can we double-check with clients that this is not the case?

jochem-brouwer:
@misilva73, (2) this is a good point and also a good exercise, I will do this today. I would be surprised if the code paths diverge here, I assume that most clients have simply swapped out the runtime code for a EIP-7702 delegated account and keep the account code (the 0xef0100..) (where EXTCODECOPY refers to). This is a good sanity check. I will check with the 5 clients to see if there is a mechanism which might cause problems for this approach.

Regarding the duplication (1): you are right, I should refactor this.

jochem-brouwer:
Converted to draft, figuring out what goes wrong here.

I have a local instance of Nethermind running to test, this is the error:

`Encountered exception Nethermind.Blockchain.InvalidTransactionException: Transaction 0xe0e9193fca0011926f1ae82d53b70698ee28c9fbc5cc57cae035dec3a776c39a at index 0 failed with error transaction nonce is too high`

misilva73:
> Converted to draft, figuring out what goes wrong here.
> 
> I have a local instance of Nethermind running to test, this is the error:
> 
> `Encountered exception Nethermind.Blockchain.InvalidTransactionException: Transaction 0xe0e9193fca0011926f1ae82d53b70698ee28c9fbc5cc57cae035dec3a776c39a at index 0 failed with error transaction nonce is too high`

When I was exploring this PR with Claude code, it noticed the following:

>_delegate_and_set_slot0 mutates authority.nonce as a side effect: The function increments authority.nonce in place. This is a hidden side effect that callers must be aware of. The two tests each call it twice in sequence, relying on this mutation. It works, but it's fragile — if someone reorders the calls or adds a third, the nonce tracking could silently break.

This didn't seem to be an issue atm, but I wonder if the transaction nonce error could have anything to do with this?

LouisTsai-Csie:
> This didn't seem to be an issue atm, but I wonder if the transaction nonce error could have anything to do with this?

@misilva73 the nonce which is manually updated could be an issue, but since we send the transaction sequentially and use the testing endpoint to build block, i personally think it might not have much impact, but i might be wrong here.

I start a local anvil node and try to run the benchmark, it was successful, but since i am not using snapshot env, there might be some slight difference. I've contacted Kamil, and we will check with him about whether there is something to do with gas-benchmarks.

Update: it seems working on the latest gas-benchmarks run ([link](https://github.com/NethermindEth/gas-benchmarks/actions/runs/23939781606/job/69823411930))

LouisTsai-Csie:
I've checked the CI run from Kamil (Please see the comment above), and prefer to merge now, so we could run the payload generation on weekend. I will track the unresolved issue (client execution path) on the board so we do not forget it.

LouisTsai-Csie (approved):
LGTM, thank you for the new tests

PR #2600, opened 2026-03-31
feat(test-benchmark): add uncacheable precompile benchmark

LouisTsai-Csie:
## 🗒️ Description

Add uncacheable benchmark variants for every precompile. Each test guarantees every single `STATICCALL` across all transactions receives unique input, preventing clients from caching precompile results.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture


tests/benchmark/compute/precompile/test_p256verify.py
```diff
@@ -88,3 +93,126 @@ def test_p256verify(
...
+        per_tx_gas = min(tx_gas_limit, remaining_gas)
+        h = int.from_bytes(calldata[:32], "big") + seed
+
+        tx_calldata = Bytes(
+            h.to_bytes(32, "big")  # hash
+            + calldata[32:]  # r, s, qx, qy
+        )
+
+        intrinsic = intrinsic_gas_calculator(
+            calldata=calldata,
```
spencer-tb: I think a small bug?
```suggestion
            calldata=tx_calldata,
```

tests/benchmark/compute/precompile/test_point_evaluation.py
```diff
@@ -57,3 +62,137 @@ def test_point_evaluation(
...
+    if precompile_address not in fork.precompiles():
+        pytest.skip("Precompile not enabled")
+
+    gsc = fork.gas_costs()
+    intrinsic_gas_calculator = fork.transaction_intrinsic_cost_calculator()
+
+    # versioned hash: data[0:32] 32 bytes
+    # x: data[32:64] 32 bytes
+    # y: data[64:96] 32 bytes
+    # commitement: data[96:144] 48 bytes
```
spencer-tb:
```suggestion
    # commitment: data[96:144] 48 bytes
```

spencer-tb (commented):
Thanks @LouisTsai-Csie!

Just wondering why we didn't add uncacheable for `test_ecrecover.py`? Is it already that it is uncacheable?

For `test_sha256_uncachable`, `test_ripemd160_uncachable` and `test_identity_uncachable` when the size is 32 bytes they are uncacheable. But when we get to 256 and 1024, it looks like most of the precompile inputs become identical, i.e the first 32 bytes are overwritten. I'm not sure if thats an issues but if we can find a way to make each input truely unique and deterministic I feel that would be a good idea!

tests/benchmark/compute/precompile/test_point_evaluation.py
```diff
@@ -57,3 +62,137 @@ def test_point_evaluation(
...
+    # proof: data[144:192] 48 bytes
+    attack_block = Op.MSTORE(
```
misilva73: I am not sure this will work. By changing y, the commitment will no longer match and thus the STATICCALL will return 0. Specifically:

- First call: correct y, valid proof → STATICCALL returns 1 → y += 1
- Second call: y+1 no longer matches commitment/proof → precompile reverts → STATICCALL returns 0 → y stays y+1
- All subsequent calls: identical y+1, identical failure → cacheable

One option is to always increment by the output if the STATICCALL plus a static +1. That way, even if the precompile reverts, we ensure a different input. Another option is fill enough valid inputs as calldata (as you did in `test_bls12_pairing_uncachable`)

tests/benchmark/compute/precompile/test_bls12_381.py
```diff
@@ -246,3 +255,335 @@ def test_bls12_pairing(
...
+            gas=Op.GAS,
+            address=precompile_address,
+            args_size=Op.CALLDATASIZE,
+            ret_size=ret_size,
+            # gas accounting
+            address_warm=True,
+        ),
+    )
+
+    attack_block_cost = attack_block.gas_cost(fork)
```
misilva73: Should this also have the `precompile_cost`? I noticed other tests do this here.

tests/benchmark/compute/precompile/test_modexp.py
```diff
@@ -415,3 +423,127 @@ def test_modexp(
...
+            gas=Op.GAS,
+            address=0x05,
+            args_size=Op.CALLDATASIZE,
+            ret_offset=96,
+            ret_size=base_length,
+            # gas accounting
+            address_warm=True,
+        ),
+    )
+    attack_block_gas = attack_block.gas_cost(fork)
```
misilva73: Should this also have the precompile_cost? I noticed other tests do this here.

misilva73 (commented):
Looks good in general. Added a few questions

LouisTsai-Csie:
@spencer-tb yes, the `test_ecrecover` one is already uncacheable.

```python
attack_block = Op.MSTORE(
    0,
    Op.ADD(
        Op.MLOAD(0),
        Op.STATICCALL(
            gas=Op.GAS,
            address=precompile_address,
            args_size=Op.CALLDATASIZE,
            ret_offset=0x80,
            ret_size=0x20,
        ),
    ),
)
```

From the `attack_block`, you could see the memory content is changing for every iteration.

tests/benchmark/compute/precompile/test_bls12_381.py
LouisTsai-Csie: Done

tests/benchmark/compute/precompile/test_modexp.py
LouisTsai-Csie: Done!

tests/benchmark/compute/precompile/test_blake2f.py
```diff
@@ -85,3 +92,115 @@ def test_blake2f_benchmark(
...
+    code = setup + loop
+    attack_contract_address = pre.deploy_contract(code=code)
+
+    base_calldata = Blake2bInput(
+        rounds=num_rounds, f=True
+    ).create_blake2b_tx_data()
+
+    txs: list[Transaction] = []
+    remaining_gas = gas_benchmark_value
+    rng = random.Random(42)
```
marioevz, line 172: What was the intention behind this `random` usage? Currently it's simply working as a hash because it's always going to return the same random-looking bytes.

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -345,14 +345,9 @@ def model_post_init(self, __context: Any, /) -> None:
...
-            self.fixed_opcode_count is not None
-            and self.code_generator is None
-            and self.target_opcode is None
-        ):
+        if self.fixed_opcode_count is not None and self.code_generator is None:
             pytest.skip(
-                "Cannot run fixed opcode count tests without a "
-                "code generator or a target opcode set"
+                "Cannot run fixed opcode count tests without a code generator"
             )
```
LouisTsai-Csie, line 374: The CI ran for over 365 minutes due to a filtering logic issue. 

The fixed-opcode-count mode triggers benchmarks if they either use a code generator OR contain the target opcode field. All uncacheable precompiles have the target opcode field, so they're selected for this mode, but this mode was designed only for code generators. Code generators takes `attack_block` and repeat for iterations, while precompile benchmarks only pass a single block as a parameter. This mismatch causes unnecessary benchmark runs, and leads to unexpected result here

tests/benchmark/compute/precompile/test_point_evaluation.py
LouisTsai-Csie: > One option is to always increment by the output if the STATICCALL plus a static +1

I chose this simpler approach, as i do not find point evaluation precompile data generator in the consensus tests

marioevz:
I've pushed a ton of fixes in the last commits:
- `Op.CALL*` now have a `inner_call_cost` metadata argument, which is now used in all precompile tests to set the cost of the precompile and be automatically added to the result of `bytecode.gas_cost(fork)`
- Added `expected_opcode_count` field to the `BenchmarkTest` spec to allow a test to set the number of times an opcode must be executed, which is now used in all cacheable tests, and uncovered issues with at least two tests.
- Fixed `_generate_bls12_g2_point` used in the BLS uncachable tests: the coefficients were inverted and this resulted in the precompile failing and exhausting all gas.
- Fixed gas calculation in `test_modexp_uncachable` for forks prior to Osaka.

To fix:
- Point Evaluation precompile uncachable test is still broken: A single call of the precompile is exhausting gas, which potentially means the inputs are incorrect (as suggested by @misilva73)

The rest of the precompile tests look good to me, but the issues found by the new `expected_opcode_count` suggest that we should only use the `WhileGas` when absolutely necessary, because otherwise we could be executing bogus tests where the gas is exhausted by an error.

cc @LouisTsai-Csie

LouisTsai-Csie:
@marioevz  thank you! I push some updates for point evaluation but needs extra eyes on it. cc @misilva73

tests/benchmark/compute/precompile/test_point_evaluation.py
```diff
@@ -57,3 +77,153 @@ def test_point_evaluation(
...
+        )
+        gas_for_loop = (
+            per_tx_gas
+            - execution_intrinsic
+            - setup_cost
+            - math.ceil(len(calldata) / 32) * gsc.GAS_COPY
+            - mem_exp(new_bytes=len(calldata) + 32)
+        )
+
+        if gas_for_loop < iteration_cost:
```
misilva73: For consistency with the BLS12 pairing test, Shouldn't this be multiplied by the number of variants? i.e.:

```suggestion
        if gas_for_loop < (per_tx_variants + 1) * iteration_cost
```

tests/benchmark/compute/precompile/test_point_evaluation.py
```diff
@@ -57,3 +77,153 @@ def test_point_evaluation(
...
+
+    iteration_cost = loop.gas_cost(fork)
+    setup_cost = setup.gas_cost(fork)
+
+    # Conservative per-variant estimate: one iteration + calldata
+    # intrinsic + CALLDATACOPY copy and linear memory expansion.
+    words_per_variant = math.ceil(INPUT_SIZE / 32)
+    per_variant_gas = (
+        iteration_cost
+        + INPUT_SIZE * 16
```
misilva73: To avoid breaking the test whenever call data costs change:

```suggestion
        + INPUT_SIZE * gsc.GAS_TX_DATA_TOKEN_FLOOR
```

misilva73 (commented):
Focussed on the point evaluation benchmark. It looks good. I only pointed two minor changes

Justfile
```diff
@@ -236,7 +236,7 @@ bench-opcode *args:
         --fork Osaka \
         -m repricing \
         -n auto --maxprocesses 10 --dist=loadgroup \
-        -k "not test_alt_bn128 and not test_bls12_381 and not test_modexp" \
+        -k "not test_alt_bn128 and not test_bls12_381 and not test_modexp and not uncachable" \
```
LouisTsai-Csie, line 239: This should be resolved in a separate PR. The CI sometimes takes over 6 hours in fixed-opcode-count mode, which is unexpected. In normal CI pipeline, running 1K opcode count benchmarks completes in ~20 minutes, while the fixed-opcode-count config takes ~3 min.

The root cause is in the collection/filtering logic. Tests without a code generator (i.e., those that provide `blocks=` or `tx=` directly) are only skipped _after_ the test body has already executed. The flow is:

1. The test body runs: pre-allocation, transaction construction, calldata generation, etc.
2. The test calls `benchmark_test(blocks=...)`, which instantiates `BenchmarkTest`.
3. `model_post_init` checks for a code generator, if absent, the test is skipped via `pytest.skip()`.

The expensive work in step 1 is wasted. For uncacheable precompile benchmarks, which construct blocks manually rather than using the code generator, this is especially costly: in fixed-opcode-count mode, `gas_benchmark_value` is set to 1 trillion gas. With Osaka's 16M transaction gas cap, the transaction-building loop iterates ~60,000 times per test: generating calldata, funding EOAs, and creating `Transaction` objecte, all before the inevitable skip.

The fix should move the skip earlier, either by filtering these tests at collection time in `pytest_collection_modifyitems`, or by skipping in the `gas_benchmark_value` fixture when it detects the test will not use a code generator.

tests/benchmark/compute/precompile/test_point_evaluation.py
LouisTsai-Csie: I update point evaluation file and the other one that is using this pattern, thanks for looking into this.

Justfile
marioevz: Created #2678 to track this issue.

tests/benchmark/compute/precompile/test_point_evaluation.py
```diff
@@ -57,3 +70,153 @@ def test_point_evaluation(
...
+        )
+        gas_for_loop = (
+            per_tx_gas
+            - execution_intrinsic
+            - setup_cost
+            - math.ceil(len(calldata) / 32) * gsc.GAS_COPY
+            - mem_exp(new_bytes=len(calldata) + 32)
+        )
+
+        if gas_for_loop < (per_tx_variants + 1) * iteration_cost:
```
marioevz: This comparison seems to be always breaking the loop because, from local runs, no transactions are actually being added to the test.

marioevz (commented):
There's one breaking issue with the test in the point evaluation that is currently make the test not execute any transactions.

marioevz (approved):

PR #2603, opened 2026-03-31
fix(execute): use configurable gas limit for funding txs (EIP-8037)

qu0b:
## Summary
- On EIP-8037 networks, simple value transfers to new accounts require more than 21000 gas due to `GAS_NEW_ACCOUNT` state gas (112 × cost_per_state_byte)
- The `execute remote` framework's default 21000 gas limit for funding transactions causes `intrinsic gas too low` errors during test setup
- This PR uses the existing `--sender-fund-refund-gas-limit` flag for all funding transactions, not just refund/cleanup ones

## Changes
- `contracts.py`: Use 200000 gas for deterministic factory deployer funding tx
- `pre_alloc.py`: Thread `sender_fund_refund_gas_limit` through to `Alloc` and use it for simple EOA funding transactions

## Usage
```bash
uv run execute remote "tests/amsterdam/" \
  --fork Amsterdam \
  --sender-fund-refund-gas-limit 200000 \
  --chain-id <id> --rpc-endpoint <url> --rpc-seed-key <key>
```

Note: The deterministic deployer's pre-signed deploy tx (100K gas) also fails on EIP-8037 networks since contract creation costs 131K+ state gas. Networks should pre-deploy the factory at `0x4e59b44847b379578588920ca78fbf26c0b4956c` in genesis using `additional_preloaded_contracts` in Kurtosis.

## Test plan
- [x] Verified `execute remote` works against bal-devnet-3 Kurtosis network (geth)
- [x] 11 EIP-8037 tests pass, remaining failures are real client divergences

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
@@ -56,7 +56,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         action="store",
         dest="sender_fund_refund_gas_limit",
         type=Wei,
-        default=21_000,
+        default=200_000,
         help=(
             "Gas limit set for the funding transactions of each worker's sender key."  # noqa: E501
         ),
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/contracts.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/contracts.py
@@ -78,6 +78,7 @@ def deploy_deterministic_factory_contract(
         fund_tx = Transaction(
             to=deploy_tx_sender,
             value=fund_amount,
+            gas_limit=200_000,
             gas_price=gas_price,
             sender=seed_key,
         )
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute_recover.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute_recover.py
@@ -24,7 +24,7 @@ def test_recover_funds(
     del index
 
     remaining_balance = eth_rpc.get_balance(eoa)
-    refund_gas_limit = 21_000
+    refund_gas_limit = 200_000
     tx_cost = refund_gas_limit * gas_price
     if remaining_balance < tx_cost:
         pytest.skip(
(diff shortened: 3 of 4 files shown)


fselmo:
Thanks @qu0b, I ended up testing and I found a few more spots where there were discrepancies. We're revisiting the branch hierarchy for devnets and eips. A week ago I would say we should target `eips/amsterdam/eip-8037` with this but this may very well have changed. Tagging @spencer-tb here for a sanity check as the `static` lint fails here are due to a necessary rebase or merge with latest `forks/amsterdam` which would also yield some conflicts here with these changes.

@spencer-tb, what's the recommended path for this change? Still PR against the `eips` branch? And should we rebase that one and then remerge it to the devnet branch? 👀

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/sender.py
```diff
@@ -56,7 +56,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         action="store",
         dest="sender_fund_refund_gas_limit",
         type=Wei,
-        default=21_000,
+        default=200_000,
```
fselmo, line 59: I figured we should bump the default here since this is already targeting a branch with EIP-8037 in it. This should make it so we don't have to pass the updated gas limit via the flag for these tests to properly run but I think this is a good configuration change to make anyhow. Both of these seem like complementary changes to me.

qu0b:
Thanks @fselmo! Applied your suggestion — bumped the default to 200k in `sender.py:59`. Also found and fixed the remaining spots:

- `pre_alloc.py:251` — `funding_gas_limit` default 21k → 200k
- `pre_alloc.py:901` — per-test refund now uses `sender_fund_refund_gas_limit` instead of hardcoded 21k
- `execute_recover.py:27` — recovery refund 21k → 200k

Re: branch targeting — happy to rebase onto whatever branch you and @spencer-tb decide. The changes are small enough to cherry-pick either way.

fselmo:
Sounds good @qu0b. I am targeting the eip branch here now as this is the right approach. I rebased as well as this was a more updated branch. Still looks like it's not the most up to date as we still have some unrelated lint errors. I'm not sure how far along the rebase process for eips branches are atm.

fselmo (approved):
This lgtm. As the linting issues are unrelated, we should be OK to merge this and resolve those via a future rebase.

spencer-tb:
Thanks @fselmo!! Just fixed the lint errors in: https://github.com/ethereum/execution-specs/pull/2612

PR #2602, opened 2026-03-31
fix(tests): resolve transition fork blob methods in EIP-4844 conftest

danceratopz:
## 🗒️ Description

Transition forks like `ShanghaiToCancunAtTime15k` inherit from `TransitionBaseClass`, not `BaseFork`, so they lack blob-related methods (`blob_gas_per_blob`, `target_blobs_per_block`, etc.). Calling these methods on a transition fork in the EIP-4844 conftest fixtures causes an `AttributeError`.

Fix: chain `.transitions_to()` before each blob method call — a no-op for regular forks, resolves to the target fork for transition forks. This is the established pattern used 30+ times across the codebase.

This bug is not caught in CI because `test_blob_type_tx_pre_fork` is marked `@pytest.mark.slow` and `tox -e py3` runs with `-m "not slow"`.

## 🔗 Related Issues or PRs

Blocks:
- #2592
- A new fixture release.

## ✅ Checklist

- [x] All: Ran fast `tox` checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    uvx tox -e static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).

#### Cute Animal Picture

image

PR #2604, opened 2026-03-31
fix(tests): resolve transition fork blob methods in EIP-7918 conftest

danceratopz:
## 🗒️ Description

Analogous to #2602 but for the EIP-7918 blob reserve price conftest.

Transition forks like `OsakaToBPO1AtTime15k` and `BPO1ToBPO2AtTime15k` inherit from `TransitionBaseClass`, not `BaseFork`, so they lack blob-related methods (`blob_gas_per_blob`, `target_blobs_per_block`, etc.). Calling these methods on a transition fork in the EIP-7918 conftest fixtures causes an `AttributeError`.

Fix: chain `.transitions_to()` before each blob method call — a no-op for regular forks, resolves to the target fork for transition forks.

As with #2602, this bug is not caught in CI because affected tests are marked `@pytest.mark.slow` and `tox -e py3` runs with `-m "not slow"`.

## 🔗 Related Issues or PRs

- Analogous to #2602.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

#### Cute Animal Picture

image

### DIFF
--- a/tests/osaka/eip7918_blob_reserve_price/conftest.py
+++ b/tests/osaka/eip7918_blob_reserve_price/conftest.py
@@ -13,19 +13,19 @@ from .spec import Spec
 @pytest.fixture
 def target_blobs_per_block(fork: Fork) -> int:
     """Return default number of target blobs per block."""
-    return fork.target_blobs_per_block()
+    return fork.transitions_to().target_blobs_per_block()
 
 
 @pytest.fixture
 def max_blobs_per_block(fork: Fork) -> int:
     """Return default number of max blobs per block."""
-    return fork.max_blobs_per_block()
+    return fork.transitions_to().max_blobs_per_block()
 
 
 @pytest.fixture
 def blob_gas_per_blob(fork: Fork) -> int:
     """Return default blob gas cost per blob."""
-    return fork.blob_gas_per_blob()
+    return fork.transitions_to().blob_gas_per_blob()
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
LGTM! Did the same for bal devnet 3! Thx sir :D

tests/osaka/eip7918_blob_reserve_price/conftest.py
```diff
@@ -13,19 +13,19 @@
 @pytest.fixture
 def target_blobs_per_block(fork: Fork) -> int:
     """Return default number of target blobs per block."""
-    return fork.target_blobs_per_block()
+    return fork.transitions_to().target_blobs_per_block()
```
marioevz, line 16: Just pointing out that this is not equivalent to the previous statement since `fork.target_blobs_per_block()` is equivalent to `fork.target_blobs_per_block(block_number=0, timestamp=0)`, so `fork.transitions_from()`. But we should take a deeper look into what the test was actually trying to do.

danceratopz: Is this is a regression due to #2442 @marioevz? If so, I can hasher before that and compare the fixtures/and or traces to this branch.

danceratopz: These tests broke with the change in #2442. I hasher-compared to the commit prior 125ed98; the fixtures have zero diffs.

danceratopz: Fill command:
```
uv run fill --fork OsakaToBPO1AtTime15k --clean \
  tests/osaka/eip7918_blob_reserve_price/test_blob_reserve_price_with_bpo_transitions.py::test_reserve_price_at_transition \
  tests/osaka/eip7918_blob_reserve_price/test_blob_reserve_price_with_bpo.py::test_blob_base_fee_with_bpo_transition \
  --output=fixtures-7918
```

marioevz (approved):
Thanks!

tests/osaka/eip7918_blob_reserve_price/conftest.py
marioevz: Then we should be safe 👍 
Thanks for checking!

danceratopz: I repeated for the full fork range to be sure:
```
uv run fill --from Osaka --until Amsterdam ...
```

PR #2614, opened 2026-04-02
fix(test-benchmark): account access subcall OOG issue

LouisTsai-Csie:
## 🗒️ Description

Fix `test_account_access` benchmark: the `While` loop lacked an exit condition, causing every transaction and subcall to OOG and revert all value transfers. This made the `value_sent=1` variants (`CALL`/`CALLCODE` × 3 account modes × 3 cache strategies = 18 test cases) fail with `BalanceMismatchError`.

- Add gas threshold to `While` loop (`condition=Op.GT(Op.GAS, 0x9000)`) for `value_sent > 0` so the loop terminates cleanly before running out of gas. The threshold (36,864) is just above the worst-case single-iteration cost (CALL + cold + value + new_account = 36,600).

- Skip Bittrex Controller nonce=1 contract via `calldata_offset` for `EXISTING_CONTRACT` mode. The nonce=1 child contract has a non-payable fallback (`INVALID` at offset 44) that reverts when receiving ETH. Nonce ≥ 2 contracts (545 bytes, identical bytecode) have a payable fallback that `STOP` immediately.

- Remove hardcoded `gas=1` from CALL/CALLCODE and `gas=1, args_size=1024` from STATICCALL/DELEGATECALL parameters that were limiting subcall gas, which leads to OOG.

- Compute `total_iterations` for future balance verification; replace `expected_benchmark_gas_used` with `skip_gas_used_validation=True` + `expected_receipt_status=1`, since we could not correctly calculate the gas cost here.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -1890,9 +1890,7 @@ def test_account_access(
         del iteration_count
         return Hash(start_iteration + calldata_offset)
 
-    attack_address = pre.deploy_contract(
-        code=attack_code, balance=10**21
-    )
+    attack_address = pre.deploy_contract(code=attack_code, balance=10**21)
 
     post: dict = {}
     cache_txs = []
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -1842,7 +1842,6 @@ def test_account_access(
         attack_call = Op.POP(
             opcode(
                 address=address_retriever.address_op(),
-                gas=1,
                 value=value_sent,
                 # Gas accounting
                 address_warm=access_warm,
@@ -1855,8 +1854,6 @@ def test_account_access(
         attack_call = Op.POP(
             opcode(
                 address=address_retriever.address_op(),
-                gas=1,
-                args_size=1024,
                 # Gas accounting
                 address_warm=access_warm,
             )
@@ -1873,6 +1870,7 @@ def test_account_access(
 
     loop_code = While(
         body=cache_op + attack_call + increment_op,
+        condition=Op.GT(Op.GAS, 0x9000) if value_sent > 0 else None,
     )
 
     attack_code = IteratingBytecode(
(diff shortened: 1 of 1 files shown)


tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -1842,7 +1842,7 @@ def test_account_access(
         attack_call = Op.POP(
             opcode(
                 address=address_retriever.address_op(),
-                gas=1,
+                gas=0,
```
jochem-brouwer: We forward 0 gas to the subcall here, so any code which consumes any gas will OOG. Either remove `gas` here as parameter if value is sent, or default it to Op.GAS.

Note that the danger with unknown contracts is that it will consume all gas and thus will immediately cost us 63/64 of the gas budget. 

For non-value-transfers the OOG is fine, for value transfers we need it to succeed in order to write it to state and let it be part of the state root calculation 😄 👍

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -1917,6 +1927,11 @@ def calldata(iteration_count: int, start_iteration: int) -> bytes:
             )
         total_gas_cost = sum(tx.gas_cost for tx in attack_txs)
 
+    if value_sent > 0:
+        post[attack_address] = Account(
+            balance=initial_balance - total_iterations * value_sent,
```
jochem-brouwer: This assumes that we calculated the total_iterations correctly. 
Note that we cannot do this, because we do not know how much gas the unknown contract will consume (yes, for repricing we target Bittrex, but we could target any contract). So our gas cost estimation will be lower than the actual cost, and therefore our total_iterations might be higher.
I think the correct test here is that the balance has decreased. Then at least one transaction succeeded and one value transfer succeeded there also.
Note that this verification does not work for contracts blocking ether sends (they will always revert/OOG)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: I remove the value here!

spencer-tb (approved):
LGTM from myside!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
spencer-tb: Will merge for now. Can follow up in another if still needs addressed.

LouisTsai-Csie: My refactor does not calculate the `total_iterations` for the post verification now. 
The current workaround is to add an exit condition for the while loop, i calculate approximate the execution cost for all account mode variants (`EXISTING_EOA`, `NON_EXISTING_EOA` and `EXISTING_CONTRACT`). 

The current approach is far from perfect, since we still lacking the verifcation to the test, i will continue on the validation part in a follow up PR.

PR #2633, opened 2026-04-08
chore(tooling,deps): update mypy to 1.20

SamWilsn:
#### Cute Animal Picture


packages/testing/src/execution_testing/base_types/tests/test_base_types.py
danceratopz, line 292: The updated type annotation is an improvement: `model_type(**json)` requires a mapping, not a string.

danceratopz (approved):
Thanks @SamWilsn, LGTM! Is it faster?! 🙏 😆

PR #2635, opened 2026-04-08
feat(benchmark): Add deep-branch SLOAD stateful test

jsign:
## 🗒️ Description
Add a deep-branch `SLOAD` stateful benchmark.

This extends the existing deep `attack(uint256)` / `SSTORE` scenario so the same pre-mined CREATE2 contracts and depth matrix can also be exercised through `getDeepest()` via `STATICCALL`. 

While for re-execution, an `SLOAD` would usually be read from a snapshot-like db, thus being O(1), this might not hold for all clients. Moreover, for zkVMs, any state access must be accompanied by an MPT proof that must be verified during execution. A deep SLOAD thus has associated work that depends on the branch length, so it is quite important to be covered.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
+++ b/tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
@@ -401,8 +401,10 @@ def build_get_deepest_orchestrator_bytecode(
         Jump over ``INVALID`` when the invariant holds.
         """
         return (
-            # Jump to the JUMPDEST after INVALID. PC points at itself, so we
-            # add the skipped payload (`INVALID`) plus ADD and JUMPI bytes.
+            # PC pushes its own offset; +3 accounts for the
+            # remaining push encoding + ADD + JUMPI bytes that
+            # follow PC, and len(Op.INVALID) accounts for the
+            # INVALID byte to skip over, landing on JUMPDEST.
             Op.JUMPI(Op.ADD(Op.PC, len(Op.INVALID) + 3), condition)
             + Op.INVALID
             + Op.JUMPDEST
--- a/tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
+++ b/tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
@@ -481,7 +481,6 @@ def build_get_deepest_orchestrator_bytecode(
     )
 
 
-@pytest.mark.valid_from("Prague")
 @pytest.mark.parametrize(
     "storage_depth,account_depth",
     DEPTH_BENCHMARK_CASES,
@@ -579,7 +578,6 @@ def test_worst_depth_stateroot_recomp(
     )
 
 
-@pytest.mark.valid_from("Prague")
 @pytest.mark.parametrize(
     "storage_depth,account_depth",
     DEPTH_BENCHMARK_CASES,
@@ -658,7 +656,6 @@ def test_worst_depth_get_deepest(
     benchmark_test(
         pre=pre,
         blocks=[Block(txs=read_txs)],
-        post=Alloc({}),
         expected_benchmark_gas_used=total_gas_cost,
         expected_receipt_status=1,
     )
--- a/tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
+++ b/tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
@@ -1,5 +1,5 @@
 """
-abstract: BloatNet worst-case attack benchmark for maximum SSTORE stress.
+abstract: BloatNet worst-case depth benchmarks for deep SSTORE and SLOAD.
 
 This test implements a worst-case scenario for Ethereum block processing
 that exploits the computational complexity of Patricia Merkle Trie
@@ -7,11 +7,12 @@ operations. It uses CREATE2 to deploy contracts at pre-mined addresses
 with shared prefixes, maximizing trie traversal depth.
 
 Key features:
-- Attacks pre-deployed contracts via CREATE2 address derivation
+- Accesses pre-deployed contracts via CREATE2 address derivation
 - Each contract has deep storage slots with configurable trie depth
-- Executes optimized attack bytecode with multiple SSTORE operations
+- Includes both a mutating `attack(uint256)` benchmark and a read-only
+  `getDeepest()` benchmark
 - Respects Fusaka tx gas limit (16M gas) and fills blocks fully
-- Verifies attack success via a storage check in each of the attack contracts
+- Verifies correctness via post-state checks or receipt-status validation
 
 Test parameters:
 - storage_depth: Depth of storage slots (e.g., 10, 11)
@@ -44,6 +45,7 @@ from execution_testing import (
     TransactionWithCost,
     While,
 )
+from execution_testing.base_types import StorageRootType
 from pydantic import BaseModel, BeforeValidator, Field
 
 # Folder path to the submodule and mined assets
@@ -52,6 +54,27 @@ MINED_ASSETS_PATH = WORST_CASE_MINER_SUBMODULE_PATH / "mined_assets"
 
 # Arbitrary value written to storage slots during attack
 DEFAULT_ATTACK_VALUE = 42
+
+# ABI selectors for the mined contracts
+ATTACK_SELECTOR = Hash(bytes.fromhex("64dd891a"), right_padding=True)
+GET_DEEPEST_SELECTOR = Hash(bytes.fromhex("d9339cc3"), right_padding=True)
+
+# From .worst_case_miner/mined_assets
+DEPTH_BENCHMARK_CASES = [
+    (10, 3),
+    (10, 4),
+    (10, 5),
+    (10, 6),
+    (10, 7),
+    (11, 3),
+    (11, 4),
+    (11, 5),
+    (11, 6),
+    (11, 7),
+    (12, 3),
+    (12, 4),
+    (12, 5),
+]
 
 
 def get_mined_asset(filename: str) -> str:
@@ -156,71 +179,100 @@ def attack_value(request: pytest.FixtureRequest) -> int:
     However during execute we should use a random value to guarantee the
     storage update that's required for the attack.
     """
-    if request.config.pluginmanager.has_plugin(
-        "execution_testing.cli.pytest_commands.plugins.execute.execute"
-    ):
+    if is_execute_mode(request):
         return int(time.time())
     return DEFAULT_ATTACK_VALUE
 
 
-@pytest.mark.valid_from("Prague")
-@pytest.mark.parametrize(
-    "storage_depth,account_depth",
-    [
-        # From .worst_case_miner/mined_assets
-        (10, 3),
-        (10, 4),
-        (10, 5),
-        (10, 6),
-        (10, 7),
-        (11, 3),
-        (11, 4),
-        (11, 5),
-        (11, 6),
-        (11, 7),
-        (12, 3),
-        (12, 4),
-        (12, 5),
-    ],
-)
-def test_worst_depth_stateroot_recomp(
-    benchmark_test: BenchmarkTestFiller,
-    fork: Fork,
-    pre: Alloc,
-    gas_benchmark_value: int,
-    storage_depth: int,
-    account_depth: int,
-    attack_value: int,
-) -> None:
+def is_execute_mode(request: pytest.FixtureRequest) -> bool:
     """
-    BloatNet worst-case SSTORE attack benchmark with pre-deployed contracts.
-
-    This test:
-    1. Derives CREATE2 addresses from initcode_hash + Nick's deployer
-    2. Deploys AttackOrchestrator that calls attack() on each target
-    3. Fills blocks with 16M gas transactions attacking contracts
-    4. Adds a verification transaction at the end to confirm success
-
-    Args:
-        benchmark_test: The benchmark test filler
-        fork: The fork to test on
-        pre: Pre-state allocation
-        gas_benchmark_value: Gas budget for benchmark
-        storage_depth: Depth of storage slots in the contract (e.g., 9)
-        account_depth: Depth of account address prefix sharing (e.g., 5)
-        attack_value: The value to be written to storage in each attack
-
+    Return ``True`` when the test is running through the execute plugin.
     """
-    # Load the mined contract file
-    mined_contract_file = MinedContractFile.load(storage_depth, account_depth)
+    return request.config.pluginmanager.has_plugin(
+        "execution_testing.cli.pytest_commands.plugins.execute.execute"
+    )
 
-    # Generate the attack orchestrator
-    factory_address = (
+
+def get_factory_address(fork: Fork) -> Address:
+    """
+    Return the deterministic deployment factory for the fork.
+    """
+    return (
         fork.deterministic_factory_predeploy_address()
         or DETERMINISTIC_FACTORY_ADDRESS
     )
 
-    # Prepare attack contract
+
+def deploy_required_mined_contracts(
+    *,
+    pre: Alloc,
+    mined_contract_file: MinedContractFile,
+    contracts_required: int,
+    storage_depth: int,
+    account_depth: int,
+    attacked_storage_value: int | None = None,
+) -> Alloc:
+    """
+    Deploy the required set of mined contracts and their auxiliary accounts.
+
+    When ``attacked_storage_value`` is provided, return the post-state
+    assertions for the deepest slot update. Otherwise, return an empty post.
+    """
+    if contracts_required > mined_contract_file.num_contracts:
+        raise RuntimeError(
+            "Depth benchmark requires more mined contracts than available for "
+            f"(storage_depth={storage_depth}, account_depth={account_depth}): "
+            f"required {contracts_required}, available "
+            f"{mined_contract_file.num_contracts}."
+        )
+
+    post = Alloc({})
+    for salt in range(contracts_required):
+        salted_contract_info = mined_contract_file.contracts[salt]
+        assert salted_contract_info.salt == salt, (
+            f"Salt out of order: {salted_contract_info.salt} != {salt}"
+        )
+
+        storage: StorageRootType = dict.fromkeys(
+            mined_contract_file.storage_keys, 1
+        )
+        deployed_contract_address = pre.deterministic_deploy_contract(
+            deploy_code=mined_contract_file.deploy_code,
+            salt=Hash(salt),
+            initcode=mined_contract_file.initcode,
+            storage=storage,
+        )
+        assert (
+            deployed_contract_address == salted_contract_info.contract_address
+        ), (
+            f"Contract address mismatch: {deployed_contract_address} != "
+            f"{salted_contract_info.contract_address}, salt: {salt}"
+        )
+
+        for auxiliary_account in salted_contract_info.auxiliary_accounts:
+            # Ensure the account exists in the state trie
+            pre.fund_address(
+                address=auxiliary_account, amount=1, minimum_balance=True
+            )
+
+        if attacked_storage_value is not None:
+            attacked_storage = storage.copy()
+            attacked_storage[mined_contract_file.storage_keys[-1]] = (
+                attacked_storage_value
+            )
+            post[salted_contract_info.contract_address] = Account(
+                storage=attacked_storage
+            )
+
+    return post
+
+
+def build_attack_orchestrator_bytecode(
+    factory_address: Address,
+) -> IteratingBytecode:
+    """
+    Build the mutating orchestrator that calls ``attack(uint256)``.
+    """
     create2_preimage = Create2PreimageLayout(
         factory_address=factory_address,
         salt=Op.CALLDATALOAD(32),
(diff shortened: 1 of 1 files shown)


tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -53,6 +54,27 @@
 # Arbitrary value written to storage slots during attack
 DEFAULT_ATTACK_VALUE = 42
 
+# ABI selectors for the mined contracts
```
jsign, line 58: While this PR adds a new test, there are some things that aren't worth fully duplicated, so some code was slightly refactored. 

I'll point out if there is any relevant part changed or similar.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -53,6 +54,27 @@
...
+ATTACK_SELECTOR = Hash(bytes.fromhex("64dd891a"), right_padding=True)
+GET_DEEPEST_SELECTOR = Hash(bytes.fromhex("d9339cc3"), right_padding=True)
```
jsign, line 60: This is the new selector for calling `getDeepset()` from the contract.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -156,71 +178,98 @@ def attack_value(request: pytest.FixtureRequest) -> int:
...
+    """
+    Return the deterministic deployment factory for the fork.
+    """
+    return (
+        fork.deterministic_factory_predeploy_address()
+        or DETERMINISTIC_FACTORY_ADDRESS
+    )
+
+
+def deploy_required_mined_contracts(
```
jsign, line 206: This diff looks a bit ugly so might recommend split-view.

The deployement of mined contracts is now extracted into this `deploy_required_mined_contracts(...)`. It does mostly the same, but the only main difference is that it optionally returns an expected `post` to be used in assertions.

For the SSTORE-attack case, works the same as before. For the SLOAD-attack case, it returns an empty `post`.

This is a good moment to explain how the test asserts that these deep-SLOADs are really happening without `post` assertions.

The way this works is that the orchestrator code (i.e. the contract that does all the `getDeepest() uint256` calls) does:
1. Read the returned value from `getDeepest()` and checks isn't zero -- thus we actually read from the deep branch, and not some nonexistent storage slot (which would be some setup error or bug).
2. If that returned value isn't expected, execute an `INVALID` opcode so the tx fails. (Why not `REVERT` will be explained in a later comment).
3. The `blockchain_test(...)` for the new test passes `expected_receipt_status=1` (L643), to assert the tx didn't fail.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -156,71 +178,98 @@ def attack_value(request: pytest.FixtureRequest) -> int:
...
+            )
+
+        if attacked_storage_value is not None:
+            attacked_storage = storage.copy()
+            attacked_storage[mined_contract_file.storage_keys[-1]] = (
+                attacked_storage_value
+            )
+            post[salted_contract_info.contract_address] = Account(
+                storage=attacked_storage
+            )
```
jsign, line 265: This is the conditional code that if is called for the existing SSTORE-attack case, it generates the returned `post`. And if not, `post` is empty (since we check with `expected_receipt_status=1` as mentioned before).

The rest of the code in this method is the same as pre-PR.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -156,71 +179,100 @@ def attack_value(request: pytest.FixtureRequest) -> int:
...
 
-    # Prepare attack contract
+    return post
+
+
+def build_attack_orchestrator_bytecode(
```
jsign, line 270: This is just an extracted of existing code.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,176 @@ def test_worst_depth_stateroot_recomp(
...
+    """
+    Build the read-only orchestrator that calls ``getDeepest()``.
+    """
+    create2_preimage = Create2PreimageLayout(
+        factory_address=factory_address,
+        salt=Op.CALLDATALOAD(0),
+        init_code_hash=Op.CALLDATALOAD(64),
+    )
+    args_offset = 96
+    return_offset = 128
```
jsign, line 374: `return_offset` is where we'll get the read deep-SLOAD value, as I mentioned before (i.e. to revert the tx or not).

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,176 @@ def test_worst_depth_stateroot_recomp(
...
+    success_offset = 160
```
jsign, line 375: To check STATICCALL didn't fail.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,176 @@ def test_worst_depth_stateroot_recomp(
...
+) -> IteratingBytecode:
+    """
+    Build the read-only orchestrator that calls ``getDeepest()``.
+    """
+    create2_preimage = Create2PreimageLayout(
+        factory_address=factory_address,
+        salt=Op.CALLDATALOAD(0),
+        init_code_hash=Op.CALLDATALOAD(64),
+    )
+    args_offset = 96
```
jsign, line 373: For this SLOAD-attack case there are no parameters in `getDeepest()` compared to the other `attack(value)` -- so only the selector is passed.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,176 @@ def test_worst_depth_stateroot_recomp(
...
+        )
+        + Op.CALLDATALOAD(32)
+    )
+    return_value_check = (
+        Op.EQ(Op.MLOAD(return_offset), 1)
+        if exact_expected_value
+        else Op.ISZERO(Op.ISZERO(Op.MLOAD(return_offset)))
+    )
+
+    def assert_with_invalid(condition: Bytecode | Op) -> Bytecode:
```
jsign, line 399: See the next comment.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,176 @@ def test_worst_depth_stateroot_recomp(
...
+            # scaffolding. This manual branch keeps that overhead minimal,
+            # and uses ``INVALID`` as the cheapest terminating failure path.
+            + assert_with_invalid(
+                condition=Op.AND(
+                    Op.AND(
+                        Op.MLOAD(success_offset),
+                        Op.EQ(Op.RETURNDATASIZE, 32),
+                    ),
+                    return_value_check,
+                )
```
jsign, line 440: As explained in the comment, the thing here is that a later `tx_iterations_by_gas_limit(...)` for this code will eventually call `Bytecode.gas_cost()`, and this method just takes all opcodes and sum gas.

If we have `Conditional` usage here, then `Bytecode.gas_cost()` will give a wrong gas sum. I actually found this when filling the test and saw the gas check failing. :)

So the code basically does the same idea of reverting the transaction, but with an INVALID which doesn't require PUSHX stuff for the REVERT (and is also zero cost). So the JUMP condition of `assert_with_invalid` has always the same gas cost, so the "just sum all opcodes gas cost" that `Bytecode.gas_cost()` does is okay.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,176 @@ def test_worst_depth_stateroot_recomp(
...
+        ),
+        condition=Op.LT(
+            Op.MLOAD(create2_preimage.salt_offset),
+            Op.DUP1,
+        ),
+    )
+
+    # Gas accounting model for the Solidity dispatch path and getDeepest body.
+    # PUSH/DUP/SWAP variants are collapsed into gas-equivalent representatives.
+    inner_call_bytecode = (
```
jsign, line 452: Validated from looking at `tests/benchmark/stateful/bloatnet/depth_benchmarks/.worst_case_miner/mined_assets` json fields with the raw bytecodes -- I think in same style as the attack bench did.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,176 @@ def test_worst_depth_stateroot_recomp(
...
+        iterating_subcall=inner_call_bytecode,
+    )
+
+
+@pytest.mark.valid_from("Prague")
+@pytest.mark.parametrize(
+    "storage_depth,account_depth",
+    DEPTH_BENCHMARK_CASES,
+)
+def test_worst_depth_stateroot_recomp(
```
jsign, line 490: Same test as before but with usage of extracted methods.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -388,3 +577,88 @@ def calldata(iteration_count: int, start_iteration: int) -> bytes:
...
         expected_benchmark_gas_used=total_gas_cost,
     )
+
+
+@pytest.mark.valid_from("Prague")
+@pytest.mark.parametrize(
+    "storage_depth,account_depth",
+    DEPTH_BENCHMARK_CASES,
+)
+def test_worst_depth_get_deepest(
```
jsign, line 587: The new test

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -388,3 +577,88 @@ def calldata(iteration_count: int, start_iteration: int) -> bytes:
...
+    )
+
+    total_gas_cost = sum(tx.gas_cost for tx in read_txs)
+
+    benchmark_test(
+        pre=pre,
+        blocks=[Block(txs=read_txs)],
+        post=Alloc({}),
+        expected_benchmark_gas_used=total_gas_cost,
+        expected_receipt_status=1,
```
jsign, line 662: Note this line as mentioned before to assert all txs succeeded, thus the orchestrator checked SLOAD read an expected value.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -388,3 +577,88 @@ def calldata(iteration_count: int, start_iteration: int) -> bytes:
...
+    BloatNet worst-case SLOAD benchmark with deep CREATE2-derived contracts.
+
+    This benchmark calls ``getDeepest()`` on pre-mined contracts and validates
+    inside the orchestrator that every call succeeds and returns the expected
+    value. Fill mode requires the exact constructor-initialized value ``1``,
+    while execute mode only requires a non-zero value to tolerate reused
+    deterministic deployments on a dirty chain.
+    """
+    mined_contract_file = MinedContractFile.load(storage_depth, account_depth)
+    exact_expected_value = not is_execute_mode(request)
```
jsign, line 606: The expected value to be read in the SLOAD is:
- 1 for fill mode
- != 0 for execution mode

I did this since I think in execution mode since we reuse contracts other `attack` calls can change the value, so expecting `1` wouldn't be correct.

For fill mode we can be tighter and expect `1` for extra correctness check.

LMK if this makes sense.

jsign:
cc @CPerezz for some eyes

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
LouisTsai-Csie: Good idea!

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
LouisTsai-Csie: I think this works fine!

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,176 @@ def test_worst_depth_stateroot_recomp(
...
+
+    return IteratingBytecode(
+        setup=setup,
+        iterating=iterating,
+        cleanup=Op.STOP,
+        iterating_subcall=inner_call_bytecode,
+    )
+
+
+@pytest.mark.valid_from("Prague")
```
LouisTsai-Csie:
```suggestion
```
All the test should start from Prague, even without explicitly being set!

Related logics:

https://github.com/ethereum/execution-specs/blob/a48e0b381d5225a6c3de2d06cd9ee7ae0b6ca9bb/tests/benchmark/conftest.py#L28-L37

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -388,3 +577,88 @@ def calldata(iteration_count: int, start_iteration: int) -> bytes:
         post=post,
         expected_benchmark_gas_used=total_gas_cost,
     )
+
+
+@pytest.mark.valid_from("Prague")
```
LouisTsai-Csie:
```suggestion
```

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -388,3 +577,88 @@ def calldata(iteration_count: int, start_iteration: int) -> bytes:
...
+            calldata=calldata,
+        )
+    )
+
+    total_gas_cost = sum(tx.gas_cost for tx in read_txs)
+
+    benchmark_test(
+        pre=pre,
+        blocks=[Block(txs=read_txs)],
+        post=Alloc({}),
```
LouisTsai-Csie:
```suggestion
```

LouisTsai-Csie (approved):
Awesome! I just leave some nit comments.

I will trigger the gas-benchmarks to verify this test works correctly on bloatnet with execute-remote before merging, thanks!

LouisTsai-Csie:
Good from my side, waiting for @CPerezz for the final approval

CPerezz:
OMW!!!

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -53,6 +55,27 @@
...
+    (10, 7),
+    (11, 3),
+    (11, 4),
+    (11, 5),
+    (11, 6),
+    (11, 7),
+    (12, 3),
+    (12, 4),
+    (12, 5),
+]
```
CPerezz, line 77: I'll need to regenerate all these after https://github.com/CPerezz/worst_case_miner/pull/7/ is merged. But it's ok to have as reference so far

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -156,71 +179,100 @@ def attack_value(request: pytest.FixtureRequest) -> int:
...
     """
-    if request.config.pluginmanager.has_plugin(
-        "execution_testing.cli.pytest_commands.plugins.execute.execute"
-    ):
+    if is_execute_mode(request):
         return int(time.time())
     return DEFAULT_ATTACK_VALUE
 
 
-@pytest.mark.valid_from("Prague")
```
CPerezz, line 166: We removed `@pytest.mark.valid_from("Prague")`for any reason in particular?

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -156,71 +179,100 @@ def attack_value(request: pytest.FixtureRequest) -> int:
...
-@pytest.mark.parametrize(
-    "storage_depth,account_depth",
```
CPerezz, line 168: same with this

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -156,71 +179,100 @@ def attack_value(request: pytest.FixtureRequest) -> int:
...
     pre: Alloc,
-    gas_benchmark_value: int,
+    mined_contract_file: MinedContractFile,
+    contracts_required: int,
     storage_depth: int,
     account_depth: int,
-    attack_value: int,
-) -> None:
+    attacked_storage_value: int | None = None,
+) -> Alloc:
```
CPerezz, line 214: I'm missing a parametrization to trigger the SLOAD or the SSTORE case no? The will require a different number of contracts & accounts mined for example.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
```diff
@@ -297,13 +349,175 @@ def test_worst_depth_stateroot_recomp(
...
+    )
+
+    def assert_with_invalid(condition: Bytecode | Op) -> Bytecode:
+        """
+        Jump over ``INVALID`` when the invariant holds.
+        """
+        return (
+            # Jump to the JUMPDEST after INVALID. PC points at itself, so we
+            # add the skipped payload (`INVALID`) plus ADD and JUMPI bytes.
+            Op.JUMPI(Op.ADD(Op.PC, len(Op.INVALID) + 3), condition)
```
CPerezz: This implies `+3 = INVALID + ADD + JUMPI`. But `len(Op.INVALID) + 3 = 1 + 3 = 4`, and `len(Op.INVALID)` already accounts for INVALID. The `+3` actually accounts for the bytes between `PC` and the skipped payload: the rest of the `PUSH` encoding for the constant, `ADD`, and `JUMPI`. The formula is correct (the offset lands on `JUMPDEST`), but the comment explains it incorrectly.


```suggestion
            # "PC pushes its own offset; +3 accounts for the remaining push encoding + ADD + JUMPI bytes 
            # that follow PC, and len(Op.INVALID) accounts for the INVALID byte to skip over, landing on
            # JUMPDEST."
            Op.JUMPI(Op.ADD(Op.PC, len(Op.INVALID) + 3), condition)
```

CPerezz (commented):
We have no post-state checks for SLOAD case. Which means that we actually don't know if we executed against valid state or not. 

We should likely fix this?

jsign:
> We have no post-state checks for SLOAD case. Which means that we actually don't know if we executed against valid state or not.
> 
> We should likely fix this?

@CPerezz, in my PR comments I explain that we don't need post-state checks to test SLOAD correctness. I check that SLOAD tests are working correctly with other checks -- in my [comments](https://github.com/ethereum/execution-specs/pull/2635/changes#r3051921857) I explain how it works.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
jsign: @LouisTsai-Csie actually asked me to remove them -- see his 9resolved) comments.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
jsign: Ditto my last response.

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
jsign: Both use the same contracts and there is a check that enough contracts are available for the intent (see the `contracts_required`).

So if we use a single set with enough contracts, I think it should be fine, no?

CPerezz:
> > We have no post-state checks for SLOAD case. Which means that we actually don't know if we executed against valid state or not.
> > We should likely fix this?
> 
> @CPerezz, in my PR comments I explain that we don't need post-state checks to test SLOAD correctness. I check that SLOAD tests are working correctly with other checks -- in my [comments](https://github.com/ethereum/execution-specs/pull/2635/changes#r3051921857) I explain how it works.

I just saw it! My bad. Makes sense the trick of 1 or 0 for the return!
LGTM!

tests/benchmark/stateful/bloatnet/depth_benchmarks/test_deep_branch.py
LouisTsai-Csie: We do not need to explicitly specify the starting fork if it is `Prague`:

https://github.com/ethereum/execution-specs/blob/3687eb564090c94ea49601adb284df4e4ff418e7/tests/benchmark/conftest.py#L28-L37

The `conftest.py` configure the starting fork to Prague, even before the test being run. You could check the `compute/` benchmark, there is no fork marker at all.

PR #2637, opened 2026-04-08
feat(tests): stateful test filling for benchmark tests

fselmo:
## 🗒️  Description

Adds a new `fill-stateful` CLI command that generates `BlockchainEngineStatefulFixture` files by running benchmark tests against a live EL client via `testing_buildBlockV1`. Replaces the gas-benchmarks mitm-proxy-based approach.

Key points:
- Connects to a running client loaded with a network snapshot (kurtosis / perfnet / bloatnet), anchored by `--snapshot-block` hash
- Records engine API payloads per test, separated into setup and execution phases; mixed-phase user blocks are split into contiguous-phase sub-blocks so benchmark gas lands on the right block
- Resets chain state between tests via `debug_setHead`, verified by post-rewind hash match; retries through transient `SYNCING` responses
- Funds test accounts via CL withdrawals — no pre-funded seed key required
- Outputs per-test fixture JSON (carries `benchmarkGasUsed` and the EL build in `_info.filling-transition-tool` for traceability) plus a `pre_run/global_setup.json` for session-level setup (factory deploy, seed funding)
- Tx gas cap defaults to `Fork.transaction_gas_limit_cap()` (EIP-7825) when `--max-gas-per-test` is unset

The generated fixtures are consumed by `benchmarkoor` to replay benchmarks against any EL client from the same snapshot.

---

Run against a client that exposes `testing`, `eth`, `debug` namespaces and supports `testing_buildBlockV1`:

```bash
uv run fill-stateful \
    --clean \
    --engine-jwt-secret-file /path/to/jwt \
    --fork Osaka \
    --snapshot-block 0x<32-byte-hash> \
    tests/benchmark/stateful/ \
    -k ...
  ```

Optional:

- `--rpc-endpoint` — defaults to `http://localhost:8545`
- `--engine-endpoint` — derived from `--rpc-endpoint` with port `8551`
- `--chain-id` — auto-detected from the client
- `--rpc-seed-key` — random key generated + funded via CL withdrawal if unset; pass `0x<64hex>` for reproducible fills
- `--snapshot-block` — accepts a 32-byte block hash (recommended), an integer block number (hex `0x...` or decimal), or unset for the client's `latest`
- `--max-gas-per-test` — overrides the fork's tx gas limit cap
- `--output` — defaults to `./fixtures`

## 🔗 Related Issues or PRs

#2560 — partial: fill-side driver via `testing_buildBlockV1`

## ✅ Checklist

    ```console
    just static
    ```
- [x] Cute animal picture 🙀 

#### Cute Animal Picture

Screenshot 2026-04-14 at 13 27 48

### CODE REVIEW

File: packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py [hunk 1]
```diff
@@ -303,3 +312,51 @@ def send_transactions(
...
+            for i, (addr, wei_amount) in enumerate(funding_targets)
+        ]
+
+        with self.block_building_lock:
+            head_block = self.get_block_by_number("latest")
+            assert head_block is not None
+            next_block_number = int(HexNumber(head_block["number"]) + 1)
+            next_timestamp = int(HexNumber(head_block["timestamp"]) + 1)
+            payload_attributes = self._payload_attributes(
+                next_block_number=next_block_number,
```
jochem-brouwer, line 349: Block number is not part of payload [attributes](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#payloadattributesv1)? See also the amsterdam version of [PayloadAttributesV4](https://github.com/ethereum/execution-apis/blob/main/src/engine/amsterdam.md#payloadattributesv4)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
```diff
@@ -303,3 +312,51 @@ def send_transactions(
...
+                next_timestamp=next_timestamp,
+                withdrawals=withdrawals,
+            )
+            new_payload = self.testing_rpc.build_block(
```
jochem-brouwer, line 353: API ref: https://github.com/ethereum/execution-apis/pull/747

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
```diff
@@ -303,3 +312,51 @@ def send_transactions(
...
+                parent_block_hash=Hash(head_block["hash"]),
+                payload_attributes=payload_attributes,
+                transactions=None,
```
jochem-brouwer, line 356: > If the `transactions` parameter is JSON `null`, the client MAY build a block from its local transaction pool (mempool). (from https://github.com/ethereum/execution-apis/blob/main/src/testing/testing_buildBlockV1.yaml)
I don't think we want this (?), I think 
```suggestion
                transactions=[],
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,602 @@
...
+@pytest.fixture(scope="session")
+def sender_funding_transactions_gas_price(eth_rpc: EthRPC) -> int:
+    """Gas price for funding transactions (replaces sender)."""
+    return eth_rpc.gas_price()
+
+
+@pytest.fixture(scope="session")
+def sender_fund_refund_gas_limit() -> int:
+    """Gas limit for fund/refund transactions (replaces sender)."""
+    return 21_000
```
jochem-brouwer, line 403: (should be read from fork config as this changes in eip2780)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,602 @@
...
+
+@pytest.fixture(scope="session")
+def snapshot_block(eth_rpc: EthRPC) -> dict:
+    """
+    Capture the raw datadir head before global setup.
+
+    Resolved before ``execute_required_contracts`` via fixture
+    dependency, so this records the pre-setup chain head.
+    """
+    block = eth_rpc.get_block_by_number("latest")
```
jochem-brouwer, line 414: I think this should not be queried by number, this should be a command line argument where you provide the blockhash. A number is ambiguous, if you have a client and you stop the client at chain head, if it then reorgs we are now building on a block which no one has 😅 \

If no argument for snapshot hash provided, it is ok (with warn/log) to read chain head and use that.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,602 @@
...
+    info = _node_to_test_info(request.node)
+    fixture_collector.add_fixture(info, fixture)
+    logger.info(f"Captured stateful fixture for {request.node.nodeid}")
+
+    # Reset chain to start block so the next test starts from
+    # identical post-global-setup state.
+    start_hex = start_block["number"]
+    logger.info(f"Resetting chain to start block {start_hex}")
+    try:
+        debug_rpc.set_head(start_hex)
```
jochem-brouwer, line 580: I'm not sure what this does, would it also set forkchoice in the client correctly? I think this is what should be done instead using standardized APIs:

Let's say we have done the setup and we have reached the start block. Build two testing_buildBlock on top of this start block. Everything should be empty, no withdrawals, no transactions) and all other parameters should be equal *except* extraData. One of the blocks will be the block hash you use as parentHash in testing_buildBlockV1 and the other one is the reorg block. We *never* build on top of this blockhash, we only use this to reorg to.

Why? If you try to fcu back to a block which is already in the current canonical chain, then most/all clients will throw. If we use this approach by creating two different blocks which are basically equal except their extraData, we can use it as fork block because everything is the same (RPC calls would also return the correct data on state, except of course last block hash).

This would use existing API and not debug_setHead as I think this differs for all clients and there might be unforeseen side effects.

BTW, this double-block approach is not my idea, it's from @kamilchodola 😄 👍 

If something unclear here let us know 😄 👍

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -822,6 +824,23 @@ class BlockchainEngineStatefulFixture(BlockchainEngineFixtureCommon):
     )
 
 
+class StatefulPreRunFixture(CamelModel):
```
jochem-brouwer, line 827: Is this a special class or a class with a special name? I think these prerun fixtures are essentially the same layout as normal fixtures right?

For benchmark/repricing effort we would need to control this presetup also. For repricing and also stateful benchmarks we cannot do test setup right before the benchmark block (this loads all state into memory cache and makes the ops super fast). We would need to create this state already in the snapshot / start block (the block where all tests start from).
To give an example, we would take a recent snapshot of mainnet and now in the prerun fill it will the desired data for repricings (basically setting up the state which we also have on perf-devnet-3, except that this entire state is much bigger and is thus worse than mainnet). 

Once this is done, we don't run any tests (yet) and now use the chain state right after this prerun as the snapshot state. This is what we distribute to benchmarkoor. We also fill the benchmark fixtures on top of this.

We do this because now the state which we target in the benchs is already in the snapshot (we do not create it when actually running the benchs). This forces the client to actually read the DB and not use some memory cache.

jochem-brouwer: (Not super urgent because can also write one big test and once test finishes just use that to get the desired snapshot)

packages/testing/src/execution_testing/rpc/rpc.py
```diff
@@ -1167,6 +1167,12 @@ def trace_call(self, tr: dict[str, str], block_number: str) -> Any | None:
             request=RPCCall(method="traceCall", params=params)
         ).result_or_raise()
 
+    def set_head(self, block_number: str) -> None:
+        """`debug_setHead`: Reset chain head to the given block."""
+        self.post_request(
+            request=RPCCall(method="setHead", params=[block_number])
```
jochem-brouwer, line 1173: The fact that this is a number and not a hash indicates that we cannot for instance perform a reorg via this RPC call (we can only roll back into the canonical chain). This would be limiting. Also don´t know if this updates the forkchoice status of client.

jochem-brouwer (commented):
Super exciting this is already close to what we would need! This is starting to come close to indeed what the MITM is doing, except now EELS is driving it 😄 👍 

Lemme know if there are any questions after this review 😄 👍

jochem-brouwer:
Okay I realized I might be too skeptical about debug_setHead :sweat_smile: Ignore that if it's not right.

packages/testing/src/execution_testing/fixtures/blockchain.py
LouisTsai-Csie: This is a special setup phase that already exists in the gas-benchmarks mechanism (which is globals setting, or something else) It's not for individual benchmark setup; rather, it ensures common contracts (like the factory contract using Nick's method) are pre-deployed on the network. Without this initial setup, all subsequent create2 address calculations and contract deployments would fail.

packages/testing/src/execution_testing/cli/pytest_commands/fill_stateful.py
```diff
@@ -0,0 +1,30 @@
...
+    """Initialize the fill-stateful command."""
+    return PytestCommand(
+        config_file="pytest-fill-stateful.ini",
+    )
+
+
+@click.command(
+    name="fill-stateful",
+    context_settings={
+        "help_option_names": ["-h", "--help"],
```
LouisTsai-Csie, line 20: I run `uv run fill-stateful --help` command and this shows up, is it expected behavior? 

```bash
(ethereum-execution) caijiacheng@caijiachengs-MacBook-Air wt-forks-op-gas-cost % uv run fill-stateful --help
Retrying execution_testing.rpc.rpc.BaseRPC._make_request in 0.5 seconds as it raised ConnectionError: HTTPConnectionPool(host='localhost', port=8545): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x10be56e40>: Failed to establish a new connection: [Errno 61] Connection refused')).
Retrying execution_testing.rpc.rpc.BaseRPC._make_request in 1.0 seconds as it raised ConnectionError: HTTPConnectionPool(host='localhost', port=8545): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x10bedc910>: Failed to establish a new connection: [Errno 61] Connection refused')).
Retrying execution_testing.rpc.rpc.BaseRPC._make_request in 2.0 seconds as it raised ConnectionError: HTTPConnectionPool(host='localhost', port=8545): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x10bedccd0>: Failed to establish a new connection: [Errno 61] Connection refused')).
```

Or i should connect to the node first?

LouisTsai-Csie: Not sure if this the root cause: the `kwargs` below (see `fill_stateful` function) is deleted and thus not being passed to the `is_help_or_collectonly_mode` function

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
LouisTsai-Csie: I agree, using the `latest` block can lead to unexpected behavior, like we saw in `test_ec_pairing`, where the block did not reset correctly to the original snapshot start point.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,602 @@
...
+    if not config.getoption("rpc_endpoint", default=None):
+        config.option.rpc_endpoint = "http://localhost:8545"
+    if not config.getoption("engine_endpoint", default=None):
+        parsed = urlparse(config.getoption("rpc_endpoint"))
+        config.option.engine_endpoint = urlunparse(
+            parsed._replace(netloc=f"{parsed.hostname}:8551")
+        )
+
+    if is_help_or_collectonly_mode(config):
+        return
```
LouisTsai-Csie, line 252: Not sure - should this check done before the `rpc_endpoint` and `engine_endpoint` check

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,602 @@
...
+    if version >= 3:
+        blob_hashes = (
+            response.blobs_bundle.blob_versioned_hashes()
+            if response.blobs_bundle is not None
+            else []
+        )
+        params.append(blob_hashes)
+        params.append(captured.payload_attributes.parent_beacon_block_root)
+    if version >= 4 and response.execution_requests is not None:
+        params.append(response.execution_requests)
```
LouisTsai-Csie, line 181:
```suggestion
    if version >= 4:
       assert response.execution_requests is not None
        params.append(response.execution_requests)
```
I checked Engine API v4+ ([link](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false)), and found the `request` field is required. Should we assert its presence instead of ignoring it when it’s None?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,602 @@
...
+                    phase = getattr(meta, "phase", None)
+            phases.add(phase)
+
+        phases.discard(None)
+        if len(phases) == 1:
+            return phases.pop()
+        # Mixed phases: SETUP takes precedence (a block containing
+        # any setup transaction is considered part of the setup).
+        if TestPhase.SETUP in phases:
+            return TestPhase.SETUP
```
LouisTsai-Csie, line 148: As I recall, some benchmarks include a single setup transaction for 7702 authorizations, followed by the benchmark transactions within the same block. In these cases, we will be labeling the block as a setup block.

This isn’t an issue for gas-benchmarks, since they keep collecting transactions within the same phase until either the buffer reaches the block gas limit or the phase changes. At that point, it flushes the transactions into a block and generate the payload.

LouisTsai-Csie (changes requested):
`fill_stateful.py` is mixing three things, types, pytest hooks, and fixtures, now without clear separation. It's a bit hard to know where to start. 

The entry point (`pytest_configure`) is at L234 behind classes and helpers, and `pytest_sessionfinish` is way down at L587. Plus, fixtures are declared out of order (e.g., `execute_required_contracts` at L291 depends on stuff at L406 and L428), so we need to jump around to understand the flow.

I would suggest adding section headers and reorganizing (maybe types → helpers → pytest hooks → fixtures by lifecycle → per-test fixtures) might make it way easier to follow. wdyt?

LouisTsai-Csie:
Quick question: who's responsible for mounting clients to the snapshot? In gas-benchmarks, Nethermind clients are mounted via overlay, a layer that captures state changes without modifying the snapshot itself. Does this PR have a similar mechanism? If not, where would that be handled?

fselmo:
@LouisTsai-Csie I appreciate the review. This is a working PoC, not something I plan to merge as is. I should've clarified it's still a work in progress and I'm actually making this a lot more `fill` based, rather than `execute` based. I will try to update this soon. Your feedback is valuable for future additions, thanks 🙏🏼.

docs/filling_tests/fill_stateful.md
```diff
@@ -0,0 +1,245 @@
+# Filling Stateful Benchmark Fixtures
+
+The `fill-stateful` command produces `BlockchainEngineStatefulFixture` JSON for benchmark tests by driving block construction on a live EL client via `testing_buildBlockV1` against a pre-loaded network snapshot. The fixtures are replayed by [`benchmarkoor`](https://github.com/ethpandaops/benchmarkoor) against the same snapshot on any EL client. This replaces the gas-benchmarks MITMProxy approach.
+
+!!! note "When to use `fill-stateful`"
+    Use the standard `fill` for t8n-based fixture generation. Use `fill-stateful` only for benchmark tests under `tests/benchmark/stateful/` that need a warm network state (perfnet / bloatnet / kurtosis snapshot) too large to bake into genesis.
```
LouisTsai-Csie: Partially correct, we could also do compute benchmark on stateful environment, to see the state size impact for the opcode.

Suggestion:

```
`fill-stateful` runs benchmarks in a stateful environment (e.g., perfnet, Kurtosis, or other snapshots) so that we can observe the state size impact. Note that benchmarks under `tests/benchmark/stateful/` can only run in such stateful environments.
```

packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-fill-stateful.ini
```diff
@@ -0,0 +1,23 @@
+[pytest]
+console_output_style = count
+minversion = 7.0
+python_files = test_*.py
+testpaths = tests/
```
LouisTsai-Csie:
```suggestion
testpaths = tests/benchmark
```

We could restrict the test path and drop the following `--ignore` parameters.

packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-fill-stateful.ini
```diff
@@ -0,0 +1,23 @@
...
+    -p execution_testing.cli.pytest_commands.plugins.execute.rpc.remote
+    -p execution_testing.cli.pytest_commands.plugins.shared.benchmarking
+    -p execution_testing.cli.pytest_commands.plugins.shared.transaction_fixtures
+    -p execution_testing.cli.pytest_commands.plugins.help.help
+    -p execution_testing.cli.pytest_commands.plugins.fill_stateful.fill_stateful
+    --tb short
+    -n 0
+    --ignore tests/cancun/eip4844_blobs/point_evaluation_vectors/
+    --ignore tests/json_loader
+    --ignore tests/evm_tools
```
LouisTsai-Csie:
```suggestion
```

These could be removed since the test path being configured to `tests/benchmark`

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,522 @@
...
+    """
+    Override of ``live_client_flags.max_gas_limit_per_test`` — fall back
+    to ``Fork.transaction_gas_limit_cap()`` (EIP-7825) when the CLI flag
+    is unset. Returns ``None`` on pre-Osaka forks (no cap).
+    """
+    cli_value = request.config.getoption("max_gas_per_test")
+    if cli_value is not None:
+        return cli_value
+    fork_at_genesis = session_fork.fork_at(block_number=0, timestamp=0)
+    return fork_at_genesis.transaction_gas_limit_cap()
```
LouisTsai-Csie, line 433: I think this becomes a problem when eip-8037 goes live in `forks/amsterdam`. It would cap the transaction gas limit to 16M if the CLI version not configured, but this value is incorrect since it does not take state gas into account.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/live_client_flags.py
```diff
@@ -0,0 +1,321 @@
...
+@pytest.fixture(scope="session")
+def dry_run(request: pytest.FixtureRequest) -> bool:
+    """Return True if the test is a dry run."""
+    return request.config.getoption("dry_run")
+
+
+@pytest.fixture(scope="session")
+def max_transactions_per_batch(request: pytest.FixtureRequest) -> int | None:
+    """Return max transactions per batch, or None for default."""
+    return request.config.getoption("max_tx_per_batch")
```
LouisTsai-Csie, line 205: Just a note, nothing needed change here:

For some of the benchmark, like ether transfer benchmark, there would be more than 750 transactions in the benchmark. This limit would cause the benchmark txs being truncated.

packages/testing/src/execution_testing/client_clis/client_backend.py
```diff
@@ -0,0 +1,343 @@
...
+        return
+
+    def set_cache(self, *, key: str) -> bool:
+        """No-op — caching not meaningful for a live client."""
+        del key
+        return False
+
+    def remove_cache(self) -> None:
+        """No-op — caching not meaningful for a live client."""
+        return
```
LouisTsai-Csie, line 139: These are t8n compatibility feature, not used in engine api use case. We could use a abstract layer to share these no-op defaults across `ClientBackend` and `t8n`

LouisTsai-Csie: FillerBackend is currently a `Protocol`. Switching to an `ABC` would let these no-ops move into the base class as shared defaults, with `evaluate` and `get_traces` marked `@abstractmethod`.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,522 @@
...
+            pytest.exit("Failed to fetch 'latest' block as snapshot anchor")
+        return block
+
+    stripped = arg.strip()
+    lower = stripped.lower()
+    is_hash = (
+        lower.startswith("0x")
+        and len(lower) == 66
+        and all(c in "0123456789abcdef" for c in lower[2:])
+    )
```
LouisTsai-Csie:
```suggestion
    is_hash = bool(re.fullmatch(r"(?i)0x[0-9a-f]{64}", stripped))
```

marioevz (approved):
Overall I think this looks amazing! A couple of thing for follow up PRs:
- We can improve `make_stateful_fixture` to let `pre_alloc` to do its thing before reaching the fixture filling
- Could create a new `pre` class specifically defined for stateful filling
- Could add two-step filling by grouping tests like we do for pre-alloc-groups

I pushed a couple of commits to do the following:
- Integrate fill-stateful into PhaseManager to avoid having to monkey-patch the supported fixture-formats in the test spec class at runtime
- Add hive support, to allow easier debugging (Use `--hive-mode --hive-simulator http://127.0.0.1:3000`)

cc @LouisTsai-Csie deferring to you to approve before merge

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -0,0 +1,724 @@
...
+        "ClientBackend fees pinned: "
+        f"gas_price={gas_price / 10**9:.2f} Gwei, "
+        f"max_fee={max_fee / 10**9:.2f} Gwei, "
+        f"priority={priority_fee / 10**9:.2f} Gwei, "
+        f"blob={blob_fee / 10**9:.2f} Gwei"
+    )
+    return backend
+
+
+def _resolve_snapshot_block(
```
LouisTsai-Csie: `snapshot_block` could be a session scope fixture, instead of being a function called by `_session_pre_run`. And i would support removing fetching `latest` block as the snapshot block, we should always explicitly provide either block number or block hash

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1158,12 +1311,205 @@ def make_hive_fixture(
...
+        - No genesis building: the client already has warm state from a
+          snapshot. The backend (typically ``ClientBackend``) owns
+          ``snapshot_block`` / ``start_block`` captured at session start.
+        - ``pre.fund_eoa`` / ``pre.deploy_contract`` calls are materialised
+          into a synthetic setup block prepended to ``self.blocks``,
+          instead of being baked into genesis alloc.
+        - Payloads are partitioned by ``FixtureEngineNewPayload.phase``
+          into ``setup_payloads`` (setup-phase txs) and ``payloads``
+          (execution-phase txs).
+        - ``verify_post_state`` is skipped: the client is the oracle.
```
LouisTsai-Csie, line 1332: We still do post verification in benchmark, like the total gas used, and the receipt check, I am not sure if these are also ignored. If so, we need to add the post verification back,

LouisTsai-Csie (approved):
I leave some comment, but not blocking this PR, i could create follow-up PR for fixes 👍🏻 

For potential refactor and missing feature, i create issue #2910 for tracking

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
marioevz: I agree the runner should always provide the block hash. It's a big change though and needs more thought, so leaving for a subsequent PR.

I've changed `_resolve_snapshot_block` into a session fixture.

PR #2644, opened 2026-04-09
Test reservoir refunded at top level when parent halts after child sp…

daniellehrner:
## 🗒️ Description
Adds a new test for EIP-8037 that checks that the state gas reservoir is refunded at top level when parent halts after child spills and reverts

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
@@ -23,6 +23,7 @@ from execution_testing import (
     BlockchainTestFiller,
     Environment,
     Fork,
+    Header,
     Op,
     StateTestFiller,
     Storage,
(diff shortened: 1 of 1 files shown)


tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
```diff
@@ -914,3 +915,101 @@ def test_call_new_account_header_gas_used(
...
+        to=parent,
+        gas_limit=gas_limit_cap,
+        sender=pre.fund_eoa(),
+    )
+
+    # Top-level exceptional halt: gas_left is zeroed, but the reservoir
+    # restored from the child refund is preserved for top-level refund.
+    # block_regular = gas_limit - sstore_state_gas (reservoir kept out)
+    # block_state   = 0 (nothing actually grown)
+    expected_gas_used = gas_limit_cap - sstore_state_gas
```
jochem-brouwer, line 958: I am confused by this (and also by spec). The regular gas here for `Op.SSTORE(0,1)` will cost 2900 (EIP-8037) + 2100 (cold access, EIP-2929) = 5000 gas, and the state gas is `32 * 1174 = 37568`. If this is fully consumed by reservoir (and thus not eating into regular gas) then we now get "free refunds" it seems. If I would spam storage creation and then invalidate it in the end I think I could even end up with negative gas here (?)

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
```diff
@@ -914,3 +915,101 @@ def test_call_new_account_header_gas_used(
...
+    reservoir.
+    """
+    gas_limit_cap = fork.transaction_gas_limit_cap()
+    assert gas_limit_cap is not None
+    sstore_state_gas = fork.sstore_state_gas()
+
+    child = pre.deploy_contract(code=(Op.SSTORE(0, 1) + Op.REVERT(0, 0)))
+
+    parent = pre.deploy_contract(
+        code=(Op.POP(Op.CALL(gas=500_000, address=child)) + Op.INVALID),
```
jochem-brouwer, line 945: Since the child frame reverts it will not consume all gas so the `gas=` param is technically not necessary here (we can send all gas). 

Also it seems that this test is almost equal to the other test, the only difference seems if either REVERT/INVALID is used in the child contract, I think we should parametrize that.

jochem-brouwer (commented):
Left some thoughts, mainly confused by the spec 😅

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
jochem-brouwer: Never mind, I misunderstood the EIP

jochem-brouwer: No wait, in this test the reservoir is initialized to 0, right? So there is no storage gas available, it will all directly eat in regular gas.

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
```diff
@@ -914,3 +915,101 @@ def test_call_new_account_header_gas_used(
...
+    )
+
+    tx = Transaction(
+        to=parent,
+        gas_limit=gas_limit_cap,
```
jochem-brouwer, line 950:
```suggestion
        gas_limit=gas_limit_cap + sstore_state_gas,
```

This inits the reservoir to have state gas. 
Note: this also means it can at most refund `sstore_state_gas`, so this might not cover all cases, I think we should add multiple gas limits, where we use `sstore_state_gas - 1` and `sstore_state_gas + 1` as extra gas on top of the max gas limit to take into account off-by-one errors.

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
```diff
@@ -914,3 +915,101 @@ def test_call_new_account_header_gas_used(
...
+
+    child = pre.deploy_contract(code=(Op.SSTORE(0, 1) + Op.INVALID))
+
+    parent = pre.deploy_contract(
+        code=(Op.POP(Op.CALL(gas=500_000, address=child)) + Op.INVALID),
+    )
+
+    tx = Transaction(
+        to=parent,
+        gas_limit=gas_limit_cap,
```
jochem-brouwer, line 1000: Gas limit should be changed here too to make gas from reservoir available!

spencer-tb:
CIFO: https://github.com/ethereum/execution-specs/pull/2644

Incorperated your comments @jochem-brouwer. Will add you as co-author @daniellehrner before merging! Thx

PR #2646, opened 2026-04-09
feat(tests): EIP-8037 - CALL with value to selfdestructed account

kclowes:
## 🗒️ Description

Adds test coverage for CALL with value to an account that was selfdestructed (per EIP-6780), both the same transaction case and the pre existing case. No spec change: EIP-8037 and EIP-6780 together already produce the correct behavior, and this PR documents it with strict `header_verify` discriminators.

Related EIP spec clarification: ethereum/EIPs#11532 (point 5).

### Behavior under test

When an account is created and then selfdestructed in the same transaction, EIP-6780 defers deletion to the end of the transaction. During the transaction the account still has `nonce = 1` from the CREATE, so it is neither empty nor nonexistent and the new account creation gate does not fire on a subsequent CALL with value, no `GAS_NEW_ACCOUNT` state gas is charged. End of the transaction destruction still removes the account regardless of any value received after SELFDESTRUCT, so value transferred by the CALL is burned along with the destroyed account. Combined with the same transaction SELFDESTRUCT state gas refund (ethereum/execution-specs#2707), the net state gas across the `CREATE + SELFDESTRUCT + CALL(value)` lifecycle is zero.

For a pre existing contract that runs SELFDESTRUCT, EIP-6780 does not queue it for destruction, so a subsequent CALL sees an existing, code carrying account and the new account creation gate does not fire either.

### Tests

All in `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py`. The `init_code_at_high_bytes` helper is moved to `spec.py` for reuse with the related PR ethereum/execution-specs#2707.

- `test_call_value_to_self_destructed_same_tx_account`, smoke test parametrized `create_opcode=[CREATE, CREATE2]`. Confirms the happy path runs to completion. Strict discrimination lives in `test_call_value_to_self_destructed_header_gas_used`.
- `test_call_value_to_self_destructed_header_gas_used`, strict `header_verify` discriminator parametrized `create_opcode=[CREATE, CREATE2]` × `selfdestruct_beneficiary=[self, external]`. Block header `gas_used` equals exactly one `GAS_NEW_ACCOUNT` (only the CREATE's state charge), proving no second charge from the CALL.
- `test_call_value_to_self_destructed_burns_value`, strict `header_verify` parametrized `create_opcode=[CREATE, CREATE2]` × `call_value=[1 wei, 1 ether]`. Block header gas used equals one `GAS_NEW_ACCOUNT`. Post state asserts the created address is `NONEXISTENT` and the orchestrator balance is zero, proving the value is burned when end of the transaction destruction runs.
- `test_call_zero_value_to_self_destructed_same_tx_account`, strict `header_verify` parametrized `create_opcode=[CREATE, CREATE2]`. Value transfer gates the new account charge, so a zero value CALL never triggers it. Block header equals one `GAS_NEW_ACCOUNT`.
- `test_call_value_to_pre_existing_selfdestructed_account`, strict `header_verify` parametrized `beneficiary_type=[eoa, contract]`. Pre deployed contract runs SELFDESTRUCT, then the orchestrator does a value bearing CALL and a sequence of cold SSTORE probes that make block state gas dominate. Block header gas used equals `num_probes * sstore_state_gas` exactly. A spurious new account charge on the value bearing CALL would push the header up by that charge.

## 🔗 Related Issues or PRs

- EIPs PR: ethereum/EIPs#11532

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
@@ -23,13 +23,17 @@ from execution_testing import (
     BlockchainTestFiller,
     Environment,
     Fork,
+    Header,
     Op,
     StateTestFiller,
     Storage,
     Transaction,
+    compute_create2_address,
+    compute_create_address,
 )
+from execution_testing.checklists import EIPChecklist
 
-from .spec import ref_spec_8037
+from .spec import init_code_at_high_bytes, ref_spec_8037
 
 REFERENCE_SPEC_GIT_PATH = ref_spec_8037.git_path
 REFERENCE_SPEC_VERSION = ref_spec_8037.version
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
LGTM!

PR #2648, opened 2026-04-09
feat(tooling,test-cli): add just checklist and `allowed_exit_codes` to `PytestCommand`

kclowes:
## 🗒️ Description
- Adds a command for checklist generation: `just checklist`,
- Extracts exit-code tolerance into a `ClassVar[List[pytest.ExitCode]]` on `PytestCommand`.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,4 +92,4 @@ env.yaml
 site/
 
 # Generated checklists
-docs/checklist/
+tmp/
--- a/.gitignore
+++ b/.gitignore
@@ -91,5 +91,5 @@ env.yaml
 
 site/
 
-# Generated checklists
+# Temporary data; used for generated checklists
 tmp/
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ logs/
 env.yaml
 
 site/
+
+# Generated checklists
+docs/checklist/
--- a/Justfile
+++ b/Justfile
@@ -263,6 +263,17 @@ bench-opcode-config *args:
         "$@" \
         tests/benchmark/compute
 
+# --- Checklists ---
+
+# Generate EIP test checklists from eip_checklist markers
+[group('checklists')]
+checklist *args:
+    #!/usr/bin/env bash
+    uv run checklist --output docs/checklist "$@"
+    rc=$?
+    # exit code 5 = "no tests ran" which is expected (collect-only)
+    if [ $rc -eq 5 ] || [ $rc -eq 0 ]; then exit 0; else exit $rc; fi
+
 # --- Docs ---
 
 # Generate documentation for EELS using docc
--- a/Justfile
+++ b/Justfile
@@ -94,6 +94,8 @@ lint-actions:
 coverage:
     uv run coverage html -d "{{ output_dir }}/fill/coverage-html"
 
+# Generate EIP test checklists from eip_checklist markers                                                                         
+[group('consensus tests')] 
 checklist *args:
     uv run checklist --output tmp/checklist "$@"
 
--- a/Justfile
+++ b/Justfile
@@ -94,6 +94,9 @@ lint-actions:
 coverage:
     uv run coverage html -d "{{ output_dir }}/fill/coverage-html"
 
+checklist *args:
+    uv run checklist --output tmp/checklist "$@"
+
 # --- Fill Tests ---
 
 # Fill the consensus tests using EELS (with Python)
@@ -262,17 +265,6 @@ bench-opcode-config *args:
         --clean \
         "$@" \
         tests/benchmark/compute
-
-# --- Checklists ---
-
-# Generate EIP test checklists from eip_checklist markers
-[group('checklists')]
-checklist *args:
-    #!/usr/bin/env bash
-    uv run checklist --output docs/checklist "$@"
-    rc=$?
-    # exit code 5 = "no tests ran" which is expected (collect-only)
-    if [ $rc -eq 5 ] || [ $rc -eq 0 ]; then exit 0; else exit $rc; fi
 
 # --- Docs ---
 
--- a/Justfile
+++ b/Justfile
@@ -89,6 +89,8 @@ lock-check:
 lint-actions:
     uv run actionlint -pyflakes pyflakes -shellcheck "shellcheck -S warning"
 
+# --- Consensus Tests ---
+
 # Generate HTML coverage report from last just fill run
 [group('consensus tests')]
 coverage:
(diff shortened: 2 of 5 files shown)


Justfile
```diff
@@ -263,6 +263,17 @@ bench-opcode-config *args:
         "$@" \
         tests/benchmark/compute
 
+# --- Checklists ---
+
+# Generate EIP test checklists from eip_checklist markers
+[group('checklists')]
```
danceratopz: Perhaps it's easier to find here? While not a test, it's closely related to testing ))
```suggestion
[group('consensus tests')]
```

Justfile
```diff
@@ -263,6 +263,17 @@ bench-opcode-config *args:
...
+checklist *args:
+    #!/usr/bin/env bash
+    uv run checklist --output docs/checklist "$@"
+    rc=$?
+    # exit code 5 = "no tests ran" which is expected (collect-only)
```
danceratopz: I wonder if it would be bad form (or even possible) to "correct" pytest's exit code within the plugin itself to avoid this boilerplate code? Not important, just a thought.

Justfile
```diff
@@ -263,6 +263,17 @@ bench-opcode-config *args:
         "$@" \
         tests/benchmark/compute
 
+# --- Checklists ---
+
+# Generate EIP test checklists from eip_checklist markers
+[group('checklists')]
+checklist *args:
+    #!/usr/bin/env bash
+    uv run checklist --output docs/checklist "$@"
```
danceratopz: Sorry, I know we discussed this, but as the docs generate this dynamically (see below) without the use of the `uv run checklist` command (and this folder), I think it's a bit misleading to write them by default underneath `./docs/`?

https://github.com/ethereum/execution-specs/blob/fa12e9978edd56a5e067e6cc58537b2dfc61f2df/docs/scripts/gen_test_case_reference.py#L59

How about?
```suggestion
    uv run checklist --output tmp/checklist "$@"
```
And .gitignoring `tmp`? Btw, `checklist` will create `tmp` if it doesn't exist.

Another option (that other recipes use) would be `.just/checklist/`. But putting them into a hidden folder is probably not great ux!

.gitignore
```diff
@@ -90,3 +90,6 @@ logs/
 env.yaml
 
 site/
+
+# Generated checklists
+docs/checklist/
```
danceratopz: Might need to be updated based on the decision below.

danceratopz (commented):
Thanks @kclowes!

Justfile
kclowes: I like `tmp/checklist`. Also communicates that they're ephemeral. Will update!

Justfile
kclowes: I'll see what I can do!

kclowes:
Updated and ready for re-review @danceratopz. Thanks!

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
```diff
@@ -1,13 +1,31 @@
...
+        Execute the command, treating pytest exit code 5 as success.
+
+        The checklist command only collects tests to analyze markers
+        and does not run them, so exit code 5 ("no tests ran") is expected.
+        """
+        executions = self.create_executions(pytest_args)
+        result = self.runner.run_multiple(executions)
+        if result == 5:
+            result = 0
+        sys.exit(result)
```
danceratopz: Yes, this is the right place to patch this in, but I think we can be more explicit and return a new `PytestExecution`. Indeed, `fill` already does this and even allows `pytest.ExitCode.NO_TESTS_COLLECTED`! I must admit, I'd completely forgotten this.

https://github.com/ethereum/execution-specs/blob/a48e0b381d5225a6c3de2d06cd9ee7ae0b6ca9bb/packages/testing/src/execution_testing/cli/pytest_commands/fill.py#L80-L86

One way could be to allow passing `allowed_exit_codes` to `create_executions()`?
https://github.com/ethereum/execution-specs/blob/a48e0b381d5225a6c3de2d06cd9ee7ae0b6ca9bb/packages/testing/src/execution_testing/cli/pytest_commands/fill.py#L32-L34

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
```diff
@@ -1,13 +1,31 @@
...
+from typing import Any, List
 
 import click
 
 from ...forks import get_development_forks
 from .fill import FillCommand
 
 
+class ChecklistCommand(FillCommand):
+    """Fill command that treats 'no tests ran' as success."""
```
danceratopz:
```suggestion
    """Checklist command that treats 'no tests ran' as success."""
```

.gitignore
```diff
@@ -90,3 +90,6 @@ logs/
 env.yaml
 
 site/
+
+# Generated checklists
```
danceratopz:
```suggestion
# Temporary data; used for generated checklists
```

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
```diff
@@ -1,13 +1,31 @@
...
+import sys
+from typing import Any, List
 
 import click
 
 from ...forks import get_development_forks
 from .fill import FillCommand
 
 
+class ChecklistCommand(FillCommand):
```
danceratopz: I wonder if we should inherit from `PytestCommand`? I don't know, please check. It could be that we need it as we need quite a lot of the `fill` collection logic.

danceratopz (commented):
Thanks @kclowes! I think we can keep it more pytestic (idiomatic pytest 😆) and more consistent with `fill`, please see my comment below.

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
kclowes: We sure can! I don't think we need any of the logic from fill.

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
kclowes: I updated to use a ClassVar here in the last commit, lmk what you think!

danceratopz: Nice, this looks great!

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
danceratopz: Ah cool, that's right! We grab the collection logic from fill's ini file, not from its CLI class.

danceratopz (approved):
Thanks @kclowes, this turned out nice! 

The checklist recipe lost it's `# ...` help comment and `[group('...')]` annotation, so I just added them back quickly to get this merged.

PR #2652, opened 2026-04-10
tests(amsterdam): add BAL missing withdrawal account tests

edg-l:
## Summary

Add tests verifying that clients reject blocks where the BAL (EIP-7928) omits an account that was modified by a withdrawal (EIP-4895).

The existing `test_bal_invalid_missing_account` only covers accounts modified by transactions. These new tests cover the withdrawal case: the BAL must include every account whose state changed during the withdrawal/request phase, not just the accounts it claims to cover.

**New tests:**
- `test_bal_invalid_missing_withdrawal_account` - block with 1 tx + 1 withdrawal, BAL missing the withdrawal recipient
- `test_bal_invalid_missing_withdrawal_account_empty_block` - empty block with 1 withdrawal, BAL missing the recipient

## Test plan

- [x] All new tests fill successfully
- [x] All 86 tests in `test_block_access_lists_invalid.py` pass (no regressions)
- [x] `ruff check` and `ruff format --check` pass

### DIFF
--- a/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_invalid.py
+++ b/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_invalid.py
@@ -25,6 +25,7 @@ from execution_testing import (
     Op,
     Storage,
     Transaction,
+    Withdrawal,
     compute_create_address,
 )
 from execution_testing.test_types.block_access_list.modifiers import (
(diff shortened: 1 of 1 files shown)


fselmo (approved):
Thanks for this! This is great coverage to have. I will iterate on some more related test cases 👍🏼

PR #2657, opened 2026-04-11
feat(tests): add test for worst-case prefetcher performance

nerolation:
## 🗒️ Description

Add `test_sload_bloated_prefetch_miss` benchmark that defeats client prefetchers by making each transaction's SLOAD range depend on state written by its predecessor. A minimal first tx writes an offset to slot 0 via calldata; subsequent max-gas txs read the previous offset, SLOAD from it, and write a new offset for the next tx.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

  ```console
  just static
#### Cute Animal Picture

<img width="320" alt="image" src="https://github.com/user-attachments/assets/39996198-384d-4526-99d6-19b3ea05cdc8" />

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -75,6 +75,22 @@ def _max_sloads_per_tx(tx_gas_limit: int, fork: Fork) -> int:
     return tx_gas_limit // cold_sload_cost
 
 
+def _sender_generator(
+    pre: Alloc, distinct_senders: bool
+) -> Generator[EOA, None, None]:
+    """
+    Yield one sender per tx.
+
+    In distinct mode, yields a fresh EOA per call. Otherwise, yields
+    the same shared sender for every call. Used by the bloated SLOAD
+    benchmarks so the BAL builder can group nonce changes by sender
+    uniformly regardless of mode.
+    """
+    sender = pre.fund_eoa()
+    while True:
+        yield sender if not distinct_senders else pre.fund_eoa()
+
+
 def delegate_with_calldata(
     pre: Alloc,
     authority: EOA,
@@ -204,7 +220,6 @@ def test_sload_bloated(
     )
 
 
-@pytest.mark.repricing
 @pytest.mark.stub_parametrize("token_name", "bloated_eoa_")
 @pytest.mark.parametrize("distinct_senders", [False, True])
 @pytest.mark.parametrize("existing_slots", [False, True])
@@ -278,18 +293,11 @@ def test_sload_bloated_prefetch_miss(
         calldata=b"\xff" * 32,
     )
 
-    # Yield one sender per tx via a generator. In distinct mode
-    # each call returns a fresh EOA; otherwise the same shared
-    # sender is reused. The senders list collects one entry per
-    # tx so the BAL builder below can group nonce changes by
-    # sender uniformly.
-    def sender_generator() -> Generator[EOA, None, None]:
-        """Yield senders, fresh if distinct, else reuse one shared."""
-        sender = pre.fund_eoa()
-        while True:
-            yield sender if not distinct_senders else pre.fund_eoa()
-
-    senders_iter = sender_generator()
+    # senders_iter yields one sender per tx (fresh per call in
+    # distinct mode, a single shared sender otherwise). The senders
+    # list collects one entry per tx so the BAL builder below can
+    # group nonce changes by sender uniformly.
+    senders_iter = _sender_generator(pre, distinct_senders)
     senders: list[EOA] = []
 
     gas_available = gas_benchmark_value
@@ -373,7 +381,6 @@ def test_sload_bloated_prefetch_miss(
     )
 
 
-@pytest.mark.repricing
 @pytest.mark.parametrize("distinct_senders", [False, True])
 @pytest.mark.parametrize("existing_slots", [False, True])
 def test_sload_bloated_multi_contract(
@@ -435,7 +442,7 @@ def test_sload_bloated_multi_contract(
     )
 
     base_offset = 1 if existing_slots else START_SLOT
-    max_sloads_per_tx = _max_sloads_per_tx(tx_gas_limit)
+    max_sloads_per_tx = _max_sloads_per_tx(tx_gas_limit, fork)
 
     # Pre-load slot 0 with the starting offset. For existing_slots,
     # also fill the slot range the loop will read so SLOADs land on
@@ -454,18 +461,11 @@ def test_sload_bloated_multi_contract(
     # + one iteration + final SSTORE, with buffer.
     min_tx_gas = intrinsic_gas + 130_000
 
-    # Yield one sender per tx via a generator. In distinct mode
-    # each call returns a fresh EOA; otherwise the same shared
-    # sender is reused. The senders list collects one entry per
-    # tx so the BAL builder below can group nonce changes by
-    # sender uniformly.
-    def sender_generator() -> Generator[EOA, None, None]:
-        """Yield senders, fresh if distinct, else reuse one shared."""
-        sender = pre.fund_eoa()
-        while True:
-            yield sender if not distinct_senders else pre.fund_eoa()
-
-    senders_iter = sender_generator()
+    # senders_iter yields one sender per tx (fresh per call in
+    # distinct mode, a single shared sender otherwise). The senders
+    # list collects one entry per tx so the BAL builder below can
+    # group nonce changes by sender uniformly.
+    senders_iter = _sender_generator(pre, distinct_senders)
     senders: list[EOA] = []
 
     gas_available = gas_benchmark_value
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -9,7 +9,7 @@ abstract: BloatNet single-opcode benchmark cases for state-related operations.
 
 from enum import Enum, auto
 from functools import partial
-from typing import Callable, List
+from typing import Callable, Generator, List
 
 import pytest
 from execution_testing import (
@@ -60,6 +60,18 @@ START_SLOT = (
     0xA4896A3F93BF4BF58378E579F3CF193BB4AF1022AF7D2089F37D8BAE7157B85F
     % (2**160)
 )
+
+
+def _max_sloads_per_tx(tx_gas_limit: int) -> int:
+    """
+    Conservative upper bound on cold SLOADs that fit in a max-gas tx.
+
+    Derived from the cold SLOAD cost (EIP-2929: 2100 gas) and used by
+    the bloated SLOAD benchmarks both as the inter-tx offset stride
+    (to keep consecutive txs' SLOAD ranges disjoint) and as the
+    per-target storage pre-load count.
+    """
+    return tx_gas_limit // 2100
 
 
 def delegate_with_calldata(
@@ -216,7 +228,10 @@ def test_sload_bloated_prefetch_miss(
     SLOAD range depends on state written by its predecessor, a
     prefetcher that predicts SLOAD targets from pre-block state
     without simulating intra-block writes will pre-warm incorrect
-    storage slots.
+    storage slots. The minimal first tx is load-bearing: it lives
+    inside the benchmark block so every subsequent max-gas tx reads
+    a slot 0 value that differs from the prefetcher's pre-block
+    snapshot, achieving a 100% miss rate.
 
     When ``distinct_senders`` is True every transaction uses a fresh
     sender. This additionally defeats per-sender prewarm
@@ -238,7 +253,12 @@ def test_sload_bloated_prefetch_miss(
     authority = pre.stub_eoa(token_name)
     runtime_address = pre.deploy_contract(code=runtime_code)
 
-    # Setup: delegate authority to the runtime contract.
+    # Setup: delegate authority to the runtime contract. Slot 0 is
+    # left at 0 (the delegation tx's calldata) so the benchmark
+    # block's pre-state has slot 0 = 0; the first benchmark tx
+    # then plants base_offset in slot 0 inside the benchmark block,
+    # forcing the prefetcher's pre-block snapshot to disagree with
+    # the actual slot 0 value seen by every max-gas tx that follows.
     delegation_tx = delegate_with_calldata(
         pre, authority, runtime_address, Hash(0)
     )
@@ -247,39 +267,49 @@ def test_sload_bloated_prefetch_miss(
 
     # Offset spacing: upper bound on SLOADs per tx ensures each
     # transaction reads a completely disjoint slot range.
-    sload_spacing = tx_gas_limit // 2100
+    max_sloads_per_tx = _max_sloads_per_tx(tx_gas_limit)
 
-    # The base offset must be at least sload_spacing away from the
-    # pre-block slot 0 value (0) so the prefetcher's predicted
+    # The base offset must be at least max_sloads_per_tx away from
+    # the pre-block slot 0 value (0) so the prefetcher's predicted
     # SLOAD range is completely disjoint from the actual range.
-    base_offset = sload_spacing if existing_slots else START_SLOT
+    base_offset = max_sloads_per_tx if existing_slots else START_SLOT
     intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(
         calldata=b"\xff" * 32,
     )
 
-    # For distinct_senders, append a fresh EOA per tx. Otherwise
-    # pre-seed the list with a single shared sender reused by all.
-    senders: list[EOA] = [] if distinct_senders else [pre.fund_eoa()]
+    # Yield one sender per tx via a generator. In distinct mode
+    # each call returns a fresh EOA; otherwise the same shared
+    # sender is reused. The senders list collects one entry per
+    # tx so the BAL builder below can group nonce changes by
+    # sender uniformly.
+    def sender_generator() -> Generator[EOA, None, None]:
+        """Yield senders, fresh if distinct, else reuse one shared."""
+        sender = pre.fund_eoa()
+        while True:
+            yield sender if not distinct_senders else pre.fund_eoa()
 
-    def next_sender() -> EOA:
-        if distinct_senders:
-            senders.append(pre.fund_eoa())
-            return senders[-1]
-        return senders[0]
+    senders_iter = sender_generator()
+    senders: list[EOA] = []
 
     gas_available = gas_benchmark_value
     txs: list[Transaction] = []
 
     # First transaction: minimal gas, only writes the initial
     # offset. Gas limit ensures remaining gas after the SLOAD +
-    # SSTORE setup falls below the 0xFFFF loop threshold.
+    # SSTORE setup falls below the 0xFFFF loop threshold so the
+    # SLOAD loop does not run. This tx's job is to change slot 0
+    # inside the benchmark block so every subsequent max-gas tx
+    # reads an offset the prefetcher's pre-block snapshot does
+    # not see, achieving a 100% prefetch miss rate on max-gas txs.
     first_tx_gas = min(gas_available, intrinsic_gas + 30_000)
+    sender = next(senders_iter)
+    senders.append(sender)
     txs.append(
         Transaction(
             gas_limit=first_tx_gas,
             to=authority,
             data=Hash(base_offset),
-            sender=next_sender(),
+            sender=sender,
         )
     )
     gas_available -= first_tx_gas
@@ -289,14 +319,14 @@ def test_sload_bloated_prefetch_miss(
     tx_index = 1
     while gas_available >= intrinsic_gas:
         tx_gas = min(gas_available, tx_gas_limit)
-        new_offset = base_offset + tx_index * sload_spacing
+        new_offset = base_offset + tx_index * max_sloads_per_tx
+        sender = next(senders_iter)
+        senders.append(sender)
         txs.append(
             Transaction(
                 gas_limit=tx_gas,
                 to=authority,
                 data=Hash(new_offset),
-                sender = next(sender_generator)
-                sender_list.append(sender)
                 sender=sender,
             )
         )
@@ -314,17 +344,17 @@ def test_sload_bloated_prefetch_miss(
             ],
         ),
     }
- sender_nonce: dict[Address, list[BalNonceChange]] = {}
-  for i, s in enumerate(senders):             
-      changes = sender_nonces.setdefault(s, [])
-      changes.append(
-          BalNonceChange(                                                                                      
-              block_access_index=i + 1,
-              post_nonce=len(changes) + 1,                                                                     
-          )                                             
-      )                                   
-  for s, nonces in sender_nonces.items():
-      expectations[s] = BalAccountExpectation(nonce_changes=nonces)
+    sender_nonces: dict[Address, list[BalNonceChange]] = {}
+    for i, s in enumerate(senders):
+        changes = sender_nonces.setdefault(s, [])
+        changes.append(
+            BalNonceChange(
+                block_access_index=i + 1,
+                post_nonce=len(changes) + 1,
+            )
+        )
+    for addr, nonces in sender_nonces.items():
+        expectations[addr] = BalAccountExpectation(nonce_changes=nonces)
     blocks.append(
         Block(
             txs=txs,
@@ -404,7 +434,7 @@ def test_sload_bloated_multi_contract(
     )
 
     base_offset = 1 if existing_slots else START_SLOT
-    sloads_per_tx = tx_gas_limit // 2100
+    max_sloads_per_tx = _max_sloads_per_tx(tx_gas_limit)
 
     # Pre-load slot 0 with the starting offset. For existing_slots,
     # also fill the slot range the loop will read so SLOADs land on
@@ -413,7 +443,7 @@ def test_sload_bloated_multi_contract(
     # gets an independent root dict, not an alias of the same one.
     storage_data: Storage.StorageDictType = {0: base_offset}
     if existing_slots:
-        for i in range(base_offset, base_offset + sloads_per_tx):
+        for i in range(base_offset, base_offset + max_sloads_per_tx):
             storage_data[i] = i
 
     intrinsic_gas = fork.transaction_intrinsic_cost_calculator()()
@@ -423,15 +453,19 @@ def test_sload_bloated_multi_contract(
     # + one iteration + final SSTORE, with buffer.
     min_tx_gas = intrinsic_gas + 130_000
 
-    # For distinct_senders, append a fresh EOA per tx. Otherwise
-    # pre-seed the list with a single shared sender reused by all.
-    senders: list[EOA] = [] if distinct_senders else [pre.fund_eoa()]
+    # Yield one sender per tx via a generator. In distinct mode
+    # each call returns a fresh EOA; otherwise the same shared
+    # sender is reused. The senders list collects one entry per
+    # tx so the BAL builder below can group nonce changes by
+    # sender uniformly.
+    def sender_generator() -> Generator[EOA, None, None]:
+        """Yield senders, fresh if distinct, else reuse one shared."""
+        sender = pre.fund_eoa()
+        while True:
+            yield sender if not distinct_senders else pre.fund_eoa()
 
-    def next_sender() -> EOA:
-        if distinct_senders:
-            senders.append(pre.fund_eoa())
-            return senders[-1]
-        return senders[0]
+    senders_iter = sender_generator()
+    senders: list[EOA] = []
 
     gas_available = gas_benchmark_value
     targets: list[Address] = []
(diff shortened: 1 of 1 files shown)


tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +187,110 @@ def test_sload_bloated(
...
+    analyzes transactions before execution will pre-warm incorrect
+    storage slots.
+    """
+    # Runtime: read old offset from slot 0, write new offset from
+    # calldata to slot 0, then SLOAD sequentially from old offset.
+    runtime_code = (
+        Op.SLOAD(Op.PUSH0)
+        + Op.CALLDATALOAD(0)
+        + Op.PUSH0
+        + Op.SSTORE
```
jochem-brouwer: This unconditially SSTOREs the calldata pointer. The first and also the txs after thus have store calldata as pointer and then do the SLOAD loop.

I think that there should be a conditional SSTORE set which only SSTOREs if SLOAD(0) equals zero.

In a setup block, delegate the code first to reset `SSTORE(0, 0)`. Then re-delegate the code to the code here (with the conditional SLOAD(0)). 

I just noticed that `delegate_and_set_slot0` is misleading, as it will delegate and insert calldata but it will depend on the code it is delegated to what it does (so if it sets slot 0). A better name would be `delegate_and_set_calldata` where the 4th arg is the calldata.

For the tx to delegate to the runtime code here, we can take out the authority gas to leave ~5 sload slots extra and not spend the 12.5k gas on the authority in the attack/bench block:

If we do this in the setup block we delegate to the runtime code (but we should not call the runtime code -> this thus needs a change of logic in the delegate_and_set_calldata method where we only delegate and do not call it) and then in bench block do the first tx where we set the first storage slot (either to 1 if existing_slots=True else a random (constant) slot with high offset) and then do the other full txs (when ran in parallel without BAL it will write to 0 the calldata and then thus SLOAD the incrementing range from the provided calldata).

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +187,110 @@ def test_sload_bloated(
...
+        + While(
+            body=(Op.DUP1 + Op.SLOAD + Op.POP + Op.PUSH1(1) + Op.ADD),
+            condition=Op.GT(Op.GAS, 0xFFFF),
+        )
```
jochem-brouwer, line 266: At the end of the while loop this should re-store the current storage pointer such that the next tx can start from here (note that when ran in parallel without BAL it should thus  all read slot 0 and therefore start at the provided calldata, while in sequential it will incrementally read slots as we will not write the pointer if slot 0 is nonzero) and instead just read this pointer

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +187,110 @@ def test_sload_bloated(
...
+        tx_index += 1
+
+    blocks.append(Block(txs=txs))
+
+    benchmark_test(
+        pre=pre,
+        blocks=blocks,
+        skip_gas_used_validation=True,
+        expected_receipt_status=True,
+    )
```
jochem-brouwer, line 381: For validation we could insert expected BAL items for Glamsterdam to see if bench works as expected

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +187,110 @@ def test_sload_bloated(
...
+
+
```
jochem-brouwer, line 383: I think the SSTORE variant here is also interesting, we could use this test as a template which tricks the prefetcher by overloading it with SLOADs which are not used in practice. For the SSTORE variant this could somewhat bottleneck reading trie nodes. (Can be added here or in another PR)

jochem-brouwer (commented):
Some comments 😄 👍

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: The conditional approach adds bytecode complexity (ISZERO + JUMPI branching) for no practical gain in prefetch-miss behavior. Both designs defeat the prefetcher equally. 
Is there a reason why you'd want it that I'm missing?

On the rename of delegate_and_set_slot0, agreed it's a better name.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: With the unconditional SSTORE, each tx's starting offset is determined by what the previous tx wrote from calldata, so no counter write-back is needed. The disjoint ranges are guaranteed by sload_spacing > actual_sloads_per_tx.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: We would need to change the logic to be able to construct the full BAL, right? Do something like `plan_benchmark()`, moving away from the `test_sload_bloated` logic?
I can add the storage_change for the slot 0 and the nonce change, but since we cannot compute exact iteration counts, we cannot list storage_reads.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: Yeah agree, let's do it in another PR to not block things. Feel free to proceed with using this PR in a follow-up PR.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
jochem-brouwer: You are right, this approach works and indeed costs less gas!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
jochem-brouwer: Yes you are right, this is a better approach as it costs less overhead gas 😄 👍

tests/benchmark/stateful/bloatnet/test_single_opcode.py
jochem-brouwer: Yes we indeed don't need to construct the full BAL, we just want to verify specific things like you are checking here, great! I think we should also add a storage read for the tx after the setup tx, to verify it reads the slot (it should read `base_offset`) (so tx1 in zero-based tx list)

jochem-brouwer (commented):
I somehow did not see that this approach solves the same problem but with less gas (I did not realize this also tricks the prefetcher). One final comment for verification.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: Good call. Added storage_reads=[base_offset] on the authority account.

jochem-brouwer (approved):
LGTM! Thanks a lot!!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,137 @@ def test_sload_bloated(
...
+    Benchmark SLOAD with calldata-driven offsets to defeat prefetching.
+
+    A small first transaction writes an initial offset into the
+    authority's slot 0 via calldata. Subsequent max-gas transactions
+    each read the previous offset from slot 0, SLOAD sequentially
+    from that offset, and overwrite slot 0 with a new offset from
+    their own calldata. Because each transaction's SLOAD range
+    depends on state written by its predecessor, a prefetcher that
+    analyzes transactions before execution will pre-warm incorrect
+    storage slots.
```
CPerezz: The docstring's ordering contradicts the actual execution order. It currently reads:
> Subsequent max-gas transactions each read the previous offset from slot 0, SLOAD sequentially from that offset, and overwrite slot 0 with a new offset from their own calldata.

Read in order, this implies SSTORE happens after the SLOAD loop. But the bytecode is:
```
  Op.SLOAD(Op.PUSH0)      # push slot[0] (old offset)
  + Op.CALLDATALOAD(0)    # push new offset from calldata
  + Op.PUSH0 + Op.SSTORE  # slot[0] := new offset   <-- happens BEFORE the loop
  + While(body=DUP1 + SLOAD + POP + PUSH1(1) + ADD, ...)
```

So slot 0 is overwritten before the SLOAD loop runs — the inline comment at the line right above runtime_code already has this right ("read old offset from slot 0, write new offset from calldata to slot 0, then SLOAD sequentially from old offset"). The docstring just needs to match.

While we're there, the claim that this defeats "a prefetcher that analyzes transactions before execution" is slightly overstated — a prefetcher that simulates intra-block writes could still chain the offsets forward, since each tx's calldata is the next slot-0 value. What the test actually defeats is any prefetcher cheaper than full execution (i.e., one that reads pre-block state without propagating SSTOREs across txs), which is the practically relevant case. Worth being explicit about the threat model.

Suggested rewrite:

> A small first transaction writes an initial offset into the authority's slot 0 via calldata. Subsequent max-gas transactions each read the previous offset from slot 0, immediately overwrite slot 0 with a new offset from their own calldata, then SLOAD sequentially from the previous offset. Because each transaction's SLOAD range depends on state written by its predecessor, a prefetcher that predicts SLOAD targets from pre-block state without simulating intra-block writes will pre-warm incorrect storage slots.

jochem-brouwer (changes requested):
I am having troubles filling these payloads and need to investigate more (can´t seem to retract my previous review 🤔 )

EDIT; this worked

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: Agree with both, fixed

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+    runs them sequentially to propagate state changes — forcing
+    every tx's prewarm scope to restart from pre-block state.
+    """
+    # Runtime: read old offset from slot 0, write new offset from
+    # calldata to slot 0, then SLOAD sequentially from old offset.
+    runtime_code = (
+        Op.SLOAD(Op.PUSH0)
+        + Op.CALLDATALOAD(0)
+        + Op.PUSH0
+        + Op.SSTORE
```
LouisTsai-Csie: nit.
```suggestion
        + Op.SSTORE(Op.PUSH0, Op.CALLDATALOAD(Op.PUSH0))
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+    # SSTORE setup falls below the 0xFFFF loop threshold.
+    first_tx_gas = min(gas_available, intrinsic_gas + 30_000)
+    txs.append(
+        Transaction(
+            gas_limit=first_tx_gas,
+            to=authority,
+            data=Hash(base_offset),
+            sender=next_sender(),
+        )
+    )
```
LouisTsai-Csie, line 323: Could we initialize storage within the `delegation_tx` instead of using a separate tx? Similar to `run_bloated_eoa_benchmark`, where `init_tx` handles both code delegation and storage setup in a single auth transaction.

Same question to `test_sload_bloated_multi_contract`

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+
+    # For distinct_senders, append a fresh EOA per tx. Otherwise
+    # pre-seed the list with a single shared sender reused by all.
+    senders: list[EOA] = [] if distinct_senders else [pre.fund_eoa()]
+
+    def next_sender() -> EOA:
+        if distinct_senders:
+            senders.append(pre.fund_eoa())
+            return senders[-1]
+        return senders[0]
```
LouisTsai-Csie: We could move out this `next_sender` function, so that both cases could re-use it. And change it to a generator like this, and reference this in the `sender` field in the transaction

```python
def sender_generator(
     pre: Alloc, distinct: bool
 ) -> Generator[EOA, None, None]:
     """Yield funded senders, fresh each time if distinct, else reuse one."""
     sender = pre.fund_eoa()
     while True:
         yield sender if not distinct else pre.fund_eoa()
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+    tx_index = 1
+    while gas_available >= intrinsic_gas:
+        tx_gas = min(gas_available, tx_gas_limit)
+        new_offset = base_offset + tx_index * sload_spacing
+        txs.append(
+            Transaction(
+                gas_limit=tx_gas,
+                to=authority,
+                data=Hash(new_offset),
+                sender=next_sender(),
```
LouisTsai-Csie: corresponding to the changes from the last comment

```suggestion
                sender = next(sender_generator)
                sender_list.append(sender)
                sender=sender,
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+    else:
+        expectations[senders[0]] = BalAccountExpectation(
+            nonce_changes=[
+                BalNonceChange(
+                    block_access_index=i + 1,
+                    post_nonce=i + 1,
+                )
+                for i in range(len(txs))
+            ],
+        )
```
LouisTsai-Csie: Could we use the following approach, so no if-else condition needed here?


```suggestion
 sender_nonce: dict[Address, list[BalNonceChange]] = {}
  for i, s in enumerate(senders):             
      changes = sender_nonces.setdefault(s, [])
      changes.append(
          BalNonceChange(                                                                                      
              block_access_index=i + 1,
              post_nonce=len(changes) + 1,                                                                     
          )                                             
      )                                   
  for s, nonces in sender_nonces.items():
      expectations[s] = BalAccountExpectation(nonce_changes=nonces)
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+    # storage. Final counter is written back to slot 0.
+    runtime_code = (
+        Op.POP(
+            Op.CALL(
+                address=offset_holder,
+                value=0,
+                args_offset=0,
+                args_size=0,
+                ret_offset=0,
+                ret_size=0,
```
LouisTsai-Csie: I think we do not need this, it would be default value here:

```suggestion
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+
+    # For distinct_senders, append a fresh EOA per tx. Otherwise
+    # pre-seed the list with a single shared sender reused by all.
+    senders: list[EOA] = [] if distinct_senders else [pre.fund_eoa()]
+
+    def next_sender() -> EOA:
+        if distinct_senders:
+            senders.append(pre.fund_eoa())
+            return senders[-1]
+        return senders[0]
```
LouisTsai-Csie: Same suggestion as `test_sload_bloated_prefetch_miss`

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+    else:
+        expectations[senders[0]] = BalAccountExpectation(
+            nonce_changes=[
+                BalNonceChange(
+                    block_access_index=i + 1,
+                    post_nonce=i + 1,
+                )
+                for i in range(len(txs))
+            ],
+        )
```
LouisTsai-Csie: Same suggestion as this comment: https://github.com/ethereum/execution-specs/pull/2657/changes#r3071379137

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+    # Setup: delegate authority to the runtime contract.
+    delegation_tx = delegate_with_calldata(
+        pre, authority, runtime_address, Hash(0)
+    )
+
+    blocks: list[Block] = [Block(txs=[delegation_tx])]
+
+    # Offset spacing: upper bound on SLOADs per tx ensures each
+    # transaction reads a completely disjoint slot range.
+    sload_spacing = tx_gas_limit // 2100
```
LouisTsai-Csie: Could we move this out of this function, so that both tests could share this value? And what about renaming it to `storage_spacing`?

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +191,342 @@ def test_sload_bloated(
...
+
+    base_offset = 1 if existing_slots else START_SLOT
+    sloads_per_tx = tx_gas_limit // 2100
+
+    # Pre-load slot 0 with the starting offset. For existing_slots,
+    # also fill the slot range the loop will read so SLOADs land on
+    # populated entries rather than empty slots. A fresh Storage
+    # instance is built per deployment (below) so that every target
+    # gets an independent root dict, not an alias of the same one.
+    storage_data: dict[int, int] = {0: base_offset}
```
LouisTsai-Csie:
```suggestion
    storage_data: Storage.StorageDictType = {0: base_offset}
```

With this, we could remove the `# type: ignore` in the contract deployment:

```python
target = pre.deploy_contract(
            code=runtime_code,
            storage=Storage(storage_data),  # type: ignore -> remove this
        )
```

LouisTsai-Csie (changes requested):
Leave some style suggestion comment, please take a look!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: Moved and renamed it to max_sloads_per_tx to fit both tests

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: I wasn't sure about this one. Would love @jochem-brouwer to take a look to.

We must make sure that it doesn't defeat the purpose of the test (first small transaction used to trick the prefetcher).

nerolation: So, the first transaction isn't just init but actually an important piece for the core logic

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +203,343 @@ def test_sload_bloated(
...
+        code=Op.SSTORE(0, Op.ADD(Op.SLOAD(0), 1)),
+    )
+
+    # Target runtime: CALL offset_holder (for the dependency), then
+    # run the same SLOAD loop as test_sload_bloated in its own
+    # storage. Final counter is written back to slot 0.
+    runtime_code = (
+        Op.POP(
+            Op.CALL(
+                address=offset_holder,
```
jochem-brouwer: This will call precompiles, we should not do this, it should initialize the initial storage to a non-precompile OR we should limit gas in this Op.CALL to 0 (note that even with gas 0 the client cannot optimize this by not reading the contract, we still need to retrieve the code)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +203,343 @@ def test_sload_bloated(
...
+    targets: list[Address] = []
+    txs: list[Transaction] = []
+
+    # Each tx targets a freshly-deployed contract with identical code
+    # and storage layout.
+    while gas_available >= min_tx_gas:
+        tx_gas = min(gas_available, tx_gas_limit)
+        target = pre.deploy_contract(
+            code=runtime_code,
+            storage=Storage(storage_data),
```
jochem-brouwer, line 481: This will likely have the problem that either this code will not fit in initcode, or the predeploy code to execute will use more than `2**24` gas. @LouisTsai-Csie we should likely create a general helper here for the predeploy in execute

jochem-brouwer:
Here is fixture output of the `sload_bloated_multi_contract`. I'll check this soon, wanted to dump this progress. The good point here is that the fixture fills succesfully so it passes any checks made on the output (BAL checks). Will do similar to the other test also.

[sload_bloated_multi_contract.json](https://github.com/user-attachments/files/26676876/sload_bloated_multi_contract.json)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: I check the codebase comment and our prior discussion, i think it is reasonable to keep it separate, ensuring the warm up miss!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -58,28 +62,40 @@
...
+def _max_sloads_per_tx(tx_gas_limit: int) -> int:
+    """
+    Conservative upper bound on cold SLOADs that fit in a max-gas tx.
+
+    Derived from the cold SLOAD cost (EIP-2929: 2100 gas) and used by
+    the bloated SLOAD benchmarks both as the inter-tx offset stride
+    (to keep consecutive txs' SLOAD ranges disjoint) and as the
+    per-target storage pre-load count.
+    """
+    return tx_gas_limit // 2100
```
LouisTsai-Csie: Instead of using the constant here, we use the `opcode.gas_cost(fork)` approach, so that even the SLOAD gas cost is changed in the future, this value would be automatically updated accordingly.

```suggestion
def _max_sloads_per_tx(tx_gas_limit: int, fork: Fork) -> int:
    """
    Conservative upper bound on cold SLOADs that fit in a max-gas tx.

    Derived from the cold SLOAD cost (EIP-2929: 2100 gas) and used by
    the bloated SLOAD benchmarks both as the inter-tx offset stride
    (to keep consecutive txs' SLOAD ranges disjoint) and as the
    per-target storage pre-load count.
    """
    cold_sload_cost = Op.SLOAD(key_warm=False).gas_cost(fork)
    return tx_gas_limit // cold_sload_cost
```

It would require the following code reference being changed

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +203,343 @@ def test_sload_bloated(
     )
 
 
+@pytest.mark.repricing
```
LouisTsai-Csie:
```suggestion
```

We label `repricing` marker for repricing analysis. For BAL specific benchmark, we could consider:
- Add a new BAL marker
- Move them under `tests/benchmark/stateful/eip7928_block_level_access_lists` (Recommend)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +203,343 @@ def test_sload_bloated(
...
+    # the actual slot 0 value seen by every max-gas tx that follows.
+    delegation_tx = delegate_with_calldata(
+        pre, authority, runtime_address, Hash(0)
+    )
+
+    blocks: list[Block] = [Block(txs=[delegation_tx])]
+
+    # Offset spacing: upper bound on SLOADs per tx ensures each
+    # transaction reads a completely disjoint slot range.
+    max_sloads_per_tx = _max_sloads_per_tx(tx_gas_limit)
```
LouisTsai-Csie:
```suggestion
    max_sloads_per_tx = _max_sloads_per_tx(tx_gas_limit, fork)
```

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +203,343 @@ def test_sload_bloated(
...
+    # Yield one sender per tx via a generator. In distinct mode
+    # each call returns a fresh EOA; otherwise the same shared
+    # sender is reused. The senders list collects one entry per
+    # tx so the BAL builder below can group nonce changes by
+    # sender uniformly.
+    def sender_generator() -> Generator[EOA, None, None]:
+        """Yield senders, fresh if distinct, else reuse one shared."""
+        sender = pre.fund_eoa()
+        while True:
+            yield sender if not distinct_senders else pre.fund_eoa()
```
LouisTsai-Csie: This function is used by both `test_sload_bloated_prefetch_miss` & `test_sload_bloated_multi_contract`. To avoid duplication, I'll extract it as a shared helper:

1. Move `sender_generator` out of the current function
2. Refactor `sender_generator` to:

```python
def sender_generator(pre, distinct_senders) -> Generator[EOA, None, None]:
    """Yield senders for each transaction.
    
    In distinct mode, yields a fresh EOA per call.
    Otherwise, reuses the same sender for all transactions.
    """
    sender = pre.fund_eoa()
    while True:
        yield sender if not distinct_senders else pre.fund_eoa()
```

3. Use in both test functions:
```python
senders_iter = sender_generator(pre, distinct_senders)
```

As suggested in comment

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +203,343 @@ def test_sload_bloated(
...
+
+    benchmark_test(
+        pre=pre,
+        blocks=blocks,
+        skip_gas_used_validation=True,
+        expected_receipt_status=True,
+    )
+
+
+@pytest.mark.repricing
```
LouisTsai-Csie: Same as [comment](https://github.com/ethereum/execution-specs/pull/2657#discussion_r3077346795)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +203,343 @@ def test_sload_bloated(
...
+        + While(
+            body=(Op.DUP1 + Op.SLOAD + Op.POP + Op.PUSH1(1) + Op.ADD),
+            condition=Op.GT(Op.GAS, 0xFFFF),
+        )
+        + Op.PUSH0
+        + Op.SSTORE
+    )
+
+    base_offset = 1 if existing_slots else START_SLOT
+    max_sloads_per_tx = _max_sloads_per_tx(tx_gas_limit)
```
LouisTsai-Csie: Same as [comment](https://github.com/ethereum/execution-specs/pull/2657#discussion_r3077348840)

tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -187,6 +203,343 @@ def test_sload_bloated(
...
+    # Yield one sender per tx via a generator. In distinct mode
+    # each call returns a fresh EOA; otherwise the same shared
+    # sender is reused. The senders list collects one entry per
+    # tx so the BAL builder below can group nonce changes by
+    # sender uniformly.
+    def sender_generator() -> Generator[EOA, None, None]:
+        """Yield senders, fresh if distinct, else reuse one shared."""
+        sender = pre.fund_eoa()
+        while True:
+            yield sender if not distinct_senders else pre.fund_eoa()
```
LouisTsai-Csie: same as [comment](https://github.com/ethereum/execution-specs/pull/2657#discussion_r3077367376)

LouisTsai-Csie (changes requested):
Thanks for the update, i did a final check, after these changes let's merge.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: For now, we could just remove the repricing marker here, no extra action needed for the two points above!

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: Please check: https://github.com/ethereum/execution-specs/pull/2657#discussion_r3077348840

tests/benchmark/stateful/bloatnet/test_single_opcode.py
LouisTsai-Csie: We already have this helper, it is not clear though:

https://github.com/ethereum/execution-specs/blob/47782b3d5b1d24e435752878954e01e978681b98/tests/benchmark/stateful/bloatnet/test_single_opcode.py#L1262

In the `test_sstore_variants` benchmark, we delegate to this helper then initialize the storage, later execute the benchmark code. This would not go out of gas, but not sure if it would take too much time or not.

tests/benchmark/stateful/bloatnet/test_single_opcode.py
nerolation: These tests aren't necessarily BAL specific though. The test is essentially the worst-case SLOAD block that we'd see today - thus we should have always had it. After BALs another test (the one where we load storage lexicopgrahically reversed) will turn into the worst-case which should be used for repricing too.

nerolation:
Everything addressed. Ready for one last review!

LouisTsai-Csie (approved):
Thank you! I will merge it for now, and refactor in PR #2672, as the current storage initialization would break at high gas configuration.

PR #2664, opened 2026-04-13
fix(test-plugins-filler): aggregate --verify-traces results across xdist workers

leolara:
## 🗒️ Description

Fix a silent data-loss bug in the `--verify-traces` pytest plugin under `pytest-xdist`.

With `-n > 1`, `TraceVerifier` stored per-test comparison results on an instance attribute (`self.test_results`). Each xdist worker has its own subprocess and its own plugin instance, so only one worker's slice of the data ever reached `pytest_terminal_summary` / `--verify-traces-json`. On a real run with `-n 10` this hid ~90% of the trace comparisons from the JSON report — `fill tests/ported_static/stCallCodes/` went from 492 JSON entries with `-n 1` to 56 with `-n 10`.

The fix follows the same pattern already used for `t8n_cache_stats` in `filler.py`:

1. `TraceDifference` / `TransactionCountMismatch` / `TraceComparisonResult` get `to_dict` / `from_dict` for xdist transfer. `TraceDifference.from_dict` dispatches on a `kind` discriminator so subclasses round-trip.
2. `TraceVerifier.pytest_sessionfinish` serializes results to `config.workeroutput` on workers (no-op on the controller).
3. A module-level `pytest_testnodedown` hook merges each worker's payload into the controller-side plugin instance.
4. `pytest_terminal_summary` is guarded with `hasattr(config, "workerinput")` so workers no longer race to overwrite the JSON file.

`node` on `pytest_testnodedown` is typed via a local `Protocol` (xdist ships without `py.typed`, so importing `WorkerController` would force a `# type: ignore`). `error: object | None` matches the upstream hookspec.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/verify_traces.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/verify_traces.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import Any, Generator
+from typing import Any, Generator, Protocol
 
 import pytest
 from _pytest.terminal import TerminalReporter
@@ -26,6 +26,21 @@ from execution_testing.client_clis.trace_report_formatter import (
     TextTracesDiffReportFormatter,
     TracesDiffReportFormatter,
 )
+
+
+class _XdistWorkerNode(Protocol):
+    """
+    Subset of ``xdist.workermanage.WorkerController`` we depend on.
+
+    Defined here as a Protocol because ``xdist`` ships without a
+    ``py.typed`` marker, so importing its concrete class would force a
+    ``# type: ignore`` on every consumer. We only need ``workeroutput``
+    (the worker→controller dict) and ``config.pluginmanager``.
+    """
+
+    workeroutput: dict[str, Any]
+    config: pytest.Config
+
 
 # ---------------------------------------------------------------------------
 # Baseline loading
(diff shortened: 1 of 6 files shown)


packages/testing/src/execution_testing/client_clis/trace_comparators.py
marioevz, line 1: Not related to the changes in this PR, but while working on EIP-8037 tests, I realized that we might also need a `EXACT_NO_STACK_NO_GAS_NO_GAS_COST` or similar, because this EIP changes the gas cost prices of many operations, and with a trace comparator like this we could verify that the EIP-8037 does not break tests from other EIPs.

packages/testing/src/execution_testing/client_clis/trace_comparators.py
```diff
@@ -47,6 +47,41 @@ class TraceDifference:
     baseline: str
     current: str
 
+    def to_dict(self) -> dict:
```
marioevz: This is the perfect usecase for Pydantic models: Just do `class TraceDifference(BaseModel)` and Pydantic will take care of all of this.

The only difference is that you'll have to use `TraceDifference.model_dump` instead of `to_dict`, and `TraceDifference.model_validate` instead of `from_dict`.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/verify_traces.py
```diff
@@ -245,13 +260,55 @@ def pytest_runtest_makereport(
...
+            return
+        if not self.test_results:
+            return
+        config.workeroutput["trace_verifier_results"] = {
+            nodeid: {
+                comp_name: result.to_dict()
+                for comp_name, result in comps.items()
+            }
+            for nodeid, comps in self.test_results.items()
+        }
```
marioevz, line 296: We can define a pydantic class for test results and this becomes:
```suggestion
        config.workeroutput["trace_verifier_results"] = self.test_results.model_dump(mode="json")
```

You might need to extra classes:
```python
class TestResult(RootModel[dict[str, TraceComparisonResult]]):
  root: dict[str, TraceComparisonResult]

class TestResults(RootModel[dict[str, TestResult]]):
  root: dict[str, TestResult]
```
In place of the current `self.test_results` type definition of `dict[str, dict[str, TraceComparisonResult]]`, but in the long term it's more maintainable because Pydantic handles any changes to the types automatically.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/verify_traces.py
```diff
@@ -265,3 +322,32 @@ def pytest_terminal_summary(
...
+    The worker has serialized its ``test_results`` into
+    ``node.workeroutput`` (see ``TraceVerifier.pytest_sessionfinish``);
+    here we deserialize and merge them into the controller-side plugin
+    instance so ``pytest_terminal_summary`` sees the combined view.
+    """
+    del error
+    payload = getattr(node, "workeroutput", {}).get("trace_verifier_results")
+    if not payload:
+        return
+    plugin = node.config.pluginmanager.get_plugin("trace-verifier")
```
marioevz, line 346:
```python
worker_test_results = TestResults.model_validate(payload)
```
to parse the results from the worker using pydantic.

marioevz (commented):
Looks mostly good to me, with the suggestion of using pydantic everywhere to make the code more maintainable.

leolara:
@marioevz 

 + I have changed it to pydantic, that is the right approach, I didn't think about that at the moment.
 + I have added the comparator mode that you suggested, it is very little code. However, I think it is better approach to don't try to do both transitions at once: not-hardcode addresses and EIP8037. I am working on the not-harcode addresses with no-stack comparator. After that we will adapt them to EIP8037 with the gas comparators

marioevz (approved):
LGTM, thanks for the changes!

PR #2665, opened 2026-04-13
fix(test-fill): resolve checklist path for EIPs referenced via `eip=[N]` kwarg

danceratopz:
## 🗒️ Description

`@pytest.mark.eip_checklist(..., eip=[N])` is the documented escape hatch for a test that covers EIP `N` but lives outside a `tests/**/eip<N>_*/` directory (introduced in ethereum/execution-spec-tests#2088). The collector's `get_eip(N)` stores such EIPs with `path=None`, relying on a later pass over items inside the EIP's own directory to back-fill the path. When those primary tests aren't collected (e.g. deselected by a `-m` filter or `--until`), the path stays `None` and `--checklist-doc-gen` aborts at `assert eip.path is not None`:

```text
INFO    -  INTERNALERROR>   File "/home/dtopz/code/github/execution-specs.specify-docs/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/eip_checklist.py", line 515,
           in pytest_collection_modifyitems
INFO    -  INTERNALERROR>     assert eip.path is not None
INFO    -  INTERNALERROR> AssertionError: assert None is not None
INFO    -  INTERNALERROR>  +  where None = EIP(number=7951, items={'general/code_coverage/eels': EIPItem(id='general/code_coverage/eels', line_number=20, descrip...when the new constraint
           is not met. ', tests=set(), not_applicable_reason='', external_coverage_reason='')}, path=None).path
INFO    -
INFO    -  ======================================================== 10 warnings in 1.16s ========================================================
ERROR   -  Documentation generation failed (exit: 3, INTERNAL_ERROR).
ERROR   -  Script '/home/dtopz/code/github/execution-specs.specify-docs/docs/scripts/gen_test_case_reference.py' caused SystemExit(<ExitCode.INTERNAL_ERROR: 3>)
```

This PR:

- Adds a filesystem fallback `_find_eip_dir(tests_root, n)` that scans `tests/` for the canonical `eip<n>_*` directory after item collection, so the checklist doc can still attach to the EIP's natural location when its primary tests weren't collected.
- Downgrades the `assert eip.path is not None` to a `logger.warning(...)` plus `continue`, so a genuinely missing `eip<n>_*` directory no longer crashes checklist doc generation.

[...]

PR #2666, opened 2026-04-13
feat(doc,test-benchmark): re-enable benchmark tests in test case reference

### LINKED ISSUE #1756: chore(docs): re-enable benchmark test documentation

danceratopz:
## 🗒️ Description

Adds benchmark tests to the "Test Case Reference" in the HTML docs.

- Use `--include-benchmark` in `gen_test_case_reference.py` to allow collection of benchmark tests together alongside consensus tests (info why this is required in #2668).
- Improve test case counting: benchmark tests were counted incorrectly as they target a different fork. Now counting depends only on the collected cases, not on the target fork.
- As a bonus:
    - The count of tests that have transition forks is fixed.
    - Forks with spaces get ordered correctly, specifically "Tangerine Whistle".

Test locally via:
```
just docs-serve
```

> [!IMPORTANT]
> Requires #2665 to land first. Without that fix, this script crashes with `INTERNALERROR> AssertionError: assert None is not None` on EIP 7951, because `tests/benchmark/compute/precompile/test_p256verify.py:34` references it via the `eip=[7951]` kwarg of `eip_checklist`.

> [!IMPORTANT]
> Requires #2668 to land first. Without that fix, it is impossible to collect both consensus tests and benchmark tests in a single run. This was preferable to multiple runs to avoid the complexity of accumulating state between runs.

image

## 🔗 Related Issues or PRs

- Closes #1756.
- Requires #2665.
- Requires #2668.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/tests/benchmark/compute/eip7928_block_level_access_lists/__init__.py
+++ b/tests/benchmark/compute/eip7928_block_level_access_lists/__init__.py
@@ -1,0 +1,1 @@
+"""EIP-7928 Block-level Access List benchmark tests."""
--- a/docs/scripts/gen_test_case_reference.py
+++ b/docs/scripts/gen_test_case_reference.py
@@ -29,21 +29,22 @@ GENERATE_UNTIL_FORK = DocsConfig().GENERATE_UNTIL_FORK
 logger = logging.getLogger("mkdocs")
 
 
-# If docs are generated while FAST_DOCS is true, then use "tests/frontier"
-# otherwise use "tests"
+# If docs are generated while FAST_DOCS is true, then use "tests/frontier";
+# otherwise include both "tests" and "tests/benchmark". The benchmark path
+# must be listed explicitly so its tests are picked up alongside the rest.
 #
 # USAGE 1 (use fast mode):
 #       export FAST_DOCS=true && uv run mkdocs serve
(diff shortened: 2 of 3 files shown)


danceratopz:
image

@LouisTsai-Csie there's a small bug here. Currently, benchmark tests target Osaka, but we show Amsterdam in the docs. As we don't add the `--until Amsterdam` flag, it's technically correct, but documents the wrong target fork. Can we live with this for now? Because I assume we will start benchmarking against Amsterdam sooon? Wdyt?

LouisTsai-Csie:
^ @danceratopz i think this is fine and we have started benchmarking on Amsterdam now also

LouisTsai-Csie (approved):
Approved from my side! I've also run locally and check the content.

tests/benchmark/compute/eip7928_block_level_access_lists/__init__.py
danceratopz, line 1: Btw, mkdocstrings failed as this was missing; the test module folder couldn't be imported as it wasn't a valid package.

PR #2668, opened 2026-04-13
feat(test-fill): add `--include-benchmark` and scan both tests + tests/benchmark in `checklist`

danceratopz:
## 🗒️ Description

TIL that we apply checklist markers to benchmark tests. This seems very reasonable in certain situations, although we should ensure that these tests are also run regularly in a standard public dashboard to ensure coverage. @kclowes wdyt? Worth making this change?

Two problems this PR fixes:

- The `checklist` CLI rejected positional paths despite the docstring example: `uv run checklist tests/prague/eip7702*` errored with `Got unexpected extra argument`, because `@click.argument("paths", ...)` was missing.
- Benchmark tests' `eip_checklist` markers were silently dropped from generated checklists. `tests/benchmark/conftest.py` hides the benchmark directory from a plain `tests/` collection unless a benchmark path is passed explicitly.

### Why the obvious fix does not work

Defaulting `paths` to `("tests", "tests/benchmark")` looks correct, but pytest deduplicates overlapping path arguments: when both `tests/` and `tests/benchmark/...` are passed, pytest drops the broader `tests/` arg before collection starts. The benchmark `pytest_ignore_collect` hook is scoped to `tests/benchmark/` and never fires for sibling paths, so it cannot rescue `tests/prague/` etc. Net effect: only benchmark tests got collected from the default.

### The fix

Add an `--include-benchmark` flag, registered in the shared `benchmarking` plugin. When set, `tests/benchmark/conftest.py::pytest_ignore_collect` short-circuits to `return False`, including `tests/benchmark/` regardless of the positional args. The `checklist` CLI now passes a single `tests/` path plus `--include-benchmark` when no paths are supplied, sidestepping pytest's path deduplication entirely. Explicit positional paths still work and override the default.

Verified:

- `uv run fill --co tests/` still excludes benchmark (~360k tests, unchanged default behavior).
- `uv run fill --co tests/ --include-benchmark` now collects both (~366k tests).
- `uv run checklist` (default) generates checklists for benchmark-only EIPs such as EIP-7928.
- `uv run checklist tests/prague/eip7702_set_code_tx` remains scoped and excludes benchmark.

> [!IMPORTANT]
> Requires #2665. Without that fix, expanding default collection to include `tests/benchmark` trips the `assert eip.path is not None` assertion on EIP-7951 because the kwarg-referenced EIP cannot resolve its canonical doc location.

## 🔗 Related Issues or PRs

- Requires #2665.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
@@ -90,7 +90,7 @@ def checklist(
         uv run checklist tests/prague/eip7702_set_code_tx
 
         # Limit to a specific fork
-        uv run checklist --until Prague
+        uv run checklist --fork Prague
 
         # Specify output directory
         uv run checklist --output ./my-checklists
--- a/packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
@@ -34,6 +34,12 @@ def _last_development_fork() -> str | None:
 
 
 @click.command()
+@click.argument(
+    "paths",
+    nargs=-1,
+    type=click.Path(),
+    default=("tests", "tests/benchmark"),
+)
 @click.option(
     "--output",
     "-o",
@@ -56,7 +62,11 @@ def _last_development_fork() -> str | None:
     help="Include upcoming forks up to and including this fork",
 )
 def checklist(
-    output: str, eip: tuple[int, ...], until: str | None, **kwargs: Any
+    paths: tuple[str, ...],
+    output: str,
+    eip: tuple[int, ...],
+    until: str | None,
+    **kwargs: Any,
 ) -> None:
     """
     Generate EIP test checklists based on pytest.mark.eip_checklist markers.
@@ -64,18 +74,21 @@ def checklist(
     This command scans test files for eip_checklist markers and generates
     filled checklists showing which checklist items have been implemented.
 
+    By default it scans `tests` plus `tests/benchmark`; pass one or more
+    paths to limit collection to a subset.
+
     By default, includes all development forks so that checklists for
     upcoming EIPs are generated without needing --until.
 
     Examples:
-        # Generate checklists for all EIPs
+        # Generate checklists for all EIPs (default: tests + tests/benchmark)
         uv run checklist
 
         # Generate checklist for specific EIP
         uv run checklist --eip 7702
 
-        # Generate checklists for specific test path
-        uv run checklist tests/prague/eip7702*
+        # Generate checklists for a specific test path
+        uv run checklist tests/prague/eip7702_set_code_tx
 
         # Limit to a specific fork
         uv run checklist --until Prague
@@ -100,10 +113,14 @@ def checklist(
     if until:
         args.extend(["--until", until])
 
+    # The default `paths` (`tests`, `tests/benchmark`) lists `tests/benchmark`
+    # explicitly because positional paths override `testpaths = tests/` in
+    # pytest-fill.ini, otherwise checklist coverage from benchmark tests
+    # gets dropped.
+    args.extend(paths)
+
     command = ChecklistCommand(
-        plugins=[
-            "execution_testing.cli.pytest_commands.plugins.filler.eip_checklist"
-        ],
+        plugins=["execution_testing.cli.pytest_commands.plugins.filler.eip_checklist"],
     )
     command.execute(args)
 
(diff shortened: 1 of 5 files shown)


kclowes (approved):
> @kclowes wdyt? Is it worth making this change?

I think so! I'm surprised that `tests/benchmark` wasn't getting picked up with `tests/`.

LouisTsai-Csie (approved):
This is good! benchmark tests should support the eip checklist. We could now start to label tests under `tests/benchmark/stateful/eip7928_block_level_access_lists`

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -71,6 +71,19 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+        dest="include_benchmark",
+        default=False,
+        help=(
+            "Include tests/benchmark/ during collection even when no "
+            "benchmark path is passed as a positional argument. Needed "
+            "when collecting both tests/ and tests/benchmark/ in a single "
+            "pytest invocation, since pytest's path deduplication drops "
+            "the broader tests/ arg."
+        ),
+    )
```
danceratopz, line 86: I opted for this solution: Either we add this flag, or we have to run checklist twice, once for consensus tests and once for benchmark tests.

danceratopz: We need to fix this in two places, the other is in #1756.

For now, the documentation will be fixed by running fill with the doc-gen plugin twice, once for Osaka (benchmark) and once for Amsterdam (consensus) to achieve clean docs. But once ./tests/benchmark get filled with Amsterdam, we can apply the same fix for the docs (i.e., use the `--include-benchmark`) and only run once.

danceratopz: It turns out a two-pass approach to accommodate for two target forks requires accumulating state which blows up the scope of that fix (#1756). We need this fix there as well, to keep the change more manageable.

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
```diff
@@ -56,29 +61,36 @@ def _last_development_fork() -> str | None:
...
         uv run checklist --eip 7702
 
-        # Generate checklists for specific test path
-        uv run checklist tests/prague/eip7702*
+        # Generate checklists for a specific test path
+        uv run checklist tests/prague/eip7702_set_code_tx
 
         # Limit to a specific fork
-        uv run checklist --until Prague
+        uv run checklist --fork Prague
```
LouisTsai-Csie: I tried to run this but failed, i think the `--fork` is not registered for `Click` command:

```
(execution-specs) execution-specs % uv run checklist --fork Prague
Usage: checklist [OPTIONS] [PATHS]...
Try 'checklist --help' for help.

Error: No such option: --for
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_include_benchmark.py
```diff
@@ -0,0 +1,126 @@
...
+        args = config.invocation_params.args or ()
+        if any(
+            benchmark_dir in Path(a).resolve().parents
+            or Path(a).resolve() == benchmark_dir
+            for a in args
+        ):
+            return False
+        return True
+    """
+)
```
LouisTsai-Csie: Is it possible to import from conftest.py directly instead of copying the content? I'm worried that if the file changes, we'll end up with inconsistent versions.

LouisTsai-Csie (changes requested):
Some suggestion for the CLI

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
danceratopz: Oh thanks Louis.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_include_benchmark.py
danceratopz: Good catch.

packages/testing/src/execution_testing/cli/pytest_commands/checklist.py
```diff
@@ -56,29 +61,36 @@ def _last_development_fork() -> str | None:
...
 
-        # Generate checklists for specific test path
-        uv run checklist tests/prague/eip7702*
+        # Generate checklists for a specific test path
+        uv run checklist tests/prague/eip7702_set_code_tx
 
-        # Limit to a specific fork
-        uv run checklist --until Prague
+        # Generate until a specific fork
+        uv run checklist --fork Prague
```
danceratopz:
```suggestion
        uv run checklist --until Prague
```

LouisTsai-Csie (approved):
LGTM

PR #2672, opened 2026-04-14
feat(test-benchmark): storage initialization helper with 7702 authorization

LouisTsai-Csie:
## 🗒️ Description

In the pre-allocation phase that deploys a contract, the framework supports passing in a dictionary that represents the storage to be initialized, and then embeds the logic inside the `initcode`. When the contract is created, it executes the initcode and performs the storage update.

However, in most of the benchmark cases, this is too inefficient, since the initialization phase requires a huge range of storage initialization, it leads to an `initcode prefix too long` issue.

### Solution

To resolve this issue, we create a helper for the stateful benchmark scenario. Users can define a storage init range like this:

```python
@dataclass(frozen=True)
class StorageInitRange:
    """One contiguous range of storage to initialize."""

    start_slot: int
    num_slots: int
    offset: int
```

Which means the storage would be initialized to `STORAGE[i] = i + offset`, where `i` starts from `start_slot` to `start_slot + num_slots`.

We leverage the EIP-7702 authorization approach for testing. The workflow now becomes:

1. Delegate the authority to the storage initialization contract (`create_sequential_sstore_initializer`).
2. Initialize the specified range — this leverages `IteratingBytecode` with `transactions_by_iteration_counts`.
3. Pack the initialization transactions into blocks under the setup phase.
4. Re-delegate the authority to the real benchmarking execution code.
5. Start the BAL benchmarking phase.

### Design decisions

- The authority EOA is created in the BAL benchmark test, not inside the helper function. This provides the flexibility to use a stub account if needed.
- Nonces use `authority.nonce` (incremented in-place) rather than hardcoded constants, ensuring compatibility with execute-remote where nonces are fetched from the live network.
- Shared helpers (`build_sequential_storage_init`, `StorageInitRange`, etc.) live in `tests/benchmark/stateful/helpers.py` so they are reusable beyond the BAL tests.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/helpers.py
+++ b/tests/benchmark/stateful/helpers.py
@@ -4,12 +4,15 @@ from collections.abc import Callable
 from enum import Enum
 
 from execution_testing import (
+    EOA,
     AccessList,
     Address,
     Alloc,
+    AuthorizationTuple,
     Block,
     Fork,
     Hash,
+    IteratingBytecode,
     Op,
     Transaction,
 )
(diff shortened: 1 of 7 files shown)


LouisTsai-Csie:
For @jochem-brouwer and @nerolation , you could focus on `create_sequential_sstore_initializer` and related logic. For `create_sstore_initializer` and `build_delegated_storage_setup`, i just move them out of the `test_single_opcode.py` file, no logic changes. You could ignore this as it is irrelevant to BAL benchmark

tests/benchmark/stateful/helpers.py
```diff
@@ -132,3 +138,313 @@ def build_cache_strategy_blocks(
...
+    initializer_calldata_generator: Callable[[int, int], bytes],
+) -> list[Block]:
+    """
+    Build setup blocks for delegated storage benchmarks.
+
+    Use EIP-7702 authorization to delegate an authority EOA first to
+    a storage-initializer contract (if *needs_init*), then to the
+    benchmark executor contract.  Return the list of setup blocks.
+    """
+    blocks: list[Block] = []
```
jochem-brouwer, line 197: Should this method be wrapped in `TestPhaseManager.setup()` ?

tests/benchmark/stateful/helpers.py
```diff
@@ -132,3 +138,313 @@ def build_cache_strategy_blocks(
...
+        authority_nonce += 1
+
+        # Calculate max slots per transaction based on gas cost
+        iteration_cost = initializer_code.tx_gas_limit_by_iteration_count(
+            fork=fork,
+            iteration_count=1,
+            start_iteration=1,
+            calldata=initializer_calldata_generator,
+        )
+        iteration_count = max(1, tx_gas_limit // iteration_cost)
```
jochem-brouwer, line 228: I think this should be `ceil(tx_gas_limit // iteration_cost)`?
```suggestion
        iteration_count = ceil(tx_gas_limit // iteration_cost)
```

tests/benchmark/stateful/helpers.py
```diff
@@ -132,3 +138,313 @@ def build_cache_strategy_blocks(
...
+                    calldata=initializer_calldata_generator,
+                )
+            )
+
+        # Pack init transactions into blocks
+        blocks.extend(pack_transactions_into_blocks(init_txs, tx_gas_limit))
+
+    # Final block: Authorize to executor
+    blocks.append(
+        Block(
```
jochem-brouwer, line 249: I think this should read the nonce for the authority EOA variable. If we use a test which has to delegate again after the setup, then we must ensure the nonce is correct 🤔

jochem-brouwer (commented):
Some questions about the storage logic 😄 👍

tests/benchmark/stateful/helpers.py
LouisTsai-Csie: I think this is fine, the block construction part is already within the `setup` syntax

tests/benchmark/stateful/helpers.py
LouisTsai-Csie: I think the nonce should be refreshed for the next delegation phase

fselmo (approved):
lgtm!

PR #2679, opened 2026-04-14
feat(test-cli): Add support for testing block building via simulator

### LINKED ISSUE #2560: Add block building correctness testing via `testing_buildBlockV1` endpoint
## Description

Following the block building discussion, we now have two things to do: verifying the block building correctness / benchmarking the block building performance. This issue is for the former topic.

We want to add correctness checks for the block building path. Currently we only validate block consumption (client receives a block and verifies it). This would add the reverse direction, given a set of transactions and block properties, verify that the client assembles the correct block.

## Approach

- Introduce a new consumer (e.g. `consume build`) alongside the existing `consume engine`
- Use the `testing_` endpoint, which guarantees transaction ordering (unlike mempool-based block building)
- Reuse existing test fixtures where possible

## Question:
- How much logic does the `testing_` endpoint share with actual network block building? We should track cases where devnet block building breaks but this consumer passes, as that shows a gap.

fselmo:
## 🗒️ Description

Adds a hive-based simulator that verifies EL clients build correct blocks via ``testing_buildBlockV1``

For each payload in a blockchain_test_engine fixture, the test sends the expected transactions and block attributes to the client's block building endpoint. The resulting block is validated against the fixture's expected output (`state_root`, `receipts_root`, etc). The fixture block is then imported via `engine_newPayloadVX` to advance the chain for subsequent blocks.

`gas_limit` is validated a bit different for now, since `testing_buildBlockV1` doesn't accept a gas limit parameter. The client picks its own within the EIP-1559 adjustment range so we validate it is within that range until we can actually account for the gas limit. For this reason, `block_hash` is also excluded from comparison as it depends on `gas_limit`.

Also looking for recs on the naming (`uv run build-block`). We _could_ go with `build` by itself but it feels vague to me. That said, I can be easily convinced here 😄.

## 🔗 Related Issues or PRs

closes #2560

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture
Screenshot 2026-04-14 at 13 23 01


packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/simulator_logic/test_via_build.py
```diff
@@ -0,0 +1,323 @@
...
+from ..helpers.timing import TimingData
+
+logger = get_logger(__name__)
+
+# Execution-dependent fields that must match exactly between the built
+# and expected payloads.  Excluded:
+# - gas_limit: testing_buildBlockV1 doesn't accept it; the client
+#   picks its own via EIP-1559 (validated separately by range check).
+# - block_hash: depends on gas_limit, so it will differ too.
+_VALIDATED_FIELDS = (
```
marioevz: We should probably take fields from `FixtureExecutionPayload` and remove the `gas_limit` and `block_hash`, just to not having to come here and update this list separately whenever we want to add a new field to the engine payload.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/simulator_logic/test_via_build.py
```diff
@@ -0,0 +1,323 @@
...
+
+def _payload_attributes(
+    payload: FixtureEngineNewPayload,
+) -> PayloadAttributes:
+    """Extract PayloadAttributes from a fixture payload."""
+    execution_payload = payload.params[0]
+    parent_beacon_block_root = (
+        payload.params[2] if len(payload.params) >= 3 else None
+    )
+    return PayloadAttributes(
```
marioevz: I guess there's no way around it in this case. Perhaps just adding a `get_payload_attributes` method to `FixtureEngineNewPayload`? Just so that it's easier to spot when adding a new field that this method has to be updated too.

marioevz (commented):
Looks great! Just two comments to try to improve maintainability.

fselmo:
@marioevz lmk if the last commit here is what you meant. I re-ran against geth `bal-devnet-3` with this cherry-picked there with some minor updates that are needed (slotnum, etc) and it's still running smooth 👍🏼

packages/testing/src/execution_testing/cli/pytest_commands/processors.py
```diff
@@ -150,6 +150,13 @@ def process_args(self, args: List[str]) -> List[str]:
                     "execution_testing.cli.pytest_commands.plugins.consume.simulators.rlp.conftest",
                 ]
             )
+        elif self.command_name == "build_block":
+            modified_args.extend(
+                [
+                    "-p",
+                    "execution_testing.cli.pytest_commands.plugins.consume.simulators.build_block.conftest",
+                ]
+            )
```
LouisTsai-Csie, line 159: This is not important but maybe we could have something like this for the entire if-else comparison:

```python
command_name = self.command_name
simulator_command_list = ["engine", "enginex", ...]

if command_name in simulator_command_list:
  modified_args.extend(
      [
          "-p",
          f"execution_testing.cli.pytest_commands.plugins.consume.simulators.{command_name}.conftest",
      ]
  )

else:
  raise ValueError(f"Unknown command name: {self.command_name}")
```

i would like to continue on reviewing the remaining part of the PR!

packages/testing/src/execution_testing/cli/pytest_commands/build_block.py
```diff
@@ -0,0 +1,46 @@
...
+def create_build_block_command() -> PytestCommand:
+    """Initialize the build-block command with paths and processors."""
+    base_path = Path("cli/pytest_commands/plugins/consume")
+    command_logic_test_paths = [
+        base_path / "simulators" / "simulator_logic" / "test_via_build.py"
+    ]
+    return PytestCommand(
+        config_file="pytest-consume.ini",
+        argument_processors=[
+            HelpFlagsProcessor("build_block"),
```
LouisTsai-Csie:
```suggestion
            HelpFlagsProcessor("consume"),
```

The `--help` command is currently not working, this is the error message after i tried:

```
ERROR: usage: build-block [options] [file_or_dir] [file_or_dir] [...]
build-block: error: unrecognized arguments: --build_block-help
  inifile: /Users/caijiacheng/Documents/Ethereum/eels-worktree/wt-eip-7778/packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-consume.ini
  rootdir: /Users/caijiacheng/Documents/Ethereum/eels-worktree/wt-eip-7778
```

After we register `build_block` in `HelpFlagsProcessor`, it would be processed to `--build_block-help` command (see `process_args` in `HelpFlagsProcessor`).

But there is no such command registered under `helper.py`, only `--consume-help`, `--fill-help` and others.

We either register `--build_block-help` or update to `HelpFlagsProcessor("consume")`.

packages/testing/src/execution_testing/cli/pytest_commands/build_block.py
```diff
@@ -0,0 +1,46 @@
...
+            ConsumeCommandProcessor(is_hive=True),
+        ],
+        command_logic_test_paths=command_logic_test_paths,
+    )
+
+
+@click.command(
+    name="build-block",
+    context_settings={
+        "help_option_names": ["-h", "--help"],
```
LouisTsai-Csie:
```suggestion
```

Do we need this? I checked `common_pytest_options` and found there is built-in `help` decorator:

```python
# check common_pytest_options function
func = click.option(
        "-h",
        "--help",
        "help_flag",
        is_flag=True,
        default=False,
        expose_value=True,
        help="Show help message.",
    )(func)
```

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -69,6 +70,9 @@
     FixtureTransactionReceipt,
 )
 
+if TYPE_CHECKING:
+    from execution_testing.rpc.rpc_types import PayloadAttributes
+
```
LouisTsai-Csie, line 75:
```suggestion
```

I think we dont need this? as you import `PayloadAttributes` in `get_payload_attributes` function, so there should be no effect of circular import.

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -472,6 +476,22 @@ def valid(self) -> bool:
...
         return self.validation_error is None
 
+    def get_payload_attributes(self) -> "PayloadAttributes":
+        """Return the ``PayloadAttributes`` corresponding to this payload."""
+        from execution_testing.rpc.rpc_types import PayloadAttributes
+
+        execution_payload = self.params[0]
+        parent_beacon_block_root = (
+            self.params[2] if len(self.params) >= 3 else None
+        )
```
LouisTsai-Csie, line 490:
```suggestion
        parent_beacon_block_root = (
            self.params[2] if self.forkchoice_updated_version >= 3 else None
        )
```

I think this is much clear. So now we know that the `parent_beacon_block_root` only exists starting from version 3.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/simulator_logic/test_via_build.py
```diff
@@ -0,0 +1,289 @@
...
+        try:
+            response = engine_rpc.forkchoice_updated_with_retry(
+                forkchoice_state=ForkchoiceState(
+                    head_block_hash=fixture.genesis.block_hash,
+                ),
+                forkchoice_version=fixture.payloads[
+                    0
+                ].forkchoice_updated_version,
+                max_attempts=30,
+                wait_fixed=1.0,
```
LouisTsai-Csie:
```suggestion
```

Removed since they are default value

packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/simulator_logic/test_via_build.py
```diff
@@ -0,0 +1,289 @@
...
+        logger.info("Calling getBlockByNumber to get genesis block...")
+        genesis_block = eth_rpc.get_block_by_number(0)
+        assert genesis_block is not None, "genesis_block is None"
+        if genesis_block["hash"] != str(genesis_header.block_hash):
+            expected = genesis_header.block_hash
+            got = genesis_block["hash"]
+            logger.fail(
+                f"Genesis block hash mismatch. "
+                f"Expected: {expected}, Got: {got}"
+            )
```
LouisTsai-Csie:
```suggestion
```

I notice this comparison logic is covered in `GenesisBlockMismatchExceptionError`:

```python
# inside GenesisBlockMismatchExceptionError
message = (
    "Genesis block hash mismatch.\n\n"
    f"Expected: {expected_header.block_hash}\n"
    f"     Got: {got_genesis_block['hash']}."
)
```

LouisTsai-Csie (dismissed):
Thank you @fselmo for this feature! I leave some comment but they are all style-related issue.

For `test_via_build.py`, i think `test_blockchain_via_build` is too large as a single function and has too much nested logic. Could we split it into smaller helpers?

My proposal:

`test_blockchain_via_build`
-> `bootstrap_engine_at_genesis`: handle genesis block FCU, verify genesis hash
-> Extract valid payload and calls `build_validate_and_advance` for each payload

`build_validate_and_advance`
-> build block
-> validate block response and gas limit
-> proceed the chain with `newPayload`
-> update the chain with FCU

---

edit: two more things for the new command
- Could we add some docs for the new command?
- Should we PR to Hive for the new simulator?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/simulator_logic/test_via_build.py
```diff
@@ -0,0 +1,289 @@
...
+                        extra_data=expected_payload.extra_data,
+                    )
+
+                # 2. Validate fields + gas limit range
+                _validate_built_block(
+                    built=build_response.execution_payload,
+                    expected=expected_payload,
+                    parent_gas_limit=parent_gas_limit,
+                )
+
```
danceratopz: We should check the execution requests here for >= v4. I.e., that `build_response.execution_requests == payload.params[3]`.

danceratopz: Potential follow-up: It would also be really nice to test `build_response.blobs_bundle`, but we don't include commitments/proofs in any fixture format yet. We just have these tests that are `execute` mode only:
https://github.com/ethereum/execution-specs/blob/c9d8be0101c6649e060950317e14fb4f515f70e9/tests/osaka/eip7594_peerdas/test_get_blobs.py#L352-L367

packages/testing/src/execution_testing/cli/pytest_commands/build_block.py
```diff
@@ -0,0 +1,46 @@
...
+            HelpFlagsProcessor("build_block"),
+            HiveEnvironmentProcessor(command_name="build_block"),
+            ConsumeCommandProcessor(is_hive=True),
+        ],
+        command_logic_test_paths=command_logic_test_paths,
+    )
+
+
+@click.command(
+    name="build-block",
```
danceratopz, line 34: This command does use the `testing_buildBlockV1` endpoint, but I think it's rather close to a `consume` simulator (it still consumes a fixture) so I wondered if it should be a `consume` sub-command? 

I.e., something like `uv run consume build-block`. 

Then the simulator in hive would be `eels/consume-build-block`.

Imo, when we add a command to fill tests using the `testing_buildBlockV1` in the future, _that command_ should be the new `build` or `build-block` command. Wdyt?

danceratopz (changes requested):
Really nice!! 🔥 

Nice review by @LouisTsai-Csie. In addition, I added one comment below about adding execution request verification for forks >= Prague.

About the naming, imo, I don't think this merits a new top-level command namespace and it that this rather belongs under `consume`, please see the comment below.

#### Docs

Also, could you please add some docs for this? Assuming you rename under `consume`, we could add it to:
- `docs/running_tests/running.md`
- `docs/consume/simulators.md`?

If we don't rename, then it will be trickier to find a spot to document this, but I think this demonstrates that it belongs to the consume family? But it could go in a new section docs/running_tests/build_block/index.md. But I'm strongly in favor of not adding a new top-level entry-point for this functionality.

#### System Test

Last request, could you add an e2e test in https://github.com/ethereum/execution-specs/blob/c9d8be0101c6649e060950317e14fb4f515f70e9/.github/workflows/hive-consume.yaml ? It should be pretty quick!

packages/testing/src/execution_testing/cli/pytest_commands/build_block.py
fselmo: Hmm I thought about this a bit but it seems like the opposite of consume imo. It receives a request to build a block and then is _validated_ against what's in the fixture, rather than sending the payloads as they are (consuming) and then validating against some post state after.

I'm not one to argue semantics tbh, all I care is that the functionality is there. I'd love to get ideas from the team and settle on something but the reason I went with a new top-level command is that this feels entirely different to me. I can see the argument since they share a lot of things though. Happy to go either way here.

fselmo:
> But I'm strongly in favor of not adding a new top-level entry-point for this functionality.

@danceratopz I gave my reasoning in your first comment but I'm curious why the strong favor here. Imo it muddies what `consume` means. `build` is also quite generic and `build` for filling tests with state seems odd to me. I think we need to have some team discussion about it tbh. Could use some more heads on it. Will raise it in the next testing call.

packages/testing/src/execution_testing/cli/pytest_commands/build_block.py
fselmo: Thanks! fixed it :)

packages/testing/src/execution_testing/fixtures/blockchain.py
fselmo: We need it for the function types. Import in the body doesn't help that and it breaks mypy othewise.

packages/testing/src/execution_testing/cli/pytest_commands/processors.py
fselmo: done 👍🏼

packages/testing/src/execution_testing/fixtures/blockchain.py
LouisTsai-Csie: Thanks, i did not notice this.

packages/testing/src/execution_testing/cli/pytest_commands/build_block.py
LouisTsai-Csie: > It receives a request to build a block and then is validated against what's in the fixture, rather than sending the payloads as they are (consuming) and then validating against some post state after.

I agree this part is quite different from `consume`

LouisTsai-Csie:
The recent 2 commits fix all my review comment, but i think we still need the e2e tests: https://github.com/ethereum/execution-specs/blob/c9d8be0101c6649e060950317e14fb4f515f70e9/.github/workflows/hive-consume.yaml 

This could be done after we decided whether it should be an individual command or under consume family

danceratopz:
> @danceratopz I gave my reasoning in your first comment but I'm curious why the strong favor here. Imo it muddies what `consume` means.

Totally fair and I agree it's a different breed of simulator! My consume rationale stems from the fact that this simulator loads a `BlockchainEngineFixture` and the client _consumes the fixture's transactions_ in order to build a block. I mentally attach "consume" to "client consumes definitions from fixtures". Through that lens, to me, the lack of `consume` prefix doesn't adequately indicate that fixtures are the input.

You have a different interpretation of "consume" (send blocks/payloads then validate against post state), which is also legitimate.

I'm happy to reconcile my understanding of `consume` if others have similar interpretations, but I think it'd still be worth a quick discussion in the call tomorrow!

The following is **out of scope for this PR**, but it'd be great to have names that are intuitive,... in that spirit, I'll just float these ideas while we're on the topic... :laughing:  

Perhaps the name consume itself is too overloaded? I think if we named it again (and perhaps we even can?), I would prefer to call the simulators (in hive path language; not CLI names) as:

- `simulators/ethereum/eels/consensus/engine`,
- `simulators/ethereum/eels/consensus/rlp`,
- `simulators/ethereum/eels/consensus/sync` ( I think that works, still consensus).

I think we discussed this naming once, @spencer-tb. This would break everyone's CIs, but if we did it when (if) we make breaking changes to the release layout, then I think it'd be reasonable if it helps everyone's understanding.

Then (as is currently in this PR) we could have: 
- `simulators/ethereum/eels/build-block` 

Or if we think we'll have other build-block simulators, then 
- `simulators/ethereum/eels/build-block/engine`

When it comes to the CLI command, a unified top-level entry point such as `uv run hivesim build-block`, `uv run hivesim execute`, `uv run hivesim consensus-engine` (albeit a bit long) would also give more clarity about these commands and all simulators could be readily discovered via `uv run hivesim --help`. Or use `hive` instead of `hivesim` but I think it's more ambiguous.

packages/testing/src/execution_testing/cli/pytest_commands/build_block.py
danceratopz: Please see the rationale in https://github.com/ethereum/execution-specs/pull/2679#issuecomment-4424487698

spencer-tb:
I honestly would like to think on naming conventions a bit harder. And align it with what ever "`test-el`" becomes in the future (fast engine test / blockchain test cli runners). For example:

The hive versions:
- `uv run hive enginetest` <- consume engine
- `uv run hive blocktest` <- consume rlp
- `uv run hive synctest` <- consume sync
- `uv run hive buildtest` <- **new block builder**

Then the native cli version, replace `el` with whatever, maybe `evm` or `native`:
- `uv run el enginetest`
- `uv run el blocktest` 
- `uv run el statetest`
- `uv run el buildtest` <- **cli version**

That being said, I'd lean on keeping what we have here and pushing for a naming overhaul/alignment once the fast cli runners are complete and actively in use!

danceratopz:
@fselmo happy to go with `build-block`! Would you like to add the end-to-end test here in this PR or as a follow-up? I think it'd be ok to go in `consume-hive.yaml` for now? We could consider renaming to `hive.yaml` in a different PR?

fselmo:
> Would you like to add the end-to-end test here in this PR or as a follow-up?

Hmm I'm not sure I understand what you and @LouisTsai-Csie mean here with e2e test because `hive` doesn't have this simulator yet. Is this what you mean or something else? We will either need to PR to hive and merge it so that `hive-consume.yaml` can use it or we can do in a followup PR. Lmk @danceratopz 👀 

Either way I need both of you to remove your `Requested Changes` and approve before this can be merged 😅

danceratopz:
> > Would you like to add the end-to-end test here in this PR or as a follow-up?
> 
> Hmm I'm not sure I understand what you and @LouisTsai-Csie mean here with e2e test because `hive` doesn't have this simulator yet. Is this what you mean or something else? We will either need to PR to hive and merge it so that `hive-consume.yaml` can use it or we can do in a followup PR. Lmk @danceratopz 👀

I won't manage a review until tomorrow, so if you want to add it in the meantime, you could add a new `dev` entry here?
https://github.com/fselmo/execution-specs/blob/ff283695f8b4a646f9049aa0ea682ca178f15465/.github/workflows/hive-consume.yaml#L109-L112

> Either way I need both of you to remove your `Requested Changes` and approve before this can be merged 😅

Will do! :)

.github/workflows/hive-consume.yaml
```diff
@@ -106,15 +107,27 @@ jobs:
...
-          #   consume_command: enginex
+          #   cli_command: consume enginex
           #   test_filter: "(fork_shanghai or fork_cancun) and push0"
+          # NOTE: build-block uses testing_buildBlockV1 on the HTTP RPC port
+          # (8545), which has a smaller body cap than the Engine API port
+          # (8551); avoid filters that target near-block-size-limit tests.
+          - name: Dev Mode Build Block
+            mode: dev
+            cli_command: build-block
+            test_filter: "Osaka and test_clz_stack_underflow"
```
danceratopz, line 125: I did wonder if there was a better test set to use here, something with multiple blocks? `test_clz_stack_underflow` is a state test, therefore single block. 

```suggestion
            test_filter: "Osaka and test_block_hashes_history"
```

danceratopz (approved):
This is looking great. I took it for a spin locally and generally got it working well out the box for `Geth/v1.17.4-unstable-da34eb59-20260513`.

I did see some fails for tests that I'd filled (with `-m eels_base_coverage --from Paris`) on this branch, but I don't think it's worth reporting them in detail here. Ship it. I think these can be investigated and fixed in follow-ups.

PR #2682, opened 2026-04-15
fix(specs,tests): EIP-7981 - Update implementation and specs

nerolation:
Refactored the EIP-7981 branch to charge every access list byte as bytes × 4 × `GAS_TX_DATA_TOKEN_FLOOR`: adding the access-list floor cost via the inherited super_fn and a _access_list_floor_tokens helper instead of weighted token counting, so the branch contains only EIP-7981 logic on top of the EIP-7623 baseline (FLOOR=10) and will combine with EIP-7976 cleanly via rebase, without ever implementing 7976 itself.

### DIFF
--- a/packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_7981.py
+++ b/packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_7981.py
@@ -22,36 +22,35 @@ class EIP7981(BaseFork):
     """EIP-7981 class."""
 
     @classmethod
-    def _access_list_token_count(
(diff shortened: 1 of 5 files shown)


felix314159:
interesting, our CI detected that we need to make a ported [static test](https://github.com/ethereum/execution-specs/blob/d206d4f2b8558d742b9ddd31407e69e5cf05e6cf/tests/ported_static/stEIP2930/test_transaction_costs.py#L150) more flexible for this change. we probably will do this in `forks/amsterdam` and then i will rebase this branch for you

PR #2683, opened 2026-04-15
feat(specs): EIP-8037 diff-at-call-return state gas charging [DRAFT]

spencer-tb:
## Summary

Prototype implementation of diff-at-call-return state gas charging for EIP-8037, as proposed in the [#eip-8037](https://discord.com/channels/1359927674746835211/1493618116809457684) discord channel by @chfast.

Instead of charging state gas at each opcode (CREATE, CALL, SSTORE, SELFDESTRUCT), compute a single state diff at each call return point and charge based on actual state growth. This eliminates most edge cases from [Maria's state gas accounting review](https://github.com/misilva73/evm-gas-repricings/blob/main/reports/eip-8037/spec_review_state_gas_accounting.md) (points 1-6).

State gas charges at call return draw from the reservoir first, when exhausted, the remainder spills into regular gas, preserving the existing spillover semantics.

## Open Spec Questions & Resolution

**1. Who pays state gas for value transfer creating a new account?**

Resolution: the callee's call frame pays.

    A calls B (non-existent address) with value = 1 wei.
    The value transfer creates B's account inside B's call frame.
    At B's call return, the state diff shows +1 new account.
    State gas is deducted from the reservoir at B's return.
    If B's gas budget cannot cover GAS_NEW_ACCOUNT, the call fails.
    A sees failure (0 on stack).

Consequence: the caller must forward enough gas to cover state creation costs at the callee's return.

**2. How does code deposit failure work?**

Resolution: CREATE is all-or-nothing at call return.

    CREATE deploys a contract with 10KB of code.
    At CREATE's call return, the state diff shows:
      +1 new account
      +10240 bytes of deployed code

    If the reservoir covers the full cost: CREATE succeeds.
    If not: the entire CREATE reverts. No account, no code.

Consequence: no partial state where an account exists without code. The call either fully succeeds or fully reverts.

**3. Can the reservoir grow beyond its initial value?**

Resolution: yes, capped at transaction level.

    Contract clears a storage slot (nonzero to zero).
    At call return, the state diff shows -1 storage slot.
    The reservoir is credited by the slot's state gas cost.

    At transaction end: tx_gas_used = max(0, tx.gas - gas_left - reservoir).
    The floor at zero prevents the user from being paid to execute.

## Test changes

These tests assumed per-opcode state gas charging. With diff-at-return, state gas is computed from the net state diff at call return, which changes gas amounts, failure modes, and who pays.

**test_block_gas_refund_eip7778_no_block_reduction**: Updated expected gas_used. The set-then-restore SSTORE pattern (0 to 1 to 0) now produces zero state gas because the net state diff is zero. Previously, state gas was charged on the first write and partially refunded via refund_counter.

**test_call_value_transfer_existing_account_no_state_gas**: Changed target from empty account (amount=0) to alive account (amount=1) and forwarded all gas via Op.GAS. With diff-at-return, the callee pays state gas at its call return, so the callee needs enough gas to cover GAS_NEW_ACCOUNT if the target is not alive.

**test_caller_reservoir_preserved_after_callee_diff_reverts**: Replaces test_code_deposit_oog_preserves_parent_reservoir. The caller CALLs a factory that does CREATE with large code. The factory's state diff (account + code) exceeds its budget, so the factory reverts. The caller's reservoir is preserved and used for a subsequent SSTORE.

**test_parent_reverts_when_create_diff_exceeds_budget**: Replaces test_nested_create_code_deposit_cannot_borrow_parent_gas. A wrapper CALLs a factory that does CREATE deploying 15000 bytes. The factory's diff cost (~17.7M) exceeds its gas budget (~16.4M), so the factory reverts. The wrapper verifies the CALL returned 0 (failure).

166/166 blockchain_test passing (excluding pre-existing fork transition failures).

## Related

- Discord discussion: [#eip-8037](https://discord.com/channels/1359927674746835211/1493618116809457684) channel, April 15 2026
- [Maria's state gas accounting review](https://github.com/misilva73/evm-gas-repricings/blob/main/reports/eip-8037/spec_review_state_gas_accounting.md)
- EIP text clarifications: ethereum/EIPs#11523


spencer-tb:
Closing as no longer using this spec change!

PR #2687, opened 2026-04-15
feat(tests, spec-specs): add full cost per state byte pricing function for EIP-8037

spencer-tb:
### Summary

- Un-hardcode the `cost_per_state_byte` pricing function in both the spec and testing framework to compute the EIP-8037 quantized value from the block gas limit
- Add `BaseFork._env_gas_limit` class variable, synced automatically by the `fork` fixture from `Environment.gas_limit` (default 120M, or `--block-gas-limit`)
- All downstream gas methods (`gas_costs()`, `sstore_state_gas()`, etc.) respond to the actual block gas limit — no parameter threading needed
- Tests that set `Environment(gas_limit=X)` or use `--block-gas-limit` get the corresponding `cost_per_state_byte` automatically
- Add parametrized tests across 10 block gas limits (1M to 1G) for SSTORE, CREATE, CALL, SELFDESTRUCT, auth, and refunds
- Add underflow boundary tests at low gas limits where the quantized result falls below the offset, verifying the floor of 1

### Gas Limit Flow

1. **Session start**, filler plugin reads `EnvironmentDefaults.gas_limit` (default 120M, or `--block-gas-limit`) via the `env_gas_limit` fixture
2. **`fork` fixture**, creates a per-test fork variant via `fork.with_env_gas_limit(env_gas_limit)` and monkey-patches `Environment` so any `Environment(gas_limit=X)` syncs `X` back to the variant's `_env_gas_limit`
3. **Test code**, calls `fork.gas_costs()`, `fork.sstore_state_gas()`, etc. which read `cls._env_gas_limit` to compute cpsb
4. **Fill pipeline**, `state.py` / `blockchain.py` set `fork._env_gas_limit = env.gas_limit` per-block before t8n execution
5. **EELS spec**, `state_gas_per_byte(gas_limit)` receives the block gas limit from the environment and computes cpsb dynamically

### Fork Calculator Split

On top of this, Mario's refactor splits the previously combined opcode-gas surface on `BaseFork` into three separate per-dimension maps:

- `opcode_state_map` — state gas cost per opcode (cpsb multiplier or callable)
- `opcode_refund_map` — regular gas refund per opcode
- `opcode_state_refund_map` — state gas refund per opcode

Tests can now query each dimension independently instead of pulling a mixed regular+state value. Paired with the dynamic cpsb, opcode state-gas/refund calculations scale with the block gas limit automatically.

### Remaining Fixes

A separate follow-up commit (`feat(tests): remaining fixes`) addresses pre-existing tests that assumed the previously-hardcoded cpsb=1174 and broke under the new dynamic computation:

- `test_state_gas_reservoir.py`: pin `fork._env_gas_limit` in four tests so every cpsb-dependent value read matches what the block uses at execution time
- `test_bls12_variable_length_input_contracts.py`: apply the default env gas limit variant in the discount-table splitter so collection-time cpsb matches runtime

### Ported Static Test Fixes

A separate commit (`fix(tests): drop inflated env gas_limit in ported_static`) removes the `gas_limit=<value>` line from `Environment(...)` calls in 576 ported_static test files. These values were inherited as boilerplate from the original static JSON fillers (ranging from 1M to max-i64) with no functional purpose,  under the old hardcoded cpsb=1174 they didn't matter, but under dynamic cpsb they caused:

- **Inflated values (1B-100B, max-i64)**: cpsb ballooned to tens of thousands, making SSTORE 0 to nonzero OOG with the standard tx budget
- **Deflated values (1M-30M)**: cpsb dropped to 1–150, shifting gas consumption enough to break `Storage` / `Account.balance` assertions

Removing the override lets `Environment` default to `EnvironmentDefaults.gas_limit = 120M`, which yields the canonical cpsb=1174 these tests were always implicitly designed against. All 576 files verified to have a single `Transaction` with `tx.gas_limit <= 14M` and no references to `env.gas_limit` in test logic, so the change preserves test semantics.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```


kclowes (approved):
LGTM!

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -625,6 +625,7 @@ def generate_block_data(
             block_number=env.number, timestamp=env.timestamp
         )
         env = env.set_fork_requirements(fork)
+        fork.env_gas_limit = int(env.gas_limit)
```
marioevz, line 636: This is a bit too late because by this time the test's function will already been executed.

There's no easy solution to intercept fork and make sure that it's passed with the updated value to the test, but I think what we could do is modify the forks plugin here:
https://github.com/ethereum/execution-specs/blob/363df20673de012c69cd927c80faf8d22854fcd3/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py#L1237
and here:
https://github.com/ethereum/execution-specs/blob/363df20673de012c69cd927c80faf8d22854fcd3/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py#L1269
to parametrize the function with a different parameter name than `"fork"` (perhaps `"parametrized_fork"`) and then create a `fork` fixture in https://github.com/ethereum/execution-specs/blob/forks/amsterdam/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py that finally modifies the value.

marioevz: Another suggestion is that I would not modify `env_gas_limit` directly in the fork object because that could at some point cause side effects to other tests.

Instead I would create a new class method that basically returns a new class with the updated value.

marioevz (commented):
Just two preliminary comments, which I think we could even apply directly to `forks/amsterdam`.

marioevz:
@spencer-tb created https://github.com/ethereum/execution-specs/pull/2690 to address the comment I made, it's not a pretty PR but it does the job, and I did an ad-hoc (not in the PR) verification in the `specs/state.py` file by comparing `env.gas_limit` to `fork._env_gas_limit` and it was exactly the same for all tests. PTAL.

packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_8037.py
fselmo: @marioevz I think the recent refactoring here introduced a double accounting on this. This part is now covered by a separate call to the abstracted `_calculate_sstore_state_refund` in `opcode_refund_calculator`, right?

spencer-tb: Fixed in the updated "refactor(test-forks): Separate calculators for state/regular gas/refunds" commit:    [8033b56](https://github.com/ethereum/execution-specs/pull/2687/commits/8033b5625efbe75390a8a32f86d8fae640339d70)

spencer-tb:
@marioevz

Added remaining 8037 test fixes here, with TODOs, feel free to check over: https://github.com/ethereum/execution-specs/pull/2687/changes/62719d89677511e88459bf6bf98dd1e63639689a

Ported static fixes here (updated in PR desc): https://github.com/ethereum/execution-specs/pull/2687/commits/a3cbac3425430bac0fb766f0d41b3a1b7ca9931a

PR desc includes the fork calculator split. We can consolidate in future PR.

Will merge once CI finally passes...

PR #2689, opened 2026-04-15
feat(spec-specs, tests): EIP-8037 - zero execution state gas on top-level failure

kclowes:
## 🗒️ Description

On a top level transaction failure (revert, exceptional halt, or out of gas) no state persists, so execution state gas should not be billed. This PR updates the spec and tests so the consumed execution state gas is returned to the reservoir and `state_gas_used` is zeroed before the transaction fee is computed.

EIPs change for reference: ethereum/EIPs#11476

### Spec change: [ad40e40](https://github.com/kclowes/execution-specs/commit/ad40e40dc1)

`src/ethereum/forks/amsterdam/fork.py` `process_transaction`, after `process_message_call` returns, when `tx_output.error` is set: `tx_output.state_gas_left += tx_output.state_gas_used` and `tx_output.state_gas_used = 0`. The billing formula `tx.gas - gas_left - state_gas_left` then sees the full reservoir restored and refunds the sender. Intrinsic state gas is untouched.

### Existing fork fixups: [7a0358f](https://github.com/kclowes/execution-specs/commit/7a0358f9e0)

Updates two pre-EIP-8037 tests to branch on `fork.is_eip_enabled(8037)`. They assumed sender balance was fully consumed on OOG, which is no longer true once the state gas portion is refunded.

### New tests: [c00d017](https://github.com/kclowes/execution-specs/commit/c00d01756d)

Six tests in `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py`.

- `test_top_level_failure_refunds_execution_state_gas`, parametrized `failure_mode=[revert, halt, oog]`. Receipt `cumulative_gas_used` excludes the SSTORE state gas that was rolled back, proving the sender is refunded.
- `test_top_level_failure_zeros_block_state_gas`, parametrized `failure_mode=[revert, halt, oog]`. `blockchain_test` + `header_verify` asserting the block header `gas_used` equals only the regular gas component.
- `test_creation_tx_failure_preserves_intrinsic_state_gas`. A creation tx whose initcode halts carries intrinsic state gas for the new account; the top level failure refund zeroes only execution state gas and preserves the intrinsic component in block accounting.
- `test_subcall_failure_does_not_zero_top_level_state_gas`. Parent calls a child that reverts and then runs its own SSTORE. The top level tx succeeds, so the refund does not apply and the parent's state gas surfaces in the block header. Proves the refund is scoped to top level failures.
- `test_top_level_failure_refunds_spilled_state_gas`. Reservoir sized to half the SSTORE state gas so the other half spills into gas_left. Proves the refund covers both portions, not just the reservoir.
- `test_top_level_failure_refunds_state_gas_propagated_from_child`. A successful subcall runs SSTORE and returns to the parent, then the parent reverts. Proves the refund catches state gas propagated up via the child success path, not just direct top frame charges.

## 🔗 Related Issues or PRs

- EIP text: ethereum/EIPs#11476

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -1050,6 +1050,10 @@ def process_transaction(
 
     tx_output = process_message_call(message)
 
+    if tx_output.error is not None:
+        tx_output.state_gas_left += tx_output.state_gas_used
+        tx_output.state_gas_used = Uint(0)
+
     tx_gas_used_before_refund = (
         tx.gas - tx_output.gas_left - tx_output.state_gas_left
     )
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py
@@ -552,13 +552,14 @@ def test_creation_tx_failure_preserves_intrinsic_state_gas(
     fork: Fork,
 ) -> None:
     """
-    Verify a failed creation tx still counts intrinsic state gas.
+    Regression test for the creation tx failure path.
 
-    A creation tx (to=None) whose initcode halts carries intrinsic
-    state gas for the new account. The top level failure refund
-    zeroes execution state gas but preserves intrinsic state gas, so
-    the block header reflects the intrinsic state gas rather than
-    zero.
+    A creation tx (to=None) whose initcode halts exercises both the
+    intrinsic state gas for the new account and the top level failure
+    refund of execution state gas. The test asserts the block header
+    `gas_used` equals `max(block_regular, intrinsic_state_gas)`,
+    guarding that the failure path does not raise and that block
+    accounting does not underflow when the refund is applied.
     """
     gas_limit_cap = fork.transaction_gas_limit_cap()
     assert gas_limit_cap is not None
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py
@@ -28,6 +28,7 @@ from execution_testing import (
     Storage,
     Transaction,
     TransactionException,
+    TransactionReceipt,
 )
 from execution_testing.checklists import EIPChecklist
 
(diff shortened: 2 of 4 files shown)


jochem-brouwer:
If execution fails (the computation goes out of gas) then the EIP-7702 authorizations still go through, and if an account does not exist, this account is created (so this is related to state cost). However I think in EIP-8037 this is taken from the intrinsic cost, and although this does create state, I think setting state gas used to 0 is correct in all cases... except maybe if the account is already there? Because then you would get a refund added to state gas, all regular gas is spent, but I think you should in the end get this refund back :thinking: 

As a side note, due to the base fees the sender account is guaranteed to exist (otherwise it could not pay for the transaction). But coinbase could be empty prior to the transaction. The gas fees paid for executing the failure could create the coinbase (although this is currently free, and should stay like so)

kclowes:
This adds a whole bunch of new fails, so converting back to draft for now.

spencer-tb (approved):
LGTM! Feel free to add more coverage!

kclowes:
@jochem-brouwer I'm not sure I understand your comment above. Are you opening these points for discussion or do you feel that there are changes that need to happen on this PR?

spencer-tb:
> If execution fails (the computation goes out of gas) then the EIP-7702 authorizations still go through, and if an account does not exist, this account is created (so this is related to state cost). However I think in EIP-8037 this is taken from the intrinsic cost, and although this does create state, I think setting state gas used to 0 is correct in all cases... except maybe if the account is already there? Because then you would get a refund added to state gas, all regular gas is spent, but I think you should in the end get this refund back 🤔
> 
> As a side note, due to the base fees the sender account is guaranteed to exist (otherwise it could not pay for the transaction). But coinbase could be empty prior to the transaction. The gas fees paid for executing the failure could create the coinbase (although this is currently free, and should stay like so)

Yes I think you get the refund back, this change only touches `state_gas_used`, so the 7702 existing account refund sits in `state_gas_left` and survives the failure path, getting refunded to the sender

My understanding here is the interaction between point 1 (this pr) and point 6 from marias doc, we'll have a seperate PR up for interactions between each point. This PR only targets point 1 spec and tests specifically :)

spencer-tb:
CI fails are expected static fails, and mirror issues on devops runners! Merging :D

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -1050,6 +1050,10 @@ def process_transaction(
 
     tx_output = process_message_call(message)
 
+    if tx_output.error is not None:
+        tx_output.state_gas_left += tx_output.state_gas_used
+        tx_output.state_gas_used = Uint(0)
```
jochem-brouwer, line 1055: I believe this is incorrect: if an EIP-7702 account is delegated and the account we are delegating already exists, we get a state gas refund which should propagate here. AFAIK all other state updates are triggered by execution, so indeed if there is an error this should be subtracted.

The gas calculation for the block is now incorrect: in case of a EIP-7702 delegation the state gas used of the block should increment. This is here set to 0.

For instance the case where the tx fails vs tx does not fail (either call an account with INVALID/REVERT(0,0) as opcode and where the tx creates an EIP-7702 account `PER_EMPTY_ACCOUNT_COST` is paid in state gas (or at least the `PER_AUTH_BASE_COST`). However via current code this sets block state gas to 0 and execution gas to the intrinsic gas cost (plus any cost for the revert / or all gas in case of invalid). This fails in case state gas > execution gas.

jochem-brouwer: Ok I see where this comes from (from EIP-8037)

> On child exceptional halt, the child’s gas_left is additionally consumed (zeroed).

This is incorrect in the EIP-7702 case, see [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702#set-code-transaction) (see Behavior section):

> Note, if transaction execution results in failure (e.g. any exceptional condition or code reverting), the processed delegation indicators is not rolled back.

kclowes: @jochem-brouwer I think this is right. The 7702 interactions are handled in `forks/amsterdam/vm/eoa_delegation.py`. I put up PR #2722 to add some tests around these scenarios. Let me know here or there if those look wrong or if I misunderstood something!

PR #2693, opened 2026-04-16
feat(test-benchmark): accept float values for `--gas-benchmark-values` flag

### LINKED ISSUE #2673: Speed-up `bench-gas` check
Currently, the `just bench-gas` check is the  bottleneck in CI at 18 minutes. 

One possible solution is to reduce the `--gas-benchmark-values` flag from `1.0` to `0.5` (or even `0.2`)? Perhaps this causes problems with tests, we should try it.

Currently, changing this results in this pydantic validation error:
```
INTERNALERROR> pydantic_core._pydantic_core.ValidationError: 1 validation error for GasBenchmarkValues
INTERNALERROR> 0
INTERNALERROR>   Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='0.5', input_type=str]
INTERNALERROR>     For further information visit https://errors.pydantic.dev/2.12/v/int_parsing
```
which can fix by changing the type from int to float > 0.001 (or something else more reasonable) - this flag is measured in Mgas.

mandeepmourya007:
Accept fractional Mgas values in --gas-benchmark-values to speed up the bench-gas CI check. Changes GasBenchmarkValues.root from List[int] to 
  List[Annotated[float, Gt(0)]], adds int() cast at the multiplication boundary so all downstream code still receives int, updates
  format_gas_limit_prefix to handle float formatting, and reduces the Justfile default from 1 to 0.5.                                           
                                                            
  🔗 Related Issues or PRs

  Fixes #2673                                                                                                                                   
   
  ✅ Checklist                                                                                                                                  
                                                            
  - All: Ran fast static checks to avoid unnecessary CI fails, see also Code Standards and Enabling Pre-commit Checks:                          
  just static
  - All: PR title adheres to the repo standard - it will be used as the squash commit message and should start type(scope):.                    
  - All: Considered updating the online docs in the ./docs/ directory.                                                                          
  - All: Set appropriate labels for the changes (only maintainers can apply labels).                                                            
  - Tests: Ran mkdocs serve locally and verified the auto-generated docs for new tests in the Test Case Reference are correctly formatted.      
  - Tests: For PRs implementing a missed test case, update the post-mortem document to add an entry the list.                                   
  - Ported Tests: All converted JSON/YML tests from ethereum/tests or tests/static have been assigned @ported_from marker.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -219,7 +219,7 @@ bench-gas *args:
     @mkdir -p "{{ output_dir }}/bench-gas/tmp" "{{ output_dir }}/bench-gas/logs"
     uv run fill \
         --evm-bin="{{ evm_bin }}" \
-        --gas-benchmark-values 1 \
+        --gas-benchmark-values 0.5 \
         --generate-pre-alloc-groups \
         --fork Osaka \
         -m "not slow" \
--- a/Justfile
+++ b/Justfile
@@ -219,7 +219,7 @@ bench-gas *args:
     @mkdir -p "{{ output_dir }}/bench-gas/tmp" "{{ output_dir }}/bench-gas/logs"
     uv run fill \
         --evm-bin="{{ evm_bin }}" \
-        --gas-benchmark-values 0.5 \
+        --gas-benchmark-values 1 \
         --generate-pre-alloc-groups \
         --fork Osaka \
         -m "not slow" \
--- a/docs/running_tests/releases.md
+++ b/docs/running_tests/releases.md
@@ -52,8 +52,8 @@ follow the same layout.
 
 When filling with `--gas-benchmark-values`, benchmark tests additionally
 include the gas limit in the subdirectory name (`for_{fork}_at_{gas}M`,
-where `{gas}` is in millions, zero-padded to four digits), with one
-subdirectory per gas value:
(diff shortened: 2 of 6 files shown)


LouisTsai-Csie:
Thank you for this PR. I have some feedback:

- Unit tests: Please add unit tests to `test_benchmark.py`, you could check `test_benchmarking_mode_configured_with_option` as an example.
- CI gas limit: The CI fails because 0.5M is too low for some benchmarks. We need to decide whether to (1) skip these benchmarks in CI, or (2) raise the gas limit to 0.7M or higher.
- Precision loss: The current `format_gas_limit_prefix` design inconsistent for 0.25M (it becomes `for_osaka_at_000_0.2M`). We should discuss the desired granularity here.

cc @danceratopz

danceratopz:
@LouisTsai-Csie you mentioned elsewhere, that, unfortunately, once we update the benchmark tests and framework for Amsterdam sub 1M gas benchmark values will no longer be viable as it will break generation of many tests.

In which case, I think we can close this PR and the corresponding issue as it makes little sense to add this. Sorry about that @mandeepmourya007; thanks for your efforts here! :pray:

PR #2694, opened 2026-04-16
feat(test-benchmark): support fractional gas benchmark values

### LINKED ISSUE #2673: Speed-up `bench-gas` check

shrirajpawar4:
## Summary

  Fixes #2673 by allowing fractional Mgas inputs for `--gas-benchmark-values`, while keeping the benchmark/test pipeline integer-based after parsing.

  The issue is not just “accept `0.5`”. The important part is to preserve benchmark semantics downstream:
  - benchmark tests consume `gas_benchmark_value` as an absolute gas budget in gas units
  - many tests use integer arithmetic and exact gas validation
  - fixture IDs and output subdirectories need to represent fractional Mgas values consistently

  ## What changed

  - Parse `--gas-benchmark-values` as decimal Mgas inputs and convert them immediately to integer gas-unit budgets
  - Keep `gas_benchmark_value` integer-based for downstream benchmark logic
  - Add readable benchmark IDs such as `benchmark-gas-value_0.5M`
  - Update gas benchmark subdirectory formatting to support fractional values while staying sortable
  - Change `just bench-gas` from `--gas-benchmark-values 1` to `0.5`
  - Update benchmark docs/examples to document fractional Mgas inputs

  ## Why this approach

  Using raw `float` values all the way through is the wrong boundary for this repo. The safe place to handle fractions is the CLI/input layer; below that, the code should continue to operate on integer
  gas budgets.

  This keeps the fix scoped to the benchmark tooling and avoids changing benchmark test semantics under `tests/benchmark/**`.

  ## Validation

  Ran targeted benchmark plugin and collection tests locally:

  ```bash
  PYTHONPATH=src:packages/testing/src .venv/bin/python -m pytest \
    packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_benchmarking.py \
    -k 'gas_benchmark_values_valid_input or gas_benchmark_values_invalid_input or gas_benchmark_value_formatting or benchmarking_mode_configured_with_option'

  PYTHONPATH=src:packages/testing/src .venv/bin/python -m pytest \
    packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py \
    -k 'benchmark_conftest_matrix and gas-benchmark-values'

  PYTHONPATH=src:packages/testing/src .venv/bin/python -m pytest \
    packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_benchmarking.py \
    -k 'fractional_benchmark_gas_values_split_into_subdirs'

  Results:

  - 10 passed, 38 deselected
  - 8 passed, 23 deselected
  - 1 passed, 47 deselected
```
 
 I did not run the full just bench-gas recipe locally because that needs the benchmark evm binary available in PATH / EVM_BIN.

### DIFF
--- a/docs/templates/base.md.j2
+++ b/docs/templates/base.md.j2
@@ -10,7 +10,7 @@ Documentation for [`{{ pytest_node_id }}@{{ short_git_ref }}`]({{ source_code_ur
 
     ```console
 {% if is_benchmark %}
-    fill -v {{ pytest_node_id }} --gas-benchmark-values 1
+    fill -v {{ pytest_node_id }} --gas-benchmark-values 0.5
 {% else %}
     fill -v {{ pytest_node_id }} --fork {{ target_or_valid_fork }}
 {% endif %}
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
@@ -2,6 +2,7 @@
 
 import re
 from abc import ABC, abstractmethod
+from decimal import Decimal, InvalidOperation
 from pathlib import Path
 from typing import ClassVar, Dict, List, Self
 
@@ -19,6 +20,8 @@ from .fixture_output import (
     format_gas_limit_prefix,
 )
 
+GAS_PER_MEGAGAS = Decimal("1000000")
+
 
 def pytest_addoption(parser: pytest.Parser) -> None:
     """Add command line options for benchmark tests."""
@@ -33,7 +36,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default=None,
         help=(
             "Gas limits (in millions) for benchmark tests. "
-            "Example: '100,500' runs tests with 100M and 500M gas. "
+            "Example: '0.5,1,2' runs tests with 0.5M, 1M, and 2M gas. "
             "Benchmark outputs are grouped under "
             f"{FORK_SUBDIR_PREFIX}{{fork}}"
             f"{SUBFOLDER_LEVEL_SEPARATOR}XXXXM/ subdirectories. "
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: docs/templates/base.md.j2 [hunk 1]
```diff
@@ -10,7 +10,7 @@ Documentation for [`{{ pytest_node_id }}@{{ short_git_ref }}`]({{ source_code_ur
 
     ```console
 {% if is_benchmark %}
-    fill -v {{ pytest_node_id }} --gas-benchmark-values 1
+    fill -v {{ pytest_node_id }} --gas-benchmark-values 0.5
```
danceratopz, line 13: Just a comment, I would avoid inflating the changeset in a PR across too many files unnecessarily. An example might be worth including, but I think this includes too many doc changes for what is otherwise, a relatively small change.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
```diff
@@ -204,6 +211,47 @@ def get_test_parameters(self, _test_name: str) -> list[ParameterSet]:
         ]
 
 
+def _parse_gas_limit_millions(value: str) -> int:
```
danceratopz, line 214: Ah I think this approach is too verbose. I believe a better idea, would be to update the pydantic model above in L77, see here:

https://github.com/ethereum/execution-specs/blob/363df20673de012c69cd927c80faf8d22854fcd3/packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py#L174-L204

danceratopz (commented):
Hey @shrirajpawar4 thanks a lot for giving this a go. I think this PR went down the wrong path. Please check my comment about updating the pydantic model instead - it would be easier to change the type there from `int` to `float` which would enable pydantic to parse these values correctly, preferably with a reasonable / heuristically determined minimum value that gets checked in a validation step.

As #2693 landed (just) before this PR and is heading down the wrong path, I will close this PR and focus on #2693. I'm sorry about that, bad luck with timing. If you'd like to tackle a different issue, feel free to reach out.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/benchmarking.py
danceratopz: I.e, change the pydantic model to allow a `List[float]` instead of a `List[int]` in L177.

shrirajpawar4: I agree the original diff was broader. The reason I did not stop at changing `GasBenchmarkValues` to `List[float]` is that I think the parsing error is only one part of the problem

 The benchmark pipeline internally works with integer gas-unit budgets, not floats. If we accept floats directly and then do int(value * 1_000_000), we silently truncate and can also end up with
  inconsistent benchmark IDs. 
  
  please correct me if i am wrong this will help me understand the system better.

PR #2701, opened 2026-04-16
fix(cli): add pytest EIP markers for valid_from tests

### LINKED ISSUE #2675: Add EIP marker for better DevEx
We've added some very nice EIP-based logic in our codebase - things like `fork.is_eip_enabled(...)` and `pytest.mark.valid_from(EIP1234)`. It would be very nice if we could auto-add an EIP marker to tests that are marked valid from certain EIPs so we could fill tests for that EIP with something like:

```
uv run fill --clean --fork=Amsterdam  -m EIP1234
```

This is a remaining thought from implementing EIP-7928 and having seen the EIP logic be merged into the code base. It made sense to add BAL expectations to older, more complex, tests in some cases. This would ideally make it so we can catch all the tests marked with the EIP7928 marker with a single `-m` command, for example, rather than looking for them and amassing some complex catch-all `-k` command.

shrirajpawar4:
## 🗒️ Description

  Auto-add plain pytest `EIP####` markers for tests that are explicitly enabled by `valid_from("EIP####")`.

  This makes EIP-scoped selection work as requested in #2675, so commands like:

  ```console
  uv run fill --clean --fork=Amsterdam -m EIP7928
```
  can select tests that are marked valid from a given EIP without relying on ad hoc -k expressions.

  This change is intentionally narrow:

  - It promotes explicit valid_from("EIP####") markers into plain pytest markers during collection.
  - It does not treat negative selectors such as valid_before("EIP####") as matching -m EIP####.
  - It registers known EIP#### markers so they are valid pytest markers.

  It also adds targeted regression tests covering:

  - function-level valid_from("EIP####")
  - param-level valid_from("EIP####")
  - the negative case where valid_before("EIP####") must not match -m EIP####

  ## 🔗 Related Issues or PRs

  Fixes #2675.

  ## ✅ Checklist

  - [ ] All: Ran fast static checks to avoid unnecessary CI fails, see also Code Standards (https://eest.ethereum.org/main/getting_started/code_standards/) and Enabling Pre-commit Checks
    (https://eest.ethereum.org/main/dev/precommit/):

    just static
  - [x] All: PR title adheres to the repo standard (https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit mes
    sage and should start type(scope):.
  - [ ] All: Considered updating the online docs in the /ethereum/execution-specs/blob/HEAD/docs/ directory.
  - [ ] Tests: Ran mkdocs serve locally and verified the auto-generated docs for new tests in the Test Case Reference (https://eest.ethereum.org/main/tests/) are correctly formatted.
  - [ ] Tests: For PRs implementing a missed test case, update the /ethereum/execution-specs/blob/HEAD/docs/writing_tests/post_mortems.md to add an entry the list.
  - [ ] Ported Tests: All converted JSON/YML tests from /ethereum/tests or /ethereum/execution-specs/blob/HEAD/tests/static have been assigned @ported_from marker.

  ### Verification

  Manual behavior check against a real test file:
```
  uv run pytest \
    -c packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-fill.ini \
    tests/frontier/validation/test_header.py \
    --until=Amsterdam \
    --collect-only -q -m EIP7928
```
  Observed result:

  - collected test_gas_limit_below_minimum[...,gas_limit_5000_1]
  - did not collect ...,gas_limit_5000_0

  This confirms -m EIP7928 selects the valid_from("EIP7928") case and not the valid_before("EIP7928") case.

  Targeted regression tests:
```
  uv run pytest \
    packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/tests/test_markers.py \
    -q \
    -k 'function_level_valid_from_eip_is_selectable_by_mark or param_level_valid_from_eip_is_selectable_by_mark or negative_eip_selectors_do_not_add_eip_markers'
```
  Observed result:

  3 passed

  #### Cute Animal Picture

  image

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
@@ -1494,7 +1494,9 @@ def blob_params_changed_at_transition(fork: Fork | TransitionFork) -> bool:
     return False
 
 
-def get_valid_from_eip_marker_names(markers: Iterator[pytest.Mark]) -> Set[str]:
+def get_valid_from_eip_marker_names(
+    markers: Iterator[pytest.Mark],
+) -> Set[str]:
     """
     Return explicit EIP names from ``valid_from`` markers.
 
@@ -1577,13 +1579,7 @@ def pytest_collection_modifyitems(
     filter_stats: Dict[str, Tuple[str, int, int]] = {}
 
     for i, item in enumerate(items):
-        try:
-            add_explicit_eip_markers(item)
-        except Exception as e:
-            pytest.exit(
-                f"Error in test '{item.name}': {e}",
-                returncode=pytest.ExitCode.USAGE_ERROR,
-            )
+        add_explicit_eip_markers(item)
 
         params = _get_item_params(item)
         if not params:
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
@@ -38,6 +38,7 @@ from execution_testing.forks import (
     get_transition_forks,
     transition_fork_to,
 )
+from execution_testing.forks.forks.eips import ALL_EIPS
 from execution_testing.logging import (
     get_logger,
 )
@@ -513,6 +514,14 @@ def pytest_configure(config: pytest.Config) -> None:
             "shown in verbose collection output."
         ),
     )
+    for eip in sorted(ALL_EIPS):
+        config.addinivalue_line(
+            "markers",
+            (
+                f"{eip.name()}: tests explicitly enabled by "
+                f"{eip.name()} via valid_from(...)"
+            ),
+        )
     for d in fork_covariant_decorators:
         config.addinivalue_line("markers", f"{d.marker_name}: {d.description}")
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by shrirajpawar4** [comment]
@fselmo do take a look at PR whenever you can sir

**Comment by shrirajpawar4** [comment]
@danceratopz can you take a look at this whenever possible?

File: packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/tests/test_markers.py [hunk 1]
```diff
@@ -521,3 +521,128 @@ def test_param_level_validity_markers(
         *pytest_args,
     )
     result.assert_outcomes(**outcomes)
+
+
+def generate_function_level_eip_marker_test() -> str:
```
marioevz: These functions seem unnecessary since they only return a constant string.
Might be better to put the string in-place.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/tests/test_markers.py
```diff
@@ -521,3 +521,128 @@ def test_param_level_validity_markers(
...
+@pytest.mark.state_test_only
+def test_negative_eip_marker(state_test):
+    pass
+"""
+
+
+def test_function_level_valid_from_eip_is_selectable_by_mark(
+    pytester: pytest.Pytester,
+) -> None:
+    """``-m EIP####`` should select tests enabled via function-level valid_from."""
```
marioevz: This fails with `E501 Line too long`. Try running `just static` to catch all lint issues.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
```diff
@@ -1543,6 +1577,14 @@ def pytest_collection_modifyitems(
     filter_stats: Dict[str, Tuple[str, int, int]] = {}
 
     for i, item in enumerate(items):
+        try:
+            add_explicit_eip_markers(item)
+        except Exception as e:
```
marioevz: Is any test currently failing like this or why is this required?

marioevz (commented):
Looks mostly good to me. I left a couple of comments that need to be addressed before merging. Thanks!

fselmo:
@marioevz part of my goal for the original issue linked here was to be able to collect any test that defines the following as an EIP-7928 test:

```py
if fork.is_eip_enabled(7928):
    # define a block_access_list_expectation and pass it to the test class
```

I think there are a couple of ways to do this but perhaps the most straightforward is something like a `@pytest.mark.eip(7928)` marker or something like this. This is different than a `valid_from` in the sense that we can add expectations to any test. And it's useful to add expectations to older tests because some of them are very complex to rewrite just to define a BAL expectation for them - other than the obvious DRY benefits.

Curious on your thoughts here ^ but I'd love to address this in this PR. It would go a long way into identifying all the EIP test cases that are defined for any variant of testing the EIP - not just if the test is only valid from that EIP but anything that validates the test in relation to the EIP, more generally.

PR #2702, opened 2026-04-16
feat(ci, test-fill): make `--generate-all-formats` optional via `feature.yaml`

spencer-tb:
## 🗒️ Description

Tarball output (`.tar.gz`) no longer auto applies `--generate-all-formats`. Release features declare the flag explicitly in the `fill-params` for only `benchmark` and `benchmark_fast`  features.

### Why?

TLDR; faster devnet releases during the interop.

This means both `mainnet` and `bal` releases will no run the `--generate-all-formats` flag, speeding up releases. We don't use any formats other than state, block and engine fixtures currently. Forcing us add enginex formats currently just add more time onto releases given that we don't use `consume enginex` yet. I propose that we add this flag back once 1) we are actively using `consume enginex` and 2) release with the `--generate-all-formats` flag is significantly faster.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/.github/actions/build-fixtures/action.yaml
+++ b/.github/actions/build-fixtures/action.yaml
@@ -44,7 +44,7 @@ runs:
 
         if [ -n "$IS_SPLIT" ]; then
           OUTPUT_ARG="--output=fixtures_${{ inputs.release_name }}"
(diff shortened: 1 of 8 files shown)


felix314159 (approved):
ty!

PR #2713, opened 2026-04-18
feat(specs,tests): EIP-7928 - Update BAL index to uint64

nerolation:
As discussed in discord, we should update the BAL index to use `uint64` instead of `uint16` to be future proof, having a large enough range to support a 300 million gas limit + EIP-2780 to be shipped.


spencer-tb (approved):
LGTM! Would like to get in the next release!

@fselmo, flag if there are any issues from your side :)

fselmo (commented):
lgtm 👍🏼

PR #2720, opened 2026-04-19
feat(test-fill): enable phase-1-only pre-alloc generation

danceratopz:
## 🗒️ Description

Splits the overloaded `--generate-pre-alloc-groups` flag so it now runs **phase 1 only**: populate the `pre_alloc/` folder and stop. This enables splitting fixture generation across runners (phase 1 on one, phase 2 via `--use-pre-alloc-groups` on another) without requiring `--generate-all-formats` as a wrapper.

Changes:

- `fill.py`: `--generate-pre-alloc-groups` alone creates a phase-1-only execution; `--generate-all-formats` remains the two-phase driver.
- `fill.py`: new `_validate_flag_combinations` rejects `--use-pre-alloc-groups` combined with `--generate-pre-alloc-groups` or `--clean` up-front with a `click.UsageError`, instead of failing later with a `FileNotFoundError` from `_load_pre_alloc_groups_from_folder`.
- `Justfile`: `bench-gas` switched to `--generate-all-formats` since `--generate-pre-alloc-groups` no longer two-phases.
- Docs: refreshed `filling_tests_command_line.md` and `blockchain_test_engine_x.md` to match the new semantics.
- Tests: cover the new phase-1-only path and both rejection paths.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    just static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).

#### Cute Animal Picture

:turtle:

PR #2721, opened 2026-04-19
feat(test-split): add grouped-split pytest plugin

danceratopz:
> [!IMPORTANT]
> Requires #2720 (`refactor(test-fill): allow phase-1-only pre-alloc generation`) to land first. Do not merge until that PR is in.

## 🗒️ Description

Add `pytest-split` integration to `fill` so a CI release build can fan fixture generation out across multiple runners while keeping each per-test-function fixture file owned by a single runner.

The new `--grouped-split` flag replaces `pytest-split`'s default splitter with a `(test_function_path, fork)`-aware algorithm. With this, every parametrization of one test function under one fork is guarenteed to land on the same runner.

Because fill's native output layout already groups per-function fixtures into one file per `(fork, function)`, this invariant makes plain `cp -r` fan-in safe: There's no need to perform a JSON-level merge to generate fixture artifacts.

**What lands:**

- `packages/testing/src/execution_testing/cli/pytest_commands/plugins/split/`: the new plugin, split into four single-concern modules.
    - `durations.py`: single source of truth for `.test_durations` handling. `strip_xdist_suffix` preserves non-`t8n-cache` `xdist_group` markers (e.g. `@bigmem`) and `@` inside parametrize values (`email@example.com`), matching `filler.py:_strip_xdist_group_suffix`. Also exports `normalize_durations`, `merge_durations`, `load_durations`, `write_durations`.
    - `grouping.py`: correctness-only module. `group_key(item)` returns the `(function_path, fork)` key. Reads the fork from `item.callspec.params["parametrized_fork"]` (the forks plugin's canonical param) when available, so a parametrize value that happens to start with `fork_` cannot hijack the key. Falls back to a nodeid `fork_*` token scan only for items without a callspec.
    - `scheduling.py`: performance-only module, three public functions composed by `assign_runners`.
        - `build_group_durations` groups keyed items and totals per-group duration (unknown items inherit the known-items mean).
        - `lpt_schedule` runs Longest-Processing-Time-first bin-packing across runners: sort groups heaviest-first, assign each to the least-loaded runner via a min-heap. Standard 4/3-approximation for makespan minimization.
        - `sort_items_within_groups` orders each group's items by individual duration DESC so that xdist workers start on the slow tests first; reduces end-of-run stragglers within one runner's `-n N` worker pool. Default-on, disabled via `assign_runners(..., sort_intra_group=False)`.
    - `plugin.py`: pytest hooks + summary plumbing only. Adds `--grouped-split`; when combined with `--splits` and `--group`, unregisters upstream `pytestsplitplugin` and partitions collected items by `group_key`. Emits a compact summary whose first line classifies the run as `mode: duration-aware (...)`, `mode: average-only (no durations ...)`, or `mode: average-only (... KEY MISMATCH)`, plus a GitHub Actions `::warning::` annotation when durations load but zero items match.
- `.github/scripts/{normalize_durations,merge_durations_files,diagnose_durations}.py`: thin CI shims over `durations.py` so workflows can call the helpers without duplicating JSON-handling logic.
- `packages/testing/pyproject.toml`: pin `pytest-split==0.11.0` exactly. The plugin unregisters by upstream's internal plugin name `pytestsplitplugin`, an implementation detail that could shift across minor releases.

**Design notes:**

- Two orthogonal concerns inside the same plugin. `grouping.py` enforces the correctness invariant (items with the same key share a runner, which is what makes plain-copy fan-in safe). `scheduling.py` handles the performance concerns: LPT across CI runners, and slowest-first within each group for intra-runner xdist workers. Swapping in a different per-group distribution rule would keep fan-in safe and only affect wallclock.
- Authoritative fork lookup. `group_key` reads the fork from `item.callspec.params["parametrized_fork"]`, the forks plugin's canonical param name (not a nodeid-token heuristic). A parametrize value like `fork_candidate` is no longer ambiguous with a real fork param.
- Intra-group ordering respects xdist. Under xdist's default scheduler settings (`--dist=load` or `--dist=loadgroup` without `--loadscopereorder`) the order set via `items[:]` is preserved to worker dispatch, verified against xdist 3.8's `LoadScopeScheduling.schedule()` source. `--loadscopereorder` is the one setting that would re-sort scopes by item count and override our order; not used in this repo's CI.
- Plugin summary travels from xdist workers to the controller via `workeroutput` (`pytest_sessionfinish` on workers, `pytest_testnodedown` on the controller) so the terminal summary emits correctly under `--dist loadgroup`.
- End-to-end verified against a non-split baseline. `fill --generate-all-formats` (no plugin) vs. `fill --grouped-split` (N groups merged by plain file copy) produces byte-equivalent output per `uv run hasher compare` across every fixture format: `blockchain_tests`, `blockchain_tests_engine`, `state_tests`, and `blockchain_tests_engine_x` per fork. The harness lives at `tmp/verify_split_matches_baseline.py` (gitignored). It also asserts the plugin's `duration coverage N/M` is non-zero, so a silent fallback to average-based splitting fails the check.

## 🔗 Related Issues or PRs

Requires #2720.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/.github/scripts/diagnose_durations.py
+++ b/.github/scripts/diagnose_durations.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+"""
+Diagnose a pytest-split ``.test_durations`` file.
+
+Report entry count, whether keys still carry ``@xdist_group`` suffixes
+(indicating incomplete normalization), and print sample keys for
+quick visual comparison against collected test nodeids.
+
+Usage::
+
+    uv run python .github/scripts/diagnose_durations.py [path]
+
+*path* defaults to ``.test_durations`` in the current directory.
+"""
+
+import sys
+from pathlib import Path
+
+from execution_testing.cli.pytest_commands.plugins.split.durations import (
+    load_durations,
+)
+
+
+def main() -> None:
+    """Entry point."""
+    path = Path(sys.argv[1]) if len(sys.argv) > 1 else Path(".test_durations")
+    if not path.exists():
+        print(f"::warning::No durations file at {path}")
+        return
+
+    data = load_durations(path)
+    has_at = sum(1 for k in data if "@" in k)
+    keys = sorted(data)
+    abs_path = path.resolve()
+
+    print(f"Durations file: {path}")
+    print(f"  Entries: {len(data)}")
+    print(f"  Keys with @ suffix: {has_at}/{len(data)}")
+    if has_at:
+        print(
+            f"  WARNING: {has_at} keys still have @ suffixes"
+            " - normalization may have failed"
+        )
+
+    for label, sample in (
+        ("First 3 keys:", keys[:3]),
+        ("Last 3 keys:", keys[-3:]),
+    ):
+        print(f"  {label}")
+        for k in sample:
+            print(f"    {k}: {data[k]:.2f}s")
+
+    print(f"  Absolute path: {abs_path}")
+    print(f"  File size: {abs_path.stat().st_size} bytes")
+
+
+if __name__ == "__main__":
+    main()
--- a/.github/scripts/merge_durations_files.py
+++ b/.github/scripts/merge_durations_files.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+"""
+Merge multiple pytest-split ``.test_durations`` files.
+
+Accept an output path and one or more input ``.test_durations`` JSON
+files and flat-merge them into one file. Splits produce disjoint test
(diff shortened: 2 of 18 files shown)


danceratopz:
This would be a nice feature to ship, but let's defer and close it for now: Releases are already much faster. 

This could be a relevant approach as newer forks land, perhaps even for Heggie. But we should optimize `fill` with for `blockchain_test_engine_x` first and then re-evaluate.

PR #2738, opened 2026-04-21
fix(evm_tools): avoid cloning unchanged fork template in ForkCache

### LINKED ISSUE #1864: Avoid creating a temporary fork clone when nothing has changed from the template
As an optimization to `ForkCache.get`, we should avoid creating a cloned fork if the clone would be identical to the template.

_Originally posted by @SamWilsn in https://github.com/ethereum/execution-specs/pull/1789#discussion_r2593528052_

JackCC703:
## 🗒️ Description

- Optimized `ForkCache.get(...)` to avoid creating a temporary fork clone when requested overrides are identical to the template.
- Added `_ForkOverrides` to organize temporary override values internally.
- Added `_template_matches(...)` to compare requested values against template defaults before cloning.
- Kept conservative fallback behavior: if `vm.gas` or `GasCosts` cannot be read, the request is treated as non-identical and still clones.
- Preserved existing behavior for changed values: still clone and cache as before.
- Expanded test coverage in `tests/evm_tools/test_fork_cache.py` to include:
  - no-override fast path returns template
  - fork criteria change triggers clone
  - per-override identical vs changed behavior (parameterized)
  - repeated changed request reuses cached clone
  - missing `vm.gas` / `GasCosts` falls back to cloning

## 🔗 Related Issues or PRs

Fixes #1864.

## ✅ Checklist

    ```console
    just static
    ```
    (Executed the equivalent `just static` command set directly via `uv run ...` because `just` is not available in local environment.)

#### Cute Animal Picture

Cute dog


src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -113,6 +113,77 @@ def __exit__(self, *args: object, **kwargs: object) -> None:
...
+        fork_mod = template.module("fork")
+
+        try:
+            gas_mod = template.module("vm.gas")
+        except ModuleNotFoundError:
+            gas_mod = None
+
+        gas_costs = getattr(gas_mod, "GasCosts", None)
+
+        def gas_default(
```
SamWilsn, line 141: My naive implementation would be to do something like:

```python
GasCosts = template.module("vm.gas").GasCosts

if gas_per_blob is not None:
    if gas_per_blob != getattr(GasCosts, "per_blob", None):
        return False

...
```

I assume this is insufficient for some reason. Why?

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -140,7 +211,17 @@ def get(
...
+        if all(x is None for x in cache_key[1:]) or self._template_matches(
+            template=template,
+            fork_criteria=fork_criteria,
+            blob_target_gas_per_block=blob_target_gas_per_block,
+            gas_per_blob=gas_per_blob,
+            blob_min_gasprice=blob_min_gasprice,
+            blob_base_fee_update_fraction=blob_base_fee_update_fraction,
+            max_blob_gas_per_block=max_blob_gas_per_block,
+            blob_schedule_target=blob_schedule_target,
+            blob_schedule_max=blob_schedule_max,
```
SamWilsn, line 223: I think it's probably time to introduce a dataclass to organize this pile of arguments 🤣

tests/evm_tools/test_fork_cache.py
```diff
@@ -0,0 +1,267 @@
...
+    if isinstance(criteria, ByTimestamp):
+        return ByTimestamp(int(criteria.timestamp) + 1)
+
+    if isinstance(criteria, Unscheduled):
+        zero_order = Unscheduled(order_index=0)
+        if criteria == zero_order:
+            return Unscheduled(order_index=1)
+        return zero_order
+
+    raise AssertionError(f"unsupported fork criteria type: {type(criteria)}")
```
SamWilsn, line 59: Should use `assert_never` here if possible.

tests/evm_tools/test_fork_cache.py
```diff
@@ -0,0 +1,267 @@
...
+
+def _gas_default(
+    gas_mod: Any,
+    module_attr: str,
+    gas_costs_attr: str | None = None,
+) -> U64 | Uint | None:
+    """Return a gas default from module-level or `GasCosts` attributes."""
+    value = getattr(gas_mod, module_attr, None)
+    if value is not None:
+        return cast(U64 | Uint, value)
```
SamWilsn, line 70: Should we use asserts here instead of casts?

SamWilsn (commented):
How does this affect test runs? Worthwhile speedup I assume?

JackCC703:
> How does this affect test runs? Worthwhile speedup I assume?

Yes, looks worthwhile for the case this PR is targeting. When the requested blob overrides are identical to the template defaults, the old path still paid the full `Hardfork.clone()` cost. On my macbook that was about `23.1-23.3s` per fresh cache, versus about `0.37s` warm (`0.67s` cold) now since it just returns the template.

For actually changed overrides, both versions still clone, and I did not see a meaningful regression there (`23.70s` vs `23.16s`).

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
JackCC703: Yeah,after I re-checking the current fork layout, comparing against `vm.gas.GasCosts` directly was enough here.

I kept the conservative fallback of cloning if `vm.gas` / `GasCosts` cannot be read, and added coverage for that case too.

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
JackCC703: Thanks. I wrapped the override values in `_ForkOverrides` so the internal plumbing is a bit less noisy without changing the external `get(...)` signature.😺

tests/evm_tools/test_fork_cache.py
JackCC703: Fixed, this now uses `assert_never`.

tests/evm_tools/test_fork_cache.py
JackCC703: Fixed, I replaced the casts here with `assert isinstance(...)` checks.

SamWilsn (approved):
I wonder if we should move the `_ForkOverrides` dataclass into `ethereum_spec_tools.forks` and use it for `Hardfork.clone` directly. The parameter list for that function is getting unwieldy.

JackCC703:
> I wonder if we should move the `_ForkOverrides` dataclass into `ethereum_spec_tools.forks` and use it for `Hardfork.clone` directly. The parameter list for that function is getting unwieldy.

That makes sense. I can open a follow-up PR for this.

PR #2742, opened 2026-04-21
fix(bls12_381): added raise for all InvalidParameter exceptions in decode_g1_scalar_pair / decode_g2_scalar_pair

### LINKED ISSUE #2734: decode_g1_scalar_pair / decode_g2_scalar_pair: missing `raise` makes length check a no-op
## Bug

In `src/ethereum/forks/prague/vm/precompiled_contracts/bls12_381/__init__.py`, both scalar pair decoders construct an `InvalidParameter` exception but never raise it:

```python
def decode_g1_scalar_pair(data: Bytes) -> Tuple[Point3D[FQ], int]:
    if len(data) != 160:
        InvalidParameter("Input should be 160 bytes long")  # should be: raise InvalidParameter(...)
    ...

def decode_g2_scalar_pair(data: Bytes) -> Tuple[Point3D[FQ2], int]:
    if len(data) != 288:
        InvalidParameter("Input should be 288 bytes long")  # should be: raise InvalidParameter(...)
    ...
```

Python silently discards the constructed object. The length guard is a no-op.

## Why it hasn't caused issues

The current callers (`bls12_g1_msm`, `bls12_g2_msm`) always validate `len(data) % LENGTH_PER_PAIR != 0` before looping, and pass exact-length slices to the decoders. So the guard is currently unreachable. But it creates a false contract: anyone calling these functions directly (tests, tooling, forks) would receive no protection.

## Fix

Add `raise` to both guards.

astrion-coder:
## 🗒️ Description

Fix missing `raise` statements in BLS12-381 scalar pair decoders.

Both `decode_g1_scalar_pair` and `decode_g2_scalar_pair` were constructing an `InvalidParameter` exception when input length checks failed, but not raising it. This resulted in the validation being a no-op, allowing execution to continue with invalid inputs.

While current callers (`bls12_g1_msm`, `bls12_g2_msm`) enforce correct input lengths and pass properly sized slices, thus making the bug effectively unreachable in existing flows, this behavior breaks the function’s contract and could lead to silent failures if these utilities are used elsewhere (e.g., tests, tooling, or future refactors).

This PR adds the missing `raise` statements to ensure proper error handling and defensive validation.

## 🔗 Related Issues or PRs

Fixes #2734

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
+++ b/src/ethereum/forks/amsterdam/vm/precompiled_contracts/bls12_381/__init__.py
@@ -414,7 +414,7 @@ def decode_g1_scalar_pair(
 
     """
     if len(data) != 160:
-        InvalidParameter("Input should be 160 bytes long")
+        raise InvalidParameter("Input should be 160 bytes long")
 
     point = bytes_to_g1(data[:128], subgroup_check=True)
 
@@ -614,7 +614,7 @@ def decode_g2_scalar_pair(
 
     """
     if len(data) != 288:
-        InvalidParameter("Input should be 288 bytes long")
+        raise InvalidParameter("Input should be 288 bytes long")
 
     point = bytes_to_g2(data[:256], subgroup_check=True)
     n = int.from_bytes(data[256 : 256 + 32], "big")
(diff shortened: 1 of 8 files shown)


SamWilsn (approved):

SamWilsn:
Because of the checks in `bls12_381_g{1,2}.py`, these lines are (AFAIK) unreachable. In theory, these should be converted to assertions or we should use `FixedByte` subclasses instead, but this is acceptable for now.

PR #2751, opened 2026-04-24
feat(test-tools): t8n file streaming optimizations

fselmo:
## 🗒️ Description

 Switches the t8n streaming execution path from stdout-buffered output to file-based output. 
 
 Replaces Python's stdout-buffered, multi-pass parse of t8n output (which held up to ~5 copies of the alloc in flight: raw stdout bytes, parsed dict, re-serialized string, re-parsed dict, validated Alloc) with a file-based `LazyAllocFile` that streams `{address: account_dict}` entries incrementally via ``ijson`` and validates each ``Account`` one
at a time, preserving every existing Pydantic validator.

This cuts per-test peak RSS on the unchunkified benchmark from 5.38 GB → 3.40 GB (measured, fixtures byte-identical checked via `hasher compare`)... ~2 GB of Python-heap pressure removed per xdist worker.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Screenshot 2026-04-24 at 12 13 19


packages/testing/src/execution_testing/client_clis/transition_tool.py
danceratopz: The check in L787 raises if `temp_dir` is `None`. Perhaps we can clean this up.
```suggestion
        temp_dir: tempfile.TemporaryDirectory,
```

packages/testing/src/execution_testing/client_clis/transition_tool.py
```diff
@@ -760,15 +773,22 @@ def safe_t8n_args(
...
         input_env: LiteralString = "--input.env=stdin"
-        output_result: LiteralString = "--output.result=stdout"
-        output_alloc: LiteralString = "--output.alloc=stdout"
-        output_body: LiteralString = "--output.body=stdout"
+        output_result: LiteralString = "--output.result=output/result.json"
+        output_alloc: LiteralString = "--output.alloc=output/alloc.json"
+        output_body: LiteralString = "--output.body=output/txs.rlp"
         trace_flag: LiteralString = "--trace"
 
+        if temp_dir is None:
```
danceratopz: Perhaps we can raise slightly earlier? with the other `raise` above?

danceratopz (commented):
Nice optimization!

Apart from the `--evm-dump-dir` issue we discussed 🙂 there's a minor nit below.

Claude also claims that the debug replay script (`t8n.sh`) is broken in non-trace stream mode, will try to reproduce with the new version you just fixed!

danceratopz (approved):
I think this is ready to go now! But if you could check the last two commits that'd be great. Commit/Point 3. below seems necessary to help heavy multi-block benchmark tests. Tthe messages have detailed explanations to motivate them.

Commits.
1. First commit fixed up `test-tests-pypy` due to a difference in pypy garbage collection.
2. Second fixed the `t8n.sh` script (this was an easy one).
3. Explicitly clean-up a consumed predecessor's temp dir after the consumer's
  `subprocess.run` returns (to prevent filling up disk in the temp dir in chained/multi-block tests).
4. Adds guard against non-object top-level JSON.

PR #2754, opened 2026-04-24
fix(tooling,ci): collect filled fixtures in json-loader pytest step

gurukamath:
## 🗒️ Description
Move fill output from .just/json-loader/fixtures to tests/json_loader/fixtures so the pytest_collect_file hook in tests/json_loader/conftest.py (which is directory-scoped) actually fires for the generated JSON files. Previously the second pytest step silently skipped all fixtures and only ran the handful of Python test modules under tests/json_loader/.

## ✅ Checklist

    ```console
    uvx tox -e static
    ```

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -160,7 +160,7 @@ json-loader *args:
         --skip-index \
         --clean \
         --ignore=tests/ported_static \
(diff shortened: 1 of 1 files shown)


danceratopz (approved):
Thanks for catching this, @gurukamath!

PR #2755, opened 2026-04-24
🚧 wip: System derived negative tests for BAL

### LINKED ISSUE #2705: suggestion: Increase invalid test coverage using the framework
## suggestion: Increase invalid test coverage using the framework

BAL validation can be broken down into three dimensions:

1. Correctness: each entry has the right value
2. Exactness: exactly the right entries exist (no more, no less)
3. Sequence: entries are in canonical order

BAL equivalence = Correctness + Exactness + Sequence.

Our tests should not assume how clients verify equivalence of computed vs provided BAL (hash, item-by-item, or otherwise).

A client that does zero BAL validation (one that simply passes through the provided BAL)  will pass every happy path test. For every happy path test, we should have a negative test and ensure client rejects them:

* happy path: Alice sends bob 1 ETH. BAL Balance change 1 ETH. accept block. test pass
* invalid path: Alice sends bob 1 ETH. BAL Balance change 2 ETH. reject block. test pass

However, instead of writing invalid tests by hand, we should automate them using the framework. We can derive these invalid tests from a single valid one: 
* Correctnes
  * corrupting an existing entry's value, 
* Exactness
  * adding a bogus entry, 
  * removing an entry,
  * duplicating an entry,  
* Sequence
  * (if there are multiple items) swapping entries. 

The required modifiers for these mutations already exists. So some kind of pytest hook over
existing tests that introspects a valid BAL expectation, enumerates its entries, produces N invalid variants,
one per applicable mutation.

this way our invalid coverage grows organically with valid tests.

_Originally posted by @raxhvl in https://github.com/ethereum/execution-specs/issues/2653#issuecomment-4235678713_

raxhvl:
## 🗒️ Description

Negative tests ensures clients prevents abuse by rejecting *corrupted* BAL. Existing negative tests are hand written, and provide limited coverage.

This PR helps scale negative test coverage organically by systematically corrupting a valid BAL, thus eliminating the need for hand picking "interesting" corruptions.

## How a valid BAL is corrupted

A valid BAL is *corrupted* to ensure clients verify Correctness and Completeness:

| Property     | Invariant                                                          | Corruption                       |
|--------------|--------------------------------------------------------------------|----------------------------------|
| Correctness  | Each entry has the right value.                                    | XOR-flip a value.                |
| Completeness | No expected entry is missing — every access and account is present. | Drop an entry or whole account.  |

That is, we either tamper a value or omit it from BAL. Additionally we also omit whole account.

The total number of negative test $N$ is:

$$N = 2C + A$$

Where:

| Symbol | Meaning                                            |
|--------|----------------------------------------------------|
| $A$    | number of accounts in the BAL                      |
| $C$    | total changes (accesses) across all accounts       |

**Example:** Alice transfers 1 ETH to Bob. The BAL has two accounts and three changes (Alice's nonce and balance; Bob's balance):

$$A = 2, \quad C = 3$$

$$N = 2(3) + 2 = 8$$

```py
 # wrong (3)
<alice>__wrong__nonce__1
<alice>__wrong__balance__1
<bob>__wrong__balance__1
# missing per change (3)
<alice>__missing__nonce__1
<alice>__missing__balance__1
<bob>__missing__balance__1
# missing per account (2)
<alice>__missing
<bob>__missing
```
## 🔗 Related Issues or PRs

closes #2705

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/test_types/block_access_list/corruptions.py
+++ b/packages/testing/src/execution_testing/test_types/block_access_list/corruptions.py
@@ -1,8 +1,8 @@
 """
 Corruption catalog for EIP-7928 Block Access Lists.
 
-Derives the full set of *corruption cases* from a valid
-`BlockAccessListExpectation`. Each case bundles:
+Derives Correctness and Completeness corruption cases from a valid
+``BlockAccessListExpectation``. Each case bundles:
 
   - a stable ``id`` for pytest parametrization,
   - a ``modifier`` that, when applied to the real t8n-produced BAL at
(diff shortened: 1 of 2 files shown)


raxhvl:
Hey @fselmo I see two kinds of invalid tests:

(1) **dynamic invalid tests** that depends on an input BAL,

(2) **static invalid tests** that check static properties and can be tested using stand alone tests, such as checking for duplicate entries or checking ordering by swapping etc.

The pure function I wrote generates type (1) invalid tests. I now need to wire this in the fill logic.

Im thinking the framework expands a valid test to 1 valid + N invalid fixtures. Perhaps a marker. What do you think?

fselmo:
> Im thinking the framework expands a valid test to 1 valid + N invalid fixtures. Perhaps a marker. What do you think?

I love this. We will be brainstorming some ways that we can fuzz test using EELS so I think this might play really nicely with that paradigm as well. One thing I can think of is either a marker or a flag when you fill. Let's say I want to test, locally at least, some tests with these corruptions - I could run with this flag and execute against hive or even the `engine test` @spencer-tb has been working on which would actually consume these very rapidly.

In this sense I think it'd also be valuable for filling and consuming. I just worry a bit that we have ~50k tests for Amsterdam or so (maybe more I forget but something like this) and if we corrupt all of them by N on releases we'd have some insane amount. This is of course a good problem to have and we can decide when and how to control the amount of fuzzing / corruption we do here... but something to think about with the sheer volume that this might produce.

On the other hand, if we fill these particularly for consumption via `engine test` and also somehow strap this up to some future fuzzer for execution on a live network, this is a really nice feature to have.

I'd love to hear your thoughts on the above and on any volume / control knob ideas.

raxhvl:
These test cases are intentional, yet generated. So its somewhere between the usual tests and fuzzing. 

I worry about the bloat this creates. But we need to make space for these for EIPs that cover large protocol surface area. Maybe an adjacent "guardrail" workflow thats meant for important milestones (pre devnet/mainnet launch etc). Fuzzing can be part of it too.

Let me think through to find a home for this bloat.

spencer-tb:
Thanks! Closing this one for now. The hand written invalid BAL cases cover all corruption classes from what I can tell :) (also to wrap up Glammy)

PR #2756, opened 2026-04-25
feat(doc): sort fork directory listings chronologically

SamWilsn:
## 🗒️ Description

Changes the rendered output so that https://ethereum.github.io/execution-specs/diffs/index.html is ordered by fork order instead of alphabetically.

#### Cute Animal Picture

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -275,9 +275,13 @@ export DYLD_FALLBACK_LIBRARY_PATH := if os() == "macos" { "/opt/homebrew/lib" } 
 
 # Generate documentation for EELS using docc
 [group('docs')]
-docs-spec:
+docs-spec $DOCC_SKIP_DIFFS="":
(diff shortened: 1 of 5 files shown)


danceratopz (commented):
Hey @SamWilsn! This is much easier to parse than an alphabetical ordering!

I'd never seen `functools.total_ordering`, so got Claude to help me to understand it. We then came up with this, it's just a suggestion to simplify the ordering logic, but perhaps I missed something!
https://github.com/SamWilsn/eth1.0-specs/pull/6

Btw, I would lean towards reversing the current order, so that the newest fork is prominently found at the top. That's how the [Test Case Reference](https://steel.ethereum.foundation/docs/execution-specs/forks/amsterdam/tests/) is ordered (with the exception of benchmark tests). This is not included in #6 above, but wdyt?

danceratopz (approved):
Thanks @SamWilsn! LGTM!

PR #2757, opened 2026-04-25
feat(tests): Add worst-case SSTORE test

nerolation:
🗒️ Description

Add test_sstore_bloated_multi_contract: SSTORE benchmark spreading writes across many fresh contracts and stride-spaced (2**240) slots within each, while a shared dependency_holder enforces an inter-tx RAW edge. Maximizes MPT breadth at post-state-root computation while keeping txs serialized.

🔗 Related Issues or PRs

N/A.

✅ Checklist

* All: Ran fast static checks to avoid unnecessary CI fails, see also Code Standards and Enabling Pre-commit Checks: just static
* All: PR title adheres to the repo standard - it will be used as the squash commit message and should start type(scope):.
* All: Set appropriate labels for the changes (only maintainers can apply labels).
* Tests: Ran mkdocs serve locally and verified the auto-generated docs for new tests in the Test Case Reference are correctly formatted.

Cute Animal Picture

TBD

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_single_opcode.py
+++ b/tests/benchmark/stateful/bloatnet/test_single_opcode.py
@@ -72,6 +72,22 @@ def _max_sloads_per_tx(tx_gas_limit: int, fork: Fork) -> int:
     return tx_gas_limit // cold_sload_cost
 
 
+def _max_sstores_per_tx(tx_gas_limit: int, fork: Fork) -> int:
+    """
+    Conservative upper bound on cold SSTOREs that fit in a max-gas tx.
+
+    Bounds by the cheapest cold-key transition (clean non-zero to a
+    different non-zero) so the count covers any pre-existing state.
+    """
+    cold_diff_sstore_cost = Op.SSTORE(
+        key_warm=False,
+        original_value=1,
+        current_value=1,
+        new_value=2,
+    ).gas_cost(fork)
+    return tx_gas_limit // cold_diff_sstore_cost
+
+
 def _sender_generator(
     pre: Alloc, distinct_senders: bool
 ) -> Generator[EOA, None, None]:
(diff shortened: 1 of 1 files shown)


tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -659,6 +675,148 @@ def test_sstore_bloated(
...
+    slots within each, to maximize MPT breadth at post-block state-
+    root computation time.
+
+    Mirrors test_sload_bloated_multi_contract: a shared
+    dependency_holder forces an inter-tx RAW edge on a single common
+    (account, slot) pair, while every transaction targets a freshly-
+    deployed contract whose writes stride by ``2**240`` so successive
+    SSTOREs land on disjoint regions of that target's storage trie.
+    """
+    # Successive SSTOREs are spaced 2**240 apart in raw key space;
```
jochem-brouwer: Is this intended to be used to make the MPT keys have more breadth/reach/range? Note that the keys in the MPT are first hashed before they are being put in, so they are by default balanced/randomly distributed

LouisTsai-Csie (approved):
Looks good from my side!

nerolation:
One issue with this test is that the contracts I'm targeting have no storage. I talked with @CPerezz and @jochem-brouwer about this but essentially we need to first deploy the needed code to some accounts (while keeping track of them), then bloating the storage slots (while keeping track of them) and then selecting accounts + storage slots that are far from each other such that the post-state root calculation becomes "worst-case".
Each transaction targets a different contract and within those contracts the tx targets distinct storage slots far from each other. The result would be worst-case root calc times for each individual account AND worst-case state root calculation over the accounts.

LouisTsai-Csie:
@nerolation I’m not sure I fully understand your comment, and I’d like to know the next steps:
- Do you need additional bloating on perf-devnet-3? If so, Jochem and Carlos can help with that.
- Would you prefer that I merge this PR first and then open a new one for future refactor? Or keep this PR open and not merge it?

For information, I recently merged a PR (#2672) that addresses the initcode prefix length issue. If that’s the current bottleneck, it might be worth taking a look at this benchmark.

fselmo:
@LouisTsai-Csie curious if this review is stale or what is the status here. No real rush, just going through PRs that are marked as approved but weren't merged so we can get them across the finish line :)

PR #2760, opened 2026-04-27
refactor(test-execute): include EIP-8037 state-gas in contract deployment

qu0b:
## Summary

In `cli/pytest_commands/plugins/execute/pre_alloc.py`, both `_deterministic_deploy_contract` and `_deploy_contract` size the deploy transaction's `gas_limit` as

```python
deploy_gas_limit  = gas_costs.TX_BASE + gas_costs.TX_CREATE
deploy_gas_limit += len(code) * gas_costs.CODE_DEPOSIT_PER_BYTE
deploy_gas_limit += memory_expansion + calldata
deploy_gas_limit *= 2  # safety
```

The `CODE_DEPOSIT_PER_BYTE` term only covers the legacy 200 gas/byte. Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) (Amsterdam), code deposit additionally pays state gas equal to `code_size * cost_per_state_byte`. With cpsb in the ~700–1200 range at typical devnet gas limits, the surcharge is hundreds of kilo-gas per kilobyte of code — well past what the `* 2` safety buffer can absorb.

## Symptom

Under Amsterdam, deploying a sufficiently large contract via `pre.deploy_contract(...)` succeeds for short bytecode but the deploy tx runs out of gas for longer bytecode. The actual contract code stays empty (`0x`) and EEST raises:

```
Exception: Deployed test contract didn't match expected code at address 0x...
(not enough gas_limit?).
Expected: ...
Actual: 0x
```

Concrete repro on bal-devnet-4 (100M block gas limit → cpsb=1174):

```
tests/amsterdam/eip8024_dupn_swapn_exchange/test_swapn.py
  ::test_swapn_basic[fork_Amsterdam-state_test-swapn_stack_235]
```

The 951-byte contract here needs ~1.12M extra state-gas (`951 * 1174`) on top of the legacy `951 * 200 = 190K` deposit, which the existing formula doesn't budget for. Smaller stack-index variants of the same test (`swapn_stack_{17..200}`, ~80–300 bytes) pass because the `* 2` margin happens to cover their smaller cpsb surcharge.

## Fix

Add `fork.code_deposit_state_gas(code_size=...)` to `deploy_gas_limit` in both call sites. The helper is already defined on `BaseFork` (default returns 0 via the `Genesis` implementation) and overridden by `EIP8037` to return `code_size * cls.cost_per_state_byte()`, so this change is a no-op on every fork outside Amsterdam.

## Test plan
- [x] Repro the failure on bal-devnet-4 against a besu node at 100M block gas: 192/193 pass, only `swapn_stack_235` fails with "Actual: 0x".
- [ ] Verify the fix: re-run `tests/amsterdam/eip8024_dupn_swapn_exchange/test_swapn.py::test_swapn_basic[fork_Amsterdam-state_test-swapn_stack_235]` against the same besu node with this branch installed → all 8 stack-index variants pass.
- [ ] CI on `forks/amsterdam`.

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -352,6 +352,13 @@ class Alloc(SharedAlloc):
             )
         deploy_gas_limit = gas_costs.TX_BASE + gas_costs.TX_CREATE
         deploy_gas_limit += len(deploy_code) * gas_costs.CODE_DEPOSIT_PER_BYTE
+        # EIP-8037: code deposit also pays state gas (`code_size * cpsb`).
+        # Returns 0 on pre-Amsterdam forks. Without this, the doubled-safety
+        # buffer below is too small for larger contracts and the deploy tx
+        # runs out of gas, leaving the contract empty (e.g. swapn_stack_235).
+        deploy_gas_limit += fork.code_deposit_state_gas(
+            code_size=len(deploy_code)
+        )
         deploy_gas_limit += memory_expansion_gas_calculator(
             new_bytes=len(initcode)
         )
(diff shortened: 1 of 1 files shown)


LouisTsai-Csie:
Looks good

LouisTsai-Csie (approved):
LGTM

PR #2622, opened 2026-04-05
feat(validate): new validate command for direct client binary testing

spencer-tb:
## 🗒️ Description

New `validate` (props to @raxhvl for the naming convention) CLI command for running EEST fixtures directly against client EVM binaries. Replaces `consume direct` with a cleaner UX — type is the subcommand, `--client` is required, no `--bin` or `--type` flags.

```
validate health                    # health check all clients
validate engine --client geth      # engine tests
validate state --client besu       # state tests
validate block --client nethermind # block tests
```

### Features
- **7 clients**: geth, besu, nethermind, erigon, reth, ethrex, nimbus
- **Per-type Pydantic models**: `StateTestResult` (stateRoot), `BlockTestResult` (lastBlockHash), `EngineTestResult` (lastPayloadStatus)
- **Exception matching**: maps client error strings to EEST exception types, verifies correct exception for every invalid test (`--no-exception-check` to disable)
- **Cross-validation**: lastBlockHash against fixture, lastPayloadStatus (VALID/INVALID) for engine tests
- **validate.toml** config for client binary paths with per-type overrides (state-bin, block-bin, engine-bin)
- **Auto-tuning**: bin-workers and xdist settings per client
- **Health checks**: version detection + sanity fixture per client per type
- **Aliases**: `go-ethereum` resolves to `geth`
- **Fully standalone**: no dependency on consume plugin
- **Removes `consume direct`**: replaced entirely by `validate`

### Pydantic Result Models

All client adapters return structured results via a shared model hierarchy in `cli_types.py`:

```
FixtureTestResult          # base — name, pass, fork, error
├── StateTestResult        # + stateRoot
├── BlockTestResult        # + lastBlockHash
│   └── EngineTestResult   # + lastPayloadStatus
```

| Model | Extra Fields | Used By |
|-------|-------------|---------|
| `FixtureTestResult` | `name`, `pass`, `fork`, `error` | base class |
| `StateTestResult` | `stateRoot` | `validate state` |
| `BlockTestResult` | `lastBlockHash` | `validate block` |
| `EngineTestResult` | `lastBlockHash`, `lastPayloadStatus` | `validate engine` |

Each client binary outputs JSON matching these schemas. The shared `validate_helpers.py` module validates results against fixture expectations:
- **stateRoot** compared to fixture's `postState` hash
- **lastBlockHash** compared to fixture's `lastblockhash`
- **lastPayloadStatus** checked as VALID (positive) or INVALID (negative test)
[...]

PR #2776, opened 2026-04-28
feat(exceptions): support external client exception mappers

### LINKED ISSUE #2631: Enable definition of client exception mappers via "external" file
Currently, client exception verification depends on exception mappers defined as Python source within execution-specs, for example:
https://github.com/ethereum/execution-specs/blob/4bf8bbe745972edac5398ac32adacc1d007fa827/packages/testing/src/execution_testing/client_clis/clis/ethrex.py#L10-L19

This creates friction for client teams and STEEL as any change in exception message strings requires a PR to execution-specs. This friction could be eliminated if we allow clients to define exception mappers in their own repo (as one of toml/yaml/json) and enable import of these files to populate our exception mappers.

Additionally,m #2622 enables exception mapping for state, blockchain and blockchain engine test fixtures. Imo, now is the time to move the primary test approach from hive (system) to module testing. In which case, clients can run these tests on every push in CI. This means, if we keep the status quo, client teams would need to PR to execution-specs in order to fix exceptions, which is clearly unacceptable. Allowing clients to enable their own exception mappers would greatly simplify this process.

h/t @edg-l who originally made this suggestion!

shrirajpawar4:
## 🗒️ Description

  Implements YAML-backed external exception mappers for client-owned error-string mappings.

  Today, client exception mappings live in EEST Python source, so client teams need an execution-specs PR whenever their error strings change. This adds a small external mapper layer so client repos can
  keep those mappings next to their client code while EEST keeps the built-in mappers as defaults.

  External mappings are additive only:
  - built-in Python mappers remain unchanged
  - external YAML mappings extend the built-in mapper
  - duplicate mapped exceptions are de-duplicated
  - unknown messages still fall back to `UndefinedException`

  YAML v1 format:

  ```yaml
  version: 1
  name: geth-ci
  substring:
    TransactionException.INSUFFICIENT_ACCOUNT_FUNDS:
      - "insufficient funds for gas * price + value"
  regex:
    BlockException.INVALID_GASLIMIT:
      - 'child gas_limit \d+ .*'

  Integration points:

  - validate.toml:

    [geth]
    bin = "../go-ethereum/build/bin/evm"
    exception-mapper = "../go-ethereum/eest-exceptions.yaml"
  - Hive consume:

    uv run consume engine --exception-mapper geth=../go-ethereum/eest-exceptions.yaml
```

  This is stacked on #2622 because that PR introduces the new direct validate workflow where client-owned exception mappings are most useful for CI.

  ## 🔗 Related Issues or PRs

  Fixes #2631.
  Depends on #2622.

  ## ✅ Checklist

  - [ ] All: Ran fast static checks to avoid unnecessary CI fails:

    just static
  - [x] All: PR title adheres to the repo standard and starts type(scope):.
  - [x] All: Considered updating the online docs in the ./docs/ directory.

  Focused checks run:

  ```
  uv run ruff check packages/testing/src/execution_testing/exceptions/external.py packages/testing/src/execution_testing/exceptions/__init__.py packages/testing/src/execution_testing/exceptions/tests/
  test_external.py packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/exceptions.py packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/
  simulators/sync/conftest.py packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/tests/test_exception_mappers.py packages/testing/src/execution_testing/cli/pytest_commands/
  plugins/validate/conftest.py packages/testing/src/execution_testing/cli/pytest_commands/plugins/validate/tests/test_exception_mappers.py

  uv run pytest packages/testing/src/execution_testing/exceptions/tests packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/tests packages/testing/src/execution_testing/cli/
  pytest_commands/plugins/validate/tests/test_exception_mappers.py

  Result: 41 passed, 5 warnings. The warnings are existing collect-only dummy consumer warnings.
  ```


felix314159:
this pr is impossible to review right now. you should wait until the pr it depends on is merged and then rebase.

shrirajpawar4:
yes okay

PR #2779, opened 2026-04-29
feat(doc): replace index with __init__.py

SamWilsn:
## 🗒️ Description

Pulls in docc 0.6, which replaces the old `index.html` with the content of `__init__.py`, making for much smoother navigation.

#### Cute Animal Picture

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -225,7 +225,7 @@ actionlint = [
     "shellcheck-py>=0.10",
 ]
 doc = [
-    "docc>=0.5.1,<0.6.0",
+    "docc>=0.6.0,<0.7.0",
     "fladrif>=0.2.0,<0.3.0",
     "mistletoe>=1.5.0,<2",
 ]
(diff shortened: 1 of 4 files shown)


danceratopz (commented):
This is so much nicer! 

Unfortunately it removes the ordering for the "forks" page that was added in #2756. I assume "diffs" is ordered, but I only ran `just docs-spec-fast", which skipped it.
Image

Also, could you try adding a minimal `src/ethereum/forks/__init__.py` stating what this sub-package contains? I think that would mean that we get a nicely rendered page instead of a "plain file listing" page?

SamWilsn:
> Also, could you try adding a minimal `src/ethereum/forks/__init__.py` stating what this sub-package contains? I think that would mean that we get a nicely rendered page instead of a "plain file listing" page?

Unfortunately adding an `__init__.py` there would break some stuff. We use a [namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/), in theory to allow third party packages to install forks and have them _Just Work_ but no one has ever tried to do that.

SamWilsn:
> Unfortunately it removes the ordering for the "forks" page that was added in #2756. I assume "diffs" is ordered, but I only ran `just docs-spec-fast", which skipped it.

Good catch! I forgot a minus sign.

danceratopz (approved):
Awesome, yup that fixed it! LGTM!

PR #2782, opened 2026-04-30
fix(test-forks): treat transition fork variants as equal to canonical

danceratopz:
## Description

`TransitionBaseClass.with_env_gas_limit` produced fresh class objects that compared unequal via the default `type.__eq__`, breaking pre-alloc group reuse during `fill --generate-pre-alloc-groups` and tripping the `add_test_pre` assertion (`Incompatible fork: CancunToPragueAtTime15k!=CancunToPragueAtTime15k`).

Mirror `BaseForkMeta` by adding `__eq__`/`__hash__` on `TransitionBaseMetaClass` keyed on the canonical identity, set `_base_fork` on the variant inside `with_env_gas_limit`, and widen `_identity`/`_maybe_transitioned` signatures to `type` so the metaclass can reuse them. Adds a regression test covering transition fork variant equality.

## Related Issues or PRs

Bug introduced by #2690 (`feat(test-forks): Gas-limit-aware Fork, fork-aware Environment`).

## Checklist

    ```console
    just static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/forks/base_fork.py
+++ b/packages/testing/src/execution_testing/forks/base_fork.py
@@ -184,7 +184,7 @@ class BaseForkMeta(ABCMeta):
         return cls.name()
 
     @staticmethod
-    def _maybe_transitioned(fork_cls: "BaseForkMeta") -> "BaseForkMeta":
+    def _maybe_transitioned(fork_cls: type) -> type:
         """
         Return the transitioned fork, if a transition fork, otherwise return
         `fork_cls`.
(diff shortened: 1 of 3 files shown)


felix314159 (approved):
lgtm, fill with `--generate-all-formats` passed for bal and the unit test passed too.

but we also have to be a bit careful because a footgun for e.g. transition forks would be sth like:

```
from execution_testing.forks import Amsterdam

# transition from bpo2 to amsterdam to get post with gas limit 120 mil
if post == Amsterdam:
    post = Amsterdam  # replaces variant with canonical Amsterdam (`_env_gas_limit=0`)
```

basically just cuz you have the same canonical fork identity it does not mean u have the same variant state

danceratopz:
> lgtm, fill with `--generate-all-formats` passed for bal and the unit test passed too.

:+1: 

> but we also have to be a bit careful because a footgun for e.g. transition forks would be sth like:
> 
> ```
> from execution_testing.forks import Amsterdam
> 
> # transition from bpo2 to amsterdam to get post with gas limit 120 mil
> if post == Amsterdam:
>     post = Amsterdam  # replaces variant with canonical Amsterdam (`_env_gas_limit=0`)
> ```
> 
> basically just cuz you have the same canonical fork identity it does not mean u have the same variant state

Agreed. If this becomes commonplace, we could look into protecting ourselves against this (something like wrapping `(fork_cls, env_gas_limit)` in a small `@dataclass(frozen=True)`. I think it's ok for now; but yeah anyone using this has to be aware of the potential consequences.

PR #2787, opened 2026-04-30
refactor(test-benchmark): support 8037 for `SSTORE` benchmark

LouisTsai-Csie:
## 🗒️ Description

Under eip-8037, `SSTORE` on a zero to nonzero transition incurs an additional state-gas cost (`32 * cost_per_state_byte`) that was not being accounted for when sizing the per-tx iteration count. This PR updates the `IteratingBytecode` to support the eip-8037 helper and refactor the `SSTORE` benchmark.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/tools/tools_code/generators.py
+++ b/packages/testing/src/execution_testing/tools/tools_code/generators.py
@@ -1002,20 +1002,19 @@ class IteratingBytecode(Bytecode):
         gas_limit: int,
         compute_gas_limit: int | None = None,
         **intrinsic_cost_kwargs: Any,
-    ) -> Tuple[bool, int]:
+    ) -> bool:
         """
         Check whether iteration_count iterations fit within the gas limits.
 
-        Returns (fits, combined_gas) where combined_gas is the
-        total regular+state gas (i.e. tx.gas) given the iteration
-        count. fits is True when both:
-          - combined_gas <= gas_limit (block-budget constraint).
+        Returns True when both:
+          - The combined regular+state gas (i.e. tx.gas) is <=
+            gas_limit (block-budget constraint).
           - The regular gas, computed as
-            combined_gas - iteration_count * iterating_state_gas,
-            respects the compute_gas_limit
+            combined - iteration_count * iterating_state_gas,
+            respects the compute_gas_limit.
         """
         if iteration_count <= 0:
-            return True, 0
+            return True
         combined = self.tx_gas_limit_by_iteration_count(
             fork=fork,
             iteration_count=iteration_count,
@@ -1023,12 +1022,12 @@ class IteratingBytecode(Bytecode):
             **intrinsic_cost_kwargs,
         )
         if combined > gas_limit:
-            return False, combined
+            return False
         if compute_gas_limit is not None:
             compute = combined - iteration_count * self.iterating_state_gas
             if compute > compute_gas_limit:
-                return False, combined
-        return True, combined
+                return False
+        return True
 
     def _binary_search_iterations(
         self,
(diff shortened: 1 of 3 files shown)


tests/benchmark/stateful/bloatnet/test_single_opcode.py
```diff
@@ -189,6 +190,7 @@ def run_bloated_eoa_benchmark(
                 cache_txs.append(
                     Transaction(
                         gas_limit=tx.gas_limit,
+                        data=tx.data,
```
jochem-brouwer, line 193: The data does not seem to be set (always empty) is this correct?

packages/testing/src/execution_testing/tools/tools_code/generators.py
fselmo, line 1171: This doesn't affect the code you're trying to change in this PR but do we need this here too?

```diff
+  compute_gas_limit=gas_limit_cap,
```

Not sure how to handle the `gas_limit` here at first pass. Perhaps we can table this but we should add a TODO if so.

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -985,61 +993,92 @@ def tx_gas_limit_by_iteration_count(
...
             **intrinsic_cost_kwargs,
+        }
+
+        ok, _ = self._iterations_fit_within_gas_limits(
+            iteration_count=1, **fits_kwargs
         )
-        if single_iteration_gas > gas_limit:
+        if not ok:
             raise ValueError(
                 "Single iteration gas cost is greater than gas limit."
```
fselmo: nit, message here could mention both limits now


```suggestion
                "Single iteration gas cost exceeds gas_limit or compute_gas_limit."
```

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -985,61 +993,92 @@ def tx_gas_limit_by_iteration_count(
...
-            start_iteration=start_iteration,
+        fits_kwargs: Dict[str, Any] = {
+            "fork": fork,
+            "start_iteration": start_iteration,
+            "gas_limit": gas_limit,
+            "compute_gas_limit": compute_gas_limit,
             **intrinsic_cost_kwargs,
+        }
+
+        ok, _ = self._iterations_fit_within_gas_limits(
```
fselmo: It looks like the second arg here (`combined_gas`) isn't used anywhere, is this the case? Can we drop that return value from a `tuple` and return just the `bool`? Is this planned to be used somewhere?

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -985,61 +993,92 @@ def tx_gas_limit_by_iteration_count(
...
+        iteration_count: int,
+        start_iteration: int,
+        gas_limit: int,
+        compute_gas_limit: int | None = None,
+        **intrinsic_cost_kwargs: Any,
+    ) -> Tuple[bool, int]:
+        """
+        Check whether iteration_count iterations fit within the gas limits.
+
+        Returns (fits, combined_gas) where combined_gas is the
```
fselmo: Mentioned elsewhere but the `combined_gas` doesn't seem to be used here. Is this maybe dead code?

fselmo (approved):
I think this makes sense @LouisTsai-Csie. If you're seeing what you need in gas-benchmarks I think we can merge this and address anything else in a future PR. Going to approve to unblock you but I left some comments for us to think about.

packages/testing/src/execution_testing/tools/tools_code/generators.py
LouisTsai-Csie: Added, this should be here

packages/testing/src/execution_testing/tools/tools_code/generators.py
LouisTsai-Csie: Yes, we could remove the `combined_gas` value, which is only used for logging.

packages/testing/src/execution_testing/tools/tools_code/generators.py
fselmo: Hmm I'm on my phone right now but is it an issue that the `gas_limit` here remains as the cap as well? Does it need to be able to accommodate for higher limits for 8037?

PR #2793, opened 2026-05-01
fix(test-consume): better xdist detection for enginex

danceratopz:
TLDR: This fixes a client clean-up bug due to failing to detect whether we're running with xdist, n>1. It is only relevant when running `uv run consume enginex` against a `./hive --dev` instance: In dev mode, users can specify workers via various command-line flags, when being ran via `./hive -sim` the method is fixed (`-n <N>`) and detection is already robust.

EngineX relies on each pre-alloc group being executed by a single xdist worker so that the per-worker client manager sees every test in the group and can stop the shared client as soon as the group completes.

The previous parallelism detection only matched '-n' as a separate argument. Common spellings such as '-n=6', '-n6', and '--numprocesses=6' did not trigger the EngineX loadgroup override, so pytest-xdist could use its default distribution and split one pre-alloc group across workers.

Detect all supported xdist parallelism spellings and ensure consume enginex always runs with '--dist=loadgroup', overriding incompatible distribution modes with a warning. Keep the behavior scoped to EngineX so consume engine and the other simulators are unchanged.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

image

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/processors.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/processors.py
@@ -162,6 +162,9 @@ class HiveEnvironmentProcessor(ArgumentProcessor):
         `--dist=loadgroup` is inert when xdist is not active (no `-n`), so
         it is safe to ensure unconditionally.
         """
+        if any("no:xdist" in arg for arg in args):
+            # Without xdist, `--dist` is an unknown argument.
+            return args[:]
         modified_args = args[:]
         found_dist = False
         changed_dist = False
@@ -172,7 +175,13 @@ class HiveEnvironmentProcessor(ArgumentProcessor):
             if arg == "--dist":
                 found_dist = True
                 if index + 1 < len(modified_args):
-                    if modified_args[index + 1] != "loadgroup":
+                    value = modified_args[index + 1]
+                    if value.startswith("-"):
+                        # Malformed `--dist <flag>`: leave it for
+                        # argparse to reject with a clear error.
+                        index += 1
+                        continue
+                    if value != "loadgroup":
                         modified_args[index + 1] = "loadgroup"
                         changed_dist = True
                     index += 2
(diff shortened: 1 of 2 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/processors.py
```diff
@@ -146,7 +143,66 @@ def _has_regex_or_sim_limit(self, args: List[str]) -> bool:
...
+            if arg == "--dist":
+                found_dist = True
+                if index + 1 < len(modified_args):
+                    if modified_args[index + 1] != "loadgroup":
+                        modified_args[index + 1] = "loadgroup"
+                        changed_dist = True
+                    index += 2
+                    continue
+                modified_args.append("loadgroup")
+                changed_dist = True
```
spencer-tb, line 190: If `--dist` is followed by another flag (the malformed --dist -n 6), this overwrites the next token, silently swallowing -n 6 and leaving a stray positional 6.
```suggestion
                if index + 1 < len(modified_args):
                    value = modified_args[index + 1]
                    if value.startswith("-"):
                        # Malformed `--dist <flag>`: leave it for
                        # argparse to reject with a clear error.
                        index += 1
                        continue
                    if value != "loadgroup":
                        modified_args[index + 1] = "loadgroup"
                        changed_dist = True
                    index += 2
                    continue
                modified_args.append("loadgroup")
                changed_dist = True
```

packages/testing/src/execution_testing/cli/pytest_commands/processors.py
```diff
@@ -146,7 +143,66 @@ def _has_regex_or_sim_limit(self, args: List[str]) -> bool:
...
+
+        EngineX client cleanup depends on each worker seeing every test in a
+        group. Any xdist distribution mode other than loadgroup can split a
+        pre-alloc group across workers, causing each worker to start its own
+        group client and defer cleanup until session teardown.
+
+        `--dist=loadgroup` is inert when xdist is not active (no `-n`), so
+        it is safe to ensure unconditionally.
+        """
+        modified_args = args[:]
```
spencer-tb, line 168: Now that `--dist loadgroup` is appended unconditionally, consume enginex -p no:xdist fails at CLI parse with an unrecognized argument, where it previously ran. Unlikely combination, but cheap to guard:
```suggestion
        if any("no:xdist" in arg for arg in args):
            # Without xdist, `--dist` is an unknown argument.
            return args[:]
        modified_args = args[:]
```

spencer-tb (approved):
Small AI found nits :)

PR #2797, opened 2026-05-04
feat(zkevm): add forks description into ChainConfig

jsign:
## 🗒️ Description

This is a PR that tries to address a solution to the problem explored in https://hackmd.io/@jsign/eth-zkevm-chainconfig-statelessinput

It extends the `ChainConfig` so that stateless inputs carry the fork configuration.
TL;DR-ish:
```
@slotted_freezable
@dataclass
class ForkActivation:
    """
    Activation point for a protocol fork.
    """

    block_number: U64 | None
    timestamp: U64 | None


@slotted_freezable
@dataclass
class BlobSchedule:
    """
    Effective blob parameters for a protocol fork.
    """

    target: U64
    max: U64
    base_fee_update_fraction: U64


@slotted_freezable
@dataclass
class ForkConfig:
    """
    Per-fork configuration needed to interpret stateless inputs.
    """

    fork: ProtocolFork
    activation: ForkActivation
    blob_schedule: BlobSchedule | None


@slotted_freezable
@dataclass
class ChainConfig:
    """
    Chain configuration needed for stateless validation.
    """

    chain_id: U64
    active_fork: ForkConfig
```

The guest program can validate if the `ChainConfig` for two situations is clean:
1. If the received `ChainConfig` refers to a fork that isn't part of its internal enum, that means the `ChainConfig` refers to a future fork not known semantically to the guest program. This is a red flag of some configuration issue, and it fails fast.
2. Given the received (and valid) `ChainConfig` and the `NewPayloadRequest`, if the guest program finds block `block_number or timestamp` rules doesn't match the active fork, then it must fail. This probably means the block corresponds to an older fork that isn't properly described in the `ChainConfig` (e.g., if `ChainConfig` doesn't describe the full history of forks, only recent ones for whatever reason).

For test fillings, we assume `ChainConfig` is only for Amsterdam for now. In the future, we can add support for the framework to automatically detect the current fork and build the chain config accordingly, but this introduces a decent amount of complexity we don't need for now, since the stateless guest is only implemented for the Amsterdam fork anyway.

I also intentionally didn't add new tests in `tests/eip8025...` folder, but now we must also plan to add tests for this in all ways this field could be invalid. e.g. `ChainConfig` containing unknown forks for the guest, block not matching any active fork definition and potentially many other cases. This is planned to be done in the near future, just to make this PR shorter and easier for ELs to take a look at as a reference for this API change.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -74,17 +75,100 @@ class NewPayloadRequestHeader:
     execution_requests: ExecutionRequests
 
 
+class ProtocolFork(StrEnum):
```
jsign, line 83: From the guest program's perspective, the main thing that matters is the correspondence between integer indices and strings, not the strings themselves.

src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -74,17 +75,100 @@ class NewPayloadRequestHeader:
...
 class ChainConfig:
     """
     Chain configuration needed for stateless validation.
-
-    TODO: Since we do not want the client to hold all possible chains,
-    we may want to add more to the chain config, like a genesis file.
     """
 
     chain_id: U64
+    forks: Tuple[ForkConfig, ...]
```
jsign: Main TL;DR of this PR is here.

Something else I thought about adding here (or in `ForkConfig`) was the address of all system contracts. This was inspired in [EIP-7910](https://eips.ethereum.org/EIPS/eip-7910), and I was mainly wondering if this might be useful for native rollups.

I'm not totally sure about it yet, so I think for now it might be okay to not include these fields. But we could easily add them if if useful for L1 or native-rollups.

src/ethereum/forks/amsterdam/stateless_ssz.py
```diff
@@ -52,13 +58,21 @@
 MAX_BYTES_PER_WITNESS_NODE = 2**20
 MAX_BYTES_PER_CODE = 2**24
 MAX_BYTES_PER_HEADER = 2**10
+MAX_FORK_CONFIGS = 32
```
jsign: This is the next power of two value from the current forks count. It gives around 7 extra forks after Amsterdam. 

Sounds like a decent amount of headroom, but we could also switch to 64.

src/ethereum/forks/amsterdam/stateless_ssz.py
```diff
@@ -52,13 +58,21 @@
...
+MAX_OPTIONAL_FORK_ACTIVATION_VALUES = 1
+MAX_BLOB_SCHEDULES_PER_FORK = 1
```
jsign, line 59: These are mainly ways to have optionals with SSZ. One way is by using `Union`s but other used style is having lists with max length 1. I preferred the latter since might be easier for libraries in general.

kevaundray: I think not using union is the better choice, see: https://github.com/ethereum/consensus-specs/issues/4763

src/ethereum/forks/amsterdam/stateless_ssz.py
kevaundray: what would be the downside of going to 64?

kevaundray (approved):

src/ethereum/forks/amsterdam/stateless_ssz.py
jsign: Nothing really since the padding shouldn't be relevant for perf. So prob I'll switch to 64 and we can forget about this being a discussion

jsign: Changed to 64.

PR #2799, opened 2026-05-04
Fix __init__.py labels in rendered diff listings

### LINKED ISSUE #2788: Fix `__init__.py` file name in rendered diffs
On, for example, [bpo5 -> amsterdam diff](https://ethereum.github.io/execution-specs/diffs/bpo5/amsterdam/index.html), in the Browse section, `__init__.py` is incorrectly labelled as `index`:

A screenshot of the browse section of the specs with __init__.py labelled as index

It should appear like:

A screenshot of the browse section of the specs with __init__.py correctly labelled as __init__.py

shrirajpawar4:
## Summary

Fix diff Browse labels for package modules so `__init__.py` is displayed correctly instead of `index`.
 ## What changed

  - split diff source metadata into:
    - a listing/display path
    - an output path
  - keep `__init__.py` as the listing path for diff sources
  - keep `index` as the emitted output path so diff URLs remain unchanged
  - add a render-level regression test for the Browse listing
  - run the targeted docc regression test in the docs workflow

  ## Why

  The diff renderer was reusing the rewritten output path (`.../index`) as the visible listing path. That caused Browse sections on diff pages to show `index` instead of `__init__.py` for package
  modules.

  This change fixes the label at the source metadata layer while preserving existing generated diff URLs like `.../index.html`.

  ## Testing

  - `uv run --group test --group doc pytest tests/spec_tools/test_docc.py`
  - `just docs-spec`

  Verified in generated docs that pages such as `diffs/bpo5/amsterdam/vm/index.html` now show `__init__.py` in the Browse section.

  Closes #2788

### DIFF
--- a/.github/workflows/docs-build.yaml
+++ b/.github/workflows/docs-build.yaml
@@ -25,6 +25,7 @@ on:
       - "src/ethereum/**"
       - "src/ethereum_spec_tools/docc.py"
       - "src/ethereum_spec_tools/forks.py"
+      - "tests/spec_tools/**"
       - "static/**"
       - "docs/**"
       - "packages/testing/**"
@@ -234,6 +235,9 @@ jobs:
           submodules: recursive
 
       - uses: ./.github/actions/setup-uv
+
+      - name: Run docc diff regression test
+        run: uv run --group test --group doc pytest tests/spec_tools/test_docc.py
 
       - name: Build spec documentation
         run: just docs-spec
--- a/.github/workflows/docs-build.yaml
+++ b/.github/workflows/docs-build.yaml
@@ -237,7 +237,7 @@ jobs:
       - uses: ./.github/actions/setup-uv
 
       - name: Run docc diff regression test
-        run: uv run --group test --group doc pytest tests/spec_tools/test_docc.py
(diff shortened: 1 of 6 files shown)


src/ethereum_spec_tools/docc.py
```diff
@@ -357,7 +372,7 @@ def relative_path(self) -> Optional[PurePath]:
         """
         Path to the Source (if one exists) relative to the project root.
         """
-        return None
+        return self._relative_path
```
SamWilsn, line 379: This highlights an interesting problem with `docc`... `relative_path` is supposed to point to an on-disk file, but we also use it to compute the display name.

.github/workflows/docs-build.yaml
```diff
@@ -235,6 +236,9 @@ jobs:
 
       - uses: ./.github/actions/setup-uv
 
+      - name: Run docc diff regression test
+        run: uv run --group test --group doc pytest tests/spec_tools/test_docc.py
+
```
SamWilsn, line 241: I take it CI failed because docc wasn't part of the test group? Would it make more sense to just add docc to the test group?

Issue #1988, opened 2026-01-07
EIP-7997 Implementation Tracker: Deterministic Factory Predeploy


marioevz:
### Target Fork

**Amsterdam**

### Instructions

- [x] Assign issue to EIP specification and testing owner(s).

> [!IMPORTANT]
> A specifications specialist and a testing specialist should ideally share ownership of the EIP.

- [x] Add the issue to the target fork milestone if applicable (i.e., the EIP is at least in the [CFI stage](https://eips.ethereum.org/EIPS/eip-7723#considered-for-inclusion)).

#### Guidance for Marking Items Complete

An item should only be checked off once the EIP is considered *stable*. In this context, stable means:

- No major issues or ambiguities are still being uncovered in the specification or tests.
- There are no open discussion points awaiting resolution.
- Client implementations have been consistently passing the tests for at least a week.

It is ultimately up to the owners' discretion to decide when an item should be marked as complete, using this guidance as the basis for that decision.

In exceptional cases, an EIP may require changes after some items have been marked complete or even after the entire issue has been completed and closed. This can happen, for example, when significant design optimizations are identified and agreed upon in ACD, or when critical security issues surface and require updates to the specification or tests.

When this occurs, owners should either unmark the relevant checkboxes if the issue is still open, or create a new tracking issue for the modifications if the original issue had already been closed.

### Specification + Testing Status

- [x] Testing complexity assessed and documented.
- [x] Specification implementation merged to `forks/amsterdam`; the separate EIP branch merge is complete.
- [x] Specification updates merged to the corresponding `forks/amsterdam` branch.
- [x] EIP updates proposed in case of architectural choices surfaced during implementation.
- [x] Required testing framework modifications implemented.
- [x] Test suite implemented in draft PR [#3614](https://github.com/ethereum/execution-specs/pull/3614), pending merge.
- [x] Full code coverage for the changed test and framework logic measured; no new EELS execution lines exist.
- [x] No regressions or failures in the reviewed prior-fork and static test runs.
- [x] [Testing checklist](https://github.com/ethereum/execution-specs/blob/HEAD/docs/writing_tests/checklist_templates/eip_testing_checklist_template.md) complete at 31/31 applicable rows.
- [x] Hardening completed through assertion review and fault-injection checks.
- [x] Benchmarking is not applicable; the factory is ordinary bytecode with no EIP-specific benchmark behavior.
[...]

PR #2802, opened 2026-05-04
feat(spec, spec-specs): EIP-7997

kclowes:
## 🗒️ Description
EIP-7997 implementation and tests

## 🔗 Related Issues or PRs

#1988

## ✅ Checklist

    ```console
    just static
    ```
- ~~[ ] Tests: For PRs implementing a missed test case, update the [post-mortem document](/ethereum/execution-specs/blob/HEAD/docs/writing_tests/post_mortems.md) to add an entry the list.~~
- ~~[ ] Ported Tests: All converted JSON/YML tests from [ethereum/tests](/ethereum/tests) or [tests/static](/ethereum/execution-specs/blob/HEAD/tests/static) have been assigned `@ported_from` marker.~~

#### Cute Animal Picture


petertdavies:
Deploying a system contract in the precompile range has the potential to cause weird bugs in clients that have longstanding hardcoded assumptions that the 0x01-0xff range is only precompiles. This isn't a precompile, it's just a regular contract. The level of testing required is higher than might superficially appear from the EIP.

The included tests look pretty good, but I could think of few more:
* Test the interaction between 0x12 and access list prewarming. The EIP is silent on prewarming, as written this PR does not prewarm.
* Try `CALLCODE`ing it.
* Give it a balance using `SELFDESTRUCT`.

kclowes:
Thanks @petertdavies! I will add those! It is very strange that this is in the precompile range, but is not a precompile.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -179,7 +196,16 @@ def apply_fork(old: BlockChain) -> BlockChain:
     new : `BlockChain`
         Upgraded block chain object for this hard fork.
 
+    [`DETERMINISTIC_FACTORY_ADDRESS`]:
+        ref:ethereum.forks.amsterdam.fork.DETERMINISTIC_FACTORY_ADDRESS
+
     """
+    code_hash = store_code(old.state, DETERMINISTIC_FACTORY_CODE)
```
marioevz: Interesting, I did not know we had this functionality!

I'm not sure though how do we signal from the tests that a block is the first block in the fork?

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,687 @@
...
+    Transaction,
+    compute_create2_address,
+)
+
+from .spec import Spec, ref_spec_7997
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_7997.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7997.version
+
+pytestmark = pytest.mark.valid_from("Amsterdam")
```
marioevz:
```suggestion
pytestmark = pytest.mark.valid_from("EIP7997")
```

marioevz: We should test the transition to the fork where this EIP is enabled, but that circles back to my question on the first comment since I'm not really sure how we signal this in the t8n. Maybe @gurukamath has more insight.

marioevz (commented):
Looks great so far, a few comments. Thanks!

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
kclowes: I _think_ that the best we can do is do a pre_alloc_mutable and do something like: `pre[0x12] - Account(code=b"", nonce=0, balance=0`, and assert that you can call the contract, and it returns success for the call, but it doesn't return anything else. It looks like we do a similar thing [here](tests/cancun/eip4788_beacon_root/test_beacon_root_contract.py::test_no_beacon_root_contract_at_transition). I'm not sure how much value that test really adds though.

src/ethereum/forks/amsterdam/fork.py
kclowes: @marioevz I'm not sure what you mean by: 

> I'm not sure though how do we signal from the tests that a block is the first block in the fork?

Can you clarify? AFAIK, the tests don't call `apply_fork`, and AIUI, we'd have to test with a pre_alloc mutable as discussed below, for inherently not a great test. This is also highlighted in the codecov report since there is not a great way to test this method. Let me know if you see another way though! I'm happy to update.

kclowes:
@marioevz @petertdavies @LouisTsai-Csie - this is ready for another look when you get a chance. I added the tests @petertdavies suggested. 

I'm not sure how to raise the question of whether this contract should be prewarmed - is that something that we'd generally put into the tests and see which clients fail or is it something we'd bring up in a sync call or comment on the EIP?

Edit to add: I don't think there is a way around the `apply_fork` coverage fail, but very happy to hear suggestions!

LouisTsai-Csie:
> Deploying a system contract in the precompile range has the potential to cause weird bugs in clients that have longstanding hardcoded assumptions that the 0x01-0xff range is only precompiles.

I agree with Peter, deploying a contract in the precompile range is confusing. Although i understand design choice, what about using `0xde` ("deploy") instead of `0x12`? If we add more precompiles later, we'd have a cleaner range without a system contract in the middle.

Could we PR to the eip?

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
LouisTsai-Csie: I dont fully understand here, why couldnt we test this state transition like other eip? Using block 14,999 -> 15,000 -> 15,001. We could just do `EXTCODEHASH`, `EXTCODESIZE`, `EXTCODECOPY` to the contract and compare the result across these three blocks?

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,848 @@
...
+FACTORY = Spec.FACTORY_ADDRESS
+
+
+def test_factory_predeploy_account(
+    state_test: StateTestFiller,
+    pre: Alloc,
+) -> None:
+    """The factory bytecode is present at `0x12` with nonce 1."""
+    caller = pre.deploy_contract(
+        Op.SSTORE(0, Op.EXTCODESIZE(FACTORY)) + Op.STOP,
```
LouisTsai-Csie: Can we test:
- `EXTCODESIZE`
- `EXTCODEHASH`
- `EXTCODECOPY`: we could copy all the code and hash it, even though it is the same effect as `EXTCODEHASH`
- `BALANCE` (optional)

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,848 @@
...
+    )
+    state_test(
+        pre=pre,
+        tx=Transaction(
+            sender=pre.fund_eoa(),
+            to=caller,
+            gas_limit=200_000,
```
LouisTsai-Csie: I would suggest we add metadata to the caller.

```
Op.SSTORE(0, Op.EXTCODESIZE(FACTORY), original_value=0, new_value=1) + Op.STOP
```

So that this hardcoded gas limit would not be affected by other repricing eip (e.g., eip-8038).

There are more instance below, what do you think? Maybe this is too much to change.

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,848 @@
...
+            storage.store_next(1, "factory_call_success"),
+            Op.CALL(
+                gas=Op.GAS,
+                address=FACTORY,
+                value=0,
+                args_offset=0,
+                args_size=Op.CALLDATASIZE,
+                ret_offset=0,
+                ret_size=32,
+            ),
```
LouisTsai-Csie, line 129: Combine my review comment for `test_factory_deploys_contract` and `test_factory_forwards_value` here.

I think we should merge this two tests, we could add a new parametrization, `forward_value`, in the `test_factory_deploys_contract`.

And i recommend reducing the forward value from `0xBA1` to 1

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,848 @@
...
+        )
+    else:
+        call_op = Op.DELEGATECALL(
+            gas=Op.GAS,
+            address=FACTORY,
+            args_offset=0,
+            args_size=Op.CALLDATASIZE,
+            ret_offset=0x100,
+            ret_size=32,
+        )
```
LouisTsai-Csie:
```suggestion
    call_op = call_opcode(
        gas=Op.GAS,
        address=FACTORY,
        args_offset=0,
        args_size=Op.CALLDATASIZE,
        ret_offset=0x100,
        ret_size=32,
    )
```
I think this is the same, if we do not explicitly specify the parameter, it would be default to zero.

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,848 @@
...
+) -> None:
+    """
+    `CREATE2` to an address that has only a balance (no code, no storage,
+    nonce 0) succeeds and preserves the existing balance.
+    """
+    salt = 0x66
+    runtime_code = Op.STOP
+    initcode = Initcode(deploy_code=runtime_code)
+    expected_address = compute_create2_address(FACTORY, salt, initcode)
+    pre_balance = 0xBA1
```
LouisTsai-Csie:
```suggestion
    pre_balance = 0x1
```

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,848 @@
...
+
+    pre[expected_address] = Account(balance=pre_balance)
```
LouisTsai-Csie: I think we could fund it via `CALL`, this way we make the test compatible with `execute` mode.

packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_7997.py
```diff
@@ -0,0 +1,55 @@
...
+    "5b60005260206000f3"
+    "5b60006000fd"
+)
+
+
+class EIP7997(BaseFork):
+    """EIP-7997 class."""
+
+    @classmethod
+    def deterministic_factory_predeploy_address(cls) -> Address | None:
```
LouisTsai-Csie, line 32: This could be a problem for benchmarking tests. We've already deployed a large number of contracts using the original factory. After the fork activates in benchmark tests, our address calculations would change, leading to mismatch. But i have no good suggestion for now, let me think about it.

LouisTsai-Csie (changes requested):
Nice work, some additional testing idea:

- Account delegate to the `0x12`
- Constructor revert with large data (16KB+) 
- Oversized initcode revert
- 0xEF-prefix deployed code revert

Since this EIP isn't SFI, should we create `eips/amsterdam/eip-7997` from `forks/amsterdam` and merge into that branch instead of directly into `forks/amsterdam`?

```
main
 └── forks/amsterdam              ← Integration branch for Amsterdam fork
      ├── eips/amsterdam/eip-7997 ← Sub-integration branch for individual EIP
      │    ├── feature-A          ← Your PR
      │    └── feature-B          ← Someone else's PR
      ├── eips/amsterdam/eip-XXXX
      └── ...
```

One more thing: should we update `eth_config` to include this system contract?

jochem-brouwer:
Cross-linking: https://github.com/ethereum/execution-specs/issues/1988

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
kclowes: oop, thank you.

spencer-tb:
Cc-ing EIP update: https://github.com/ethereum/EIPs/pull/11783

packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_7997.py
kclowes: @LouisTsai-Csie - I think this comment is now obsolete?

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
kclowes: I think Mario's incoming PR will help with this. I'll wait for it to land and then can update here accordingly (possibly in another PR).

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
kclowes: Nice update, thanks!

kclowes:
Thanks for the comments @LouisTsai-Csie! I think they're all addressed and this is ready for another look!

> One more thing: should we update eth_config to include this system contract?

Do you mean adding it to execution-apis or somewhere in this repo?

packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_7997.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @classmethod
+    def system_contracts(cls) -> List[Address]:
+        """Add the deterministic factory contract."""
+        return [
+            Address(
+                DETERMINISTIC_FACTORY_PREDEPLOY_ADDRESS,
+                label="DETERMINISTIC_FACTORY_PREDEPLOY_ADDRESS",
+            ),
+        ] + super(EIP7997, cls).system_contracts()
```
LouisTsai-Csie: Is this factory a system contract? The EIP doesn't explicitly specify, so I'm uncertain.

If it is, I understand we need to add a system contract entry to the `eth_config` response ([eip-7910](https://eips.ethereum.org/EIPS/eip-7910)). So for Amsterdam, we'd need to add a new entry for this factory.

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_eip_mainnet.py
```diff
@@ -0,0 +1,52 @@
...
+    pre: Alloc,
+) -> None:
+    """
+    Deploy a contract via the deterministic factory by sending
+    `salt || initcode` from an EOA directly to the factory address.
+    """
+    salt = 0x42
+    runtime = Op.PUSH1(0x01) + Op.PUSH1(0x00) + Op.RETURN
+    initcode = Initcode(deploy_code=runtime)
+    expected_address = compute_create2_address(
```
LouisTsai-Csie, line 37: I've verified there is no contract deployed there: https://etherscan.io/address/0x7eb1007d453d50bab90123faebc3cc5d5a625177

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_eip_mainnet.py
```diff
@@ -0,0 +1,52 @@
...
+    Transaction,
+    compute_create2_address,
+)
+
+from .spec import Spec, ref_spec_7997
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_7997.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7997.version
+
+pytestmark = [pytest.mark.valid_at("Amsterdam"), pytest.mark.mainnet]
```
LouisTsai-Csie: Just for consistency!
```suggestion
pytestmark = [pytest.mark.valid_at("EIP7997"), pytest.mark.mainnet]
```

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,812 @@
...
+    `SHA3`, and `BALANCE`.
+    """
+    caller = pre.deploy_contract(
+        Op.SSTORE(0, Op.EXTCODESIZE(FACTORY))
+        + Op.SSTORE(1, Op.EXTCODEHASH(FACTORY))
+        + Op.EXTCODECOPY(FACTORY, 0, 0, Op.EXTCODESIZE(FACTORY))
+        + Op.SSTORE(2, Op.SHA3(0, Op.EXTCODESIZE(FACTORY)))
+        + Op.SSTORE(3, Op.BALANCE(FACTORY))
+        + Op.STOP,
+    )
```
LouisTsai-Csie, line 75: Some suggestion for simplification

```suggestion
    storage = Storage()
    caller = pre.deploy_contract(
        Op.SSTORE(
            storage.store_next(len(Spec.FACTORY_BYTECODE)),
            Op.EXTCODESIZE(FACTORY),
        )
        + Op.SSTORE(
            storage.store_next(keccak256(Spec.FACTORY_BYTECODE)),
            Op.EXTCODEHASH(FACTORY),
        )
        + Op.EXTCODECOPY(FACTORY, 0, 0, Op.EXTCODESIZE(FACTORY))
        + Op.SSTORE(
            storage.store_next(keccak256(Spec.FACTORY_BYTECODE)),
            Op.SHA3(0, Op.EXTCODESIZE(FACTORY)),
        )
        + Op.SSTORE(
            storage.store_next(0),
            Op.BALANCE(FACTORY),
        )
        + Op.STOP,
    )
```

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,812 @@
...
+                balance=0,
+                code=Spec.FACTORY_BYTECODE,
+            ),
+            caller: Account(
+                storage={
+                    0: len(Spec.FACTORY_BYTECODE),
+                    1: keccak256(Spec.FACTORY_BYTECODE),
+                    2: keccak256(Spec.FACTORY_BYTECODE),
+                    3: 0,
+                },
```
LouisTsai-Csie:
```suggestion
            caller: Account(
                storage={
                    0: len(Spec.FACTORY_BYTECODE),
                    1: keccak256(Spec.FACTORY_BYTECODE),
                    2: keccak256(Spec.FACTORY_BYTECODE),
                    3: 0,
                },](caller: Account(storage=storage),)
```

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,812 @@
...
+            gas_limit=1_000_000,
+        ),
+        post={
+            caller: Account(storage=storage),
+            target: Account(nonce=1, code=bytes(runtime_code)),
+        },
+    )
+
+
+def test_factory_different_salts_produce_different_addresses(
```
LouisTsai-Csie, line 215: Suggest using the Storage object with store_next helper; Otherwise, the test implementation looks good!

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,812 @@
...
+    caller's context, so `CREATE2`'s deployer is the caller — not the
+    factory. The contract is deployed at the address derived from the
+    caller, and the factory-derived address is empty.
+    """
+    salt = 0x44
+    runtime_code = Op.STOP
+    initcode = Initcode(deploy_code=runtime_code)
+    factory_derived = compute_create2_address(FACTORY, salt, initcode)
+
+    call_op = Op.DELEGATECALL(
```
LouisTsai-Csie:
```suggestion
    call_op = call_opcode(
```

This should be updated, or `call_opcode` would never be referenced

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,812 @@
...
+    the cold-vs-warm address access cost.
+
+    - Without access list: difference is 2,500.
+    - With access list including the factory: difference is 0.
+    """
+    # Identical measurement block around each EXTCODESIZE: GAS, op, POP,
+    # GAS, SWAP1, SUB. Same operations on both sides cancels overhead.
+    measure = (
+        Op.GAS + Op.POP(Op.EXTCODESIZE(FACTORY)) + Op.GAS + Op.SWAP1 + Op.SUB
+    )
```
LouisTsai-Csie: We could use this helper object instead: `CodeGasMeasure`

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
```diff
@@ -0,0 +1,812 @@
...
+                code=bytes(runtime_code),
+            ),
+        },
+    )
+
+
+@pytest.mark.parametrize(
+    "use_access_list,expected_delta",
+    [
+        pytest.param(False, 2500, id="without_access_list"),
```
LouisTsai-Csie: Hardcoded the access list value might be a problem in the future, if there is more access list repricing.

The delta could be configured in the test as below:

```python
delta = 0 if use_access_list else fork.COLD_ACCOUNT_ACCESS - fork.WARM_ACCESS
```

LouisTsai-Csie (changes requested):
Thanks for the update! I left some comments and suggestion.

Note that i did not review test related to block access list

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_eip_mainnet.py
kclowes: Oops, thank you!

packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_7997.py
kclowes: Technically, it's not, but this is analogous to the deposit contract, and so I think it should be added to the `eth_config` response. I believe there is some machinery in adding it here where it will also get added to the `eth_config` response, but will double check that.

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
kclowes: 🤦‍♀️

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_factory.py
kclowes: good idea!

packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_7997.py
LouisTsai-Csie: Nice, thank you

LouisTsai-Csie (approved):
Looks good from my side, confirm all the suggestion applied.

One small thing is this comment: https://github.com/ethereum/execution-specs/pull/2802#discussion_r3391205772

We could do in a separate PR that remove the hardcoded gas limit

packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_7997.py
marioevz:
```suggestion
```
Not a sys contract, needs to be removed.

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -3188,7 +3188,15 @@ def deposit_contract_initial_storage() -> Storage:
...
-@pytest.mark.with_all_system_contracts
+@pytest.mark.with_all_system_contracts(
+    # Skip the EIP-7997 deterministic factory: its CREATE2 increments the
+    # delegated account's nonce and deploys a contract, neither of which
+    # this test's post-state expects. Set-code-to-factory is covered in
+    # tests/amsterdam/eip7997_deterministic_factory_predeploy/
+    # test_factory.py::test_factory_via_eip7702_delegation.
+    selector=lambda c: c
+    != Address(0x4E59B44847B379578588920CA78FBF26C0B4956C),
+)
```
marioevz:
```suggestion
@pytest.mark.with_all_system_contracts
```

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/eth_config/tests/test_execute_eth_config.py
marioevz, line 1: Can be removed.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/eth_config/tests/test_execute_eth_config.py
```diff
@@ -291,6 +291,50 @@
...
+    "RIPEMD160": "0x0000000000000000000000000000000000000003",
+    "SHA256": "0x0000000000000000000000000000000000000002"
+    },
+    "systemContracts": {
+    "BEACON_ROOTS_ADDRESS": "0x000f3df6d732807ef1319fb7b8bb8522d0beac02",
+    "CONSOLIDATION_REQUEST_PREDEPLOY_ADDRESS":
+                           "0x0000bbddc7ce488642fb579f8b00f3a590007251",
+    "DEPOSIT_CONTRACT_ADDRESS": "0x00000000219ab540356cbb839cbe05303d7705fa",
+    "DETERMINISTIC_FACTORY_PREDEPLOY_ADDRESS":
+                           "0x4e59b44847b379578588920ca78fbf26c0b4956c",
```
LouisTsai-Csie:
```suggestion
```

This should be remove based on Mario's suggestion

marioevz (approved):
This is great, this will allow us to rely more on this contract for testing in any devnet/testnet. Thanks!

PR #2811, opened 2026-05-06
test(test-benchmark): add `bn128_add` precompile special case benchmarks

GottfriedHerold:
## 🗒️ Description
This adds 2 new benchmark tests regarding the ecAdd precompile. Notably,
we include test cases for the special cases where we
- double a point
- add the negative of a point to itself
These cases are highly likely to require different code paths in most implementations, to we should benchmark them separately.


LouisTsai-Csie (approved):
LGTM, thank you for the new cases

Issue #2812, opened 2026-05-06
EIP7928: missing test case when multiple consecutive SSTOREs change same slot in one single transaction


0xEllie:
The output for BALs should contain only the final value change to the slot (intermediate changes should not appear in the final BALs). 

Note that `test_bal_multiple_storage_writes_same_slot` Only test when the same slot is changed through cross transactions, which also increases the nonce and is a different scenario than the one I claimed is missing here.

PR #2813, opened 2026-05-06
feat(tests): add EIP-7928 BALs missing test when multiple consecutive…

0xEllie:
Add EIP-7928 BALs missing test when multiple consecutive SSTOREs change same slot in one tx

### Description

As mentioned in the related issue, a test case for a single transaction that executes consecutive SSTOREs within the same storage slot is missing. 

form the issue : 
"Note that `test_bal_multiple_storage_writes_same_slot` Only test when the same slot is changed through cross transactions, which also increases the nonce and is a different scenario than the one I claimed is missing here."

### Related Issues or PRs
Fixes issue #2812


fselmo (approved):
Thanks @0xEllie! I added a bit more here. Looks good to me.

PR #2824, opened 2026-05-07
🧪 feat(tests):  EIP-7928 BAL test for 7702 account call with low balance

raxhvl:
## 🗒️ Description

Adds coverage for BAL tests targeting 7702 accounts. Checks if authority and delegate are recorded when called into with insufficient call value.

## ✅ Checklist

    ```console
    just static
    ```


fselmo:
@raxhvl can you run `just fix` and `just static` and fix the lint here? 👀

fselmo (approved):
@raxhvl thanks for this! I made two additions, specifically in two separate commits here so we can track it:

- first commit: I added some parametrization across warm/cold target and delegation which doesn't change anything but good to add I think. We also don't need the `bal=1`. Can do with / without changing that but the non-empty code (delegation designation) is enough for the account to not be empty.

- second commit: I actually looked to add CALLCODE case as well and then thought to parametrize it all. I think it reads clean but lmk what you think... adds extra coverage here.

Also added the test to `test_cases.md`. Approving but lmk what your final review on my pushes are and we can merge or edit before merging 👀

raxhvl (commented):
Beautiful! Thank you.

PR #2825, opened 2026-05-07
feat(zkevm): fill public_keys execution witness field

jsign:
## 🗒️ Description

Adds transaction public keys to Amsterdam stateless validation input and threads them through payload/block execution.

- Populate `StatelessInput.public_keys` with one 65-byte uncompressed SEC1 public key per payload transaction.
- Validate public-key count and SSZ public-key length before stateless execution.
- Allow Amsterdam execution to verify sender derivation from supplied transaction public keys.
- Extend the blockchain filler with public-key verification/mutation hooks for stateless input tests.
- Add coverage for valid public-key construction, invalid public-key rejection, SSZ length checks, and public-key count mismatches.

So, although I added some test cases to check tha valid construction works fine, and basic invalid public key are rejected by the guest program, I believe we need to add some extra tests here. Mainly regarding potential wrong public keys reg ECDSA parity checks, and probably other cases. We will do that in other upcoming PRs.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/specs/blockchain.py
+++ b/packages/testing/src/execution_testing/specs/blockchain.py
@@ -95,6 +95,9 @@ from execution_testing.test_types.execution_witness import (
     ExecutionWitnessHeadersExpectation,
     ExecutionWitnessStateExpectation,
 )
+from execution_testing.test_types.execution_witness.modifiers import (
+    PublicKeyModifier,
+)
 
 from .base import BaseTest, FillResult, OpMode, verify_result
 from .debugging import print_traces
@@ -215,16 +218,17 @@ def with_execution_witness_implicit_codes(
     return expectation.model_copy(update={"codes_present": codes_present})
 
 
-def rerun_amsterdam_stateless_guest_with_witness(
+def rerun_amsterdam_stateless_guest_with_overrides(
     *,
     fork: Fork,
     block_number: int,
     timestamp: int,
     original_stateless_input_bytes: Bytes,
-    execution_witness: ExecutionWitness,
+    execution_witness: ExecutionWitness | None = None,
+    public_keys: Tuple[Bytes, ...] | None = None,
 ) -> tuple[Bytes, Bytes, bool]:
     """
-    Rebuild the stateless input with the emitted witness and rerun the guest.
+    Rebuild the stateless input with test overrides and rerun the guest.
 
     Amsterdam is currently the only fork with stateless guest support in this
     repository, so the rerun path is kept Amsterdam-specific.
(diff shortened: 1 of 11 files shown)


packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -215,16 +218,17 @@ def with_execution_witness_implicit_codes(
...
-def rerun_amsterdam_stateless_guest_with_witness(
+def rerun_amsterdam_stateless_guest_with_overrides(
     *,
     fork: Fork,
     block_number: int,
     timestamp: int,
     original_stateless_input_bytes: Bytes,
-    execution_witness: ExecutionWitness,
+    execution_witness: ExecutionWitness | None = None,
+    public_keys: Tuple[Bytes, ...] | None = None,
```
jsign, line 228: Mainly adding `public_keys` as a new field in the execution witness that can be modified by tests to check if the guest program is correctly validating its correctness. (The `execution_witness` change is a simplification to make both potentially modified `execution_witness` and `public_keys` feel symmetrical at this method API design).

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -283,6 +295,83 @@ def rerun_amsterdam_stateless_guest_with_witness(
...
+    payload_transactions = tuple(
+        Bytes(bytes(transaction))
+        for transaction in (
+            stateless_input.new_payload_request.execution_payload.transactions
+        )
+    )
+    return public_keys, payload_transactions
+
+
+def verify_stateless_input_public_keys(
```
jsign, line 333: Whenever a test is filled, we always check the `public_keys` generated by t8n matches an explicit list of public keys recovered with ecrecover.

This happens automatically for all tests and do not need an explicit signal from the test creator. This might be a bit harshish on the filler to check this always, but it feels way safer since it is an invariant that must always hold.

I filled locally and didn't notice any observable slowdown because of ecrecovers, so looks should be okay. Eventually if we conclude otherwise, we can avoid doing it always and expect the test writer to provide explicit signal to do this check for partiular cases. NBD.

src/ethereum/forks/amsterdam/execution_engine/new_payload.py
```diff
@@ -71,6 +72,7 @@ def execute_new_payload_request(
     new_payload_request: NewPayloadRequest,
     pre_state: PreState,
     chain_context: ChainContext,
+    transaction_public_keys: Optional[Tuple[Bytes, ...]] = None,
```
jsign, line 75: `execute_new_paylaod_request` can receive an optional list of `transaction_public_keys`.

The call of this method from the guest program perspective always provide the values. But remember this method is also called from the "Engine API" context which is stateful, in that case it isn't passed.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -554,7 +569,12 @@ def check_transaction(
     if tx_blob_gas_used > blob_gas_available:
         raise BlobGasLimitExceededError("blob gas limit exceeded")
 
-    sender_address = recover_sender(block_env.chain_id, tx)
+    if sender_public_key is None:
+        sender_address = recover_sender(block_env.chain_id, tx)
+    else:
+        sender_address = recover_sender_from_public_key(
+            block_env.chain_id, tx, sender_public_key
+        )
```
jsign, line 577: Key point: if the public key ways provided, we use a `recover_sender_from_public_key` which will check if the `sender_public_key` provided is correct or not (and theoretically leverage it to make the signature check faster  -- "theoretically" because of a comment I'll do later).

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -293,6 +297,13 @@ def execute_block(
     if len(rlp.encode(block)) > MAX_RLP_BLOCK_SIZE:
         raise InvalidBlock("Block rlp size exceeds MAX_RLP_BLOCK_SIZE")
 
+    if transaction_public_keys is not None and len(
```
jsign, line 300: Here in `execute_block` we do an early check of length matching between `public_keys` and txs count. Later when processing tx, each concrete pubkey will be verified against the tx signature.

src/ethereum/forks/amsterdam/transactions.py
```diff
@@ -676,6 +680,57 @@ def recover_sender(chain_id: U64, tx: Transaction) -> Address:
...
+    Optimized implementations may avoid full public-key recovery, but must
+    still verify that the supplied key validates the signature and is
+    consistent with the transaction's recovery id / y-parity bit. Otherwise,
+    another valid recovery candidate could derive a different sender address.
+
+    Returns the sender address derived from the verified public key.
+    """
+    if public_key != recover_transaction_public_key(chain_id, tx):
+        raise InvalidSignatureError
+    return _sender_address_from_public_key(public_key)
```
jsign, line 718: This is the "theoretically" comment I made before.

This is the method now called when processing transactions to get the address from the tx plus provided public key. Note the method comment.

For the spec, I want things to be simple, so I simply do `ecrecover` on the tx signature and check it against the provided public key in the witness. This is the simplest way to comply with the spec of verifying pubkey correctness.

As mentioned in this method comment, a real EL should leverage the public key to verify the ECDSA signature against it since that's faster than a full ecrecover. But I also add an important reminder point that that's not enough and it should also check the provided public key matches the parity bit from the tx signature. This is relevant since our case here is not really a normal ECDSA-verification that we trust the public key being correct, but should match what the yParity says in the signature to not have the symmetric EC point of `r`-X-coordinate.

TL;DR: in the spec, I'm doing the easiest non-optimized thing. But leaving the comment for EL authors reading this since might not be obvious.

I'm also planning to add a test to trigger this border case since it is far from obvious, and could lead to an incorrect address derivation of the sender.

cc @kevaundray

tests/amsterdam/eip8025_optional_proofs/test_witness_public_keys.py
jsign, line 1: The two simple tests for valid and invalid cases to do minimal coverage that things are working as expected.

kevaundray (approved):

PR #2826, opened 2026-05-07
feat(zkevm): add schema-id prefix to stateless guest program input bytes

jsign:
## 🗒️ Description
This PR adds a `schema-id` two-byte big-int prefix to the guest program input bytes.

Before this PR, the guest input was SSZ-bytes corresponding to `StatelessInput`. SSZ encoded bytes does not describe which is the target struct to deserialize to. As a concrete example, different forks would have different `ExecutionPayload` thus indirectly different `StatelessInput` definitions.

Guest programs are meant to support multiple forks so they are useful between fork boundaries, thus they need some clear way to distinguish which target `StatelessInput` should be used to deserialize the received input bytes. While this can be done by using some assumptions about SSZ field ordering to read `timestamp` against the `ChainConfig`, this might feel a bit hacky.

Note that `schema-id` isn't really a signal of a target fork -- it can even mean that the rest of bytes after `schema-id` byte have other encoding formats, so should be interpreted as `[schema-id-byte][rest input bytes...]` instead of `[schema-id-byte][ssz-encoded-stateless-input-for-fork-XXX...]` (although for a while this might be the right interpretation in practice).

Other notes:
- `schema-id` is a two-byte prefix. This should provide enough capacity for the near future. Thought about LEB128, but feels like overkill.
- Decided that `schema-id == 0` is not a valid value, mainly to avoid being used by error since it is the default value of a byte.
- Amstedram being `0x01` is mainly the most obvious value, although in the near future, if we end up supporting historical forks, we might want to redefine this if another ordering feels more natural.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


src/ethereum/forks/amsterdam/stateless_guest.py
```diff
@@ -25,8 +26,15 @@ def serialize_stateless_output(
 
 
 def deserialize_stateless_input(data: Bytes) -> StatelessInput:
```
jsign, line 29: Crux of the PR.

src/ethereum/forks/amsterdam/stateless_guest.py
```diff
@@ -25,8 +26,15 @@ def serialize_stateless_output(
...
-    """Deserialize a StatelessInput from SSZ bytes."""
-    ssz_obj = SszStatelessInput.decode_bytes(data)
+    """Deserialize a StatelessInput from schema-prefixed SSZ bytes."""
+    if len(data) == 0:
+        raise ValueError("Stateless input is missing schema id")
+    schema_id = data[0]
+    if schema_id != STATELESS_INPUT_SCHEMA_ID:
+        raise ValueError(
+            f"Unsupported stateless input schema id: 0x{schema_id:02x}"
+        )
```
jsign, line 40: In the spec, we don't have a single guest program entry point that supports multiple-forks. This is mainly related on how `execution-specs` is architected having full specs per fork in different folders.

In real EL guest programs which support multiple forks, they will probably have a `switch` here to act accordingly.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
```diff
@@ -1070,10 +1070,11 @@ def test_execution_witness_in_blockchain_fixture(
     assert len(witness["codes"]) > 0, "executionWitness.codes is empty"
     assert len(witness["headers"]) > 0, "executionWitness.headers is empty"
 
-    # statelessInputBytes and statelessOutputBytes are non-empty hex strings
+    # statelessInputBytes is a schema-prefixed guest input hex string.
     assert "statelessInputBytes" in block
     sib = block["statelessInputBytes"]
     assert isinstance(sib, str) and sib.startswith("0x") and len(sib) > 2
+    assert sib.startswith("0x01")
```
marioevz: This is ok but working with strings makes it a bit more difficult to maintain.

If we import `Fixtures` from `execution_testing.fixtures.file` the entire file in `fixture_path` can be parsed automatically and every field will be in the correct type:

```python
from execution_testing.fixtures.file import Fixtures
...
    with open(fixture_path, "r") as f:
        fixture_data = Fixtures.model_validate_json(f.read())
```

Might be out of the scope for this PR though, but it could be a suggestion for a follow up PR.

marioevz (approved):
LGTM, the single comment is a suggestion that could be addressed later.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
jsign: Ah good suggestion, done in https://github.com/ethereum/execution-specs/pull/2826/changes/224df696aa3e4b1db829a1f4ab9dfbcb0fd20430

rubo (approved):
My only concern is the `schema-id` size being a single byte. I'm leaning more toward 2-byte or `varint`.

jsign:
> My only concern is the `schema-id` size being a single byte. I'm leaning more toward 2-byte or `varint`.

Yep, sounds good. I think I'll switch to a fixed-two-byte value, mostly because something like LEB128 might be an overkill of fiddling with bits inside zkVM (although compared to the big picture shouldn't be a big deal either). Two-bytes would give us >65k values which should be enough.

jsign:
@rubo, made the change in https://github.com/ethereum/execution-specs/pull/2826/commits/e1eb780a37536ef839160eaa0f1e37f1bb81561f if you want to take a final look (two-byte big-endian encoding).

rubo (approved):

PR #2750, opened 2026-04-23
refactor(specs): port state refactor to older forks

gurukamath:
## 🗒️ Description
Finish the State refactor (started in #2381, #2207 and #2264) and make one change to state for the older pre-Cancun forks related to EIP-6780. This PR tackles the following changes:

1. Apply state refactor to the older forks
2. track storage clears across pre-EIP-6780 destructions — see note below.
3. wire t8n through ForkLoad accessors — drop the amsterdam hardcoding.
4. adjust json_loader tests for the state refactor — repoint tests at shared modules; skip test_optimized_state until the optimized state is redesigned (tracked by #2256).

## 🔗 Related Issues or PRs
#2381
#2207
#2264
#2256
#2245 

## ✅ Checklist
- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    just static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.

#### Cute Animal Picture

Cute Animals - 4 of 12

PR #2832, opened 2026-05-11
fix(traces): tracing consistency between input and output alloc formatting

felix314159:
## 🗒️ Description
while looking into how our different tracing flags work I noticed that the same values are represented differently in input vs output alloc. for instance, a storage key would be `0x22` in the input (no padding) but `0x000..22` in the output (32 byte padded). another example is how an empty storage in the input would be shown as `{}` whereas empty fields are completely emitted in the output (same for empty balances and empty code). also nonces in the input are shown like e.g. `0x01` (always at least two chars)  but shown as `0x1` in the output (minimal representation)

this PR fixes all these inconsistencies and adds a unit tests that ensures input and output alloc formatting remains consistent. the output of course can still be different, this PR is only about the format in which values are shown

i think this is a good feature because if we ever want to actually look at the diff between input and output alloc we currently would have a lot of noise, after the PR only the actual differences would be left (less bloating of our human context windows)

i don't think this PR breaks anything, because it only changes the input alloc formatting and leaves the output alloc (that might be read be client software) exactly as is

## Command to run the new unit test

`uv run pytest packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_alloc_dump_canonical.py::test_input_alloc_matches_output_alloc_format -v`

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Cute Animal Picture

### DIFF
--- a/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/spec.py
+++ b/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/spec.py
@@ -1,0 +1,24 @@
+"""Defines EIP-7778 specification constants and functions."""
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class ReferenceSpec:
+    """Defines the reference spec version and git path."""
+
+    git_path: str
+    version: str
+
+
+ref_spec_7778 = ReferenceSpec(
+    "EIPS/eip-7778.md", "ce17d00b8341032a946301944124c4a6013032d6"
+)
+
+
+@dataclass(frozen=True)
+class Spec:
+    """
+    Parameters from the EIP-7778 specifications as defined at
+    https://eips.ethereum.org/EIPS/eip-7778.
+    """
--- a/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
+++ b/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
@@ -24,8 +24,10 @@ from execution_testing import (
 from execution_testing.base_types import HashInt
 from execution_testing.vm import Op
 
-REFERENCE_SPEC_GIT_PATH = "EIPS/eip-7778.md"
-REFERENCE_SPEC_VERSION = "ce17d00b8341032a946301944124c4a6013032d6"
+from .spec import ref_spec_7778
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_7778.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7778.version
 
 
 def build_refund_tx(


src/ethereum_spec_tools/evm_tools/t8n/__init__.py
marioevz, line 1: What's the purpose of the changes to this file?

packages/testing/src/execution_testing/client_clis/cli_types.py
```diff
@@ -512,6 +512,39 @@ def validate(self) -> Alloc:
         return Alloc.model_validate(accumulated)
 
 
+def _alloc_to_canonical_dict(alloc: Alloc) -> Dict[str, Dict[str, Any]]:
```
marioevz: One small downside of this function is that it's decoupled from the `Alloc` model, so if we change it (very unlikely) this function would break.

marioevz (commented):
It looks ok to me, just a couple of comments. I do agree that having the input and output in the same folder be in different formats is annoying.

I think ideally we should coalesce types used in both sides of the t8n to get rid of this issue from the root.

cc @gurukamath since he is working on refactoring t8n.

gurukamath:
> It looks ok to me, just a couple of comments. I do agree that having the input and output in the same folder be in different formats is annoying.
> 
> I think ideally we should coalesce types used in both sides of the t8n to get rid of this issue from the root.
> 
> cc @gurukamath since he is working on refactoring t8n.

That's right. Once we manage to replace the current t8n's `Alloc` parsing mechanism and just start using the EEST `Alloc`, we should be able to have consistent input and output formats. In that world, the pydantic models alone will be responsible for json serialization/de-serialization, leading to a fundamental consistency. Incidentally, that is the next agenda on my list. Just waiting for #2750 to be merged first.

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
felix314159: i had to add the second commit to fix CI failure:

36 tests in `test-tests` and `test-tests-pypy` failed. in all cases`test_filler.py` gave the same exception:

```
E   FileNotFoundError: [Errno 2] No such file or directory:
    '/tmp/tmpmepzcz2k/trace-0-0x1997...jsonl'
    at .../evm_trace/eip3155.py:324: json_file = open(output_path, "w")
```

there was a pre-existing bug that was exposed by the first commit of this PR.
so first `T8N.__init__` called `trace.set_evm_trace(tracers)` (in `t8n/init.py:336`). but `set_evm_trace` is global (there's one tracer slot for the whole process, and if the next t8n does not use traces the previous trace slot is still live). so every time a T8N was constructed it used that slot, and once a T8N's `run()` ended the slot was left pointing at that T8N's tracer, holding that T8N's `tempdir` in its `output_basedir` attribute. then the next T8N (constructed by the next test) would create a fresh `tempfile.TemporaryDirectory()` but never re-installed its own tracer if tracing was disabled for that call. EVM execution still fired the leaked previous tracer, which tried to write `trace-*.jsonl` into a tempdir that had since been cleaned up by `TemporaryDirectory.__exit__`. so that's why we got `FileNotFoundError` in CI

and it is a bit weird that `output_traces` never used `mkdir` for the `output_basedir`. so even in the happy path, the trace writer assumed someone else had created the directory.

so the second commit fixes both the t8n-leak-between-instances and also ensures the directory we try to write to exists

felix314159:
Closing this as it turned out guru was also working on it

gurukamath:
#2924 Should make the input and output alloc serialization consistent since they both use the pydantic framework

Issue #2830, opened 2026-05-10
EIP-7928 BAL: missing test coverage for direct DELEGATECALL/CALLCODE to a precompile (bal-devnet-7)


benaadams:
## Context

While reviewing Nethermind PR [NethermindEth/nethermind#11547](https://github.com/NethermindEth/nethermind/pull/11547) we identified an EIP-7928 BAL coverage gap in EEST that is worth considering for `bal-devnet-7` alignment. Filing in the spirit of the [`bal-devnet-7` tracker (#2804)](https://github.com/ethereum/execution-specs/issues/2804).

## The gap

When an execution client wraps its code-info repository with a precompile-caching decorator (e.g. Nethermind's `PrecompileCachedCodeInfoRepository`, used when block prewarming is enabled) the decorator's `IsPrecompile` fast-path returns the cached `CodeInfo` without going through the underlying world state. Without an explicit `AddAccountRead` on that fast-path, code lookups on precompile addresses can be missing from the generated BAL.

Existing `test_pointer_to_precompile` catches this for the **EIP-7702 EOA-delegated-to-precompile via CALL** scenario, but the gap is broader:

- **Direct DELEGATECALL or CALLCODE to a precompile address**: the call sets `target == ExecutingAccount` (NOT `codeSource`), so the indirect `AccountExists(target)` records the caller, not the precompile. The `codeSource` (precompile) is therefore missing from BAL.

For plain CALL / STATICCALL the gap is masked because `target == codeSource` and `AccountExists(target)` records the precompile incidentally. For DELEGATECALL / CALLCODE there is no fallback path.

## Why pyspec doesn't currently catch it

`tests/prague/eip7702_set_code_tx/test_set_code_txs_2.py::test_pointer_to_precompile` only exercises 7702-delegated CALL paths into a precompile. There is no pyspec test (that we can find) that:

- has a contract execute `DELEGATECALL` or `CALLCODE` to a precompile address (e.g. `0x02` Sha256), and
- asserts that `<precompile_addr>` appears in the generated `BlockAccessList`.

A node running with the precompile-caching decorator active would compute a BAL hash that omits the precompile address, diverging from clients that record it. This becomes a consensus risk if/when EIP-7928 is consensus-active.

## Suggested coverage

State / blockchain test (one variant per opcode):

1. **Pre-state**: contract whose runtime code does `DELEGATECALL <precompile_addr> <gas>`; sender EOA with funds.
2. **Tx**: sender invokes the contract.
[...]

PR #2834, opened 2026-05-11
feat(test): Increase BAL coverage with tests and BAL expectations for some existing

fselmo:
## 🗒️ Description

- Increase BAL coverage. Some expectations and tests in first commit here were based on the claims from this repository: https://github.com/catwith1hat/evm-breaker, though investigation led to most of the reported cases already being covered in the tests here and clients seem to be passing these. Either way, BAL expectations were added to some tests that cover similar scenarios as guardrails, and some novel test scenarios came out of this as well and were added here. Test cases that seem genuinely novel from `evm-breaker` audit that were added in this PR (5/18):
  - `#167`: `test_bal_7702_invalid_authority_has_code_authorization` — auth fails because authority has pre-existing non-delegation code; authority loaded then rejected, must appear in BAL with empty change
   set.
  - `#180`: `test_bal_7702_multi_hop_delegation_chain` — A→B→C delegation chain; CALL(A) resolves one hop only, third address must not appear in BAL.
  - `#165`: `test_bal_create_storage_op_then_selfdestruct_same_tx` — same-tx CREATE/CREATE2 + SLOAD/SSTORE on slot B + SELFDESTRUCT; slot B must appear in storage_reads, never storage_changes, and
  destroyed contract has no nonce/code changes.
  - `#165`: `test_bal_create2_selfdestruct_then_recreate_same_block` — Tx1 CREATE2+SELFDESTRUCT then Tx2 CREATE2 at same address; Tx1 entry has no nonce/code changes, Tx2 entry has fresh nonce+code at the
  same address.
  - `#165`: `test_selfdestruct_send_to_sender` — pre-existing contract SELFDESTRUCTs to tx.sender; sender's BAL entry coalesces nonce_changes (as sender) and balance_changes (as beneficiary).
- Since I was already here I did a quick scan for some related tests of interest against what exists and added a few more cases here in a second commit to increase the coverage even more.
- The third commit here closes out missing test cases reported in #2830 from Nethermind precompile-specific paths.

---

Bonus: I changed all the `fork.header_bal_hash_required()` checks in tests that append a BAL expectation to `fork.is_eip_enabled(7928)`, now that this API is available, as I think it is more readable / desirable this way.

---

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Screenshot 2026-05-11 at 11 43 29


tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7702.py
```diff
@@ -515,6 +515,134 @@ def test_bal_7702_invalid_nonce_authorization(
...
+    tx = Transaction(
+        sender=relayer,
+        to=bob,
+        value=10,
+        gas_limit=1_000_000,
+        gas_price=0xA,
+        authorization_list=[
+            AuthorizationTuple(
+                address=oracle,
+                nonce=Spec7702.MAX_NONCE,
```
marioevz: Here the authorization might be rejected because the nonce does not match the current account's nonce. We could have another case where the account's nonce is indeed the maximum nonce.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -3150,3 +3151,253 @@ def test_bal_create_early_failure(
...
+        initcode=initcode_bytes,
+        opcode=create_opcode,
+    )
+
+    tx0 = Transaction(
+        sender=alice,
+        to=target_a,
+        value=fund_amount,
+        gas_limit=100_000,
+    )
```
marioevz: We could also `pre.fund_address` instead of using two transactions, unless the test requires this.

marioevz (commented):
LGTM, just a couple of comments, thanks!

fselmo:
> LGTM, just a couple of comments, thanks!

Thanks! Both great suggestions. Applied here just now 👀

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
raxhvl: unrelated nit: Use `@pytest.mark.with_all_precompiles`

Since we are so close, would be nice to to simplify this concoction

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7702.py
```diff
@@ -515,6 +515,146 @@ def test_bal_7702_invalid_nonce_authorization(
     )
 
 
+@pytest.mark.pre_alloc_mutable()
+def test_bal_7702_invalid_authority_has_code_authorization(
```
raxhvl, line 523: suggestion:  Check for malformed delegations

Also try code that *looks* like a delegation, to check if clients conflate "starts with magic byte" to be a delegation without verifying if it follows with a valid address.

| Code                     | Bytes | Probe                          |
|--------------------------|-------|--------------------------------|
| 0x00                     | 1     | Current `Op.STOP`              |
| 0xef0100 + 19 zero bytes | 22    | Prefix matches, addr too short |
| 0xef0100 + 21 zero bytes | 24    | Prefix matches, addr too long  |


This should also be checked in the base 7702 tests.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7702.py
```diff
@@ -515,6 +515,146 @@ def test_bal_7702_invalid_nonce_authorization(
...
+    blockchain_test: BlockchainTestFiller,
+    account_nonce: int,
+) -> None:
+    """
+    Pre-load rejection isolator: authorization tuple with
+    `nonce = 2**64 - 1` is skipped before the authority is added to
+    `accessed_addresses` per EIP-7702. Per EIP-7928, the authority MUST
+    NOT be in BAL.
+
+    Parametrized over `account_nonce` to isolate the step-2 rejection
```
raxhvl: Optional nit: Simplify doc, param ids, and test case description

The intent of the test case is difficult to understand from the comment.

Mario's case essentially ensures that nonce match (step 6) is correctly done AFTER max nonce check (step 2).   So our cases are `correct_max_nonce` and `incorrect_max_nonce` and authority is never read regardless of correctness of nonce. 

Simplify the comment:

Parameterize the nonce values such that the auth nonce either matches or differs from the account nonce. When the auth nonce is set to `MAX_NONCE`, the authority must not be loaded, regardless of whether the nonce matches. This ensures clients validate the nonce boundary before checking nonce correctness.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7702.py
```diff
@@ -639,6 +779,108 @@ def test_bal_7702_delegated_via_call_opcode(
     )
 
 
+def test_bal_7702_multi_hop_delegation_chain(
```
raxhvl, line 717: suggestion: Consider looping back to authority

| Case  | Destination | Probe                                                                                                                                                                      |
|-------|-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| chain | C           | Second delegation fails. C not recorded.                                                                                                                                   |
| loop  | A           | Authority is not treated as a special case, and `CALL` to A still fails. Note: A will still be recorded as part of the first delegation, we only assert `CALL` to A fails. |

Set the `entry_address` storage to a non-zero value like `0xDEAD` and assert `slot[0]==0` for both cases. and `destination` should be excluded in case of "chain".

---

EDIT: [See this comment](https://github.com/ethereum/execution-specs/pull/2834/changes#r3227598705)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -3150,3 +3151,245 @@ def test_bal_create_early_failure(
...
+    if is_create2:
+        deploy_op = Op.CREATE2(
+            value=0, offset=0, size=Op.CALLDATASIZE, salt=salt
+        )
+    else:
+        deploy_op = Op.CREATE(value=0, offset=0, size=Op.CALLDATASIZE)
+
+    factory_code = (
+        Op.CALLDATACOPY(0, 0, Op.CALLDATASIZE)
+        + Op.SSTORE(0, deploy_op)
```
raxhvl, line 3200: Issue: Unused result

The factory stores the result of deployment but its never used. Verify the result or drop it.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_opcodes.py
```diff
@@ -3150,3 +3151,245 @@ def test_bal_create_early_failure(
...
+    alice = pre.fund_eoa()
+    beneficiary = pre.fund_eoa(amount=0)
+    salt = 0
+
+    runtime = Op.SELFDESTRUCT(beneficiary)
+    runtime_bytes = bytes(runtime)
+    initcode_bytes = bytes(Initcode(deploy_code=runtime))
+
+    # CALLVALUE=0 -> CREATE2 only; CALLVALUE>0 -> CREATE2 + CALL.
+    factory_code = Conditional(
```
raxhvl: (optional) suggestion: Branch based on state and not call value

Use the state to call the deployed contract exactly once.

```
...
create2()
if slot(1) != 1: 
   sstore(1,call(contract))
```
i.e 
tx1 = `create2` + `call`
tx2 = `create2`

Which gives us two nice to haves:
1. A storage check to confirm the call went through, in addition to the beneficiary balance
2. Simplifies the test to use identical txs `Block( txs=[tx] * 2 ...)`

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -130,7 +137,7 @@
...
+| `test_selfdestruct_send_to_sender` (TangerineWhistle) | Ensure BAL tracks SELFDESTRUCT sending all funds back to the tx sender (no burn). Multi-fork: valid_from TangerineWhistle, BAL expectations applied from Amsterdam via `is_eip_enabled(7928)`. File: `tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py`. | Parametrized over `originator_balance: [0, 100]`. Pre-deployed contract `victim` with `code=SELFDESTRUCT(CALLER)` and the given balance. Alice (tx.sender) sends tx directly to `victim`; no intermediate caller. Pre-Cancun: victim destroyed. >=Cancun (EIP-6780, non-same-tx): victim preserved with balance 0. | When EIP-7928 is enabled, BAL **MUST** include Alice with `nonce_changes` coalesced with `balance_changes` (Alice is both sender and beneficiary). Victim has `balance_changes` 100→0 when funded, otherwise empty changes — and **MUST NOT** have `code_changes` or `nonce_changes`. | ✅ Completed |
```
raxhvl:
```suggestion
| `test_selfdestruct_send_to_sender`  | Ensure BAL tracks SELFDESTRUCT sending all funds back to the tx sender (no burn). Multi-fork: valid_from TangerineWhistle, BAL expectations applied from Amsterdam via `is_eip_enabled(7928)`. File: `tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py`. | Parametrized over `originator_balance: [0, 100]`. Pre-deployed contract `victim` with `code=SELFDESTRUCT(CALLER)` and the given balance. Alice (tx.sender) sends tx directly to `victim`; no intermediate caller. Pre-Cancun: victim destroyed. >=Cancun (EIP-6780, non-same-tx): victim preserved with balance 0. | When EIP-7928 is enabled, BAL **MUST** include Alice with `nonce_changes` coalesced with `balance_changes` (Alice is both sender and beneficiary). Victim has `balance_changes` 100→0 when funded, otherwise empty changes — and **MUST NOT** have `code_changes` or `nonce_changes`. | ✅ Completed |
```

chore:  Clean up table

The first row should exactly be test name since its parsed by downstream tools.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7702.py
fselmo: This can be in a separate PR. I have some tests for this that we should port over from the stealth tests repo.

tests/tangerine_whistle/eip150_operation_gas_costs/test_eip150_selfdestruct.py
```diff
@@ -1195,3 +1195,83 @@ def test_initcode_selfdestruct_to_self(
         blocks=[Block(txs=[tx], expected_block_access_list=expected_bal)],
         post=post,
     )
+
+
+@pytest.mark.parametrize(
```
raxhvl, line 1200: Issue: Incorrect test; consider a simplified test under Amsterdam

The sole intent of this test  "SELFDESTRUCT to sender" is NOT tested. We never check that Alice receives the funds, either via BAL or post state.

EIP-6780 account destruction behavior is already covered elsewhere and should not be part of this test.

Fund transfer behavior of SELFDESTRUCT remains unchanged, so Tangerine Whistle is also an odd placement.

This test should instead be under the Amsterdam instead.

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -1163,6 +1164,36 @@ def test_call_into_self_delegating_set_code(
...
+        expected_block_access_list = BlockAccessListExpectation(
+            account_expectations={
+                tx.sender: BalAccountExpectation(
+                    nonce_changes=[
+                        BalNonceChange(block_access_index=1, post_nonce=1)
+                    ],
+                ),
+                entry_address: BalAccountExpectation(
+                    storage_changes=[],
+                    storage_reads=[return_slot],
```
raxhvl, line 1181: Issue: Make the failure explicit

When the call fails it does a no-op 0 -> 0 write. Which is demoted to a BAL read. This is confusing and can be avoided by using a non-zero pre value like `0xDEAD` for the slot so we can clearly assert call failures by `slot[0] == 0`.

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -1223,6 +1257,51 @@ def test_call_into_chain_delegating_set_code(
         sender=pre.fund_eoa(),
     )
 
+    expected_block_access_list = None
```
raxhvl, line 1260: Optional chore: Collect all the chaining test?

So we have 3 cases that essentially test the same thing after the initial delegation resolution, EVM halts when it sees any delegation pointer:

1. A => A (loopback, one-hop) 
2. A=> B => A  (loopback, two-hop)
3. A => B => C ( chain, two-hop) 

Collect all these tests in one place? Its now split between 7702 and BAL. BONUS: Parameterize into one test

fselmo: scope creep here... I'm going to leave this alone in this PR.

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
```diff
@@ -2677,6 +2787,83 @@ def test_nonce_validity(
...
+    if fork.is_eip_enabled(7928):
```
raxhvl, line 2791: Thought: We have overlapping test for max nonce

This test overlaps a with `test_bal_7702_invalid_nonce_field_max_authorization`. We should keep one.

raxhvl (commented):
Great work on increasing BAL coverage :)

tests/prague/eip7702_set_code_tx/test_set_code_txs.py
fselmo: Yes good point. Removed!

fselmo:
Thanks @raxhvl. I've addressed everything here. I think some of these can come in another PR. We should really do a review back through the `Planned` cases in `test_cases.md` and add them here.

Can you do a final review here?

raxhvl (approved):
Everything except [this](https://github.com/ethereum/execution-specs/pull/2834/changes#r3227506505) is resolved.Its a non-blocking nice to have.

Agree on closing tests from `test_cases.md`. I (and claude) have found great value in `test_cases.md` for BAL. Could we consider this for other EIPs.

fselmo:
> Agree on closing tests from `test_cases.md`. I (and claude) have found great value in `test_cases.md` for BAL. Could we consider this for other EIPs.

I found it useful for BALs for sure. If you can suggest this in a testing meeting, we can see how the team feels 👀

fselmo:
Ah I still need an approve from someone with `write` privileges. @marioevz when you get a chance can you re-review so we can merge this.

spencer-tb (approved):
LGTM!

PR #2835, opened 2026-05-11
feat(zkevm): avoid stateless fields generation for fixtures that use RLP-modifiers

jsign:
## 🗒️ Description
Some fixtures use RLP-modifies to "corrupt" the RLP representation of blocks. Conceptually, this is an encoding-mismatch for the stateless executor since its input is SSZ-encoded. While some particular mutations might be translatable to SSZ-mutations (e.g. raw concrete field mutation), others that might target RLP-nature can't apply. Trying to automate this sounds quite tricky, so this PR auomatically skips fixtures with RLP-modifiers.

The current invalid input test cases are "semantic-invalidations" (e.g. mutate the MPT proofs or headers to be invalid by missing data, or similar). But we are also planning to add more "raw corruption" cases which is exactly the same idea of having RLP-mutators but SSZ-ones.

Edit: has also many unrelated lint fixtures since `projects/zkevm` was rebased.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
@@ -1028,6 +1028,76 @@ test_module_execution_witness_missing_expected = textwrap.dedent(
 )
 
 
+test_module_execution_witness_rlp_modifier = textwrap.dedent(
+    """\
+    import pytest
+
+    from execution_testing import (
+        Account,
+        Alloc,
+        Block,
+        BlockchainTestFiller,
+        Header,
+        Op,
+        Transaction,
+    )
+
+    @pytest.mark.valid_at("Amsterdam")
+    def test_execution_witness_rlp_modifier(
+        pre: Alloc,
+        blockchain_test: BlockchainTestFiller,
+    ) -> None:
+        contract = pre.deploy_contract(code=Op.SSTORE(0, 1) + Op.STOP)
+        sender = pre.fund_eoa()
+        tx = Transaction(to=contract, gas_limit=100_000, sender=sender)
+
+        blockchain_test(
+            pre=pre,
+            post={
+                contract: Account(storage={0: 1}),
+                sender: Account(nonce=1),
+            },
+            blocks=[
+                Block(
+                    txs=[tx],
+                    rlp_modifier=Header(extra_data=b"mutated"),
+                )
+            ],
+        )
+    """
+)
+
+
+test_module_execution_witness_rlp_modifier_stateless_intent = textwrap.dedent(
+    """\
+    import pytest
+
+    from execution_testing import (
+        Alloc,
+        Block,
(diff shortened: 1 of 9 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
```diff
@@ -1028,6 +1028,76 @@ def test_execution_witness_missing_expected(
...
+    def test_execution_witness_rlp_modifier_stateless_intent(
+        pre: Alloc,
+        blockchain_test: BlockchainTestFiller,
+    ) -> None:
+        blockchain_test(
+            pre=pre,
+            post={},
+            blocks=[
+                Block(
+                    rlp_modifier=Header(extra_data=b"mutated"),
```
jsign, line 1092: Add a test taht if a fixture has an `rlp_modifier`, then the stateless fields in the output aren't here.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -987,6 +987,40 @@ def generate_block_data(
...
+        ):
+            raise AssertionError(
+                "skip_stateless_validation cannot be combined with "
+                "execution witness expectations, stateless input public-key "
+                "modifiers, or "
+                "expected_stateless_validation_success"
+            )
+        skip_stateless_for_block = (
+            self.skip_stateless_validation or omit_stateless_artifacts
+        )
```
jsign, line 1022: Most of the diff is moving the block below upwards, plus adding this `skip_stateless_for_block` boolean that extends the feature of signaling t8n to avoid generating stateless output.

Before, this was only run for the `@pymarker` that we added recently to allow test authors to explicitly ask for this. Now, we leverage the same feature to automatically activate it if an RLP-modifier is used.

Another option is not doing it this way and allow t8n to generate it anyways, but after it is returned override the fields with `None`, since those generated fields aren't technically correct if the testing framework modified the RLP representation. I find this option quite messy and somewhat wasteful since we already know before calling t8n that we won't be using the outputs (which are technically incorrect).

finterne (commented):
Review summary: no blocking issues found.

I reviewed the changes around skipping stateless artifact generation for blocks using RLP modifiers, the rejection path for conflicting stateless expectations, and the related import/lint/dead-code cleanups. The new behavior looks consistent with the intent: avoid emitting misleading pre-mutation stateless artifacts, while still failing fast when a fixture explicitly asks for stateless validation together with an RLP modifier.

Focused local checks passed:
- test_filler.py -k 'rlp_modifier': 2 passed
- test_incremental_mpt.py + test_witness_state.py: 65 passed
- git diff --check: passed

One non-blocking note: there is a pre-existing readability/precedence concern around the `failing_tx_count` assignment expression in `blockchain.py`; it does not appear introduced by this PR, so I would treat it separately if worth cleaning up.

PR #2842, opened 2026-05-12
feat(spec-spec, tests): Implement eip-8246 and testing scenario

LouisTsai-Csie:
## 🗒️ Description

Implement EIP-8246: https://forkcast.org/eips/8246

## 🔗 Related Issues or PRs

issue #2821

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


tests/amsterdam/eip8246_selfdestruct_no_burn/test_selfdestruct_no_burn.py
```diff
@@ -0,0 +1,103 @@
...
+    Transaction,
+    compute_create_address,
+)
+
+from .spec import ref_spec_8246
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_8246.git_path
+REFERENCE_SPEC_VERSION = ref_spec_8246.version
+
+pytestmark = pytest.mark.valid_from("Amsterdam")
```
marioevz:
```suggestion
pytestmark = pytest.mark.valid_from("EIP8246")
```

tests/amsterdam/eip8246_selfdestruct_no_burn/test_selfdestruct_no_burn.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@pytest.mark.parametrize(
+    "child_balance",
+    [0, 100_000],
```
marioevz:
```suggestion
    [0, 1],
```
Keep the value low (non-zero) for `execute`.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -448,6 +448,20 @@ def destroy_account(tx_state: TransactionState, address: Address) -> None:
...
+    Clear nonce and code for an account, but preserve balance.
+
+    Parameters
+    ----------
+    account :
+        The account to modify.
+
+    """
+    account.nonce = Uint(0)
+    account.code_hash = EMPTY_CODE_HASH
```
marioevz: Might be worth to double check whether this is sufficient to clear the code. cc @gurukamath as he probably knows the answer.

marioevz (commented):
Looks good overall. I think we need to take care of failing tests in `./tests/cancun/eip6780_selfdestruct` before eventually merging into `forks/amsterdam` (if this gets included in a devnet).

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -448,6 +448,20 @@ def destroy_account(tx_state: TransactionState, address: Address) -> None:
     set_account(tx_state, address, None)
 
 
+def preserve_account_balance(account: Account) -> None:
```
gurukamath: How about creating a `convert_to_balance_only_account` (or something similar) which has all the semantics of `destroy_account` minus the balance bit and simply calling that in `fork.py`?

src/ethereum/forks/amsterdam/state_tracker.py
gurukamath: Yes. This is equivalent to resetting the code.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -434,7 +434,7 @@ def destroy_account(tx_state: TransactionState, address: Address) -> None:
     This function is made available exclusively for the ``SELFDESTRUCT``
```
gurukamath: Docstring to be updated

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -1087,7 +1090,8 @@ def process_transaction(
     block_output.block_logs += tx_output.logs
 
     for address in tx_output.accounts_to_delete:
-        destroy_account(tx_state, address)
+        destroy_storage(tx_state, address)
+        modify_state(tx_state, address, preserve_account_balance)
```
gurukamath: If modify_state ends up calling destroy_account (zero-balance case), destroy_storage runs twice. Not a huge issue but something to be kept in mind

marioevz:
Checking the list of failing tests for the current fill in Amsterdam, they all look like expected failures because of the behavior change.

I recommend we track these in the EIP tracking issue #2821, tackle them in separate PRs to merge to [eips/amsterdam/eip-8246](https://github.com/ethereum/execution-specs/tree/eips/amsterdam/eip-8246), and merge this PR as is after addressing the spec comments.

tests/cancun/eip6780_selfdestruct/test_selfdestruct.py
```diff
@@ -354,7 +356,16 @@ def test_create_selfdestruct_same_tx(
     for address, balance in sendall_final_balances.items():
         post[address] = Account(balance=balance, storage={0: 1})
 
-    post[selfdestruct_contract_address] = Account.NONEXISTENT  # type: ignore
+    if fork >= Amsterdam and selfdestruct_contract_current_balance > 0:
```
marioevz:
```suggestion
    if fork.is_eip_enabled(8246) and selfdestruct_contract_current_balance > 0:
```
So we can know exactly which EIP changed this behavior. Same for the rest of the tests.

LouisTsai-Csie: Updated. I like it, very clear.

src/ethereum/forks/amsterdam/state_tracker.py
LouisTsai-Csie: I like this idea, i am bad at naming.

src/ethereum/forks/amsterdam/fork.py
LouisTsai-Csie: I will keep it as is, but let me know if you have any suggestion!

LouisTsai-Csie:
@marioevz @gurukamath I've updated according to the comment. Now the PR contains spec & test for eip-8246 and refactor the failing legacy tests. Please take a look!

I will stop adding new tests here, and wait for the PR to merge first.

src/ethereum/forks/amsterdam/fork.py
gurukamath: Why not just call `destroy_storage` in `convert_to_balance_only_account` like `destroy_account` currently does? That way, the diff in `fork.py` looks cleaner as well. Otherwise, this looks good to me.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -431,10 +431,9 @@ def destroy_account(tx_state: TransactionState, address: Address) -> None:
     """
     Completely remove the account at ``address`` and all of its storage.
 
-    This function is made available exclusively for the ``SELFDESTRUCT``
-    opcode. It is expected that ``SELFDESTRUCT`` will be disabled in a
-    future hardfork and this function will be removed. Only supports same
-    transaction destruction.
+    Invoked by ``modify_state`` (and the coinbase fee-credit path) to
+    clean up an account that has become empty (zero nonce, empty
+    code, and zero balance) so it does not appear in the post-state.
```
SamWilsn, line 438: There's something funky about this. I've opened #2908.

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -448,6 +447,20 @@ def destroy_account(tx_state: TransactionState, address: Address) -> None:
     set_account(tx_state, address, None)
 
 
+def convert_to_balance_only_account(account: Account) -> None:
```
SamWilsn: This function name needs work. The word "convert" implies that something is changing about the kind of account, not about its contents. For example, you "convert a gas-powered car to electric", but you don't "convert a dirty car to clean" when you go through a car wash.

`convert_to_balance_only_account` implies that the account is changing "kinds" from a regular account to one that can only have a balance. As far as I am aware, there's nothing special about a balance only account. It's equivalent to an account that has received ether but has not sent any transactions.

Some ideas:

- `reset_account` (pairs nicely with `destroy_account`, but also might be confusing)
- `make_account_deployable` (this is more semantically meaningful I think, but a mouthful)
- `clear_account_nonce_and_code_hash` (says what it does, but not what it intends; might make future refactors more difficult)

SamWilsn: The pattern for other functions in this file is to create a function that operates on a frozen account, that defines an nested function to apply the transformation. Is there a reason this function breaks from that pattern?

If not, I'd suggest something like:

```python
def convert_to_balance_only_account(tx_state: TransactionState, ​​address: Address) -> None:
    def clear_account(account: Account) -> None:
        account.nonce = Uint(0)
        account.code_hash = EMPTY_CODE_HASH

    modify_state(tx_state, address, clear_account)
```

SamWilsn (dismissed):
A quick review of only the spec related changes.

src/ethereum/forks/amsterdam/fork.py
```diff
@@ -1087,7 +1090,8 @@ def process_transaction(
     block_output.block_logs += tx_output.logs
 
     for address in tx_output.accounts_to_delete:
-        destroy_account(tx_state, address)
+        destroy_storage(tx_state, address)
+        modify_state(tx_state, address, convert_to_balance_only_account)
```
Carsons-Eels: Might be worth a comment here noting that the explicit destroy_storage exists for EIP-7928 block-access-list completeness...

Carsons-Eels (commented):
I don't see anything wrong with the spec code that hasn't been commented on, will review the tests tomorrow. 🚀

src/ethereum/forks/amsterdam/fork.py
LouisTsai-Csie: Great idea, i will add a new commit.

src/ethereum/forks/amsterdam/state_tracker.py
LouisTsai-Csie: I applied Guru's [feedback](https://github.com/ethereum/execution-specs/pull/2842#discussion_r3274914480) by moving `destroy_storage` into the `convert_to_balance_only_account` function, and your suggestion [here](https://github.com/ethereum/execution-specs/pull/2842#discussion_r3299250886). Now the refactored function becomes:

```py
def convert_to_balance_only_account(
    tx_state: TransactionState, address: Address
) -> None:
    def clear_account(account: Account) -> None:
        account.nonce = Uint(0)
        account.code_hash = EMPTY_CODE_HASH
    destroy_storage(tx_state, address)
    modify_state(tx_state, address, clear_account)
```

Since it now clears storage, nonce, and code hash, the name `clear_account_nonce_and_code_hash` no longer fits, while `reset_accounts` is a bit confusing?

I would support make_account_deployable compared to other two suggestions, i also come up with:
- `nullify_account_code_storage_nonce`
- `strip_to_balance_only`

What do you think?

src/ethereum/forks/amsterdam/fork.py
LouisTsai-Csie: I dont fully understand, could you expand more on this?

I applied Guru and Sam's feedback on the refactor, would this break anything?

Carsons-Eels: This is just a comment request. No functional change requested.

marioevz (approved):
I pushed two commits:
- Remove all references and helper functions for burn logs
- Fix remaining tests by removing burn logs from them

Should be good to merge!

PR #2851, opened 2026-05-13
test(tests): inter-dependent tx tests for BAL parallelization

fselmo:
## 🗒️ Description

From ETH R&D @matkt:

https://discord.com/channels/595666850260713488/1364000387195076608/1504026078107275274

Add five new cross-tx parallelization-correctness tests:
- `test_bal_cross_tx_storage_chain` — 8-tx fibonacci chain on one contract... each tx i ≥ 2 writes `slot[i] = SLOAD(i-1) + SLOAD(i-2)` -- transactions are dependent on previous 2. Catches clients that fail to apply prior txs BAL storage changes.
- `test_bal_cross_tx_deploy_then_call` (`@with_all_create_opcodes`) — tx1 `CREATE`/`CREATE2`s a contract; tx2 `CALL`s it. Catches clients that hit an empty account because tx1's `code_changes` weren't applied.
- `test_bal_cross_tx_balance_dependency` — tx1 sends value to a contract; tx2 calls the contract and `SSTORE`s `SELFBALANCE`. Catches clients that read the pre-block balance.
- `test_bal_7702_cross_tx_delegation_then_call` — tx1 installs an EIP-7702 delegation on `alice`; tx2 and tx3 call `alice` to increment slot `0`, using different senders. Catches clients that don't apply the delegation or skip the intermediate increment.
- `test_bal_cross_tx_funding_chain` — Five-tx chain `alice`→`bob`→`charlie`→`dan`→`eunice`→`target` where each intermediate starts empty and only becomes solvent once the prior tx's BAL `balance_changes` are applied; `oog_minus_1` additionally exercises target's `storage_reads` when eunice's `SSTORE` OOGs one gas short.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Screenshot 2026-05-13 at 11 14 41


tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
```diff
@@ -2188,6 +2190,444 @@ def test_bal_cross_tx_storage_revert_to_zero(
...
+    blockchain_test: BlockchainTestFiller,
+) -> None:
+    """
+    Verify clients apply Tx1's balance change before executing Tx2 in
+    the same block. Tx1 sends value to a contract; Tx2 invokes the
+    contract which records its `SELFBALANCE` to storage. A client that
+    parallelizes Tx2 without applying Tx1's `balance_changes` would
+    record the pre-block balance, yielding a different state root.
+    """
+    transferred = 10**18
```
marioevz:
```suggestion
    transferred = 1
```
We could keep this low just to be mindful of `execute` users in testnets/devnets.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
```diff
@@ -2188,6 +2190,444 @@ def test_bal_cross_tx_storage_revert_to_zero(
...
+            if_false=Op.SSTORE(0, Op.SELFBALANCE),
+        ),
+    )
+
+    tx_send = Transaction(
+        sender=alice,
+        to=contract,
+        value=transferred,
+        gas_limit=100_000,
+    )
```
marioevz: Alternative test: Same idea but transfer the funds via `SENDALL` (Selfdestruct) so that the recipient contract's code is not executed when the funds are received, could perhaps catch some exotic optimization.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
```diff
@@ -2188,6 +2190,444 @@ def test_bal_cross_tx_storage_revert_to_zero(
...
+        post={
+            contract: Account(balance=transferred, storage={0: transferred}),
+        },
+    )
+
+
+@pytest.mark.parametrize(
+    "eunice_oog",
+    [False, True],
+    ids=["success", "oog_minus_1"],
```
marioevz: It would also be interesting to expect an insufficient-balance error instead of an OOG (invalid block): eunice provides a given gas-limit, but the balance passed by the previous transaction is off-by-one insufficient to pay gas_limit * gas_price.

Malicious case: (Not sure if possible) The BAL is corrupted to indicate there's enough balance for eunice to pay gas_limit*gas_price, but in reality Dan passed insufficient value (by one wei).

marioevz (commented):
Looks good, I suggested some extra variations of the tests included. Let me know if the description of these tests are clear. Thanks!

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
fselmo: Added. On the malicious case idea, I wonder... malicious BAL corruption is already covered by `test_bal_invalid_balance_value` (and some other `test_bal_invalid_*` tests). The 5-tx chain variant would surface as the same hash mismatch via the same path, no?

Lmk if I am missing something 👀. I think these valid cases help verify that we can't just run transactions in parallel in any order. What's tricky is all the different valid ways in which a client might optimize for checking "is this tx parallelizable or not".

marioevz: I just verified locally and the malicious case I suggested is not possible since there's no `balanceReads` field in the BAL, so it was a misunderstanding of mine. The idea I had was to have a correct balance-change output for Dan's tx, but then corrupt a (non-existent) balance-read field in the BAL for Eunice's tx with a different value that is sufficient to execute it.

marioevz (approved):
Excellent tests, thanks!

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
fselmo: Posting from Discord just to keep the history here in case this is helpful:

> I see. I think even if there was a `balanceRead`, we would end up in the scenario where this would produce a `pytest.mark.exception_test` with `INVALID_BLOCK_ACCESS_LIST` and this is where the client would fail. Which would be a decent enough invalid test but it wouldn't test the same surface area I was going for in this PR where we provide successful test cases that might trick a client into thinking the transactions are parallelizable in any order but they depend on each other in tricky ways so we can try to catch optimization bugs.

PR #2859, opened 2026-05-14
feat(tests): eip7928 checklist v1

fselmo:
## 🗒️ Description

These are the last remaining tests that I believe get us to a really good spot for the testing checklist for EIP-7928 (BALs).

As mentioned in the testing calls, I think there are some very real missing gaps that the current version of the checklist doesn't cover for complex EIPs that now have different behavior at different opcode gas boundaries, etc. As we discussed, this will likely be on some V2 of the checklist and I'm already working on a version of this with EIP-7928 and some of EIP-8037 as good candidates for what surface areas to expand upon.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Screenshot 2026-05-14 at 15 54 38


tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py
```diff
@@ -0,0 +1,215 @@
...
+    tx = Transaction(sender=sender, to=receiver, value=100, gas_price=10)
+
+    blockchain_test(
+        pre=pre,
+        post={},
+        blocks=[
+            Block(
+                timestamp=FORK_TIMESTAMP - 1,
+                txs=[tx],
+                rlp_modifier=Header(block_access_list_hash=Hash(0)),
```
marioevz, line 113: I think this is for some reason not kicking in (`blockchain_test_engine` fixture): 
```json
{
                        "parentHash": "0x1cfcb2a6818881c1572fc14c78d7bb7e4d0f2956a97481a32c99d7ea66bd4371",
                        "feeRecipient": "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba",
                        "stateRoot": "0x55a8a074769d31265dc733b6b7b4939f8163ce3f4c5fcd11dbb894d95bcdb80e",
                        "receiptsRoot": "0x056b23fbba480696b65fe5a59b8f2148a1299103c4f57df839233af2cf4ca2d2",
                        "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
                        "blockNumber": "0x1",
                        "gasLimit": "0x7270e00",
                        "gasUsed": "0x5208",
                        "timestamp": "0x3a97",
                        "extraData": "0x",
                        "prevRandao": "0x0000000000000000000000000000000000000000000000000000000000000000",
                        "baseFeePerGas": "0x7",
                        "blobGasUsed": "0x0",
                        "excessBlobGas": "0x0",
                        "blockHash": "0xb44ffeeef8f1d2ad377c0d802eec7f2041a765121b939c54564a9b745ca630e4",
                        "transactions": [
                            "0xf85f800a82520894c0f6dc9e5836f54caadbf59cc69346c508e1992b648026a058a81333901bde13feafa1503458f1eda427c32ea68db07f17a4a5257950db8ba078bf41c87ed70d87a6af3d3f6c65624a026a9ae4c9ca3129091363f1c7850515"
                        ],
                        "withdrawals": []
                    },
```
In a normal Amsterdam block I see here `blockAccessList` (and `slotNumber` but this is a separate issue).

The problem might be that `rlp_modifier` is not influencing the `engine_newPayload`, so perhaps we need to add a new field in `Block`.

tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py
```diff
@@ -0,0 +1,215 @@
...
+
+    blockchain_test(
+        pre=pre,
+        post={},
+        blocks=[
+            Block(
+                timestamp=FORK_TIMESTAMP,
+                txs=[tx],
+                rlp_modifier=Header(
+                    block_access_list_hash=Header.REMOVE_FIELD,
```
marioevz, line 145: Similar issue to previous test but the inverse, `blockAccessList` is appearing in the `blockchain_test_engine` fixture's payload.

marioevz (commented):
Looks great overall. I've flagged two issues with the fork transition tests. It might be an issue with our framework.

tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py
fselmo: Ah really nice catch. The issue that I see here is that if we add a new engine field modifier and we forget to define it, then we only modify the `blockchain_test` version of this and not the `blockchain_engine` version (current buggy state here) -- which are two entirely different tests now.

We have to properly enforce that if one of them is set, the other one has to be too. Or we should really only have to set one, tbh. Thoughts on this @marioevz?

fselmo: What I committed here is the automatic passing on of this from `blockchain_test(rlp_modifier=...)`, which guarantees the engine test behavior mimics the intention... but it feels like cheating on the explicitness of what the action is doing. It is no longer just an `rlp_modifier`.

This feels like the best approach for now, but perhaps we have to rename this field in order to make this more explicit - which I would suggest we file as an issue to tackle later. Or perhaps this is OK? Lmk what you think.

tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py
fselmo: See previous comment... addressed for now but in a not-so-ideal way I think? Thanks for flagging.

marioevz (approved):
LGTM, thanks for making the changes 👍

PR #2860, opened 2026-05-14
feat(test-specs): Test mutators

marioevz:
## 🗒️ Description
Introduce a spec-test mutator framework that lets fill apply systematic mutations to existing spec tests so we can hunt for edge cases without writing new tests. Mutators are fork-gated: each fork declares which mutators it supports via `BaseFork.spec_test_mutators()`, and the user opts in per fill invocation.

### Usage

```console
fill ... --test-mutators=EIP_7702_ALL_CONTRACTS_AS_DELEGATIONS
fill ... --test-mutators=all
```

The value selected for a given test is the intersection of what the user requested with what the test's fork supports — so requesting a mutator on a fork that doesn't enable it is a no-op rather than an error, and tests under a fork range get mutated only on the forks where the mutation is meaningful.

### First mutator: EIP_7702_ALL_CONTRACTS_AS_DELEGATIONS (Prague+)

Every `pre.deploy_contract(...)` call is rewritten so that:

1. The original bytecode is placed in a separate account.
2. The address returned to the test is a delegated EOA `(0xef0100 || <code-address>)` pointing at that account.

This stresses the EIP-7702 delegation path against the full body of existing state/blockchain tests. Edge cases surfaced this way are expected to be promoted into standalone tests as a follow-up — the flag is an exploration aid, not part of regular fill runs.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -27,7 +27,7 @@ from execution_testing.base_types.conversions import (
     BytesConvertible,
     NumberConvertible,
 )
-from execution_testing.forks import Fork, TransitionFork
+from execution_testing.forks import Fork, SpecTestMutator, TransitionFork
 from execution_testing.logging import get_logger
 from execution_testing.rpc import EthRPC
 from execution_testing.rpc.rpc_types import TransactionByHashResponse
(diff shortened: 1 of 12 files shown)


felix314159:
afaik you mainly intend this feature to be used with `fill`, but when you run `execute remote` with `--test-mutators=EIP_7702_ALL_CONTRACTS_AS_DELEGATIONS` it will accept the argument but then silently not mutate? it is a bit dangerous/misleading and we should try to have fill and execute be consistent with each other whenever possible

felix314159:
the new flag `--test-mutators` is completely missing from `fill --help`.

and when you use `uv run fill --pytest-help` it does show up, but it is useless because it does not tell u which `TEST_MUTATORS` actually exist, e.g. currently u get

```
Arguments to specify spec test mutators:                                 
  --test-mutators=TEST_MUTATORS                                                                                                                    
                        Parametrize spec tests with the given comma-separated list of mutators.
```

PR #2724, opened 2026-04-20
feat(test-consume): add engine-witness simulator for witness verification

spencer-tb:
## 🗒️ Description

**Fully vibe coded experimental.**

Adds a new `consume engine-witness` Hive simulator that drives Amsterdam blockchain-engine fixtures through a witness-emitting Engine API endpoint and verifies the **client-generated** `ExecutionWitness` against the
fixture's expected witness.

Supports two transports on a single simulator:

| Mode | Endpoint | Witness encoding | Clients today |
| --- | --- | --- | --- |
| **Default** (no flag) | JSON-RPC `engine_newPayloadWithWitnessVX` | RLP | [go-ethereum](https://github.com/ethereum/go-ethereum/pull/30069) |
| `--ssz` | REST `POST /new-payload-with-witness` | SSZ | Ethrex [`feat/zkengine-http`](https://github.com/developeruche/ethrex/tree/feat/zkengine-http) prototype only |

Both paths converge on a shared `NewPayloadWithWitnessResponse` dataclass and the same `assert_witness_matches` helper. Clients that do not implement the chosen transport skip cleanly:
- REST: HTTP `404` / `405` raises `EngineWitnessEndpointNotImplementedError`
- JSON-RPC: `-32601 Method not found` is caught and turned into `pytest.skip`

The simulator also skips whole fixtures with no `executionWitness` on any payload (i.e. pre-Amsterdam forks).

### Usage

**Still to test!**

```bash
# Fill witness-bearing fixtures
uv run fill tests/amsterdam/ --fork=Amsterdam --output=fixtures_witness

# Hive in dev mode (separate shell)
./hive --dev --client go-ethereum
export HIVE_SIMULATOR=http://127.0.0.1:3000

# Default: JSON-RPC + RLP (Geth)
uv run consume engine-witness --input=fixtures_witness

# Alternative: REST + SSZ (Ethrex zkengine-http branch)
uv run consume engine-witness --input=fixtures_witness --ssz
```

## 🔗 Related Issues or PRs

- Consumes fixture infrastructure from #2667 (`jsign-support-blockchain-test-engine`), already merged on `projects/zkevm`.
- REST+SSZ endpoint spec: [execution-apis#773](https://github.com/ethereum/execution-apis/pull/773).
- JSON-RPC endpoint it targets by default: [go-ethereum#30069](https://github.com/ethereum/go-ethereum/pull/30069).

## ✅ Checklist

[...]

PR #2872, opened 2026-05-18
feat(zkevm): add engine-witness simulator and add executionWitnessMutated flag into fixtures

jsign:
## 🗒️ Description
This PR:
- Add support for `uv run consume engine-witness` simulator to allow running execution witness fixtures in Hive. This is based on https://github.com/ethereum/execution-specs/pull/2724 with a couple of extra fixes needed to finish it.
- Add `executionWitnessMutated` boolean flag to fixtures (more about this below).

Regarding the last point: the existing `executionWitness` in fixtures signals the expected execution witness that a stateful-EL should create for the given fixture. This is similar to an "expected BAL" for the block.

But there is a catch: the execution witness isn't part of the execution payload for stateful-ELs; the execution witness is only used as input for the stateless validator.

We have many tests that mutate the correct execution witness for a block, so that we can check that the stateless validator fails when executing the block.

This means that:
- For fixtures that do not mutate the execution witness: the fixture should pass both for stateful and stateless execution clients.
- For fixtures that do mutate the execution witness to make them invalid: the fixture should fail for the stateless validator (since that was the goal of the test), but actually don't make sense to be run in stateful-ELs. Mostly because the mutation happens in the testing framework after the real filling was done, and as said, the execution witness isn't used at all as input in the stateful run so it won't fail.

This means that fixtures that perform execution-witness mutation shouldn't be run with `uv run consume engine-witness`, which targets only stateful-ELs, not stateless validator runs.

The way to handle this now is that I added an `executionWitnessMutated` field in the fixtures, so this can be noted. If it is `true`, then `uv run consume engine-witness` will skip them since those fixtures aren't really a target to test stateful-EL correctness in constructing an execution witness.

I've the feeling this situation is just adding a bit more to the tension of using `blockchain_tests` as a format. We have been pushing back a bit on a new fixture format to avoid complexity in the testing framework and the need for multiple ELs to support it. But at some point after some things settle, we can reconsider if maybe it is a good strategy. e.g. keep `blockchain_tests` for things that only make sense for stateful-ELs, and create a separate format for stateless runs, which would indirectly avoid needing this flag plus avoiding some current bloating of redundant data (in different encoding formats) in `blockchain_tests`. Created https://github.com/ethereum/execution-specs/issues/2873 for tracking discussion.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


Dyslex7c:
`Could not find a part of the path '/var/folders/.../nethermind-eest/PyTests/tests-zkevm@v0.4.1/fixtures_zkevm/fixtures/blockchain_tests_engine/for_amsterdam/eip7928_block_level_access_lists` will these be included in the next release? @jsign

jsign:
@Dyslex7c, the `blockchain_tests_engine` are in the release:
image

is this what are you refering to?

jsign:
@Dyslex7c, I think tehre is a missing folder in your path.

Dyslex7c:
> @Dyslex7c, I think tehre is a missing folder in your path.

right that fixed it, thanks

PR #2883, opened 2026-05-19
feat(tests): EIP-7928 BAL withdrawal predeploy balance read across txs

edg-l:
## Summary

Adds a cross-tx test in `tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_cross_index.py` covering a transaction reading the post-state balance of `WITHDRAWAL_REQUEST_PREDEPLOY` after a prior transaction in the same block paid the withdrawal fee.

## Coverage gap

The existing EIP-7002 / EIP-7251 BAL tests cover the per-tx storage and balance entries for the predeploys in isolation, and the post-execution NOOP-filtering paths, but they do not exercise the case where:

- a user transaction mutates a predeploy's account info (here: balance, via the withdrawal fee),
- and a subsequent transaction in the same block reads that field through an account-info opcode (`BALANCE`).

This is the cross-tx propagation of a BAL `balance_change` to a later transaction's view of the account, for an account that is also touched during pre-execution system calls.

## Test

`test_bal_withdrawal_predeploy_balance_observed_cross_tx`:

1. `tx 0`: EOA sends 1 wei to `WITHDRAWAL_REQUEST_PREDEPLOY` with valid withdrawal-request calldata. The predeploy retains the fee, so its balance transitions `0 -> 1`, recorded as a `balance_change` at `block_access_index = 1`.
2. `tx 1`: calls a contract whose code is `SSTORE(0, BALANCE(WITHDRAWAL_REQUEST_PREDEPLOY)) + STOP`.
3. Post-state: `reader.storage[0] == 1`.

The expected BAL declares the predeploy `balance_change` at index 1 and the reader's `storage_change` at index 2.

Fee retention is consistent with existing BAL tests in the same suite: `test_bal_7002_clean_sweep` in `test_block_access_lists_eip7002.py` already asserts `Account(balance=withdrawal_request.fee)` in post-state after a single fee-paying transaction.

## Notes

- Happy to trim the `BlockAccessListExpectation` if a slimmer declaration is preferred, or to extend to `EXTCODESIZE` / `EXTCODEHASH` / `EXTCODECOPY` as parametrised variants.
- `CONSOLIDATION_REQUEST_PREDEPLOY` (EIP-7251) admits the same shape; can mirror if useful.


fselmo (approved):
Thank you! This lgtm. Will wait for CI check and merge 👍🏼

Issue #2736, opened 2026-04-20
Tracker: Faster, More Targeted Test Fixture Releases


danceratopz:
This issue encompasses 2 goals for test fixture releases:
1. Higher `mainnet` release cadence, i.e., more regular releases of test cases for stable forks.
2. Faster release builds, in particular, for development releases. i.e., we can quickly generate releases that target specific development work (e.g., `bal-devnet-3` or `benchmark`).

Goal 2. could also help unlock more finely grained releases, e.g., for every individual EIP, e.g., `eips/amsterdam/eip-2780`, but we should verify with client teams whether this would be beneficial - our releases should match client teams branches as close as possible.

> [!IMPORTANT]
>Implication: If we only fill for the development fork in dev releases, clients need to consume two artifacts from two separate releases for full coverage and no regressions:
> 1. `fixtures_mainnet.tar.gz`.
> 2. `fixtures_<dev_feature_name>.tar.gz`.

#### Subtasks

- [ ] #2720
- [ ] #2721
- [ ] Implement use of pytest-split in [release_fixture_feature.yaml](https://github.com/ethereum/execution-specs/blob/5502bd80829eabc5e6c585bdfbc7f55c69901f26/.github/workflows/release_fixture_feature.yaml).
- [ ] Enable multi-release support to `consume cache --input` (i.e., allow the input to be a comma-separated list of releases) to allow full coverage runs in the `consume` Hive simulators.
- [ ] Socialize new release model with client teams: New `mainnet` cadence and the two-artifact consumption model.
- [ ] Follow up with client teams on their development branch structures: Input for whether finer-grained releases (e.g., per-EIP) would be useful.
- [ ] Update docs for removal of `fixtures_stable`/`fixtures_develop` -> `fixtures_mainnet`.
#### Higher Mainnet Release Cadence

#2592 mainly solved 1., but requires team agreement/alignment and some follow-up work (only docs?). Essentially, as of now, we can readily create `fixtures_mainnet` (~1hr30). More info below, but TLDR: `fixtures_mainnet` replaces `fixtures_stable` and `fixtures_develop` has been removed.
#### Faster Release Builds

This issue proposes that we achieve faster development release builds by:
1. Only filling the development fork in development releases — e.g., for devnet releases like `bal-devnet-4`, `--until=Amsterdam` (previously used for client-team convenience) would change to `--fork=Amsterdam`.
[...]

PR #2888, opened 2026-05-20
feat(ci): overhaul fixture releases

spencer-tb:
## 🗒️ Description

This PR outlines a detailed description and updates our test fixture release process **for non-benchmark** tests.

### Key Changes

- Release configs are cleaned up:
  - `evm-impl.yaml` is merged into `evm.yaml` and each entry now carries its `impl`, `repo`, `ref`, `evm-bin` and `xdist`.
  - A `benchmark` client alias is kept so `feature.yaml` workflows can reference a client by name, and `build-evm-base` resolves which builder to run via the `impl` field.
  - The ethereumjs t8n tool is removed from the release path.
- We now have `tests`, `devnet` and `benchmark` features (in `forks/amsterdam`).
  - `devnet` is specifically chosen so we don't ever need to update `feature.yaml` for future devnet features. The `devnet` keyword is used as a substring match on release tagging, and `<feat>-devnet` keeps its friendly name.
- `release_fixture_feature.yaml` is replaced with `release_fixtures.yaml`:
  - Triggered now via `workflow_dispatch` with explicit `feature` + `version` inputs. The git tag is `tests-<feature>@<version>` (the `tests` feature tags as `tests@<version>`) and the release title is `<feature>@<version>`.
  - Input validation fails fast: `version` must match `vX.Y.Z`, `feature` must be non-empty, `*-devnet` requires a `branch`, and an `evm` override must be a key in `evm.yaml`.
  - Optional `evm` / `evm_repo` / `evm_ref` inputs override the client impl and the t8n tool repo/ref for a one off release.
  - Releases are tagged with the gh cli on success only, no tag push, no delete/recreate process moving forward.
  - Releases are only drafted in EELS (EEST mirror is removed, EEST is now archived).
  - LLLC and Solc dependencies are removed.
- Multi-runner split: BPO forks are filled within the `osaka` range (no standalone `bpo` split).

### Tagging A Release

To tag releases we must now use the Github CLI. This has the benefit of only creating tags in EELS if the fixture building process is successful. No more tag deletion and recreation, only workflow triggers.

The following can be ran locally or optionally triggered with the github actions website UI.
 ```bash
 gh workflow run release_fixtures.yaml -f feature=tests -f version=v20.3.1
 # devnet releases additionally require the branch to release from:
 gh workflow run release_fixtures.yaml -f feature=bal-devnet -f version=v7.0.0 -f branch=bal-devnet-7
 # optional: override the client / t8n repo+ref for a one-off release
 gh workflow run release_fixtures.yaml -f feature=tests -f version=v20.3.1 \
   -f evm=geth -f evm_repo=ethereum/go-ethereum -f evm_ref=master
 ```

### Downloading A Release

Fixtures can be downloaded by 2 seperate methods. `gh release download` for the raw tarball, or `consume cache` if you want tag resolution (`@latest`), local caching, and `--input` integration with consume subcommands.

```bash
# via gh cli
gh release download tests@v20.3.1 --repo ethereum/execution-specs
gh release download tests-bal-devnet@v7.0.0 --repo ethereum/execution-specs

# latest tests release by publish time:
LATEST=$(gh release list --repo ethereum/execution-specs --limit 100 \
  --json tagName --jq '[.[] | select(.tagName | startswith("tests@v")) | .tagName][0]')
gh release download "$LATEST" --repo ethereum/execution-specs

# via consume cache
uv run consume cache --input=tests@v20.3.1
uv run consume cache --input=tests@latest
uv run consume cache --input=bal-devnet@v7.0.0
```

#### `tests@vX.Y.Z`
In the past (pre-Weld) we had the following release features: `stable` & `develop`, where `stable` was a subset for `develop`. To converge on these 2 features we now define the `tests` feature. This is the invariant to the `benchmark` feature.

The `tests` feature acts as our mainnet set of tests. That is for now `fill --until BPO4`, all tests for all forks until last mainnet fork. This will be released weekly on any change or addition to the tests always from the latest development branch: `forks/amsterdam` currently. Clients will use this release on their main/master branches in CI, and eventually this release will contain all tests from `ethereum/tests` & `ethereum/legacy-tests` (allowing us to archive both of these repos). TLDR; one tag type to verify you will not break mainnet. Additionally this will be ran in our Hive CI (under what is currently labelled generic).

Here we define a new semantic versioning type for our `tests` releases:
- X is the fork number, BPO2 is at index 20,
  - This makes it clear what fork the release is up to date with.
- Y is a consensus-breaking spec change targeting fork X. Should rarely occur.
- Z is a non-breaking change (refactor), new or modified tests.

The first release of this type in EELS will be `tests@v20.0.0`, to catch up from the last fork. For the fork under development (Amsterdam) the first `tests` release will be tagged once all CFI'd EIPs are deemed successful in a devnet (purposely ambiguous here, things change in ethereum). Typically this will resolve to the last devnet before the first testnet; this first release can be viewed as the testnet release. For Amsterdam we will tag `tests@v24.0.0` likely after `glamsterdam-devnet-6` is deemed successful. Spec changes can still occur and that is why we have Y in the new fixture versioning scheme.

#### `tests-<feat>-devnet@vX.Y.Z`
This release type will follow on from the current devnet release process but more explicitly. Today `<feat>` is `bal` and soon it will be `glamsterdam`. `<feat>` can essentially be any keyword but typically is the fork name or the headliner feature.

The `devnet` feature is entirely for test releases during the fork development process for the upcoming fork. Here we still fill for all forks so clients can make sure they do not introduce any regressions, currently `fill --until Amsterdam`, all tests for all forks until the development fork. Clients will use this release on their `devnet` branches in CI; they must pass all of these tests before being included in the devnet that the release is tagged for. This release will be ran in Hive CI under the same naming scheme as we do currently.

For the `bal` devnets today we use the tag `tests-bal@vX.Y.Z`. This PR will change the process to `tests-bal-devnet@vX.Y.Z`. As `bal-devnet-7` is the last of the `bal`'s we will start the new tagging scheme for `glamsterdam-devnet-5`. The devnet releases will additionally follow a new versioning scheme:
- X is the devnet number, so for `glamsterdam-devnet-5` this would be 5,
  - This makes it clear what devnet the release targets.
- Y/Z follow the same semantics as the `tests` releases.

Following the latter, the first `glamsterdam-devnet-5` release will be tagged as `tests-glamsterdam-devnet@v5.0.0`. All devnet releases must be tagged from a devnet branch in EELS, in this case `glamsterdam-devnet-5`. Here we are specifically choosing to diverge from the EELS `/` branch naming scheme to align every repo under the same devnet name.

## 🔗 Related Issues or PRs

- https://github.com/ethereum/execution-specs/issues/2736

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


.github/configs/feature.yaml
```diff
@@ -1,19 +1,12 @@
 # Unless filling for special features, all features should fill for previous forks (starting from Frontier) too
-mainnet:
+consensus:
   evm-type: eels
-  fill-params: --until=BPO2 --generate-all-formats
+  fill-params: --until=BPO4
```
danceratopz: Can we keep `--generate-all-formats` please? :)

spencer-tb:
This PR needs to be merged to fully test out the new workflow, but I did some smaller smoke tests on my fork:
- **consensus@v20.0.0** - https://github.com/spencer-tb/execution-specs/actions/runs/26400607639
- **bal-devnet@v7.0.0** - https://github.com/spencer-tb/execution-specs/actions/runs/26400609657

To keep the workflow runnable in minutes on a fork (no `gigachungus` runners), I scoped both features to a small  Cancun test module and pointed `build` at `ubuntu-latest`. The split/combine/release wiring is identical to production, the only difference is the fill scope and runner diffs.

Smoke test releases can be found here: https://github.com/spencer-tb/execution-specs/releases

chfast:
1. "consensus" name not very good. I skip the section on the first read because I though this is for CL. I don't want very picky, but some of my suggestions: just "tests", "tests-main", "tests-stable" (in a sense for stable spec).
2. I don't really care about the version number so you can chose any versioning. However, you are adding implicit matching rule: bal-devnet-7 → tests-bal-devnet@v7.x.y. This breaks semver (which I don't care about, as mentioned). Make sure you really want this.
3. Having weekly / bi-weekly cadence would be big improvement for me. I often contribute new test cases for "tests-consensus" and want to integrate them to my CI as soon as possible. Do releases often. Skip if nothing new to release.

.github/configs/feature.yaml
```diff
@@ -1,19 +1,12 @@
...
 benchmark:
   evm-type: benchmark
-  fill-params: --fork=Osaka --generate-all-formats --gas-benchmark-values 1,5,10,30,60,100,150 ./tests/benchmark/compute --maxprocesses=30 --dist=worksteal
-  feature_only: true
-
-benchmark_fast:
-  evm-type: benchmark
-  fill-params: --fork=Osaka --generate-all-formats --gas-benchmark-values 100 ./tests/benchmark/compute
-  feature_only: true
+  fill-params: --fork=Amsterdam --gas-benchmark-values 1,10,30,60,100,150 ./tests/benchmark
```
LouisTsai-Csie: We're using geth to fill the tests, but as we discussed, I don't think it has t8n tool support yet. 

I did a partial review focusing mostly on benchmarking, and I don't see any other issues from my side.

marioevz:
> 1. "consensus" name not very good. I skip the section on the first read because I though this is for CL. I don't want very picky, but some of my suggestions: just "tests", "tests-main", "tests-stable" (in a sense for stable spec).
> 
>     2. I don't really care about the version number so you can chose any versioning. However, you are adding implicit matching rule: bal-devnet-7 → [tests-bal-devnet@v7.x.y](mailto:tests-bal-devnet@v7.x.y). This breaks semver (which I don't care about, as mentioned). Make sure you really want this.
> 
>     3. Having weekly / bi-weekly cadence would be big improvement for me. I often contribute new test cases for "tests-consensus" and want to integrate them to my CI as soon as possible. Do releases often. Skip if nothing new to release.

1. I had not thought about, I think it's important. Maybe we should rollback to "stable" or similar?
2. I agree that breaks semver but also think that it should not be a dealbreaker. I see these releases as more "ephemeral" in the sense that these are probably not going to be in client's CI workflows for more than a month or two, so this versioning scheme is ok for now IMO.
3. I think this could be the next step: Once we have a good release process, we can start looking for automation for a certain release cadence (run a workflow that lists the stable tests for the current commit, compares against the list from the previous release, if any addition, release and list the changes).

.github/configs/feature.yaml
fselmo: Geth should have t8n support after [this](https://github.com/ethereum/go-ethereum/pull/35025) was merged, curious if there any issues

jsign: > Geth should have t8n support after [this](https://github.com/ethereum/go-ethereum/pull/35025) was merged, curious if there any issues

Oh thanks, that fixes something I pointed out in https://github.com/ethereum/go-ethereum/pull/34972#issuecomment-4522274993 :)

taratorio:
This is great, thanks. Just one question about benchmark releases. Does it make sense for those to also have benchmark-consensus and benchmark-devnet variants? For example, recently ive been hitting a slight complication which ive had to work around - the benchmark fixtures dont have BALs in them so I couldnt use them locally for optimisation work on top of the devnet changes. I had to synthesise my own fixtures essentially. If we had benchmark-devnet variant for those I wouldve not had to do this workaround

.github/configs/feature.yaml
danceratopz: Should we append 200M here to reflect the minimum glammie target value from the interop? Or other values @LouisTsai-Csie 

Can also be a follow-up!

.github/configs/evm.yaml
danceratopz, line 1: The fact that we configure a `fill` flag here is initially surprizing, a boolean `supports-xdist` instead of `x-dist` would show clearer intent. But probably not worth the overhead.

danceratopz, line 1: Nit: Can we rename `x-dist` to `xdist`? I've never seen it hyphenated anywhere else )

.github/configs/evm.yaml
```diff
@@ -1,13 +1,39 @@
...
+  ref: master
+  evm-bin: evm
+  x-dist: auto
+consensus:
+  impl: eels
+  repo: null
+  ref: null
+  evm-bin: null
+  x-dist: auto
+
```
danceratopz: For `consensus` this seems like an unnecessary redirct? I'd skip this obfuscation and just use `eels`, respectively `geth` in `.github/configs/feature.yaml`.

And, same for `benchmark`, assuming we only have one type of `benchmark` feature (as is the case in this PR - it removes `benchmark_fast`).

.github/configs/feature.yaml
danceratopz: I think we still need `--generate-all-formats` here. The changes to `packages/testing/src/execution_testing/specs/benchmark.py` only remove `blockchain_test_engine` fixture formats for `benchmark` tests.
```suggestion
  fill-params: --fork=Amsterdam --generate-all-formats --gas-benchmark-values 1,10,30,60,100,150 ./tests/benchmark
```

.github/configs/feature.yaml
```diff
@@ -1,19 +1,12 @@
 # Unless filling for special features, all features should fill for previous forks (starting from Frontier) too
```
danceratopz:
```suggestion
# Release feature definitions consumed by the release_fixtures workflow.
#
# Each top-level key is either:
#   1. A literal feature name used verbatim in the release tag, e.g.
#      `consensus` -> `tests-consensus@vX.Y.Z`, `benchmark` -> `tests-benchmark@vX.Y.Z`.
#   2. The shared `devnet` entry, which is resolved by `-devnet` suffix:
#      any workflow input feature ending in `-devnet` (e.g. `bal-devnet`,
#      `glamsterdam-devnet`) maps here while keeping its `<feat>-devnet`
#      name in the tag (`tests-<feat>-devnet@vX.Y.Z`). The devnet index
#      is encoded in the version (X), not in the feature name -- so
#      `glamsterdam-devnet-5` (branch) releases as feature
#      `glamsterdam-devnet` at version `v5.0.0`, yielding the tag
#      `tests-glamsterdam-devnet@v5.0.0`. This means `feature.yaml` does
#      not need to be updated for new devnets.
#
# Unless filling for special features, all features should fill for previous forks (starting from Frontier) too
```

.github/configs/feature.yaml
```diff
@@ -1,19 +1,12 @@
...
-  feature_only: true
-
-benchmark_fast:
-  evm-type: benchmark
-  fill-params: --fork=Osaka --generate-all-formats --gas-benchmark-values 100 ./tests/benchmark/compute
-  feature_only: true
+  fill-params: --fork=Amsterdam --gas-benchmark-values 1,10,30,60,100,150 ./tests/benchmark
 
-bal:
+devnet:
```
danceratopz, line 23:
```suggestion
# Shared entry for all `<feat>-devnet` releases; matched by `-devnet` suffix.
devnet:
```

.github/workflows/benchmark.yaml
```diff
@@ -83,18 +83,3 @@ jobs:
 
   # TODO: Add execute remote tests with --gas-benchmark-values
   # TODO: Add execute remote tests with --fixed-opcode-count
-
-  build-artifact:
```
danceratopz, line 92: This is quite slow and we should remove it if no-one is using the artifacts, but this has been a nice benchmark release sanity check. Also for hive startup and genesis file production.

Perhaps we can come up with a faster replacement (lower gas limit and not artifact production)?

.github/workflows/release_fixtures.yaml
```diff
@@ -17,14 +39,46 @@ jobs:
...
+            exit 1
+          fi
+
+          # `<feat>-devnet` releases (e.g. bal-devnet) resolve to the shared
+          # `devnet` feature but keep their friendly name, and must specify
+          # the branch to release from.
+          if [ "$FEATURE_NAME" = "devnet" ] || [ "$FEATURE_NAME" = "-devnet" ]; then
+            echo "::error::devnet releases require a <feat>- prefix, e.g. bal-devnet"
+            exit 1
+          fi
```
danceratopz: Would this be too much to check the unwanted/unrequired suffix path? I could imagine falling for this one ))
```suggestion
          if [ "$FEATURE_NAME" = "devnet" ] || [ "$FEATURE_NAME" = "-devnet" ]; then
            echo "::error::devnet releases require a <feat>- prefix, e.g. bal-devnet"
            exit 1
          fi
          # Reject `<feat>-devnet-<n>` early: the devnet index belongs in
          # the version (X of vX.Y.Z), not in the feature name.
          if [[ "$FEATURE_NAME" == *-devnet-* ]]; then
            SUGGESTED_FEATURE="${FEATURE_NAME%-*}"
            SUGGESTED_INDEX="${FEATURE_NAME##*-}"
            echo "::error::devnet index must go in 'version', not the feature name; did you mean feature=${SUGGESTED_FEATURE} version=v${SUGGESTED_INDEX}.0.0?"
            exit 1
          fi
```

danceratopz: In general, though I would prefer to put as many of these checks into the Python script and have it validated in Python and made testable. At least for the checks that only validate the `$FEATURE_NAME` input.

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -316,7 +315,6 @@ class BenchmarkTest(BaseTest):
         Sequence[FixtureFormat | LabeledFixtureFormat]
     ] = [
         BlockchainFixture,
-        BlockchainEngineFixture,
```
danceratopz, line 320: This means it'll be impossible to generate `blockchain_test_engine` fixtures from a benchmark test. Which is fine as no one uses it! 👍 

Tbh, in general though, I would love to have an explicit and clean way to tell `fill` which outputs to generate. Currently, the only real way to specify fixture format outputs is via `-m` which is a bit clumsy, so I'd prefer something like `--formats=blockchain_test,blockchain_test_enginex`. For another PR.

I would prefer this kind of change to be a smaller, targeted PR, it gets a bit lost here.

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
 # EEST Fixture Releases
 
+!!! warning "Two repositories — do not confuse them"
```
danceratopz: I don't think this needs to be top-level content. People rarely need older releases and hopefully in a month when everyone has moved to execution-specs releases, this will be defunct and unnecessary.

It also bunldes a spec release and test release warning in a "Two repositories" warning (spec releases are a different topic).

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
-        ├── prague/
-        │   └── eip7702_set_code_tx/...
-        └── osaka/                    # tests introduced in Osaka
-            └── eip7692_eof_v1/...
-```
+| Blob Transaction Tests                                               | - using the [eels/execute-blobs Simulator](./execute/hive.md#the-eelsexecute-blobs-simulator)                                                                                                                                                                                                                         | None; executed directly from Python source,</br>using a release tag |
 
-Other format directories (`state_tests/`, `blockchain_tests_engine/`)
-follow the same layout.
+## Release Tracks
```
danceratopz: Tracks feels like jargon? I can't imagine saying "tracks" when explaining this to someone ))
```suggestion
## Test Release Types
```

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
 # EEST Fixture Releases
 
+!!! warning "Two repositories — do not confuse them"
+    - **`ethereum/execution-specs`** publishes Python *spec package* releases tagged `vX.Y.Z` (e.g. [`v2.20.0`](https://github.com/ethereum/execution-specs/releases/tag/v2.20.0)). **These contain no test fixtures** — they are the executable specification package only.
+    - **Test fixtures** are published as **feature-scoped releases** on the same repository — `consensus@vX.Y.Z`, `<feat>-devnet@vX.Y.Z`, `benchmark@vX.Y.Z` — and are *not* attached to the `vX.Y.Z` package tags.
+    - The legacy `fixtures_stable.tar.gz` / `fixtures_develop.tar.gz` artifacts (previously on `ethereum/execution-spec-tests`) are being retired in favour of the feature-scoped releases described here.
+
 ## Formats and Release Layout
```
danceratopz: Release layout is no longer document in this section.

I think we can move the formats and directory layout docs to below the "Release Tracks" or "Test Release Types".

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
-### Benchmark fixture layout
+Fixtures are released on independent tracks. Each track has its own tag namespace, artifact, and cadence.
 
-When filling with `--gas-benchmark-values`, benchmark tests additionally
-include the gas limit in the subdirectory name (`for_{fork}_at_{gas}M`,
-where `{gas}` is in millions, zero-padded to four digits), with one
-subdirectory per gas value:
+| Track     | Tag                  | Artifact                        | Scope                                                              | Built from        |
+| --------- | -------------------- | ------------------------------- | ------------------------------------------------------------------ | ----------------- |
+| Consensus | `consensus@vX.Y.Z`   | `fixtures_consensus.tar.gz`     | All forks, all tests (including legacy tests)                       | latest `forks/*` branch |
```
danceratopz: Not yet but sooon! How about?
```suggestion
| Consensus | `consensus@vX.Y.Z`   | `fixtures_consensus.tar.gz`     | All forks, all tests (including ethereum/tests state tests) | latest `forks/*` branch |
```

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
-`{N}` is in thousands and may include decimals):
-
-```text
-fixtures/
-└── blockchain_tests/
-    ├── for_osaka_at_opcount_10K/     # 10K opcodes
-    │   └── benchmark/compute/...
-    └── for_osaka_at_opcount_20K/     # 20K opcodes
-        └── benchmark/compute/...
+## Creating a Fixture Release
```
danceratopz: This is great, but not interesting for the general public, I'd move this to `docs/dev/releasing_tests.md`.

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
+| Benchmark | `benchmark@vX.Y.Z`   | `fixtures_benchmark.tar.gz`     | EVM benchmarking tests                                             | latest `forks/*` branch |
 
-```text
-fixtures/
-└── blockchain_tests/
-    ├── for_osaka_at_0001M/           # 1M gas benchmark
-    │   └── benchmark/compute/...
-    └── for_osaka_at_0002M/           # 2M gas benchmark
-        └── benchmark/compute/...
-```
```
danceratopz, line 65: I think this is helpful and would move it (and the "Formats" section) below "Release Tracks" (or "Test Release Types" if we rename).

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
+4. **Review and publish the draft.** Open the draft on the [releases page](https://github.com/ethereum/execution-specs/releases), check the auto-generated notes (anchored at the prior release on the same track via `--notes-start-tag`), and click *Publish release* when ready.
 
-Devnet releases should be treated as WIP and may not yet contain full test coverage (or even coverage for all EIPs). The coverage provided by these releases is detailed in the corresponding release notes.
+!!! tip "Release features opt into all fixture formats via `feature.yaml`"
+    Tarball output (`.tar.gz`) does not by itself include the pre-allocation group formats (`BlockchainEngineXFixture`, `BlockchainEngineStatefulFixture`). A release feature requests them by adding `--generate-all-formats` to its `fill-params` in `.github/configs/feature.yaml`:
+    ```console
+    # Automatically enables --generate-all-formats due to .tar.gz output
+    uv run fill --output=fixtures_consensus.tar.gz tests/
+    ```
```
danceratopz: The paragraph and the code comment contradict each other? 

> `# Automatically enables --generate-all-formats due to .tar.gz output`

This was removed in https://github.com/ethereum/execution-specs/pull/2702

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
-- `<MINOR>`: Additional coverage (new tests, or a new format) have been added to the release.
-- `<PATCH>`: A bug-fix release; an error in the tests or framework has been patched.
+### Devnet releases
 
-Please see below for an explanation of the optional `<pre_release_name>` that is used in pre-releases.
+Devnet releases must use a `<feat>-devnet` feature name (e.g. `feature=bal-devnet`) and must specify the branch to release from:
 
-### Standard Releases
+```bash
+gh workflow run release_fixtures.yaml -f feature=bal-devnet -f version=v1.0.0 -f branch=bal-devnet-7
```
danceratopz: I believe this would be an invalid combination of tag and branch by definition? Unless the MAJOR version would get automatically populated.
```suggestion
gh workflow run release_fixtures.yaml -f feature=bal-devnet -f version=v7.0.0 -f branch=bal-devnet-7
```

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
+3. Publishes a **draft pre-release** to [`ethereum/execution-specs`](https://github.com/ethereum/execution-specs/releases), titled `<feature>@vX.Y.Z` (no `tests-` prefix), with the fixture tarball(s) attached.
 
-!!! tip "Release features opt into all fixture formats via `feature.yaml`"
-    Tarball output (`.tar.gz`) does not by itself include the pre-allocation group formats (`BlockchainEngineXFixture`, `BlockchainEngineStatefulFixture`). A release feature requests them by adding `--generate-all-formats` to its `fill-params` in `.github/configs/feature.yaml`:
-    ```console
-    uv run fill --generate-all-formats --output=fixtures_stable.tar.gz tests/
-    ```
+| Example dispatch | Git tag | Release title | Artifact |
+| ---------------- | ------- | ------------- | -------- |
+| `feature=consensus version=v1.2.3` | `tests-consensus@v1.2.3` | `consensus@v1.2.3` | `fixtures_consensus.tar.gz` |
```
danceratopz: Just to help cognitive expectations :)
```suggestion
| `feature=consensus version=v20.2.3` | `tests-consensus@v20.2.3` | `consensus@v20.2.3` | `fixtures_consensus.tar.gz` |
```

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
+| `feature=bal-devnet version=v1.0.0 branch=bal-devnet-7` | `tests-bal-devnet@v1.0.0` | `bal-devnet@v1.0.0` | `fixtures_bal-devnet.tar.gz` |
```
danceratopz: Invalid version/tag I believe? It must be compatible with the target devnet version.
```suggestion
| `feature=bal-devnet version=v7.0.0 branch=bal-devnet-7` | `tests-bal-devnet@v7.0.0` | `bal-devnet@v7.0.0` | `fixtures_bal-devnet.tar.gz` |
```

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
+The release is created as a draft; review and publish it from the GitHub releases page.
 
-### Pre-Release and Devnet Releases
+### Tagging A Release
 
-Intermediate releases that target specific subsets of features or tests under active development are published at @ethereum/execution-specs [releases](https://github.com/ethereum/execution-specs/releases).
+To cut a new release:
 
-These releases are tagged using the format `<pre_release_name>@vX.Y.Z`.
+1. **Pick the next version** per the [Versioning Scheme](#versioning-scheme) for the track you're releasing on (e.g. the next consensus release after `consensus@v20.3.0` is `consensus@v20.3.1` for a tests-only bump, or `consensus@v20.4.0` for a behaviour change).
```
danceratopz: I think this could be less ambiguous: 
- tests-only = new tests?
- behavior change = fixed tests/specs?

docs/running_tests/releases.md
```diff
@@ -1,127 +1,129 @@
...
+2. **Dispatch the workflow** from the [Actions tab](https://github.com/ethereum/execution-specs/actions/workflows/release_fixtures.yaml) or via the CLI:
 
-<!-- TODO: These example tags live in the legacy @ethereum/execution-spec-tests repo; replace with current @ethereum/execution-specs examples once available. -->
-Examples:
+   ```bash
+   gh workflow run release_fixtures.yaml -f feature=consensus -f version=v20.3.1
+   # devnet releases additionally require the branch to release from:
+   gh workflow run release_fixtures.yaml -f feature=bal-devnet -f version=v1.0.0 -f branch=bal-devnet-7
```
danceratopz: Perhaps this should also be tested by the workflow that the branch "suffix" matches the target release major version?
```suggestion
   gh workflow run release_fixtures.yaml -f feature=bal-devnet -f version=v7.0.0 -f branch=bal-devnet-7
```

.github/workflows/release_fixtures.yaml
```diff
@@ -1,10 +1,32 @@
...
   workflow_dispatch:
+    inputs:
+      feature:
+        description: "Feature name, e.g. consensus, benchmark, bal-devnet"
+        required: true
+        type: string
+      version:
+        description: "Release version, e.g. v20.0.0"
+        required: true
+        type: string
```
danceratopz, line 13: How about making this `Y.Z` and:
1. Devnet release type: Extracting `X` from `devnets/bal/X` (or in potentially in the future `devnets-bal-X`. 
2. Consensus release type: Hard-coding in `.github/configs/feature.yaml` (close to where `--until=<fork>` is defined).
3. Benchmark release type: Hard-code in `.github/configs/feature.yaml`.

LouisTsai-Csie:
I support @taratorio 's idea if this does not complicate the workflow too much

.github/workflows/release_fixtures.yaml
```diff
@@ -17,14 +39,46 @@ jobs:
...
       - uses: ./.github/actions/setup-uv
 
-      - name: Generate build matrix
+      - name: Validate input and generate build matrix
         id: matrix
         shell: bash
+        env:
+          INPUT_FEATURE: ${{ inputs.feature }}
+          INPUT_VERSION: ${{ inputs.version }}
+          INPUT_BRANCH: ${{ inputs.branch }}
```
danceratopz, line 56: This is somewhat of an edge case, but it's valid and there's an easy fix that's probably worth adding.

From codex:

There's a potential race condition when `INPUT_BRANCH` is set, e.g. `devnets/bal/7`):
  - setup, build (matrix), combine, release **each call actions/checkout with: `ref: ${{ inputs.branch }}`** → resolves the current branch tip independently at each checkout (lines 42, 97, 118, 174).
  - Release step: `TARGET="${INPUT_BRANCH:-$GITHUB_SHA}"` → `INPUT_BRANCH` (a branch name string) → `gh release create --target bal-devnet-7` → resolves the current branch tip again at tag creation time (line 221).
  - That's 5+ independent snapshots of the branch tip, any of which can disagree if the branch advances during the (potentially 24h =job timeout) build.
  - `GITHUB_SHA` doesn't enter into this path at all — it's whatever the dispatch ref's tip was at dispatch time, which is typically a different ref (e.g. `forks/amsterdam`) and never gets consulted in this path.

No problem when `INPUT_BRANCH` is empty:
  - Every actions/checkout is called with `ref: ${{ inputs.branch }}` → empty → checkout falls back to `GITHUB_SHA`.
  - Release step: `TARGET="${INPUT_BRANCH:-$GITHUB_SHA}"` → `GITHUB_SHA`. 
  - `GITHUB_SHA` is immutable for the duration of the run.
  - Race-free. All jobs see the same SHA, tag lands on the same SHA. Done.

Suggested fix:
```
  # in `setup` job, after checkout:
  - name: Resolve target SHA
    id: target_sha
    run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
```
Add `target_sha` to `setup.outputs`, then replace every `ref: ${{ inputs.branch }}` with `ref: ${{ needs.setup.outputs.target_sha }}` (in build, combine, release), and `--target "$TARGET"` with `--target "${{ 
  needs.setup.outputs.target_sha }}"`. Now there's exactly one tip resolution; everything downstream is SHA-pinned.

danceratopz (commented):
Thanks so much for this effort! Looking forward to getting all releases in execution-specs!

I feel a little bit weird about hijacking the MAJOR version for fork/devnet version (fork version is MINOR in EELS releases, so another slight inconsistency). We do have to hope PandaOps never launches `feat-devnet-alphaone` or similar to avoid invalid versions 😆 

On face value, it seems to fit well. And i think one of your major motivations is to avoid hard-coding releases in the [hive-tests](https://github.com/ethpandaops/hive-tests) runner config repos...
https://github.com/ethpandaops/hive-tests/blob/52cf2fcfa6bb698c11344994d8e223e69dd3a969/.github/workflows/hive-devnet-7.yaml#L161-L162

But it might be worth thinking through a little more. If we have two devnet versions (`bal-devnet-3`,`bal-devnet-7`) running at the same time, does it still simplify artifact download? I guess we never have hive runner configs for two devnets simultaneously?

As-is we could already do:
```
consume cache --input=bal-devnet-7@latest
```
and if we aim to deprecate `consume cache` then both versions are equally convenient with `gh release download`, I think?

Versioning scheme in this PR:
```bash
#!/usr/bin/env bash

TAG=$(gh release list --repo ethereum/execution-specs --limit 100 --json tagName \
    --jq 'map(select(.tagName | startswith("tests-bal-devnet@v7."))) | .[0].tagName')
gh release download "$TAG" --repo ethereum/execution-specs --pattern '*.tar.gz'
```


Versioning scheme currently:
```bash
#!/usr/bin/env bash
TAG=$(gh release list --repo ethereum/execution-specs --limit 100 --json tagName \
    --jq 'map(select(.tagName | startswith("tests-bal-devnet-7@v"))) | .[0].tagName')
gh release download "$TAG" --repo ethereum/execution-specs --pattern '*.tar.gz'
```

If this does not greatly help downstream convenience I would opt for explicit hard-coded release names here `bal-devnet-7`) and not add our own custom convention to versioning.

If we keep it, I wonder, at the risk of complicating the workflows, we should automate/hard-code the major version to avoid user error (see the docs corrections below) as suggested [here](https://github.com/ethereum/execution-specs/pull/2888/changes#r3302365615). This would avoid duplicating this in the case of the devnet branch and put it close to the `fill --until=<fork>` config in the case of a fork branch. The [comment](https://github.com/ethereum/execution-specs/pull/2888/changes#r3302256151)s on the invalid tag/version highlight that this is error prone. 

Could you restructure the docs to keep all (or most of) the "Formats and Release Layout" section, but move it below the "Release Tracks" (or "Test Release Types" if we rename)?

I think removing `blockchain_test_engine` from the `benchmark` spec deserves its own PR as it gets a bit lost here.

danceratopz:
> This is great, thanks. Just one question about benchmark releases. Does it make sense for those to also have benchmark-consensus and benchmark-devnet variants? 

Hey @taratorio, thanks for asking and the feedback! Yes, this def makes sense in general and we can add these as required. 

> For example, recently ive been hitting a slight complication which ive had to work around - the benchmark fixtures dont have BALs in them so I couldnt use them locally for optimisation work on top of the devnet changes. I had to synthesise my own fixtures essentially. If we had benchmark-devnet variant for those I wouldve not had to do this workaround

Were you filling these fixtures yourself? The latest release https://github.com/ethereum/execution-specs/releases/tag/tests-benchmark%40v0.0.9 is only filled for Osaka. There might have been a bit of flux here due to incompatible/incomplete t8n tools for EELS and geth (benchmark releases have been using geth), but as-is today :laughing: on `forks/amsterdam` with https://github.com/ethereum/go-ethereum/pull/35025 both regular and benchmark test fixtures have `blockAccessList`. They should be in the next benchmark release, which will target Amsterdam!

taratorio:
> > This is great, thanks. Just one question about benchmark releases. Does it make sense for those to also have benchmark-consensus and benchmark-devnet variants?
> 
> Hey @taratorio, thanks for asking and the feedback! Yes, this def makes sense in general and we can add these as required.
> 
> > For example, recently ive been hitting a slight complication which ive had to work around - the benchmark fixtures dont have BALs in them so I couldnt use them locally for optimisation work on top of the devnet changes. I had to synthesise my own fixtures essentially. If we had benchmark-devnet variant for those I wouldve not had to do this workaround
> 
> Were you filling these fixtures yourself? The latest release https://github.com/ethereum/execution-specs/releases/tag/tests-benchmark%40v0.0.9 is only filled for Osaka. There might have been a bit of flux here due to incompatible/incomplete t8n tools for EELS and geth (benchmark releases have been using geth), but as-is today 😆 on `forks/amsterdam` with [ethereum/go-ethereum#35025](https://github.com/ethereum/go-ethereum/pull/35025) both regular and benchmark test fixtures have `blockAccessList`. They should be in the next benchmark release, which will target Amsterdam!

yes, I filled a few that I was interested in with erigon but not via t8n but my own quick and hacky way

danceratopz:
Just discussed with @spencer-tb and @LouisTsai-Csie, this is our suggestion going forward:

1. Instead of changing what is now "mainnet" to "consensus" (in this PR), we just tag these releases as `tests@vX.Y.Z`. No stable or other labelling required, they're just the tests. We aim for a on-demand release schedule, which could be as frequently as weekly if tests are added/fixed. The forks covered/filled-for in these releases should be slightly ahead of clients' testnet/mainnet release schedules. I.e., we include Amsterdam in good time before for the first testnet releases.
2. The benchmark changes will move to another PR. Benchmark devnet releases:
    - A: For devnet 7, merge 8037 PR, bump Osaka->Amsterdam, pick/decide on an EVM :), then create the release from forks/amsterdam.
    - B: In general, for now, if necessary just add a new entry to feature.yaml (Move benchmark changes to another PR). I.e., if we need a benchmark-glamsterdam-devnet-5@v1.0.0, we can create a new entry for that.
4. We simplify release the versioning scheme that can be applied to any test fixture release to:
    X: fix spec (spec change)
    Y: fix test
    Z: new test
5. The next/new release versions start at:
    - `tests@v1.0.0`
    - `benchmark@v1.0.0` maybe from the devnet-7 release on / maybe the next one TBD ??
    - next devnet release -` <feat>-devnet-<N>@v1.0.0`

marioevz:
> Just discussed with @spencer-tb and @LouisTsai-Csie, this is our suggestion going forward:
> 
>     1. Instead of changing what is now "mainnet" to "consensus" (in this PR), we just tag these releases as `tests@vX.Y.Z`. No stable or other labelling required, they're just the tests. We aim for a on-demand release schedule, which could be as frequently as weekly if tests are added/fixed. The forks covered/filled-for in these releases should be slightly ahead of clients' testnet/mainnet release schedules. I.e., we include Amsterdam in good time before for the first testnet releases.
> 
>     2. The benchmark changes will move to another PR. Benchmark devnet releases:
>        
>        * A: For devnet 7, merge 8037 PR, bump Osaka->Amsterdam, pick/decide on an EVM :), then create the release from forks/amsterdam.
>        * B: In general, for now, if necessary just add a new entry to feature.yaml (Move benchmark changes to another PR). I.e., if we need a benchmark-glamsterdam-devnet-5@v1.0.0, we can create a new entry for that.
> 
>     3. We simplify release the versioning scheme that can be applied to any test fixture release to:
>        X: fix spec (spec change)
>        Y: fix test
>        Z: new test
> 
>     4. The next/new release versions start at:
>        
>        * `tests@v1.0.0`
>        * `benchmark@v1.0.0` maybe from the devnet-7 release on / maybe the next one TBD ??
>        * next devnet release -` <feat>-devnet-<N>@v1.0.0`

I mostly agree with this comment except for points 3 and 4, since I think the MAJOR should refer to the devnet/fork:

My suggestion is as follows:

For `tests@vX.Y.Z`:
- X: Fork-based number
- Y: Consensus-breaking spec change targeting fork X
- Z: Non-breaking spec change (refactoring), new tests, modified tests

For `<feat>-devnet@vX.Y.Z`:
- X: Devnet number
- Y: Consensus-breaking spec change targeting devnet X
- Z: Non-breaking spec change (refactoring), new tests, modified tests

I.e. if a client is targeting to join fork/devnet X, it should target MAJOR equal to X, must take the latest MINOR, and should ideally take the latest PATCH.

With fork/devnet as MAJOR, the version number alone tells a client whether a release is relevant to them. Under the alternative (MAJOR tracks any spec change), you'd have to combine the version and the -N suffix in the release name to figure out whether a spec change targets your devnet or a different one. Devnet compatibility shouldn't require parsing two fields IMO.

Concretely, the mindset just from looking at the version (ignoring the name) should be: I'm a client dev targeting devnet 7, and I was passing tests contained in`v7.0.0`, but now `v7.1.0` has been released, which means there was a spec change in the devnet my client was targeting, hence I should read the release notes to figure out if the spec change affects my client's ability to join the devnet.

On the topic of parallel maintenance of two different devnets, this scheme handles this naturally IMO. E.g. if we are maintaining devnets 3 & 7 for the same feature, releasing v3.0.1 after or alongside v7.0.0 is not a problem (think Python 2.x vs 3.x). It is a well-known Semver pattern, and Semver is good at this. We should simply make this rule clear and follow it so we are predictable.

On `consume cache`, it's a solvable problem. We will need to update our tooling, but it's not a big issue.

On benchmarking, major and minor should mirror the feature they are targeting, while the patch moves freely at its own pace.

danceratopz:
:ship: :shipit:

.github/configs/feature.yaml
spencer-tb: Restored!

.github/configs/feature.yaml
spencer-tb: Let's make it a follow up, I'm pulling the benchmark changes out of this PR

.github/configs/evm.yaml
spencer-tb: Keeping as-is for now.

spencer-tb: Done!

.github/configs/evm.yaml
spencer-tb: No longer using consensus.

spencer-tb: Benchmark is kept, as without it we have to update the evm version is 2 places. 

Keeping the alias here means only one place needs updated, and shows clearly what t8n we are currently using for benchmarking which is a moving target atm

.github/configs/feature.yaml
spencer-tb: Moving these changes to a benchmarking follow up. I don't think we need the flag after local testing.

.github/configs/feature.yaml
spencer-tb: Added, and applied with the `consensus` to `tests` rename. Tweaked the tag example since the `tests` feature tags as `tests@vX.Y.Z` directly.

.github/configs/feature.yaml
spencer-tb: Added!

.github/workflows/benchmark.yaml
spencer-tb: Reverted out of this PR, and adding to a follow up.

.github/workflows/release_fixtures.yaml
spencer-tb: Added this but implemented in `generate_build_matrix.py`, PTAL: https://github.com/spencer-tb/execution-specs/blob/93a501c950cfa78213ac01fae4830d9737414851/.github/scripts/generate_build_matrix.py#L76-L125

packages/testing/src/execution_testing/specs/benchmark.py
spencer-tb: Will add the explicit flag instead, moved to a follow up PR.

.github/workflows/release_fixtures.yaml
spencer-tb: Done, as per above.

docs/running_tests/releases.md
spencer-tb: Done!

docs/running_tests/releases.md
spencer-tb: Done!

docs/running_tests/releases.md
spencer-tb: Done!

docs/running_tests/releases.md
spencer-tb: Applied with the consensus to tests rename

docs/running_tests/releases.md
spencer-tb: Done!

docs/running_tests/releases.md
spencer-tb: Done!

docs/running_tests/releases.md
spencer-tb: Fixed!

docs/running_tests/releases.md
spencer-tb: Fixed, all devnet examples now use `version=v7.0.0` with `branch=bal-devnet-7`

docs/running_tests/releases.md
spencer-tb: Updated!

docs/running_tests/releases.md
spencer-tb: Applied!

docs/running_tests/releases.md
spencer-tb: Updated this, replaced the ambiguous wording with the scheme from Marios last comment: `Y` = consensus breaking spec change targeting fork/devnet `X`, `Z` = non breaking change (refactor) / new / modified tests.

docs/running_tests/releases.md
spencer-tb: Done, the check is now in `validate_inputs()` via the python script.

.github/workflows/release_fixtures.yaml
spencer-tb: Went a different route here as per Mario's last comment. We now validate X for devnet releases only. We can add further validation to a follow up imo.

.github/workflows/release_fixtures.yaml
spencer-tb: Added the exact fix here thanks!

spencer-tb:
PR is now consolidated. Description updated. Benchmark release changes are moved to this PR: https://github.com/ethereum/execution-specs/pull/2954.

docs/running_tests/releases.md
danceratopz: Still missing!

docs/dev/releasing_tests.md
```diff
@@ -0,0 +1,108 @@
...
+  number in the branch (so `feature=bal-devnet branch=bal-devnet-7` must use `version=v7.*.*`).
+
+## Devnet releases
+
+Devnet releases must use a `<feat>-devnet` feature name (e.g. `feature=bal-devnet`) and must
+specify the branch to release from. The `version` major must match the devnet number in the
+branch:
+
+```bash
+gh workflow run release_fixtures.yaml -f feature=bal-devnet -f version=v7.0.0 -f branch=bal-devnet-7
```
danceratopz: This (and the changes in `.github/scripts/`) is forward looking to an EELS branch scheme rename? 

`devnets/bal/7` -> `bal-devnet-7`.

I just asked in the chat ;-)

.github/workflows/release_fixtures.yaml
danceratopz: Even better!

danceratopz: Still missing a unicode rejection check ;-)

packages/testing/src/execution_testing/specs/benchmark.py
danceratopz: This can be added as a small separate feature:
- https://github.com/ethereum/execution-specs/issues/2959

.github/scripts/generate_build_matrix.py
```diff
@@ -61,11 +71,70 @@ def load_config(path: Path) -> dict:
...
+            )
+        devnet_number = int(match.group(1))
+        major = int(version.lstrip("v").split(".")[0])
+        if major != devnet_number:
+            minor_patch = version.split(".", 1)[1]
+            fail(
+                f"version major (v{major}) must equal the devnet number "
+                f"({devnet_number}) from branch '{branch}'; "
+                f"did you mean version=v{devnet_number}.{minor_patch}?"
+            )
```
danceratopz, line 134: This validation assumes execution-specs uses `<feat>-devnet-<X>` scheme, currently we use `devnets/bal/7`.

We need to sort this out first, there's some debate in the aforementioned chat.

danceratopz (commented):
Thanks for handling all the comments, respectively pushing the benchmark changes to #2954!

I would like to go over the docs again before merge, but posting this beforehand as the larger blocker right now: Both the validation scripts and docs assume our branches have the fork `bal-devnet-7` (as the rest of the ecosystem uses), whereas we currently use `devnets/bal/7` in EELS.

docs/dev/releasing_tests.md
danceratopz: Updated for the existing scheme (`devnets/<feat>/<x>`) in:
- https://github.com/spencer-tb/execution-specs/pull/20

.github/scripts/generate_build_matrix.py
danceratopz: Updated for the existing scheme (`devnets/<feat>/<x>`) in:
- https://github.com/spencer-tb/execution-specs/pull/20

.github/configs/feature.yaml
```diff
@@ -13,7 +19,7 @@ benchmark_fast:
   fill-params: --fork=Osaka --generate-all-formats --gas-benchmark-values 100 ./tests/benchmark/compute
   feature_only: true
 
-glamsterdam-devnet:
+# Shared entry for all `<feat>-devnet` releases; matched by `-devnet` suffix.
+devnet:
   evm-type: eels
-  fill-params: --fork=Amsterdam
-  feature_only: true
+  fill-params: --until=Amsterdam --generate-all-formats
```
danceratopz: Should we remove this for the first release to keep it quick and allow a faster iteration while testing? We can add it back quickly enough if need be.
```suggestion
  fill-params: --until=Amsterdam
```

danceratopz: Just on devnet releases. I would leave it for the stable release.

danceratopz (approved):
Thanks @spencer-tb This LGTM! Approved, just one last question.... I don't mind either way!

docs/running_tests/releases.md
danceratopz: Added back in:
- https://github.com/spencer-tb/execution-specs/pull/20

.github/configs/feature.yaml
spencer-tb: Applied!

PR #2889, opened 2026-05-20
feat(zkevm): support external t8n stateless input/output byte generation

jsign:
## 🗒️ Description
zkEVM added three important fields to fixtures: `executionWitness`, `statelessInputBytes` and `statelessOutputBytes`. The last two are the main crucial ones since are the raw bytes received by the stateless validator (i.e. guest programs) to validate a block (and thus generate a proof).

Currently, we can fill test fixtures but not benchmark fixtures -- mainly because filling benchmarks with the Python spec is not possible given that these are quite heavy blocks and is not doable. This isn't something particular to zkEVM efforts, but a general one (e.g. for Glamsterdam efforts, benchmarks aren't filled with Python either).

This means that benchmark fixtures should be filled with an external t8n implementation, usually based on some real EL that is performant enough. Geth is currently one of the main clients used in EEST for this purpose.

This PR:
- Allows to fill any fixture (in particular, benchmark fixtures) using an external t8n tool with a minimal requirement that `executionWitness` is returned as t8n output (not necessarily `stateless{Input|Output}Bytes`)
- The `stateless{Input|Ouput}Bytes` are constructed in the testing framework.

Ideally, the t8n should also return the `stateless{Input| Output}Bytes`, and it would make this PR a simpler (i.e. basically almost no diff at all), but currently Geth doesn't have good SSZ support to do this. But, it has good support for `executionWitness` generation, thus still doing the SSZ-work on this side. Eventually, we could push `stateless{Input|Output}Bytes` generation to Geth-t8n, and remove this on EEST side, which would be nice.

I also added a GH workflow that does a quick smoke test, checking that a particular benchmark fixture target filled with Python and with Geth leads to exactly the same output. See [here](https://github.com/ethereum/execution-specs/actions/runs/26174837461/job/77002149002?pr=2889) for a run.

With this change, we could create zkEVM benchmark releases. These would target spec-compliant guest programs, plus also on the latest devnet target.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/.github/workflows/benchmark-fill-parity.yaml
+++ b/.github/workflows/benchmark-fill-parity.yaml
@@ -1,0 +1,134 @@
+name: Benchmark Fill Parity
+
+on:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      geth_repo:
+        description: "Geth repository to build"
+        required: false
+        type: string
+        default: "jsign/go-ethereum"
+      geth_branch:
+        description: "Geth branch to build"
+        required: false
+        type: string
+        default: "jsign-master-t8n-fill-executionwitness"
+      benchmark_file:
+        description: "Benchmark test file to fill"
+        required: false
+        type: string
+        default: "tests/benchmark/compute/instruction/test_memory.py"
+      fill_fork:
+        description: "Fork to fill"
+        required: false
+        type: string
+        default: "Amsterdam"
+      xdist_workers:
+        description: "Number of fill workers"
+        required: false
+        type: string
+        default: "auto"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: ${{ github.ref_name != github.event.repository.default_branch }}
+
+env:
+  GETH_REPO: ${{ inputs.geth_repo || 'jsign/go-ethereum' }}
+  GETH_BRANCH: ${{ inputs.geth_branch || 'jsign-master-t8n-fill-executionwitness' }}
+  BENCHMARK_FILE: ${{ inputs.benchmark_file || 'tests/benchmark/compute/instruction/test_memory.py' }}
+  FILL_FORK: ${{ inputs.fill_fork || 'Amsterdam' }}
+  XDIST_WORKERS: ${{ inputs.xdist_workers || 'auto' }}
+
+jobs:
+  compare-fills:
+    name: Compare Python and Geth fills
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          submodules: true
+
+      - name: Setup uv
+        uses: ./.github/actions/setup-uv
+        with:
+          enable-cache: "false"
+
+      - name: Install dependencies
+        run: uv sync --no-progress
+
+      - name: Build Geth evm
+        uses: ./.github/actions/build-evm-client/geth
+        with:
+          repo: ${{ env.GETH_REPO }}
+          ref: ${{ env.GETH_BRANCH }}
+
+      - name: Validate benchmark file input
+        shell: bash
+        run: |
+          case "$BENCHMARK_FILE" in
+            ./*)
+              echo "benchmark_file must be relative to the repository root without a leading ./"
+              exit 1
+              ;;
+            tests/benchmark/*.py)
+              ;;
+            *)
+              echo "benchmark_file must be a Python file under tests/benchmark/"
+              exit 1
+              ;;
+          esac
+
+          if [ ! -f "$BENCHMARK_FILE" ]; then
+            echo "benchmark_file does not exist: $BENCHMARK_FILE"
+            exit 1
+          fi
+
+      - name: Prepare fill directories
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark-fill-parity.yaml
```diff
@@ -0,0 +1,131 @@
...
+  XDIST_WORKERS: ${{ inputs.xdist_workers || 'auto' }}
+
+jobs:
+  compare-fills:
+    name: Compare Python and Geth fills
+    runs-on: ubuntu-latest
+    timeout-minutes: 120
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
jsign, line 51: I'm using these particular commit references since they are also used in other GH workflows.

.github/workflows/benchmark-fill-parity.yaml
```diff
@@ -0,0 +1,131 @@
...
+      geth_branch:
+        description: "Geth branch to build"
+        required: false
+        type: string
+        default: "jsign-master-t8n-fill-executionwitness"
+      benchmark_file:
+        description: "Benchmark test file to fill"
+        required: false
+        type: string
+        default: "tests/benchmark/compute/instruction/test_memory.py"
```
jsign, line 21: This is just a file that fills ~150 fixtures -- feels like a light enough signal to compare Geth and Python are filling the same thing.

I think this should actually be a workflow useful for `forks/amsterdam` -- but we could think about upstreaming eventually.

.github/workflows/benchmark-fill-parity.yaml
```diff
@@ -0,0 +1,131 @@
...
+          if [ ! -f "$BENCHMARK_FILE" ]; then
+            echo "benchmark_file does not exist: $BENCHMARK_FILE"
+            exit 1
+          fi
+
+      - name: Fill benchmark with Python spec
+        run: |
+          uv run fill \
+            --clean \
+            --gas-benchmark-values 1 \
```
jsign, line 97: I'm using 1M gas value to make it easy for Python filler. I think with this we get the signal we need reg fixture matching. Also, higher gas limit won't really change much `executionWitness`/`stateless{Input|Output}Bytes` mismatchs if they exist.

.github/workflows/benchmark-fill-parity.yaml
```diff
@@ -0,0 +1,131 @@
...
+      geth_repo:
+        description: "Geth repository to build"
+        required: false
+        type: string
+        default: "jsign/go-ethereum"
+      geth_branch:
+        description: "Geth branch to build"
+        required: false
+        type: string
+        default: "jsign-master-t8n-fill-executionwitness"
```
jsign, line 16: Of course, I had to do some work in Geth to:
- Be aware of the existing new t8n-input `BlockHeaders`, since it is required for `headers` execution witness construction.
- Add `ExecutionWitness` and BAL fields as part of t8n-output

Nothing new here reg t8n API definition -- is just making Geth be aware of these new optional fields and use/return them correctly.

It also has an important fix for correct execution witness generation, but that's something more related to Geth than EEST. I'm planning to eventually create the PR in geth, so we can try using Geth for this without my fork.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1281,6 +1524,32 @@ def generate_block_data(
         stateless_output_bytes = (
             transition_tool_output.result.stateless_output_bytes
         )
+        if (
+            not skip_stateless_for_block
+            and t8n_witness is not None
+            and bal is not None
+            and (
+                stateless_input_bytes is None or stateless_output_bytes is None
+            )
```
jsign, line 1533: If hte t8n output does have `stateless_input_bytes` and `stateless_output_bytes`, then we are probably filling with Python t8n so "nothing new is done".

But if that isn't the case, then we are probably using t8n output from Geth which only returns `executionWitness`. What `build_amsterdam_stateless_artifacts_from_t8n(...)` does is constructing `StatelessInput` and `StatelessValidationResult` from parts, and doing the SSZ serialization to get the `stateless{Input|Output}Bytes`.

If at some point Geth supports returning `stateless{Input|Output}Bytes` in the t8n output, we can basically remove all the diffs in this file.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -372,6 +372,249 @@ def verify_stateless_input_public_keys(
...
+        ),
+        execution_requests=execution_requests,
+        block_access_list=_convert_amsterdam_block_access_list(
+            block_access_list
+        ),
+        chain_id=U64(chain_id),
+    )
+    stateless_input_bytes = serialize_stateless_input(stateless_input)
+    # TODO: Replace this shortcut with a client-provided validation result
+    # or a real stateless guest run once Geth witness generation is complete.
```
jsign, line 603: For now we are only going to use Geth for filling benchmarks which are always valid blocks. 

If Geth adds the `successful_validation` boolean t8n-output, we could remove this todo. But to do that meaningfully, it requires some extra changes in Geth. We should try to do it, but it isn't worth being blocked on filling benchmark tests because of this.

PR #2893, opened 2026-05-21
feat(tests): EIP-7928 self destruct to system address with 0 value

taratorio:
## Summary

Adds `test_bal_selfdestruct_to_system_address_zero_balance` to `tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip4788.py`.

The test covers the intersection of two BAL rules that previously had no joint coverage:

- **EIP-7928 inclusion list**: "Beneficiary addresses for `SELFDESTRUCT`" MUST appear in the BAL — regardless of value transfer.
- **EIP-7928 SystemAddress carve-out**: `SYSTEM_ADDRESS` (`0xff...fffe`) "MUST NOT be included unless it experiences state access itself" — and a SELFDESTRUCT to it _is_ such an access, so the exception fires.

## Scenario

A single CREATE transaction whose init code is `Op.SELFDESTRUCT(SYSTEM_ADDRESS)`, with the new contract's balance still zero (the CREATE tx carries no value). Per EIP-6780 the contract is actually deleted because creation and destruction happen in the same transaction. The resulting BAL must contain the `SYSTEM_ADDRESS` entry with every change-set empty — the SELFDESTRUCT is the only access on `SYSTEM_ADDRESS` in this block.

## Motivation

This case was identified via a real-world client divergence on a Block Access List devnet — one client dropped the `SYSTEM_ADDRESS` entry under this combination of conditions, producing a BAL hash mismatch with other clients. Existing BAL SELFDESTRUCT tests in this directory all use either a beacon-root / history-storage / withdrawal-queue / consolidation-queue beneficiary or an ordinary EOA — none of them target `SYSTEM_ADDRESS` itself, so the carve-out's "unless it experiences state access" clause was never exercised end-to-end.

## Test plan

- [x] `uv run fill tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip4788.py::test_bal_selfdestruct_to_system_address_zero_balance --fork Amsterdam` — fills against EELS t8n.
- [x] Generated blockchain_test fixture executes successfully against a spec-correct client.
- [x] Generated blockchain_test fixture deterministically reproduces the divergence against a non-spec-correct client (BAL hash mismatch, missing `SYSTEM_ADDRESS` entry).


fselmo:
Thanks @taratorio. I pointed this to `forks/amsterdam` as this is our default branch at the moment... and it can trickle back up to devnet branches (that's the force push here). I also cleaned it up to use dynamic gas limit and added this to `test_cases.md`. Will wait for CI.

fselmo (approved):
lgtm, will wait for CI 👍🏼

PR #2894, opened 2026-05-21
feat(tests,ci): Verify filled benchmark fixtures against EELS via json_loader

danceratopz:
## 🗒️ Description

Benchmark tests are filled with an external EVM (geth, evmone,...) as the EELS EVM is extremely slow at filling these compute-heavy tests. This has led to a consensus testing gap: The canonical spec (EELS) is not part of the benchmark workflow.

This PR updates the `bench-gas` recipe to additionally run the externally generated fixtures against EELS via the `json_loader`; this is then verified upon push in CI.

It does not update the target fork used for benchmarks (currently Osaka); this should be bumped!

### `bench-gas` recipe / `benchmark.yaml` workflow

This PR adds a Step 2 to `bench-gas`: after filling with geth, re-execute the resulting `blockchain_test` fixtures against EELS via the `json_loader`. EELS' `state_transition` validates each block's state root against the header internally (`src/ethereum/forks/osaka/fork.py:266`), and the `lastblockhash` assertion provides the final cryptographic check. Any gas, state, or opcode-behavior divergence between geth and EELS will now fail the verify step.

Modifications:

1. `feat(tests): Add --allow-post-state-hash option to json_loader`: previously the json_loader xfailed any fixture without a full `postState` dict. Benchmark fixtures only carry `postStateHash` (since `include_full_post_state_in_output` defaults to `False`). With the new flag, EELS' state transition runs against these fixtures and `lastblockhash` validates the resulting state root via the block header. Behavior is unchanged when the flag is absent, so the existing `json-loader` recipe is unaffected.
2. `feat(bench-gas): Verify filled fixtures against EELS via json_loader`: extends the recipe with a clearly-labeled Step 2. Symlinks the filled output under `tests/json_loader/` so the path-scoped conftest applies, then runs pytest with `--allow-post-state-hash --fork Osaka` under PyPy 3.11.
3. `ci(bench): Install PyPy for bench-gas json_loader verify`: pre-installs PyPy 3.11 in the CI job so the recipe doesn't pay the on-demand download cost. Gated to the `bench-gas` matrix entry.

PyPy gives a ~9x speedup over CPython on this workload (`1082 passed in 2:08` vs `19:32` locally), so the verification adds about two minutes to the existing CI job rather than twenty. The bottleneck is pure-Python Merkle Patricia Trie hashing and RLP encoding inside EELS' state transition, which is exactly the workload PyPy's JIT was designed for; the `fill` step still runs on CPython since it bottlenecks on subprocess I/O to the EVM binary, not on hot Python loops.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

image

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -71,6 +71,11 @@ jobs:
         with:
           enable-cache: "false"
 
+      - name: Install PyPy for bench-gas json_loader verify
+        if: matrix.recipe == 'bench-gas'
+        shell: bash
+        run: uv python install pypy3.11
+
       - uses: ./.github/actions/build-evm-base
         id: evm-builder
         with:
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,7 @@ pip-delete-this-directory.txt
 
 .tox/
 .just/
+tests/json_loader/bench_gas_fixtures
 
 /doc/_autosummary
 
--- a/Justfile
+++ b/Justfile
@@ -212,10 +212,11 @@ test-ci-scripts *args:
 
 # --- Benchmarks ---
 
-# Fill benchmark tests with --gas-benchmark-values
+# Fill benchmark tests with --gas-benchmark-values, then verify with EELS
 [group('benchmark tests')]
 bench-gas *args:
     @mkdir -p "{{ output_dir }}/bench-gas/tmp" "{{ output_dir }}/bench-gas/logs"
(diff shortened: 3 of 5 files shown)


SamWilsn (approved):

danceratopz:
I considered adding the same check to benchmark artifact generation/release flows, but leaving as-is for now as the release flow is in a it of flux #2888.

PR #2896, opened 2026-05-21
feat(test-benchmark): add more modexp precompile benchmark scenario

GottfriedHerold:
## 🗒️ Description
Add more tests for the modexp precompile.
Notably, this adds tests with essentially random moduli and exponents of a prescribed bitlength and tests for moduli that are a power of 2.
The power-of-2 cases are just for completeness sake; we do not expect implementation to fail on those.
For the random moduli: this addresses the issue that many of the explicitly defined test cases have moduli of the very special form 0xFFFF...FF00 or 0xFFFF...FF01. With the base of the form 0xFFF...FF of the same bitlength, this results in
base mod modulus being a small (8-bit!) number. If an implementation uses variable-length numbers in its precomputation with a windowed exponentation, this leads to atypical speedups due these special shapes.
Furthermore, with moduli this close to a power of two, the constants that appear in Montgomery multiplication become small as well, so these unusual lengths persists for Montgomery multiplication. Similarly phenonena may appear in Barret reduction.

Adding random test cases solves this. The exponents are chosen at 17/33/65 bit due to the fact that gas computation looks at bit lengths, but real cost in implementation may be at byte/word length granularity, so this is more likely to catch worst-cases. Indeed, the chosen parameter ranges cover the worst-case regimes that came up when extensively benchmarking the Go standard library implementation against gas cost earlier this year.

Note: Some of the test_ids are generated using an f-string from parameters. This means grep won't find it. Feel free to fix this if it violates coding standards.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
- [X] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    just static
    ```
#### Cute Animal Picture

### DIFF
--- a/tests/benchmark/compute/precompile/test_modexp.py
+++ b/tests/benchmark/compute/precompile/test_modexp.py
@@ -514,6 +514,7 @@ def create_modexp_test_cases() -> list[ParameterSet]:
     ]
     return regular_cases + special_cases
 
+
 @pytest.mark.parametrize(
     ["mod_exp_input"],
     create_modexp_test_cases(),
--- a/tests/benchmark/compute/precompile/test_modexp.py
+++ b/tests/benchmark/compute/precompile/test_modexp.py
@@ -33,23 +33,14 @@ def create_random_modexp_test_case(
     modulus_2arity: int | None = None,
 ) -> tuple[str, str, str, str]:
     """
-    Build a random MODEXP test case with exactly the prescribed bit lengths.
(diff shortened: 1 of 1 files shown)


LouisTsai-Csie (approved):
I push a commit to update comment and refactor the helper a bit, thank you!

PR #2900, opened 2026-05-22
feat(amsterdam): defer EIP-7702 delegate warming until entry checks pass

nerolation:
## 🗒️ Description

Restructure CALL/CALLCODE/DELEGATECALL/STATICCALL into pre-state gas -> free depth/balance preconditions → post-state gas + warming + BAL insertion, per EIP-7928. The target's cold-access surcharge and accessed_addresses.add(...) are deferred until after preconditions pass; EIP-2929 cost -warm pairing is preserved. generic_call's depth check is removed (callers handle it).

Additional tests still needed across CALL family - {cold/warm, depth-OK/exceeded, balance-OK/insufficient, delegated/not}.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

* [x] All: Ran fast static checks:

  ```console
  just static
  ```

* [x] All: PR title adheres to the repo standard.

* [x] All: Considered updating the online docs in the ./docs/ directory.

* [ ] All: Set appropriate labels for the changes.

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -298,14 +298,9 @@ def generic_call(
     """
     Perform the core logic of the `CALL*` family of opcodes.
     """
-    from ...vm.interpreter import STACK_DEPTH_LIMIT, process_message
+    from ...vm.interpreter import process_message
 
     evm.return_data = b""
-
-    if evm.message.depth + Uint(1) > STACK_DEPTH_LIMIT:
-        evm.gas_left += gas
-        push(evm.stack, U256(0))
-        return
 
     tx_state = evm.message.tx_env.state
     code_hash = get_account(tx_state, code_address).code_hash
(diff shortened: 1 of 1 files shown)


src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -307,6 +307,11 @@ def generic_call(
         push(evm.stack, U256(0))
         return
 
+    # EIP-7702: warm the delegated address only after all entry preconditions
+    # (gas, balance, depth) have passed. Idempotent for non-delegation paths,
+    # where `code_address` is the call target the caller already warmed.
+    evm.accessed_addresses.add(code_address)
```
gurukamath: If we move the delegation warming to a later point, in a case where the call does not have enough funds for value transfer, we will be charging the gas for warming the delegation but not actually adding it to the accessed list.

fselmo: Agree here. The gas cost needs to be done before the balance check so as long as the warming was paid for, we need to warm it. This is different than the BAL where it shows _when it was accessed_, more like a witness.

nerolation: updated it based on dragans comment:

https://discord.com/channels/595666850260713488/1506543443172921464/1510908205558005903

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -386,37 +381,45 @@ def call(evm: Evm) -> None:
...
+    # Pre-state preconditions: stack depth and (if value > 0) sender balance.
+    from ...vm.interpreter import STACK_DEPTH_LIMIT
 
-    # STATE ACCESS
     tx_state = evm.message.tx_env.state
+    sender_balance = get_account(tx_state, evm.message.current_target).balance
+    if evm.message.depth + Uint(1) > STACK_DEPTH_LIMIT or (
+        value > U256(0) and sender_balance < value
+    ):
+        charge_gas(evm, extra_gas + extend_memory.cost)
```
gurukamath, line 399: A call that does not have enough balance to send value, ends up paying only the warm cost to access the target account, even if it is cold.
In that case, the target account ends up in BAL but is never added to the access list. If we are going for a "BAL is a strict subset of access list" assumption, this scenario violates it.

nerolation: Yeah right now it looks like we won't go for this change anyway

gurukamath: @nerolation Is there consensus on a final shape for this?

nerolation: the default was not changing and clients now had 2-3 weeks to chime in. I'll let clients take the shot on this with the default closing this PR.

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -386,37 +381,45 @@ def call(evm: Evm) -> None:
...
 
-    # check static gas before state access
-    check_gas(
-        evm,
-        access_gas_cost + transfer_gas_cost + extend_memory.cost,
-    )
+    # Pre-state extra gas (warm-equivalent access cost only; cold surcharge
+    # is deferred to post-state).
+    extra_gas = GasCosts.WARM_ACCESS + transfer_gas_cost
+    check_gas(evm, extra_gas + extend_memory.cost)
```
marioevz, line 389: Maybe I'm missing something but I don't quite understand this charge. Couldn't we just `check_gas(evm, extend_memory.cost)` and immediately `if evm.message.depth + Uint(1) > STACK_DEPTH_LIMIT or (value > U256(0) and sender_balance < value)`, and charge no warm access?

marioevz: Answering my own comment: With bare `check_gas(evm, extend_memory.cost)`, `CALL` with insufficient balance would be cheaper to expand the memory (full of zeros) than `MSTORE`.

gurukamath:
The ACDT makes this discussion makes this a no-op. I think we can close this for now

PR #2904, opened 2026-05-22
chore(specs): refactor code deployment functions

SamWilsn:
## 🗒️ Description

- Combines `account_has_code_or_nonce` and `account_has_storage` into a single function, `account_deployable`.
- Expands an EIP-7610 test to cover Frontier+
- Adds support to EEST for multiple `valid_from` marks (taking the more strict mark).

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/tests/test_bad_validity_markers.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/tests/test_bad_validity_markers.py
@@ -8,19 +8,6 @@ invalid_merge_marker = "Marge"  # codespell:ignore marge
 
 invalid_validity_marker_test_cases = (
     (
-        "too_many_valid_from_markers",
-        (
-            """
-            import pytest
-            @pytest.mark.valid_from("Paris")
-            @pytest.mark.valid_from("Paris")
-            def test_case(state_test):
-                assert 0
-            """,
-            "Too many 'valid_from' markers applied to test",
-        ),
-    ),
-    (
         "too_many_valid_until_markers",
         (
             """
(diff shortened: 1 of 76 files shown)


gurukamath (dismissed):

src/ethereum/forks/amsterdam/state_tracker.py
```diff
@@ -280,28 +280,18 @@ def account_exists(tx_state: TransactionState, address: Address) -> bool:
...
-    address :
-        Address of the account that needs to be checked.
-
-    Returns
-    -------
-    has_code_or_nonce : ``bool``
-        True if the account has non-zero nonce or non-empty code,
-        False otherwise.
-
+    Check if an account's code can be written to.
```
LouisTsai-Csie, line 285: I noticed the remaining functions (e.g., `account_has_storage`, `account_exists` and more ) in this file all have `Parameters` and `Returns` sections in their docstrings, but this one doesn't. Is that intentional? I see the same issue across all forks.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/tests/test_bad_validity_markers.py
```diff
@@ -7,19 +7,6 @@
...
-            import pytest
-            @pytest.mark.valid_from("Paris")
-            @pytest.mark.valid_from("Paris")
-            def test_case(state_test):
-                assert 0
-            """,
-            "Too many 'valid_from' markers applied to test",
-        ),
-    ),
     (
```
LouisTsai-Csie, line 23: The latest changes support multiple markers, so these two test cases are no longer needed. Could we add new tests for the updated behavior.

LouisTsai-Csie (approved):
I add two comments for non-blocking suggestion, thank you!

src/ethereum/forks/amsterdam/state_tracker.py
SamWilsn: We're slowly migrating away from the Parameters and Returns sections

packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/tests/test_bad_validity_markers.py
SamWilsn: Absolutely! Where would the best place to add one be?

src/ethereum/forks/amsterdam/state_tracker.py
LouisTsai-Csie: Okay! I see

packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/tests/test_bad_validity_markers.py
LouisTsai-Csie: What about adding in `test_param_level_validity_markers` function? We could add a new helper, something like `generate_param_level_valid_from_merge_test`?

PR #2906, opened 2026-05-23
feat(tooling, tests): detect and mark gas-checking tests

leolara:
## 🗒️ Description

Adds a pytest plugin that identifies tests asserting specific gas values, and applies a `@pytest.mark.gas_check` marker to the tests it finds.

The plugin (`gas_taint.py`) is enabled with `--detect-gas-checks` in fill mode only. It uses two detection modes:

- **`post[addr].storage[slot]`** — taint propagation. A `GasTainted(int)` carrier is returned by `Bytecode.gas_cost`, per-fork `gas_costs` / `opcode_gas_calculator` / `transaction_intrinsic_cost_calculator`, and propagated through arithmetic. `Number.__new__` and `FixedSizeHexNumber.__new__` are patched so taint survives `HashInt` / `HexNumber` construction.
- **Receipt / header / block-level / benchmark fields** — field-presence. The field name itself (`cumulative_gas_used`, `expected_gas_used`, `blob_gas_used`, …) is the assertion signal; if set, flag.

OOG-expecting tests (`tx.error`, `block_exception`, or `@pytest.mark.exception_test`) are excluded — their gas value lives on `tx.gas_limit`, an input, not a sink.

Worker results are aggregated to the master via `workeroutput` and `pytest_testnodedown`. Output is a JSON report; default path is `gas_check_report.json`.

`Block.expected_gas_used` and `BaseTest.expected_benchmark_gas_used` change from `int` to `HexNumber` so they pass through the patched `Number.__new__`. Both are internal assertion-only fields, not serialised to fixtures, so the type change has no fixture impact.

`scripts/mark_tests.py` is an idempotent script that applies a configurable `@pytest.mark.<name>` decorator to test functions listed in a JSON file. Three input shapes accepted: bare mapping, bare list of node IDs, or wrapped form with embedded marker name. Used here to apply `@pytest.mark.gas_check` to the 54 functions identified by a full-suite scan; re-running is a no-op.

### Result on the current suite

Full-suite run (`fill --detect-gas-checks --until=Amsterdam -m "not slow" -n 8 tests/`): 184,683 passed, 0 failures, 0 errors. 12,890 parametrized test cases flagged, collapsing to 54 unique functions across 31 files / 16 EIP directories.

| Sink kind | Hits | Detection |
|---|---|---|
| storage | 10,184 | taint |
| header.blob_gas_used | 5,784 | presence |
| receipt.cumulative_gas_used | 1,338 | presence |
| block_expected_gas_used | 68 | presence |

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/gas_taint.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/gas_taint.py
@@ -35,9 +35,30 @@ from __future__ import annotations
 import json
 from dataclasses import fields, replace
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Set, Tuple
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    List,
+    Set,
+    Tuple,
+)
 
 import pytest
+
+from execution_testing.specs.base import BaseTest
+from execution_testing.specs.benchmark import BenchmarkTest
+from execution_testing.specs.blockchain import Block, BlockchainTest
+from execution_testing.specs.state import StateTest
+from execution_testing.test_types.account_types import Alloc
+from execution_testing.test_types.transaction_types import Transaction
+from execution_testing.test_types.utils import Removable
+from execution_testing.vm.bases import OpcodeBase
+
+if TYPE_CHECKING:
+    from xdist.workermanage import WorkerController
 
 # ---------------------------------------------------------------------------
 # Taint carrier
@@ -168,7 +189,7 @@ def _wrap_intrinsic(c: type, fn: Callable, args: tuple, kwargs: dict) -> Any:
 def _wrap_opcode_calc(c: type, fn: Callable, args: tuple, kwargs: dict) -> Any:
     inner = fn(c, *args, **kwargs)
 
-    def calc(opcode: Any) -> Any:
+    def calc(opcode: OpcodeBase) -> int:
         result = inner(opcode)
         if isinstance(result, int) and not _is_tainted(result):
             name = getattr(opcode, "_name_", None) or str(opcode)
@@ -260,9 +281,11 @@ def uninstall_taint() -> None:
 
 
 def _record_tainted(
-    hits: List[dict], kind: str, location: str, value: Any
+    hits: List[dict], kind: str, location: str, value: int | None
 ) -> None:
     """Record only if value carries gas-source taint (used for storage)."""
+    if value is None:
+        return
     origins = _origins_of(value)
     if origins is None:
         return
@@ -277,12 +300,12 @@ def _record_tainted(
 
 
 def _record_present(
-    hits: List[dict], kind: str, location: str, value: Any
+    hits: List[dict], kind: str, location: str, value: int | None
 ) -> None:
     """Record if value is set (field-name-implies-gas-assertion sinks)."""
     if value is None:
         return
-    hit: Dict[str, Any] = {
+    hit: Dict[str, object] = {
         "kind": kind,
         "location": location,
         "value": int(value),
@@ -293,21 +316,14 @@ def _record_present(
     hits.append(hit)
 
 
-def _walk_storage(hits: List[dict], post: Any) -> None:
+def _walk_storage(hits: List[dict], post: Alloc | None) -> None:
     """Storage slots are general-purpose; only flag tainted values."""
     if post is None:
         return
-    try:
-        items = post.items()
-    except AttributeError:
-        return
-    for address, account in items:
+    for address, account in post.items():
         if account is None:
             continue
-        storage = getattr(account, "storage", None)
-        if storage is None:
-            continue
-        for slot, value in dict(storage.root).items():
+        for slot, value in account.storage.root.items():
             _record_tainted(
                 hits,
                 "storage",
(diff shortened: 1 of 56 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_gas_taint.py
```diff
@@ -0,0 +1,490 @@
...
+        storage=SimpleNamespace(root=storage_dict) if storage_dict else None
+    )
+
+
+def _make_post(accounts: dict) -> dict:
+    """Return a dict that quacks like ``Alloc`` (provides ``.items()``)."""
+    return accounts
+
+
+def _make_test(
```
marioevz: My main concern with the tests that use this function (and the mock that it returns) is that, e.g. if we change `post` from `BlockchainTest` or `StateTest`, this unit test will not detect it because `SimpleNamespace` will still have a `post` field in it – I know it's unlikely that we rename `post` but we can make this unit test resist these types of changes.

Can we use the type of unit testing used in `packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_benchmarking.py` or  `packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py`, where a mock pytest file is created and then the results analyzed?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/gas_taint.py
```diff
@@ -0,0 +1,489 @@
...
+    _record_present(
+        hits,
+        "block_expected_gas_used",
+        f"block[{i}].expected_gas_used",
+        getattr(block, "expected_gas_used", None),
+    )
+
+
+def _is_oog_test(test: Any, node: pytest.Item | None) -> bool:
+    tx = getattr(test, "tx", None)
```
marioevz: Is there any way we can infer the type here instead of trying to fetch attributes? It's a genuine question as it might be not possible and then simply making the unit tests more robust like I recommend in the other comment might be the way to go.

scripts/mark_tests.py
```diff
@@ -0,0 +1,366 @@
...
+       }
+
+   ``--marker`` on the command line overrides the value in the file.
+
+USAGE
+-----
+
+::
+
+    uv run python scripts/mark_tests.py REPORT.json --marker gas_check
```
marioevz, line 59: A couple of requests for this script:
- Make it available in the `just` command (so we can `just mark_gas_tests ./tests/amsterdam/...` and such)
- Allow a test file as input argument, so we can reduce the scope of which tests we want to analyze to a single file or function.

marioevz (commented):
Nice! A couple of comments.

One thing that I've noticed is that this requires the test to be already using `Bytecode.gas_cost`, which is ok, but another step could be that we also catch (perhaps with a separate script) tests that are sensible to gas changes and do not use this construct (think `static_tests`). It might be unfeasible but worth exploring in a future PR.

leolara:
@marioevz I have addressed your comments.

I was thinking abotu it more like a temporary fix, but this way is more useful for the long run.

 + I made the code typed
 + I made the tests with real object of the type that should be
 + I added pytester to test the plugin like in the benchmarks
 + I added the just command that you said

leolara:
Mario is working on having less tests with a gas_limit. The gas_limit triggers this marker, so I would do this after the reduction of tests depending on gas_limit is done.

leolara:
## Marker diff after rebasing onto `forks/amsterdam` (47 commits)

Re-ran `just mark_gas_tests` on the rebased tree. Net **+60 markers across 11 files**, captured in commit `01baba6` (`chore(tests): re-mark gas-checking tests after rebase`).

Unique test functions marked per directory:

| Directory | Old | New | Δ |
|---|--:|--:|--:|
| amsterdam/eip7778_block_gas_accounting_without_refunds | 5 | 5 | 0 |
| amsterdam/eip7954_increase_max_contract_size | 1 | 1 | 0 |
| amsterdam/eip7976_increase_calldata_floor_cost | 9 | 9 | 0 |
| amsterdam/eip7981_increase_access_list_cost | 1 | 1 | 0 |
| **amsterdam/eip8037_state_creation_gas_cost_increase** | 0 | 58 | **+58 (new EIP)** |
| berlin/eip2929_gas_cost_increases | 5 | 5 | 0 |
| berlin/eip2930_access_list | 2 | 2 | 0 |
| byzantium/eip196_ec_add_mul | 1 | 1 | 0 |
| byzantium/eip197_ec_pairing | 1 | 1 | 0 |
| cancun/eip1153_tstore | 2 | 2 | 0 |
| cancun/eip4844_blobs | 8 | 8 | 0 |
| cancun/eip5656_mcopy | 2 | 2 | 0 |
| **frontier/opcodes** | 6 | 5 | **−1** |
| osaka/eip7918_blob_reserve_price | 2 | 2 | 0 |
| osaka/eip7939_count_leading_zeros | 1 | 1 | 0 |
| **ported_static/stEIP2930** | 0 | 1 | **+1 (new)** |
| **ported_static/stPreCompiledContracts** | 0 | 1 | **+1 (new)** |
| prague/eip7623_increase_calldata_cost | 3 | 3 | 0 |
| prague/eip7702_set_code_tx | 4 | 4 | 0 |
| shanghai/eip3651_warm_coinbase | 1 | 1 | 0 |
| **TOTAL** | **54** | **114** | **+60** |

### Notes

- **Where the +60 came from**: almost entirely from the new EIP-8037 (state-creation gas cost increase) tests added upstream — 58 of the 60. Two more from a refresh of `ported_static`.
- **The single −1** in `frontier/opcodes`: one function we previously detected is no longer flagged. Its marker is still in the file (`mark_tests.py` only adds, never removes), so it's now stale — happy to remove it in a follow-up if reviewers prefer.
- **Parametrized entry counts** (entries in the JSON report, not unique functions) went from 12,890 → 13,188. Net of new EIP-8037 additions offset by reductions in `prague/eip7702_set_code_tx` (369 → 248), `prague/eip7623_increase_calldata_cost` (243 → 216), and `amsterdam/eip7778_block_gas_accounting_without_refunds` (68 → 48). Those reductions are the visible footprint of #2969 (implicit `tx.gas_limit`) — the test functions still assert gas, just with fewer parametrize cases.
- The detector deliberately ignores `tx.gas_limit`, so #2969's bulk cleanup of explicit-limit assignments doesn't move marker counts on its own; only when a parametrize axis got dropped did entry counts shift, and even then the unique functions per directory stayed the same.

leolara:
## Walker extension: `StateTest.blockchain_test_header_verify`

While resolving the rebase conflicts, I noticed `test_selfdestruct_state_gas_refilled_on_ancestor_revert` in `test_state_gas_selfdestruct.py` asserts a gas value via `state_test(..., blockchain_test_header_verify=Header(gas_used=expected_regular))`, and the walker was not detecting it. The field is defined on `StateTest` and gets copied verbatim onto the generated block's `header_verify` when the state test is promoted to a blockchain test, so it carries the same kind of gas assertion the walker already handles for `Block.header_verify`.

### Fix

Two commits on top of the rebase:

1. `d5aa3cf74 feat(tooling): walk StateTest.blockchain_test_header_verify too`
   - Extracts `_walk_header(hits, header, location_prefix)`, so both `Block.header_verify` (already covered) and `StateTest.blockchain_test_header_verify` go through the same recording path, including the `Removable` guard on `blob_gas_used`.
   - Adds a unit test that constructs a `StateTest` with `blockchain_test_header_verify=Header(gas_used=...)` and verifies the walker records the header hit.

2. `6772d439f chore(tests): mark tests using blockchain_test_header_verify`
   - Second re-mark pass with the extended walker. Nineteen additional tests now get the marker.

### Result

| | Before | After |
|---|--:|--:|
| Unique test functions flagged | 113 | 132 |
| Markers added in this pass | — | 19 |
| Already present | — | 113 |

Nineteen new markers land in nine files, mostly EIP-8037 state-gas tests that assert block-header gas via a state test:

- `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_create.py`: three tests
- `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_selfdestruct.py`: two tests, including `test_selfdestruct_state_gas_refilled_on_ancestor_revert`
- `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py`: three tests
- `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_sstore.py`: two SSTORE restoration tests
- `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_set_code.py`: four EIP-7702 auth tests
- `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py`: two block-level tests
- `tests/ported_static/stCreateTest/`: two collision tests
- `tests/ported_static/stEIP3607/`: one auth-account test

Sibling PR #2941 has been rebased on top of the new tip. Unit and end-to-end tests still pass, `just static` clean.

leolara:
@marioevz check again please

PR #2908, opened 2026-05-25
chore(specs): refactor block reward and fee refunds

SamWilsn:
## 🗒️ Description

- Refactors block reward and fee refunds to use `create_ether` instead of `set_account_balance`.
- Removes a redundant `destroy_account`

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/fork.py
+++ b/src/ethereum/forks/amsterdam/fork.py
@@ -70,7 +70,6 @@ from .requests import (
 from .state_tracker import (
     BlockState,
     TransactionState,
-    account_exists_and_is_empty,
     create_ether,
     destroy_account,
     extract_block_diff,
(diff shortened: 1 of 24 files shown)


src/ethereum/forks/amsterdam/fork.py
```diff
@@ -1089,11 +1079,6 @@ def process_transaction(
 
     all_logs = tx_output.logs + tuple(finalization_logs)
 
-    if coinbase_balance_after_mining_fee == 0 and account_exists_and_is_empty(
-        tx_state, block_env.coinbase
-    ):
-        destroy_account(tx_state, block_env.coinbase)
-
```
SamWilsn, line 1095: Would you mind double checking that this is correct to remove? I believe it's handled automatically in [`set_account_balance`](https://ethereum.github.io/execution-specs/src/ethereum/forks/amsterdam/state_tracker.py.html#ethereum.forks.amsterdam.state_tracker.set_account_balance:0) (via [`modify_state`](https://ethereum.github.io/execution-specs/src/ethereum/forks/amsterdam/state_tracker.py.html#ethereum.forks.amsterdam.state_tracker.modify_state:0)).

LouisTsai-Csie (commented):
I noticed that from `frontier` through `tangerine_whistle`, the miner fee transfer:

```python
# transfer miner fees
coinbase_balance_after_mining_fee = get_account(
    tx_state, block_env.coinbase
).balance + U256(transaction_fee)
set_account_balance(
    tx_state, block_env.coinbase, coinbase_balance_after_mining_fee
)
```

was not refactored to the shorter form used from `paris` through `amsterdam`:

```
# transfer miner fees
create_ether(tx_state, block_env.coinbase, U256(transaction_fee))
```

In `paris` through `amsterdam`, `modify_state` implements EIP-161 semantics and deletes the account when it becomes empty after the update. From `frontier` through `tangerine_whistle`, `modify_state` does not perform that cleanup because EIP-161 is not yet active, so replacing the two-line pattern with `create_ether` here would be a pure simplification with no behavioral change. Not sure if i am missing something

SamWilsn:
I think I just missed that while backporting!

src/ethereum/forks/amsterdam/fork.py
LouisTsai-Csie: That's reasonable, but I should clarify: account removal isn't triggered by `set_account_balance`. It's triggered by `create_ether`, which calls `modify_state` with the `increase_balance` hook.

LouisTsai-Csie (approved):
Looooks good 👍🏻

Issue #2910, opened 2026-05-26
Stateful filling refactors & missing feature tracker


LouisTsai-Csie:
Tracking suggested features and refactor for PR #2637 .

- [ ] In PR #2637, hive support is added, but the newly created environment doesn't include the stub accounts needed for stateful benchmarks. We could add these accounts to the genesis file so they're imported, like how state-actor (e.g., PR https://github.com/ethereum/state-actor/pull/76) handle the genesis file. This would let us verify the benchmark implementation's correctness (though not performance yet).
- [x] To support ether transfer benchmark introduced in PR #2837, we need to add extra 15K withdrawal starting with specific pk. => PR #2923 
- [x] There is no gas bump feature that could control that bump the block gas limit. We only need to support this in perf-devnet-3, as block gas limit in Jochemnet starts at 1G. We also need to investigate into the recent proposal that gives CL full control over the block gas limit, instead of EL. => PR #2923 
- [ ] Batch PRC call for `_fetch_receipts` helper in ClientBackend object, this was significantly blocking gas-benchmarks payload generation in the past, for `test_sstore_invariatns` only, it took ~2 hr to complete.
- [ ] There is currently a max batch transaction size, which is 750 txs. When benchmark generate transactions more than this amount (e.g., ether transfer, contract creation benchmark), the framework would build multiple block with max batch tx size, instead of building one single block.
- [ ] Post verification being ignored, but we still need such mechanism for benchmarking (e.g., total gas used comparison, final receipt check and more.).
- [x] Our current approach use `debug_setHead` at the end of the benchmark, so that the next benchmark begins at `start_block` (`snapshot_block` + some global initialization). Based on Jochem's comment, we need to investigate whether it works as expected. In gas-benchmarks, we use FCU to control the rollback. => PR #2923 
- [ ] Add necessary metadata information for fixture, including the snapshot block, gas bump block number, sender pool size and more.
- [ ] Handle opcode counting in stateful filler, make benchmarkoor able to consume the opcode count file (related benchmarkoor PR https://github.com/ethpandaops/benchmarkoor/pull/182#issuecomment-4554991482)

PR #2923, opened 2026-05-27
feat(test-fill): align `fill-stateful` with `gas-benchmarks` implementation

LouisTsai-Csie:
## 🗒️ Description

Add a few features & refactors to the stateful filling implementation.

### Commit [d198221](https://github.com/ethereum/execution-specs/pull/2923/commits/d19822166524b988556274907a73055aea93bc69): Replace `debug_setHead` with FCU approach.

`debug_setHead` only supports in Geth, Besu and Erigon. Nethermind implements `debug_resetHead` that takes block hash as input. Reth has the method registered but not implemented ([Link](https://github.com/paradigmxyz/reth/blob/fd96cb2bd65023a0b8cc1ee328f325e64f65ebb1/crates/rpc/rpc/src/debug.rs#L1112)), Nimbus comments out the method ([Link](https://github.com/status-im/nimbus-eth1/blob/375f43fd6336feaf706bdf2c845365645356b8d2/execution_chain/rpc/debug.nim#L158)), no implementation either.

Consider some features (e.g., built-in opcode tracing) is now only supported in Nethermind master branch, this PR changes the `debug_setHead` to FCU version, in order to set the canonical chain.

### Commit [0252bec](https://github.com/ethereum/execution-specs/pull/2923/commits/0252bec93f69e85d35db564bc7493f28b2948268): Add a new CLI flag `--gas-bump-blocks`.

For perf-devnet-3 snapshot, we need to bump the block gas limit to ~1T block gas limit. This is done by sending empty blocks in gas-benchmarks. User could use `--gas-bump-blocks` to assign number of empty block to increase the block gas limit. 

In gas-benchmarks, we send 5000 empty blocks to bump the block gas limit for perf-devnet-3. For Jochmnet, we do not need to bump the gas limit, as the block gas limit is already high enough (~1 gigagas).

### Commit [05f3c16](https://github.com/ethereum/execution-specs/pull/2923/commits/05f3c16b2af478e5a87701aa3a732efb9d15f2d4): Add deterministic sender pooling.

For `test_ether_transfers_onchain_receivers` benchmark, to prevent the sender account being cached, we pre-fund the account via CL withdrawal. We inject 15K withdrawal into funding phase, starting with this pk range:

```python
SENDER_BASE_KEY = int.from_bytes(
    keccak256(b"gas-repricings-private-key"), "big"
)
```

### Commit [54f43ad](https://github.com/ethereum/execution-specs/pull/2923/commits/54f43ad277880bf82f0e9c506249df2b6ef5534c): Batch tx receipt request

Batch tx receipt RPC request, to avoid large amount of RPC request at the same time. This was a bottleneck in the past when using gas-benchmarks.

### Commit [8477bfe](https://github.com/ethereum/execution-specs/pull/2923/commits/8477bfe48b8eaeebd7727bf0efc5d3626ce40d2e): refactor `fill_stateful.py` 

Split the hive setup version from the original `fill_stateful.py` file, to keep the file small and focus on the filler feature.

These senders are then consumed by `test_ether_transfers_onchain_receivers`, this aligns changes in gas-benchmarks PR https://github.com/NethermindEth/gas-benchmarks/pull/146

### Verification
Verify the test via Hive mode, works successfully.

## 🔗 Related Issues or PRs

Issue https://github.com/ethereum/execution-specs/issues/2910

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
@@ -337,27 +337,27 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
 
     def bump_block_gas_limit(
         self,
-        block_count: int,
+        target_gas_limit: int,
     ) -> List[EnginePayloadMetadata]:
         """
         Build empty block to increase the block gas limit to target.
         """
-        if block_count <= 0:
-            return []
         assert self.testing_rpc is not None, (
             "bump_block_gas_limit requires testing_rpc"
         )
         captured: List[EnginePayloadMetadata] = []
         with self.block_building_lock:
-            for _ in range(block_count):
-                head_block = self.get_block_by_number("latest")
-                assert head_block is not None
-                next_timestamp = int(HexNumber(head_block["timestamp"]) + 1)
+            head_block = self.get_block_by_number("latest")
+            assert head_block is not None
+            parent_hash = Hash(head_block["hash"])
+            parent_timestamp = int(HexNumber(head_block["timestamp"]))
+            gas_limit = int(HexNumber(head_block["gasLimit"]))
+            while gas_limit < target_gas_limit:
                 payload_attributes = self._payload_attributes(
-                    next_timestamp=next_timestamp,
+                    next_timestamp=parent_timestamp + 1,
                 )
                 new_payload = self.testing_rpc.build_block(
-                    parent_block_hash=Hash(head_block["hash"]),
+                    parent_block_hash=parent_hash,
                     payload_attributes=payload_attributes,
                     transactions=[],
                     extra_data=Bytes(b""),
@@ -368,6 +368,14 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
                         payload_attributes.parent_beacon_block_root,
                     )
                 )
+                new_gas_limit = int(new_payload.execution_payload.gas_limit)
+                assert new_gas_limit > gas_limit, (
+                    f"block gas limit stalled at {new_gas_limit} before "
+                    f"reaching target {target_gas_limit};"
+                )
+                gas_limit = new_gas_limit
+                parent_hash = new_payload.execution_payload.block_hash
+                parent_timestamp = int(new_payload.execution_payload.timestamp)
         return captured
 
     def set_canonical_head(self, head_block_hash: Hash) -> None:
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
@@ -19,6 +19,7 @@ from execution_testing.base_types import (
 )
 from execution_testing.client_clis.cli_types import EnginePayloadMetadata
 from execution_testing.forks import Fork, TransitionFork
+from execution_testing.forks.transition_base_fork import TransitionBaseClass
 from execution_testing.rpc import EngineRPC, TestingRPC
 from execution_testing.rpc import EthRPC as BaseEthRPC
 from execution_testing.rpc.rpc_types import (
@@ -38,7 +39,10 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
     pending transactions or a block generation interval.
     """
 
-    fork: Fork | TransitionFork
+    fork: Fork
+    fcu_version: int
+    new_payload_version: int
+    get_payload_version: int
     engine_rpc: EngineRPC
     get_payload_wait_time: float
     block_building_lock: FileLock
@@ -63,7 +67,25 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
             transaction_wait_timeout=transaction_wait_timeout,
             max_transactions_per_batch=max_transactions_per_batch,
         )
+        assert not issubclass(fork, TransitionBaseClass), (
+            "ChainBuilderEthRPC does not support transition forks"
+        )
         self.fork = fork
+        fcu_version = fork.engine_forkchoice_updated_version()
+        assert fcu_version is not None, (
+            "Fork does not support engine forkchoice_updated"
+        )
+        self.fcu_version = fcu_version
+        new_payload_version = fork.engine_new_payload_version()
+        assert new_payload_version is not None, (
+            "Fork does not support engine new_payload"
+        )
+        self.new_payload_version = new_payload_version
+        get_payload_version = fork.engine_get_payload_version()
+        assert get_payload_version is not None, (
+            "Fork does not support engine get_payload"
+        )
+        self.get_payload_version = get_payload_version
         self.engine_rpc = engine_rpc
         parsed = urlparse(rpc_endpoint)
         self.block_building_lock = FileLock(
@@ -87,26 +109,15 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
                 # Get the head block hash
                 head_block = self.get_block_by_number("latest")
                 assert head_block is not None
-                block_number = HexNumber(head_block["number"])
-                timestamp = HexNumber(head_block["timestamp"])
-                head_fork = self.fork.fork_at(
-                    block_number=block_number, timestamp=timestamp
-                )
                 # Send initial forkchoice updated
                 forkchoice_state = ForkchoiceState(
                     head_block_hash=head_block["hash"],
-                )
-                forkchoice_version = (
-                    head_fork.engine_forkchoice_updated_version()
-                )
-                assert forkchoice_version is not None, (
-                    "Fork does not support engine forkchoice_updated"
                 )
                 for _ in range(initial_forkchoice_update_retries):
                     response = self.engine_rpc.forkchoice_updated(
                         forkchoice_state,
                         None,
-                        version=forkchoice_version,
+                        version=self.fcu_version,
                     )
                     if (
                         response.payload_status.status
@@ -137,9 +148,8 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
         withdrawals: List[Withdrawal] | None = None,
     ) -> PayloadAttributes:
         """Build payload attributes for a block at ``next_timestamp``."""
-        next_fork = self.fork.fork_at(block_number=0, timestamp=next_timestamp)
         return PayloadAttributes.for_fork(
-            next_fork,
+            self.fork,
             timestamp=next_timestamp,
             withdrawals=withdrawals,
         )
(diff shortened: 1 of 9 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
```diff
@@ -347,3 +347,65 @@ def fund_via_withdrawals(
...
+
+    def set_canonical_head(self, head_block_hash: Hash) -> None:
+        """
+        Reorg the canonical head to head_block_hash via
+        ``engine_forkchoiceUpdated``.
+        """
+        head_block = self.get_block_by_hash(head_block_hash)
+        assert head_block is not None, (
+            f"cannot reset head to unknown block {head_block_hash}"
+        )
```
marioevz: Maybe this is unnecessary, since the client itself will make the same complaint if it cannot find the block.
It's just one less RPC call to make IMO.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
```diff
@@ -347,3 +347,65 @@ def fund_via_withdrawals(
...
+        head_fork = self.fork.fork_at(
+            block_number=HexNumber(head_block["number"]),
+            timestamp=HexNumber(head_block["timestamp"]),
+        )
```
marioevz: I see now the reason behind the previous call, but perhaps the solution is that we should never expect `TransitionFork` here and assume that `ChainBuilderEthRPC` is always going to be initialized with `Fork`, otherwise error out.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
```diff
@@ -347,3 +347,65 @@ def fund_via_withdrawals(
...
+        fcu_version = head_fork.engine_forkchoice_updated_version()
+        assert fcu_version is not None, (
+            "Fork does not support engine forkchoice_updated"
+        )
```
marioevz: We could do this once during initialization of `ChainBuilderEthRPC`, save `fcu_version` as a field of the chain builder instance, and then just use it as `self.fcu_version`.
This is also under the assumption that `fork` is always `Fork` and never `TransitionFork`.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -693,35 +557,32 @@ def t8n(
 @pytest.fixture(autouse=True, scope="function")
 def _reset_chain_between_tests(
```
marioevz, line 562: Maybe out of scope for this PR, but the way this fixture is auto-used feels off.
I think a better structure would be to have a `session_client_backend` fixture that is session-scoped, and a `client_backend` fixture that is function-scoped.

The `session_client_backend` fixture initializes the object for the entire session, most of what `_session_pre_run` and `client_backend` fixtures do now.

The new `client_backend` fixture prepares and cleans up the `ClientBackend` object before and after every test, so what `_reset_chain_between_tests` currently does.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -693,35 +557,32 @@ def t8n(
...
-    start_hex = client_backend.start_block["number"]
     expected_hash = client_backend.start_block["hash"]
-    # Skip when head already at start (geth rejects same-block setHead).
+
     current_head = eth_rpc.get_block_by_number("latest")
     if current_head is not None and current_head["hash"] == expected_hash:
         return
     try:
-        debug_rpc.set_head(start_hex)
+        eth_rpc.set_canonical_head(Hash(expected_hash))
```
marioevz, line 579: This call feels like we are conflating two different objects to achieve a purpose.

We could have `client_backend.reset_to_start_block` or similar because `client_backend` already knows  `start_block` so it can do everything by itself.

This also makes the code more maintainable because if we ever modify the `client_backend` structure, you only have to modify methods inside of the self-contained `ClientBackend` class.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -117,6 +114,29 @@ def pytest_addoption(parser: pytest.Parser) -> None:
...
+        action="store",
+        dest="gas_bump_blocks",
+        default=5000,
+        type=int,
+        help=(
+            "Empty blocks at start to increase block gas limit."
+            "Each block increase ≤ parent_gas_limit/1024."
+            "Default: 5000 blocks."
+        ),
+    )
```
marioevz: This is the right idea but the wrong abstraction.

A user normally only knows the gas limit their test needs, not how many empty blocks produce it. With this parameter they still have to (a) find the starting block gas limit of the snapshot, and (b) compute the number of blocks needed to ramp up to their target. That's our math to do, not theirs.

Two suggestions:

- For the MVP, make the parameter `--block-gas-limit` (the target), and internally derive how many empty blocks are needed to reach it from the start block's gas limit. The `≤ parent_gas_limit/1024` per-block constraint is exactly the kind of detail we should be hiding behind this.
- Longer term, and probably out of scope here: even that parameter may be unnecessary. We already have a default `Environment` with a default `gas_limit`, so we could just target that. A test that deviates inherently needs a different setup, and therefore a different number of gas-bump-blocks anyway.

tests/benchmark/stateful/bloatnet/test_transaction_types.py
```diff
@@ -16,13 +16,14 @@
...
     compute_create_address,
+    keccak256,
 )
 
 # Deterministic sender pool of 15K accounts.
 # Funded via system contract withdrawals (funding.txt) in payload generation.
 # Placed outside pre-allocation to ensure accounts remain uncached.
-SENDER_BASE_KEY = (
-    0x1111111111111111111111111111111111111111111111111111111111111111
+SENDER_BASE_KEY = int.from_bytes(
```
marioevz: My main concern here is that this sidesteps the standard `pre.fund_eoa` path. That abstraction exists so we can change how funding works under the hood without touching tests — and this is the one test that opts out, which means every framework change now has to account for this special case separately. Case in point: This PR adds a whole new parameter just to deal with this (`--sender-pool-size`).

It also makes the test order-dependent: yield_distinct_sender() hands out senders based on execution order, so the accounts a given test gets aren't stable.

I'd rather we solve the underlying thing generally. If the goal is funding via withdrawals to keep accounts uncached, that's something `pre.fund_eoa` could do for all accounts — at which point this becomes automatic and the special case (and the new parameter) goes away.

marioevz (changes requested):
Thanks for doing this, I added a couple of comments which I feel need to be resolved before merging.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
marioevz: See example of `session_t8n` and `t8n` session and function scoped fixtures that use the same object internally: https://github.com/ethereum/execution-specs/blob/7771d278b7aa754e9a26368554cd01ebc479d6c8/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py#L1093

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
LouisTsai-Csie: Removed the lookup, an unknown head now surfaces as a non-VALID FCU status instead.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
LouisTsai-Csie: I review who constructs `ChainBuilderEthRPC` instance and found:

- `execute-hive` ([Link](https://github.com/ethereum/execution-specs/blob/26ff8a649e048cb3d77c24c9646dde491aab3108/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/hive.py#L447)) and `execute-remote` ([Link](https://github.com/ethereum/execution-specs/blob/26ff8a649e048cb3d77c24c9646dde491aab3108/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/remote.py#L199)): both set `skip_transition_forks` to `True` ([Link](https://github.com/ethereum/execution-specs/blob/26ff8a649e048cb3d77c24c9646dde491aab3108/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py#L110))
- `fill-stateful`: also sets `skip_transition_forks` to `True`. ([Link](https://github.com/ethereum/execution-specs/blob/26ff8a649e048cb3d77c24c9646dde491aab3108/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py#L263))
- E2E tests: hardcode Osaka
- Consumer simulators: unaffected (they share ruleset module but don't use `ChainBuilderEthRPC`)

Changes Made: Refactored all calls to `get_fcu_version`, `new_payload_version`, and `get_payload_version` to be retrieved at initialization (`__init__`) time

Breaking Change: Previously, for execute/fill-stateful, explicit fork specs like `--fork BPO2ToAmsterdamAtTime15k` worked because the parser accepts transition names and a single fork bypasses `skip_transition_forks`. Now this fails fast at session start with "ChainBuilderEthRPC does not support transition forks" instead of proceeding with errors.

Added in commit [c5882a8](https://github.com/ethereum/execution-specs/pull/2923/commits/c5882a833f7d62fc05ca3e17e5befbb42e0eeadf)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
LouisTsai-Csie: Same as [comment](https://github.com/ethereum/execution-specs/pull/2923/changes/BASE..8477bfe48b8eaeebd7727bf0efc5d3626ce40d2e#r3394590112)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
LouisTsai-Csie: I added the `--block-gas-limit` flag to replace the old flag. If not provided, it defaults to `env.gas_limit`. For consensus tests, the default is 2× mainnet block limit; for benchmarks, it would be 1 gigagas.

Added in commit [5947fd4](https://github.com/ethereum/execution-specs/pull/2923/commits/5947fd4d30aded0828cedc028540e1cc250814ae)

tests/benchmark/stateful/bloatnet/test_transaction_types.py
LouisTsai-Csie: Not all pre-allocated accounts need to be uncached, only specific benchmarks do. I added `pre.uncached_eoa()` to hand out EOAs from a deterministic pool.

Configuration is now centralized in one place: `UNCACHED_EOA_BASE_KEY` and `UNCACHED_EOA_POOL_SIZE`. Tests no longer define `SENDER_BASE_KEY` or use `--sender-pool-size`.
These accounts stay out of pre-allocation (anything funded during setup is warm, not what we want in the benchmark). 

Allocation resets per test, and `forkchoiceUpdate` restores nonces/balances, so each test gets identical account state regardless of order.

Added in commit [2074875](https://github.com/ethereum/execution-specs/pull/2923/commits/20748752fe18be5edeb687ef280c05db4ff19066)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
LouisTsai-Csie: Could I combine this refactor with the [comment](https://github.com/ethereum/execution-specs/pull/2923#discussion_r3350620703) in a subsequent PR? They both touch `ClientBackend`, so a single PR would be cleaner, and the current PR already bundles lots of changes from different aspects.

PR #2932, opened 2026-05-28
feat(tests): EIP-7778 admission gate uses pre-refund gas

danceratopz:
## 🗒️ Description

Add `test_extra_tx_admission_uses_pre_refund_gas`: a 2-tx regression covering the case where a refund-bearing first tx is followed by a tx whose `gas_limit` exceeds its actual usage. The existing `test_multi_transaction_gas_accounting` parametrisations pin `extra_tx.gas_limit` to the intrinsic cost, so a post-refund admission gate is masked by the subsequent `header.gas_used > gas_limit` check. The added slack makes the bypass observable: a buggy implementation admits the trailing tx and computes a `header.gas_used` that no longer matches the fixture.

## 🔗 Related Issues or PRs
This coverage gap was discovered by:
- https://github.com/NethermindEth/nethermind/pull/11794.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
+++ b/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
@@ -558,28 +558,13 @@ def test_extra_tx_admission_uses_pre_refund_gas(
     refund_tx_reverts: bool,
 ) -> None:
     """
-    Test the EIP-7778 admission gate against the pre-refund accumulator
-    when the trailing tx has gas_limit above its actual gas usage.
+    Test that the admission gate uses the pre-refund accumulator when
+    the trailing tx's gas_limit exceeds its actual usage.
 
-    Distinguishes a correct implementation from one that uses the
-    post-refund (receipt) accumulator at the admission gate but the
-    pre-refund accumulator for ``header.gas_used``. Without slack
-    between ``tx.gas_limit`` and the tx's actual gas usage, both
-    implementations end up rejecting the block (the buggy one via the
-    later ``header.gas_used > gas_limit`` check), so the bug is masked.
-    The slack here makes the buggy implementation accept the block,
-    which then mismatches the expected-invalid fixture.
-
-    See https://github.com/NethermindEth/nethermind/pull/11794.
-
-    Layout (``P`` = ``refund_tx`` pre-refund gas; ``R`` = refund applied;
-    ``G`` = ``extra_tx.gas_limit``; ``U`` = ``extra_tx`` actual usage):
-
-    - ``block.gas_limit = P + G - 1``
-    - Correct admission: ``P + G > L``  -> reject extra tx
-    - Buggy admission:   ``(P - R) + G <= L``  -> admit extra tx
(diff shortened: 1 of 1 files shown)


danceratopz:
This a tricky one, so adding Claude's explanation for dummies :laughing: 

## Setup

**Two transactions in one block.**

### Tx 1 — the "refund tx"

Hits a contract that performs 10 `SSTORE` clears (slots from `1` → `0`).
Two things result:

- It *charges* roughly **P = 71,050 gas** (the "pre-refund" amount — the
  work actually done).
- It *earns* a refund of **R** gas (~12,000) because clearing storage
  gets a partial rebate.

The receipt for this tx shows `P − R` gas used (what the sender pays
for). But the **block** counts the full `P`, because EIP-7778 says
block-level accounting ignores refunds.

After tx 1 the block has two "running totals":

- Pre-refund running total: `P` ← this is what `header.gas_used` will be.
- Post-refund running total: `P − R` ← this is what receipts add up to.

### Tx 2 — the "trailing tx"

Just calls a contract that runs `STOP`. Almost nothing happens — its
actual gas use is the intrinsic **21,000**.

The trick: we give tx 2 a `gas_limit` of **42,000** (twice what it'll
actually use). That extra headroom is the whole point.

### Block

`block.gas_limit = P + 42,000 − 1 = 113,049`. Just one gas short of
fitting tx 2's *declared* gas limit on top of tx 1's pre-refund cost.

## What the two clients do

| Step | Correct client (pre-refund gate) | Buggy client (post-refund gate) |
|---|---|---|
| At tx 2's admission check, what does "used so far" mean? | `P` (= 71,050) | `P − R` (= ~59,050) |
| Available room = `gas_limit − used so far` | `41,999` | `~53,999` |
| Tx 2's declared gas (42,000) vs available | `42,000 > 41,999` → **reject** | `42,000 ≤ 53,999` → **admit** |
| What happens next? | Block invalid — bad tx in it. **Done.** | Executes tx 2. It uses only 21,000, so block's pre-refund total becomes `92,050`. |
| End-of-block: compare computed `header.gas_used` to what's in the block header (`71,050`). | (Never reached.) | `92,050 ≠ 71,050` → **block invalid, but for the wrong reason.** |

## The point

Both clients reject the block — but **for different reasons**, at
different points in the pipeline.

- Correct: "tx 2 doesn't fit, reject at admission."
- Buggy: "I admitted tx 2 (oops), and now my totals don't match the
  header at the end."

The `got 92,050, expected 71,050` mismatch the buggy binary prints is
the smoking gun: that extra 21,000 is exactly tx 2's STOP, which proves
it was admitted when it shouldn't have been.

The whole trick of the test is **the slack on tx 2's gas_limit**.
Without it, the buggy code's "wrong" admission still ends up overflowing
the block gas limit at the end, which masks the bug as a
different-but-similar error. With the slack, the buggy code admits *and*
stays under the block limit, so the only thing that catches it is the
header-vs-computed-gas-used comparison — a distinct, distinguishable
error.

svlachakis (approved):

spencer-tb (approved):
LGTM! Made some small tweaks... will add for the next devnet release :)

PR #2934, opened 2026-05-28
chore(specs): add final decorator

SamWilsn:
## 🗒️ Description

mypyc can use the `@final` decorator for optimizations, so might as well add it.

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/crypto/blake2.py
+++ b/src/ethereum/crypto/blake2.py
@@ -2,7 +2,7 @@
 
 import struct
 from dataclasses import dataclass
-from typing import Final, List, Tuple
+from typing import Final, List, Tuple, final
 
 from ethereum_types.numeric import Uint
 
@@ -248,6 +248,7 @@ class Blake2:
 
 
 # Parameters specific to the Blake2b implementation
+@final
 @dataclass
 class Blake2b(Blake2):
     """
--- a/src/ethereum/forks/amsterdam/block_access_lists.py
+++ b/src/ethereum/forks/amsterdam/block_access_lists.py
@@ -13,7 +13,7 @@ See [`BlockAccessList`][bal] for more detail.
 """
 
 from dataclasses import dataclass, field
-from typing import Dict, List, Optional, Set, Tuple, TypeAlias
+from typing import Dict, List, Optional, Set, Tuple, TypeAlias, final
 
 from ethereum_rlp import rlp
 from ethereum_types.bytes import Bytes, Bytes32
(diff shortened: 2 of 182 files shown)


danceratopz (approved):
LGTM! I asked Claude if any classes were accidentally left out, but we're good to go. 

I'm interested what you got cooking with `mypyc` and what the results are!

I added Claude's summary below on what which classes get the decorator and which classes are omitted (and why) as documentation.

## What got `@final`

491 classes, all leaf `@slotted_freezable @dataclass` (or plain `@dataclass`) types, across all 24 fork directories plus 5 shared modules (`crypto/blake2.py`, `genesis.py`, `merkle_patricia_trie.py`, `state.py`, `trace.py`). The decorator is placed outermost (above `@slotted_freezable` / `@dataclass`), which is correct: `slotted_freezable` returns a freshly built class object, and since decorators apply bottom-up, the outermost `@final` marks that final object.

These data types are the ones that benefit: `@dataclass` generates `__init__`, `__eq__`, and `__hash__`, which are real calls in the codebase (equality, dict/set keys), and `@final` lets mypyc bypass the vtable and emit direct native calls for them.

## What was omitted, and why it is correct

mypyc's `@final` optimization only bypasses the vtable for method calls and property accessors. A class with no methods or properties has nothing to devirtualize, so the following omissions cost nothing:

- Leaf exception classes (`OutOfGasError`, `StackUnderflowError`, `TransactionTypeError`, etc.): `pass` bodies, no methods or properties, and never method-called in hot paths.
- KZG scalar types in `crypto/kzg.py` (`KZGCommitment`, `KZGProof`, `BLSFieldElement`, `VersionedHash`): also `pass` bodies, nothing to optimize.
- `ForkCriteria` subclasses in `fork_criteria.py` (`ByBlockNumber`, `ByTimestamp`, `Unscheduled`): these do have methods (`check`, `__repr__`), but they are invoked through the `ForkCriteria` base type (so devirtualization would not fire at the call sites) and only at fork-scheduling time, not a hot path.

These are also correct on the merits regardless of mypyc:

- `GasCosts`: a plain namespace class carrying the comment "may be patched at runtime by a future gas repricing utility", so it is intentionally left mutable.
- `Ops` and other `Enum` subclasses: enums with members cannot be subclassed anyway, so `@final` would be redundant.
- `PreState` and `EvmTracer`: `typing.Protocol` subclasses, where `@final` would be semantically wrong since protocols are meant to be implemented.

## Sources

- mypyc devirtualizes method and property calls on final classes: [python/mypy#17886](https://github.com/python/mypy/pull/17886).
- mypyc native classes overview: https://mypyc.readthedocs.io/en/latest/native_classes.html.
- `typing.final` semantics: https://mypy.readthedocs.io/en/stable/final_attrs.html.

danceratopz:
Hey @SamWilsn, I added a follow-up here:
- https://github.com/ethereum/execution-specs/pull/2965

PR #2945, opened 2026-06-01
✨ feat(tests): Disallow empty change set for storage slot

raxhvl:
## 🗒️ Description

Adds a test to ensure clients [reject storage slots with empty changes](https://github.com/ethereum/EIPs/pull/11750)

    ```console
    just static
    ```


packages/testing/src/execution_testing/test_types/block_access_list/modifiers.py
```diff
@@ -481,6 +481,41 @@ def transform(bal: BlockAccessList) -> BlockAccessList:
...
+
+        found_address = False
+        new_root = []
+        for account_change in bal.root:
+            if account_change.address == address:
+                found_address = True
+                new_account = account_change.model_copy(deep=True)
+                new_account.storage_changes.append(
+                    BalStorageSlot(slot=slot, slot_changes=[])
+                )
```
spencer-tb, line 504: Without this, I think slot 0 is in both `storage_reads` and `storage_changes`, so the block becomes invalid for a diff rule?
```suggestion
                new_account.storage_changes.append(
                    BalStorageSlot(slot=slot, slot_changes=[])
                )
                new_account.storage_reads = [
                    s for s in new_account.storage_reads if int(s) != slot
                ]
```

spencer-tb (approved):
LGTM! Thanks :)

One small change if it makes sense!

packages/testing/src/execution_testing/test_types/block_access_list/modifiers.py
raxhvl: > Without this, I think slot 0 is in both storage_reads and storage_changes, so the block becomes invalid for a diff rule?

This is intentional.  Case `demoted_noop` MUST record `storage_reads=[0x00]` since we are overwriting the same value. This is a valid entry and it MUST not be touched by the modifier.

The modifier appends a *stray* empty set to a valid BAL. In this particular case returns:

```py
storage_reads=[0x00] //valid change
storage_changes=[(0x00, [])]  // extra empty slot
```

The goal of this test is to check if clients clean up storage changes during a no-op write and correctly detect empty change set. 

I also understand that clients may reject the BAL because the storage key is repeated. Since our tests dont pin error codes we cant differentiate  "rejected because empty change" vs "rejected because key-in-both"  or say "rejected because hash mismatch." 

So how we design our exception tests is *subjective* and i lean towards *isolating* a corruption.

cc: @marioevz - im okay what you both decide.

spencer-tb: Thanks! Appreciate the reply here! Sgtm will merge :)

PR #2946, opened 2026-06-02
feat(tests): EIP-7928 catch per-tx no-op SSTORE after prior tx write

pk910:
## 🗒️ Description
Adds `test_bal_intra_tx_round_trip_after_prior_tx_write` to verify that a per-tx no-op SSTORE is classified as a read even when an earlier transaction in the same block already wrote to the slot.

Two txs call a contract that does
`SSTORE(slot=1, 0xff); SSTORE(slot=1, 0x42); STOP`. Tx 1 changes slot 1 from 0 to 0x42 (real change). Tx 2 round-trips 0x42 -> 0xff -> 0x42 (per-tx no-op). Per EIP-7928 §Storage, "MUST check the pre-transaction value" — and pre-transaction is per-tx, not per-block — so only tx 1 must appear in storage_changes.

Regression test for the bal-devnet-7 chain split (slot 96338 / EL block 94626). Reproduced live on a 2-node nethermind + erigon kurtosis devnet running the bal-devnet-7 images: erigon rejects with `block access list mismatch` and forks off.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

        /\_/\
      =( o.o )=
        > ^ <

### DIFF
--- a/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_cross_index.py
+++ b/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_cross_index.py
@@ -268,6 +268,79 @@ def test_bal_noop_write_filtering(
     )
 
 
+def test_bal_intra_tx_round_trip_after_prior_tx_write(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+) -> None:
+    """
+    A per-tx no-op SSTORE sequence is demoted to a read even when an
+    earlier tx in the same block already wrote to the slot.
+
+    Per EIP-7928 §Storage, "Implementations MUST check the
+    pre-transaction value to correctly distinguish between actual writes
+    and no-op writes." Pre-transaction is per-tx, not per-block, so a
+    prior tx's write to the slot does not change tx 2's classification.
+
+    Both txs call the same contract whose runtime SSTOREs 0xff then 0x42
+    to slot 1. Tx 1 changes slot 1 from 0 to 0x42 (real change). Tx 2
+    round-trips it 0x42 -> 0xff -> 0x42 (net-zero, per-tx no-op). Only
+    tx 1 must appear in `storage_changes`.
+    """
+    # Runtime: write 0xff to slot 1, then write 0x42 to slot 1, STOP.
+    # The two SSTOREs hit the journal at every call, but the net effect
+    # on the slot is `pre_value -> 0x42` — a no-op when pre_value == 0x42.
+    contract_code = Bytecode(Op.SSTORE(1, 0xFF) + Op.SSTORE(1, 0x42) + Op.STOP)
+    contract = pre.deploy_contract(code=contract_code)
+
+    sender_a = pre.fund_eoa()
+    sender_b = pre.fund_eoa()
+
+    # Both txs go into the same block; tx 1 makes the real 0 -> 0x42
+    # change, tx 2 starts from 0x42 and ends at 0x42 (per-tx no-op).
+    tx_1 = Transaction(sender=sender_a, to=contract, gas_limit=200_000)
+    tx_2 = Transaction(sender=sender_b, to=contract, gas_limit=200_000)
+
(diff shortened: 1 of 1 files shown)


tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_cross_index.py
danceratopz, line 18:
```suggestion
    BalAccountAbsentValues,
    BalAccountExpectation,
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_cross_index.py
```diff
@@ -268,6 +268,79 @@ def test_bal_noop_write_filtering(
...
+                storage_changes=[
+                    BalStorageSlot(
+                        slot=1,
+                        slot_changes=[
+                            BalStorageChange(
+                                block_access_index=1, post_value=0x42
+                            ),
+                        ],
+                    ),
+                ],
```
danceratopz, line 320: Not essential, this addition does not change the fixture output. It's just a defensive guard when filling against a buggy reference spec implementation.
```suggestion
                ],
                # `storage_changes` is only verified as a sub-sequence
                # at fill time, so this additional check guards against a
                # reference regression that emits tx 2's no-op as a
                # spurious index-2 change.
                absent_values=BalAccountAbsentValues(
                    storage_changes=[
                        BalStorageSlot(
                            slot=1,
                            slot_changes=[
                                BalStorageChange(
                                    block_access_index=2, post_value=0x42
                                ),
                            ],
                        ),
                    ],
                ),
```

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_cross_index.py
```diff
@@ -268,6 +268,79 @@ def test_bal_noop_write_filtering(
...
+    Per EIP-7928 §Storage, "Implementations MUST check the
+    pre-transaction value to correctly distinguish between actual writes
+    and no-op writes." Pre-transaction is per-tx, not per-block, so a
+    prior tx's write to the slot does not change tx 2's classification.
+
+    Both txs call the same contract whose runtime SSTOREs 0xff then 0x42
+    to slot 1. Tx 1 changes slot 1 from 0 to 0x42 (real change). Tx 2
+    round-trips it 0x42 -> 0xff -> 0x42 (net-zero, per-tx no-op). Only
+    tx 1 must appear in `storage_changes`.
+    """
```
danceratopz, line 291: Claude nit'd with this docstring as it quotes a stale version of the spec that was updated here in https://github.com/ethereum/EIPs/commit/1505de07519930e6af4f97439d8042ae37858d50#diff-b4ae44357044a8606fc10a6d8433e1ee2135b1111042c88876bd58d6f11c1d6dL103-L220 😆 So sorry for the nit, but I think this reads truer to the spec/behavior.
```suggestion
    """
    Verify a per-tx no-op SSTORE round-trip is not recorded as a storage
    change when an earlier tx in the same block wrote the slot.

    Per EIP-7928 §Storage, a write is compared against "the storage value
    as of immediately before the current `block_access_index` (i.e., the
    cumulative state from all prior indices, falling back to the pre-block
    state)", and "a no-op write MUST NOT remove `storage_changes` entries
    from earlier indices for the same slot".

    Both txs call the same contract whose runtime SSTOREs 0xff then 0x42
    to slot 1. Tx 1 changes slot 1 from 0 to 0x42 (real change). Tx 1's
    write becomes tx 2's baseline, so tx 2's 0x42 -> 0xff -> 0x42 nets to
    a no-op and only tx 1 appears in `storage_changes` (with tx 1's entry
    left intact).
    """
```

danceratopz (commented):
Thanks for finding and adding this test, @pk910!

Just a couple of minor nits below.

danceratopz (approved):
Thanks!

PR #2947, opened 2026-06-02
refactor(test-benchmark): update `test_account_access`, add `keccak256` overhead scenario

LouisTsai-Csie:
## 🗒️ Description

The original `test_account_access` targets contracts created by the `Bittrex` controller factory. This requires `CREATE` address calculations, introducing significant overhead and it is hard to precisely control over `EXP` and `KECCAK256` inputs.

We introduced several changes in this PR:
- Move the `test_account_access` from `test_single_opcode.py` to `test_account_query.py`
- Remove the `EXISTING_CONTRACT` parametrization, adding `EXISTING_CONTRACT_MINIMAL`, `EXISTING_CONTRACT_SAME` and `EXISTING_CONTRACT_DIFF` parametrization.
- Add two helper files under tests/benchmark/helper: `account_creator.py` and `account_sender_receiver.py` -> we should remove `tests/benchmark/{compute, stateful}/helpers.py` in the future to share the helper function.
- Add `test_deploy_existing_contracts` that would be used for contract deployment, this includes deploying `EXISTING_CONTRACT_*` instance and necessary delegation.

deployed contract and initcode for each parametrization:

**EXISTING_CONTRACT_MINIMAL**
initcode: `RETURN(PUSH1(0), PUSH1(1))`
deployed code: `STOP`

**EXISTING_CONTRACT_SAME**
initcode: 
```
seed 32 bytes of JUMPDEST (5b) at mem[0]
for size in 32, 64, ..., 16384:
    MCOPY mem[0:size] -> mem[size:2*size] # doubles filled region

mem[0] = 0x00 # MSTORE8: STOP

RETURN mem[0:max_code_size]
```
deployed code:

STOP JUMPDEST JUMPDEST ... ... ... JUMPDEST (fill until max code size)

**EXISTING_CONTRACT_DIFF**
initcode:
```
seed 32 bytes of JUMPDEST (5b) at mem[0]
for size in 32, 64, ..., 16384:
    MCOPY mem[0:size] -> mem[size:2*size] # doubles filled region

mem[0] = ADDRESS # MSTORE: ADDRESS

RETURN mem[0:max_code_size]
```

deployed code:

STOP STOP ... STOP ADDRESS JUMPDEST ... ... ... JUMPDEST (fill until max code size)

- code[0:12] = STOP
- code[12:32] = code address, which should be different for each contract

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


LouisTsai-Csie:
I've rebased the PR and updated the description, please review and integrate contract deployment into state-actor. cc @jochem-brouwer

tests/benchmark/stateful/helpers.py
```diff
@@ -48,6 +49,121 @@ class CacheStrategy(str, Enum):
...
+
+def account_mode_initcode(fork: Fork, account_mode: AccountMode) -> bytes:
+    """
+    Return the initcode bytes for the account mode's contracts.
+    """
+    if account_mode == AccountMode.EXISTING_CONTRACT_MINIMAL:
+        # The deployed runtime is a single zero byte (STOP).
+        return bytes(Op.RETURN(Op.PUSH1(0), Op.PUSH1(1)))
+    if account_mode == AccountMode.EXISTING_CONTRACT_JUMPDEST:
+        return bytes(build_unique_contract_initcode())
```
jochem-brouwer: The default returns the this code, I believe it should explicitly check for `AccountMode` and error if there is no match found for this `account_ mode` .

This avoid multiple `account_mode` being classified as the "default" (likely not what we want)

This method should thus have clear guards here, which throws if we use an `AccountMode` which is not supported.
It also looks like this method can  be created as Map (it is essentially a `Map` from `AccountMode` to `initcode`)

tests/benchmark/stateful/helpers.py
```diff
@@ -48,6 +49,121 @@ class CacheStrategy(str, Enum):
...
 
+class AccountMode(Enum):
+    """Target Account Mode."""
+
+    EXISTING_CONTRACT_MINIMAL = auto()
+    EXISTING_CONTRACT_SAME = auto()
+    EXISTING_CONTRACT_DIFF = auto()
+    EXISTING_CONTRACT_JUMPDEST = auto()
+    EXISTING_EOA = auto()
+    NON_EXISTING_ACCOUNT = auto()
```
jochem-brouwer: These helpers should have descriptions, so it is clear what these `AccountMode`s means (can now only guess from the name)

tests/benchmark/stateful/helpers.py
```diff
@@ -48,6 +49,121 @@ class CacheStrategy(str, Enum):
...
+    for size in (1 << s for s in range(5, 15)):
+        code += Op.MCOPY(size, 0, size)
+
+    if account_mode == AccountMode.EXISTING_CONTRACT_DIFF:
+        code += Op.MSTORE(0, Op.ADDRESS)
+    else:
+        code += Op.MSTORE8(0, 0)
+    code += Op.RETURN(0, max_code_size)
+
+    return code
```
jochem-brouwer: This code is ambiguous because it should have guard for what AccountMode it can be called.
If it gets called with `EXISTING_CONTRACT_JUMPDEST` then the returned initcode is incorrect (and the method should not be called with it). It seems the only ones we can call it with is `EXISTING_CONTRACT_DIFF` and `EXISTING_CONTRACT_SAME`

tests/benchmark/stateful/helpers.py
```diff
@@ -48,6 +49,121 @@ class CacheStrategy(str, Enum):
...
+    Embedded ADDRESS makes runtime unique per contract;
+    initcode and its CREATE2 hash is shared across all salts.
+    """
+    max_code_size = JOCHEMNET_RUNTIME_SIZE
+
+    # MCOPY fills MEM[0:0x8000] with JUMPDEST.
+    # Runtime only uses MEM[0:0x6000].
+    code = Op.MSTORE(0, bytes(Op.JUMPDEST * 32))
+    for size in (1 << s for s in range(5, 15)):
+        code += Op.MCOPY(size, 0, size)
```
jochem-brouwer: It helps if this has the hint `1 <<  5 = 32` (so the first `MCOPY` is `Op.MCOPY(32, 0, 32)` (so it copies the just MSTOREd 32 bytes of JUMPDEST from 0-32 to offset 32, i.e. doubles the memory which has JUMPDEST, and the initial code for this is correct)

jochem-brouwer (commented):
Some small comments!

tests/benchmark/stateful/helpers.py
LouisTsai-Csie: I do a huge refactor, please check `AccountCreator` for more details.

tests/benchmark/stateful/helpers.py
LouisTsai-Csie: Added in the refactored version, please check if this is clear to you

tests/benchmark/stateful/helpers.py
LouisTsai-Csie: Done in the latest refactor!

tests/benchmark/helper/account_creator.py
```diff
@@ -0,0 +1,319 @@
...
+    Create2PreimageLayout,
+    Hash,
+    Op,
+    SequentialAddressLayout,
+    keccak256,
+)
+
+# Runtime code size of the jochemnet contract (EIP-170 limit).
+JOCHEMNET_RUNTIME_SIZE = 0x6000
+
```
LouisTsai-Csie: Based on my current understanding, we want to do the repricing based on the old max code size value (0x6000), could you please confirm? cc @jochem-brouwer 

This is applied to all the following contract variants. Note that if i use `fork.max_code_size()` instead, this would be affected by eip-7954.

tests/benchmark/helper/account_sender_receiver.py
```diff
@@ -0,0 +1,80 @@
...
+# the accounts stay uncached.
+SENDER_BASE_KEY = int.from_bytes(
+    keccak256(b"gas-repricings-private-key"), "big"
+)
+
+# Deterministic EIP-7702 delegate authorities:
+# Authority i delegates to EXISTING_CONTRACT_DIFF receiver i.
+DELEGATE_BASE_KEY = int.from_bytes(
+    keccak256(b"gas-repricings-7702-delegate"), "big"
+)
```
LouisTsai-Csie, line 26: Starting from this range, we delegate to EXISTING_CONTRACT_DIFF contracts for a certain count. See `test_deploy_existing_contracts`. I need to verify if the current count is sufficient.

tests/benchmark/helper/account_creator.py
jochem-brouwer: Yes, the repricing is based on Osaka, so without any changes in Amsterdam 😄 👍

jochem-brouwer: We could also replace this constant by `fork(Osaka).max_code_size` (convert this to the correct values) instead of using this constant

tests/benchmark/helper/account_creator.py
```diff
@@ -0,0 +1,319 @@
...
+
+class AccountMode(Enum):
+    """Benchmark target account variant."""
+
+    # Minimal contract: single STOP byte.
+    EXISTING_CONTRACT_MINIMAL = auto()
+
+    # Max-size contract: byte-identical across copies.
+    EXISTING_CONTRACT_SAME = auto()
```
jochem-brouwer:
```suggestion
    EXISTING_CONTRACT_SAME_MAX = auto()
```

This is a more descriptive name. Maybe we should also drop the `EXISTING_` part, I think if there are any benchmarks related to state left which query state deployed in the setup phase, we should change those benchmarks to target existing state (state-actor, or state which should exist on the target chain (mainnet/jochemnet for instance))

tests/benchmark/helper/account_creator.py
```diff
@@ -0,0 +1,319 @@
...
+
+    # Max-size contract: ADDRESS-embedded, each copy unique.
+    EXISTING_CONTRACT_DIFF = auto()
```
jochem-brouwer:
```suggestion
    EXISTING_CONTRACT_DIFF_MAX = auto()
```

tests/benchmark/helper/account_creator.py
```diff
@@ -0,0 +1,319 @@
...
+
+    # Max-size contract: exercises JUMPDEST analysis.
```
jochem-brouwer:
```suggestion
    # Max-size contract: exercises JUMPDEST analysis. The code is unique.
```

It is not directly clear if this is an unique or non-unique JUMPDEST contract.

tests/benchmark/stateful/bloatnet/test_transaction_types.py
```diff
@@ -145,59 +51,66 @@ def test_ether_transfers_onchain_receivers(
...
+        case "diff_to_self":
+            receivers = senders
+        case "diff_to_nonexistent":
+            receivers = yield_distinct_nonexistent_receiver()
+        case "diff_to_existent":
+            receivers = yield_distinct_existent_receiver()
+        case "diff_to_contract":
+            receivers = yield_distinct_contract_receiver()
+            # Runtime code is the same across all the receivers
+            # Example contract: https://etherscan.io/address/0xa888df3ef62286dde06a79395760b9bce6c83c83#code
```
jochem-brouwer, line 73: It should be made clear here that this runtime is the code what is being executed, not the actual runtime code (this specific compiled code would try to JUMPI to 0x49 which has non-existent code and would thus OOG because of invalid jump)

tests/benchmark/stateful/bloatnet/test_transaction_types.py
```diff
@@ -1,139 +1,45 @@
...
     "case_id",
     [
+        "diff_to_self",
         "diff_to_nonexistent",
         "diff_to_existent",
         "diff_to_contract",
         "diff_to_unique_code_jumpdest_contract",
+        "diff_to_contract_minimal",
+        "diff_to_contract_same",
+        "diff_to_contract_diff",
```
jochem-brouwer: For the test description `diff_to_contract_same_max` and `diff_to_contract_diff_max` should be used so it is clear that it uses max-sized contracts (Osaka max size)

jochem-brouwer: Beyond the scope of this PR, but we should also include max-sized contract tests for Glamsterdam (which means we have to do something about the naming here too, as it is confusing what the "max" then refers to. Osaka or Amsterdam?)

jochem-brouwer (commented):
Some minor doc comments! Looks good in general! One comment about a follow-up PR to also include Amsterdam tests (we should target those for state-actor)

jochem-brouwer (commented):
Some comments!

tests/benchmark/helper/account_creator.py
LouisTsai-Csie: We could refactor all the naming issue in a follow-up PR!

tests/benchmark/stateful/bloatnet/test_transaction_types.py
LouisTsai-Csie: Yes! Let's discuss this during next benchmarking call!

tests/benchmark/helper/account_creator.py
```diff
@@ -0,0 +1,337 @@
...
+    """Benchmark target account variant."""
+
+    # Minimal contract: single STOP byte.
+    EXISTING_CONTRACT_MINIMAL = auto()
+
+    # Max-size contract: byte-identical across copies.
+    EXISTING_CONTRACT_SAME_MAX = auto()
+
+    # Max-size contract: ADDRESS-embedded, each copy unique.
+    EXISTING_CONTRACT_DIFF_MAX = auto()
```
jochem-brouwer, line 32: We should Glamsterdam versions in a later PR (and we likely want to make the "code size" part of the constructor of these generators). For repricings we want Osaka-sized contracts, but for the generic Amsterdam benchmarks (which verifies pre-fork if we can safely raise the gas limit to gas limit X, based on the generic benchmarks (so all benchmarks, not only the specific repricing ones)). For these we want to also target Amsterdam max-sized contracts to verify that clients can handle this throughput (the max code size increases by a factor `2+2/3` (64/24) ([EIP-7954](https://eips.ethereum.org/EIPS/eip-7954)).

(we should do this soon, this also makes me realize that we want to run the benchmarks ASAP 👀 Let's attempt to get this up within 2 weeks)

tests/benchmark/helper/account_creator.py
```diff
@@ -0,0 +1,337 @@
...
+    Initcode whose deployed runtime embeds its own contract ADDRESS.
+
+        offset    size   contents
+        ------    ----   --------------------------------
+        0x0000       4   PUSH2 0x5FFF; JUMP   <- entry
+        0x0004      28   JUMPDEST padding
+        0x0020      12   JUMPDEST padding
+        0x002C      20   contract ADDRESS     <- unique
+        0x0040   24512   JUMPDEST             <- 0x5FFF lands here
+        0x6000           STOP
```
jochem-brouwer:
```suggestion
```

This is incorrect, attempting to deploy bytecode at offset 0x6000 would deploy a contract sized of 0x6001, which violates EIP-170 (bytecode starts at offset 0)

The STOP opcode is implicit whenever bytecode runs out-of-bounds. This is also why running empty bytecode (i.e. non-delegated EOAs) the (empty) bytecode which is being ran is thus implicitly the STOP opcode (0 gas). (nit: this does not turn up in debug_traceX outputs since the opcode is not "actually executed" since the opcode does not "exist" at the account)

(suggestion removes this line)

tests/benchmark/helper/account_sender_receiver.py
jochem-brouwer: (this is a response on this comment: https://github.com/ethereum/execution-specs/pull/2947/changes#r3434132846) 

----

This should [be](https://github.com/ethereum/execution-specs/pull/2947/changes#diff-2e05f30d8d88cfa54bf15f04da6dd2c9318b2bb917468b99dffb50039d8be520R105-R118) (`test_transaction_types.py`):

```
    iteration_cost = (
        fork.transaction_intrinsic_cost_calculator()(
            sends_value=sends_value,
            recipient_type=recipient_type,
        )
        + fork.transaction_top_frame_gas_calculator()(
            sends_value=sends_value,
            recipient_type=recipient_type,
        )
        + fork.transaction_top_frame_state_gas(
            sends_value=sends_value,
            recipient_type=recipient_type,
        )
        + receiver_execution_gas
    )
```

Where the amount of target accounts is `gas_benchmark_value // iteration_cost`. 

Assuming this calculation is correct and yields the max amount of accounts we can reach given the scenario (target accounts) then this yields the amount of accounts which should pre-exist. 

Is this formula correct for the situations we test here? If we target one account per transaction then this is correct, if we do this via e.g. `CALL` opcodes then this is not correct as it assumes we keep paying the top frame costs each iteration.

jochem-brouwer (commented):
Some small comments, CI fails also, once this is fixed I will do one extra pass over the devnet-6 specs to see if this is correct.

tests/benchmark/helper/account_sender_receiver.py
LouisTsai-Csie: This is correct because we access these deployed contracts via transaction, not via `CALL`!

tests/benchmark/helper/account_creator.py
LouisTsai-Csie: Done

tests/benchmark/helper/account_creator.py
LouisTsai-Csie: I will refactor the generator to receive code size parameter.

tests/benchmark/stateful/bloatnet/test_transaction_types.py
LouisTsai-Csie: I rename it to `execute_code`!

jochem-brouwer (approved):
Great refactor!! 😄 👍 LGTM 🎉

PR #2949, opened 2026-06-02
fix(tests): accept tx-level exception for over-budget block (EIP-7778)

chfast:
## 🗒️ Description

In test_multi_transaction_gas_accounting, a block whose included transaction exceeds the remaining block gas was expected to fail only with the block-level BlockException.GAS_USED_OVERFLOW. A client may legitimately report the same condition at the transaction level as TransactionException.GAS_ALLOWANCE_EXCEEDED (the best I found).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
+++ b/tests/amsterdam/eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py
@@ -327,7 +327,10 @@ def test_multi_transaction_gas_accounting(
         blocks=[
             Block(
                 txs=txs,
-                exception=BlockException.GAS_USED_OVERFLOW
+                exception=[
+                    BlockException.GAS_USED_OVERFLOW,
+                    TransactionException.GAS_ALLOWANCE_EXCEEDED,
+                ]
                 if exceed_block_gas_limit
                 else None,
                 expected_gas_used=total_block_gas_used


danceratopz (approved):
This is the right fix.

At least for now, but this is something we should definitely revisit when we attack the exception test topic properly.

The specs themselves also raise a transaction-level exception for this and it bothered me that we don't dogfood our own fixtures to actually verify this and avoid this issue. Indeed, it would appear other clients have essentially worked around this exception expectation of this test by avoiding distinct substring matches for both the transaction-level and the block-level exception. That's the only reason why no one else flagged this, but this is the better fix. I.e., 5 of the client exception mappers more or less define the following entries to ensure they pass this class of test:

```python
  GAS_ALLOWANCE_EXCEEDED: "Block gas limit exceeded"  # TransactionException
  GAS_USED_OVERFLOW: "Block gas limit exceeded"  # BlockException
```

On the specs side:
```
  state_transition(chain, block)        fork.py:229   ← block import/validation
  entrypoint
    └─ apply_body(...)                   fork.py:320/809
         └─ for each tx: process_transaction(...)   fork.py:854/926
              └─ check_transaction(...)             fork.py:974/482
                   └─ raise GasUsedExceedsLimitError   fork.py:551  ← here
```
but this is not verified rigorously anywhere against the fixture's exception or we would have caught this earlier. Two issues:

1. Nowhere do the specs currently consume **all new** test fixtures in CI.
2. Even where specs consume fixtures, the `json-loader` only verifies that an exception occurs, but not that it's the correct one, so we don't hold the specs to the same standards as the clients.

chfast:
I only started matching error codes this week. And this is only partial for now.

danceratopz:
> I only started matching error codes this week. And this is only partial for now.

Just fyi, we could consider adding it `consume direct`. This would save you the work on the test runner side but likely be slower.

PR #2951, opened 2026-06-03
feat(consume direct, tests): add erigon support, also add test to showcase erigon statetest bug

felix314159:
## 🗒️ Description
firstly, this title allows us to run `consume direct` against erigon (a bit tricky, cuz just from the version output alone you can't differentiate it reliably from geth's evm). but i check for certain strings in the binary so now it is robust.

secondly, differential fuzzing found a bug in erigon's evm (not relevant for mainnet, but annoying during fuzzing):

### problem

erigon, when it has to reconstruct and sign a state-test transaction itself (no `txbytes` in fixture, pre-London fork), skips the pre-execution validity check (only happens pre-London) and executes the transaction instead of rejecting it for insufficient funds. all other clients reject the tx.

note: because txbytes it missing, the clients reconstruct that field from the human-readable tx object. so the bug only happens in this specific code path which is not relevant for mainnet. but it is important to get this fixed to reduce noise during differential fuzzing.

## how to reproduce

step 0: get latest erigon main and build the t8n via `make all`

---

step 1: fill fixtures using:
`uv run fill tests/frontier/validation/test_transaction.py::test_sender_balance_insufficient_state_test --clean -v -s`

---

step 2: artifically remove 'txbytes' keys from fixtures (not possible in execution-specs, but requirement to trigger this bug in erigon's statetest t8n):

`find fixtures/state_tests -path '*sender_balance_insufficient*' -name '*.json' -exec sh -c 'jq "del(.. | .txbytes?)" "$1" > "$1.tmp" && mv "$1.tmp" "$1"' _ {} \;`

this step is necessary because our framework would never delete those keys, so this standalone command is simpler than adding it as a feature in execution-specs

---

step 3: run consume direct:
`uv run consume direct --bin /home/$USER/Documents/go-ethereum/build/bin/evm --input ./fixtures --sim.limit ".*sender_balance_insufficient_state_test.*"`

-> erigon will fail 6/12 tests with: `AssertionError: State test failed: expected error "TransactionException.INSUFFICIENT_ACCOUNT_FUNDS", got no error` (will fail for all 6 pre-London forks)

-> geth and other clients with pass all 12/12

#### Cute Animal Picture

Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/client_clis/__init__.py
+++ b/packages/testing/src/execution_testing/client_clis/__init__.py
@@ -14,6 +14,15 @@ from .cli_types import (
 )
 from .client_backend import ClientBackend, ClientBackendExceptionMapper
 from .clis.besu import BesuFixtureConsumer, BesuTransitionTool
+
+# NOTE: erigon is imported before geth so it is registered (and thus probed)
+# first. Both expose an `evm` binary printing `evm version ...`; go-ethereum's
+# detection matches that banner unconditionally, so it would otherwise claim an
+# Erigon binary. ErigonEvm.detect_binary positively fingerprints Erigon (via
+# the `enginextest` subcommand) and declines anything else, so a go-ethereum
+# binary checked here falls through to GethEvm — the ordering only gives Erigon
+# first look, it does not by itself decide identity.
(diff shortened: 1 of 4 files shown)


marioevz (approved):
🚀

PR #2965, opened 2026-06-09
feat(spec-tool): enforce `@final` on leaf dataclasses

danceratopz:
## 🗒️ Description

Follow-up to #2934. Add a `FinalDecoratorHygiene` rule to `ethereum-spec-lint` that enforces `@final` on leaf dataclasses across the specification.

PR #2934 added `@final` to leaf dataclasses (`@dataclass` and `@slotted_freezable` classes that are never subclassed) so `mypyc` can bypass the vtable for method calls and property accessors. This rule keeps that coverage from drifting as new forks are copied forward or new dataclasses are added.

The rule:

- Flags any leaf dataclass that is missing `@final`. Classes that are subclassed are skipped because they are not leaves, as are non-dataclass types such as enums, protocols, exceptions, and constant namespaces.
- Scans the whole specification once: every fork's modules plus the shared modules such as `ethereum.state` and `ethereum.trace`, including each package's `__init__.py`.
- Is registered in `vulture_whitelist.py`, since lints are discovered dynamically.

Note: Because the check is a whole-specification invariant, it runs once at the first fork position, so diagnostics are grouped under the earliest fork's header (`frontier`). Each message names the full module path, for example `ethereum.forks.amsterdam.state_tracker`, so the location is unambiguous.

Example output:
image

## 🔗 Related Issues or PRs

Related to #2934.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

image

### DIFF
--- a/src/ethereum_spec_tools/lint/lints/final_decorator.py
+++ b/src/ethereum_spec_tools/lint/lints/final_decorator.py
@@ -1,0 +1,147 @@
+"""
+Final Decorator Lint.
+
+Ensures that leaf dataclasses are decorated with `@final`.
+"""
+
+import ast
+import importlib
+import inspect
+import pkgutil
+from typing import Generator, List, Optional, Sequence, Set, Tuple
+
+from ethereum_spec_tools.forks import Hardfork
+from ethereum_spec_tools.lint import Diagnostic, Lint
+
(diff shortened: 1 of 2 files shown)


SamWilsn (approved):
Good idea!

PR #2901, opened 2026-05-22
feat(spec-specs, tests): merge EIP-8037 to `forks/amsterdam`

spencer-tb:
## 🗒️ Description

This PR merges EIP-8037 into `forks/amsterdam` including framework, specs and all test changes.

**Review Plan**
- **Specs** - `src/ethereum/forks/amsterdam/` @SamWilsn
- **Tests batch 1** - Targeted 8037 cases, `tests/amsterdam/`, 54 files @marioevz 
- **Tests batch 2** - Cross fork non-static, `tests/osaka` to `tests/frontier`, 78 files  
  - tests/berlin - @kclowes 
  - tests/byzantium - @kclowes 
  - tests/cancun - @marioevz
  - tests/common - @marioevz 
  - tests/constantinople  - @marioevz 
  - tests/frontier - @marioevz 
  - tests/homestead - @marioevz 
  - tests/istanbul - @marioevz 
  - tests/osaka - @kclowes 
  - tests/paris - @marioevz 
  - tests/prague @marioevz 
  - tests/shanghai @marioevz 
  
- **Tests batch 3** - Cross fork static-ported, `tests/ported_static` audit @leolara 
- **Testing Framework** - `packages/testing` @marioevz 

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    just static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).

Issue #2957, opened 2026-06-04
`consume direct`: xfail geth for 7610


felix314159:
if you fill 

`uv run fill -v -s --no-html --clean ./tests/paris/eip7610_create_collision/test_revert_in_create.py  -k test_collision_with_create2_revert_in_initcode`

and then consume with:

`uv run consume direct -v -s --bin=/home/user/Documents/erigon/build/bin/evm`

erigon (any other clients) pass all tests, but if you run it against `geth` you get 10 failures. that is because geth hardcodes the few relevant cases that exist for `eip-7610` (so it is not an actual problem, their approach is valid and correct). i don't think we can easily fix this test, and our framework does not have the feature where you can specify expected fails (xfail) on a client-to-client granularity. this issue is about the question whether we want such a feature, or whether we can rework this test in some other way that makes all clients happy.

also the fact that we do not catch these failures in an automated way supports the argument that we should give more love to `consume direct` and maybe even include it in CI in some form. i tripped over this when fuzzing btw

spencer-tb:
Duplicate of #1530. The ask here is a per-client expected-failure marker for `consume`, which is exactly what #1530 tracks, and the geth behaviour itself is valid as you noted. Closing in favour of #1530.

PR #2969, opened 2026-06-09
feat(test-specs): Allow automatic transaction gas-limit

### LINKED ISSUE #2248: Suggestion: For tests, let the default tx gas limit be the maximum allowed value
Most tests do not meter gas usage. And they use an arbitrary gas limit to make sure they pass.

https://github.com/ethereum/execution-specs/blob/c708ab6075973b6b928d17ec073dd43a7fe161b6/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_invalid.py#L159-L163

Set the default to the EIP-7925 maximum value so:

* Implementers can fall back to a sane limit defined by the fork instead of an arbitrary value.
* It can still be overridden if required.

marioevz:
### Summary

While reviewing #2901 I noticed that the bulk of its test changes were simply bumping the `gas_limit` of transactions in tests that don't actually care about the exact gas limit — they only care that the transaction executes in full without running out of gas.

This PR makes the transaction `gas_limit` optional. When a test omits it, the filling/execution tooling calculates it automatically based on fork properties.

#### Tests

| folder | Owner |
| - | - |
| tests/amsterdam/eip7928_block_level_access_lists | @raxhvl |
| tests/amsterdam/eip8037_state_creation_gas_cost_increase | @spencer-tb |
| tests/amsterdam/* (others) | @spencer-tb |
| tests/berlin | @kclowes |
| tests/byzantium | '' |
| tests/cancun | '' |
| tests/common | '' |
| tests/constantinople | '' |
| tests/frontier | '' |
| tests/homestead | @felix314159 |
| tests/istanbul | '' |
| tests/osaka | '' |
| tests/paris | '' |
| tests/prague | '' |
| tests/shanghai | @danceratopz |
| tests/ported_static | @spencer-tb |

#### Testing Framework

| folder | Owner |
| - | - |
| packages/testing/src/execution_testing/cli/pytest_commands/plugins | @leolara |
| packages/testing/src/execution_testing/execution | @leolara |
| packages/testing/src/execution_testing/fixtures | @leolara |
| packages/testing/src/execution_testing/forks | @leolara |
| packages/testing/src/execution_testing/specs | @leolara |
| packages/testing/src/execution_testing/test_types | @leolara |
| packages/testing/src/execution_testing/tools | @leolara |

#### `Transaction.gas_limit` now defaults to `None` instead of `HexNumber(21_000)`.

`Transaction.set_gas_limit()` resolves an unset limit in place, and signing now raises `ValueError("gas_limit must be set to sign a transaction")` if the limit is still unset by the time the transaction is signed.

This lead to many typing issues that have now been resolved, but it can be argued that a better choice is to leave `HexNumber` as the type for `Transaction.gas_limit` and rely instead on `model_fields_set` to determine whether the gas limit has been set by the tester.

#### `BlockchainTest` and `StateTest` Fillers

For State tests (`StateTest`), an unset `gas_limit` is filled with the fork's `transaction_gas_limit_cap()`, falling back to the environment `gas_limit` when there is no cap.

For Blockchain tests (`BlockchainTest`), the block's remaining gas (`env.gas_limit` minus the gas already claimed by transactions with explicit limits) is split evenly across the transactions that leave `gas_limit` unset, clamped to the fork's transaction gas-limit cap. It raises a clear "test correctness" error if no gas remains.

#### Execute Tests

The same logic is implemented for the execute path (`calculate_max_transaction_gas_limit` in `execution/base.py`, applied in `transaction_post.py` and `blob_transaction.py`), so implicit gas limits work when running against a live client, not just during filling.

It can be argued that there's a better path to resolve the transaction gas limit in the case of execute, such as:
- Use `eth_estimateGas` and use the value returned, with the minor caveat that the contracts that the transaction requires have to be already on chain before calling this RPC method.
- Run the T8N tool interface with the transaction to obtain the gas used.

Both methods require more work and could be implemented in a future PR.

#### State gas reservoir (EIP-8037)
  
Rather than special-casing EIP-8037 (`fork.is_eip_enabled(8037)`), this PR introduces a fork-level concept:

- New abstract fork method `Fork.state_gas_reservoir_enabled()` (alongside the existing `transaction_gas_limit_cap()`), defaulting to False and turned on by EIP-8037, currently in Amsterdam.
- New `Transaction.state_gas_reservoir` field. When a fork enables the reservoir, the transaction gas-limit cap is treated as removed and `set_gas_limit()` accounts for the requested reservoir (transaction_gas_limit_cap + state_gas_reservoir) when sizing the implicit limit.
- EIP-8037 tests were refactored to express their requirement through state_gas_reservoir instead of hand-computing gas limits (`tx.state_gas_reservoir = state_gas_reservoir` instead of `tx.gas_limit = gas_limit_cap + state_gas_reservoir`).

### Test cleanup

The implicit gas limit was applied across the test suite, removing explicit `gas_limit` assignments from transactions wherever the precise limit was irrelevant to what the test exercises. This touches ~509 test files (net ~7,200 lines removed) spanning Frontier through Amsterdam and many individual EIPs (1153, 4844, 5656, 6780, 7516, 7708, 7843, 7928, 8024, Blake2, etc.).
  
Tests that should rely on the implicit gas limit are those that only need the transaction to execute fully without running out of gas. Tests that intentionally probe gas-limit behavior continue to set it explicitly.

### Incidental fixes surfaced while porting tests

  - `pre.fund_eoa` default bumped from `10**21` to `10**27` to fix deposit tests that were under-funded.
  - EIP-7883 was not being activated in Osaka; this is now fixed.

## 🔗 Related Issues or PRs

Closes #2248:
- #2248

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


raxhvl:
Sweet. I think this closes https://github.com/ethereum/execution-specs/issues/2248

marioevz:
> Sweet. I think this closes #2248

This is exactly your suggestion indeed! Feel free to pick up some folder to review, perhaps BAL tests?

packages/testing/src/execution_testing/specs/state.py
```diff
@@ -356,7 +356,13 @@ def make_state_test_fixture(
         )
 
         env = self.env.set_fork_requirements(fork)
-        tx = self.tx.with_signature_and_sender(keep_secret_key=True)
+        tx = self.tx
```
leolara: Perhaps, it cannot cause problems, because this self is not used after make_stte_fixture. But, `with_signature_and_sender` has non mutation sematics, so self.tx is not mutated before, now we are mutating self.tx.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -823,7 +851,23 @@ def generate_block_data(
             block_number=env.number, timestamp=env.timestamp
         )
         env = env.set_fork_requirements(fork)
-        txs = [tx.with_signature_and_sender() for tx in block.txs]
+        txs = block.txs[:]
```
leolara, line 826: the same as in the other comment, tx.with_signature_and_sender() will return a new copy, in the new code we are mutating inside blocks.txs

leolara:
blockchain.py `calculate_max_transaction_gas_limit` and base.py `calculate_max_transaction_gas_limit` look the same, perhaps we can eliminate this duplication

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip4895.py
```diff
@@ -759,8 +739,29 @@ def test_bal_withdrawal_to_coinbase(
...
+    tx_value = 5
+    tx = Transaction(
+        sender=alice,
+        to=bob,
+        value=tx_value,
+        gas_limit=intrinsic_gas,
+        **tx_kwargs,
+    )
+
+    tip_to_coinbase = priority_fee * intrinsic_gas
```
raxhvl, line 762: idea: Abstract transaction tip calculation

In future, perhaps expose a `tx.tip()` method that computes the tip based on gas limit and tx type.

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -837,13 +842,50 @@ def set_gas_price(
...
+        transaction_gas_limit_cap: int | None,
+        state_gas_reservoir_enabled: bool = False,
+    ) -> None:
+        """Set the transaction gas limit if unset."""
+        if "gas_limit" not in self.model_fields_set or self.gas_limit is None:
+            tx_gas_limit = max_gas_limit
+            if state_gas_reservoir_enabled:
+                if "state_gas_reservoir" in self.model_fields_set:
+                    assert transaction_gas_limit_cap is not None, (
+                        "Impossible to set calculate the tx gas limit for the "
```
leolara:
```suggestion
                        "Impossible to calculate the tx gas limit for the "
```

leolara:
@marioevz 

I have been thinking about `gas_limit: HexNumber | None` breaks the typing because the parent doesn't have `None`. I think changing the parent would have to many downstream effects, perhaps there are tests that read `gas_limit`.

So, perhaps it is lighter to take out the `None` and use pydantic's `model_fields_set`.

Do you want me to try and see how it looks?

packages/testing/src/execution_testing/execution/base.py
```diff
@@ -42,18 +43,60 @@ def __pydantic_init_subclass__(cls, **kwargs: Any) -> None:
...
+        """
+        available_gas = int(env.gas_limit)
+        unset_gas_limit_tx_count = 0
+        for tx in txs:
+            if tx.gas_limit is None:
+                unset_gas_limit_tx_count += 1
+            else:
+                available_gas -= int(tx.gas_limit)
+
+        if unset_gas_limit_tx_count == 0 or available_gas <= 0:
```
danceratopz: Returning `0` conflates "no implicit-gas-limit txs" (benign) with "no remaining gas" (test bug), and the execute formats never check the return value, so an over-subscribed block would send transactions signed with `gas_limit=0`. Proposed fix: danceratopz/execution-specs@375551170d13df815cf378810cbc808304586396

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -803,6 +803,34 @@ def make_genesis(
             ).with_rlp(txs=[]),
         )
 
+    @staticmethod
+    def calculate_max_transaction_gas_limit(
```
danceratopz: Verbatim copy of `BaseExecute.calculate_max_transaction_gas_limit()`; the two encode the same sizing policy and will drift. Proposed dedup into `transaction_types`, next to `Transaction.set_gas_limit()`: danceratopz/execution-specs@fc016b98e80cd0c690337c66f943adbc179d9e53

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -823,7 +851,23 @@ def generate_block_data(
             block_number=env.number, timestamp=env.timestamp
         )
         env = env.set_fork_requirements(fork)
-        txs = [tx.with_signature_and_sender() for tx in block.txs]
+        txs = block.txs[:]
+        if any(tx.gas_limit is None for tx in block.txs):
+            max_tx_gas_limit = self.calculate_max_transaction_gas_limit(
+                txs, env, fork
+            )
+            if max_tx_gas_limit == 0:
```
danceratopz, line 834: This caller-side check is what disambiguates the `0` sentinel, but it only protects this call site; moving the raise into the function protects the execute formats too: danceratopz/execution-specs@7b864ab56f7d31b2503c30fe41c4772b9a2199e8

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -331,6 +331,8 @@ def strip_hash_from_t8n_output(cls, data: Any) -> Any:
 
     expected_receipt: TransactionReceipt | None = Field(None, exclude=True)
 
+    state_gas_reservoir: int = Field(0, exclude=True)
```
danceratopz: The three-state semantics (unset / explicit `0` / positive) are subtle and the unset-vs-zero distinction lives invisibly in `model_fields_set`; proposed field description and `set_gas_limit` docstring: danceratopz/execution-specs@cbcb47482dc1b5f2fe63664ef2a3f09bb553782d

packages/testing/src/execution_testing/test_types/transaction_types.py
danceratopz: Typo ("set calculate"), and the message presumes the source of the inconsistency; proposed symptom-first rewording: danceratopz/execution-specs@cbcb47482dc1b5f2fe63664ef2a3f09bb553782d

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -837,13 +842,50 @@ def set_gas_price(
...
+                    assert transaction_gas_limit_cap is not None, (
+                        "Impossible to set calculate the tx gas limit for the "
+                        "required state gas reservoir without a gas limit cap"
+                    )
+                    if self.state_gas_reservoir > 0:
+                        minimum_gas_with_reservoir = (
+                            transaction_gas_limit_cap
+                            + self.state_gas_reservoir
+                        )
+                        assert tx_gas_limit >= minimum_gas_with_reservoir
```
danceratopz: This bare assert is reachable through normal test authoring (a crowded block pushes the even split below `cap + reservoir`) and fails with no message; proposed "test correctness" error with the numbers: danceratopz/execution-specs@cbcb47482dc1b5f2fe63664ef2a3f09bb553782d

packages/testing/src/execution_testing/test_types/transaction_types.py
```diff
@@ -837,13 +842,50 @@ def set_gas_price(
...
+                        tx_gas_limit = minimum_gas_with_reservoir
+                    else:
+                        if tx_gas_limit > transaction_gas_limit_cap:
+                            tx_gas_limit = transaction_gas_limit_cap
+            else:
```
danceratopz: A positive `state_gas_reservoir` on a fork without the reservoir is silently dropped here (clamped to cap on Osaka, ignored pre-Osaka), so a fork-parametrized test can silently stop testing what it claims; proposed guard (keeps the `state_gas_reservoir=0` idiom working): danceratopz/execution-specs@366732508d8a41f9a3c19d691d63e0a70ae965f0

felix314159:
i looked at the changes for `homestead`, `istanbul`, `osaka`, `paris`, `prague` and `shanghai`, they are fine.
`fill` of modified files still works, and almost all tests also passed `consume direct` against geth. the only failures were 140 failures for ` tests/paris/eip7610_create_collision` but those are expected (i made an [issue](https://github.com/ethereum/execution-specs/issues/2957) for this recently). it is a good PR, we should have done this earlier

danceratopz (commented):
Hey @marioevz, this is a great addition. Here's a review of the framework changes (only covered the testing framework so far; with partial review of tests for context: `tests/shanghai/`, `tests/amsterdam/eip7928_*`). Few comments below, each links to a proposed fix; the full set, including a new unit test module for the dynamic gas-limit machinery is in marioevz/execution-specs#6.

marioevz:
@LouisTsai-Csie could you check if `split_transaction` from the benchmark tests still work fine?

packages/testing/src/execution_testing/specs/blockchain.py
marioevz: Addressed above on Leo's comment.

packages/testing/src/execution_testing/test_types/transaction_types.py
marioevz: Addressed on Leo's comment above.

marioevz:
@leolara @danceratopz pushed the fixes to all of your comments, thank you for the reviews!

danceratopz:
> @leolara @danceratopz pushed the fixes to all of your comments, thank you for the reviews!

Thanks @marioevz, I think you missed the unit tests added in https://github.com/marioevz/execution-specs/pull/6/changes/ebbe563a9be150622f3a44535fec5a815000b681

raxhvl (approved):
Reviewed 7928 and 7708

kclowes (commented):
Berlin-Frontier look good to me. This is such a nice change!

leolara:
@marioevz I think there are places where we are setting Transaction(gas_limit=None) but setting it to a variable that can be `None`.

leolara:
> @marioevz I think there are places where we are setting Transaction(gas_limit=None) but setting it to a variable that can be `None`.

Like here: https://github.com/marioevz/execution-specs/blob/f82f657ed2ff7f4d0a5ecd39b3aadb4aaae41d16/tests/prague/eip7002_el_triggerable_withdrawals/helpers.py#L117

leolara:
I think there are no unit tests for with_gas_limit, _calculate_implicit_gas_limit, calculate_max_gas_limit and state_gas_reservoir. Perhaps we don't need it because they are called from other tests.

leolara (approved):

marioevz:
> > @marioevz I think there are places where we are setting Transaction(gas_limit=None) but setting it to a variable that can be `None`.
> 
> Like here: https://github.com/marioevz/execution-specs/blob/f82f657ed2ff7f4d0a5ecd39b3aadb4aaae41d16/tests/prague/eip7002_el_triggerable_withdrawals/helpers.py#L117

Nice catch, uploaded a new commit https://github.com/ethereum/execution-specs/pull/2969/commits/a2e13fb71164d37df9f1df962dc50a925200c608. Just as a side note, I think this specific tests need an update because they are no longer really accurate (they hard-code assumed gas costs for the system contracts, which no longer hold).

marioevz:
Once @spencer-tb signs off we can proceed to merge.

spencer-tb (approved):
Did a final check this morning! All LGTM :)
#2984 feels appropriate to me too. Lets merge!

PR #2973, opened 2026-06-10
feat(ci): deploy default branch docs at /docs/execution-specs/

danceratopz:
TLDR:

We currently deploy the default branch's docs (forks/amsterdam) at:
- https://steel.ethereum.foundation/docs/execution-specs/forks/amsterdam/

This PR and https://github.com/ethereum/steel-website/pull/59 changes this to deploy at:
- https://steel.ethereum.foundation/docs/execution-specs/

This is so that when we bump the default branch, the URL doesn't change: The most up-to-date version of the docs will always be available at https://steel.ethereum.foundation/docs/execution-specs/

## Summary

- Adds a `default_branch` field to `.github/configs/docs-branches.yaml` to name the branch whose docs are deployed at the root of the product namespace on steel.ethereum.foundation (rather than under a `<branch>/` segment).
- Updates `docs-build.yaml` to set `SITE_URL` without a `<branch>/` segment when building the default branch, so canonical, OG, Twitter, and sitemap URLs match where the artifact is actually published.
- Removes `devnets/bal/4` from the allowlist (no longer required).

## Commits

1. `feat(ci): deploy default branch docs at /docs/execution-specs/` — the substantive change.
2. `refactor(ci): resolve SITE_URL once in check-should-publish` — cleanup pass: `SITE_URL` is now derived once and emitted as a job output; the publish-summary line and the html-docs Build step both consume it. Replaces two separate default-vs-non-default conditionals. Same behavior, html-docs Build drops from 14 lines to 4.

## Why

Today the default development branch (currently `forks/amsterdam`) ships at `steel.ethereum.foundation/docs/execution-specs/forks/amsterdam/`. The URL becomes ephemeral at each fork rollover and the docs landing page has to recommend `mainnet` for stable incoming links, which lags reality.

Pairing this PR with the steel-website change moves the default branch's docs to `steel.ethereum.foundation/docs/execution-specs/`, which is stable across rollovers and is the right URL to share. This PR is the upstream half: it ensures `SITE_URL` (and therefore canonical/OG/sitemap URLs baked into the built HTML) matches the new publish path.

## Pairing

- Requires https://github.com/ethereum/steel-website/pull/59, which stages the default branch artifact at `site/docs/execution-specs/` instead of `site/docs/execution-specs/<branch>/`.
- Order is flexible: until both land, the default branch URL is cosmetically inconsistent (canonicals point at one URL, deploy is at another) but functionally fine because in-page links are relative.

## Behavior

- `forks/amsterdam` (default): `SITE_URL=https://steel.ethereum.foundation/docs/execution-specs/`.
- `mainnet`, other allowlisted branches: `SITE_URL=https://steel.ethereum.foundation/docs/execution-specs/<branch>/` (unchanged).
- Pull requests: `SITE_URL=https://example.com/docs/<branch>/` (unchanged).
- The `check-should-publish` job validates that `default_branch` is present in `branches[]` and fails fast on misconfiguration.

### DIFF
--- a/.github/configs/docs-branches.yaml
+++ b/.github/configs/docs-branches.yaml
@@ -5,14 +5,19 @@
 # `docs-config.yml`; a branch that publishes here but isn't listed there
 # is dispatched but dropped by the aggregator.
 #
+# `default_branch` names the branch whose docs are deployed at the root
+# (steel.ethereum.foundation/docs/execution-specs/) instead of a
+# branch-nested path. It must also appear in `branches[]` below and must
+# match steel-website's `default_branch` setting in `docs-config.yml`.
+#
 # Each entry defines:
 #   - path: The branch name (must match the Git branch exactly)
 #   - label: Human-readable label for the version switcher UI
+
+default_branch: forks/amsterdam
 
 branches:
   - path: "mainnet"
     label: "Mainnet (BPO2)"
   - path: "forks/amsterdam"
     label: "Amsterdam"
-  - path: "devnets/bal/4"
-    label: "bal-devnet-4"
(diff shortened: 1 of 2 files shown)


kclowes (approved):
I tested locally and this all looks good to me!

PR #2974, opened 2026-06-10
feat(consume): send slot_number in build-block payload attributes

edg-l:
`get_payload_attributes()` never sets `slot_number`, so the build-block simulator always sends slot 0 to `testing_buildBlockV1`. Blocks whose expected slot is non-zero can't be built, and SLOTNUM (EIP-7843) only ever gets exercised returning 0.

The expected payload already has the slot, so just forward it. The field is already on `PayloadAttributes` (optional, serializes as `slotNumber`), so no other changes needed.

Noticed while running build-block against ethrex on the Amsterdam BAL fixtures.


marioevz (approved):
LGTM, thanks for catching this!

PR #2975, opened 2026-06-10
fix(ci,tooling): include json-loader in coverage reports

SamWilsn:
## 🗒️ Description

Coverage wasn't being collected for json_loader, so files like `genesis.py` were showing as 0%.

#### Cute Animal Picture

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -163,10 +163,17 @@ json-loader *args:
         --output="tests/json_loader/fixtures" \
         --cov-config=pyproject.toml \
         --cov=ethereum \
-        --cov-fail-under=80
+        --cov-branch \
+        --cov-report=term \
+        --cov-fail-under=85
     uv run pytest \
         -m "not slow" \
         -n auto --maxprocesses 6 --dist=loadfile \
+        --cov-config=pyproject.toml \
+        --cov=ethereum \
+        --cov-branch \
+        --cov-report=term \
+        --cov-report "xml:{{ output_dir }}/json-loader/coverage.xml" \
         --basetemp="{{ output_dir }}/json-loader/tmp" \
         "$@" \
         tests/json_loader
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -145,6 +145,12 @@ jobs:
         run: just json-loader
         env:
           PYTEST_XDIST_AUTO_NUM_WORKERS: auto
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
+        with:
+          files: .just/json-loader/coverage.xml
+          flags: unittests
+          token: ${{ secrets.CODECOV_TOKEN }}
 
   test-tests:
     runs-on: [self-hosted-ghr, size-xl-x64]


Justfile
```diff
@@ -163,10 +163,17 @@ json-loader *args:
         --output="tests/json_loader/fixtures" \
         --cov-config=pyproject.toml \
         --cov=ethereum \
+        --cov-branch \
+        --cov-report=term \
         --cov-fail-under=85
```
danceratopz, line 168: This comment is orthogonal to the main discussion, but... 

This one confused me a little initially as it's not applied anywhere else. I guess this was the main historical codecov gate prior to The Weld.

From the [codecov output](https://github.com/ethereum/execution-specs/actions/runs/27294915337/job/80625141610#step:5:815): 
```
Required test coverage of 85% reached. Total coverage: 87.53%
```
this would trigger a fail at that step in CI. As the tests marked with `eels_base_coverage` are out of sync with the current dev fork (unless we start updating this set regularly using minimize-tests), this could at some point trigger and would have little meaning as this test set doesn't cover new features.

danceratopz (commented):
Hey Sam, the changes look good, but, assuming it's the `90.53%` number (below is copied from the [codecov bot comment](https://github.com/ethereum/execution-specs/pull/2975#issuecomment-4672952286)) you want to improve, we haven't achieved that goal yet.
```
@@                 Coverage Diff                 @@
##           forks/amsterdam    #2975      +/-   ##
===================================================
- Coverage            90.53%   88.81%   -1.72%     
===================================================
  Files                  535      620      +85     
  Lines                32893    38706    +5813     
  Branches              3021     3310     +289     
===================================================
+ Hits                 29780    34378    +4598     
- Misses                2595     3520     +925     
- Partials               518      808     +290     
```

This is because `./tests/json_loader/` imports more modules that weren't previously imported by running `just fill`. So although we gain coverage for `src/ethereum/genesis.py`, we lose coverage absolutely due to these new imports.

Looking into it, the coverage report now newly includes `./src/ethereum/forks/spurious_dragon/` and `./src/ethereum/forks/tangerine_whistle/` which leads to the coverage drop. `json_loader` imports these forks, but `fill` doesn't: These forks are currently configured to be ignored by `fill` (`ignore=True`), so didn't previously get imported or contribute to the coverage report:
https://github.com/ethereum/execution-specs/blob/26ff8a649e048cb3d77c24c9646dde491aab3108/packages/testing/src/execution_testing/forks/forks/forks.py#L1341-L1348

Wdyt? Should we configure codecov to ignore these forks, too? That seems reasonable to me. We could add them here:
https://github.com/ethereum/execution-specs/blob/26ff8a649e048cb3d77c24c9646dde491aab3108/pyproject.toml#L299-L304

Justfile
danceratopz: It's out of scope for this PR, but perhaps we can re-think where and how we catch coverage regressions, but this doesn't seem like the best place.

SamWilsn: I believe @gurukamath added the minimum coverage there to force us to rerun minimize-tests if coverage drops too low.

---

Python code coverage remains a mystery to me... Not importing a module should definitely count as not executing it 🤣

If those forks aren't executed, they should count as not executed for code coverage purposes. Unlike the DAO, glacier, and BPO forks, Spurious Dragon and Tangerine Whistle have meaningful changes.

danceratopz: No import, no problem! :laughing: Ok, yeah that doesn't actually make any sense. I dug a bit deeper...

We use `--cov=ethereum`, so pytest-cov should provide a coverage report based on all sub-packages, but as `src/ethereum/forks` is a PEP 420 namespace package (and has no `__init__.py`) and it turns out Coverage.py silently [skips namespace packages by default](https://coverage.readthedocs.io/en/7.14.1/config.html#report-include-namespace-packages). I fixed up our configuration for that as a targeted PR to see what our coverage lands at: https://github.com/ethereum/execution-specs/pull/2982

If you merge, #2982 first, then coverage will get worse and your PR can be the hero that improves it again. But it'll likely still be much worse than the `90.53%` as of the [report above](https://github.com/ethereum/execution-specs/pull/2975#issuecomment-4672952286).

To improve on that (while adding coverage for these two forks), I think we'd need to create a follow-up PR that:
 
1. Modifies `fill` to allow optional inclusion via flag of `SpuriousDragon` and `TangerineWhistle`, but don't change the current default behavior.
2. Uses this flag in the preliminary fill step in `just json-loader` to create fixtures with `eels_base_coverage` for these forks -> `pytest tests/json_loader` will use them.
3. -> Coverage back up.

This isn't perfect but should do the trick. In the future, we can also include them as part of minimize-tests? 

Then again, depending on minimize-tests to increase coverage of `src/ethereum/forks` seems a bit of a contradiction...The full solution would be to run them as part of `just fill` them in CI, but as we've never releases fixtures for these tests :upside_down_face: and no one ever complained, this seemingly unnecessarily adds more load to CI? 

Thoughts? :smile:

danceratopz: lol, #2982 is worse than i thought and [failed the 85% coverage gate](https://github.com/ethereum/execution-specs/actions/runs/27359246896/job/80842102260#step:5:902):
```
FAIL Required test coverage of 85% not reached. Total coverage: 81.54%
```

danceratopz: Btw, I applied #2982's change on top of this branch locally and only achieved 78.34% :sob:

danceratopz: Perhaps we should do the [suggestion above](https://github.com/ethereum/execution-specs/pull/2975#discussion_r3397372631) to allow filling Tangerine Whistle and Spurious Dragon (for the json-loader tests) as a preliminary PR and then tackle this. But let me know what you think about the json-loader vs full fill approach.

SamWilsn: `fill` is the entirety of EELS' test coverage. Nothing should be disabled. If we refactor Tangerine Whistle or Spurious Dragon (which we do quite often), we need to test those changes!

Can we enable _all_ forks for `just fill`?

danceratopz: - https://github.com/ethereum/execution-specs/pull/2992

danceratopz: We're now at 93.02% :guitar: 

https://app.codecov.io/gh/ethereum/execution-specs/pull/2975?dropdown=coverage

danceratopz (approved):
We're so back
image
from https://app.codecov.io/gh/ethereum/execution-specs/pull/2975?dropdown=coverage

PR #2986, opened 2026-06-14
feat(test-benchmark): add invalid jump destination edge case benchmarks

Mgabal:
## What this PR does

Adds benchmark tests for JUMP and JUMPI instructions when targeting 
invalid destinations (positions that are not JUMPDEST opcodes).

## Why

The existing benchmark tests cover the happy path for JUMP and JUMPI 
but do not exercise the InvalidJumpDestError path. These tests ensure 
the error handling branch is benchmarked and verified under the 
execution spec.

## Tests added

- `test_jump_to_invalid_destination` — JUMP targeting a non-JUMPDEST position
- `test_jumpi_to_invalid_destination` — JUMPI with true condition targeting a non-JUMPDEST position

### DIFF
--- a/tests/benchmark/compute/instruction/test_control_flow.py
+++ b/tests/benchmark/compute/instruction/test_control_flow.py
@@ -115,3 +115,56 @@ def test_jumpdests(
         target_opcode=Op.JUMPDEST,
         code_generator=JumpLoopGenerator(attack_block=Op.JUMPDEST),
     )
+
+
+def test_jump_to_invalid_destination(
+    benchmark_test: BenchmarkTestFiller,
+    pre: Alloc,
+) -> None:
+    """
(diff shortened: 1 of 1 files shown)


LouisTsai-Csie (changes requested):
I don't think this is the worst case. Two reasons:

- The program counter is too small (0x03)
- For every run, the transaction reverts immediately after the tx executes, the intrinsic cost overhead is too high.
- The destination stays the same for every run, not sure if there is cache issues (Highly unlikely but worth verifying)

For worst-case scenarios, i would suggest investigate how clients handle `JUMPDEST` analysis. Reference `test_jumpdest_analysis` and add variants for invalid scenarios.

spencer-tb:
Closing.

PR #2987, opened 2026-06-15
feat(tests, spec-specs): raise EIP-7954 max code size to 64KiB

spencer-tb:
## 🗒️ Description

Apply the EIP-7954 constant update from ethereum/EIPs#11540.

Required changes:
- **`Initcode` generator**: widen the deploy-code-length `PUSH2` to `PUSH3` once code exceeds 64 KiB-1.
- **`test_max_code_size_fork_transition`**: deploy `parent.max_code_size() + 1` so the code size boundary is exercised via the code deposit check in isolation from the initcode-size limit.
- **`test_cases.md`**: drop the stale `test_max_initcode_size_gas_metering_via_create` row (moved to the EIP-8037 suite), document `test_warm_after_failed_create_over_max_code_size`, and update the transition test row.

## 🔗 Related Issues or PRs
EIP-7954 tracker:
- https://github.com/ethereum/execution-specs/issues/2028

## ✅ Checklist

- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    just static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).

#### Cute Animal Picture

cute animal

Issue #2915, opened 2026-05-26
Test Release Tracker: `glamsterdam-devnet@v6.0.0`


spencer-tb:
### Overview

This tracker is for the first `glamsterdam-devnet@v6.0.0` test release and will be updated async.

- **Target release tag**: `tests-glamsterdam-devnet@v6.0.0`
- **EELS branch**: [`devnets/glamsterdam/6`](https://github.com/ethereum/execution-specs/tree/devnets/glamsterdam/6) 
- **Test release date**: 2026-06-19
- **Devnet launch date**: 2026-06-24 (tbd)

#### Related trackers

- https://github.com/ethereum/execution-specs/issues/3001

New EIP implementations will be merged to their respective `eips/amsterdam/eip-xxxx` branch and then used to create the `glamsterdam-devnet-6` branch from `forks/amsterdam`.

### Scope

The most difficult changes will be the EIP-8037 / 8038 / 2780 integration.

#### New Repricing EIPs

These will be merged to [`eips/amsterdam/eip-2780-8038`](https://github.com/ethereum/execution-specs/compare/forks/amsterdam...eips/amsterdam/eip-2780-8038), and added to [`devnets/glamsterdam/6`](https://github.com/ethereum/execution-specs/compare/forks/amsterdam...devnets/glamsterdam/6) accordingly.

- [x] **[EIP-8038](https://eips.ethereum.org/EIPS/eip-8038) - @danceratopz**
  - Preliminary numbers: ethereum/EIPs#11802
  - EELS PR: #2972
- [x] **[EIP-2780](https://eips.ethereum.org/EIPS/eip-2780) - @gurukamath**
  - Specs/Tests PR: #3017
  - Alignment with 8037 / 8038: ethereum/EIPs#11645

#### New Spec Only EIPs

These will be merged to there own `eips/amsterdam/eip-xxxx` branch and added to [`devnets/glamsterdam/6`](https://github.com/ethereum/execution-specs/compare/forks/amsterdam...devnets/glamsterdam/6) accordingly.
- [x] **[EIP-8246](https://eips.ethereum.org/EIPS/eip-8246) - @LouisTsai-Csie / @Carsons-Eels**
  - [x] EELS Specs/Tests PR: #2842 
- [x] **[EIP-7997](https://eips.ethereum.org/EIPS/eip-7997) - @kclowes**
  - [x] EELS Specs/Tests PR: #2802
[...]

Issue #2028, opened 2026-01-15
EIP-7954 Implementation Tracker: Increase Maximum Contract Size


marioevz:
### Target Fork

**Amsterdam**

### Instructions

- [x] Assign issue to EIP specification and testing owner(s).

> [!IMPORTANT]
> A specifications specialist and a testing specialist should ideally share ownership of the EIP.

- [x] Add the issue to the target fork milestone if applicable (i.e., the EIP is at least in the [CFI stage](https://eips.ethereum.org/EIPS/eip-7723#considered-for-inclusion)).

#### Guidance for Marking Items Complete

An item should only be checked off once the EIP is considered *stable*. In this context, stable means:

- No major issues or ambiguities are still being uncovered in the specification or tests.
- There are no open discussion points awaiting resolution.
- Client implementations have been consistently passing the tests for at least a week.

It is ultimately up to the owners' discretion to decide when an item should be marked as complete, using this guidance as the basis for that decision.

In exceptional cases, an EIP may require changes after some items have been marked complete or even after the entire issue has been completed and closed. This can happen, for example, when significant design optimizations are identified and agreed upon in ACD, or when critical security issues surface and require updates to the specification or tests.

When this occurs, owners should either unmark the relevant checkboxes if the issue is still open, or create a new tracking issue for the modifications if the original issue had already been closed.

### Specification + Testing Status

- [x] Testing complexity assessed and documented.
- [x] Specification implementation merged to `eips/amsterdam/eip-7954` *(skip if the fork branch merge below is already complete)*.
- [x] Specification updates merged to the corresponding `forks/amsterdam` branch.
- [x] EIP updates proposed in case of architectural choices surfaced during implementation.
- [x] Required testing framework modifications implemented.
- [x] Test suite implemented.
- [x] Full code coverage for all changes.
- [x] No regressions or failures in tests from prior forks (including static tests).
- [x] [Testing checklist](https://github.com/ethereum/execution-specs/blob/HEAD/docs/writing_tests/checklist_templates/eip_testing_checklist_template.md) complete.
- [x] Hardening session completed.
- [x] Benchmarking tests written and results documented.
- [x] Ran tests using `execute` to ensure compatibility, and marked specific tests to be skipped when they cannot be executed on live networks.
[...]

PR #2993, opened 2026-06-16
feat(tests): add EIP-7954 jumpdest test past the old code-size limit

danceratopz:
## 🗒️ Description

EIP-7954 raises `MAX_CODE_SIZE` to 64 KiB and the initcode limit to 128 KiB. Before this fork, deployed runtime code could never exceed 24 KiB, so no existing test ever executes a contract at a program counter past the old limit. The EIP-7954 max-size contracts are either deploy-only or run a short prefix that `STOP`s before their `JUMPDEST` padding, and `EXTCODECOPY`/`CODECOPY` only read the code rather than run it.

A client whose jumpdest analysis or code execution is still bounded by the old limit would pass the existing suite yet break on a real 64 KiB contract.

`test_max_code_size_high_jumpdest` deploys a `MAX_CODE_SIZE` contract and jumps near the new limit, far beyond the old 24 KiB code and 48 KiB initcode limits:

- Valid `JUMPDEST`: The jump succeeds and the code runs at the high offset, storing a sentinel.
- Non-`JUMPDEST` byte: The jump is rejected and nothing is stored.

The contract is called through a caller that records the call result, so a correct client is told apart from one that truncated its jumpdest analysis (the valid jump fails) or accepts any high offset (the invalid jump wrongly succeeds).

**Final version**

image
image

## 🔗 Related Issues or PRs

- EIP-7954 tracker: #2028
- Test release tracker: #2915 (`glamsterdam-devnet@v6.0.0`)
- Follow-up to #2987, which raised the EIP-7954 limits to 64 KiB code / 128 KiB initcode.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

cute animal


tests/amsterdam/eip7954_increase_max_contract_size/test_max_code_size.py
```diff
@@ -349,3 +349,69 @@ def test_warm_after_failed_create_over_max_code_size(
...
+    target = pre.deploy_contract(target_code)
+    caller = pre.deploy_contract(
+        Op.SSTORE(0, Op.CALL(gas=Op.GAS, address=target)) + Op.STOP
+    )
+
+    tx = Transaction(
+        sender=pre.fund_eoa(),
+        to=caller,
+        gas_limit=fork.transaction_gas_limit_cap(),
+    )
```
marioevz:
```suggestion
    tx = Transaction(sender=pre.fund_eoa(), to=caller)
```

tests/amsterdam/eip7954_increase_max_contract_size/test_max_code_size.py
```diff
@@ -349,3 +349,69 @@ def test_warm_after_failed_create_over_max_code_size(
...
+    full new code range fails one of the two cases. No existing test
+    executes a contract at a program counter beyond the old limit.
+    """
+    if valid_jumpdest:
+        # Lands on a JUMPDEST and stores a sentinel at the high offset.
+        tail = Op.JUMPDEST + Op.SSTORE(0, 1) + Op.STOP
+    else:
+        # First byte is PUSH1 (0x60), not a JUMPDEST: jumping here is invalid.
+        # A client that wrongly accepts it would execute this and store 2.
+        tail = Op.SSTORE(0, 2) + Op.STOP
```
marioevz:
```suggestion
    if valid_jumpdest:
        tail = Op.JUMPDEST
    else:
        # First byte a STOP, not a JUMPDEST: jumping here is invalid.
        # A client that wrongly accepts it would execute this and store 2.
        tail = Op.STOP
```
Suggestion: Do the SSTORE at the prefix, so there's less code at the end and just a bare JUMPDEST/STOP.

tests/amsterdam/eip7954_increase_max_contract_size/test_max_code_size.py
```diff
@@ -349,3 +349,69 @@ def test_warm_after_failed_create_over_max_code_size(
...
+
+    dest = fork.max_code_size() - len(tail)
+    push_size = (dest.bit_length() + 7) // 8
+    push_op = getattr(Op, f"PUSH{push_size}")
+    prefix = push_op(dest) + Op.JUMP
```
marioevz:
```suggestion
    prefix = Op.SSTORE(0, 1) + push_op(dest) + Op.JUMP
```
Follow up to previous comment.

marioevz (approved):
Looks good, just a couple of suggestions, feel free to apply.

LouisTsai-Csie (approved):
Good points from Mario, LGTM once those are addressed. No other comments from my side.

Follow-up PR idea on `JUMPDEST` analysis: clients run it to ensure a jump target doesn't land on an immediate byte. We could add a scenario where the jump targets a `0x5b` immediate byte of a `PUSH`, `SWAPN`, `DUPN`, or `EXCHANGE` instruction, asserting it's rejected as an invalid jump destination. The contract size should be at the new max code size, so the analysis is done well past the increase.

tests/amsterdam/eip7954_increase_max_contract_size/test_max_code_size.py
danceratopz: Thanks, of course :)

tests/amsterdam/eip7954_increase_max_contract_size/test_max_code_size.py
danceratopz: This is cleaner, thanks!

tests/amsterdam/eip7954_increase_max_contract_size/test_max_code_size.py
danceratopz: Follow-up:
- https://github.com/ethereum/execution-specs/pull/2995

danceratopz:
> Follow-up PR idea on `JUMPDEST` analysis: clients run it to ensure a jump target doesn't land on an immediate byte. We could add a scenario where the jump targets a `0x5b` immediate byte of a `PUSH`, `SWAPN`, `DUPN`, or `EXCHANGE` instruction, asserting it's rejected as an invalid jump destination. The contract size should be at the new max code size, so the analysis is done well past the increase.

Thanks for the review @LouisTsai-Csie! About the idea above, I think it has low additional value, but happy to add if you disagree. I think it would only catch a client that had hard-coded the wrong contract sizes in EIP-8024 context?

| Behavior (the jump target byte) | Whose rule | Already tested in `forks/amsterdam`? |
|---|---|---|
| `0x5B` buried in `PUSH` data → **rejected** | classic JUMPDEST analysis (pre-EIP) | Rule: [runtime.py#L62-L67](https://github.com/ethereum/execution-specs/blob/c00006fb09a3975ded1906fee368d6c4402df553/src/ethereum/forks/amsterdam/vm/runtime.py#L62-L67). No dedicated test in the 8024/7954 suites (ancient behavior). |
| `0x5B` after `DUPN` → **accepted** | EIP-8024 | ✅ [`test_dupn_jump_to_immediate_byte_0x5b_succeeds`](https://github.com/ethereum/execution-specs/blob/c00006fb09a3975ded1906fee368d6c4402df553/tests/amsterdam/eip8024_dupn_swapn_exchange/test_dupn.py#L280-L315) |
| `0x5B` after `SWAPN` → **accepted** | EIP-8024 | ✅ [`test_swapn_jump_to_immediate_byte_0x5b_succeeds`](https://github.com/ethereum/execution-specs/blob/c00006fb09a3975ded1906fee368d6c4402df553/tests/amsterdam/eip8024_dupn_swapn_exchange/test_swapn.py#L356-L391) |
| `0x5B` after `EXCHANGE` → **accepted** | EIP-8024 | ✅ [`test_exchange_jump_to_immediate_byte`](https://github.com/ethereum/execution-specs/blob/c00006fb09a3975ded1906fee368d6c4402df553/tests/amsterdam/eip8024_dupn_swapn_exchange/test_exchange.py#L344-L399) (immediate sweep, incl. `0x5B`) |
| valid immediate skipped → jump **fails** | EIP-8024 | ✅ [`test_dupn_jump_to_valid_immediate_fails`](https://github.com/ethereum/execution-specs/blob/c00006fb09a3975ded1906fee368d6c4402df553/tests/amsterdam/eip8024_dupn_swapn_exchange/test_dupn.py#L316-L350), [`test_swapn_jump_to_valid_immediate_fails`](https://github.com/ethereum/execution-specs/blob/c00006fb09a3975ded1906fee368d6c4402df553/tests/amsterdam/eip8024_dupn_swapn_exchange/test_swapn.py#L392-L426) |
| any of the above **at 64 KiB** (`MAX_CODE_SIZE`) | EIP-7954 (size only) | ❌ Not covered. The 8024 tests above use tiny contracts; the basic max-size jump case is in [#2993](https://github.com/ethereum/execution-specs/pull/2993). |

danceratopz:
@LouisTsai-Csie I added the test case here  https://github.com/ethereum/execution-specs/pull/2998

PR #2994, opened 2026-06-16
refactor(test-forks): Introduce `minimum_block_gas_limit`

marioevz:
## 🗒️ Description

Noticed during the implementation of EIP-8282 that it is clunky to update the minimum block gas limit when a new system contract is introduced due to the Block-level-access-list max element constrain.

This PR introduces a `Fork.minimum_block_gas_limit()` helper so tests no longer hardcode the minimum valid block gas limit, which now varies per fork because of EIP-7928 (Block-Level Access Lists).

Previously the minimum was a fixed `5000`, and EIP-7928's empty-block BAL cost was a single hardcoded `empty_block_bal_item_count() = 15` constant living in `EIP7928`. With EIP-8282 adding two more system contracts, that constant no longer holds, and the value was duplicated across several tests.

This refactor:

- **Adds `minimum_block_gas_limit()`** (abstract in `BaseFork`, implemented in `Frontier`). It returns `max(5000, empty_block_bal_item_count() * BLOCK_ACCESS_LIST_ITEM)`, only applying the BAL-derived floor when `header_bal_hash_required()` is set.
- **Composes `empty_block_bal_item_count()` per EIP** via `super()` chaining instead of a single hardcoded total. Each system-contract EIP contributes its own items: EIP-4788 `+3`, EIP-2935 `+2`, EIP-7002 `+5`, EIP-7251 `+5`, EIP-8282 `+10`. The base returns `0`. This reproduces the previous total (15) for Prague and correctly yields 15 for Amsterdam (25 after EIP-8282).
- **Updates tests** to call `fork.minimum_block_gas_limit()` instead of recomputing or hardcoding the value:
  - `tests/frontier/validation/test_header.py`: `test_gas_limit_below_minimum` → `test_block_gas_limit_below_minimum`, reparametrized via `parametrize_by_fork` so the boundary cases (`minimum_minus_one`, `minimum`) track the active fork. Gas limits below the protocol floor (`5000`) expect `INVALID_GASLIMIT`; on EIP-7928 forks they accept either `INVALID_GASLIMIT` or `BLOCK_ACCESS_LIST_GAS_LIMIT_EXCEEDED`, since a client may reject on the header gas-limit floor before reaching the BAL check.
  - `tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py`: use the helper for the size-constraint boundary.
  - `tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py`: scale the number of filler txs so the block gas limit meets the fork minimum while preserving the off-by-one boundary check.
  - `tests/ported_static/stStaticCall/test_static_internal_call_hitting_gas_limit2.py`: drop the `Environment` `gas_limit=47766`, which would fall below the Amsterdam minimum and invalidate the block; tagged `@manually-enhanced` so the porter won't re-add it.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/forks/forks/forks.py
+++ b/packages/testing/src/execution_testing/forks/forks/forks.py
@@ -324,6 +324,22 @@ class Frontier(
         return wrapper
 
     @classmethod
+    def minimum_block_gas_limit(cls) -> int:
+        """
+        Return the minimum gas limit for the block to be considered valid.
+        """
+        minimum_block_gas_limit = 5_000
+        bal_minimum_block_gas_limit = 0
+        if cls.header_bal_hash_required():
+            # The block gas limit is influenced by the minimum amount of
+            # block level access elements the system contracts contain.
+            bal_minimum_block_gas_limit = (
+                cls.empty_block_bal_item_count()
+                * cls.gas_costs().BLOCK_ACCESS_LIST_ITEM
+            )
+        return max(minimum_block_gas_limit, bal_minimum_block_gas_limit)
+
+    @classmethod
     def opcode_gas_map(
         cls,
     ) -> Dict[OpcodeBase, int | Callable[[OpcodeBase], int]]:
--- a/packages/testing/src/execution_testing/forks/base_fork.py
+++ b/packages/testing/src/execution_testing/forks/base_fork.py
@@ -457,6 +457,12 @@ class BaseFork(ForkOpcodeInterface, metaclass=BaseForkMeta):
 
     @classmethod
     @abstractmethod
+    def minimum_block_gas_limit(cls) -> int:
+        """Return the minimum block gas limit for it to be considered valid."""
+        pass
+
+    @classmethod
+    @abstractmethod
     def opcode_gas_map(
         cls,
     ) -> Dict[OpcodeBase, int | Callable[[OpcodeBase], int]]:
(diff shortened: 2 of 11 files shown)


packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -323,6 +323,22 @@ def wrapper(opcode: OpcodeBase) -> int:
...
+        minimum_block_gas_limit = 5_000
+        bal_minimum_block_gas_limit = 0
+        if cls.header_bal_hash_required():
+            # The block gas limit is influenced by the minimum amount of
+            # block level access elements the system contracts contain.
+            bal_minimum_block_gas_limit = (
+                cls.empty_block_bal_item_count()
+                * cls.gas_costs().BLOCK_ACCESS_LIST_ITEM
+            )
+        return max(minimum_block_gas_limit, bal_minimum_block_gas_limit)
```
kclowes, line 340: I am probably missing something but I think this could just return `minimum_block_gas_limit`?

kclowes (approved):
LGTM! I left one small comment on the new `minimum_block_gas_limit` classmethod, but not blocking 🚀

packages/testing/src/execution_testing/forks/forks/forks.py
marioevz: `minimum_block_gas_limit` is not overwritten by the `if` code path, so we still need to compare and return the maximum of the two.

PR #2996, opened 2026-06-17
chore(tests): reduce Amsterdam ported_static skip list (480 to 258)

leolara:
## 🗒️ Description

Reduce the Amsterdam `tests/ported_static/` skip list from 480 to 258
entries. PR #2969 made the transaction gas limit automatic, adding a
state-gas reservoir on EIP-8037 forks, which fixes a large class of
these tests directly.

Every change here is independent of the exact EIP-8037 gas constants. It
either removes a skip entry or unsets an incidental transaction
`gas_limit`, so the test rides the implicit limit and the reservoir from
#2969. The reduction comes in four parts:

1. **Unskip cases that now pass (33).** After #2969, tests that leave the
   transaction gas limit unset get the implicit limit plus reservoir.

2. **Drop obsolete entries (10).** Size-limit tests deselected by
   EIP-7954, and stale parametrization ids.

3. **Unset incidental tx gas_limit (116).** An explicit gas_limit that
   was incidental: enough on older forks, short on Amsterdam.

4. **Fork-conditional per-slot unset (63).** Mixed files where only some
   gas parametrizations broke; unset the affected slot on EIP-8037 only.

The gas-measurement, `warm_after_fail`, and refund re-pins, which pin
values derived from the current EIP-8037 constants, are split out to
`leolara/fix-ported-static-amsterdam-repins` for a follow-up PR once the
constants settle. Keeping them out of this PR makes it stable against
another recalibration.

Every change was filled across all valid forks and passes strict mypy
and ruff.

## 🔗 Related Issues or PRs

Follows #2969.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

cute animal


leolara:
## Progress summary (updated)

This PR is now scoped to the changes that are independent of the exact EIP-8037 gas constants. Each one is either an unskip or an incidental-`gas_limit` unset, so the test rides the automatic transaction gas limit and the state-gas reservoir from #2969. None pins a constant-derived value, so a future recalibration cannot break this PR. Current state: skip list **480 to 258** (222 cleared), filled across all valid forks.

### Cleared, by approach

| Approach | Entries |
|---|---:|
| Now pass after #2969 (no edit) | 33 |
| Obsolete entries removed | 10 |
| Unset incidental transaction gas_limit | 116 |
| Fork-conditional per-slot gas unset | 63 |
| **Total** | **222** |

### Split out to a follow-up

The gas-measurement, `warm_after_fail`, and refund re-pins (the further reduction from 258 to 229 in the earlier version of this PR) are moved to `leolara/fix-ported-static-amsterdam-repins`. They pin values derived from the current EIP-8037 constants, so they belong in a separate PR filled once the constants settle.

### Still skipped (258)

Dominated by:
- `multi_param_per_d` — the MCOPY sweep and the `point_mul` precompile sweeps, whose OoG boundary needs a per-`d` recompute.
- gas-measurement and refund tests whose stored value or whole transaction shifts under EIP-8037.
- `oog_by_design`, `bytecode_baked`, `other`.

Reviving these needs a gas-budget change (not the same scenario), a per-`d` boundary recompute, or per-frame state-gas work.

tests/ported_static/stCreateTest/test_create_address_warm_after_fail.py
```diff
@@ -418,9 +418,9 @@ def test_create_address_warm_after_fail(
                         3: 1,
                         4: 1,
                         5: 1,
-                        12: 32028,
+                        12: (7028 if fork.is_eip_enabled(8037) else 32028),
```
marioevz: This doesn't look like a permanent solution IMO – we are going have to come back on the next gas repricing to re-disable it.

I would rather only re-enable these tests if we are doing dynamic recalculation of these values via a rewrite.

marioevz (changes requested):
I think this PR does a good job in pointing out which tests try to store gas at a certain point during execution, but the proposed solution is to simply introduce a new constant instead of fixing the issue and making it future proof.

I think the steps forward should be:
1) We keep the changes in this PR that simply remove the `gas_limit` from the transaction, since those effectively remove constants and make the tests more resilient.
2) Move the changes that introduce new constants into a new branch/issue so we can start tackling them one by one via a proper update to the test.

For (2) I think we need to create a claude skill and this should be an agentic process.

tests/ported_static/stCreateTest/test_create_address_warm_after_fail.py
leolara: I understand, I will change this to only send the tests that don't need specific repricing.

Then, perhaps I can think of a way to abstract this out, like to say declaratively what we want to happen with the gas and that gets automatically calculated, I don't know how difficult is that.

marioevz (approved):
LGTM, I ported the failing tests to the newer python ECADD+ECMUL framework, so we can proceed to merge. Thanks!

PR #3000, opened 2026-06-17
fix(specs): check chain id

SamWilsn:
## 🗒️ Description

EELS didn't check incoming transaction's chain id

#### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/transactions.py
+++ b/src/ethereum/forks/amsterdam/transactions.py
@@ -735,7 +735,25 @@ def count_tokens_in_data(data: bytes) -> Uint:
     return num_zeros + num_non_zeros * Uint(4)
 
 
-def recover_sender(chain_id: U64, tx: Transaction) -> Address:
+def chain_id(tx: Transaction) -> None | U64:
+    """
+    Extract the chain identifier from a transaction. See [EIP-155].
+
+    [EIP-155]: https://eips.ethereum.org/EIPS/eip-155
+    """
+    if isinstance(tx, LegacyTransaction):
+        if tx.v == 27 or tx.v == 28:
+            return None
+
+        if tx.v < U256(35):
+            raise InvalidSignatureError("bad v")
+
+        return U64((tx.v - U256(35)) >> U256(1))
+    else:
+        return tx.chain_id
+
+
+def recover_sender(tx: Transaction) -> Address:
     """
     Extracts the sender address from a transaction.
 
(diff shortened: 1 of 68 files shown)


src/ethereum/forks/arrow_glacier/transactions.py
```diff
@@ -376,7 +376,21 @@ def calculate_intrinsic_cost(tx: Transaction) -> Uint:
...
+    Extract the chain identifier from a transaction. See [EIP-155].
+
+    [EIP-155]: https://eips.ethereum.org/EIPS/eip-155
+    """
+    if isinstance(tx, LegacyTransaction):
+        if tx.v == 27 or tx.v == 28:
+            return None
+        return U64((tx.v - U256(35)) >> U256(1))
+    else:
+        return tx.chain_id
```
LouisTsai-Csie, line 394:
```suggestion
    if isinstance(tx, LegacyTransaction):
        if tx.v == 27 or tx.v == 28:
            return None
        return U64((tx.v - U256(35)) >> U256(1))
    return tx.chain_id
```

Not very important suggestion.

src/ethereum/forks/berlin/transactions.py
```diff
@@ -319,14 +333,15 @@ def recover_sender(chain_id: U64, tx: Transaction) -> Address:
                 r, s, v - U256(27), signing_hash_pre155(tx)
             )
         else:
-            chain_id_x2 = U256(chain_id) * U256(2)
-            if v != U256(35) + chain_id_x2 and v != U256(36) + chain_id_x2:
+            tx_chain_id = U64((v - U256(35)) >> U256(1))
+            v = (v - U256(35)) & U256(1)
+            if v != 0 and v != 1:
```
LouisTsai-Csie: Will this always be false? since `v` masks with `U256(1)`, so it would always be either 0 or 1.

LouisTsai-Csie (dismissed):
Leave one question, thank you

src/ethereum/forks/berlin/transactions.py
SamWilsn: good catch!

src/ethereum/forks/amsterdam/transactions.py
```diff
@@ -735,7 +735,21 @@ def count_tokens_in_data(data: bytes) -> Uint:
...
+def chain_id(tx: Transaction) -> None | U64:
+    """
+    Extract the chain identifier from a transaction. See [EIP-155].
+
+    [EIP-155]: https://eips.ethereum.org/EIPS/eip-155
+    """
+    if isinstance(tx, LegacyTransaction):
+        if tx.v == 27 or tx.v == 28:
+            return None
+        return U64((tx.v - U256(35)) >> U256(1))
```
marioevz, line 751: This test generates an overflow error in this line:
```python
@pytest.mark.valid_from("Istanbul")
@pytest.mark.exception_test
def test_chainid_invalid_v(
    state_test: StateTestFiller,
    pre: Alloc,
    chain_config: ChainConfig,
) -> None:
    """Test CHAINID opcode."""
    chain_id = chain_config.chain_id
    contract_address = pre.deploy_contract(Op.SSTORE(1, Op.CHAINID) + Op.STOP)

    tx = Transaction(
        ty=0,
        chain_id=chain_id,
        to=contract_address,
        v=34,
        r=1,
        s=1,
        sender=Address(0x0),
        error=TransactionException.INVALID_SIGNATURE_VRS,
    )

    post = {
        contract_address: Account(storage={1: chain_id}),
    }

    state_test(pre=pre, post=post, tx=tx)
```
We still need to check that the `v` value is equal to 35 or greater. Same for the rest of the forks I believe.

marioevz (commented):
Just one comment I think needs addressing. Thanks!

marioevz (approved):
LGTM, thanks!

PR #3004, opened 2026-06-17
chore(ci): update mypy

SamWilsn:
## 🗒️ Description

Grab a newer mypy version and enable more checks.

#### Cute Animal Picture


Justfile
```diff
@@ -130,7 +130,7 @@ fill *args:
 [group('integration tests')]
 fill-pypy *args:
     @mkdir -p "{{ output_dir }}/fill-pypy/tmp" "{{ output_dir }}/fill-pypy/logs"
-    uv run --python pypy3.11 fill \
+    uv run --python pypy3.11 --no-dev --group test fill \
```
marioevz, line 133: Seems like this can also be applied to cPython just recipe.

Justfile
```diff
@@ -197,7 +197,7 @@ test-tests *args:
 [group('unit tests')]
 test-tests-pypy *args:
     @mkdir -p "{{ output_dir }}/test-tests-pypy/tmp"
-    cd packages/testing && uv run --python pypy3.11 pytest \
+    cd packages/testing && uv run --python pypy3.11 --no-dev --group test pytest \
```
marioevz, line 200: Same as previous comment.

marioevz (approved):
LGTM, just left one suggestion. Thanks!

Justfile
danceratopz: Sam, I was indeed wondering why `pypy3.11` got singled out? I wondered if mypy had dropped 3.11 support in 2.1.0, but that's not the case. Either way, it's def cleaner to drop `test` deps here.

Justfile
danceratopz: I think this only works because our exotic testing package explicitly depends on `pytest` btw (`fill` needs `pytest`).

danceratopz (approved):
Thanks for this! The source changes it caught are nice. I was wondering if it's also quicker?

SamWilsn:
There is no pre-compiled wheel for mypy on pypy. Up until now, we were able to compile the Rust components of the crate using the rustc on the runner image, but with mypy 2.0 that is no longer the case.

We don't typecheck on pypy, so it is safe to drop from that interpreter.

Issue #2314, opened 2026-02-25
py3 CI job timing regression (~18min → ~38min) after PR #2252


danceratopz:
## Summary

The `py3` CI job doubled in duration from ~18 minutes to ~38 minutes after PR #2252 merged on Feb 23. The regression is isolated to the "Run py3 tests" step (16m42s → 35m37s) — queue and setup times are unchanged.

**Root cause**: Expensive calibration code in `eip7934_block_rlp_limit/test_max_block_rlp_size.py`, amplified by coverage instrumentation overhead (`sys.settrace` in Python 3.12).

## What changed

PR #2252 (commit `916d723`) added several computationally intensive functions to `tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py`:

### 1. Binary search allocating large byte strings (~20 iterations per call)

[`max_zero_calldata_bytes_for_gas()`](https://github.com/ethereum/execution-specs/blob/916d723/tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py#L71-L117) performs a binary search calling `calculator(calldata=b"\x00" * N)` approximately 20 times per invocation, allocating progressively larger byte strings each iteration.

### 2. Calibration wrapper calling the binary search

[`calibrated_block_gas_limit()`](https://github.com/ethereum/execution-specs/blob/916d723/tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py#L119-L166) calls the above binary search as part of its calibration logic.

### 3. Autouse fixture running for EVERY test in the module

[`calibrate_gas_limit`](https://github.com/ethereum/execution-specs/blob/916d723/tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py#L168-L182) is an `autouse=True` fixture, meaning the calibration runs before every single test in the module — not once per session or module.

### 4. Rewritten transaction builder with 6x RLP-encoding per call

[`_exact_size_transactions_impl()`](https://github.com/ethereum/execution-specs/blob/916d723/tests/osaka/eip7934_block_rlp_limit/test_max_block_rlp_size.py#L351-L610) was rewritten to call `get_block_rlp_size()` six times per invocation, each constructing a `FixtureHeader` and RLP-encoding a full block.

The `@lru_cache` on `_exact_size_transactions_cached` only helps when arguments are identical — tests with withdrawals or specific transactions bypass the cache entirely.

## Why coverage amplifies the cost

[...]

PR #3010, opened 2026-06-18
chore(ci,tooling): speed up json-loader job

danceratopz:
## 🗒️ Description

The `json-loader` job takes ~23 min. This PR applies two small quick wins to reduce this:

- Run the job on Python 3.14, matching `fill`, so `coverage` uses `sys.monitoring` instead of the slower C tracer. Coverage was enabled for `json-loader` in #2975, but the job still runs on the default 3.13, reintroducing the C-tracer overhead that #2310 removed elsewhere (see the regression tracked in #2314).
- Run the fixture-consuming `pytest` phase with the same worker count as the fill phase (`-n {{ xdist_workers }}`, which is `auto` in CI). It was pinned to `--maxprocesses 6`, capping it at 6 workers instead of scaling with the runner. The cap carried over from the `tox` configuration in #2555.

## 🔗 Related Issues or PRs

- #2975
- #2310
- #2314

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

:cat:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -140,6 +140,8 @@ jobs:
         with:
           submodules: recursive
       - uses: ./.github/actions/setup-uv
+        with:
+          python-version: "3.14"
       - uses: ./.github/actions/setup-env
       - name: Fill and run json-loader tests
         run: just json-loader
--- a/Justfile
+++ b/Justfile
@@ -165,6 +165,7 @@ json-loader *args:
         --cov=ethereum \
         --cov-branch \
         --cov-report=term \
+        --durations=50 \
         --cov-fail-under=85
     uv run pytest \
         -m "not slow" \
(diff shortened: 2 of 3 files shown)


SamWilsn (approved):

danceratopz:
Hey @SamWilsn, thanks for the review here. I reverted the attempts to speed-up the json-loader job by splitting up the `test_tools_new_fork.py::test_end_to_end` test across different workers. This didn't seem to be working as intended. I think we can leave this as for a follow-up and get the other quick wins in for speed-up.

The `new_fork` tests take ~225s each and it looks like they get placed on one xdist worker, leading to poorly-balanced workers/starvation. As these tests are unit tests for the new fork tool and have nothing todo with JSON, I'd consider moving these out of the json-loader tests; there might be other candidates, will have to check. This would make it easier to split them and is cleaner.

danceratopz:
@SamWilsn added an issue here: 
- https://github.com/ethereum/execution-specs/issues/3022

PR #3011, opened 2026-06-18
feat(tests): EIP-8037 state-gas refund on failed CREATE2 with init storage

bshastry:
## 🗒️ Description

A `CREATE2` whose init code creates new storage slots and then **fails its code deposit** — either because the deposited code is rejected by EIP-3541, or because the EIP-8037 code-deposit state gas cannot be paid — reverts the create frame. On that revert, *all* of the init's state-creation gas must be refunded: the new account **and** every storage slot the init wrote.

Existing `eip8037` coverage exercises the create-init storage charge on the *success* path (`test_state_gas_create.py`) and same-tx slot-reset refunds (`test_state_gas_sstore.py`), but not the refund of **storage-slot** state gas on a create *failure*. Because the new-account state-gas refund on failure is already correct, a failing create with no init storage agrees across clients — only "failing create + init storage" exposes the gap.

This adds a parametric `state_test` over:

- `fail_mode`: `eip3541` (init returns `0xEF`-prefixed code) and `oog_deposit` (init returns max-size code whose deposit cost cannot be paid);
- `slots`: `0`, `1`, `3` init storage slots.

`slots=0` is a negative control (account-creation refund only) that must not diverge; `slots>=1` isolates the storage-slot state-gas refund and scales the discrepancy with the slot count. A post-mortem entry is included per the missed-test-case process. The scenario was surfaced by goevmlab differential fuzzing on Amsterdam (bal-devnet-7).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

- [x] All: Ran the relevant static checks on the new test (`ruff format`, `ruff check`, `codespell`, `mypy`) — all pass.
- [x] All: PR title adheres to the repo standard — `feat(tests): ...`.
- [ ] All: Considered updating the online docs in ./docs/.
- [ ] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs in the Test Case Reference are correctly formatted.
- [x] Tests: For PRs implementing a missed test case, updated the post-mortem document (`docs/writing_tests/post_mortems.md`).
- [ ] Ported Tests: All converted JSON/YML tests have been assigned `@ported_from` (N/A — new test, not ported).

#### Cute Animal Picture

cute animal


bshastry:
### Client root-cause note (go-ethereum @ bal-devnet-7) — for the geth team; not part of this spec change

The reference spec (EELS `forks/amsterdam` @ `5f8c109f7`) and revm / nethermind / besu / erigon / ethrex all agree on these fixtures; **go-ethereum alone diverges** on the `slots>=1` cases (post-state-root mismatch) and passes `slots=0`. Source pointers (geth `9e04883`, bal-devnet-7):

- The per-op EIP-8037 SSTORE state-creation **charge is correct** — `core/vm/gas_table.go:655-661` (`gasSStore8037`) charges `StorageCreationSize(64) * CostPerStateByte(1530) = 97920` for a new slot (applied at `core/vm/interpreter.go:230`); the symmetric same-tx `0→x→0` reset refund is at `gas_table.go:677-684`. A plain top-level new-slot SSTORE matches all clients, so this is not the bug.
- The divergence is in the **create-frame-failure revert of state gas**: on a CREATE/CREATE2 that fails after its init created storage, geth refunds the new-account state-creation gas but does **not** revert the `UsedStateGas` charged by the init's storage SSTOREs. The tx-level state dimension (`core/state_transition.go:813-861`, `txStateGas = UsedStateGas`) over-counts by `num_slots * 97920`, inflating gas used → sender/coinbase balances → state root.
- Empirical confirmation (raw bal-devnet-7 binaries, gasUsed / post-state root):
  - `slots=0` (account creation only, no init storage): geth == eels == revm. ✓
  - `slots>=1`: geth over-reports gas used by `num_slots * 97920` and diverges. Reproduces on **both** fail modes (EIP-3541 reject *and* code-deposit OOG) — the failure trigger is irrelevant; the bug is the unrefunded storage-slot state gas on the create revert.

The exact line in geth's create-failure / journal-revert path that fails to roll back the init's storage `UsedStateGas` was not pinpointed from static reading alone (the charge and the same-tx reset refund are both correct), so flagging the revert path for the geth team to confirm. **Unreleased Amsterdam / bal-devnet-7 discrepancy, not a mainnet issue.**

spencer-tb (approved):
Thanks! Appreciate the finding, I wrapped the test to an existing file :)

PR #3012, opened 2026-06-18
chore(ci): run benchmark sanity checks on size-xl-x64 runner

danceratopz:
## 🗒️ Description

Run the benchmark `sanity-checks` jobs (`Benchmark Gas Values`, `Fixed Opcode Count CLI`, `Fixed Opcode Count Config`) on the `size-xl-x64` self-hosted runner instead of `ubuntu-latest`.

The standard `ubuntu-latest` runner only exposes 2 cores, so `fill`/`pytest` with `-n auto --maxprocesses 10` was capped at 2 workers, making `Benchmark Gas Values` take roughly 28 minutes. The XL runner has enough cores for `-n auto` to scale toward the existing `--maxprocesses 10` cap, matching the runner already used by `test.yaml`.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -71,6 +71,8 @@ jobs:
         with:
           enable-cache: "false"
 
+      - uses: ./.github/actions/setup-env
+
       - name: Install PyPy for bench-gas json_loader verify
         if: matrix.recipe == 'bench-gas'
         shell: bash
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -35,8 +35,6 @@ jobs:
           submodules: true
 
       - uses: ./.github/actions/setup-uv
-        with:
-          enable-cache: "false"
 
       - uses: ./.github/actions/build-evm-base
         id: evm-builder
@@ -68,8 +66,6 @@ jobs:
           submodules: true
 
       - uses: ./.github/actions/setup-uv
-        with:
-          enable-cache: "false"
 
       - uses: ./.github/actions/setup-env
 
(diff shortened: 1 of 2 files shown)


LouisTsai-Csie (approved):

danceratopz:
Thanks for jumping in with a review @LouisTsai-Csie. It didn't lead to an improvement (which is curious). Will revisit when I get the chance, I think we need https://github.com/ethereum/execution-specs/pull/2693

danceratopz:
> Thanks for jumping in with a review @LouisTsai-Csie. It didn't lead to an improvement (which is curious). Will revisit when I get the chance, I think we need #2693

My friend told me that:

> ... the Phase 2 "slowest durations" has a cliff:
>
> ```
> 465.28s  test_keccak_max_permutations[...blockchain_test...]@t8n-cache-bc5796d4
> 464.94s  test_keccak_max_permutations[...blockchain_test_engine_x...]@t8n-cache-bc5796d4
> 408.27s  test_keccak_max_permutations[...blockchain_test_engine...]@t8n-cache-bc5796d4
>  12.29s  test_point_evaluation_uncachable   ← next slowest
> ```
> 
>
> One test eats ~465s per format; everything else is ≤12s. And note the shared @t8n-cache-bc5796d4 suffix: that's the --dist=loadgroup key, so all three format-variants are pinned to a single xdist worker and run serially → ~1338s (~22 min) on one core. The other 15 cores finish the entire rest of the suite and then sit idle.

We only ignore these test cases so far:
image

An easy quick win would be to mark this test slow:
- https://github.com/ethereum/execution-specs/pull/3058

But tbh, I think there's an easy optimization to be had here that would help releases; will work on that. Perhaps it will even supersede #3058.

danceratopz:
#3057 and #3060 have brought the run times down to ~5 min (for now); let's leave this on the Github runners. We can always bump later if need be.

PR #3005, opened 2026-06-17
feat(spec, spec-tests): 8037 Calldata floor accounting alignment

kclowes:
## 🗒️ Description
Apply changes from EIP-8037 spec change [#11706](https://github.com/ethereum/EIPs/pull/11706/) that don't conflict with [#11807](https://github.com/ethereum/EIPs/pull/11807)

Update EIP-8037 block gas accounting and CALL state-gas refills

Implements two spec clarifications from EIPs PR #11706:
  1. Calldata floor is tx-level only. Removed the redundant block-level application of the EIP-7623 calldata floor. Block regular gas is now tx_gas_used_before_refund - tx_state_gas, matching EIP-8037's "Integration with EIP-7778" formula.
  1. CALL refunds NEW_ACCOUNT on pre-frame failures. When a value-bearing CALL to a non-existent account fails before the child frame (insufficient balance or stack depth limit), the pre-charged NEW_ACCOUNT state gas is now refunded to the reservoir, matching the spec text.

  Tests in eip7778_block_gas_accounting_without_refunds/test_gas_accounting.py updated to drop calldata-floor terms from block expectations and bumped from valid_from("EIP7778") to valid_from("EIP8037") where the new expectations require the integration rules.

## 🔗 Related Issues or PRs

#2915 

## ✅ Checklist

- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    just static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).
- [x] Tests: Ran `mkdocs serve` locally and verified the auto-generated docs for new tests in the [Test Case Reference](https://eest.ethereum.org/main/tests/) are correctly formatted.

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3013, opened 2026-06-18
feat(amsterdam): refund CALL new-account state gas on soft-fail (EIP-8037)

bshastry:
## 🗒️ Description

EIP-8037's prose says the new-account **state gas** charged for a `CALL*`
is refilled to the `state_gas_reservoir` (and `execution_state_gas_used`
decreases) when the operation is unsuccessful before entering the call
frame, and it draws **no distinction** between `CALL*` and
`CREATE`/`CREATE2` ([Specification], "Gas accounting for new accounts"):

> If the respective call frame reverts or halts exceptionally or if the
> operation is unsuccessful before entering the call frame (e.g., due to
> insufficient balance or due to the stack depth), the charged state-gas
> is refilled back to the `state_gas_reservoir` and
> `execution_state_gas_used` decreases by the same amount.

The executable spec only honored this for `CREATE`/`CREATE2`:
`generic_create` calls `credit_state_gas_refund(evm, NEW_ACCOUNT)` on
each early-fail path (depth / balance / nonce / collision / child-revert).
The `CALL` paths charged `NEW_ACCOUNT` (`call()`, before the balance
check) but on the soft-fail restored only the child reservoir — the
new-account charge was **retained**, contrary to the prose.

This PR aligns `call()` with the prose:

- capture the charged amount in `call()` and refund it on the
  insufficient-balance soft-fail (symmetric with `CREATE`);
- thread it through `GenericCall` so the stack-depth soft-fail in
  `generic_call()` refunds it too.

`CALLCODE`/`DELEGATECALL`/`STATICCALL` never charge new-account state gas
(self/alive target, or no value), so they are unaffected —
`new_account_state_gas` defaults to `0`, making the refund a no-op.

**Tests:** `test_call_insufficient_balance_returns_reservoir` is updated
(the fresh-target reservoir no longer needs the extra new-account
allowance; the docstring previously claimed "the call fails before any
state gas is charged", which contradicted both the code and its own
reservoir sizing). A new `test_call_softfail_refund_reused_by_new_account_call`
pins the `CALL`/`CREATE` symmetry by reusing the refunded reservoir for a
subsequent account-creating `CALL`.

### ⚠️ Note for reviewers — this intentionally diverges from current clients

This is a **prose-alignment** change. As of `bal-devnet-7`, geth
(`9e04883a`), revm (`ab5e3a4c`), besu, and nethermind all **retain** the
charge on a value-`CALL` soft-fail (verified: identical `gasUsed 213,920`
/ post-state root for the minimal case), matching the *previous*
executable-spec behavior, not the prose. So this PR makes EELS follow the
EIP text and, for now, diverge from the clients.

Either the prose is authoritative (this PR; clients would then need to
follow) or the prior behavior is intended (and EIP-8037's text should
instead be amended to state the `CALL*`-vs-`CREATE` asymmetry explicitly,
consistent with legacy `CALL` value-failure semantics). Flagging so the
text and the executable spec say the same thing before activation —
happy to invert this into a test-only PR that keeps current behavior if
the authors decide the prose should change instead.

Scope is `forks/amsterdam` only (the sole fork with EIP-8037).

[Specification]: https://eips.ethereum.org/EIPS/eip-8037

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

cute animal

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -325,6 +325,12 @@ class GenericCall:
     memory_output_size: U256
     code: Bytes
     disable_precompiles: bool
+    new_account_state_gas: Uint = Uint(0)
+    """
+    EIP-8037 new-account state gas charged before the frame, refunded to
+    the reservoir if the call fails the stack-depth check. Non-zero only
+    for a value-bearing `CALL` to a non-existent account.
+    """
 
 
 def generic_call(evm: Evm, params: GenericCall) -> None:
@@ -338,6 +344,9 @@ def generic_call(evm: Evm, params: GenericCall) -> None:
     if evm.message.depth + Uint(1) > STACK_DEPTH_LIMIT:
         evm.gas_left += params.gas
         evm.state_gas_left += params.state_gas_reservoir
+        # EIP-8037: refund new-account state gas charged before the frame
+        # when the stack-depth check fails (symmetric with CREATE).
+        credit_state_gas_refund(evm, params.new_account_state_gas)
         push(evm.stack, U256(0))
         return
 
@@ -457,8 +466,10 @@ def call(evm: Evm) -> None:
     code = get_code(tx_state, code_hash)
 
     charge_gas(evm, extra_gas + extend_memory.cost)
+    new_account_state_gas = Uint(0)
     if value != 0 and not is_account_alive(tx_state, to):
-        charge_state_gas(evm, StateGasCosts.NEW_ACCOUNT)
+        new_account_state_gas = StateGasCosts.NEW_ACCOUNT
+        charge_state_gas(evm, new_account_state_gas)
 
     message_call_gas = calculate_message_call_gas(
         value,
@@ -483,6 +494,10 @@ def call(evm: Evm) -> None:
         evm.return_data = b""
         evm.gas_left += message_call_gas.sub_call
         evm.state_gas_left += call_state_gas_reservoir
+        # EIP-8037: the new-account state gas is charged before the value
+        # transfer; an insufficient-balance soft-fail never enters the
+        # frame, so refund it to the reservoir (symmetric with CREATE).
+        credit_state_gas_refund(evm, new_account_state_gas)
     else:
         generic_call(
             evm,
@@ -501,6 +516,7 @@ def call(evm: Evm) -> None:
                 memory_output_size=memory_output_size,
                 code=code,
                 disable_precompiles=is_delegated,
+                new_account_state_gas=new_account_state_gas,
             ),
         )
 
--- a/docs/writing_tests/post_mortems.md
+++ b/docs/writing_tests/post_mortems.md
@@ -72,6 +72,35 @@ None required - the existing framework supported writing these tests.
 
 ---
 
+## 2026-06 - CALL Soft-Fail New-Account State-Gas Refund - Amsterdam
+
+### Description
+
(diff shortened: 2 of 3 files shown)


bshastry:
⚠️ **Heads-up: do not merge as-is — this will fail every client today.**

This PR makes the reference (EELS) follow the EIP-8037 prose, which is the
*opposite* of what every client currently implements. On `bal-devnet-7`,
geth (`9e04883a`), revm (`ab5e3a4c`), besu, and nethermind all **retain**
the new-account state gas on a value-`CALL` soft-fail (verified: identical
`gasUsed 213,920` and post-state root for the minimal case). They match
the *previous* executable-spec behavior, not the prose.

So if this is merged, the regenerated EIP-8037 fixtures will **fail
against all of those clients** until each one adds the `CALL*` refund.
It should not land before either (a) the clients adopt the prose
behavior in a coordinated way, or (b) the decision goes the other way and
EIP-8037's text is amended to state the `CALL*`-vs-`CREATE` asymmetry
explicitly (in which case this becomes a test-only PR that keeps current
behavior).

Opening it as a draft purely to surface the prose-vs-executable-spec
divergence and get an authors' decision on which side is intended.

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -338,6 +344,9 @@ def generic_call(evm: Evm, params: GenericCall) -> None:
     if evm.message.depth + Uint(1) > STACK_DEPTH_LIMIT:
         evm.gas_left += params.gas
         evm.state_gas_left += params.state_gas_reservoir
+        # EIP-8037: refund new-account state gas charged before the frame
+        # when the stack-depth check fails (symmetric with CREATE).
+        credit_state_gas_refund(evm, params.new_account_state_gas)
```
kclowes, line 349: I think this needs to be gated behind `new_account_charged` because "the account creation charge is applied conditionally... Only when the target account does not exist and a postive value is transferred", but let me know if you think that's incorrect. I have a PR up (#3005) with this change.

src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -483,6 +494,10 @@ def call(evm: Evm) -> None:
         evm.return_data = b""
         evm.gas_left += message_call_gas.sub_call
         evm.state_gas_left += call_state_gas_reservoir
+        # EIP-8037: the new-account state gas is charged before the value
+        # transfer; an insufficient-balance soft-fail never enters the
+        # frame, so refund it to the reservoir (symmetric with CREATE).
+        credit_state_gas_refund(evm, new_account_state_gas)
```
kclowes, line 500: Same here, I think this needs to be conditionally applied.

kclowes (commented):
Thanks @bshastry! I commented inline too, but I think the call `credit_state_gas_refund` needs to be applied conditionally when the target account does not exist and a positive value is transferred, like [here](https://github.com/ethereum/execution-specs/pull/3005/changes#diff-28a1b575fd7c3d82832c0826cf58a881101643543d35c123c78ca65202152c23R346-R347). I pulled your test onto my #3005 PR and it does pass/fail as expected, so I do think that is valid and we should pull it in. /cc @spencer-tb

bshastry:
Thanks for the review @kclowes ! Please feel free to close this as superseded by #3005

spencer-tb:
CIFO #3005!

PR #3018, opened 2026-06-18
chore(tests,test-forks): Address #2990 review comments (pkgutil, type: ignore, Spec dataclass)

marioevz:
## 🗒️ Description

Follow-up cleanups addressing @SamWilsn review comments from #2990. Three independent refactors:

### 1. Load contract binaries via `pkgutil.get_data`

Replace the `Path(realpath(__file__)).parent / "contracts" / "...".read_bytes()` pattern used to load system-contract bytecode in the fork definitions with `pkgutil.get_data`, accessing the `.bin` files as proper package data. Reading via `__file__` only works when the package is installed unpacked on disk; it breaks when `execution_testing` is imported from a zip/wheel or any non-extracted location. The `.bin` files are already declared as package data (`"execution_testing.forks" = ["forks/eips/**/*.bin"]`), so `pkgutil.get_data` is the correct, install-layout-independent way to read them.

- Add `forks/bytecode.py` with a reusable helper, `load_contract_bytecode(module_name, filename)`, which reads `contracts/<filename>` relative to the calling module and raises if the resource is missing.
- Migrate the EIP-2935, EIP-6110, EIP-7002, and EIP-7251 fork definitions to use it, dropping the now-unused `os.path.realpath` / `pathlib.Path` imports.

### 2. Drop `# type: ignore` on `pre_allocation_blockchain` merges

The `{...} | super().pre_allocation_blockchain()` merge required a `# type: ignore` because `dict.__or__` is typed to only accept another `dict`, not the abstract `Mapping` return type, even though the value is a real `dict` at runtime. Merging via `**` unpacking (`{..., **super().pre_allocation_blockchain()}`) is accepted by mypy with identical runtime semantics, so the suppression can be removed. Applied across EIP-4788, EIP-6110, EIP-2935, EIP-7002, and EIP-7251.

### 3. Remove inert `@dataclass(frozen=True)` from constant-only `Spec` classes

The per-EIP `Spec` classes are pure constant namespaces — never instantiated, their members are un-annotated class variables (so not even dataclass fields), and they are never serialized. `@dataclass(frozen=True)` therefore provides no benefit (`frozen` only guards instance attribute assignment, which never happens). Removed the decorator from every `class Spec` across the test suite (34 `spec.py` files), and dropped the `dataclass` import where `Spec` was its only user. `ReferenceSpec` and other genuine dataclasses are left untouched.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/forks/bytecode.py
+++ b/packages/testing/src/execution_testing/forks/bytecode.py
@@ -1,0 +1,19 @@
+"""Helper to load predeploy contract bytecode bundled as package data."""
+
+import pkgutil
+
+
+def load_contract_bytecode(module_name: str, filename: str) -> bytes:
+    """
+    Load predeploy contract bytecode bundled as package data.
+
+    `module_name` is the importing module's `__name__`; the bytecode is read
+    from a `contracts/<filename>` resource located alongside that module.
+    """
+    resource = f"contracts/{filename}"
+    bytecode = pkgutil.get_data(module_name, resource)
+    if bytecode is None:
+        raise FileNotFoundError(
+            f"Unable to read bytecode `{resource}` from `{module_name}`"
+        )
+    return bytecode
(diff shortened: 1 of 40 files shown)


marioevz:
I'll probably add more backports from the rest of the comments in the other PR review to this same PR. Converting to draft in the meantime.

LouisTsai-Csie (approved):
LGTM

SamWilsn (approved):

PR #3024, opened 2026-06-19
fix(test-cli): raise a clear error when `gentest` can't find `ruff`

danceratopz:
## 🗒️ Description

When `ruff` is missing from the environment, gentest raised an opaque `<Result FileNotFoundError(2, 'No such file or directory')>` that never mentioned `ruff`. The error is now caught and re-raised with a clear hint to install the dev environment.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/cli/gentest/source_code_generator.py
+++ b/packages/testing/src/execution_testing/cli/gentest/source_code_generator.py
@@ -85,7 +85,7 @@ def format_code(code: str) -> str:
         config_path = AppConfig().ROOT_DIR.parent / "pyproject.toml"
 
         try:
-            result = subprocess.run(
+            subprocess.run(
                 [
                     str(formatter_path),
                     "format",
@@ -96,6 +96,7 @@ def format_code(code: str) -> str:
                 ],
                 capture_output=True,
                 text=True,
+                check=True,
             )
         except FileNotFoundError as e:
             raise FileNotFoundError(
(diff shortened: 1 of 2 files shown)


packages/testing/src/execution_testing/cli/gentest/source_code_generator.py
```diff
@@ -84,20 +84,28 @@ def format_code(code: str) -> str:
...
+                text=True,
+            )
+        except FileNotFoundError as e:
+            raise FileNotFoundError(
+                f"Could not run the 'ruff' formatter at '{formatter_path}'. "
+                f"gentest requires 'ruff' to format generated tests; ensure "
+                f"the development environment is installed, e.g. with "
+                f"'uv sync'."
+            ) from e
         if result.returncode != 0:
```
SamWilsn: Would [`check_output`](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) be better than `run`?

SamWilsn (approved):

packages/testing/src/execution_testing/cli/gentest/source_code_generator.py
danceratopz: Thanks for the suggestion @SamWilsn, I didn't know `check_output()`. I think your motivation for this suggestion was to improve readability by handling both error modes via exception handling? I took a look and settled with `run(check=True)` over `check_output()` in the end (last commit). There's almost no difference between the two tbh, I just preferred to keep `run()` for familiarity.

PR #3032, opened 2026-06-22
feat(tests): CREATE/CREATE2 and CALL clear return data on failed pre-checks

omerfirmak:
## Motivation

The existing CREATE/CREATE2 return-data tests (e.g. `eip1014_create2/test_create_returndata.py::test_create2_return_data`, the `stCreateTest` returndata ports, and the `eip7610` collision test) all either execute the create's initcode or exercise only the *collision* pre-check. None cover the **depth / nonce / balance pre-check abort** paths, where the create fails **before any initcode runs**.

Per the spec, entering a CREATE always resets the return-data buffer — unconditionally, including those pre-check failures. A client that clears the buffer only *after* the pre-checks would leave stale return data from a preceding CALL observable via `RETURNDATASIZE` / `RETURNDATACOPY`. This is an easy implementation mistake that the current suite does not catch.

## What this adds

A state test (parametrized over `CREATE` and `CREATE2`) that:
1. CALLs a callee returning 32 bytes, so the return-data buffer holds 32 bytes (`RETURNDATASIZE == 32`),
2. executes a create with `value` greater than the creator's balance, failing the **insufficient-balance pre-check** before initcode runs,
3. asserts `RETURNDATASIZE == 0` afterward (buffer cleared), along with the create result being `0`.

A buggy client that only clears the buffer after the balance check returns the stale `32`, which changes the value the contract SSTOREs and is observable in the post-state.

## Verification

Fills cleanly. Verified against an independent client implementation (zevm) which had exactly this bug: it **fails** when the return-data clear happens after the balance pre-check, and **passes** when the clear happens on create entry.

The corresponding client fix that this test exercises: https://github.com/omerfirmak/zevm/pull/37/commits/b200243d53c50229c0ed9662867f31725c819c6e (moves the return-data buffer reset to the top of `create()`, before the depth/nonce/balance pre-checks). This regression was originally found on a `glamsterdam-devnet-5` block, where a CREATE2 failing with "insufficient balance for transfer" left a preceding precompile call's 64-byte return data observable, corrupting downstream gas/state accounting.

### DIFF
--- a/tests/byzantium/eip211_return_data/test_call.py
+++ b/tests/byzantium/eip211_return_data/test_call.py
@@ -1,0 +1,91 @@
+"""Test CALL return data buffer behavior on pre-check failure."""
+
+import pytest
+from execution_testing import (
+    Account,
+    Alloc,
+    Op,
+    StateTestFiller,
+    Transaction,
+)
+
+from .spec import ref_spec_211
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_211.git_path
+REFERENCE_SPEC_VERSION = ref_spec_211.version
+
+
+@pytest.mark.valid_from("Byzantium")
+def test_call_clears_return_data_on_insufficient_balance(
+    pre: Alloc,
+    state_test: StateTestFiller,
+) -> None:
+    """
+    Test that a CALL clears the return-data buffer even when the call fails the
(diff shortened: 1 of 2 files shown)


LouisTsai-Csie (approved):
I push some refactor, thank you for this PR.

PR #3033, opened 2026-06-22
test(amsterdam): add EIP-8038 state-access gas cost tests

CPerezz:
## 🗒️ Description

Adds a comprehensive test suite for **EIP-8038 (State-access gas cost update)** under
`tests/amsterdam/eip8038_state_access_gas_cost_increase/`, covering the **regular-gas**
state-access repricing on the Amsterdam fork:

- `SLOAD`/`SSTORE` cold/warm access, the SSTORE access + `STORAGE_WRITE` split, the full
  original/current/new matrix, and the refund schedule (incl. the clear-refund reversal) with the
  20% cap;
- the `EXTCODESIZE`/`EXTCODECOPY` code-read `WARM_ACCESS` surcharge vs `BALANCE`/`EXTCODEHASH`;
- `CALL`/`CALLCODE`/`DELEGATECALL`/`STATICCALL` access, `CALL_VALUE`, the value→new-account seam,
  EIP-7702 delegation double-access (all four opcodes, single-hop), the 63/64 forwarding, and the
  stipend;
- `CREATE`/`CREATE2` `CREATE_ACCESS` (+ collision), `SELFDESTRUCT` `ACCOUNT_WRITE` (incl. EIP-6780
  same-tx-created burn/transfer cases);
- the EIP-7702 authorization regular intrinsic + dual-channel refunds (incl. mixed validity);
- access-list pricing, transient-storage non-repricing, exact-balance no-fallback boundaries,
  a mainnet smoke subset, and fork-transition before/after Amsterdam.

> This PR also includes one small **test-framework** correction in`forks/eips/amsterdam/eip_8037.py`: `SELFDESTRUCT`'s new-account funding cost (`NEW_ACCOUNT`) was
folded into the *regular* gas dimension and absent from the state-gas opcode map, whereas the EELS
spec charges it as *state* gas (the EIP-8038 `ACCOUNT_WRITE` surcharge stays regular). 
The fix mirrors the existing `CREATE` handling and leaves the total `gas_cost` unchanged; without it the
regular/state split for `SELFDESTRUCT` is mis-modelled and a block-level `gas_used = max(regular, state)` assertion could be silently wrong.

The whole `tests/amsterdam` tree fills green on `--fork Amsterdam` with no regressions; the new
suite is 540 fixtures.

## 🔗 Related Issues or PRs
Tests for EIP-8038, whose spec and test-framework support were added in:
- #2972.

EIP-8038 tracker:
- https://github.com/ethereum/execution-specs/issues/1941

Issue #1941, opened 2025-12-18
EIP-8038 Implementation Tracker: State-access gas cost update


marioevz:
### Target Fork

**Amsterdam**

### Instructions

- [x] Assign issue to EIP specification and testing owner(s).

> [!IMPORTANT]
> A specifications specialist and a testing specialist should ideally share ownership of the EIP.

- [x] Add the issue to the target fork milestone if applicable (i.e., the EIP is at least in the [CFI stage](https://eips.ethereum.org/EIPS/eip-7723#considered-for-inclusion)).

#### Guidance for Marking Items Complete

An item should only be checked off once the EIP is considered *stable*. In this context, stable means:

- No major issues or ambiguities are still being uncovered in the specification or tests.
- There are no open discussion points awaiting resolution.
- Client implementations have been consistently passing the tests for at least a week.

It is ultimately up to the owners' discretion to decide when an item should be marked as complete, using this guidance as the basis for that decision.

In exceptional cases, an EIP may require changes after some items have been marked complete or even after the entire issue has been completed and closed. This can happen, for example, when significant design optimizations are identified and agreed upon in ACD, or when critical security issues surface and require updates to the specification or tests.

When this occurs, owners should either unmark the relevant checkboxes if the issue is still open, or create a new tracking issue for the modifications if the original issue had already been closed.

### Specification + Testing Status

- [x] Testing complexity assessed and documented: [EIP-8038.md](https://github.com/ethsteel/pm/blob/main/complexity_assessments/EIPs/EIP-8038.md) - 🔴 20.
- [x] Specification implementation merged to `eips/amsterdam/eip-8038` *(skip if the fork branch merge below is already complete)*.
- [x] Specification updates merged to the corresponding `forks/amsterdam` branch. #3052 
- [x] EIP updates proposed in case of architectural choices surfaced during implementation. - ethereum/EIPs#12343
- [x] Required testing framework modifications implemented. #3052 
- [x] Test suite implemented.
- [x] Full code coverage for all changes.
- [x] No regressions or failures in tests from prior forks (including static tests).
- [x] [Testing checklist](https://github.com/ethereum/execution-specs/blob/HEAD/docs/writing_tests/checklist_templates/eip_testing_checklist_template.md) complete.
- [x] Hardening session completed.
- [x] Benchmarking tests written and results documented.
- [x] Ran tests using `execute` to ensure compatibility, and marked specific tests to be skipped when they cannot be executed on live networks.
[...]

PR #3044, opened 2026-06-24
fix(tests): update EIP-8038 SELFDESTRUCT tests for EIP-8246 burn log removal

danceratopz:
Update the EIP-8038 `SELFDESTRUCT` regular-gas test so it is correct on forks with and without EIP-8246. The same-transaction self-destruct-to-self expectations are gated on `fork.is_eip_enabled(8246)`: pre-EIP-8246 a `Burn` log is emitted and the originator is deleted; from EIP-8246 the burn is removed, so the balance stays in the emptied originator and no log is emitted. `burn_log` is imported lazily because EIP-8246 deletes it from the EIP-7708 spec.

## 🔗 Related Issues or PRs
Follow-up to:
- #3033

Implementation trackers:
- #1941

### DIFF
--- a/tests/amsterdam/eip8038_state_access_gas_cost_increase/test_selfdestruct_gas.py
+++ b/tests/amsterdam/eip8038_state_access_gas_cost_increase/test_selfdestruct_gas.py
@@ -51,12 +51,13 @@ from execution_testing import (
     StateTestFiller,
     Storage,
     Transaction,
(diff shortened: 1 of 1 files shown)


danceratopz:
`test-tests` failed due to an evmone build failure (probably a flake); not waiting for benchmark tests (irrelevant for this change).

Issue #3043, opened 2026-06-24
Test Release Tracker: `glamsterdam-devnet@v6.1.0`


danceratopz:
### Overview

- **Target release tag**: `tests-glamsterdam-devnet@v6.1.0`
- **EELS branch**: [`devnets/glamsterdam/6`](https://github.com/ethereum/execution-specs/tree/devnets/glamsterdam/6) 
- **Test release date**: 2026-06-25
- **Devnet launch date**: Week 26, ~June 25th 2026
- **Related Trackers**
  - #2915 
  - #3001
  - #3049 

### Scope

Specs & tests bug fix release for a subtle [EIP-2780](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2780.md) inconsistency with [EIP-161](https://eips.ethereum.org/EIPS/eip-161): As v6.1.0, EELS charges account creation costs if value is transferred to a precompile with previously zero-balance.
- https://github.com/ethereum/execution-specs/pull/3048

#### New EIPs
None

#### EIP Updates
None

#### Spec Fixes
- [x] **[EIP-2780](https://eips.ethereum.org/EIPS/eip-2780):** Resource-based intrinsic transaction gas 
  - https://github.com/ethereum/execution-specs/pull/3048
  - > for the empty recipient and transaction carrys the value, if the recipient is the precompile, shall we charge the account-creation cost? From: https://discord.com/channels/595666850260713488/1504670887851327648/1519527453230501998

#### Test Fixes
None

#### New Test Cases
- A singular new EIP-2780 test case was added checks value transfer to a zero-balance precompile at the oog boundary (also in #3048).

### Status
- [x] ~~All relevant PRs merged to `forks/amsterdam`.~~
- [x] ~~`glamsterdam-devnet-6` branch created and rebased on `forks/amsterdam`.~~
- [x] Fixtures generated and released.
- [x] [`hive-tests`](https://github.com/ethpandaops/hive-tests/tree/master/.github/workflows) updated to run the new fixtures.
- [x] [`hive-ui`](https://github.com/ethpandaops/hive-ui/blob/master/public/discovery.json) discovery updated to the latest `glamsterdam-devnet` workflow.
- [x] Linked the test fixture release tag within the issue and closed it.

danceratopz:
For speed and ease of ensuring correctness this release was also created by cherry-picking the fix from `eips/amsterdam/2780-8038` (#3048):
[...]

PR #3048, opened 2026-06-25
fix(spec-specs): EIP-2780 charge `NEW_ACCOUNT` for value transfer to zero balance precompile

gurukamath:
## 🗒️ Description
EIP-2780 charges the NEW_ACCOUNT state cost when a transaction transfers value to a recipient that is empty per [EIP-161](https://eips.ethereum.org/EIPS/eip-161). The top-frame charge previously carved out precompile recipients, but neither EIP-2780 nor EIP-161 authorizes that exemption:

- EIP-2780 does not mention precompiles; its rule keys solely on "empty per EIP-161 and tx.value > 0".
- EIP-161 defines empty structurally (no code, zero nonce, zero balance) with no precompile exception, so an unfunded precompile is empty and is created by the value transfer like any other account.

Remove the `recipient_is_precompile` carve-out from the top-frame charge so an empty precompile receiving value pays NEW_ACCOUNT, drop the matching special-case from the testing framework's `transaction_top_frame_state_gas`, and rewrite `test_value_move_to_precompiles` to assert the charge fires for the not-funded precompile while a pre-funded (alive) precompile remains exempt by virtue of being non-empty.

## Related Issues

- #1940 
- #3043 

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_2780.py
+++ b/packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_2780.py
@@ -146,15 +146,10 @@ class EIP2780(BaseFork):
         """
         Return the state gas charged at the top-level transaction
         frame. Charges ``NEW_ACCOUNT`` when value is transferred to an
-        empty recipient; zero otherwise. Precompile recipients are
-        exempt: they are protocol-inherent rather than created by a
-        value transfer, so they never trigger ``NEW_ACCOUNT`` even
-        when state-empty.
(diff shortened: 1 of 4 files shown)


danceratopz (commented):
The change looks good to me! I added a new test that tests the behavior at the oog boundary, ptal here:
- https://github.com/gurukamath/execution-specs/pull/13

danceratopz (approved):
Thanks @gurukamath!

PR #3050, opened 2026-06-25
fill-stateful: make the per-test chain rewind optional (add debug_resetHead)

skylenet:
## Summary

The per-test chain rewind (`_reset_chain_between_tests`) hard-required geth's `debug_setHead`. This makes it optional and client-agnostic:

- **adds `debug_resetHead`** (by block hash) as an alternative — for clients like Nethermind that don't implement `debug_setHead`;
- **makes the debug rewind optional**: if a client implements neither, fill-stateful keeps going instead of aborting — each test's first block is built on its explicit `start_block` parent, so the client reorgs onto it without any debug call.

So fill-stateful is no longer tied to the geth `debug` namespace. Nethermind is the concrete client this was validated against, and it now works as a filler.

## Details

- **Three-tier per-test rewind** in `rewind_head`, probed once and cached:
  1. `debug_setHead` (by number) — geth; **truncates** the chain, keeping the block tree small;
  2. else `debug_resetHead` (by hash) — e.g. Nethermind;
  3. else **nothing** — rely on the explicit `start_block` parent; the client reorgs onto it.
- **Read per-test state at `start_block`, not `"latest"`** (the seed-nonce sync and the post-rewind check). A `debug_resetHead`-style rewind (or no rewind) restores the build state but leaves `latest` at the previous test's tip, so reading `"latest"` there returns a stale nonce and every funding tx is rejected (`Invalid nonce - expected 0`). Identical to `"latest"` on clients (geth) whose `debug_setHead` moves it.
- Two small base fill fixes (deploy gas safety-buffer cap; skip redundant seed withdrawal funding).

## Validation

Against the bloatnet stateful repricing prestate (Amsterdam), with `nethermindeth/nethermind:master` as the filler:

- **Fills 225/225** (0 failed), via `debug_resetHead` and, on a separate run, via the no-debug ("none") path.
- Both fixture sets **replay clean on geth and Nethermind** — 0 `InvalidBlockLevelAccessList`, 0 `invalid block access list`, 0 validation failures, 450/450 rollbacks.

Notes:
- `debug_setHead` truncates on geth (bounded block tree). `debug_resetHead` only repoints the head on Nethermind, so per-test forks still accumulate there (same as the "none" path) — measured identical reorg counts with and without it.
- For Nethermind, the filler must be a build with **both** `testing_buildBlockV1` *and* correct EIP-7928 BALs — `master` works; the older `testing_build_block_with_opcode_tracing` image builds invalid BALs that both geth and Nethermind reject on replay.
- Geth filling is unaffected (it takes the `debug_setHead` tier, and `latest` already tracks the rewind).

### DIFF
--- a/docs/filling_tests/fill_stateful.md
+++ b/docs/filling_tests/fill_stateful.md
@@ -13,7 +13,7 @@ The target client must expose:
 
 - `testing` (`testing_buildBlockV1`) — block construction with explicit transaction ordering.
 - `engine` — `engine_newPayloadVX`, `engine_forkchoiceUpdatedVX`.
-- `eth`, `debug` — chain queries and `debug_setHead` for between-test rewind.
+- `eth`, `debug` — chain queries and `debug_setHead` (or `debug_resetHead` on clients like Nethermind that lack `debug_setHead`) for between-test rewind.
 - `web3` (optional) — `web3_clientVersion` is recorded into the fixture's `_info.filling-transition-tool` for traceability.
 
 The production-ready filler is `ethpandaops/geth:master`.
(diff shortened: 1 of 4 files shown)


marioevz (approved):
Nice, very good additions. I think we need to bring them to `forks/amsterdam` eventually.

PR #3053, opened 2026-06-25
fix(spec-tests): register JSON fixture fork marker

### LINKED ISSUE #1407: Register per-fork custom marks in `pytest_configure` or `pyproject.toml`
### What was wrong?

We generate a ton of warnings while running `pytest`:

```
tests/json_infra/test_transaction.py:25: PytestUnknownMarkWarning: Unknown pytest.mark.fork - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.fork(fork_name)
```

### How can it be fixed?

If we register our mark, we can silence this warning. If we can register the whole `fork` mark at once, we should do it statically in `pyproject.toml`, but if we have to register each fork separately (`fork(osaka)`), we should do it in [`pytest_configure`](https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_configure).

JackCC703:
## 🗒️ Description

Register the `fork(name)` pytest marker used by JSON fixture tests.

JSON fixture collection adds `pytest.mark.fork(...)` to collected tests, but the marker was not registered in the root pytest configuration. This caused `PytestUnknownMarkWarning` during pytest runs. Registering the marker in `pyproject.toml` silences the warning without changing test behavior.

## 🔗 Related Issues or PRs

Fixes #1407.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Cute animal picture

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -292,6 +292,7 @@ markers = [
     "json_state_tests: marks tests as json_state_tests (deselect with '-m \"not json_state_tests\"')",
     "vm_test: marks tests as vm_test (deselect with '-m \"not vm_test\"')",
     "eels_base_coverage: Minimized subset selected to preserve high EELS line-coverage parity (select with '-m eels_base_coverage')",
+    "fork: marks JSON fixture tests by fork",
     "repricing: marks tests for gas repricing analysis",
     "stub_parametrize: parametrize test from address stubs by prefix",
 ]


LouisTsai-Csie (approved):

pyproject.toml
```diff
@@ -292,6 +292,7 @@ markers = [
     "json_state_tests: marks tests as json_state_tests (deselect with '-m \"not json_state_tests\"')",
     "vm_test: marks tests as vm_test (deselect with '-m \"not vm_test\"')",
     "eels_base_coverage: Minimized subset selected to preserve high EELS line-coverage parity (select with '-m eels_base_coverage')",
+    "fork(name): marks JSON fixture tests by fork",
```
danceratopz: I'm not sure this would actually match any marker as `name` is not a valid fork name (as noted in #1407).

danceratopz (changes requested):
Sorry, #1407 is an old issue and I don't think it's required any longer. Could you share the command you're using and the warnings that this change fixes? I.e., can you still reproduce them locally before this change. I had a quick go and couldn't create the warnings.

JackCC703:
Oh, Sorry. I think the difference is that my checkout has generated JSON fixtures under `tests/json_loader/fixtures/`, which are ignored by git. A clean checkout does not have those files, so the warning won't appear unless that generated fixture collection workflow is used.

So this PR only fixes the warning for collecting generated JSON fixtures, not for a clean default pytest run. Since the issue is old, if that workflow is no longer considered supported or relevant, I'm happy to close this PR.

pyproject.toml
JackCC703: Thank you. I updated the marker registration from `fork(name)` to `fork: ...` to
avoid implying that `name` is a fork value.

JackCC703:
On `origin/forks/amsterdam`, with a generated JSON fixture present under
`tests/json_loader/fixtures/`, I can reproduce it with:

```console
.venv/bin/pytest --collect-only -q tests/json_loader/fixtures/blockchain_tests/for_frontier/frontier/validation/transaction/sender_balance.json -W "error::pytest.PytestUnknownMarkWarning"
```

This fails before the change with:

```console
E   pytest.PytestUnknownMarkWarning: Unknown pytest.mark.fork - is this a typo?  You can register custom marks to avoid this warning
```

The warning is triggered from:

```python
self.add_marker(pytest.mark.fork(self.fork_name))
```

in `tests/json_loader/helpers/load_blockchain_tests.py`.

After this PR, the same command passes with `PytestUnknownMarkWarning`
configured to fail the run, which confirms that the warning is no longer emitted.

danceratopz (approved):
Thanks @JackCC703 for fixing the marker and the steps to reproduce, appreciated!

PR #3055, opened 2026-06-26
feat(tests): add more EIP-2780 tests

gurukamath:
## 🗒️ Description

Expands the EIP-2780 (Reduce Transaction Intrinsic Cost) test suite on top of the `eip-2780-8038` base, pinning the decomposed-intrinsic, top-frame state-gas, and calldata-floor interactions not covered by the existing tests:

- **Fork transition** — identical transactions straddling the Amsterdam boundary assert the per-tx gas changes by the EIP-2780 amount only once the fork activates (flat 21,000 → decomposed intrinsic); self-transfers collapse to `TX_BASE`.
- **Top-frame `NEW_ACCOUNT` as state gas** — a block-header `gas_used` assertion pins that the empty-recipient value-transfer charge is attributed to the *state* dimension (`max(block_regular, block_state)`), which balance-only state tests cannot distinguish.
- **EIP-7708 transfer log** — fires exactly when `TRANSFER_LOG_COST` is charged (non-self value transfers), and never for self-transfers or zero-value transactions.
- **Intrinsic & floor boundaries** — `gas_limit = intrinsic − 1` rejection for both plain calls and contract creation; the EIP-7623/7976 calldata floor binding (masking the decomposed value charges) plus below-floor rejection; and a nonce-only-alive recipient confirming the `NEW_ACCOUNT` gate keys on `is_account_alive`, not `balance == 0`.

Test-only; no spec changes. All fixtures fill green against Amsterdam.

## 🔗 Related Issues or PRs
#3017 

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Cute Animals - 6 of 7

### DIFF
--- a/tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
+++ b/tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
@@ -1,20 +1,4 @@
-"""
-EIP-2780 interaction with the EIP-7623/7976 calldata floor.
-
-A transaction's gas accounting uses ``max(intrinsic, calldata_floor)``.
-EIP-2780 decomposes the intrinsic (``TX_BASE`` + recipient access +
-value-transfer charges) and lowers ``TX_BASE`` to 12_000; that lowered
-base also feeds the calldata floor. These tests pin the data-heavy
-regime where the floor dominates:
-
-- The floor binds, so ``gas_used`` equals the floor and the
-  recipient/value charges folded into the intrinsic are masked: the
-  gas paid is identical for a zero-value and a value-bearing
-  transaction of the same calldata size.
-- One gas below the floor, the transaction is rejected with
-  ``INTRINSIC_GAS_BELOW_FLOOR_GAS_COST`` even though it covers the
-  (smaller) decomposed intrinsic.
-"""
+"""EIP-2780 interaction with the EIP-7623/7976 calldata floor."""
 
 import pytest
 from execution_testing import (
@@ -77,12 +61,12 @@ def _floor_dominating_calldata(fork: Fork) -> Bytes:
 
 
 @pytest.mark.parametrize(
-    "outcome",
+    "gas_modifier",
     [
-        pytest.param("floor_binds", id="floor_binds"),
+        pytest.param(0, id="at_floor"),
         pytest.param(
-            "below_floor",
-            id="below_floor_rejected",
+            -1,
+            id="below_floor",
             marks=pytest.mark.exception_test,
         ),
     ],
@@ -98,17 +82,16 @@ def test_calldata_floor(
     fork: Fork,
     pre: Alloc,
     state_test: StateTestFiller,
-    outcome: str,
+    gas_modifier: int,
     value: int,
 ) -> None:
     """
     A data-heavy transaction to an existing EOA whose calldata floor
     exceeds the decomposed value-transfer intrinsic.
 
-    - ``floor_binds``: with a gas limit above the floor, ``gas_used``
+    - ``at_floor``: with a gas limit exactly at the floor, ``gas_used``
       pins to the floor, so the value-transfer charges
       (``TRANSFER_LOG_COST + TX_VALUE_COST``) folded into the intrinsic
-      are masked -- the gas paid is identical at ``value == 0`` and
       ``value == 1`` and only the moved wei differs.
     - ``below_floor``: a gas limit one short of the floor still covers
       the (smaller) decomposed intrinsic, so the floor -- built on the
@@ -126,42 +109,35 @@ def test_calldata_floor(
     gas_price = 1_000_000_000
 
     post: dict[Address, Account] = {}
-    if outcome == "below_floor":
-        # ``gas_limit`` one short of the floor still covers the
-        # decomposed intrinsic, so the floor is the only thing that can
-        # reject it; the post state is empty (transaction rejected).
-        intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(
-            calldata=calldata,
-            sends_value=bool(value),
-            recipient_type=RecipientType.EOA,
-            return_cost_deducted_prior_execution=True,
-        )
-        gas_limit = calldata_floor - 1
-        assert intrinsic_gas <= gas_limit, (
-            "gas_limit must still cover the decomposed intrinsic so the "
-            "rejection is pinned to the calldata floor"
-        )
-        tx = Transaction(
-            sender=sender,
-            to=target,
-            value=value,
-            data=calldata,
-            gas_limit=gas_limit,
-            gas_price=gas_price,
-            error=TransactionException.INTRINSIC_GAS_BELOW_FLOOR_GAS_COST,
-        )
-    else:
-        # ``floor_binds``: no explicit gas limit (auto-fills above the
-        # floor). The gas component is the floor regardless of value
-        # (charges masked); only the transferred wei changes the
-        # balance.
-        tx = Transaction(
-            sender=sender,
-            to=target,
-            value=value,
-            data=calldata,
-            gas_price=gas_price,
-        )
+    gas_limit = calldata_floor + gas_modifier
+    # Even at the reduced limit the decomposed intrinsic is still
+    # covered, so the calldata floor is the sole gate on the
+    # transaction.
+    intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(
+        calldata=calldata,
+        sends_value=bool(value),
+        recipient_type=RecipientType.EOA,
+        return_cost_deducted_prior_execution=True,
+    )
+    assert intrinsic_gas <= gas_limit, (
+        "gas_limit must still cover the decomposed intrinsic so the "
+        "outcome is pinned to the calldata floor"
+    )
+
+    tx = Transaction(
+        sender=sender,
+        to=target,
+        value=value,
+        data=calldata,
+        gas_limit=gas_limit,
+        gas_price=gas_price,
+        error=(
+            TransactionException.INTRINSIC_GAS_BELOW_FLOOR_GAS_COST
+            if gas_modifier < 0
+            else None
+        ),
+    )
+    if gas_modifier == 0:
         sender_final_balance = (
             sender_initial_balance - value - calldata_floor * gas_price
         )
--- a/tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
+++ b/tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
@@ -1,0 +1,173 @@
+"""
+EIP-2780 interaction with the EIP-7623/7976 calldata floor.
+
+A transaction's gas accounting uses ``max(intrinsic, calldata_floor)``.
+EIP-2780 decomposes the intrinsic (``TX_BASE`` + recipient access +
+value-transfer charges) and lowers ``TX_BASE`` to 12_000; that lowered
+base also feeds the calldata floor. These tests pin the data-heavy
+regime where the floor dominates:
+
+- The floor binds, so ``gas_used`` equals the floor and the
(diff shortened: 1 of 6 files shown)


tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
```diff
@@ -0,0 +1,173 @@
...
+base also feeds the calldata floor. These tests pin the data-heavy
+regime where the floor dominates:
+
+- The floor binds, so ``gas_used`` equals the floor and the
+  recipient/value charges folded into the intrinsic are masked: the
+  gas paid is identical for a zero-value and a value-bearing
+  transaction of the same calldata size.
+- One gas below the floor, the transaction is rejected with
+  ``INTRINSIC_GAS_BELOW_FLOOR_GAS_COST`` even though it covers the
+  (smaller) decomposed intrinsic.
```
LouisTsai-Csie: The comment here is similar to the test's docstring. What do you think about combining these two?

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
```diff
@@ -0,0 +1,173 @@
...
+    "outcome",
+    [
+        pytest.param("floor_binds", id="floor_binds"),
+        pytest.param(
+            "below_floor",
+            id="below_floor_rejected",
+            marks=pytest.mark.exception_test,
+        ),
+    ],
+)
```
LouisTsai-Csie, line 73: Could we use `gas_modifier` as the parametrization instead of `outcome`?

```python
@pytest.mark.parametrize(
    "gas_modifier",
    [
        pytest.param(0, id="at_floor"),
        pytest.param(
            -1,
            id="below_floor",
            marks=pytest.mark.exception_test,
        ),
    ],
)
```

The idea is to simplify the logic: instead of an `if`/`else` that picks the
gas limit per outcome, fold it into a single expression

```python
gas_limit = calldata_floor + gas_modifier
```

That way both cases share one `Transaction` construction, and whether the transaction carries `error=INTRINSIC_GAS_BELOW_FLOOR_GAS_COST` is decided directly by `gas_modifier < 0`. It also makes the relationship to the floor explicit: `at_floor` hands the transaction exactly `calldata_floor` gas and still succeeds, which proves `gas_used == floor` more directly than the
auto-filled limit did.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
```diff
@@ -0,0 +1,173 @@
...
+        # floor). The gas component is the floor regardless of value
+        # (charges masked); only the transferred wei changes the
+        # balance.
+        tx = Transaction(
+            sender=sender,
+            to=target,
+            value=value,
+            data=calldata,
+            gas_price=gas_price,
+        )
```
LouisTsai-Csie:
```suggestion
    gas_limit = calldata_floor + gas_modifier
    # Even at the reduced limit the decomposed intrinsic is still
    # covered, so the calldata floor is the sole gate on the
    # transaction.
    intrinsic_gas = fork.transaction_intrinsic_cost_calculator()(
        calldata=calldata,
        sends_value=bool(value),
        recipient_type=RecipientType.EOA,
        return_cost_deducted_prior_execution=True,
    )
    assert intrinsic_gas <= gas_limit, (
        "gas_limit must still cover the decomposed intrinsic so the "
        "outcome is pinned to the calldata floor"
    )

    tx = Transaction(
        sender=sender,
        to=target,
        value=value,
        data=calldata,
        gas_limit=gas_limit,
        gas_price=gas_price,
        error=(
            TransactionException.INTRINSIC_GAS_BELOW_FLOOR_GAS_COST
            if gas_modifier < 0
            else None
        ),
    )
```

After applying the above suggestion, we could share the transaction construction.

LouisTsai-Csie (changes requested):
Just some style suggestions, thanks!

Two more suggestions that could be done in a follow-up PR:
- Should we add self transfer parametrization to test_calldata_floor?
- Could we reduce the funding and gas price within the test?

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
gurukamath: Yeah. Updated to include a single source of truth

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
gurukamath: Good point. Updated

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
gurukamath: Related to the above point. Updated

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
```diff
@@ -0,0 +1,153 @@
...
+    pre: Alloc,
+    state_test: StateTestFiller,
+    gas_modifier: int,
+    value: int,
+) -> None:
+    """
+    A data-heavy transaction to an existing EOA whose calldata floor
+    exceeds the decomposed value-transfer intrinsic.
+
+    - ``floor_binds``: with a gas limit above the floor, ``gas_used``
```
danceratopz: This should now be `at_floor`. 
```suggestion
    - ``at_floor``: with a gas limit exactly at the floor, ``gas_used``
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
```diff
@@ -0,0 +1,153 @@
+"""
+EIP-2780 interaction with the EIP-7623/7976 calldata floor.
+
+"""
```
danceratopz:
```suggestion
"""
EIP-2780 interaction with the EIP-7623/7976 calldata floor.
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_fork_transition.py
```diff
@@ -0,0 +1,153 @@
...
+        expected_post += post_gas_costs.COLD_ACCOUNT_ACCESS
+        if value:
+            expected_post += (
+                post_gas_costs.TRANSFER_LOG_COST + post_gas_costs.TX_VALUE_COST
+            )
+
+    timestamps = [PRE_FORK_TIMESTAMP, POST_FORK_TIMESTAMP]
+    expected_intrinsics = [expected_pre, expected_post]
+    blocks = []
+    post: dict = {}
```
danceratopz:
```suggestion
    post: dict[Address, Account] = {}
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_fork_transition.py
```diff
@@ -0,0 +1,153 @@
...
+    recipient is an existing EOA (or the sender itself for
+    ``self_transfer``), so neither block runs EVM bytecode and
+    ``gas_used`` equals the intrinsic exactly.
+
+    The per-fork intrinsic returned by the calculator is also checked
+    against a hand-derived decomposition built from each fork's gas
+    constants, so a calculator regression fails here with a clear
+    message rather than only as a downstream balance mismatch.
+    """
+    gas_price = 10**9
```
danceratopz: For consistency with other tests
```suggestion
    gas_price = 1_000_000_000
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_top_frame_charges.py
```diff
@@ -159,6 +162,140 @@ def test_top_frame_state_charge_empty_precompile(
...
+        f"({new_account_state_gas}) dominates the regular intrinsic "
+        f"({intrinsic_regular})"
+    )
+
+    # No EVM bytecode runs (empty recipient), so the only regular gas
+    # is the intrinsic and the only state gas is the top-frame
+    # ``NEW_ACCOUNT`` charge.
+    expected_gas_used = max(intrinsic_regular, new_account_state_gas)
+
+    gas_price = 10**9
```
danceratopz: For consistency with other tests
```suggestion
    gas_price = 1_000_000_000
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_top_frame_charges.py
```diff
@@ -159,6 +162,140 @@ def test_top_frame_state_charge_empty_precompile(
     state_test(pre=pre, tx=tx, post=post)
 
 
+def test_top_frame_new_account_charged_as_state_gas(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    fork: Fork,
```
danceratopz, line 166: The ordering stood out to me here, there's no rule here, but we tend to order as:

```suggestion
def test_top_frame_new_account_charged_as_state_gas(
    fork: Fork,
    pre: Alloc,
    blockchain_test: BlockchainTestFiller,
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
danceratopz, line 1: Any particular reason why you didn't add 7708 log test expectations for other two tests: `test_value_contract_creation_tx()` and `test_value_move_to_precompiles()`? Would it be worth adding there, too?

danceratopz (commented):
These look good @gurukamath! Below are some very minor nits and one question about some more 7708 tests.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
gurukamath: Added

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_calldata_floor.py
gurukamath: Updated

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_fork_transition.py
gurukamath: Updated

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_fork_transition.py
gurukamath: Updated

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_top_frame_charges.py
gurukamath: Updated

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_top_frame_charges.py
gurukamath: Updated

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_value_moving_transactions.py
gurukamath: self-transfer is the main edge case that I wanted to capture here. But, yeah. There is no harm in adding an extra check in the other cases too. So, I've added log expectations to the other tests now

danceratopz (approved):
Thanks @gurukamath!

LouisTsai-Csie (approved):

PR #3057, opened 2026-06-29
chore(ci): speed up bench-gas by filling only the `blockchain_test` format

danceratopz:
## 🗒️ Description

`bench-gas` (the `Benchmark Gas Values` leg of the `Benchmarking` workflow) is the longest job in CI. It is a correctness smoke check: It fills the `tests/benchmark/compute` suite with the configured EVM (geth) and replays the fixtures through EELS, confirming the two implementations agree. It is not the released benchmark artifact, which is built separately from the release matrix in `.github/configs/feature.yaml` and is left untouched here.

This PR restructures the recipe into three steps:

1. `fill --generate-pre-alloc-groups`: Generate the pre-allocation groups. This keeps the `BlockchainEngineX` code path smoke-tested, since pre-alloc generation is the part unique to the all-formats run and where its risk is concentrated.
2. `fill -m "blockchain_test and (not derived_test) and (not slow)"`: Fill only the `blockchain_test` format, the one EELS consumes.
3. Replay the fixtures through EELS via `json_loader`. Added `--no-dev --group test` so the PyPy environment builds without the dev Rust crates, matching the `fill-pypy` recipe.

## 🔗 Related Issues or PRs

Addresses #2673. Alternative to #2693 (accepting float `--gas-benchmark-values`): Lowering the gas value cannot help, since it is floored at roughly 0.9M by the most expensive single benchmark op, and fill time is dominated by per-format generation rather than gas magnitude. Also an alternative to #3012 (running on a larger runner), which did not improve the time because the two-phase fill does not scale with cores.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -221,17 +221,30 @@ test-ci-scripts *args:
 
 # --- Benchmarks ---
 
-# Fill benchmark tests with --gas-benchmark-values, then verify with EELS
+# Smoke-test benchmark tests: fill blockchain_test fixtures, then verify against EELS.
 [group('benchmark tests')]
 bench-gas *args:
     @mkdir -p "{{ output_dir }}/bench-gas/tmp" "{{ output_dir }}/bench-gas/logs"
-    @echo "==> Step 1/2: Filling benchmark fixtures with configured EVM (EVM_BIN={{ evm_bin }})"
+    @echo "==> Step 1/3: Generating pre-alloc groups (smoke-tests the BlockchainEngineX path)"
+    uv run fill \
+        --generate-pre-alloc-groups \
+        --evm-bin="{{ evm_bin }}" \
+        --gas-benchmark-values 1 \
+        --fork Osaka \
+        -m "not slow" \
+        -n auto --maxprocesses 10 --dist=loadgroup \
+        --output="{{ output_dir }}/bench-gas/pre-alloc" \
+        --basetemp="{{ output_dir }}/bench-gas/tmp" \
+        --log-to "{{ output_dir }}/bench-gas/logs" \
+        --clean \
+        "$@" \
+        tests/benchmark/compute
+    @echo "==> Step 2/3: Filling blockchain_test fixtures with configured EVM (EVM_BIN={{ evm_bin }})"
     uv run fill \
         --evm-bin="{{ evm_bin }}" \
         --gas-benchmark-values 1 \
-        --generate-all-formats \
         --fork Osaka \
-        -m "not slow" \
+        -m "blockchain_test and (not derived_test) and (not slow)" \
         -n auto --maxprocesses 10 --dist=loadgroup \
         --output="{{ output_dir }}/bench-gas/fixtures" \
         --basetemp="{{ output_dir }}/bench-gas/tmp" \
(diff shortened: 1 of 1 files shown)


danceratopz:
This cuts CI run time from ~25-26 min to ~12 min, based on the run here (recent run without this change: https://github.com/ethereum/execution-specs/actions/runs/28234305627/job/83645530740).

Not filling the `engine_x` format for benchmark tests is less than ideal, but seems like a necessary evil for this quick win.

Justfile
```diff
@@ -221,28 +221,41 @@ test-ci-scripts *args:
 
 # --- Benchmarks ---
 
-# Fill benchmark tests with --gas-benchmark-values, then verify with EELS
+# Smoke-test benchmark tests: fill blockchain_test fixtures, then verify against EELS.
 [group('benchmark tests')]
 bench-gas *args:
     @mkdir -p "{{ output_dir }}/bench-gas/tmp" "{{ output_dir }}/bench-gas/logs"
-    @echo "==> Step 1/2: Filling benchmark fixtures with configured EVM (EVM_BIN={{ evm_bin }})"
+    @echo "==> Step 1/3: Generating pre-alloc groups (smoke-tests the BlockchainEngineX path)"
```
spencer-tb, line 228: How essential is this phase for CI?

Justfile
```diff
@@ -221,28 +221,41 @@ test-ci-scripts *args:
...
+        --output="{{ output_dir }}/bench-gas/pre-alloc" \
+        --basetemp="{{ output_dir }}/bench-gas/tmp" \
+        --log-to "{{ output_dir }}/bench-gas/logs" \
+        --clean \
+        "$@" \
+        tests/benchmark/compute
+    @echo "==> Step 2/3: Filling blockchain_test fixtures with configured EVM (EVM_BIN={{ evm_bin }})"
+    uv run fill \
+        --evm-bin="{{ evm_bin }}" \
+        --gas-benchmark-values 1 \
```
spencer-tb, line 245: If we can use 0.9 here then I think we might as well?

spencer-tb (commented):
Happy with this overall! Thanks for the quick wins.

Justfile
danceratopz: I believe this checks the most fragile part of `engine_x` filling by checking the prealloc groups are good. For example, that no hard-coded addresses were inadvertently added to the tests (that could lead to address collisions).

Justfile
danceratopz: It's a small gain and it's not entirely for free (requires PR, but the changeset is not huge):
- https://github.com/ethereum/execution-specs/pull/2693

I was leaning towards skipping this, will re-check it.

danceratopz: I closed #2693 after a comment from Louis that the Amsterdam repricings make < 1M gaslimit values infeasible for many tests.

danceratopz: https://github.com/ethereum/execution-specs/pull/2693#issuecomment-4841874046

danceratopz:
After rebasing on `forks/amsterdam` with 
- #3060

this PR further reduces `just bench-gas` to ~5 mins in CI (#3060 took it from ~26min to 18min). 

It just goes to show how slow filling the enginex format is, will cook an optimization there.

Justfile
spencer-tb: All good lets skip! Thx

spencer-tb (approved):
Woo nice speed up!!

PR #3059, opened 2026-06-29
fix(test-types,test-forks): import `ethereum` lazily so xdist fill coverage measures it

danceratopz:
## 🗒️ Description

This PR improves the coverage measurement for `ethereum` and fixes the following warning that every `fill` job logs, once per pytest-xdist worker:

> `CoverageWarning: Module ethereum was previously imported, but not measured (module-not-measured)`

image
From https://github.com/ethereum/execution-specs/actions/runs/28472970517/job/84390462134

### Root cause

pytest-cov starts an xdist **worker's** coverage session late — in `pytest_sessionstart`. The `execution_testing` framework, however, imports the measured `ethereum` package at **import time**, well before that, from four module locations:

- `base_types` and `test_types.trie` import `ethereum.crypto.hash.keccak256` at module level — executed as soon as the `fill` CLI and its plugins are imported;
- `ExecutionSpecsTransitionTool` imports the EELS t8n (`T8N` / `ForkCache` / `create_parser` / `get_supported_forks` / `ethereum`) at module level, and the `forks` plugin calls `t8n.is_fork_supported()` from `pytest_configure`.

So by the time a worker starts measuring, `ethereum` is already in `sys.modules`. `pytest-cov`'s `Central`/`DistMaster` engines suppress this warning; `DistWorker` does not — hence it surfaces only on xdist workers.

### Coverage impact

~~Pre-importing the **source package** also makes coverage **under-count** `ethereum` — in **every** mode, `-n 0` and `-n 1` alike.~~

~~On a small sample, deferring the import raises the measured statement total from **36,926 → 39,000** (+2,074 across 152 files). The xdist warning was the visible symptom of a measurement gap that affected all runs.~~ 

**Update:** the ~2,074 figure above was a measurement error (the "baseline" was a dirty checkout). A same-checkout A/B shows the statement **denominator is unchanged** (39,000). The real effect is small and **constant**: `fill` measures **~300 more `ethereum` statements** — the import-time/header lines of modules loaded before the worker's coverage starts — ≈304 on a narrow Frontier slice and ≈303 on a 2,709-test Frontier→Cancun run, so it does not scale with the test set. At <1% it rounds to a ~0% codecov diff (see [comment](https://github.com/ethereum/execution-specs/pull/3059#issuecomment-4832391239)).

### The fix

Defer all four `ethereum` imports to first use, so they run after the worker's coverage session has started (collection / test execution / the `trylast` report header). Specifically: lazy `keccak256` imports; a lazy `ForkCache` property + in-function t8n imports + `importlib.metadata` for the tool version; and a lazily-computed, cached unsupported-fork set. **Nothing is silenced** — no `filterwarnings`, no `--disable-warnings`, no `disable_warnings`.

### Results

- `module-not-measured` gone — verified on the minimal repro, repeated `-n 4` runs, and a 2,709-test Frontier→Cancun fill.
- `-n 0` and `-n 1` report identical coverage; `fill` measures ~300 more `ethereum` statements (import-time lines) on the same 39,000-statement denominator — a constant <1% gain, not test-set-dependent.
- `eels_base_coverage` `--cov-fail-under=85` gate passes (91.45%).
- 637 framework unit tests pass; `just static` clean.

> ⚠️ **For reviewers:** the statement denominator is unchanged; the coverage % moves by <1% (≈0% codecov diff), and the `--cov-fail-under=85` gate still passes comfortably (91.45%).

### Alternative considered

A runner-only approach was prototyped (start coverage earlier via `COVERAGE_PROCESS_START` + a dedicated `coverage_fill.cfg` in the `Justfile`, dropping pytest-cov's `--cov` and combining manually). It works, but we preferred fixing the source because it:

- removes the warning for *every* `--cov=ethereum` invocation, not just the `just fill` recipe (direct `pytest`/`fill` runs, `json-loader`, IDE runs, contributors);
- keeps the pytest-cov integration rather than hand-rolling `coverage erase/combine/report/xml` and a second coverage config to keep in sync with `pyproject.toml`;
- addresses the actual root cause — a test framework shouldn't import the package-under-test at import time.

<sub>The prototype lives on a local branch `fix-coverage-warnings-start-coverage-earlier` (currently based off the wrong base), kept only for reference.</sub>

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory — no user-facing change.

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
@@ -627,8 +627,6 @@ def pytest_configure(config: pytest.Config) -> None:
             returncode=pytest.ExitCode.USAGE_ERROR,
         )
 
-    # The unsupported-fork set is computed lazily in `get_unsupported_forks`
-    # (see its docstring for why it is deferred out of `pytest_configure`).
 
 
 def get_unsupported_forks(
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
@@ -628,7 +628,6 @@ def pytest_configure(config: pytest.Config) -> None:
         )
 
 
-
 def get_unsupported_forks(
     config: pytest.Config,
 ) -> FrozenSet[Fork | TransitionFork]:
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
@@ -24,7 +24,7 @@ from typing import (
 
 import pytest
 from _pytest.mark.structures import ParameterSet
-from pytest import Mark, Metafunc
+from pytest import Mark, Metafunc, StashKey
 
 from execution_testing.client_clis import TransitionTool
 from execution_testing.forks import (
@@ -45,6 +45,10 @@ from execution_testing.logging import (
 )
 
 logger = get_logger(__name__)
+
+# Session-scoped cache for the lazily-computed unsupported-fork set
+# (see `get_unsupported_forks`).
+unsupported_forks_key: StashKey[FrozenSet[Fork | TransitionFork]] = StashKey()
 
 
 def pytest_addoption(parser: pytest.Parser) -> None:
@@ -636,15 +640,13 @@ def get_unsupported_forks(
     """
     Return the selected forks not supported by the configured t8n tool.
 
-    The result is computed once and cached on ``config``. Computation is
+    The result is computed once and cached in ``config.stash``. Computation is
     deferred out of ``pytest_configure`` (where it previously lived) so that
     the ``ethereum`` package, imported when the t8n tool is queried, is only
     imported after pytest-cov has started the xdist worker's coverage session.
     Importing it earlier left it "previously imported, but not measured".
     """
-    cached: FrozenSet[Fork | TransitionFork] | None = getattr(
-        config, "_unsupported_forks", None
-    )
+    cached = config.stash.get(unsupported_forks_key, None)
     if cached is not None:
         return cached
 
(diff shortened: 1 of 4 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
```diff
@@ -622,18 +627,41 @@ def get_fork_option(
             returncode=pytest.ExitCode.USAGE_ERROR,
         )
 
-    config.unsupported_forks: Set[Fork | TransitionFork] = set()  # type: ignore
+    # The unsupported-fork set is computed lazily in `get_unsupported_forks`
+    # (see its docstring for why it is deferred out of `pytest_configure`).
```
spencer-tb:
```suggestion
```

packages/testing/src/execution_testing/client_clis/clis/execution_specs.py
```diff
@@ -49,18 +48,42 @@ def __init__(
...
+            self._fork_cache.__exit__()
 
     def version(self) -> str:
         """Version of the t8n tool."""
-        return ethereum.__version__
+        # Read the version from the installed package metadata instead of
+        # `ethereum.__version__`. This method is called during the filler's
+        # `pytest_configure`, which on xdist workers runs before pytest-cov
+        # starts measuring; importing `ethereum` here (even lazily) would make
+        # coverage report it as "module-not-measured".
```
spencer-tb:
```suggestion
        # Use package metadata rather than `ethereum.__version__` to avoid
        # importing `ethereum` here (see `__init__` for why it must stay lazy).
```

spencer-tb (approved):
LGTM, minus some comments. Feel free to merge after updating the comments :)

packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
danceratopz: Thanks, I should get stricter on these. The noise/repetition gets bad.

PR #3060, opened 2026-06-29
refactor(test-benchmark): speed-up the keccak max-permutations search

danceratopz:
## 🗒️ Description

`test_keccak_max_permutations` runs a discovery loop over `range(1, 1_000_000, 32)` to find the input size that maximizes keccak permutations per block. Each of the 31,250 iterations rebuilt the `Op.POP(Op.SHA3(Op.PUSH0, Op.DUP1, data_size=i))` bytecode and recomputed `.gas_cost(fork)`, which dominated the fill at roughly 164s per search.

Only SHA3's dynamic per-word cost depends on the input size `i`, so the per-call gas is affine in the keccak word count: `base + per_word * ((i + 31) // 32)`. This computes `base` once (the gas cost at `data_size=0`) and reads `per_word` from the fork's `GasCosts`, then adds the per-word term arithmetically in the loop. A guard asserts the affine model against the real `.gas_cost(fork)` for sample input sizes, so a future fork that reprices keccak non-linearly fails loudly instead of silently discovering a different optimum.

The discovered `optimal_input_length` is unchanged, so the filled fixtures are byte-identical: filling before and after into separate directories and running `uv run hasher compare` reports no differences at the root or per test. The search drops from roughly 164s to roughly 0.03s, which primarily speeds up release generation: The release benchmark matrix fills this test 21 times (7 gas values times 3 formats), each previously paying the full search.

## 🔗 Related Issues or PRs

Complements #3057 and #3058. Relates to #2673.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/tests/benchmark/compute/instruction/test_keccak.py
+++ b/tests/benchmark/compute/instruction/test_keccak.py
@@ -35,13 +35,11 @@ def test_keccak_max_permutations(
         """Return the keccak attack block hashing `input_length` bytes."""
         return Op.POP(Op.SHA3(Op.PUSH0, Op.DUP1, data_size=input_length))
 
-    # The per-call gas of `POP(SHA3(PUSH0, DUP1, data_size=i))` is affine in
-    # the keccak word count `(i + 31) // 32`: only SHA3's dynamic per-word
-    # cost depends on the input size `i`. Precompute the `i`-independent base
-    # once and add the per-word cost arithmetically below, rather than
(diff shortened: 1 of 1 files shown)


danceratopz:
This PR reduces `just bench-gas` in CI from ~26 mins to ~18 mins.

spencer-tb (approved):
LGTM!

PR #3066, opened 2026-06-30
feat(tests): EIP-7997 fork-transition test

barnabasbusa:
## 🗒️ Description

Adds a fork-transition test for [EIP-7997: Deterministic Deployment Factory](https://eips.ethereum.org/EIPS/eip-7997). There were no tests under `tests/amsterdam/` for EIP-7997.

EIP-7997 requires the chain state to include the factory at `FACTORY_ADDRESS` (`0x4e59…4956c`) with **nonce 1** and the canonical runtime code. Its Backwards Compatibility section states that on chains which already have the factory, *"satisfying this requirement is a no-op"*.

`test_existing_factory_preserved_across_transition` seeds the factory pre-fork with `pre_fork_nonce ∈ {1, 2, 32}` (1 = the EIP value; 2/32 = nonces accrued from earlier CREATE2 deployments), runs a pre-fork block + the Amsterdam transition block, and asserts the factory nonce is **unchanged** afterward.

A client that re-applies the EIP-7997 injection at the transition — resetting an already-deployed (and possibly already-used) factory back to nonce 1 — diverges its post-state root from every other client and halts at the transition block. This was a real bug found on `glamsterdam-devnet-6`: the factory reached nonce 32 by block 956, and the buggy client reset it to 1 at the Amsterdam activation block 957 (state root `0x1558…` vs canonical `0x90c6…`).

## ✅ Checklist

- [x] Fills cleanly (`fill --fork Amsterdam`, 6/6 pass against the reference EELS t8n).
- [x] `ruff format` / `ruff check` / `codespell` clean.
- [x] Reference spec version pinned to the latest `EIPS/eip-7997.md` commit.
- [x] Generated fixtures confirm the expected post-state nonce is preserved (e.g. `0x20` for the nonce-32 case).

Possible follow-up: a companion test that exercises the factory via CREATE2 across the fork (nonce continues `2 → 3`, deterministic address unchanged), proving it stays functional and not merely un-reset.


LouisTsai-Csie:
supersedes by PR #3077

danceratopz:
Entschuldigung Louis.

PR #3067, opened 2026-06-30
feat(tests): EIP-8037 CALL new-account state gas exact-fit boundary

chfast:
## 🗒️ Description

A value CALL to a non-alive account charges NEW_ACCOUNT state gas (STATE_BYTES_PER_NEW_ACCOUNT * COST_PER_STATE_BYTE = 183_600) in the caller frame. With an empty reservoir the charge spills into gas_left, so the caller must hold at least that much regular gas at the charge.

This test sets gas_limit so the charge lands exactly at the available gas (the CALL succeeds and the target is materialized) and one gas short (the caller frame goes out of gas, so the target is never created and the value transfer is rolled back). It is the CALL counterpart of test_code_deposit_state_gas_exact_fit_boundary.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
@@ -28,6 +28,7 @@ from execution_testing import (
     StateTestFiller,
     Storage,
     Transaction,
+    TransactionReceipt,
     compute_create2_address,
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
LGTM! Pinned the gas used with tx reciept for assertion check by framework when filling :D

PR #3070, opened 2026-06-30
feat(spec,tests): Implement EIP-8282

marioevz:
## 🗒️ Description

Implements the current spec iteration of this PR: https://github.com/ethereum/EIPs/pull/11760.

Matches glamsterdam-devnet-6 specs.

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist
    ```console
    just static
    ```

#### Cute Animal Picture


LouisTsai-Csie (approved):
Good from my side, some suggestion that could done in follow-up PRs:

- Add builder deposit / exit system contract to `eth_config` request.
- Do we need a `test_eip_mainnet.py` / fork transition test (so that we could test out )?

packages/testing/src/execution_testing/test_types/system_contract_request_types.py
```diff
@@ -389,3 +389,141 @@ def update_pre(self, pre: Alloc) -> Self:
             contract_address=contract_address,
             entry_address=entry_address,
         )
+
+
+# Scratch memory offsets for the out-of-gas measurement, placed above the
+# largest supported request calldata so they never overlap the copied calldata.
+_MEASURE_TOTAL_SLOT = 0x400
+_MEASURE_OVERHEAD_SLOT = 0x420
```
LouisTsai-Csie, line 397: It is possible that `self.extra_code` overrides these memory slots, but it is not a big problem now.

tests/amsterdam/eip8282_builder_execution_requests/test_modified_builder_contract.py
```diff
@@ -0,0 +1,171 @@
...
+    prepared = builder_exit_request_transaction.update_pre(pre=pre)
+    # get transaction list
+    txs: List[Transaction] = prepared.transactions()
+
+    blockchain_test(
+        pre=pre,
+        blocks=[
+            Block(
+                txs=txs,
+                requests_hash=Requests(*requests_list),
```
LouisTsai-Csie:
```suggestion
```

This looks like a no-op?

`requests_hash` on `Block` is a field inherited from `Header`. We can pass it into `Block.requests_hash`, but it is never verified against anything:

The header used for verification is rebuilt from the transition-tool (t8n) output.

https://github.com/ethereum/execution-specs/blob/2d68578e587cf9e2b5c1af2f61ff4c22f646b184/packages/testing/src/execution_testing/specs/blockchain.py#L891-L909

`requests_hash` is an output field, so it isn't a t8n input (it's not carried into the `Environment`) and it isn't merged into the re-built header either, that is, `header.requests_hash` always comes from the transition output.

https://github.com/ethereum/execution-specs/blob/2d68578e587cf9e2b5c1af2f61ff4c22f646b184/packages/testing/src/execution_testing/specs/blockchain.py#L933-L942

I've confirmed this by updating  to `Block(requests_hash=requests_hash=Requests(*[]),)`. It sill fills successfully and the fixture header keeps the real t8n value, so the field configuration is unused.

LouisTsai-Csie (approved):
I have one question related to `requests_hash `, otherwise it looks good! I like the design of `SystemContractInteractionMeasuredOutOfGasContract`, amazing!

packages/testing/src/execution_testing/test_types/system_contract_request_types.py
marioevz: Not really because the storage location is written and read in the same chunk of code before `extra_code`. So the memory is safe to overwrite then.

tests/amsterdam/eip8282_builder_execution_requests/test_modified_builder_contract.py
marioevz: Yep, and I can confirm that this is a copy-paste error because it's also in the source 7002 test file. We should fix this in the #3071.

spencer-tb (approved):
LGTM!

Feel free to add these before merge: https://github.com/ethereum/execution-specs/pull/3090

PR #3073, opened 2026-06-30
feat(tests): EIP-8037 CREATE-onto-alive refunds NEW_ACCOUNT to gas_left

chfast:
## 🗒️ Description

A CREATE2 onto an already-alive (pre-funded, code-less) address spills the NEW_ACCOUNT state charge from the empty reservoir into gas_left, succeeds, and refunds it. EIP-8037 source-based refunds are LIFO, so the spilled charge must return to gas_left, not the reservoir.

gas_limit is sized so the refunded NEW_ACCOUNT is exactly the gas left after the create, so the following SSTORE runs only if the refund landed in gas_left. A reservoir-only refund leaves gas_left at zero, out-of-gases the SSTORE, and diverges on the state root.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_create.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_create.py
@@ -2282,7 +2282,6 @@ def test_create_tx_collision_refunds_reservoir(
     )
 
 
-@pytest.mark.pre_alloc_mutable()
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
LGTM! :D

spencer-tb (approved):
Pushed a tweak to `fund_address` instead of `pre[target] =`

PR #3074, opened 2026-07-01
fix(tooling): honor exported `DOCC_SKIP_DIFFS` in `docs-spec` recipe

danceratopz:
## 🗒️ Description

This PR fixes a regression that has led to (`on_push`) docs builds taking 20 min instead of the previous ~8 min. This regression was introduced in #2756 (2026-04-28), which added the recipe parameter alongside the new `docs-spec-fast` shortcut.

The `docs-spec` recipe declared `$DOCC_SKIP_DIFFS=""`, whose empty parameter default shadows the environment variable the `Build Spec Docs` CI job exports, so the `DOCC_SKIP_DIFFS=1` set on pull requests and non-default-branch pushes never reached `docc` and the slow per-fork diff render always ran.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/.github/workflows/docs-build.yaml
+++ b/.github/workflows/docs-build.yaml
@@ -36,6 +36,7 @@ on:
       - "mkdocs.yml"
       - "uv.lock"
       - "pyproject.toml"
+      - "Justfile"
       - ".github/workflows/docs-build.yaml"
       - ".github/configs/docs-branches.yaml"
 
--- a/Justfile
+++ b/Justfile
@@ -308,7 +308,7 @@ export DYLD_FALLBACK_LIBRARY_PATH := if os() == "macos" { "/opt/homebrew/lib" } 
 
 # Generate documentation for EELS using docc
 [group('docs')]
-docs-spec $DOCC_SKIP_DIFFS="":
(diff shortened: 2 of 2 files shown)


.github/workflows/docs-build.yaml
```diff
@@ -36,6 +36,7 @@ on:
       - "mkdocs.yml"
       - "uv.lock"
       - "pyproject.toml"
+      - "Justfile"
```
jochem-brouwer, line 39: What does this do? Does it include the Justfile in the docs generator? Or is the Justfile added to trigger the docs build CI?

jochem-brouwer: Ah right, it is the latter.

Is there a reason to include the Justfile or was this used to check the docs build speed?

LouisTsai-Csie (approved):

.github/workflows/docs-build.yaml
danceratopz: I think the addition is just-ified, even if it'll trigger more often. If you create a fix in the Justfile for the docs build, you want to see that it's fixed in CI :)

PR #3087, opened 2026-07-02
feat(tests): EIP-7928 records an absent system contract in the BAL

chfast:
## 🗒️ Description

A pre-execution system call reads its target account unconditionally, before any code check, so the address is recorded in the block access list with an empty AccountChanges even when the contract was never deployed.

Adds `test_bal_2935_absent_contract` to the EIP-2935 file and `test_bal_4788_absent_contract` to the EIP-4788 file (alongside their deployed-contract BAL tests): overriding the genesis contract with an empty account drops it from the pre-state, and the block-start system call still records the absent address with an empty AccountChanges (`BalAccountExpectation.empty()`). 

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```


spencer-tb (approved):
LGTM, clean approve :)

PR #3093, opened 2026-07-03
perf(test-consume): align enginex engine-API flow and skip redundant genesis check

danceratopz:
## 🗒️ Description

Two enginex preamble changes to `test_via_engine.py`, developed together and kept in one PR because they touch the same code:

- `chore`: Align the enginex engine-API flow with the engine simulator. Enginex now always sends the initial forkchoice update to genesis and the forkchoice update after each valid payload instead of skipping them for reused clients, so both modes follow the same sequence.
- `perf`: Skip the redundant `getBlockByNumber(0)` genesis verification on reused clients. Genesis is immutable per client, so it is verified once per client instead of once per test, removing one RPC round-trip from the large majority of enginex tests.

Measured with go-ethereum and four workers over the full fixture release, skipping the genesis check cut wall-clock from `51:15` to `38:29` (about 25% faster) with identical results in both runs (`248 failed, 50316 passed`), confirming the change is behavior-preserving. Both runs used the same two commands, differing only in whether the genesis-skip was applied:

```console
./hive --dev --client go-ethereum --client-file ../execution-specs/.github/configs/hive/master.yaml --docker.buildoutput --sim.loglevel 5
uv run consume enginex --input=fixtures/ -n 4 -v
```

## 🔗 Related Issues or PRs
N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

cat

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/base.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/base.py
@@ -25,6 +25,18 @@ def eth_rpc(client: Client) -> EthRPC:
     return EthRPC(f"http://{client.ip}:8545")
 
 
+@pytest.fixture(scope="session")
+def genesis_verified_clients() -> set[str]:
+    """
+    Return the set of client ids whose genesis block has been verified.
+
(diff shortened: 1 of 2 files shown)


spencer-tb (approved):
LGTM! Changes/comments look clean to me!

PR #3094, opened 2026-07-03
fix(test-consume,test-rpc): close RPC sessions on teardown to fix fd leak

danceratopz:
## 🗒️ Description

The consume simulators created a `requests.Session` per test that was never closed, so against a reused enginex client the keep-alive sockets accumulated at two file descriptors per test until the worker hit its open-file limit and failed with `OSError: [Errno 24] Too many open files`. `BaseRPC` now owns its session lifecycle (a `close()` method plus context-manager support) and every consume `*_rpc` fixture releases it on teardown, which fixes the leak across the enginex, engine, rlp, sync, and build-block simulators. Unit tests cover the new session lifecycle (`close()` and context-manager exit).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

cat

### DIFF
--- a/packages/testing/src/execution_testing/rpc/rpc.py
+++ b/packages/testing/src/execution_testing/rpc/rpc.py
@@ -8,7 +8,16 @@ import time
 from contextlib import AbstractContextManager, nullcontext
 from itertools import count
 from pprint import pprint
-from typing import Any, Callable, ClassVar, Dict, List, Literal, Sequence
+from typing import (
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    List,
+    Literal,
+    Self,
+    Sequence,
+)
 
 import requests
 from jwt import encode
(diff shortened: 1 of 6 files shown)


packages/testing/src/execution_testing/rpc/rpc.py
```diff
@@ -210,6 +219,24 @@ def __init__(
         self.response_validation_context = response_validation_context
         self.session = requests.Session()
 
+    def close(self) -> None:
```
spencer-tb, line 222: Where is this used other than `__exit__()`?

spencer-tb (approved):
LGTM, only one comment. Non-blocking.

packages/testing/src/execution_testing/rpc/rpc.py
danceratopz: No-where else tbh, it was added becaues it matches standard resource-owner design like that used in `socket` and file objects. It also makes the public API complete (for users that don't want to use a context manager).

PR #3096, opened 2026-07-03
perf(spec-tools,ci): speed up json-loader and fill jobs

### LINKED ISSUE #3022: Move new-fork CLI tests out of the json-loader job
The json-loader job's phase 2 is dominated by two tests in `tests/json_loader/test_tools_new_fork.py` (`test_end_to_end[amsterdam]` and `[osaka]`, ~190-240s each). They test the `ethereum-spec-new-fork` CLI, not EELS or fixtures: they exercise `ethereum_spec_tools` + libcst and never import `ethereum.*`, so they add ~0 to the job's `--cov=ethereum`. 

They also pin to one xdist worker and run back-to-back, and no scheduler reliably splits them. (`tests/json_loader/test_tools.py`, the linter tests, is similarly misplaced.). This leads to worker starvation, slowing down the json-loader job.

**Fix:** run the tooling tests in their own job, parallel to json-loader: a `just test-new-fork` recipe (`pytest -n auto`) plus a `test-new-fork` workflow job. They must run under plain `pytest`, since `fill` errors on non-fixture tests (`No spec type format found in the test item`).

**Impact:** json-loader phase 2 ~7-8 min to ~1-2 min (~11 to ~5-6 min total); the tooling tests run concurrently, off the critical path.

**Bonus:** the CLI itself is slow (`new_fork/builder.py` runs ~12 libcst codemod passes, all `-j1`, each re-parsing the whole fork dir). Dropping `-j1` and scoping `SetConstant` to its target file would speed real fork creation too.

spencer-tb:
## 🗒️ Description

Speed up the per-PR CI by:

- Moving the new-fork CLI and linter tool tests out of the json-loader job (into `tests/evm_tools`, run by test-tests)
- Switch coverage to the `sys.monitoring` core
- Fill "one" fixture format per test in the fill jobs.

`json-loader` drops from ~7m to ~4m30s (was the current bottleneck) and `fill (amsterdam)` from ~5m30s to ~4m (the next bottleneck).

## 🔗 Related Issues or PRs

Additionally fixes #3022.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


Justfile
```diff
@@ -185,13 +185,18 @@ json-loader *args:
 # Run the testing package unit tests (with Python)
 [group('unit tests')]
 test-tests *args:
-    @mkdir -p "{{ output_dir }}/test-tests/tmp"
+    @mkdir -p "{{ output_dir }}/test-tests/tmp" "{{ output_dir }}/test-tests/tmp-evm-tools"
```
SamWilsn: Might be worth defining a Just variable for the temp dir, or better yet, use something like [`recipe_name`](https://github.com/casey/just/blob/c613447527aa4ab82ba3601ab059c61a25e5c434/README.md?plain=1#L4970) to factor out making temp dirs from all the recipes.

.github/workflows/test.yaml
```diff
@@ -105,9 +105,12 @@ jobs:
           python-version: "3.14"
       - uses: ./.github/actions/setup-env
       - name: Run fill (${{ matrix.label }})
-        run: just fill --from ${{ matrix.from_fork }} --until ${{ matrix.until_fork }}
+        run: >
+          just fill --from ${{ matrix.from_fork }} --until ${{ matrix.until_fork }}
+          -m "not slow and (blockchain_test or state_test_only or blockchain_test_engine_only)"
         env:
           PYTEST_XDIST_AUTO_NUM_WORKERS: auto
+          COVERAGE_CORE: sysmon
```
SamWilsn: Is there a reason you shouldn't define this in the Justfile? Seems like it would be just as useful for local builds.

danceratopz: Setting this here is redundant as sysmonis enabled by default in `coverage` 7.13.4 (which we use) with 3.14:
- https://github.com/coveragepy/coveragepy/blob/4f78d57f83ff8a4976043e8a8fcea24b91891840/coverage/env.py#L48-L49

So I agree with Sam and it makes all the more sense to enable locally (for all recipes in the `Justfile`), in case people haven't pinned 3.14.

Justfile
```diff
@@ -185,13 +185,18 @@ json-loader *args:
...
     cd packages/testing && uv run pytest \
         -n {{ xdist_workers }} \
         --basetemp="{{ output_dir }}/test-tests/tmp" \
         --ignore=src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_benchmarking.py \
         "$@" \
         src
+    uv run pytest \
```
danceratopz: Moving this out `json-loader` is a nice simple change and means we don't need `--dist=loadfile` for these tests, allowing them to run in parallel! 🚀 

However, they're not `test-tests`, they're `spec-tests`. I'd prefer a new recipe for it. That gets run in parallel with the others in CI. More targeted, less scope confusion.

Justfile
```diff
@@ -185,13 +185,18 @@ json-loader *args:
...
+        -n {{ xdist_workers }} \
+        --basetemp="{{ output_dir }}/test-tests/tmp-evm-tools" \
+        --ignore=tests/evm_tools/test_count_opcodes.py \
+        tests/evm_tools
```
danceratopz: Fix for arg forwarding:
```suggestion
    uv run pytest \
        -n {{ xdist_workers }} \
        --basetemp="{{ output_dir }}/test-tests/tmp-evm-tools" \
        --ignore=tests/evm_tools/test_count_opcodes.py \
        tests/evm_tools \
        "@"
```

.github/workflows/test.yaml
```diff
@@ -147,6 +150,7 @@ jobs:
         run: just json-loader
         env:
           PYTEST_XDIST_AUTO_NUM_WORKERS: auto
+          COVERAGE_CORE: sysmon
```
danceratopz: Unnecessary here, respectively move to `Justfile` (same as above comment).

.github/workflows/test.yaml
```diff
@@ -105,9 +105,12 @@ jobs:
           python-version: "3.14"
       - uses: ./.github/actions/setup-env
       - name: Run fill (${{ matrix.label }})
-        run: just fill --from ${{ matrix.from_fork }} --until ${{ matrix.until_fork }}
+        run: >
+          just fill --from ${{ matrix.from_fork }} --until ${{ matrix.until_fork }}
+          -m "not slow and (blockchain_test or state_test_only or blockchain_test_engine_only)"
```
danceratopz: This excludes the previously included `transaction_test` tests. This is the cleanest way to achieve it:
```suggestion
          -m "not slow and not derived_test"
```

Note, `derived_test` currently has a bug meaning that this would exclude our single `blockchain_test_engine_only` test 😆 , I'll add a fix...

Justfile
danceratopz: Nice idea @SamWilsn, but apparently `recipe_name()` is dynamic, so it can't move the `mkdir` into a shared recipe.

Alternative:
```just
[private]
_tmp name:
    @mkdir -p "{{ output_dir }}/{{ name }}/tmp"

json-loader *args: (_tmp "json-loader")
    uv run fill ...
```

If we bumped min version from 1.43 to 1.53 (which I'd prefer to avoid), I think we could write:
```just
json-loader *args: (_tmp recipe_name())
```

This is a follow-up issue/PR.

danceratopz (commented):
Niiice that you also hit this optimization! I found this trying to speed-up `json-loader` in #3010 and added #3022 as a follow-up! 

Few comments below!

.github/workflows/test.yaml
danceratopz: This orthogonal fix has landed 
- https://github.com/ethereum/execution-specs/pull/3108

spencer-tb: Applied!

.github/workflows/test.yaml
spencer-tb: Applied!

.github/workflows/test.yaml
spencer-tb: Applied!

Justfile
spencer-tb: New recipe applied `evm-tools`! We are cooking 🥘

Justfile
spencer-tb: Done!

spencer-tb:
@danceratopz @SamWilsn I've addressed you comments in https://github.com/ethereum/execution-specs/pull/3096/commits/fc7fafc6a08d743f64cfcec947bfc8f0bcfb67ba

#3022 should be fully resolved the addition in https://github.com/ethereum/execution-specs/pull/3096/commits/2d8301b3e029837968fde39166bcd54afbadf81f

PTAL! :)

Justfile
danceratopz: Could we call it `spec-tools`? To match the label? I would change it from `spec-tool` to `spec-tools`.

danceratopz: Thanks, label renamed from `A-spec-tool` to `A-spec-tools`.

danceratopz (approved):
Thanks for this!

Justfile
danceratopz: Create follow-up here:
- https://github.com/ethereum/execution-specs/issues/3112

.github/workflows/test.yaml
danceratopz: Thanks for the suggestion here, @SamWilsn!

PR #3099, opened 2026-07-03
refactor(spec-specs): credit CREATE refund before child incorporation

chfast:
## 🗒️ Description

On the CREATE success path, generic_create credited the NEW_ACCOUNT state-gas refund (target-already-alive case) after incorporate_child_on_success, which folds the child's spill into the parent. The refund's LIFO gas_left/reservoir split was therefore computed against the combined parent+child spill, unlike the error path (whose incorporate_child_on_error does not fold the child's spill).

Credit before incorporating the child so the refund reverses only the parent's own NEW_ACCOUNT spill, matching the error path. The split is tx-unobservable: the total refund is preserved, and over-cap frames carry gas_left near TX_MAX_GAS_LIMIT so the routing never gates a downstream charge. Verified by filling the eip8037 suite: 1567 fixtures byte-for-byte identical before and after.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

- [ ] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) and [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/):
    ```console
    just static
    ```
- [ ] All: PR title have the form `<type>(<area>):`, where `<type>` and `<area>` come from an approrpriate `C-<type>`, respectively `A-<area>`, label. The title should match the a target squash commit message.
- [ ] Ported Tests: Add the following docstring to manually enhanced tests from `./tests/ported_static/`:
    ```text
	@manually-enhanced: Do not overwrite. Post-state expectations corrected
	manually (see PR #2784).
	````

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/instructions/system.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/system.py
@@ -162,9 +162,12 @@ def generic_create(
         evm.return_data = child_evm.output
         push(evm.stack, U256(0))
     else:
-        incorporate_child_on_success(evm, child_evm)
         if target_alive:
+            # Target already existed: no new account, refund the state gas.
+            # Credit before incorporating the child so the refund reverses
+            # the parent's own spill only (as the error path does).
             credit_state_gas_refund(evm, StateGasCosts.NEW_ACCOUNT)
+        incorporate_child_on_success(evm, child_evm)
         evm.return_data = b""
         push(evm.stack, U256.from_be_bytes(child_evm.message.current_target))
 


src/ethereum/forks/amsterdam/vm/instructions/system.py
```diff
@@ -162,9 +162,12 @@ def generic_create(
         evm.return_data = child_evm.output
         push(evm.stack, U256(0))
     else:
-        incorporate_child_on_success(evm, child_evm)
         if target_alive:
+            # Target already existed: no new account, refund the state gas.
+            # Credit before incorporating the child so the refund reverses
+            # the parent's own spill only (as the error path does).
```
gurukamath, line 168: If the parent's `StateGasCosts.NEW_ACCOUNT` charge is covered by the `state_gas_reservoir` but the child's initcode spills state charges into `gas_left` and then successfully deploys onto an already-alive address, the two orderings route the refund differently:

1. The old code credits the child's spill back into `gas_left` while  refunding `StateGasCosts.NEW_ACCOUNT`
2. The new code credits everything in `StateGasCosts.NEW_ACCOUNT` to the reservoir. 

Since the `GAS` opcode excludes the reservoir, this is observable behaviour. Probably needs a test case.

gurukamath:
Also, should we do the refund first in `generic_call` as well?

chfast:
Added test, let's review that one first: https://github.com/ethereum/execution-specs/pull/3163.

gurukamath:
> Added test, let's review that one first: #3163.

#3163 merged

spencer-tb:
Accidental close?

chfast:
No. Not equivalent as proven by the test added in #3163.

PR #3100, opened 2026-07-03
feat(ci): fill mainnet fixtures nightly, draft cached `tests@` releases

spencer-tb:
## 🗒️ Description

**TLDR:** Fill the fixtures for mainnet forks nightly through the release workflow, and draft a `tests@` release on demand from any nightly artifact in minutes.

Run the fixture release workflow on a nightly cron (02:00 UTC). It fills the mainnet `tests` feature (all tests, slow included, all fixture formats) through the exact release pipeline but stops after `combine`: No tag and no release, just a rotating `fixtures_<commit>` workflow artifact with 5 day retention that is always available.

A unit-tested gate skips quiet nights with no new commits since the last nightly that actually filled. A failed or skipped nightly never advances that baseline, so no commit slips through unfilled, and a quiet stretch still refills after 4 days (or once the artifact has expired) so a live artifact always exists.

The nightly doubles as a `tests@` release candidate on demand: Dispatching the workflow with `cached=true` (or `commit=<sha>` to pick a specific nightly) skips `build` and `combine` and drafts the release from the nightly's tarball in minutes, tagged at the commit the nightly built. A unit-tested resolver guards the shortcut: `tests` on the default branch only, the version must exceed the newest `tests@` tag, the nightly must contain the previous release and be an ancestor of the current head. The step summary lists any commits the release will not contain.

Also extracts a shared `just fill-release` recipe, moves the evm-override validation into `generate_build_matrix.py`, removes a duplicate `benchmark` entry from `evm.yaml`, and exports `PYTEST_XDIST_AUTO_NUM_WORKERS` empty from the Justfile so pytest-xdist stops warning on non-numeric values such as `auto`.

**Follow-ups**

Deliberately not included: Dev-fork nightly fills (#3178) and failure notifications for red nightlies.

## 🔗 Related Issues or PRs

Complements the per-PR fill trim in #3096.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture


.github/workflows/nightly-fill.yaml
```diff
@@ -0,0 +1,74 @@
...
+            until_fork: Cancun
+          - label: prague
+            from_fork: Prague
+            until_fork: Prague
+          - label: osaka
+            from_fork: Osaka
+            until_fork: Osaka
+          - label: amsterdam
+            from_fork: Amsterdam
+            until_fork: Amsterdam
```
danceratopz: Would be nice to remove this duplication if possible.

.github/workflows/nightly-fill.yaml
danceratopz, line 1: What do you think about publishing an artifact here?
Will mostly be ignored, but could be useful. Use cases:
1. Validating clients against new incoming tests before publish.
2. We have a test that regressed. It fills correctly, invalidity is verifiable via trace. These artifacts could be used to pin down the regression (maybe we can't blame the test itself; it's in the framework).

.github/workflows/nightly-fill.yaml
```diff
@@ -0,0 +1,74 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      run: ${{ steps.check.outputs.run }}
+    steps:
+      - name: Check for commits in the last 25 hours
+        id: check
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          since=$(date -u -d "25 hours ago" +%Y-%m-%dT%H:%M:%SZ)
```
danceratopz: It could be useful to get the ref since the last successful build and then provide a list in the github summary fo these commits. Or perhaps there's an easier mechanism in github for this; I guess I'm looking for exactly the output that Github auto-generates in the release notes since the last tag.

danceratopz (commented):
Could we add a cron trigger to the release workflow and avoid triggering the `release` stage/tagging if trigged by cron? Or are you trying to be considerate of runner usage? Being kind to the gigachungi is valid; just asking!

Otherwise, couple of comment below.

.github/workflows/nightly-fill.yaml
spencer-tb: Removed!

.github/workflows/nightly-fill.yaml
spencer-tb: Nice one! Added as `fixtures_<commit>`.

.github/workflows/nightly-fill.yaml
spencer-tb: Here we go: https://github.com/ethereum/execution-specs/pull/3100/changes#diff-ea2d5c7ba7c25605936f749d3af7e29f20db8479e7141c34f9d64609d6b9fa12R31

spencer-tb:
Using rotating artifacts with FIFO of size 5. Ready for re-review :)

.github/workflows/nightly-fill.yaml
```diff
@@ -0,0 +1,163 @@
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh run download "${{ github.run_id }}" -p "fixtures__*" --dir split \
+            || echo "No per-range fixture artifacts found."
+      - name: Merge into a single fixtures tree
+        run: |
+          mkdir -p "fixtures_${{ steps.sha.outputs.short }}"
+          for d in split/fixtures__*/; do
+            [ -d "$d" ] || continue
+            rsync -a "$d" "fixtures_${{ steps.sha.outputs.short }}/"
```
danceratopz: I think we overwrite previously sync'd `.meta/index.json` files here. This is already solved in the main fixture release workflow.

.github/workflows/nightly-fill.yaml
```diff
@@ -0,0 +1,163 @@
...
+      actions: write
+    steps:
+      - name: Resolve short commit SHA
+        id: sha
+        run: echo "short=${GITHUB_SHA:0:8}" >> "$GITHUB_OUTPUT"
+      - name: Download per-range fixtures
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh run download "${{ github.run_id }}" -p "fixtures__*" --dir split \
```
danceratopz: Apparently, `gh` resolves the target repo from a git remote in the cwd or `GH_REPO` var. This means this step would fail before even calling the API. Easy fix, set `GH_REPO`.

danceratopz (commented):
Thanks for fixing these up! It still didn't sit right with me to add a new workflow that almost duplicates the workflow that we already had. I understand that the priority is to ensure that tests fill correctly, but by re-using the existing workflow, we can test that much more of the release machinery is working in the nightly. Also, if/when we move the release flow to pytest-split, this will also be present in the nightly, etc. I PR'd this suggestion so that `.github/workflows/release_fixtures.yaml` is used for nightly builds, too.

As a follow-up we could consider how we can create `tests@..` from the nightly. Then we can basically have mainnet releases ready/on-demand.

https://github.com/spencer-tb/execution-specs/pull/28

danceratopz (commented):
I think we're super close! Thanks a lot for the bug fix regarding "successful" empty runs (as no new commits)!


I think your last commit that removes the `nightly` features and simply uses `tests@` as the nightly build makes a bunch of sense, let's do that!

There's just one thing missing imo to make this a killer feature that I think is worth shipping with this PR: The ability to basically one-click create a release from the last successful nightly artifact. I added a "Promote nightly" workflow here:
- https://github.com/spencer-tb/execution-specs/pull/31

Using `tests` reduces coverage slightly (over the previous version), but I think that getting that back will require a follow-up. We could run multiple nightlies of relevant branches. But this is fine for now, especially at the late stage of Amsterdam fork dev and the fact that we'll probably bump the `tests` release feature to `Amsterdam` pretty soon Follow-up issue:
- https://github.com/ethereum/execution-specs/issues/3178

I also had this idea to make the nightly flow triggerable for easier testing but this is less important, just close if it's too much noise.
- https://github.com/spencer-tb/execution-specs/pull/32

spencer-tb:
@danceratopz I pushed to 31 as discussed!

danceratopz (approved):
Thanks @spencer-tb!! Enjoyed working on this one with you!

Please excuse me sneaking in the last commit. It was totally unrelated, but it does silence an irrelevant error.

spencer-tb:
Looks good thx! Just updating the PR desc/title to reflect everything then will merge :)

PR #3101, opened 2026-07-03
perf(spec-tools,ci): parallel PR spec-doc builds, publish-only social cards

spencer-tb:
## 🗒️ Description

Aims to improve **per PR** docs CI workflows.

Spec doc builds now run as four parallel docc shards over consecutive fork sets cutting `Build Spec Docs` to ~4m. 

MkDocs social cards are now publish only, trimming `Build HTML Docs` to ~5mins. 

Note, forward / cross-shard references are validated only by the serial default-branch build, and optionally log what PruneReferencesTransform prunes so the coverage boundary is visible in CI, more info in https://github.com/ethereum/execution-specs/pull/3101#discussion_r3529001737.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

cat


mkdocs.yml
```diff
@@ -22,7 +22,10 @@ plugins:
             members_order: source
             group_by_category: false
   - search
-  - social
+  # Social cards render a PNG per page (~2m for the full site) and need
+  # native cairo; only publish builds enable them (see docs-build.yaml).
+  - social:
+      enabled: !ENV [DOCS_SOCIAL_CARDS, false]
```
danceratopz, line 28: Great idea!

Justfile
```diff
@@ -316,6 +316,37 @@ docs-spec $DOCC_SKIP_DIFFS=env_var_or_default("DOCC_SKIP_DIFFS", ""):
...
 
+# Build spec docs in parallel shards of consecutive forks (PR validation).
+# Each shard overlaps its predecessor fork by one so every previous-fork
+# reference is validated in some shard; outputs are per-shard and not merged.
+[group('docs')]
+docs-spec-parallel shards="4":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    read -ra forks <<< "$(uv run python -c 'from ethereum_spec_tools.forks import Hardfork; print(" ".join(h.short_name for h in Hardfork.discover()))')"
+    total=${#forks[@]}
```
danceratopz: Ensure we fail loudly if we get an empty list of forks for whatever reason. 
```suggestion
    read -ra forks <<< "$(uv run python -c 'from ethereum_spec_tools.forks import Hardfork; print(" ".join(h.short_name for h in Hardfork.discover()))')"
    total=${#forks[@]}
    [ "$total" -gt 0 ] || { echo "fork discovery failed" >&2; exit 1; }
```

Justfile
```diff
@@ -316,6 +316,37 @@ docs-spec $DOCC_SKIP_DIFFS=env_var_or_default("DOCC_SKIP_DIFFS", ""):
 [group('docs')]
 docs-spec-fast: (docs-spec "1")
 
+# Build spec docs in parallel shards of consecutive forks (PR validation).
+# Each shard overlaps its predecessor fork by one so every previous-fork
+# reference is validated in some shard; outputs are per-shard and not merged.
```
danceratopz: Codex found this, Claude confirmed it and provided this explanation....

This "every previous-fork reference is validated" invariant holds only for **backward** references. Forward / multi-fork references get pruned in every shard that renders the source (via `PruneReferencesTransform`), so they are never validated on PRs.

Concretely today, `src/ethereum/forks/prague/transactions.py:472-476` forward-references amsterdam:

```
See [`has_access_list`][hal] and [`Access`][a] for more details.
[hal]: ref:ethereum.forks.amsterdam.transactions.has_access_list
[a]: ref:ethereum.forks.amsterdam.transactions.Access
```

With the default 24 forks / 4 shards, prague lands in shard 2 (`arrow_glacier..osaka`) and amsterdam only in shard 3, so both links are pruned in prague's shard and never checked. A rename of `has_access_list` / `Access` in amsterdam would pass PR CI here and only fail post-merge on the serial `docs-spec` build (which gates the docs deploy). A scan of every `ethereum.forks.*` reference confirms this is the only such pair today (2 sites).

It's a small, self-correcting regression from the old full-`docs-spec` PR step. Two ways to resolve:

- **Accept it** — reword this comment to say forward / cross-shard references are validated only by the serial default-branch build, and optionally `log` what `PruneReferencesTransform` prunes so the coverage boundary is visible in CI.
- **Close it** — include the tip fork in every shard's `DOCC_ONLY_FORKS` (every cross-fork reference in the tree currently points at the newest fork), at a small per-shard rendering cost.

Justfile
```diff
@@ -316,6 +316,37 @@ docs-spec $DOCC_SKIP_DIFFS=env_var_or_default("DOCC_SKIP_DIFFS", ""):
...
+        echo "shard $i: $shard"
+        DOCC_SKIP_DIFFS=1 DOCC_ONLY_FORKS="$shard" \
+            uv run docc --output "{{ output_dir }}/docs-spec-parallel/shard-$i" &
+        pids+=($!)
+    done
+    fail=0
+    for pid in "${pids[@]}"; do
+        wait "$pid" || fail=1
+    done
+    exit $fail
```
danceratopz: This is a pretty ugly and hard-to-maintain bash block in the `Justfile` can we make this pure Python?

Suggestion here: 
- https://github.com/spencer-tb/execution-specs/pull/24

Justfile
danceratopz: This comment also serves as the doc for the recipe in `just --list` (or just `just` with our config). Multiline comments seem to get jumbled. This was also addressed in:
- https://github.com/spencer-tb/execution-specs/pull/24

danceratopz (commented):
Thanks for tackling this. Adds quite some complexity, but it really does help shorten runtimes. Few comments below. I made one suggestion directly in:

- https://github.com/spencer-tb/execution-specs/pull/24

spencer-tb:
Merged your PR in! Thx

Justfile
danceratopz: This is unnecessary with:
- https://github.com/spencer-tb/execution-specs/pull/24

Justfile
danceratopz: I think we can accept the compromise.

danceratopz: Added a comment to the PR desc.

danceratopz (commented):
There's one clean-up required that was missed in https://github.com/spencer-tb/execution-specs/pull/24. Let me commit that real quick.

tests/evm_tools/test_docc_shards.py
danceratopz, line 1: This is currently not ran as part of CI.

If we move this to `tests/evm_tools/test_docc_shards.py`, then we'd collect and run this as part of the new `just spec-tools` recipe that we added as part of:
- #3096

danceratopz: Done.

danceratopz (approved):
Thanks for this!

Confirmed that the last commit does actually include these tests in the `spec-tools` CI. All good!

PR #3107, opened 2026-07-06
fix(test-rpc): bound JSON-RPC requests with a default timeout

yperbasis:
## 🗒️ Description

`BaseRPC` sent every JSON-RPC request with no HTTP timeout (`post_request`/`post_batch_request` default `timeout=None` passed straight to `requests`). A request whose packets get silently dropped therefore blocks until the kernel abandons TCP retransmission — **~15.4 minutes** on Linux (`tcp_retries2=15`) — even when the client is perfectly healthy.

We hit exactly this in erigon's hive-eest CI ([run](https://github.com/erigontech/erigon/actions/runs/28774682425/job/85315814600)): under heavy container churn (~12k containers per run, IPs recycled every ~1.5 s), a `consume engine` test's `engine_forkchoiceUpdatedV3` was black-holed by docker network state churn milliseconds after a successful `engine_newPayloadV4` on the same keep-alive connection. The client's logs prove the request never arrived; the sim wedged for 925 s, and by the time the kernel gave up, the tenacity retries — which open fresh connections and would have succeeded — fired too late. One flaky flow, one failed test out of 20964, red CI.

What this PR does:

- Every request now carries a default `(connect=10 s, read=300 s)` timeout. The `None` → default resolution lives in `_make_request`, the single transport chokepoint, so no call path can bypass the bound. `DEFAULT_REQUEST_TIMEOUT` and the `TimeoutType` alias are exported from `execution_testing.rpc`.
- The bound is overridable per client (`request_timeout=...` on any `BaseRPC` subclass, `ChainBuilderEthRPC` included; `None` disables it entirely) and per call (existing `timeout=` argument, unchanged semantics except `None` now means "use the client default").
- `_make_request` retries `requests.Timeout` the same way it already retries `ConnectionError` (`ConnectTimeout` subclasses `ConnectionError`, but `ReadTimeout` does not, so read timeouts previously escaped the retry entirely). Re-sending cannot corrupt state: the methods used by the framework either read state or re-broadcast the same signed transaction. At worst, a re-sent transaction whose first delivery succeeded draws a duplicate-transaction error instead of its hash — a flow that had already lost its response — while retrying recovers every flow where the request itself was lost.

With this in place, a black-holed flow recovers on a fresh connection after at most ~300 s instead of failing the test after ~15 minutes.

**Risk note:** any single request that legitimately needs longer than 300 s between response bytes will now time out and be retried (5 attempts). The default is deliberately generous for slow clients / huge payloads; happy to tune it or plumb it through simulator config if maintainers prefer a different bound.

Includes unit tests for the default, the per-client and per-call overrides (including forwarding through `ChainBuilderEthRPC`), the `None` escape hatch, read-timeout retry (without real sleeps), and the batch path.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) and [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/) (scoped to the touched package: `ruff check`, `ruff format --check`, `mypy`, and the `rpc` module's tests — 10 passed).

#### Cute Animal Picture

A cat to keep the request company while it waits, briefly

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
@@ -19,7 +19,12 @@ from execution_testing.base_types import (
 )
 from execution_testing.client_clis.cli_types import EnginePayloadMetadata
 from execution_testing.forks import Fork, TransitionFork
-from execution_testing.rpc import EngineRPC, TestingRPC
+from execution_testing.rpc import (
+    DEFAULT_REQUEST_TIMEOUT,
+    EngineRPC,
+    TestingRPC,
+    TimeoutType,
+)
 from execution_testing.rpc import EthRPC as BaseEthRPC
 from execution_testing.rpc.rpc_types import (
     ForkchoiceState,
@@ -55,6 +60,7 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
         initial_forkchoice_update_retries: int = 5,
         transaction_wait_timeout: int = 60,
         max_transactions_per_batch: int | None = None,
+        request_timeout: TimeoutType = DEFAULT_REQUEST_TIMEOUT,
         testing_rpc: TestingRPC | None = None,
     ):
         """Initialize the Ethereum RPC client for the hive simulator."""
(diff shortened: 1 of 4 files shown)


spencer-tb (approved):
LGTM! I have no changes to request here :)

Will leave @danceratopz for a review and merge!

danceratopz (approved):
Thanks for this! It's somewhat of an edge case, but Codex and I were slightly concerned that, for `execute`, in the case of a transaction that was successfully submitted (i.e., known to the client) but who's http response got black-holed, the transaction would fail upon being resubmitted, despite being correctly known to the client. I added two commits to account for this case and some unit tests. A genuinely rejected transaction will still correctly raise an exception.

PR #3108, opened 2026-07-06
fix(test-fill): fix `derived_test` marking for single-format tests

danceratopz:
## 🗒️ Description
`pytest_generate_tests` marked every fixture format after the first in `supported_fixture_formats` as `derived_test` purely by list position, so a test whose first format is dropped, either by a single-format marker such as `blockchain_test_engine_only` or by a session-level format filter such as `--generate-pre-alloc-groups`, had all of its surviving formats marked `derived_test` and was deselected entirely by `-m "not derived_test"`. This PR applies the spec-level marker discard directly in `pytest_generate_tests` so that the first generated parameter is the primary and remains unmarked, drops the unused `fork` argument from `BaseTest.discard_fixture_format_by_marks()`, and adds pytester tests covering both scenarios.

## 🔗 Related Issues or PRs
Unblocks the use of `-m "not slow and not derived_test"` in #3096, see [this review comment](https://github.com/ethereum/execution-specs/pull/3096#discussion_r3527529480).

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

I'm the primary now

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -26,6 +26,7 @@ from typing import TYPE_CHECKING, Any, Dict, Generator, List, Self, Set, Type
 import pytest
 import xdist
 from _pytest.compat import NotSetType
+from _pytest.mark.structures import ParameterSet
 from _pytest.terminal import TerminalReporter
 from filelock import FileLock
 from pytest_metadata.plugin import metadata_key
@@ -1779,20 +1780,36 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     FillerFile.collect() in ./static_filler.py for more details.
     """
     session: FillingSession = metafunc.config.filling_session  # type: ignore[attr-defined]
+    markers = list(metafunc.definition.iter_markers())
     for test_type in BaseTest.spec_types.values():
         if test_type.pytest_parameter_name() in metafunc.fixturenames:
-            parameters = []
-            for i, format_with_or_without_label in enumerate(
-                test_type.supported_fixture_formats
-            ):
+            parameters: List[ParameterSet] = []
+            for (
+                format_with_or_without_label
+            ) in test_type.supported_fixture_formats:
                 if not session.should_generate_format(
                     format_with_or_without_label
+                ):
+                    continue
+                fixture_format = (
+                    format_with_or_without_label.format
+                    if isinstance(
+                        format_with_or_without_label, LabeledFixtureFormat
+                    )
+                    else format_with_or_without_label
+                )
+                if test_type.discard_fixture_format_by_marks(
+                    fixture_format, markers
                 ):
                     continue
                 parameter = labeled_format_parameter_set(
                     format_with_or_without_label
                 )
-                if i > 0:
+                # The first surviving format is the test's primary; the
+                # rest are derived from it (e.g. a BlockchainTest derived
+                # from a StateTest) and can be deselected with
+                # `-m "not derived_test"`.
+                if parameters:
                     parameter.marks.append(pytest.mark.derived_test)  # type: ignore
                 parameters.append(parameter)
             metafunc.parametrize(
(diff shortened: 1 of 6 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -1779,20 +1780,36 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
...
                     format_with_or_without_label
                 )
-                if i > 0:
+                # The first surviving format is the test's primary; the
+                # rest are derived from it (e.g. a BlockchainTest derived
+                # from a StateTest) and can be deselected with
+                # `-m "not derived_test"`.
+                if parameters:
                     parameter.marks.append(pytest.mark.derived_test)  # type: ignore
                 parameters.append(parameter)
```
danceratopz, line 1814: A breakdown of the reworked loop for reviewers.

```python
markers = list(metafunc.definition.iter_markers())
```

The test function's full marker set (decorator, class, and module-level marks), hoisted once per test function since it doesn't vary per format.

Then, per entry in `supported_fixture_formats`, three gates in order:

1. **Session gate** (`should_generate_format()`, pre-existing): is this format being generated in this fill session at all? This is CLI/config-level (e.g. a `--generate-pre-alloc-groups` session only generates EngineX) and has nothing to do with the individual test.
2. **Marker gate** (`discard_fixture_format_by_marks()`, moved up from `pytest_collection_modifyitems`): does this test's own marker set veto the format? E.g. `blockchain_test_engine_only` returns `True` for `BlockchainFixture`, so the parameter is never created.
3. **Primary/derived split** (`if parameters:`): if anything already survived both gates, a primary exists, so this format is derived. The first survivor lands with `parameters` still empty and stays unmarked.

Concrete run for a test marked `blockchain_test_engine_only`: `BlockchainFixture` fails gate 2; `BlockchainEngineFixture` passes with `parameters` empty, so it is the primary and stays unmarked; `BlockchainEngineXFixture` passes with `parameters` non-empty, so it is marked `derived_test`; the sync format fails gate 2 (the engine-only check keeps only Engine/EngineX). For an unmarked test, the first format in the list is the primary and the rest are marked, identical to the old positional behavior, which is why the rest of the suite is unaffected.

The marker gate could move up here because the spec-level `discard_fixture_format_by_marks()` never actually used its `fork` argument (every override did `del fork`), and markers are available at generate time via `metafunc.definition.iter_markers()`: they are properties of the function/class/module, independent of parametrization. Everything genuinely fork-dependent stays in `pytest_collection_modifyitems`: the pre-Paris hive removal, `supports_fork()`, and the fixture-level discard (e.g. sync fixtures requiring `verify_sync`). The spec-level call there is now a no-op for dynamic tests but load-bearing for static fillers, which bypass `pytest_generate_tests` entirely.

Note that the old `if i > 0:` indexed the full format list, so in a fill session whose format filter excluded the primary (gate 1), every generated format was marked `derived_test`; now the first format surviving both gates is the unmarked primary. That is the second bug this PR fixes (see `test_not_derived_test_selects_session_filter_survivor`).

spencer-tb (changes requested):
The new tests before and after the fix pass. Can we update the tests to catch the bug?

`engine_only_survivor_is_primary` and `session_filter_survivor` only, others should pass on both.

Loop change looks good to me (from what I understand)!

danceratopz:
> The new tests before and after the fix pass. Can we update the tests to catch the bug?
> 
> `engine_only_survivor_is_primary` and `session_filter_survivor` only, others should pass on both.
> 
> Loop change looks good to me (from what I understand)!

Yeah, two tests are for reproducing the bug, the two others provide are new regression tests that also pass before.

```bash
src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_derived_test_marker.py::test_not_derived_test_selects_primary_survivor[normal_test_still_marks_derived] FILLED [2/4]
src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_derived_test_marker.py::test_not_derived_test_selects_primary_survivor[state_only_survivor_is_primary] FILLED
```
The first guards the fix against over-correcting (if the loop stopped marking derived formats entirely, only this case would catch it), and the second pins the single-format marker whose survivor coincides with the positional-first format, where the old behavior was already correct.

```
git checkout 6cbabb1f656
just test-tests src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_derived_test_marker.py -n 1
```
->
```
FAILED src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_derived_test_marker.py::test_not_derived_test_selects_primary_survivor[engine_only_survivor_is_primary] - AssertionError: Collection failed:
FAILED src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_derived_test_marker.py::test_not_derived_test_selects_session_filter_survivor - AssertionError: Collection failed:
==================================================================== 2 failed, 2 passed, 32 warnings in 3.44s ====================================================================
```
The warnings aren't from this PR; we should fix them as a follow-up.

spencer-tb:
Whoops, my "pass before and after" on the two repro tests were stale runs! Sorry!

spencer-tb (approved):
LGTM!

PR #3110, opened 2026-07-06
refactor(test-benchmark): deploy contract gas usage

LouisTsai-Csie:
### Problem
Stateful filler fails immediately, geth rejects the setup block with: `intrinsic gas too low: have 15000, want 21000`

### Root cause
The funding transaction gas limit is calculated as bare intrinsic cost (15,000), but funding txs transfer value to empty EOAs, which costs more:
- 21,000 regular gas (12,000 base + 3,000 cold account + 6,000 value transfer)
- 183,600 state gas (EIP-8037 account creation charge)

Total: 204,600, thus 15,000 limit is too low for any client.

## Fix

Calculate gas for what the transaction actually is (value transfer to empty account):

```python
intrinsic = fork.transaction_intrinsic_cost_calculator()(
    sends_value=True,
    recipient_type=RecipientType.EMPTY_ACCOUNT,
)
return intrinsic + fork.transaction_top_frame_state_gas(
    sends_value=True,
    recipient_type=RecipientType.EMPTY_ACCOUNT,
)
```

| Fork | Before | After |
|------|--------|-------|
| Amsterdam | 15,000 ❌ | **204,600** ✓ |
| Osaka/Prague | 21,000 | 21,000 (unchanged) |

### Verification

Please download benchmarkoor and add this configuration file, follow the instruction:

```
# Build benchmarkoor
make build-core

# Build datadirs and eest payloads
./bin/benchmarkoor build --config examples/configuration/simple.yaml --force
```

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -261,7 +261,12 @@ def _compute_deploy_gas_limit(
 
     # Regular portion, bound by the gas cap.
     regular_gas = intrinsic_regular_gas
-    regular_gas += deploy_code_size * gas_costs.CODE_DEPOSIT_PER_BYTE
+    if fork.state_gas_reservoir_enabled():
+        regular_gas += gas_costs.OPCODE_KECCAK256_PER_WORD * (
+            (deploy_code_size + 31) // 32
+        )
+    else:
+        regular_gas += deploy_code_size * gas_costs.CODE_DEPOSIT_PER_BYTE
     regular_gas += memory_expansion_gas_calculator(
         new_bytes=len(bytes(initcode))
     )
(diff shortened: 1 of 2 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -261,7 +261,12 @@ def _compute_deploy_gas_limit(
 
     # Regular portion, bound by the gas cap.
     regular_gas = intrinsic_regular_gas
-    regular_gas += deploy_code_size * gas_costs.CODE_DEPOSIT_PER_BYTE
+    if fork.state_gas_reservoir_enabled():
+        regular_gas += gas_costs.OPCODE_KECCAK256_PER_WORD * (
+            (deploy_code_size + 31) // 32
+        )
+    else:
+        regular_gas += deploy_code_size * gas_costs.CODE_DEPOSIT_PER_BYTE
```
LouisTsai-Csie, line 269: `_compute_deploy_gas_limit` was charging the full code-deposit cost (200/byte) to regular gas. For a 64 KiB contract, this exceeded the EIP-7825 cap (16.7M), causing deploys to be rejected.

However, under EIP-8037, the cost model changed: the per-byte charge (1,530/byte) moves to state-gas, while only the hash cost stays in regular gas. The fix: use the new model instead of the old 200/byte rule.

spencer-tb (approved):
LGTM!

spencer-tb:
execute `--sender-fund-refund-gas-limit` default is still `200_000`? it should be bumped i think

PR #3111, opened 2026-07-06
fix(specs,tests): check SSTORE access cost before the implicit read

nerolation:
## 🗒️ Description

EIP-7928 pre-state validation for `SSTORE` requires the slot's access cost to be covered, in addition to the EIP-2200 stipend sentry, before the implicit read of the current storage value. Since EIP-8038 raised `COLD_STORAGE_ACCESS` (3000) above `CALL_STIPEND` (2300), the sentry alone no longer implies the access cost is covered: a cold `SSTORE` in a frame with 2301–2999 gas passed the sentry, performed the implicit read, and recorded the slot in the block access list before failing in `charge_gas`. Storage reads survive frame rollback by design, so the unaffordable slot ended up in the BAL and its hash.

This PR checks the cold/warm access cost before the implicit read in the Amsterdam `SSTORE`, so such a slot is neither read nor recorded. Gas usage, refunds, and post-state are unchanged: the frame runs out of gas either way. The BAL is the only observable difference.

Test changes in `test_bal_sstore_and_oog`:

* `oog_at_eip2200_stipend_plus_1` now expects no BAL entry, previously expected a storage read.
* New `oog_at_cold_access_cost_minus_1` case: no BAL entry.
* New `oog_at_cold_access_cost` case: storage read in BAL, implicit `SLOAD` done, OOG on the write cost.

## 🔗 Related Issues or PRs

EIP-7928 wording update [here](https://github.com/ethereum/EIPs/pull/11854).

## ✅ Checklist

- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) and [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/).

* [x] All: PR title has the form `<type>(<area>):`, where `<type>` and `<area>` come from an appropriate `C-<type>` and `A-<area>` label. The title should match the target squash commit message.

* [x] All: Considered updating the online docs in the [[./docs/](https://chatgpt.com/ethereum/execution-specs/blob/HEAD/docs/)](/ethereum/execution-specs/blob/HEAD/docs/) directory.

* [ ] All: Set appropriate labels for the changes, only maintainers can apply labels.


nerolation:
Close in favor of https://github.com/ethereum/execution-specs/pull/3064

PR #3115, opened 2026-07-06
perf(spec-specs): Run Blake2 compression on plain `int` words

danceratopz:
## 🗒️ Description

`test_blake2b_large_gas_limit` (EIP-152, `rounds=100_000`) holds the top three slots in the fill run's slowest-test list at ~21s per fixture format. Profiling showed `Blake2.compress` spending nearly all of its time constructing `Uint` objects: every `+`/`^`/`>>`/`<<`/`%` in the mixing loop allocates and bounds-checks a new `Uint` (~43M constructions per 100k rounds), and the `max_word` property recomputed `Uint(2) ** self.w` on every access.

This PR runs the compression on plain `int` words:

- `compress` converts the `Uint` state and message words to `int` once on entry; `G` mixes plain `int`s. The arithmetic expressions in `G` are textually unchanged.
- The class configuration (`w`, `mask_bits`, `R1`..`R4`), `IV`, and `MIX_TABLE` hold plain `int`s; `max_word` and `w_R1`..`w_R4` become `cached_property`s.
- The public interface is unchanged: `get_blake2_parameters` still returns `Uint` values (the precompile's gas math relies on this) and `compress` keeps its signature.

`compress` on the 100k-round test vector drops from 66s to 3.4s; filling `tests/istanbul/eip152_blake2/` (`--fork Amsterdam -m "not derived_test"`) drops from 62s to 18s. Any other test touching the `blake2f` precompile benefits as well.

A follow-up commit removes the now-stale `slow` marker from `test_blake2b`: The worst of its 66 marked cases fills in 0.88s locally post-fix, in the same band as the module's unmarked tests, so the `-m "not slow"` CI jobs now cover the basic EIP-152 vectors. `test_blake2b_large_gas_limit` keeps the marker (its 100k-round cases still fill in 2.3-4.3s locally) and its 8M-round cases remain skipped; fixtures verified unchanged via `hasher`.

### Correctness

- Fixtures are byte-identical: Before/after fills of `tests/istanbul/eip152_blake2/` (`--fork Amsterdam -m "not derived_test"`, 73 fixtures) produce the same `hasher` root hash, `0x9e30eaf1a70989c1fedb752999c9494bbcfb55ecf6f1369d8f7ae3670b937573`.
- The old and new `compress` were also cross-checked standalone for equal output at rounds 0, 1, 12, 100, and 100k.

### Why the constants change type too

Two variants that keep partial `Uint` usage show that residual casts in the hot loop give back a large share of the win (10k rounds, identical output in all variants; absolute factors vary with machine load between sessions, the variant ordering is the stable result):

| Variant | Time | Speedup |
| ------- | ----: | ------: |
| Original (all `Uint`) | 1.073s | 1.0x |
| This PR | 0.100s | 10.7x |
| This PR, but `@property` instead of `@cached_property` | 0.155s | 6.9x |
| This PR, but `MIX_TABLE` entries as `Uint` | 0.179s | 6.0x |

`MIX_TABLE` values are used as list indices ~22 times per `G` call and each `Uint` subscript dispatches through `__index__`; `max_word` is read 6 times per `G` call, 4.8M times per 100k rounds. Beyond cost, the value domains cannot be half-converted: Mixed `int`/`Uint` arithmetic either raises or promotes results back into `Uint`, reinstating the per-operation construction this PR removes.

### Readability

- The `G` mixing body and all docstrings are unchanged, so the correspondence to RFC 7693 reads exactly as before, and the literal tables lose 32 `Uint(...)` wrappers.
- The trade-off: This module no longer gets `Uint`'s runtime unsigned-range checking; a comment in `compress` marks the deviation and why. The normative fork precompiles (`vm/precompiled_contracts/blake2f.py`) are untouched.
- Incidental fix: The old work-vector setup (`v[8:15] = self.IV` assigns 8 elements into a 7-element slice) silently grew the 16-element list to 17 with a dead trailing element; the new construction is exact.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Uint-free and fast

### DIFF
--- a/src/ethereum/crypto/blake2.py
+++ b/src/ethereum/crypto/blake2.py
@@ -2,6 +2,7 @@
 
 import struct
 from dataclasses import dataclass
+from functools import cached_property
 from typing import Final, List, Tuple, final
 
 from ethereum_types.numeric import Uint
@@ -40,48 +41,48 @@ class Blake2:
     https://datatracker.ietf.org/doc/html/rfc7693
     """
 
-    w: Uint
-    mask_bits: Uint
+    w: int
+    mask_bits: int
     word_format: str
 
-    R1: Uint
-    R2: Uint
-    R3: Uint
-    R4: Uint
+    R1: int
+    R2: int
+    R3: int
+    R4: int
 
-    @property
-    def max_word(self) -> Uint:
+    @cached_property
+    def max_word(self) -> int:
         """
         Largest value for a given Blake2 flavor.
         """
-        return Uint(2) ** self.w
+        return 2**self.w
 
-    @property
-    def w_R1(self) -> Uint:
+    @cached_property
+    def w_R1(self) -> int:
         """
         (w - R1) value for a given Blake2 flavor.
         Used in the function G.
         """
         return self.w - self.R1
 
-    @property
-    def w_R2(self) -> Uint:
+    @cached_property
+    def w_R2(self) -> int:
         """
         (w - R2) value for a given Blake2 flavor.
         Used in the function G.
         """
         return self.w - self.R2
 
-    @property
-    def w_R3(self) -> Uint:
+    @cached_property
+    def w_R3(self) -> int:
         """
         (w - R3) value for a given Blake2 flavor.
         Used in the function G.
         """
         return self.w - self.R3
 
-    @property
-    def w_R4(self) -> Uint:
+    @cached_property
+    def w_R4(self) -> int:
         """
         (w - R4) value for a given Blake2 flavor.
         Used in the function G.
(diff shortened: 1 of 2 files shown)


spencer-tb (commented):
Looks good from my side but would like @gurukamath to review/approve!

danceratopz:
> Looks good from my side but would like @gurukamath to review/approve!

Yes, thanks! I wanted to ask Guru whether we can/should accept the inconsistent cast style?

SamWilsn:
Is sacrificing the clarity around signedness worth the... minute of test time?

This also highlights the importance of optimizing `ethereum_types`. I wonder if the `mypyc` annotations that disable the garbage collector for `Uint` would help with the allocation overhead.

gurukamath (commented):
Would it be better to use `blake2b-py` library as an alternative pathway here?

danceratopz:
> Is sacrificing the clarity around signedness worth the... minute of test time?

Fair. Thanks for fighting for readability, @SamWilsn!

> This also highlights the importance of optimizing `ethereum_types`. I wonder if the `mypyc` annotations that disable the garbage collector for `Uint` would help with the allocation overhead.

Noted, will look into it.

danceratopz:
> Would it be better to use `blake2b-py` library as an alternative pathway here?

Great idea @gurukamath I didn't think to question the status quo! Yes, it's 640x faster than this PR. We out source keccak and secp256k1, so I guess it's fine to do it here. Follow-up:
- https://github.com/ethereum/execution-specs/issues/3181

However, there's no wheels for 3.14/pypy3.11 and the sdist doesn't compile. The repo was amongst those transfered to ApeWorX, I've PR'd it here:
- https://github.com/ApeWorX/blake2b-py/pull/28

included in:
- https://github.com/ApeWorX/blake2b-py/pull/29

danceratopz:
Closing in favor of #3181.

PR #3116, opened 2026-07-06
feat(spec-specs, tests): charge EIP-8037 account creation at access

spencer-tb:
### Description

Implements the EIP-8037 charge at access model for `CREATE`/`CREATE2` and creation transactions from ethereum/EIPs#11858.

Account creation state gas is now charged conditionally at the destination access, before the 63/64ths split, rather than charged unconditionally before the create frame and refilled. Failing a pre access check (balance, nonce, stack depth) reads no destination and charges nothing, an already alive target charges nothing, the charge is decided by existence alone, so a storage only collision is charged and refilled, and a failed create frame still refills.

Creation transactions follow the same rule: the create component of the intrinsic state gas is seeded into the reservoir rather than pre-consumed, charged at the deployment-address access only if the destination does not exist, and refilled on failure.

The storage only collision is untestable, the fixture loader rejects empty with storage pre state accounts (`StateWithEmptyAccount`), hence the codecov patch miss on the refill lines.

### Related Issues or PRs

ethereum/EIPs#11858

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/vm/opcodes.py
+++ b/packages/testing/src/execution_testing/vm/opcodes.py
@@ -5425,6 +5425,7 @@ class Opcodes(Opcode, Enum):
     - init_code_size: size of the initialization code in bytes (default: 0)
     - new_memory_size: memory size after expansion in bytes (default: 0)
     - old_memory_size: memory size before expansion in bytes (default: 0)
+    - account_new: whether creating a new account (default: True)
 
     Source: [evm.codes/#F0](https://www.evm.codes/#F0)
     """
@@ -5765,6 +5766,7 @@ class Opcodes(Opcode, Enum):
     - init_code_size: size of the initialization code in bytes (default: 0)
     - new_memory_size: memory size after expansion in bytes (default: 0)
     - old_memory_size: memory size before expansion in bytes (default: 0)
+    - account_new: whether creating a new account (default: True)
 
     Source: [evm.codes/#F5](https://www.evm.codes/#F5)
     """
--- a/packages/testing/src/execution_testing/vm/opcodes.py
+++ b/packages/testing/src/execution_testing/vm/opcodes.py
@@ -5381,6 +5381,7 @@ class Opcodes(Opcode, Enum):
             "init_code_size": 0,
             "new_memory_size": 0,
             "old_memory_size": 0,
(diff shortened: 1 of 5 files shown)


spencer-tb:
cc @chfast @misilva73

spencer-tb:
FYI the storage only collision is untestable, the fixture loader rejects empty with storage pre state accounts (StateWithEmptyAccount), hence the codecov patch miss on the refill lines. Please double check, these lines in the spec.

Opcode-level refill (`generic_create` collision branch):
https://github.com/ethereum/execution-specs/blob/b5fa770df96f7c069666697bb6509236fbbc9856/src/ethereum/forks/amsterdam/vm/instructions/system.py#L122-L123

Transaction-level refill (`process_message_call` collision branch):
https://github.com/ethereum/execution-specs/blob/b5fa770df96f7c069666697bb6509236fbbc9856/src/ethereum/forks/amsterdam/vm/interpreter.py#L146-L149

gurukamath (commented):
This largely looks good to me. But I have a comment on how we should merge this in.

A lot of the changes to `src/ethereum/forks/amsterdam/fork.py` seem to be transitional scaffolding to bridge over the fact that the latest version of EIP-2780 isn't already in `forks/amsterdam`. EIP-2780 gets rid of the intrinsic state gas (see [this](https://github.com/gurukamath/execution-specs/pull/17))
The real meat of this PR lies in the `system.py` changes and some `interpreter.py` changes.

So, I would say, we should first get #3126 merged in, then rebase this PR and try to get that in next. @spencer-tb your thoughts?

spencer-tb:
Sgtm!!

packages/testing/src/execution_testing/vm/opcodes.py
marioevz, line 5427:
```suggestion
    - old_memory_size: memory size before expansion in bytes (default: 0)
    - account_new: whether creating a new account (default: True)
```
I forgot to add this to the docstring 🙈

marioevz (approved):
LGTM on the tests side! I just left one comment, because I forgot to update `CREATE`/`CREATE2` docstrings in the PR I submitted to your branch (sorry!)

gurukamath (approved):
The specs side looks good to me.

PR #3120, opened 2026-07-06
fix(spec-specs, tests): anchor calldata floor on decomposed EIP-2780 base

spencer-tb:
### Description

Spec alignment PR, due to a spec/EIP mismatch. Anchor the EIP-7623 calldata floor on the decomposed EIP-2780 regular intrinsic base (`TX_BASE` + recipient access + value) rather than `TX_BASE` alone, matching the EIP.

Flagged in https://github.com/ethereum/go-ethereum/pull/35279#issuecomment-4873297759 by @rjl493456442.

Threads the tx recipient/value context through the framework floor calculator and re-baselines the affected fixtures.

### Related Issues or PRs

N/A

### Checklist

### Cute Animal Picture

Screenshot 2026-07-06 at 23 00 21


spencer-tb:
cc @danceratopz for review plz

src/ethereum/forks/amsterdam/transactions.py
```diff
@@ -725,15 +725,22 @@ def calculate_intrinsic_cost(
     # Total floor tokens.
     total_floor_tokens = floor_tokens_in_calldata + tokens_in_access_list
 
+    # Decomposed regular-gas intrinsic base (EIP-2780): sender, recipient
+    # access, and value primitives, excluding calldata, init code, access
+    # list, and authorizations.
+    base_regular_gas = GasCosts.TX_BASE + recipient_regular_gas
```
danceratopz, line 733: The `calculate_intrinsic_cost` docstring still describes the floor as based only on the calldata size. Worth a line there noting the floor is now anchored on this decomposed base. (While at it: the docstring says the function takes `gas_limit` as a parameter, but the signature is `(tx, sender)`; pre-existing.)

packages/testing/src/execution_testing/forks/forks/eips/amsterdam/eip_2780.py
```diff
@@ -41,6 +42,48 @@ def gas_costs(cls) -> GasCosts:
...
+        cls,
+    ) -> TransactionDataFloorCostCalculator:
+        """
+        Anchor the calldata floor on the decomposed regular-gas intrinsic
+        base (EIP-2780).
+
+        The inherited floor base is ``TX_BASE`` alone; add the recipient
+        access and value-transfer primitives so the floor never undercuts
+        the transaction's own intrinsic base. Calldata, init code, access
+        list, and authorizations are excluded.
```
danceratopz: "access list ... excluded" is slightly misleading: the access-list floor tokens (EIP-7981) are still included via `super_fn`; only the access list's regular intrinsic cost is excluded. Suggest rewording to make that distinction clear.

packages/testing/src/execution_testing/forks/base_fork.py
```diff
@@ -67,6 +67,9 @@ def __call__(
         *,
         data: BytesConvertible,
         access_list: List[AccessList] | None = None,
+        contract_creation: bool = False,
+        sends_value: bool = False,
+        recipient_type: RecipientType = RecipientType.CONTRACT,
```
danceratopz, line 72: The `CONTRACT` default silently adds `COLD_ACCOUNT_ACCESS` to the floor when a caller passes no context. Correct for the common case, but a short note in the Protocol docstring would help future call sites for create/self/value transactions remember to pass these.

danceratopz (approved):
Looks good. Some docstring only suggestions here:
- https://github.com/spencer-tb/execution-specs/pull/26

PR #3122, opened 2026-07-06
feat(test-fill): pack Engine X pre-alloc groups

danceratopz:
### Impact

This PR will help speed-up `consume enginex` by removing 2179 pre-alloc groups (5458 -> 3279). Besu's pure startup time is ~6.7 s, Nethermind's is 5 s (ignoring docker overhead), this would lead to the following approx savings per test run:

| Client     | Startup time per group | Churn saved per consume run (x 2,179) | With `--sim.parallelism=4` |
| ---------- | ---------------------- | ------------------------------------- | -------------------------- |
| Nethermind | ~5.0 s                 | ~3.0 h                                | ~45 min                    |
| Besu       | ~6.8 s                 | ~4.1 h                                | ~1.0 h                     |
| geth       | ~0.8 s                 | ~29 min                               | ~7 min  

### Background

`consume enginex` boots one execution client per pre-allocation group and amortizes that boot across the group's tests via repeated `engine_newPayload` calls, so the tail of a run is dominated by the fixed per-group container-startup cost and the main lever on runtime is the group count. Phase 1 keys a test's group on the exact content of any hard-coded accounts it sets, which is a conservative proxy for genuine address conflicts: it leaves 65% of Engine X groups holding a single test, most of them in `tests/ported_static`, where nearly every test pins accounts to canonical addresses.

### Approach in this PR

Please check this HTML doc that provides an overview:

[prealloc-packing.html](https://github.com/user-attachments/files/29977886/prealloc-packing.html)

This PR adds a `pack_pre_alloc_groups` pass that runs on the master process at the end of phase 1 and merges the fine-grained groups into fewer, larger ones without changing what any test executes:

- Groups are bucketed by everything a shared genesis requires (fork, chain id, and genesis environment) and by the explicit `pre_alloc_group` marker salt, so a test that demands its own genesis keeps it.
- Within a bucket, an address is *reserved* if it is a precompile (`< 0x100`, implicitly callable by any test) or is allocated by more than one group (a shared or canonical address rather than one private to a single test).
- Two groups merge only when their reserved accounts match exactly, so a funded precompile or scratch contract is never introduced into a test that assumes that address is empty. Test-private addresses merge freely, which is where most of the reduction comes from; within a bucket a non-reserved address has exactly one allocator and reserved accounts are byte-identical, so the merged pre-allocation cannot collide by construction.
- Packing is deterministic (a packed group's id derives from its sorted test ids) and writes a test id to group hash index file, which phase 2 uses to look up each test's group (the packed hash can no longer be recomputed per test).
- Any leak the reserved-address rule might miss fails the fill loudly: After phase 2, every Engine X fixture's payloads are compared against the test's `blockchain_test_engine` sibling, which is filled against the test's own pre-allocation with an independent `t8n` execution (Engine X fixtures never share the transition tool output cache). A difference in any payload field other than the state-root-derived ones (`stateRoot`, `blockHash`, `parentHash`) aborts the fill naming the offending tests, so a leaked account cannot change a test's behavior without failing the fill. Fills that generate no sibling fixtures (e.g. `-m blockchain_test_engine_x`) log a warning that the check was skipped; release fills generate all formats and always verify. The check adds well under a minute even to a full fill (measured: `0.8s` for 6126 fixtures).

### Results

Full fill across all forks:

```console
uv run fill --until=Amsterdam --output=fixtures-prealloc-all --generate-pre-alloc-groups -m blockchain_test_engine_x -n 6 --clean
```

| Metric | `forks/amsterdam` | This PR | Change |
| --- | ---: | ---: | ---: |
| Pre-alloc groups (one client boot each) | 5458 | 3279 | -40% |
| Single-test groups | 3564 (65.3%) | 1690 (51.5%) | -53% |
| Total genesis accounts | 137622 | 123908 | -10% |
| Tests | 74673 | 74673 | unchanged |

On the heavily fragmented forks the packing roughly doubles the boot amortization (average tests per group: Cancun 9.8 to 17.8, Prague 12.2 to 21.4, Osaka 12.7 to 21.1, Amsterdam 16.4 to 25.9).

The reduction comes almost entirely (98%) from `tests/ported_static`; the remainder comes for free from EIP test directories whose tests pin distinct fixed addresses and can now share a genesis (`paris/eip7610_create_collision`, `constantinople/eip1052_extcodehash`, and several `amsterdam` EIP tests):

| Groups by source directory | `forks/amsterdam` | This PR | Change |
| --- | ---: | ---: | ---: |
| `ported_static` | 4436 | 2309 | -2127 |
| `paris` | 154 | 132 | -22 |
| `amsterdam` | 94 | 76 | -18 |
| `constantinople` | 60 | 48 | -12 |
| `cancun` | 209 | 203 | -6 |
| Groups spanning multiple directories | 58 | 64 | +6 |
| All other directories | 447 | 447 | unchanged |

### Verification

All A/B comparisons below are against a base-branch fill at the same commit (`80cc337bf36`).

- Packing strictly merges whole phase-1 groups: Mapping all 74673 tests from their phase-1 group to their packed group shows that no phase-1 group is ever split or reassigned.
- Two-phase fills pass on both the ported and the non-ported merge paths: `tests/ported_static` at Osaka (`-m "blockchain_test_engine_x and (not slow)"`, 6126 passed, 0 failed) and `tests/paris/eip7610_create_collision` plus `tests/constantinople/eip1052_extcodehash` at Paris (125 passed, 0 failed).
- Execution is provably unchanged: Comparing every Engine X fixture between the two fills shows all payload fields equal except `stateRoot`, `blockHash`, and `parentHash` (which change with the genesis) for all 6126 Osaka tests (2282 of which received a different genesis from packing) and all 125 Paris tests.
- The `state_test`, `blockchain_test`, and `blockchain_test_engine` formats are untouched by construction (they are generated from a test's own pre-allocation, and packing runs entirely inside the Engine X pre-alloc pipeline): Re-filling `tests/ported_static` at Osaka on both branches yields identical fixture hashes for all 18378 fixtures of these formats, with byte differences only in `_info` metadata (the framework version string differs across branches).

### Reviewer notes

- The reserved-address rule is sound-leaning rather than proven (it cannot see an address a test reads but never allocates), but a leak cannot ship silently: The post-fill execution check above turns any behavioral change into a hard fill error, and an affected test can then be isolated with `@pytest.mark.pre_alloc_group` (whose salt the packing respects) and re-filled.
- The remaining single-test groups are the deliberate price of the exact-match rule, not unfinished work: A group whose reserved-address footprint matches no other group's stays alone, and a genuinely irreducible set remains (e.g. tests that overwrite a system contract at its canonical address, where no two variants can share a genesis). Packing on direct conflicts only, with an automatic split-and-retry for any test the execution check flags, is the natural follow-up to cut the count further.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

iA well-packed pre-allocation group sharing a single genesis at Salisbury Plain, South Georgia

### DIFF
--- a/packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
+++ b/packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
@@ -290,10 +290,11 @@ def packed_group_hash_for_test(
     return entry.group_hash
 
 
-# Addresses below this value are precompiles / the reserved range. A ported
-# state test can call them without ever declaring them in its pre, so an
-# account introduced there by another test in the group silently changes its
-# execution.
+# Blanket-reserved low address range. A ported state test can blindly call
+# a low address without ever declaring it in its pre, so an account
+# introduced there by another test in the group silently changes its
+# execution. Precompiles at or above this range (EIP-7951 puts P256VERIFY
+# at 0x100) are reserved via the bucket fork's precompile list instead.
 _RESERVED_ADDRESS_CEILING = 0x100
 
 
@@ -305,15 +306,22 @@ def _reserved_addresses(builders: List["PreAllocGroupBuilder"]) -> Set[str]:
     A ported state test only declares the accounts it sets and assumes all
     other addresses are empty, so introducing an account at an address it
     quietly depends on (a precompile, a canonical scratch contract, ...)
-    changes its result. Two kinds of address are therefore reserved: the low
-    precompile range, and any address more than one group allocates (i.e. a
-    shared/canonical address rather than one private to a single test).
+    changes its result. Three kinds of address are therefore reserved: the
+    blanket low range, the fork's precompile addresses (which extend beyond
+    that range from EIP-7951's P256VERIFY at ``0x100`` on), and any address
+    more than one group allocates (i.e. a shared/canonical address rather
+    than one private to a single test).
     """
+    # Packing buckets by fork, so every builder shares this one; a
+    # transition fork reserves the post-transition precompiles, matching
+    # the genesis built by `PreAllocGroupBuilders.add_test_pre`.
+    fork = builders[0].fork.transitions_to()
+    reserved = {str(address) for address in fork.precompiles()}
     frequency: Dict[str, int] = defaultdict(int)
     for builder in builders:
         for address in builder.pre.root:
             frequency[str(address)] += 1
-    return {
+    return reserved | {
         address
         for address, count in frequency.items()
         if count > 1 or int(address, 16) < _RESERVED_ADDRESS_CEILING
--- a/packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
+++ b/packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
@@ -43,6 +43,13 @@ class PreAllocGroupBuilder(CamelModel):
     )
     fork: Fork | TransitionFork = Field(..., alias="network")
     chain_id: int = DEFAULT_CHAIN_ID
+    group_salt: str | None = Field(
+        None,
+        description=(
+            "Explicit isolation salt from the `pre_alloc_group` marker; "
+            "groups only pack with groups carrying the same salt."
+        ),
+    )
     pre: Alloc
 
     def get_pre_account_count(self) -> int:
@@ -77,6 +84,7 @@ class PreAllocGroupBuilder(CamelModel):
             environment=self.environment,
             fork=self.fork,
             chain_id=self.chain_id,
+            group_salt=self.group_salt,
             pre=self.pre.model_dump(),
             pre_account_count=self.get_pre_account_count(),
             test_count=self.get_test_count(),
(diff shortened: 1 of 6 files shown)


spencer-tb (approved):
Nothing looks out of the ordinary to me, nice that we are verifying this at fill time.

I added a post state check here: https://github.com/danceratopz/execution-specs/pull/60/

Feel free to use it! Approving :)

danceratopz:
> Nothing looks out of the ordinary to me, nice that we are verifying this at fill time.
> 
> I added a post state check here: [danceratopz#60](https://github.com/danceratopz/execution-specs/pull/60)
> 
> Feel free to use it! Approving :)

Thanks for the suggestion :pray: I just pushed some other fixes. I think your PR needs a bit more work to be useful, but let me take another look before discounting it.

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
```diff
@@ -180,6 +192,264 @@ def merge_partial_group_files(folder: Path) -> None:
...
+    shared/canonical address rather than one private to a single test).
+    """
+    frequency: Dict[str, int] = defaultdict(int)
+    for builder in builders:
+        for address in builder.pre.root:
+            frequency[str(address)] += 1
+    return {
+        address
+        for address, count in frequency.items()
+        if count > 1 or int(address, 16) < _RESERVED_ADDRESS_CEILING
```
marioevz, line 327: This currently excludes the EIP-7951 precompile.

danceratopz:
> > Nothing looks out of the ordinary to me, nice that we are verifying this at fill time.
> > I added a post state check here: [danceratopz#60](https://github.com/danceratopz/execution-specs/pull/60)
> > Feel free to use it! Approving :)
> 
> Thanks for the suggestion 🙏 I just pushed some other fixes. I think your PR needs a bit more work to be useful, but let me take another look before discounting it.

Hey @spencer-tb, I took another look at your PR. TLDR: I think it's useful; needs a bugfix to work; let's do those as two separate PRs. More details in the review https://github.com/danceratopz/execution-specs/pull/60

marioevz (approved):
Just one comment!

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
danceratopz: Thanks, nice catch! We now explicitly add the fork's dedicated list of precompiles to the set of reserved addresses via:
```python 
reserved = {str(address) for address in fork.precompiles()}
```
The check above was kept as a blanket check for low addresses, as some tests probe these low addresses without declaring them in their pre (to check for precompile absence, for example).

danceratopz:
Thanks for the reviews @spencer-tb and @marioevz!

@spencer-tb, I added follow-up issues to fix the `postStateDiff` field in `blockchain_test_engine_x` and for your additional proposed check you PR'd in https://github.com/danceratopz/execution-specs/pull/60. It's a good idea, thanks!

- https://github.com/ethereum/execution-specs/issues/3160
- https://github.com/ethereum/execution-specs/issues/3161

PR #3124, opened 2026-07-07
feat(test-cli): add `--extract-opcode-count` opcode tracing

skylenet:
## 🗒️ Description

Adds an opt-in `--extract-opcode-count` flag to `fill-stateful`.

When set, each **execution-phase** block built against the live client is traced via `debug_traceBlockByHash` using a small JS opcode-counting tracer, and the per-opcode execution counts are recorded in the fixture's `_info.metadata.opcode_counts` — an array with one entry per `engineNewPayloads` block (`opcode_counts[i]` is the count for `engineNewPayloads[i]`, or `null` if its trace was unavailable), so multi-block benchmarks keep per-payload granularity.

Details:

- `DebugRPC.trace_block_by_hash` wraps `debug_traceBlockByHash`.
- `ClientBackend` gains optional `debug_rpc` + `extract_opcode_count` ctor params, a JS tracer constant, and `extract_block_opcode_count()` (aggregates per-tx opcode counts across a block).
- `make_stateful_fixture` traces **execution-phase blocks only** (setup blocks are skipped — no wasted trace) and records one opcode count per execution payload into `_info.metadata.opcode_counts` via `FillResult.metadata` — no filler/fixture-model changes needed. `reset_opcode_count()` is a no-op: the filler's aggregate `opcode_count` is unused, so that singular key is not emitted.
- `fill_stateful` plugin: the `--extract-opcode-count` flag and `debug_rpc` wired into the backend.
- Docs updated in `docs/filling_tests/fill_stateful.md`.

The flag is fully opt-in: with it off, no trace RPCs are issued and the `opcode_counts` key is omitted. It requires the client to expose the `debug` namespace with JS tracer support; the tracer re-executes each block, so it is slow and intended for analysis runs.

Opcode-count verification: when a benchmark test declares a target opcode count (`fixed_opcode_count`/`expected_opcode_count`) and `--extract-opcode-count` is on, the live-client trace is verified against that target and the fill fails on >5% divergence — surfacing benchmarks that don't execute what the author expected. With the flag off, `benchmark_opcode_count` stays `None` and verification is skipped, exactly as before.

Robustness:

- Non-fatal extraction: a `debug_traceBlockByHash` RPC error is logged and skipped rather than aborting the fill.
- Opcode-name normalization: geth emits `opcode 0xNN not defined` for undefined opcodes, which `OpcodeCount` could not parse; such names are reduced to the bare `0xNN` (`UndefinedOpcode`), and anything still unrecognized is dropped with a warning.

`just static` (mypy, ruff, vulture, ethereum-spec-lint, actionlint, codespell) passes.

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

cute cat


jochem-brouwer:
EDIT/NOTE: written with Claude, reflects my thoughts.

Preliminary comment — this feature is definitely something we want.

One concern: debug_traceBlockByHash is [in](https://github.com/ethereum/execution-apis/blob/a6dea42bc624084755f4af989c92c9cdef68773a/src/debug/trace.yaml#L180) execution-apis, but custom JS tracers are not — they're a client-specific extension with unspecified output. That's fine today since the trace only runs at fill time against geth, but it locks opcode extraction to JS-tracer-capable clients, and _normalize_opcode_name() already shows the next-fork fragility (geth emitting "opcode 0xNN not defined" for unknown opcodes).

Longer term: since we already have testing_buildBlock / testing_commitBlock, a testing_ method returning opcode counts directly would remove the JS-tracer dependency entirely — and it's much cheaper than standardizing full trace output.

Question: is the motivation parity with the standard fill benchmark path (diff_opcode_counts.py), or also feeding Nethermind's gas-benchmarks flow?

Not blocking — for Glamsterdam we can't wait for a standardized endpoint, so this is the right pragmatic step. Will re-visit this week 👍 😄

skylenet:
@jochem-brouwer the motivation from my side is mainly getting the opcode counts to be able to show them on benchmarkoor. But the good side effect is that we can also detect opcode divergence early. 

The feature is behind a flag and not enabled by default so it shouldn't interfere with existing tooling. 

I agree that maybe some custom testing API endpoint would be the best/fastest, but for now I think it's fine if we do the following:
 
1 ) **Extract via JS tracer.** Small response body. Works with nethermind, geth. 
 
2 ) **Fallback to a counting them from the structLogs**.  Big response body, but should work with other clients that don't support JS tracers, e.g. Besu. 
 
I'm working now on testing the fallback approach.

skylenet:
So I've got this working with geth, nethermind (using JS tracer) and besu (using the fallback)

jochem-brouwer:
(have rebased)

packages/testing/src/execution_testing/client_clis/client_backend.py
```diff
@@ -209,6 +253,155 @@ def evaluate(
...
+
+        Failures are non-fatal: a trace RPC error or an unrecognized
+        opcode name is logged and skipped rather than aborting the fill,
+        since this is an opt-in analysis feature.
+        """
+        if not self.extract_opcode_count or self.debug_rpc is None:
+            return None
+
+        if not self._js_tracer_unsupported:
+            counts = self._count_via_js_tracer(block_hash)
```
jochem-brouwer: If `_count_via_js_tracer` returns an empty object (happens on blocks with 0 txs, assuming that system calls are not logged in the trace) then the tracer will permanently flip to tracer unsupported. An empty object should be fine and not treat the client as having no js trace support.
Also, if the RPC throws/timeouts, it will likely also permanently mark it as unsupported js trace 🤔

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1485,21 +1485,37 @@ def make_stateful_fixture(
...
-            # what the client actually built.
-            client_hash = Hash(
-                built_block.engine_payload.payload_response.execution_payload.block_hash
-            )
+                    # Feed the traced count to benchmark opcode
+                    # verification: with --extract-opcode-count on, a
+                    # live-client opcode count that diverges from the
+                    # test's declared target fails the fill. Stays None
+                    # when the flag is off, which skips verification.
+                    benchmark_opcode_count = block_opcode_count
```
jochem-brouwer, line 1510: This writes to this variable each loop, so the final execution block "wins". At some point we want to support multi-block benchmarks, so we should likely "merge" the opcode counts together if the opcode count is already set.

jochem-brouwer (changes requested):
Two comments! 😄 👍

packages/testing/src/execution_testing/specs/blockchain.py
skylenet: Good point. I think we could use `_info.metadata.opcode_counts: []`  (note the plural `counts`)   instead of a single `_info.metadata.opcode_count: {}` .  

So every element in `_info.metadata.opcode_counts: []` would match the test/benchmark `engineNewPayloads:[]`

packages/testing/src/execution_testing/client_clis/client_backend.py
LouisTsai-Csie: I've added some error handling for this, it should separate the RPC timeout issue from non-existing JS tracer.

LouisTsai-Csie (approved):
Good from my side - i push some commits, waiting for Jochem for the final review

jochem-brouwer (approved):
LGTM!

I keep forgetting that "request changes" is blocking until I dismiss or approve 😅

PR #3125, opened 2026-07-07
fix(test-fixtures): emit chainId in state test fixture transaction

spencer-tb:
### Description

The state test fixture `transaction` object carries no chain id, so the wrong-chain-id fixtures shipped in `tests-glamsterdam-devnet@v6.1.1` (#3000) can only be detected by decoding `txbytes`.

Emit `chainId` in the state fixture transaction so JSON field consumers can see it, omitting it for unprotected type 0 transactions which encode no chain id.

### Related Issues or PRs

Follow-up to #3000.

### Checklist

### Cute Animal Picture

### DIFF
--- a/docs/running_tests/test_formats/state_test.md
+++ b/docs/running_tests/test_formats/state_test.md
@@ -108,6 +108,10 @@ Excess blob gas of the block where the transaction is executed.
 
 ### `FixtureTransaction`
 
+#### - `chainId`: [`ZeroPaddedHexNumber`](./common_types.md#zeropaddedhexnumber)
+
+Chain id of the transaction; omitted for unprotected (non-EIP-155) type-0 transactions
+
 #### - `nonce`: [`ZeroPaddedHexNumber`](./common_types.md#zeropaddedhexnumber)
 
 Nonce of the account that sends the transaction
--- a/packages/testing/src/execution_testing/fixtures/state.py
+++ b/packages/testing/src/execution_testing/fixtures/state.py
@@ -46,6 +46,7 @@ class FixtureTransaction(TransactionFixtureConverter):
     # via model_dump(), which includes many fields not in this model.
     model_config = CamelModel.model_config | {"extra": "ignore"}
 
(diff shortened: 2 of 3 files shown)


danceratopz (approved):
Nice work following up on #3000!

I was worried that the extra field might break consumers, but neither `evmone` nor `geth` have any trouble parsing the JSON (I didn't test others). Geth parses `txbytes` from `post` and `evmone` does parse the `chainid`. Although, looks like evmone would legitimately fail, once this is merged here's what Claude said about that, cc @chfast:
> To be precise about evmone: evmone-statetest (0.22.0-14) fails 27 of the 29 test_invalid_chain_id state fixtures with "unexpected valid transaction" (the two type-4 cases pass only coincidentally, rejected for their empty authorizationList rather than the wrong chain id). Results are identical with and without the new chainId field, so this is a pre-existing gap rather than a regression: evmone already parses chainId from the transaction JSON (defaulting to 1 when absent) but never validates it. These aren't xfailed in evmone's CI: its execution-spec-tests job is still pinned to tests-glamsterdam-devnet@v6.1.0, which predates the #3000 fixtures, so the failures will surface on their next release bump. Notably evmone-blockchaintest passes all 29 blockchain-test variants, since there the wrong chain id is carried by the block-embedded transaction signature: the JSON state-test path is the only gap, and this PR provides exactly the field needed to close it. 

Interestingly,  `test_invalid_chain_id` isn't marked `eels_base_coverage`, so the `json-loader` job never runs these fixtures, but this does now fix EELS consumption with fixtures filled post #3000 as it brings the expectation inline that PR. 

This will change all the `state_test` hashes, but nothing we can do about that!

chfast:
I think in general this is good change. The chainid is grey area in testing because many parts implicitly expect chain id of 1. To the point evmone was failing execution on testnets where chainid was not 1 any more. I understand, this doesn't fix the problem completely but still nice addition.

In evmone there are likely some shortcuts, but I can easily adapt when consuming next release of EEST.

spencer-tb:
Great! Will merge :)

Issue #3034, opened 2026-06-22
Test Release Tracker: `glamsterdam-devnet@v7.0.0`


marioevz:
### Overview

This tracker is for the first release for Glamsterdam Devnet 7, for which discussions are still ongoing regarding EIP changes to be included, and will be based on the latest `glamsterdam-devnet-6` spec.

- **Target release tag**: `tests-glamsterdam-devnet@v7.0.0`
- **EELS branch**: [`devnets/glamsterdam/7`](https://github.com/ethereum/execution-specs/tree/devnets/glamsterdam/7)
- **Test release date**: ~~(ASAP after ACDT 2026-07-06)~~ 2026-07-08
- **Devnet launch date**: ~2026-07-09
- **Related Trackers**
  - #3049
  - #2915

### Scope

Intrinsic gas rework for EIP-2780/EIP-8037, a pre-access check for EIP-8038/EIP-7928, and final EIP-8282 contract constants & disable switch.

Repricing numbers (EIP-8038, EIP-2780) are treated as stable for devnet-7 but will likely be changed still.

New mined EIP-8282 addresses:
- Builder Deposit Contract: `0x0000BFF46984E3725691FA540A8C7589300D8282`
- Builder Exit Contract: `0x000064D678505AD48F8CCB093BC65613800E8282`

#### EIP Updates

Some PRs cherry-sqaushed to `glamsterdam/devnets/7`.

- [x] **[EIP-2780](https://eips.ethereum.org/EIPS/eip-2780) & [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037)**: @spencer-tb / @gurukamath
  - https://github.com/ethereum/EIPs/pull/11844 & https://github.com/ethereum/EIPs/pull/11891 @gurukamath
    - [x] Roll back all authorizations, stamp of approval from lightclient: https://discord.com/channels/595666850260713488/745077610685661265/1523709794501005402
    - [x] EELS PR: #3126 (cherry squashed to `devnets/glamsterdam/7`)
  - https://github.com/ethereum/EIPs/pull/11858 @spencer-tb
    - [x] EELS PR: https://github.com/ethereum/execution-specs/pull/3116 (cherry squashed to `devnets/glamsterdam/7`)
[...]

PR #3126, opened 2026-07-07
feat(spec-specs,test-tests): update the EIP-2780 implementation

gurukamath:
## 🗒️ Description
Implements the EIP-2780 rework from [ethereum/EIPs#11844](https://github.com/ethereum/EIPs/pull/11844): state dependent transaction costs move out of the intrinsic gas — which over-charged up front and refunded after — into runtime charges at the top frame, keyed on the transaction's pre-state.

- **Intrinsic** is now state-independent only (and remains the sole validity input): no `NEW_ACCOUNT` for creates, just `REGULAR_PER_AUTH_BASE_COST` per authorization.
- **Authorizations** (`set_delegation`): `NEW_ACCOUNT` if the authority's leaf doesn't exist; `ACCOUNT_WRITE` on the transaction's first write to the authority (sender exempt, repeats pay once); `AUTH_BASE` once per authority for a net-new delegation, never credited back.
- **Dispatch prep** (`prepare_dispatch`): recipient/creation-target `NEW_ACCOUNT` keyed on pre-tx aliveness; delegation resolution charged warm/cold.
- **Failure semantics**: prep out-of-gas rolls back all applied delegations and halts without dispatching. After prep, the authorization state gas is folded into the frame baseline, so a dispatch revert/halt can't refill it — delegations persist per EIP-7702, and their state gas stays consumed. Recipient/creation-target charges stay refillable since their state rolls back with the frame.
- **Removed**: the intrinsic `NEW_ACCOUNT` refund paths and `MessageCallOutput.state_refund` / `created_target_alive` plumbing.

## 🔗 Related Issues or PRs
#3034

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

Cute Animals - 5 of 7


tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7702.py
```diff
@@ -485,6 +486,67 @@ def test_bal_7702_delegated_storage_access(
...
+        recipient_type=RecipientType.DELEGATION_7702,
+    )
+
+    if outcome == "oog":
+        gas_limit = intrinsic_gas + top_frame_gas - 1
+        # The delegation charge fails, so the target is never accessed.
+        delegated_to_expectation = None
+    else:
+        gas_limit = intrinsic_gas + top_frame_gas + 1000
+        delegated_to_expectation = BalAccountExpectation.empty()
```
marioevz, line 530: Minor nit:
```suggestion
    gas_limit = intrinsic_gas + top_frame_gas
    if outcome == "oog":
        gas_limit -= 1
        # The delegation charge fails, so the target is never accessed.
        delegated_to_expectation = None
    else:
        delegated_to_expectation = BalAccountExpectation.empty()
```

src/ethereum/forks/amsterdam/transactions.py
marioevz, line 720: I believe the concept of intrinsic state gas has been completely removed from the EIP with Dragan's PR. We should simplify this function and `IntrinsicGasCost` by removing all references to state intrinsic gas. It's not present in any prior forks, so should be easy to remove.

gurukamath:
I have rebased this branch to fix the conflicts emerging from #3120 merge

src/ethereum/forks/amsterdam/vm/eoa_delegation.py
```diff
@@ -191,92 +192,83 @@ def validate_authorization(
...
-        refund = StateGas(Uint(0))
+        if not account_exists(tx_state, authority):
+            charge_state_gas(evm, StateGasCosts.NEW_ACCOUNT)
 
-        if account_exists(tx_state, authority):
-            refund += StateGasCosts.NEW_ACCOUNT
-            # The new-account ACCOUNT_WRITE charged at intrinsic time is
-            # not needed: refund it to the regular refund counter.
-            regular_refund += GasCosts.ACCOUNT_WRITE
+        if authority not in written_accounts:
```
jochem-brouwer, line 253: **The `ACCOUNT_WRITE` rule diverges from (part of) the EIP — which reading is intended?**

The implementation charges `ACCOUNT_WRITE` on the transaction's first write to *any* authority — including authorities whose leaf already exists — with the sender always exempt and `tx.to` exempt only when the transaction is value-bearing. The EIP text is internally inconsistent on this rule:

- *Specification → Runtime gas costs*: "if this is the first write to the authority within the transaction (i.e. the authority differs from `tx.to`), charge `ACCOUNT_WRITE`" — supports first-write tracking, but the parenthetical exempts `tx.to` **unconditionally** (this impl charges a zero-value `tx.to` authority) and never mentions the sender exemption (which this impl applies).
- *Interactions → EIP-7702*: "…plus `ACCOUNT_WRITE` **for a non-existent authority**…" — ties `ACCOUNT_WRITE` to the new-leaf case only.
- *Test Cases #8*: "a non-existent authority additionally pays `STATE_BYTES_PER_NEW_ACCOUNT × CPSB` state and `ACCOUNT_WRITE` regular, **an existing authority pays neither**" — under this reading an existing authority pays no runtime `ACCOUNT_WRITE` at all. That matches the old intrinsic-charge-plus-refund model this PR removes, and disagrees with this implementation by +8,000 gas for every existing authority.

`test_account_write_authority_is_recipient`'s docstring says it "resolves the EIP text's parenthetical", i.e. the tests encode an interpretation rather than the letter of the EIP. First-write tracking is arguably the more defensible reading under EIP-8038 first-write semantics (the per-auth base only prices *warm* writes), but the spec repo shouldn't settle this unilaterally — can we confirm the intended rule with the EIP authors and land a follow-up EIPs PR making Specification / Interactions / Test Cases consistent (whichever way it goes)?

Note: we should not attempt to resolve this here (IMO) but if this is not clear in the EIP text/spec then this should be fixed in the EIP text, and we should raise this as then it is not possible to write a correct implementation based on the EIP spec (then one would run our tests, figure out what the correct implementation is (according to tests, not the EIP text), and then edit the implementation to that spec)

src/ethereum/forks/amsterdam/vm/eoa_delegation.py
```diff
@@ -191,92 +192,83 @@ def validate_authorization(
...
+    message = evm.message
     tx_state = message.tx_env.state
-    state_refund = Uint(0)
-    regular_refund = Uint(0)
+    # Accounts this transaction has already written: the sender's leaf
+    # was written at inclusion (nonce bump and fee deduction). The
+    # recipient is written when value is transferred.
+    written_accounts: Set[Address] = {message.tx_env.origin}
+    if evm.message.tx_env.value > U256(0):
+        written_accounts.add(evm.message.current_target)
```
jochem-brouwer, line 240: Two follow-ups on the value-bearing recipient exemption:

1. At authorization-processing time the value transfer hasn't happened yet — `move_ether` runs when the frame is entered, after prep. So the exemption is really "the write is already *priced* by the transaction" (`TX_VALUE_COST`), not "the leaf was already written", and it holds even if the dispatch later reverts (intrinsic charges aren't refunded). The comment's "The recipient is written when value is transferred" reads as if the write precedes the auths; worth stating the prepaid-not-yet-performed nature explicitly.
2. Edge case to sanity-check: authority == `tx.to`, `value > 0`, and the authority's leaf does **not** exist pre-tx. The authorization pays `NEW_ACCOUNT` state gas but is exempt from `ACCOUNT_WRITE`, so this brand-new leaf's first write is priced at `TX_VALUE_COST` (4,244) instead of `ACCOUNT_WRITE` (8,000) — cheaper than the identical new-leaf authority at any other address. Intended, or should the exemption only apply to pre-existing recipients?

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -267,6 +243,72 @@ def process_create_message(message: Message) -> Evm:
...
+    back the whole preparation -- including the applied authorizations
+    -- and halts the frame without dispatching.
+    """
+    message = evm.message
+    tx_state = message.tx_env.state
+
+    if message.target == Bytes0(b""):
+        if (
+            get_pre_state_account(tx_state, message.current_target)
+            == EMPTY_ACCOUNT
```
jochem-brouwer, line 281: The create-target check uses the transaction *pre-state* while the recipient check below uses live (post-authorization) state — both are correct: `process_create_message` has already bumped the target's nonce by the time this runs, so a live existence check would always see the account, and type-4 transactions can't create, so pre-state ≡ post-auth state here. But the asymmetry is easy to trip over: the EIP phrases the create-target rule as a runtime load ("if it does not already exist"), and the PR description says both charges are "keyed on pre-tx aliveness" when only the create target is (the recipient is deliberately post-auth, so an authority-materialized `tx.to` isn't double-charged). Suggest a sentence in the docstring explaining why the create branch must use `get_pre_state_account`, and fixing the PR description.

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -267,6 +243,72 @@ def process_create_message(message: Message) -> Evm:
...
+        ):
+            charge_state_gas(evm, StateGasCosts.NEW_ACCOUNT)
+    else:
+        recipient = message.current_target
+        if message.value > U256(0) and not is_account_alive(
```
jochem-brouwer, line 286: Note: `prepare_dispatch` has no self-transfer exemption, so a delegated self-recipient (sender == `tx.to`, sender delegated) resolves the delegation and pays the warm/cold target access. That matches the EIP's reference-case row "ETH transfer to self, sender 7702 delegated → runtime `COLD_ACCOUNT_ACCESS` + execution" (new in 11844), but contradicts the EIP's prose, which scopes the recipient runtime charges with "unless the transaction is a self-transfer or a contract creation". I believe the table — and this implementation — are what's intended (`NEW_ACCOUNT` can't fire on a self-transfer anyway, since the sender is necessarily alive), but the prose should be fixed in the same follow-up EIPs PR as the `ACCOUNT_WRITE` wording.

jochem-brouwer (commented):
Hi @gurukamath, this is the first time I have instructed Claude to create a draft review for me, which I have then checked before publishing it. I normally completely review PRs myself, but this is something new I wanted to try.

If the quality of this review is not of expected quality, or if there is general feedback on it (Claude generates a lot of text) then please say so, I will correct this in future reviews or drop reviews-by-AI completely. 

I have checked the text and edited some parts before sending, so this is not a "blind review" where I just instruct AI to create a review and then blindly send it.

A general note: there seems to be inconsistencies in the EIP, or some behavior is not (well) defined. I believe this is not our task to fix, this is the domain where the EIP should govern the spec, and if something is missing or inconsistent then this should be fixed ASAP on the EIP side. There is a comment below on `tx.to` and EIP-7702 delegations which I think is currently indeed not well specified.

----

Reviewed against [ethereum/EIPs#11844](https://github.com/ethereum/EIPs/pull/11844) (now merged into EIP-2780), re-checked after the rebase onto the #3120 merge and the two follow-up commits (36805cc / 5d6ca1f). The implementation incorporates the rework faithfully on almost every axis I checked:

**Verified against the EIP**

- Intrinsic gas is state-independent only: creations no longer pre-pay `NEW_ACCOUNT` state gas, authorizations pay only `REGULAR_PER_AUTH_BASE_COST`, and validity is decided solely by the intrinsic check. The "Remove intrinsic state gas concept" commit takes this to its logical end — `IntrinsicGasCost.state` and the `tx_env` plumbing are gone entirely, so there's no dead always-zero path left. Nice cleanup.
- EIP-7623 floor: `data_floor_gas_cost` is now anchored on `base_regular_gas` (`TX_BASE` plus the recipient/value primitives — `COLD_ACCOUNT_ACCESS + TX_VALUE_COST + TRANSFER_LOG_COST` for value transfers, `CREATE_ACCESS` for creations), with initcode, auth, and access-list costs correctly excluded from the base. This matches the EIP's 7623 interaction, including its validity check.
- Runtime authorization charges (`set_delegation`): `NEW_ACCOUNT` when the authority leaf doesn't exist (at most once — later auths see the created leaf), `AUTH_BASE` only for a net-new delegation indicator (once per authority, never credited back; pre-tx-delegated re-point/clear-then-set and same-tx set–clear–set all correctly pay zero/one `AUTH_BASE`).
- Dispatch prep: recipient `NEW_ACCOUNT` keyed on post-authorization aliveness (so an authority-materialized `tx.to` doesn't double-pay), delegation-target access charged warm/cold per the 2929 model (the `WARM_ACCESS` case is new in 11844 and correctly honored), create-to-pre-existing-target pays no `NEW_ACCOUNT` (new reference case in 11844).
- Failure semantics: pre-execution OOG → included, all gas consumed, everything (including applied delegations) rolled back, frame never dispatched. For creates I confirmed `process_create_message` restores its own pre-nonce-bump snapshot on `evm.error`, so the target's nonce bump and destroyed storage roll back too. Post-prep, the auth state gas is folded into the frame baseline so a dispatch revert/halt can't refill it (delegations persist per 7702), while recipient/create-target `NEW_ACCOUNT` stays refillable — LIFO via `state_gas_spilled`, matching the refill language 11844 added.
- Authorities are added to `accessed_addresses` right after signature recovery (`validate_authorization`), per the EIP.
- Testing-package calculators and the migrated 8037/8038/7778/7928 suites mirror the new model; the OOG rollback matrix in `test_authorization_oog.py` is thorough, and the delegation-pointer test fix in 36805cc (inert-authorization flags) closes a real phantom-`AUTH_BASE` hole.

**Main open question** (inline at `set_delegation`): the runtime `ACCOUNT_WRITE` rule. The implementation charges it on the transaction's first *priced* write to the authority (sender exempt, value-bearing `tx.to` exempt, existing authorities **do** pay), but the EIP's Specification bullet, its EIP-7702 interactions paragraph, and its Test Case 8 give three mutually inconsistent versions of this rule — one of which ("an existing authority pays neither") disagrees with this implementation by 8,000 gas per existing authority. The tests openly "resolve" the EIP's parenthetical rather than follow it. I'd like us to pin down the intended behavior and fix the EIP (or the impl) before this merges — details inline.

gurukamath:
> **Main open question** (inline at `set_delegation`): the runtime `ACCOUNT_WRITE` rule. The implementation charges it on the transaction's first _priced_ write to the authority (sender exempt, value-bearing `tx.to` exempt, existing authorities **do** pay), but the EIP's Specification bullet, its EIP-7702 interactions paragraph, and its Test Case 8 give three mutually inconsistent versions of this rule

This has been discussed [here](https://github.com/ethereum/EIPs/pull/11891/changes)

src/ethereum/forks/amsterdam/vm/eoa_delegation.py
gurukamath: I think this has been resolved after a discussion and will be clarified in a subsequent EIP update.
See [this](https://github.com/ethereum/EIPs/pull/11891/changes) and [this](https://discord.com/channels/595666850260713488/1504670887851327648/1524582370114535535)

src/ethereum/forks/amsterdam/vm/eoa_delegation.py
gurukamath: Agree with point 1.
Point 2 is intended. More context [here](https://discord.com/channels/595666850260713488/1504670887851327648/1524384947551338607)

src/ethereum/forks/amsterdam/vm/interpreter.py
gurukamath: Yes. The EIP text reads very awkwardly here. Have already raised this in a [comment](https://github.com/ethereum/EIPs/pull/11891#discussion_r3543743546) upstream

gurukamath:
Thanks @jochem-brouwer for the review. I will make a couple of docstring updates to make the intent clearer. And I agree. Some of this needs upstream clarification on the EIP text and we have raised the comments on the relevant PR

src/ethereum/forks/amsterdam/vm/interpreter.py
```diff
@@ -191,9 +169,7 @@ def process_message_call(message: Message) -> MessageCallOutput:
         return_data=evm.output,
         state_gas_left=evm.state_gas_left,
         regular_gas_used=evm.regular_gas_used,
-        state_gas_used=frame_state_gas_used(evm),
-        state_refund=state_refund,
-        created_target_alive=target_alive,
+        state_gas_used=frame_state_gas_used(evm) + evm.auth_state_gas_used,
```
marioevz: This addition seemed a bit out of place to me so I asked Claude for their opinion and it came up with a good solution (IMO):

## Suggestion: fold the auth state-gas split into a named `commit_frame_state_gas` helper

The current top-frame handling snapshots the auth state gas, moves the refill
baseline down, and zeroes the spill counter inline, then has to re-add the
snapshot to the total outside `frame_state_gas_used`:

```python
evm.auth_state_gas_used = frame_state_gas_used(evm)
message.state_gas_reservoir = evm.state_gas_left
evm.state_gas_spilled = Uint(0)
...
state_gas_used=frame_state_gas_used(evm) + evm.auth_state_gas_used,
```

Splitting the total into two halves and manually re-joining them outside the
function is hard to follow. Naming the concept — state gas that's been
**committed** (locked in, non-refillable because the state it paid for outlives
a later dispatch failure) — lets `frame_state_gas_used` stay self-contained and
turns the inline dance into a single call.

```python
def frame_state_gas_used(evm: Evm) -> int:
    """
    Return the net state gas consumed by a finished frame, including
    any state gas committed as non-refillable earlier in the frame.
    """
    return (
        int(evm.message.state_gas_reservoir)
        - int(evm.state_gas_left)
        + int(evm.state_gas_spilled)
        + evm.committed_state_gas
    )


def commit_frame_state_gas(evm: Evm) -> None:
    """
    Mark the state gas consumed so far as non-refillable and reset the
    refill baseline so a later ``refill_frame_state_gas`` restores only
    the consumption that follows this commit.
    """
    evm.committed_state_gas = frame_state_gas_used(evm)
    evm.message.state_gas_reservoir = evm.state_gas_left
    evm.state_gas_spilled = Uint(0)
```

With this, the top-frame call site becomes `commit_frame_state_gas(evm)` after
`set_delegation`, and the total is just `state_gas_used=frame_state_gas_used(evm)`
— no external addition. (The `auth_state_gas_used` field is renamed to
`committed_state_gas`.)

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/helpers.py
```diff
@@ -15,6 +28,186 @@
...
+                address=target,
+                nonce=0,
+                signer=authority,
+                creates_account=True,
+            )
+            applied = Account(nonce=1, balance=0, code=designation(target))
+            original = None
+        case AuthorizationAction.SETS_NEW_DELEGATION:
+            target = pre.deploy_contract(code=Op.STOP)
+            authority = pre.fund_eoa(amount=balance)
```
marioevz, line 99:
```suggestion
            assert balance > 0, (
                "An existing account must have non-zero balance"
            )
            authority = pre.fund_eoa(amount=balance)
```
Just to make sure that there's no new account created.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_charges.py
```diff
@@ -0,0 +1,567 @@
...
+    - ``CLEARS_DELEGATION``: an already-delegated EOA is cleared (the
+      authorization target is the null address); no ``AUTH_BASE``
+      accrues and the delegation code is removed.
+    - ``INVALID``: the authorization nonce does not match the
+      authority's account nonce, so ``validate_authorization`` skips it.
+      The intrinsic base cost is still paid; no top-frame charge (not
+      even ``ACCOUNT_WRITE``) accrues and the authority is untouched.
+    """
+    sender_initial_balance = 10**18
+    sender = pre.fund_eoa(sender_initial_balance)
```
marioevz:
```suggestion
    sender = pre.fund_eoa()
```
Can be applied if the next comment is applied (and can be applied to most of the tests in this file.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_charges.py
```diff
@@ -0,0 +1,567 @@
...
+
+    tx = Transaction(
+        sender=sender,
+        to=recipient,
+        value=0,
+        authorization_list=authorization_list,
+        gas_limit=total_gas_cost,
+        max_fee_per_gas=GAS_PRICE,
+        max_priority_fee_per_gas=GAS_PRICE,
+    )
```
marioevz, line 81: We could check the receipts instead to avoid requiring a hard-coded gas price for this test:

```suggestion
    tx = Transaction(
        sender=sender,
        to=recipient,
        value=0,
        authorization_list=authorization_list,
        gas_limit=total_gas_cost,
        expected_receipt=TransactionReceipt(
            gas_used=total_gas_cost,
        ),
    )
```

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_charges.py
```diff
@@ -0,0 +1,567 @@
...
+
+    post = {
+        sender: Account(
+            nonce=1,
+            balance=sender_initial_balance - total_gas_cost * GAS_PRICE,
+        ),
```
marioevz:
```suggestion
```
Can be removed entirely if the previous comment is applied.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+            BalCodeChange(
+                block_access_index=1,
+                new_code=scenario.applied_account.code,
+            )
+        ],
+    )
+
+
+@pytest.mark.parametrize(
+    "oog_charge", ["new_account", "account_write", "auth_base"]
```
marioevz: Can we add a `"succeeds"` case, just to verify that the off-by-one case of `auth_base` is correct (since that's the last charge that happens for the authorization).

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+
+    tx = Transaction(
+        sender=sender,
+        to=recipient,
+        value=0,
+        authorization_list=authorization_list,
+        gas_limit=gas_limit,
+        max_fee_per_gas=GAS_PRICE,
+        max_priority_fee_per_gas=GAS_PRICE,
+    )
```
marioevz:
```suggestion
    tx = Transaction(
        sender=sender,
        to=recipient,
        value=0,
        authorization_list=authorization_list,
        gas_limit=gas_limit,
        expected_receipt=TransactionReceipt(
            gas_used=gas_limit,
        ),
    )
```
Similar to a comment in the `test_authorization_charges.py` file, the hard-coded gas prices can be removed if we instead check the transaction receipt.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+    """
+    regular = fork.transaction_top_frame_gas_calculator()(
+        recipient_type=RecipientType.CONTRACT,
+        authorizations=authorizations,
+    )
+    state = fork.transaction_top_frame_state_gas(
+        recipient_type=RecipientType.CONTRACT,
+        authorizations=authorizations,
+    )
+    return regular + state
```
marioevz, line 93: We have an implicit convention on naming for these gas calculators:
- `transaction_top_frame_gas_calculator`: It should return the total amount of gas that is consumed given the input parameters, i.e. the sum of both dimensions, regular and state. Normally used by tests that don't care much about exact costs and just want a number that will make the transaction pass.
- `transaction_top_frame_regular_gas`/`transaction_top_frame_state_gas`: Returns the specific dimension.

I think we should commit this change at some point perhaps in another PR.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+        tx=tx,
+        post=post,
+        expected_block_access_list=expected_block_access_list,
+    )
+
+
+@pytest.mark.parametrize(
+    "recipient_charge", ["new_account", "delegation_access"]
+)
+def test_recipient_charge_oog_rolls_back_delegations(
```
marioevz, line 374: We can add another parametrization here to test for a successful execution when the tx does not run out of gas.
```python
@pytest.mark.parametrize(
    "succeeds",
    [
        pytest.param(False, id="fails"),
        pytest.param(True, id="succeeds"),
    ],
)
```

src/ethereum/forks/amsterdam/vm/eoa_delegation.py
marioevz: The only thing I don't like about point 1 is that this assumption holds _now_, but in the future (perhaps with frames), since this is not tightly coupled, we could forget to update this part of the code.
Perhaps we need to have something similar to `written_accounts` in `tx_env` for accounts whose write charge is already accounted for in the intrinsic gas cost.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+    The many-authorization reservoir transaction, now moving value to an
+    empty recipient, carries both classes of state charge at once --
+    and a failure settles each class by whether its state survives.
+
+    The value transfer adds the recipient's ``NEW_ACCOUNT`` dispatch
+    charge alongside the authorizations' ``NEW_ACCOUNT`` + ``AUTH_BASE``
+    charges. The recipient is the bn254 pairing precompile: the one
+    empty account that still executes, so an execution-phase failure is
+    reachable (a 1-byte input makes the precompile exceptionally halt
+    after the value has moved). An empty recipient runs no code of its
```
marioevz: Picking this precompile is ok, just maybe rewording that we are picking _a_ precompile, not the only one that works for this example (e.g. blake2 fails with empty data so that one could also work in theory).

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+            balance=sender_initial_balance - gas_used * GAS_PRICE,
+        ),
+        auth.authority: auth.applied_account,
+        recipient: Account(code=halt_code, balance=0),
+    }
+
+    state_test(pre=pre, tx=tx, post=post)
+
+
+def test_auth_state_gas_in_header_on_dispatch_revert(
```
marioevz, line 1187: We might need a negative test for a block that is correctly rejected because the last transaction barely does not fit in the block gas limit boundaries because of state gas consumed in the dispatch of the prior transactions.

Edit: *Prior reverted transactions

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+    # reservoir, only the authorization's state gas is consumed -- its
+    # delegation persists -- and the unused remainder returns.
+    gas_used = cap + auth_state_gas
+
+    tx = Transaction(
+        sender=sender,
+        to=recipient,
+        value=0,
+        authorization_list=authorization_list,
+        gas_limit=gas_limit,
```
marioevz:
```suggestion
        state_gas_reservoir=reservoir,
```
Should also work here.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+    # limit must still cover the state charge while it is outstanding.
+    gas_used = intrinsic_regular + auth_charges + exec_regular
+    gas_limit = gas_used + exec_state + 10_000
+
+    tx = Transaction(
+        sender=sender,
+        to=recipient,
+        value=0,
+        authorization_list=authorization_list,
+        gas_limit=gas_limit,
```
marioevz:
```suggestion
```
Gas limit can be completely omitted here.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+
+    gas_used = regular_total + auth_state
+    gas_limit = gas_used + 10_000
+
+    tx = Transaction(
+        sender=sender,
+        to=recipient,
+        value=0,
+        authorization_list=authorization_list,
+        gas_limit=gas_limit,
```
marioevz:
```suggestion
```
Can be removed.

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
```diff
@@ -0,0 +1,1381 @@
...
+    # budget returns.
+    gas_used = intrinsic_regular + auth_charges + revert_exec_gas
+    gas_limit = gas_used + 10_000
+
+    tx = Transaction(
+        sender=sender,
+        to=recipient,
+        value=0,
+        authorization_list=authorization_list,
+        gas_limit=gas_limit,
```
marioevz:
```suggestion
```
Can be removed.

marioevz (commented):
Thanks, some comments but overall looks great!

src/ethereum/forks/amsterdam/vm/eoa_delegation.py
gurukamath: Actually, the "not tightly coupled" part is not new to the Evm. In all calls, the charge for value transfer happens at the parent frame level whereas the actual value transfer is at the child frame.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7702.py
gurukamath: Applied — and tightened the success case to exact gas (the delegated STOP costs nothing), so it now pins the boundary from above.

src/ethereum/forks/amsterdam/vm/interpreter.py
gurukamath: applied

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
gurukamath: applied

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
gurukamath: done

tests/amsterdam/eip2780_reduce_intrinsic_tx_gas/test_authorization_oog.py
gurukamath: Added test_reverted_dispatch_state_gas_counts_toward_block_limit: the reverted tx's persisted authorization state gas squeezes out the last tx (exceeded rejected with GAS_ALLOWANCE_EXCEEDED, exact_fit control accepted, rejection pinned to the state dimension).

marioevz (approved):
Thanks for the fixes. I've pushed another commit to fix the outstanding failing ported-static tests, so now all tests are passing.

PR #3127, opened 2026-07-08
feat(fill-stateful): make the per-test chain rewind optional (add debug_resetHead)

skylenet:
## 🗒️ Description

Ports #3050 (merged into `devnets/bal/7-bench`) to `forks/amsterdam`.

The per-test chain rewind (`_reset_chain_between_tests`) hard-required geth's `debug_setHead`. This makes it optional and client-agnostic:

- **adds `debug_resetHead`** (by block hash) as an alternative — for clients like Nethermind that don't implement `debug_setHead`;
- **makes the debug rewind optional**: if a client implements neither, fill-stateful keeps going instead of aborting — each test's first block is built on its explicit `start_block` parent, so the client reorgs onto it without any debug call.

So fill-stateful is no longer tied to the geth `debug` namespace. Nethermind is the concrete client this was validated against (on the original PR), and it now works as a filler.

## Details

- **Three-tier per-test rewind** in `rewind_head`, probed once and cached:
  1. `debug_setHead` (by number) — geth; **truncates** the chain, keeping the block tree small;
  2. else `debug_resetHead` (by hash) — e.g. Nethermind;
  3. else **nothing** — rely on the explicit `start_block` parent; the client reorgs onto it.
- **Read per-test state at `start_block`, not `"latest"`** (the seed-nonce sync and the post-rewind check). A `debug_resetHead`-style rewind (or no rewind) restores the build state but leaves `latest` at the previous test's tip, so reading `"latest"` there returns a stale nonce and every funding tx is rejected (`Invalid nonce - expected 0`). Identical to `"latest"` on clients (geth) whose `debug_setHead` moves it.
- Two small base fill fixes (deploy gas safety-buffer cap; skip redundant seed withdrawal funding).

## Validation

The behavior was validated on the original PR (#3050) against the bloatnet stateful repricing prestate (Amsterdam) with `nethermindeth/nethermind:master`: 225/225 fills via both the `debug_resetHead` and the no-debug (`"none"`) path, replaying clean on geth and Nethermind. This port is the same net diff applied onto `forks/amsterdam` (applied cleanly with no conflicts); geth filling is unaffected (it takes the `debug_setHead` tier, and `latest` already tracks the rewind).

## 🔗 Related Issues or PRs

Ports #3050.

## ✅ Checklist

    ```console
    just static
    ```

#### Cute Animal Picture

cute cat

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
@@ -271,7 +271,24 @@ def _compute_deploy_gas_limit(
         new_bytes=len(bytes(initcode))
     )
     regular_gas += storage_slots * sstore_regular_gas
-    regular_gas *= 2
+
+    # Double as a safety buffer since gas estimation is approximate. The buffer
+    # must not, by itself, push a contract that genuinely deploys within the
+    # EIP-7825 regular-gas cap over it: when the unbuffered estimate still fits
+    # the cap, clamp the limit to the cap instead. The deploy then runs with a
+    # cap-sized regular limit and consumes only its (smaller) actual gas.
+    # Only a contract whose unbuffered estimate exceeds the cap is truly
+    # undeployable (the caller raises on that).
+    buffered_regular_gas = regular_gas * 2
+    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
+    if (
+        tx_gas_limit_cap is not None
+        and buffered_regular_gas > tx_gas_limit_cap
+        and regular_gas <= tx_gas_limit_cap
+    ):
+        regular_gas = tx_gas_limit_cap
+    else:
+        regular_gas = buffered_regular_gas
 
     # State portion, from the block reservoir.
     state_gas = intrinsic_state_gas
--- a/docs/filling_tests/fill_stateful.md
+++ b/docs/filling_tests/fill_stateful.md
@@ -13,7 +13,7 @@ The target client must expose:
 
 - `testing` (`testing_buildBlockV1`) — block construction with explicit transaction ordering.
 - `engine` — `engine_newPayloadVX`, `engine_forkchoiceUpdatedVX`.
-- `eth`, `debug` — chain queries and `debug_setHead` for between-test rewind.
+- `eth`, `debug` — chain queries and `debug_setHead` (or `debug_resetHead` on clients like Nethermind that lack `debug_setHead`) for between-test rewind.
 - `web3` (optional) — `web3_clientVersion` is recorded into the fixture's `_info.filling-transition-tool` for traceability.
 
 The production-ready filler is `ethpandaops/geth:master`.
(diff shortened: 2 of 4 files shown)


jochem-brouwer:
EDIT: let me revisit this comment as it might have assumptions which do not hold anymore

One comment, on https://github.com/NethermindEth/gas-benchmarks we do rewinds using the engine API, and not the debug endpoint. This is not in the scope for this PR, but we should add this feature soon-ish. Gas-benchmarks inserts an empty "anchor" block on top of the block we fork on, we do not build on this block, but we can use `engine_forkchoiceUpdated` to rollback to this anchor block. Most clients do not allow rewinding to a canonical block (i.e. a parent block of the current canonical chain) it instead has to be a new chain, forked off some parent block (and can have multiple blocks on top of it).
I can open an issue for this, stateful should rewind via this strategy and not via the non-standardized `debug` namespace. Should I do that @danceratopz ?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
```diff
@@ -271,7 +271,24 @@ def _compute_deploy_gas_limit(
...
+    buffered_regular_gas = regular_gas * 2
+    tx_gas_limit_cap = fork.transaction_gas_limit_cap()
+    if (
+        tx_gas_limit_cap is not None
+        and buffered_regular_gas > tx_gas_limit_cap
+        and regular_gas <= tx_gas_limit_cap
+    ):
+        regular_gas = tx_gas_limit_cap
+    else:
+        regular_gas = buffered_regular_gas
```
jochem-brouwer, line 291: I am not entirely sure here, the state reservoir introduced by EIP-8037 should actually make most (not all) contracts deployable. The "only" reason why a contract cannot get deployed is when the total regular gas spent is more than the EIP-7825 gas limit cap (this caps the *regular gas used*), or if the total state gas spent is more than the block gas limit. Note that the tx gas limit is the "regular gas limit", the state gas limit is (on tx level) infinite, and is "only" bounded by the block gas limit (state gas spent should fit within this gas limit)

If we thus have a `tx_gas_limit_cap` (I think we should rename this to something more descriptive, e.g. `tx_regular_gas_limit_cap` (beyond the scope of this PR), cc @danceratopz ?) where the regular gas is more than the `tx_gas_limit_cap` then we can already throw here, as there is "no way" (as of Glamsterdam fork) to deploy this contract.

jochem-brouwer (commented):
I left previous [comment](https://github.com/ethereum/execution-specs/pull/3127#issuecomment-4925561973) without reading this PR thorougly, which I should not do. I opened an [issue](https://github.com/ethereum/execution-specs/issues/3134) explaining the next steps. For now, aside one comment, this looks correct. I'd rather have someone who is "deeper" in the fill implementation to also take a look, so will not explicitly approve here.

The clients which do not support those debug endpoints likely will throw errors on some specific benchs which require state reading. All state reads which read from "latest" could fail on that client, in particular those benchs which query nonces (benchs using EIP-7702 delegations). The issue linked should fix that, but this is definitely a step in the correct direction 😄 👍 

Thanks @skylenet 😄 👍

spencer-tb (approved):
From a framework perspective looks fine to me.

@LouisTsai-Csie you'll need to cherry-pick to `denvets/glamsterdam/6-benchmark` I think?

Feel free to merge from my side! Can address the follow up issue in the future

packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/pre_alloc.py
spencer-tb: Regular makes sense to me!

jochem-brouwer (approved):
LGTM, thanks @skylenet 😄 👍

Issue #3128, opened 2026-07-09
Benchmark test wrapper compatibility with `glamsterdam-devnet-6` spec


LouisTsai-Csie:
## Summary

Benchmark tests use a custom test wrapper with built-in code generators (`JumpLoopGenerator`, `ExtCallGenerator`) and transaction-splitting (`split_transaction`) features. This analysis examines how `glamsterdam-devnet-6` spec (especially repricing EIPs) affects these components.

### Code Generators
`JumpLoopGenerator` fills a contract with the target opcode up to max code size. During execution, it runs through the contract and executes a `JUMP` back to the beginning, repeating this loop until the transaction OOG.

`ExtCallGenerator` creates two contracts: a loop contract that repeatedly calls into a target contract, which executes the given operations ~1024 times (until EVM stack limit). `ExtCallGenerator` is preferred when opcodes push items onto the stack. For example, `CALLVALUE` pushes the call value; with `JumpLoopGenerator`, you would need to repeat `POP(CALLVALUE)` in the loop, or you will quickly hit the stack limits. With `ExtCallGenerator`, the loop contract `STATICALL`s into the target, which executes only `CALLVALUE` up to stack size. This is confirmed to be optimized version of the `CALLVALUE` benchmark in Osaka.

These two patterns are used extensively for compute benchmarks involving non-state-changing operations (`CREATE`, `SSTORE` are avoided). Such operations don't depend on state changes, so OOG isn't a concern.

### Transaction Splitting
Under the benchmark test wrapper, users pass a code generator or transaction. The framework uses `split_transactions` to fill blocks respecting the per-transaction gas cap (16M). With a 60M block limit and 16M tx cap, it generates 3 full (16M each) + 1 partial (12M) transaction.

### Concern 1

Do the increased account access costs in repricing EIPs break the `ExtCallGenerator` advantage over `JumpLoopGenerator`?

If the access cost of `*CALL` increases significantly beyond the overhead introduced by `POP`, we should consider replacing `ExtCallGenerator` with `JumpLoopGenerator`. Currently, `ExtCallGenerator` is preferred because it avoids stack-pushing opcodes that requires the `POP` overhead.

Based on EIP-8038, `WARM_ACCESS` stays the same, and `COLD_ACCOUNT_ACCESS` increases from 2600 to 3000, a minimal overhead increase. 

[...]

PR #3098, opened 2026-07-03
refactor(test-benchmark): update implementation, align `glam-devnet-6` spec

LouisTsai-Csie:
## 🗒️ Description

Updated geth binary from `bal-devnet-7` to `glamsterdam-devnet-6`. The bench-gas job (`--gas-benchmark-values 1`) failed on 13 cases.

Root cause: benchmark framework helper assumed single-dimensional gas; EIP-8037 introduces state-gas, breaking this assumption -> `generators.py`: now accounts for state gas (not just regular gas)

Test changes:
- test_storage_access_cold: handle OOG and revert scenarios for gas refunding
- test_selfdestruct_created / test_creates_collisions: correct gas accounting.
- More refactor.

Follow Ups:
- Intrinsic state gas (e.g. from authorizations) is deliberately not modeled in gas_cost; noted in the `gas_cost` docstring (`packages/testing/src/execution_testing/tools/tools_code/generators.py:794`).

## 🔗 Related Issues or PRs

N/A.

## ✅ Checklist

- [ ] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) and [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/):
    ```console
    just static
    ```
- [ ] All: PR title have the form `<type>(<area>):`, where `<type>` and `<area>` come from an approrpriate `C-<type>`, respectively `A-<area>`, label. The title should match the a target squash commit message.
- [ ] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [ ] All: Set appropriate labels for the changes (only maintainers can apply labels).
- [ ] Tests: For PRs implementing a missed test case, update the [post-mortem document](/ethereum/execution-specs/blob/HEAD/docs/writing_tests/post_mortems.md) to add an entry the list.
- [ ] Ported Tests: Add the following docstring to manually enhanced tests from `./tests/ported_static/`:
    ```text
	@manually-enhanced: Do not overwrite. Post-state expectations corrected
	manually (see PR #2784).
	````

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3135, opened 2026-07-09
fix(test-benchmark): make audited compute benchmarks measure real work

spencer-tb:
### Description

Fixes benchmarks in #3098 that filled green but silently measured nothing, their target operations were faulting, running out of gas, or being rejected before doing any real work.

- **`test_storage_access_cold_benchmark` / `test_storage_access_warm_benchmark` (SSTORE cases) and `test_tload` (`fixed_value=False`)**. The `ExtCallGenerator` wrapped every target in `STATICCALL`, so the SSTORE/TSTORE faulted on the first op and the benchmark did zero work, the caller now uses `CALL` when the target contains a state changing opcode. The fix is at the generator, so it also covers any future state changing `ExtCallGenerator` benchmark.
- **`test_deploy_then_interact`**. The deploy budget used the pre-8037 200/byte code-deposit cost, so every deploy ran out of gas and the interact txs hit empty accounts. It now budgets the creation code's actual gas cost.
- **`test_ether_transfers_to_precompile`**. The `gas_limit` was intrinsic only, leaving zero budget so every precompile call OOG'd, adds each precompiles empty input cost and pre funds the precompile so value transfers don't pay the one time new account state gas.
- **`test_alt_bn128`, the `ec_pairing_2_sets`, `ec_pairing_3_pair`, `ec_pairing_1_pair` and `ec_pairing_1_pair_empty` cases**. These used off curve points, wrong length input, or empty calldata and were rejected before any pairing ran. Replaced with valid on curve, correctly sized pairs so each does real Miller loop work.
- **`test_bls12_381_uncachable`, the `bls12_g1msm` / `bls12_g2msm` cases**. The MSM scalar was exactly the subgroup order `r`, so `r·P = O` and the output fed back loop multiplied infinity forever, uses `r - 2` to keep every call a real scalar multiplication over a non-repeating sequence.

### Related Issues or PRs

Follow-up to #3098, related to #3128.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/benchmark/benchmark_code_generator.py
+++ b/packages/testing/src/execution_testing/benchmark/benchmark_code_generator.py
@@ -100,19 +100,26 @@ class ExtCallGenerator(BenchmarkCodeGenerator):
 
         # Create caller contract that repeatedly calls the target contract
         # attack = POP(
-        #             STATICCALL(GAS, target_contract_address, 0, 0, 0, 0)
+        #             (STATIC)CALL(GAS, target_contract_address, ...)
         #          )
         #
         # setup + JUMPDEST + attack + attack + ... + attack +
         # JUMP(setup_length)
+        #
+        # The target must be entered via CALL when it contains
+        # state-changing opcodes: a STATICCALL'd frame faults on the
+        # first one and the target executes nothing.
+        call_opcode = (
+            Op.CALL if self.uses_state_changing_opcode() else Op.STATICCALL
+        )
         code_sequence = Op.POP(
-            Op.STATICCALL(
-                Op.GAS,
-                self._target_contract_address,
-                Op.PUSH0,
-                Op.CALLDATASIZE,
-                Op.PUSH0,
-                Op.PUSH0,
+            call_opcode(
+                gas=Op.GAS,
+                address=self._target_contract_address,
+                args_offset=Op.PUSH0,
+                args_size=Op.CALLDATASIZE,
+                ret_offset=Op.PUSH0,
+                ret_size=Op.PUSH0,
             )
         )
 
--- a/packages/testing/src/execution_testing/specs/benchmark.py
+++ b/packages/testing/src/execution_testing/specs/benchmark.py
@@ -64,6 +64,23 @@ class OpcodeTarget:
         return self.name
 
 
+STATE_CHANGING_OPCODES = (
+    Op.SSTORE,
+    Op.TSTORE,
+    Op.CREATE,
+    Op.CREATE2,
+    Op.CALL,
+    Op.CALLCODE,
+    Op.SELFDESTRUCT,
+    Op.LOG0,
+    Op.LOG1,
+    Op.LOG2,
+    Op.LOG3,
+    Op.LOG4,
+)
+"""Opcodes that are illegal inside a static context (EIP-214)."""
+
+
 @dataclass(kw_only=True)
 class BenchmarkCodeGenerator(ABC):
     """Abstract base class for generating benchmark bytecode."""
@@ -81,6 +98,18 @@ class BenchmarkCodeGenerator(ABC):
     def deploy_contracts(self, *, pre: Alloc, fork: Fork) -> Address:
         """Deploy any contracts needed for the benchmark."""
         ...
+
+    def uses_state_changing_opcode(self) -> bool:
+        """
+        Return whether the setup or attack block contains an opcode that
+        is illegal in a static context, in which case the target contract
+        must be entered via CALL instead of STATICCALL (a STATICCALL'd
+        frame faults on the first such opcode and executes nothing).
+        """
+        target_code = bytes(self.setup) + bytes(self.attack_block)
+        return any(
+            bytes(opcode) in target_code for opcode in STATE_CHANGING_OPCODES
+        )
 
     def deploy_fix_count_contracts(self, *, pre: Alloc, fork: Fork) -> Address:
         """Deploy the contract with a fixed opcode count."""
@@ -107,27 +136,10 @@ class BenchmarkCodeGenerator(ABC):
             Op.PUSH0, Op.PUSH0, Op.CALLDATASIZE
         ) + Op.PUSH4(iterations)
 
-        is_state_changing_set = [
-            Op.SSTORE,
-            Op.TSTORE,
-            Op.CREATE,
-            Op.CREATE2,
-            Op.CALL,
-            Op.CALLCODE,
-            Op.SELFDESTRUCT,
-            Op.LOG0,
-            Op.LOG1,
-            Op.LOG2,
-            Op.LOG3,
-            Op.LOG4,
-        ]
-
         # Select CALL for state-changing opcodes, STATICCALL otherwise
-        uses_state_changing_opcode = any(
-            bytes(opcode) in bytes(self.attack_block)
-            for opcode in is_state_changing_set
+        call_opcode = (
+            Op.CALL if self.uses_state_changing_opcode() else Op.STATICCALL
         )
-        call_opcode = Op.CALL if uses_state_changing_opcode else Op.STATICCALL
 
         opcode = (
             prefix
--- a/tests/benchmark/compute/eip7928_block_level_access_lists/test_block_access_list.py
+++ b/tests/benchmark/compute/eip7928_block_level_access_lists/test_block_access_list.py
@@ -629,16 +629,6 @@ def test_deploy_then_interact(
 
     runtime_code, setup_gas, _, reserve_gas = _build_keccak_chain(fork)
 
-    creation_code = Initcode(
-        deploy_code=runtime_code,
-        initcode_prefix=Op.SSTORE(0, 1),
-    )
-
-    intrinsic_gas_create = intrinsic_gas_calculator(
-        calldata=bytes(creation_code),
-        contract_creation=True,
-    )
-
     initcode_sstore = Op.SSTORE(
         0,
         1,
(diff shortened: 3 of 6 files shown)


packages/testing/src/execution_testing/benchmark/benchmark_code_generator.py
```diff
@@ -100,19 +100,26 @@ def deploy_contracts(self, *, pre: Alloc, fork: Fork) -> Address:
...
+        )
         code_sequence = Op.POP(
-            Op.STATICCALL(
-                Op.GAS,
-                self._target_contract_address,
-                Op.PUSH0,
-                Op.CALLDATASIZE,
-                Op.PUSH0,
-                Op.PUSH0,
+            call_opcode(
```
jochem-brouwer, line 116: This STATICCALL on state-changing targets is something what thus also was broken in the past, which we did not catch, again raises the question how we should catch these. We should really be confident that our benchmarks do what we expect them to do, and keep finding situations where the bench is clearly broken 🤔

packages/testing/src/execution_testing/benchmark/benchmark_code_generator.py
```diff
@@ -100,19 +100,26 @@ def deploy_contracts(self, *, pre: Alloc, fork: Fork) -> Address:
...
+                gas=Op.GAS,
+                address=self._target_contract_address,
```
jochem-brouwer, line 118: If `call_opcode == Op.CALL` then inject `value=Op.PUSH0` (this saves 1 gas per call opcode (PUSH1 vs PUSH0))

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -64,6 +64,23 @@ def __str__(self) -> str:
         return self.name
 
 
+STATE_CHANGING_OPCODES = (
+    Op.SSTORE,
+    Op.TSTORE,
+    Op.CREATE,
+    Op.CREATE2,
+    Op.CALL,
+    Op.CALLCODE,
```
jochem-brouwer, line 73: CALL is only state changing if value > 0 (not sure if we should enforce this).
EIP-214:

> As an exception, CALLCODE is not considered state-changing, even with a non-zero value.

So CALLCODE should be removed (the value field of this broken/deprecated/bugged opcode sets msg.value in the new message frame, but does not actually move value (it "moves" value from the current account to the target account where the current context execute, which is the current account, so there is no net transfer, and there is also no situation where a net transfer from an account to a distinct one is possible)

packages/testing/src/execution_testing/specs/benchmark.py
```diff
@@ -82,6 +99,18 @@ def deploy_contracts(self, *, pre: Alloc, fork: Fork) -> Address:
...
+    def uses_state_changing_opcode(self) -> bool:
+        """
+        Return whether the setup or attack block contains an opcode that
+        is illegal in a static context, in which case the target contract
+        must be entered via CALL instead of STATICCALL (a STATICCALL'd
+        frame faults on the first such opcode and executes nothing).
+        """
+        target_code = bytes(self.setup) + bytes(self.attack_block)
+        return any(
+            bytes(opcode) in target_code for opcode in STATE_CHANGING_OPCODES
```
jochem-brouwer, line 111: I think this state changing property should be part of the `Op`. Not a blocker for this PR, can refactor later, but can also introduce it here.
We would then also have to distinguish between CALL opcodes where the value is 0 or not, which in some cases cannot be statically analyzed (for instance it reads the value to send from a memory position), not sure what to do there, always return that it changes in case we cannot statically analyze? Or return a type which indicates it "maybe" is state changing?

tests/benchmark/compute/eip7928_block_level_access_lists/test_block_access_list.py
```diff
@@ -647,12 +637,22 @@ def test_deploy_then_interact(
...
         new_value=1,
     )
-    initcode_exec_gas = initcode_sstore.gas_cost(fork)
-    code_deposit_gas = 200 * len(runtime_code)
+    creation_code = Initcode(
+        deploy_code=runtime_code,
+        initcode_prefix=initcode_sstore,
+    )
+
+    intrinsic_gas_create = intrinsic_gas_calculator(
```
jochem-brouwer, line 645: This has the problem with EIP-2780 that any "extra" costs (account create costs charged as top-frame costs, "after" the intrinsic gas cost deduction) are not covered here. These extra costs might be 0 which means this will cover the costs, however, we should have a mechanism to easily get the required gas limit costs (something like `eth_estimateGas`) to succesfully deploy.

tests/benchmark/compute/precompile/test_alt_bn128.py
```diff
@@ -251,25 +251,38 @@
             )
             # Second pairing
             + PointG1(
-                x=0x0000000000000000000000000000000000000000000000000000000000000013,
-                y=0x0644E72E131A029B85045B68181585D97816A916871CA8D3C208C16D87CFD451,
+                x=0x17072B2ED3BB8D759A5325F477629386CB6FC6ECB801BD76983A6B86ABFFE078,
```
jochem-brouwer, line 254: Can you add a comment here what these numbers are and, if they are generated, how these were generated?

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -277,6 +277,24 @@ def test_ether_transfers(
...
+EMPTY_INPUT_PRECOMPILE_GAS: dict[int, int] = {
+    0x01: 3_000,  # ECRECOVER charges fully, recovers nothing.
+    0x02: 60,  # SHA2-256 base cost.
+    0x03: 600,  # RIPEMD-160 base cost.
+    0x04: 15,  # IDENTITY base cost.
+    0x05: 500,  # MODEXP minimum (EIP-7883).
+    0x06: 150,  # BN128ADD (zero-padded input adds two infinities).
+    0x07: 6_000,  # BN128MUL.
+    0x08: 45_000,  # BN128PAIRING base (zero pairs).
+    0x100: 6_900,  # P256VERIFY charges fully, verifies nothing.
```
jochem-brouwer, line 289: I think we should read these constants from fork config (just like how we would query intrinsic costs or opcode gas costs)

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -277,6 +277,24 @@ def test_ether_transfers(
...
+}
+"""
+Gas each precompile consumes for empty calldata. Precompiles absent from
```
jochem-brouwer, line 292: (shouldn´t this comment be above the table or is this some lint rule which enforces multi-strings below what is targeted? 🤔 )

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -288,11 +306,28 @@ def test_ether_transfers_to_precompile(
...
 ) -> None:
     """Test a block full of ether transfers to a precompile address."""
-    # A precompile already exists, so a value transfer pays the EIP-2780
-    # value-transfer charge but never the new-account state gas.
-    iteration_cost = fork.transaction_intrinsic_cost_calculator()(
-        sends_value=transfer_amount > 0,
-        recipient_type=RecipientType.PRECOMPILE,
+    precompile_id = int.from_bytes(bytes(Address(precompile)), "big")
+    if precompile_id not in EMPTY_INPUT_PRECOMPILE_GAS:
+        pytest.skip(
```
jochem-brouwer, line 311: If this is the case we should use some default calldata which the precompile accepts for the transfer to succeed (precompile does not OOG/REVERT on such calldata), instead of skipping it 🤔

jochem-brouwer (changes requested):
Some comments and some stuff which I believe should be changed 😄 👍 

Not all of this has to be in this PR, we can do a follow-up PR, of course open for discussion if all my comments are needed. If we do a follow-up, we should (likely) track this with an issue so we do not forget 😄 👍

spencer-tb:
On where the bn128 pairing points came from: they're **generated with `py_ecc`, not copied from an external source**, so they're guaranteed on-curve and in the prime-order subgroup.

```python
from py_ecc.bn128 import G1, G2, multiply

# G1: PointG1(x, y) = k * G1  (affine)
p = multiply(G1, k)

# G2: k * G2, with the FQ2 coeffs swapped to match the precompile's
# (x_imag, x_real) decode order in bytes_to_g2
q = multiply(G2, k)                       # q[i].coeffs == (c0, c1)
x = (q[0].coeffs[1], q[0].coeffs[0])
y = (q[1].coeffs[1], q[1].coeffs[0])
```

Scalars used (small, arbitrary):

| case | points |
|------|--------|
| `ec_pairing_1_pair` | `3·G1`, `5·G2` |
| `ec_pairing_2_sets` | pair-0 unchanged, pair-1 = `7·G1`, `11·G2` |
| `ec_pairing_3_pair` | `(2·G1, 3·G2)`, `(5·G1, 7·G2)`, `(11·G1, 13·G2)` |
| `ec_pairing_1_pair_empty` | `bytes(192)` — one correctly sized `(∞, ∞)` pair |

I verified each assembled calldata through the actual EELS `bytes_to_g1` / `bytes_to_g2` decoders plus the subgroup check (`curve_order · P == ∞`), and confirmed each pair drives a real Miller loop over the *filled* fixtures (the previous constants were off-curve / wrong-length and were rejected before any pairing ran).

tests/benchmark/compute/precompile/test_alt_bn128.py
spencer-tb: https://github.com/ethereum/execution-specs/pull/3135#issuecomment-4957124001, sry the comment was on the original PR, forgot to copy it over!

spencer-tb:
@jochem-brouwer, i added your requested fixes in this commit: https://github.com/ethereum/execution-specs/pull/3187/commits/1076d6ee9058a6c7d14adc79ebe0d2cabf1fb8ec

Closing this PR in favor of it! Feel free to comment or ping any issues :)

PR #3136, opened 2026-07-09
fix(test-benchmark): correct compute benchmarks that ran below worst case

spencer-tb:
### Description

Follow-up to #3135. Where #3135 fixes benchmarks that measured *nothing*, this fixes three that measured the "wrong thing" where the gas is exact, but the work is below worst case.

- **`test_arithmetic`, drop the MOD/SMOD/ADDMOD/MULMOD cases.** Their operand chain collapses to `x mod 0` after ~150 iterations, so >99.99% of the loop is trivial (the params are even commented "not suitable ... values quickly become zero"). The dedicated `test_mod` / `test_mod_arithmetic` already benchmark these at worst case and carry the `repricing` marker, so these cases are redundant duplicates.
- **`test_bls12_g1_msm` / `test_bls12_g2_msm`, use distinct points and scalars.** The input was a single point `P1` repeated with scalar = the subgroup order `r`, so every term was `r·P1 = O` and an MSM can bucket the identical scalars / factor out the shared point. Distinct points paired with distinct full width scalars below `r` force the full k term computation. The exactly redundant k=128/k=64 cases in the `test_bls12_381` sweep (which carry no `repricing` marker) are removed.

### Related Issues or PRs

Follow-up to #3135, related to #3128.

### Checklist

### Cute Animal Picture

### DIFF
--- a/tests/benchmark/compute/instruction/test_memory.py
+++ b/tests/benchmark/compute/instruction/test_memory.py
@@ -20,7 +20,12 @@ from execution_testing import (
 
 
 @pytest.mark.repricing(mem_size=1)
-@pytest.mark.parametrize("mem_size", [0, 1, 1_000, 100_000, 1_000_000])
+# MSIZE is O(1), so its cost does not vary with memory size. mem_size only
+# controls the one-time expansion in setup; capping it at 100_000 keeps that
+# expansion a minority of the measured gas. A larger value (e.g. 1_000_000)
+# would make the quadratic expansion dwarf the MSIZE loop, benchmarking
+# expansion rather than MSIZE.
+@pytest.mark.parametrize("mem_size", [0, 1, 1_000, 100_000])
 def test_msize(
     benchmark_test: BenchmarkTestFiller,
     mem_size: int,
--- a/tests/benchmark/compute/instruction/test_memory.py
+++ b/tests/benchmark/compute/instruction/test_memory.py
@@ -11,33 +11,50 @@ Supported Opcodes:
 
 import pytest
 from execution_testing import (
+    BenchmarkCodeGenerator,
     BenchmarkTestFiller,
     Bytecode,
     ExtCallGenerator,
+    Fork,
     JumpLoopGenerator,
     Op,
 )
 
 
 @pytest.mark.repricing(mem_size=1)
-# MSIZE is O(1), so its cost does not vary with memory size. mem_size only
-# controls the one-time expansion in setup; capping it at 100_000 keeps that
-# expansion a minority of the measured gas. A larger value (e.g. 1_000_000)
-# would make the quadratic expansion dwarf the MSIZE loop, benchmarking
-# expansion rather than MSIZE.
-@pytest.mark.parametrize("mem_size", [0, 1, 1_000, 100_000])
+# MSIZE should be O(1), but sweep mem_size so a size-dependent
+# implementation shows up as a regression. ExtCallGenerator re-expands
+# memory in every call frame, so once the expansion outweighs a frame's
+# MSIZE work (~16 KiB), loop in a single frame instead, paying one POP
+# per MSIZE but expanding only once.
+@pytest.mark.parametrize("mem_size", [0, 1, 1_000, 100_000, 1_000_000])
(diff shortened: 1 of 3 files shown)


tests/benchmark/compute/instruction/test_memory.py
```diff
@@ -20,7 +20,12 @@
 
 
 @pytest.mark.repricing(mem_size=1)
-@pytest.mark.parametrize("mem_size", [0, 1, 1_000, 100_000, 1_000_000])
+# MSIZE is O(1), so its cost does not vary with memory size. mem_size only
```
jochem-brouwer: > MSIZE is O(1), so its cost never varies with memory size.

This assumes the *implementation* of the MSIZE opcode is O(1). There could be (naive) implementations of EVM memory where the memory size is not O(1) and for instance does some kind of counting, before returning the memory size. This is something what would be flagged by benchmarks due to a regression. I believe we should keep testing this, with at least 3 distinct values so the benchmark results make it clear that something is off

tests/benchmark/compute/instruction/test_memory.py
```diff
@@ -20,7 +20,12 @@
...
+# controls the one-time expansion in setup; capping it at 100_000 keeps that
+# expansion a minority of the measured gas. A larger value (e.g. 1_000_000)
+# would make the quadratic expansion dwarf the MSIZE loop, benchmarking
+# expansion rather than MSIZE.
```
jochem-brouwer: I disagree, for 1 million the memory expansion cost is:

```
words = 1_000_000 // 32
expansion_cost = words * 3 + (words * words) // 512
```

This costs 2_001_098. The regular gas cost is capped at ~16.7 million, so we still have ~14.7 million gas left to spend on MSIZE (~11.9% is spent on the memory expansion)

I think we should keep `[1, 1000, 1_000_000]` (or possibly swap 1 million with 100_000)

EDIT: nvm, in current setup this is correct, due to ExtCallGenerator we keep instantiating contracts with empty memory so we indeed pay mostly the memory expansion costs. We should port the variants where more is paid on memory expansion to the JumpCallGenerator instead of the ExtCallGenerator. At ~16384 memory size (in bytes) the expansion cost is ~2048, which is equal to 1024 (max stack size) times the MSIZE cost (2), so that is the point where we should re-use the context at the cost of using Op.POP to avoid paying memory expansion costs.

tests/benchmark/compute/precompile/test_bls12_381.py
```diff
@@ -279,6 +247,40 @@ def _generate_bls12_g2_point(seed: int) -> Bytes:
...
+    """
+    Build a k-pair G1MSM input that resists MSM shortcuts.
+
+    Identical points would let the precompile factor out (sum sᵢ)·P, and
+    identical scalars would let Pippenger bucket every term together. Pairing
+    distinct points with distinct full-width scalars below the subgroup order
+    (so no term collapses to the point at infinity) forces the full k-term
+    computation.
+    """
+    rng = random.Random(0)
```
jochem-brouwer, line 260: We should further optimize this benchmark (in a later PR) where we move this calldata into contract code so we do not waste gas on the calldata itself. To further bench, we should figure out if we can port the `_generate_bls12_g1_point` to EVM: the goal is to create distinct G1/G2 points where we chain each transaction depending on the result of the earlier one (so without BALs you cannot parallelize them as you need the result of the previous txs). 
(not a blocker, should be addressed in a new PR)

jochem-brouwer (changes requested):
I disagree with removing different-sized MSIZE benchs as this assumes a correct implementation of MSIZE in O(1). (Always open for discussion 😄 👍 )

tests/benchmark/compute/instruction/test_memory.py
spencer-tb: Very fair point! Lets keep it, regressions are important. Will add it back :)

spencer-tb:
Updated @jochem-brouwer! Thanks for the review :D

jochem-brouwer (approved):
LGTM! 🎉

PR #3140, opened 2026-07-09
fix(tests): EIP-8037: failing test

marioevz:
### Description

Fixes test that broken during the spec changes of #3120:

Since the floor data cost increased due to the anchor change, the test was always consuming the data floor instead of the expected gas. 

Fix was to increase the regular used by the initcode via memory increment.

### Related Issues or PRs
N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_create.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_create.py
@@ -2123,7 +2123,9 @@ def test_create_account_charge_reduces_child_gas(
 @pytest.mark.parametrize(
     "init_code",
     [
-        pytest.param(Op.REVERT(0, 0), id="revert"),
+        pytest.param(
+            Op.REVERT(0, 10_000, new_memory_size=10_000), id="revert"
+        ),
         pytest.param(Op.INVALID, id="halt"),
     ],
 )
@@ -2139,6 +2141,9 @@ def test_failed_create_tx_refunds_intrinsic_new_account(
     refunded on creation-tx revert/halt. Block state-gas excludes it
     so header gas_used reflects only the regular component, and the
     sender's receipt reflects the same refund via cumulative_gas_used.
+
+    Gas consumed must be above the floor for the test to work, hence
+    the increased memory consumption in some of the initcodes.
     """
     intrinsic_calc = fork.transaction_intrinsic_cost_calculator()
     create_state_gas = fork.create_state_gas(code_size=0)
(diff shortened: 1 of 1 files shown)


tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_create.py
```diff
@@ -2158,11 +2163,10 @@ def test_failed_create_tx_refunds_intrinsic_new_account(
...
     # regular gas consumed, pinning gas_used to the floor.
-    floor = fork.transaction_data_floor_cost_calculator()(
-        data=bytes(init_code), contract_creation=True
-    )
-    expected_gas_used = max(expected_gas_used, floor)
-    expected_cumulative = max(expected_cumulative, floor)
+    data_floor_calc = fork.transaction_data_floor_cost_calculator()
+    floor = data_floor_calc(data=init_code, contract_creation=True)
+    assert expected_gas_used > floor
+    assert expected_cumulative > floor
```
jochem-brouwer, line 2169:
```suggestion
    assert expected_gas_used >= floor
    assert expected_cumulative >= floor
```

Might not be relevant for the current test vectors, but there could be situations where this is exactly the floor?

jochem-brouwer (commented):
To check: there is a problem where we get hit with the calldata floor cost https://eips.ethereum.org/EIPS/eip-7976? 

Why is this problem being "found" now and not previously? (Trying to understand why things break now)

One tiny suggestion which makes the assertions "more correct" (?)

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_create.py
marioevz: I think for this particular case we would want the expected gas used to be strictly higher than the floor so we can guarantee that the floor did not kick in

marioevz:
> To check: there is a problem where we get hit with the calldata floor cost https://eips.ethereum.org/EIPS/eip-7976? 
> 
> Why is this problem being "found" now and not previously? (Trying to understand why things break now)
> 
> One tiny suggestion which makes the assertions "more correct" (?)

It's because the behavior of 7976 was affected by 2780, and specifically #3120 was a fix to make this behavior match what the EIP currently describes.

The problem actually is that we didn't catch this in that PR, and afaik it's because we don't run Amsterdam in CI iirc, but I think we should change that soon. Cc @spencer-tb

jochem-brouwer (approved):
Thanks for the clarifications @marioevz 😄 👍 

LGTM! 🎉

gurukamath (approved):

PR #3141, opened 2026-07-10
feat(test-fill): collect per-test opcode counts in fill-stateful via `debug_traceBlockByNumber`

jochem-brouwer:
Made using Claude, Draft PR as I have not thoroughly checked this myself.

----

Ports the opcode-count collection feature from [gas-benchmarks](https://github.com/NethermindEth/gas-benchmarks) (its `eest_stateful_generator.py --trace-json` flag / `run.sh` `TRACE_BLOCKS` env var) to `fill-stateful`.

New flags (**disabled by default**):

- `--trace-opcodes` — after each block is built via `testing_buildBlockV1` and made canonical, trace it with `debug_traceBlockByNumber` using the unigram tracer (per-opcode execution counts) and, at session end, write a JSON file mapping each test to its summed opcode counts.
- `--trace-opcodes-output` — output path override (default: `<output-dir>/opcodes_tracing.json`).

The output file format is identical to gas-benchmarks' trace output:

```json
{
  "test_single_opcode.py__test_sload_erc20_generic[fork_Amsterdam-...]": {
    "PUSH1": 55674,
    "SLOAD": 11132
  }
}
```

Implementation notes:

- **Client compatibility**: geth, erigon and nethermind resolve the built-in `unigramTracer` tracer by name; reth does not, but runs the equivalent inline JS tracer source (its `js-tracer` feature is enabled in default builds). `DebugRPC.trace_block_opcode_counts` probes the name first, falls back to inline JS, and — for clients supporting neither (besu has no JS tracer engine) — warns once and continues filling without traces. The working tracer is cached per session, mirroring the existing `rewind_head` probe.
- **Phase scoping**: only execution-phase payloads count toward a test's totals; per-test setup blocks (funding, deploys) are excluded, matching gas-benchmarks, which derives counts from the `testing/` payloads only. Counts are summed across a test's execution blocks.
- **Name parity**: `OpcodeCount` canonicalizes `KECCAK256` to the `Opcodes` enum's `SHA3`; the writer maps it back to `KECCAK256` so the output matches what clients report and what gas-benchmarks files contain.
- Tracing failures are never fatal to the fill: trace errors and unparsable responses are logged and skipped.
- As a side effect, `benchmark_opcode_count` is now populated in stateful benchmarking fills, so `fixed_opcode_count` / `expected_opcode_count` verification works on the stateful path when tracing is enabled.

Usage:

```console
uv run fill-stateful --fork=Amsterdam \
  --rpc-endpoint http://localhost:8545 \
  --engine-endpoint http://localhost:8551 \
  --engine-jwt-secret-file ./jwt.hex \
  --trace-opcodes \
  --output fixtures/erc20_mixed \
  tests/benchmark/stateful/bloatnet/test_multi_opcode.py
# -> fixtures/erc20_mixed/opcodes_tracing.json
```

Includes unit tests for the trace-response parsing (per-tx shapes of all clients), the name→JS→warn-once fallback chain, key derivation from pytest node ids, the exact output file format, per-test merging, and `ClientBackend.evaluate` tracing off-by-default/on-when-enabled behavior.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

Crow image

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
@@ -61,6 +61,7 @@ from ..execute.rpc.hive import (
 )
 from ..shared.helpers import is_help_or_collectonly_mode
 from ..shared.live_client_flags import FEE_BUMP_MULTIPLIER
+from .opcode_tracing import write_opcode_trace_file
 
 # 1 billion ETH. Withdrawals are Gwei (u64-capped); much higher risks
 # overflow on some clients, this is plenty for any single fill session.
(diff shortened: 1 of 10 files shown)


jochem-brouwer:
This PR attempts the same as https://github.com/ethereum/execution-specs/pull/3124, will close here.

Issue #3106, opened 2026-07-06
Benchmarking Update to Devnet 6, 7 Spec


LouisTsai-Csie:
## Overview

This tracker manages benchmark refactoring tasks to add glamsterdam-devnet-6 support. Currently, benchmarks only work with bal-devnet-7. We're prioritizing glamsterdam-devnet-6; glamsterdam-devnet-7 is a future enhancement.

- Payload Generation: Rafael generates via Benchmarkoor (not EELS)
- EELS Branch: `forks/amsterdam`. Ensure new spec changes sync to `devnets/glamsterdam/7`
- Timeline: ~1 day for payload generation; Expect benchmarks start running on Friday.

## Scope

This includes refactoring benchmark cases for glamsterdam-devnet-6 and adding new ones. We also need to refactor existing the infrastructure to support this.

### Add Benchmark
- [x] account access benchmark + ether transfer benchmark
  - https://github.com/ethereum/execution-specs/pull/2947 -> This should refactor to align with EIP-2780, EIP-8037, EIP-8038.

### Benchmark Refactor
- [x] Refactor compute benchmark
  - Louis: https://github.com/ethereum/execution-specs/pull/3098
    - Mario: https://github.com/LouisTsai-Csie/execution-specs/pull/10
    - Spencer: https://github.com/LouisTsai-Csie/execution-specs/pull/11
- Follow up PRs:
  - PR #3135 , PR #3136 , not blocking initial payload generation
- [x] Refactor stateful benchmark, both repricing and non-repricing cases
  - For  non-repricing cases, we could ignore scenario that interacts with ERC20
  - `test_account_query.py` cc @LouisTsai-Csie 
  - `test_create2_access.py` cc @LouisTsai-Csie 
  - ~~`test_delegatecall_chain.py`~~
  - `test_extcodesize_bytecode_sizes.py` cc @LouisTsai-Csie 
  - `test_multi_opcode.py` cc @LouisTsai-Csie 
  - `test_single_opcode.py` cc @LouisTsai-Csie 
  - `test_transaction_types.py` cc @LouisTsai-Csie 
  - `test_transient_storage.py` cc @LouisTsai-Csie 
- [x] Review benchmark test wrapper
  - Re-consider `JumpLoopGenerator` and `ExtCallGenerator` under EIP-2780, EIP-8038. Does `CALL*` reflect its actual gas calculation -> Clarify in issue #3128 

### Framework Refactor
[...]

PR #3142, opened 2026-07-10
fix(fill-stateful): verify receipt status when filling stateful fixtures

jochem-brouwer:
### Description

The receipt-status verification added in #2527 (`BaseTest.validate_receipt_status`, driven by `expected_receipt_status` on benchmark tests) was wired into `make_fixture` and `make_hive_fixture`, but never into `make_stateful_fixture`. As a result, a stateful benchmark whose transactions silently OOG or revert — rolling back state and invalidating the benchmark — filled "successfully", even though `ClientBackend` already fetches the real receipts (including `status`) from the live client via `eth_getTransactionReceipt`.

This PR wires the check into the stateful fill path and adds tests at two levels:

**Fix**

- `make_stateful_fixture` now calls `validate_receipt_status` on every execution-phase block's receipts.
- Setup-phase blocks (`pre.fund_eoa` / `pre.deploy_contract` funding and deploy transactions) are exempt: `expected_receipt_status` describes the test's own transactions, and a test that legitimately expects reverts (`expected_receipt_status=0`) would otherwise falsely fail on its always-succeeding setup transactions.
- `TestingBuildBlock` gets `__test__ = False` so pytest does not try to collect it when imported into test modules.

**Unit tests** (`packages/testing/src/execution_testing/specs/tests/test_stateful_receipt_status.py`)

Drive `make_stateful_fixture` with a stubbed `ClientBackend` and a monkeypatched `generate_block_data` feeding crafted receipt statuses through the loop:

- status 0 with `expected_receipt_status=1` raises;
- status 1 with `expected_receipt_status=0` raises;
- one failed receipt among successes raises;
- matching statuses (0/0 and 1/1) fill normally;
- setup-phase blocks are exempt;
- unset `expected_receipt_status` skips validation.

All three raise-path tests fail without the wiring fix and pass with it.

**Runtime canaries** (`tests/benchmark/stateful/test_receipt_status.py`)

Filler self-tests that run in any real fill session (including fill-stateful against a live client, where they additionally prove statuses are read back correctly over RPC):

- `test_receipt_status_mismatch_fails_fill`: a reverting transaction with `expected_receipt_status=1`, marked `xfail(strict=True)` — the session passes only when the filler raises the mismatch; if verification ever silently stops working, the canary XPASSes and fails the session.
- `test_receipt_status_expected_revert_fills`: the control — the same reverting transaction with `expected_receipt_status=0` must fill cleanly.

Verified end-to-end with `fill --fork Prague --gas-benchmark-values 1`: normally 2 passed + 2 xfailed; with `validate_receipt_status` temporarily neutered, the canaries flip to FAILED (strict XPASS) as designed.

### Related Issues or PRs

Part of #3106 ("Add a test for receipt status verification works for stateful filler"). Follow-up to #2527.

### Checklist

### Cute Animal Picture

Grutto

(this is a Grutto, a bird seen in The Netherlands, but it is getting less common to see it here due to changes)


tests/benchmark/stateful/test_receipt_status.py
```diff
@@ -0,0 +1,80 @@
...
+
+   These tests exist to prove, during a real fill session, that the
+   filler verifies receipt statuses against ``expected_receipt_status``.
+   The mismatch canary is marked ``xfail(strict=True)``: the session
+   passes only when the filler raises the receipt-status mismatch. If
+   verification ever silently stops working, the canary XPASSes and
+   fails the session. They run in every fill mode; against a live
+   client (fill-stateful) they additionally prove statuses are read
+   back correctly via ``eth_getTransactionReceipt``.
+"""
```
LouisTsai-Csie: We already have e2e coverage in `test_stateful_receipt_status.py`, so this test may be redundant. Also, framework tests under `tests/benchmark/stateful` could be confusing, so better to keep them separate.

I will remove this file in a follow commit, but let me know if we should still keep it!

LouisTsai-Csie (approved):
Thanks for the update, i push a commit for some suggestion. Otherwise, we could merge this PR.

PR #3143, opened 2026-07-10
docs(test-consume): fix and augment enginex docs

danceratopz:
### Description

Fixes and extends the `consume enginex` documentation in `docs/running_tests/running.md`, which still described the Engine API flow as it was before #3093.

- Correct the claim that enginex skips the forkchoice update (FCU) to genesis and the FCU after valid payloads. Since #3093 both `consume engine` and `consume enginex` send the identical Engine API sequence; the only step enginex skips is the `eth_getBlockByNumber(0)` genesis block hash check, which now runs once per client, instead of once per test, as genesis is immutable.
- Expand the `consume engine` and `consume enginex` step lists to include the initial FCU to genesis, the genesis block hash check and the FCU sent after each valid payload.
- Document enginex's implicit chain reorg coverage in a new section: the FCU at the start of each test rolls the client's head back from the previous test's chain tip to genesis, so the next payload arrives as a sibling of a block the client already considers canonical, and the FCU that follows it re-canonicalizes a new branch. Every test after the first in a pre-allocation group therefore exercises the client's reorg path, which `consume engine` never reaches. The section notes that this coverage is a side-effect of client reuse and not something the fixtures describe, and that an enginex-only failure points at the client's reorg, head state rollback or block caching logic.
- Add `chain reorgs (implicit)` to the components tested by `consume enginex` in the top-level comparison table, and add a `Chain reorgs` row to the "Engine vs EngineX" table.

### Related Issues or PRs

Fixes docs following:
- #3093.

### Checklist

### Cute Animal Picture

HTTP 200 OK dog

### DIFF
--- a/docs/running_tests/running.md
+++ b/docs/running_tests/running.md
@@ -59,10 +59,12 @@ The `consume engine` command:
 
 1. **Initializes the execution client** with genesis state.
 2. **Connects via Engine API** (port 8551), primitively mocking a consensus client.
-3. **Sends a forkchoice update** to establish the chain head.
-4. **Submits payloads** using `engine_newPayload` calls.
-5. **Validates responses** against expected results.
-6. **Tests error conditions** and exception handling.
+3. **Sends a forkchoice update** to the genesis block to establish the chain head.
+4. **Verifies the client's genesis block hash** via `eth_getBlockByNumber(0)`.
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
Fuego! :D

PR #3144, opened 2026-07-10
fix(spec-specs, tests): apply calldata floor to block-level regular gas

nerolation:
implementing the fix from https://github.com/ethereum/EIPs/pull/11908


spencer-tb (approved):
LGTM!

PR #3145, opened 2026-07-10
feat(test-consume): set HIVE_EXPECT_DEEP_REORGS in the multi-test client environment

danceratopz:
Multi-test client workloads (currently `consume enginex`) reuse one client across a pre-alloc group's tests and rewind the head to genesis between tests, which trips client-specific reorg depth limits — e.g. geth refuses rewinds deeper than 32 blocks since ethereum/go-ethereum#34767, permanently poisoning the shared client:

```
JSONRPCError(code=-38006, message=Too deep reorg, data={'err': 'reorg depth 37 exceeds limit 32'})
```

Setting `HIVE_EXPECT_DEEP_REORGS=1` in the multi-test client environment gives hive client wrapper scripts an intent-named, simulator-agnostic way to detect such workloads and raise those limits only where needed, e.g. `export GETH_ENGINE_MAXREORGDEPTH=512` in ethereum/hive#1572.

416 Range Not Satisfiable


spencer-tb (approved):
Muy buen!

PR #3146, opened 2026-07-10
fix(test-consume): accept bad-block-cache errors for resubmitted blocks

danceratopz:
### Description

In enginex mode a client is reused across all tests of a pre-allocation group, so its bad-block cache persists between tests. When two tests in a group contain an identical invalid block, the client validates the first submission for real and returns the specific validation error, but answers the resubmission from its cache with a generic one: reth returns "links to previously rejected block", which maps to no known exception and fails the test under strict exception matching, even though the client correctly returns `INVALID` both times. This causes 18 systematic enginex-only failures on `tests@v20.0.0` and is the entire `consume-engine` vs `consume-enginex` diff for reth.

Fix: A new `BlockRejectionTracker` records the first validation error each client returns per invalid block hash. When a rejection does not match the expected exception, it is verified against the client's first (real) rejection of the same block: accepted and logged if that one matched, failed as before otherwise. This is sound because an identical block hash implies an identical block on an identical parent chain, so the validation outcome is deterministic. It is also xdist-safe: entries are scoped by client id, mirroring the per-process bad-block cache they model, and `--dist loadgroup` pins each group's client to one worker. Classic `consume engine` (fresh client per test) is unaffected.

#### Reproducing the Issue

Reproduction against reth via hive `--dev`: without this PR 18 of 342 tests fail; with it all pass, and the acceptance path logs "Accepting mismatched validation error" exactly 18 times.

```console
./hive --dev --client reth --client-file ../execution-specs/.github/configs/hive/master.yaml --docker.buildoutput --sim.loglevel 5 --dev.addr 127.0.0.1:3020
```

```console
export HIVE_SIMULATOR=http://127.0.0.1:3020
uv run consume enginex --input=tests@v20.0.0 -n 4 -v --sim.limit ".*(test_invalid_excess_blob_gas_(above_target_)?change|test_invalid_zero_excess_blob_gas_in_header).*"
```

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

208 Already Reported

### DIFF
--- a/docs/running_tests/running.md
+++ b/docs/running_tests/running.md
@@ -107,7 +107,7 @@ Every test after the first in a pre-allocation group consequently exercises the 
 
 ### Bad-Block Cache Handling
 
-Clients cache the blocks they reject. Because a client is reused across a pre-allocation group, its bad-block cache persists between tests: if two tests in a group contain an identical invalid block, the client validates the first submission for real and returns the specific validation error, but answers the resubmission from its cache with a generic error (e.g. reth's "links to previously rejected block") that maps to no known exception.
+Clients typically cache the blocks they reject. Because a client is reused across a pre-allocation group, its bad-block cache persists between tests: if two tests in a group contain an identical invalid block, the client validates the first submission for real and returns the specific validation error, but may answer the resubmission from its cache with a generic error (e.g. geth's and reth's "links to previously rejected block" or Nethermind's "is known to be a part of an invalid chain") that maps to no known exception.
 
 The simulator therefore remembers the first validation error each client returns per invalid block. When a rejection does not match the test's expected exception, it is verified against the client's first rejection of the same block: it is accepted and logged ("Accepting mismatched validation error") if that first rejection matched the expected exception, and fails the test as before otherwise. This is sound because an identical block hash implies an identical block built on an identical parent chain, so the real validation outcome is deterministic. `consume engine` starts a fresh client per test and is unaffected.
 
--- a/docs/running_tests/running.md
+++ b/docs/running_tests/running.md
@@ -105,6 +105,12 @@ Every test after the first in a pre-allocation group consequently exercises the 
 
     It does mean, however, that a test which fails under `consume enginex` but passes under `consume engine` is more likely to indicate a bug in the client's reorg, head state rollback or block caching logic than in its EVM or block validation logic.
 
+### Bad-Block Cache Handling
+
+Clients cache the blocks they reject. Because a client is reused across a pre-allocation group, its bad-block cache persists between tests: if two tests in a group contain an identical invalid block, the client validates the first submission for real and returns the specific validation error, but answers the resubmission from its cache with a generic error (e.g. reth's "links to previously rejected block") that maps to no known exception.
+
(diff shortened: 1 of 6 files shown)


danceratopz:
image

danceratopz:
Draft: Requires docs, will wait for this unrelated enginex docs update to land:
- https://github.com/ethereum/execution-specs/pull/3143

danceratopz:
Added docs, the bad block cache addition will also fix 18 Nimbus `enginex`-specific fails (alongside reth's).

Update: It will also fix tests failing due to the same behavior in geth and Nethermind.

spencer-tb (approved):
LGTM! I don't see this as an issue. Some of our tests will have the same bad block (and should afaik).

Tagging @marioevz here, but will merge nonetheless.

danceratopz:
> LGTM! I don't see this as an issue. Some of our tests will have the same bad block (and should afaik).
> 
> Tagging @marioevz here, but will merge nonetheless.

Thanks @spencer-tb! Yeah, I also wondered if we should avoid multiple tests that generate blocks with the same (invalid) hash. Here's what Claude said:

> 
> Short answer: I'd keep the tracker. The duplication is real but it's emergent from parameter grids, not authored redundancy, and pruning it would be brittle while fixing only some instances of the problem. Here's the reasoning, grounded in the two colliding tests:
> 
> Why the duplicates exist. The two tests express different intents over overlapping grids (test_excess_blob_gas.py:406-437 and :651-696):
> 
> - test_invalid_zero_excess_blob_gas_in_header tests an absolute boundary: header excessBlobGas = 0 when the correct value is nonzero.
> - test_invalid_excess_blob_gas_change sweeps relative offsets: header = correct + delta·GAS_PER_BLOB for every nonzero delta in [-target, +target].
> 
> For the specific combo where correct_excess_blob_gas happens to equal 2·GAS_PER_BLOB (Osaka, parent_blobs_0), the delta -2 case lands exactly on zero — byte-identical to the zero-test's block. At other grid points the same two parametrizations produce entirely distinct blocks. So each parametrization does cover something; they only coincide at fork-dependent boundary values, and the collision set shifts whenever blob schedule constants change (Cancun→Prague→Osaka targets, BPO forks). Excluding the collisions would mean fork-aware cross-function pruning logic that computes "does my delta land on another test's absolute value" — brittle, and it obscures each test's intent. There's precedent for within-function pruning (both tests raise "invalid test case" when header == correct), but cross-function dedup is a different beast.
> 
> Why the resubmission isn't testing nothing. In enginex mode the duplicate submission is currently the only thing exercising the client's bad-block cache: a client that accepted the resubmitted invalid block would still fail the status check. That's squarely within enginex's stated purpose (implicit reorg/caching coverage) — the right move is for the simulator to grade a cached rejection correctly, which is exactly what #3146 does, not to remove the stimulus.
> 
> Why dedup wouldn't fully solve it anyway. Removing today's collisions doesn't remove the class: any future grid can collide again, and Nethermind issue #3 showed same-hash cases that aren't even equal submissions (identical block hash, different executionRequests param). Fixtures are also consumed by many tools; in consume engine, statetests, etc., the duplicates are harmless redundancy, so reshaping the test source around one simulator mode's client-cache artifact inverts the layering.
> 
> The scale also argues against it: Family A was 18 failures across the full reth run — a small tail, so the fill/runtime savings from dedup would be marginal.
> 
> That said, your instinct points at something worth having: visibility. The tracker's "Accepting mismatched validation error" log line now enumerates every collision in an enginex run for free — if after a few full runs the acceptance count looks large for some grid, that's the signal to revisit that test's parametrization specifically. A fill-time report of identical block hashes within a pre-alloc group would be the more principled tool than test removal, if you ever want to audit the grids systematically.as well.

PR #3156, opened 2026-07-13
feat(tests): add type-0 transaction RLP validity tests

chfast:
### Description

Port the core malformation classes of the legacy TransactionTests suites (ttWrongRLP, ttNonce, ttValue, ttRSValue, ttVValue, ttAddress), which were never converted because the ported-static pipeline only handles state-test fillers and the raw malformed bytes cannot round-trip through a structured transaction model.

A local RLP encoder builds each corruption deliberately, since a correct encoder cannot emit non-canonical forms: per-field leading zeros, 33-byte field overflows, 19 and 21 byte addresses, fields encoded as lists, structural corruptions of the outer list (truncation, trailing bytes, wrong element counts, header size mismatches, size with leading zeros), and well-encoded but invalid signature values. A valid re-encoded control case anchors the encoder to the framework's byte-exact output.

The transaction_test fixture format records the declared exception without consulting the transition tool, so all 36 (35 invalid) cases were verified externally by feeding the generated fixture bytes through EELS decode_transaction, recover_sender and validate_transaction at Frontier, London and Cancun: every invalid vector is rejected and the control is accepted with the matching sender. Notably the gas limit and gas price are unbounded scalars in the spec, so their oversized encodings are valid at the transaction level; the overflow cases cover the 256-bit bounded fields (nonce, value, r, s) only.

### Related Issues or PRs

Found by evmone-fuzz.

### Checklist

### DIFF
--- a/tests/frontier/validation/test_transaction_rlp.py
+++ b/tests/frontier/validation/test_transaction_rlp.py
@@ -194,11 +194,16 @@ def test_field_overflow(
 ) -> None:
     """
     Encode one integer field as a 33-byte value (2**256), exceeding the
-    field's width (64 bits for the nonce, 256 bits for the others).
+    256-bit width of the field. The spec decodes the nonce as a 256-bit
+    scalar as well; its 64-bit bound is a validation rule (EIP-2681),
+    not a decoding one, and clients that store the nonce in 64 bits
+    reject the oversized encoding all the same.
 
     The gas limit and gas price are unbounded scalars in the spec, so
     oversized values there are not a decoding error and are rejected
-    only in block context.
+    only in block context. The signature v is also a bounded 256-bit
+    field, but has no field-specific decoding exception, so its
+    oversized encoding is not covered here.
     """
     fields = signed_tx_fields(pre, fork)
     corrupted = rlp_bytes(int_payload(2**256))
--- a/tests/frontier/validation/test_transaction_rlp.py
+++ b/tests/frontier/validation/test_transaction_rlp.py
@@ -253,6 +253,8 @@ def test_to_address_size(
         "RLPElementIsListWhenItShouldntBeCopier.json",
         f"{LEGACY_TX_TESTS}/ttWrongRLP/"
         "RLPElementIsListWhenItShouldntBe2Copier.json",
+        f"{LEGACY_TX_TESTS}/ttWrongRLP/TRANSCT_rvalue_GivenAsListCopier.json",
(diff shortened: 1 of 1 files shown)


danceratopz (commented):
Thanks @chfast! I think we have a bit of work to do regarding awareness of and ease of consuming the `transaction_test` format. Could you test these against `evmone`? If so, mind sharing how? 

The other way to run `transaction_test` is via `hive execute`. I think we have a bit of work to do if that's the goal. No need to dive into this now, I'm not sure that should be the goal for these (invalid) tests, that should get rejected at the JSON RPC layer.

danceratopz (commented):
Added some suggestions here https://github.com/chfast/execution-specs/pull/2. Thanks!

chfast:
This is discussion starter for transaction decoding testing. I don't consume the transaction tests in evmone, but I plan to do so soon as I'm adding full transaction decoding capability. But this seems to be working with `evmone t8n`.

The alternative to specific `transaction_test` format is to reuse `state_test` format. However, then you can only have `txbytes` representation of the transaction (no transaction decomposed to a JSON object). For me this is fine as well as I plan to add exactly support for `txbytes` in evmone. I believe some Clients should be already compatible.

marioevz:
@danceratopz  friendly bump to get this merged 😄 
I think it will be useful to test FOCIL if we add the `pytest.mark.inclusion_test` marker.

danceratopz:
> @danceratopz friendly bump to get this merged 😄 I think it will be useful to test FOCIL if we add the `pytest.mark.inclusion_test` marker.

@chfast let's get this merged. Could you please:

1. TAL at https://github.com/chfast/execution-specs/pull/2 and merge if you agree; I think it's worthwhile.
2. Rebase on forks/amsterdam to get `@pytest.mark.inclusion_test`, cf #3337.
3. Apply `@pytest.mark.inclusion_test` to these tests via a module level mark:
    ```python
       pytestmark = [
          pytest.mark.valid_from("Frontier"),
          pytest.mark.inclusion_test,
      ]
    ```
    
In parallel I'd like to add an infra PR to ensure that `transaction_test` get ran against EELS.

danceratopz (approved):
Thanks @chfast! This LGTM! Guru and I just got this quick PR merged, which ensures that the value-transferring tests in this PR generate correct fixtures for Amsterdam via the framework:
- https://github.com/ethereum/execution-specs/pull/3491

PR #3159, opened 2026-07-13
refactor(spec-specs): organize gas logic in amsterdam EIP-8037/8038/2780

gurukamath:
### Description

Two behavior-preserving refactors of the Amsterdam two-dimensional
gas implementation:

- **Bundle frame gas into `GasMeter`** — gather the flat gas fields on `Evm` into a `GasMeter` dataclass in `vm/gas.py`, replacing scattered inline arithmetic with named operations for the state-gas rollback/commit scheme, child-frame grants, the EIP-8037 execution gas split, and transaction settlement. t8n tracer protocols gain gas-layout accessors so both `Evm` shapes trace through one path.
- **Name the gas stages in opcodes** — `CALL*`/`CREATE*` wrappers price in labeled stages (`GAS (STATE-INDEPENDENT)`, `STATE ACCESS (STATE-DEPENDENT GAS)`, `STATE GAS`, `CHILD GRANT`, `OPERATION`), also applied to `sstore` and `selfdestruct`; `generic_call` /`generic_create` reduce to the child-frame lifecycle (PREFLIGHT,
  DESTINATION ACCESS, DISPATCH, OUTCOME). Currently, it looks like an un-organised blob of operations.

Audited against EIP-8037/8038/2780 text for functional and naming consistency.

## Verification

No behavior change: the relative order of every charge, check, and trace event is preserved. Verified with the EIP-8037/2780/7778/8038
Ran `uv run hasher compare` on the full Amsterdam fixtures to make sure the fixtures did not change.

### Checklist

### Cute Animal Picture

### DIFF
--- a/.claude/commands/implement-eip.md
+++ b/.claude/commands/implement-eip.md
@@ -28,9 +28,21 @@ Each fork lives at `src/ethereum/forks/<fork_name>/`. Explore the latest fork di
 ## Adding a New Opcode
 
 1. Add to `Ops` enum in `vm/instructions/__init__.py` with hex value
-2. Implement function in appropriate `vm/instructions/<category>.py` — follows pattern: STACK → GAS (`charge_gas`) → OPERATION → PROGRAM COUNTER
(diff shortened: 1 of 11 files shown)


marioevz (approved):
No objections from my side, it's a great change and looks much more readable. Thanks!

PR #3170, opened 2026-07-14
feat(tests): cover invalid-BAL content canonicality (no-op / missing / phantom-read entries)

edg-l:
### Description

Adds invalid Block-Level Access List (EIP-7928) tests for canonicality violations the current suite doesn't cover. The existing invalid-BAL tests inject values that differ from execution state, so entries that match the start-of-tx value, omitted writes, and phantom reads on destroyed accounts all pass today.

New cases, all expecting `INVALID_BLOCK_ACCESS_LIST`:

- no-op change entries (post value == start-of-tx value) for storage, balance, nonce, code
- a storage write omitted from the BAL
- a CREATE's `code_changes` omitted
- a slot change recorded only at a later index than the tx that wrote it
- a phantom `storage_read` on an account created and destroyed in the same tx

Also adds a `remove_slot_change` modifier and lists the new cases in `test_cases.md`. All cases fill against the reference spec and the full amsterdam eip7928 suite fills with no regressions. 

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

cute animal


spencer-tb (approved):
LGTM! Thanks :)

Some of these tests are nice coverage!

PR #3171, opened 2026-07-15
chore(ci): only install a native build toolchain for PyPy jobs

danceratopz:
### Description

TLDR: #2374 replaced `coincurve` with `spec256k1` and removed the last required package (outside the scope of `ethereum-spec-sync`) that needed creating a wheel and C compilation for CPython runs. This gives us an opportunity to simplify our CI dependencies and slightly reduce CI times.

More specifically, after #2374 every locked dependency installs as a wheel on x86-64 Linux for CPython 3.13 and 3.14; the only remaining sdist builds are pure Python (the workspace packages, `actionlint-py`, and `pyyaml-ft`). A native toolchain is only genuinely needed by the PyPy jobs, where `libcst` (Rust) and `pycryptodome`, `pyyaml`, `regex`, and `markupsafe` (C) publish no wheels.

- Drop `setup-env` from all CPython jobs (`fill`, `json-loader`, `spec-tools`, `test-tests`, both hive workflows, and the `build-fixtures` action) and drop the `static` job's inline apt step; slim the action to `build-essential` plus Rust for the two PyPy jobs, removing the `coincurve` leftovers `libsecp256k1-dev` and `pkg-config`.
- Move `ethereum-execution[optimized]` from the `dev` group to a new opt-in `optimized` dependency group: `ethash` has no CPython 3.14 wheels (cf #3173) and nothing in CI imports `ethereum_optimized`. `just typecheck` installs the group for mypy; the sync tool is now invoked as:
  ```
  uv run --group optimized ethereum-spec-sync
  ```
- Make `build-evmone` install `build-essential` itself on cache miss (it previously came implicitly from the calling job's `setup-env`).

The win is mainly simplification: CPython jobs are now provably wheels-only, and the remaining `setup-env` documents exactly why it exists. Time savings are modest: the step took ~17-18s per job across 8 jobs per `test.yaml` run, and the `static` job's step sat on the critical path since all other jobs gate on it via `needs: static`.

Verified by running cold-cache `uv sync --locked` for every CI environment shape in containers with no C compiler or Rust installed. The tip commit sets `UV_NO_CACHE=1` so this PR's CI also installs everything with a cold cache; it will be reverted before merge. Since the `setup-uv` cache key is the exact `uv.lock` hash (no prefix fallback), any future lock change re-exercises wheel coverage automatically.

### Related Issues or PRs

Enabled by:
- #2374.

### Checklist

### Cute Animal Picture

204 No Content: nothing left to install

### DIFF
--- a/.github/actions/setup-env-pypy/action.yaml
+++ b/.github/actions/setup-env-pypy/action.yaml
@@ -1,4 +1,4 @@
-name: Setup Environment
+name: Setup PyPy Build Toolchain
 description: |
   Install the native toolchain needed to build sdists for PyPy jobs.
   Several dependencies publish no PyPy wheels and are built from source:
--- a/.github/actions/setup-env-pypy/action.yaml
+++ b/.github/actions/setup-env-pypy/action.yaml
@@ -10,5 +10,6 @@ runs:
     - name: Install C and Rust toolchains
       shell: bash
       run: |
-        sudo DEBIAN_FRONTEND=noninteractive apt-get install --yes --force-yes build-essential rustup
+        sudo apt-get -q update
+        sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install build-essential rustup
         rustup update --no-self-update 1.89.0 && rustup default 1.89.0
--- a/Justfile
+++ b/Justfile
@@ -64,8 +64,7 @@ deadcode:
 format-check *args:
     uv run ruff format --check "$@"
 
-# Run type checking with mypy (the optimized group provides the typed
-# rust-pyspec-glue and ethash packages imported by ethereum_optimized)
+# Run type checking with mypy (installs the optimized dependency group)
 [group('static analysis')]
 typecheck *args:
     uv run --group optimized mypy "$@"
--- a/Justfile
+++ b/Justfile
@@ -64,10 +64,11 @@ deadcode:
 format-check *args:
     uv run ruff format --check "$@"
 
-# Run type checking with mypy
+# Run type checking with mypy (the optimized group provides the typed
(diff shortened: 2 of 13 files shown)


Justfile
```diff
@@ -64,10 +64,11 @@ deadcode:
 format-check *args:
     uv run ruff format --check "$@"
 
-# Run type checking with mypy
+# Run type checking with mypy (the optimized group provides the typed
+# rust-pyspec-glue and ethash packages imported by ethereum_optimized)
```
spencer-tb: `just --list` only picks up the last comment line, so the recipe doc renders as `# rust-pyspec-glue and ethash packages imported by ethereum_optimized)`.

```suggestion
# Run type checking with mypy (installs the optimized dependency group)
```

.github/actions/setup-env-pypy/action.yaml
```diff
@@ -0,0 +1,14 @@
...
+  Several dependencies publish no PyPy wheels and are built from source:
+  libcst needs Rust; pycryptodome, PyYAML, regex, and MarkupSafe need a C
+  compiler. CPython jobs install wheels only and must not use this action.
+runs:
+  using: "composite"
+  steps:
+    - name: Install C and Rust toolchains
+      shell: bash
+      run: |
+        sudo DEBIAN_FRONTEND=noninteractive apt-get install --yes --force-yes build-essential rustup
```
spencer-tb:
```suggestion
        sudo apt-get -q update
        sudo DEBIAN_FRONTEND=noninteractive apt-get -qy install build-essential rustup
```
`--force-yes` is deprecated in apt-get; updating first matches `build-evmone`.

spencer-tb (approved):

PR #3176, opened 2026-07-15
fix(tests,test-fill): isolate the scenarios BLOCKHASH program pre-alloc group

danceratopz:
### Description

Follow-up to #3122: The v20.0.1 release fill failed the Engine X execution drift check with a single mismatch, `test_scenarios[fork_Paris-blockchain_test_engine_x-test_program_program_BLOCKHASH-debug]` ([failing job](https://github.com/ethereum/execution-specs/actions/runs/29406408314/job/87322887242)). This PR fixes the root cause; pre-alloc group packing itself is safe, and the check caught exactly the kind of issue it was built for.

**Why this one test drifts.** The scenarios `BLOCKHASH` program feeds `BLOCKHASH(0)` (the genesis hash) to the scenario suite's gas-hash contract, which deliberately burns gas proportional to its input so that values unknowable at fill time still leave a trace in `gasUsed`. The genesis hash depends on every account in the genesis, so merging any other test's accounts into the shared pre-allocation changes this test's `gasUsed` and `receiptsRoot`. No grouping scheme can be transparent for a test that observes the genesis hash itself; it needs its own genesis. The drift even predates packing: the fine-grained phase 1 groups already shared this test's genesis, and the check introduced in #3122 merely exposed it.

**The fix.** Mark just the `BLOCKHASH` program parametrization with `@pytest.mark.pre_alloc_group("separate")`, so its Engine X fixture is filled against the test's own genesis on every fork; its execution then matches its `blockchain_test_engine` sibling exactly and the drift check passes. The suite's other gas-hash programs (`GASPRICE`, `TIMESTAMP`, `DIFFICULTY`, `BASEFEE`, `BLOBBASEFEE`) feed environment-fixed values and were verified to fill drift-free when packed into a single group.

Using the marker (this is its first real user) surfaced two more issues, fixed here:

- The filler used the literal marker argument as the group salt, so every test marked `"separate"` would have silently shared one named group instead of getting its own genesis. It now salts `"separate"` (and the bare marker) with the test's node id, matching the marker's documented semantics. The regression cases were committed first and fail without the fix.
- The drift error message claimed an account leak was the only possible cause; it now also names genesis-hash observation and points at the `"separate"` marker as the remedy.

Fixture impact: `blockchain_tests` and `blockchain_tests_engine` fixtures are hash-identical before and after this change (verified with `hasher compare`); only Engine X fixtures re-key, since the affected pre-alloc groups' contents change.

### Related Issues or PRs

Follow-up to #3122.

### Checklist

### Cute Animal Picture

HTTP 409 Conflict

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -1657,10 +1657,15 @@ def base_test_parametrizer(cls: Type[BaseTest]) -> Any:
                     "pre_alloc_group"
                 ):
                     # Get the group name/salt from marker args
-                    if pre_alloc_group_marker.args:
+                    if (
+                        pre_alloc_group_marker.args
+                        and pre_alloc_group_marker.args[0] != "separate"
+                    ):
                         group_salt = str(pre_alloc_group_marker.args[0])
                     else:
-                        # We got the marker but unspecified, pass test name
+                        # "separate" (or a bare marker): salt with the
+                        # test's node id so the test gets its own genesis
+                        # instead of a group named literally "separate".
                         group_salt = _strip_xdist_group_suffix(
                             request.node.nodeid
                         )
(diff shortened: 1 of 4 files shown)


spencer-tb (approved):

spencer-tb:
LGTM! Lets get the new release out 🚀 :)

PR #3085, opened 2026-07-02
chore(test-consume): resolve fixture release tag for consume cache

spencer-tb:
## 🗒️ Description

Fix `consume cache` resolution for the fixture release scheme from ethereum/execution-specs#2888.

Release matching is now tag-namespace only (`tests@vX.Y.Z`, `tests-<feature>@vX.Y.Z`), so pre-tests-tag releases (`stable`/`develop`) no longer resolve, and `@latest` resolves the highest version rather than the most recently published.

### Usage

All release inputs below work with `consume cache` (and any consume subcommand's `--input`):

```bash
# mainnet (tests) releases
uv run consume cache --input=latest
uv run consume cache --input=v20.0.0
uv run consume cache --input=tests@latest
uv run consume cache --input=tests@v20.0.0

# feature releases, with or without the tests- tag prefix
uv run consume cache --input=bal@v7.3.2
uv run consume cache --input=tests-bal@v7.3.2
uv run consume cache --input=glamsterdam-devnet@latest
uv run consume cache --input=benchmark@latest
uv run consume cache --input=zkevm@latest
```

## 🔗 Related Issues or PRs

Follow up to ethereum/execution-specs#2888

## ✅ Checklist

- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) and [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/):
    ```console
    just static
    ```
- [x] All: PR title have the form `<type>(<area>):`, where `<type>` and `<area>` come from an approrpriate `C-<type>`, respectively `A-<area>`, label. The title should match the a target squash commit message.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [x] All: Set appropriate labels for the changes (only maintainers can apply labels).

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3182, opened 2026-07-15
fix(consume): make release resolution robust to GitHub API rate limits

danceratopz:
### Description

`consume` commands that resolve fixtures from a release spec crashed whenever the release-information cache was older than 4 hours and the GitHub API request failed:

```console
❯ uv run consume cache --input=tests@v20.0.1
...
INTERNALERROR>   File ".../plugins/consume/releases.py", line 250, in download_release_information
INTERNALERROR>     response.raise_for_status()
INTERNALERROR> requests.exceptions.HTTPError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/ethereum/execution-spec-tests/releases?per_page=100
```

The root cause: `get_release_information()` deleted the stale cache file *before* attempting the refresh. A failed download (the unauthenticated GitHub API allows 60 requests/hour per IP, shared by every tool on the machine) crashed the run and left no cache behind, so every subsequent run hit the API cold and the failure became self-perpetuating.

This PR makes release resolution robust and, for pinned versions, offline after the first lookup:

- Pinned versions (e.g. `tests@v20.0.1`) resolve from the cache regardless of its age and without any API requests: Release tags are immutable, so a cached entry cannot become outdated.
- A stale cache is kept as a fallback when the refresh fails (with a logged warning) instead of being deleted upfront; only lookups with no usable cache still hard-fail.
- A corrupt cache file (e.g. a partially-written download) is re-downloaded instead of wedging every run until manually deleted; the cache is now written atomically via a temporary file.
- API requests are authenticated with `GITHUB_TOKEN` when set: 5,000 requests/hour instead of 60.
- `get_release_page_url()` now receives the release spec instead of the asset URL, so the release-page lookup shares the pinned-version fast path instead of triggering a second resolution.

The `per_page=100` pagination introduced in #3085 was initially suspected but is not the cause: the request count per refresh is unchanged (still at most 2 pages per repo), and larger pages mean fewer, not more, requests.

The first commit adds the tests on their own; the four that cover the new behavior fail without the fix.

### Related Issues or PRs

Related: #3085.

### Checklist

### Cute Animal Picture

403 Forbidden

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/consume.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/consume.py
@@ -43,10 +43,9 @@ from execution_testing.tools.utility.versioning import (
 
 from .releases import (
     ReleaseTag,
-    get_release_page_url,
-    get_release_url,
     is_release_url,
     is_url,
+    resolve_release,
 )
 
 CACHED_DOWNLOADS_DIRECTORY = (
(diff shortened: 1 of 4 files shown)


spencer-tb (approved):
LGTM! Added a small addition

PR #3189, opened 2026-07-18
fix(test-specs): use integer division for genesis base fee

cuiweixie:
### Description

When deriving the genesis `base_fee_per_gas` from a state test's block#1 env (`* 8 / 7`), the previous code used float division (`/` → `float`), then cast back with `int(...)`.

IEEE-754 `float64` only has a **53-bit mantissa**, so integers larger than `2^53` are not represented exactly. For large `base_fee_per_gas` values this can silently round before the integer conversion, producing an incorrect genesis base fee.

This change switches to pure integer arithmetic (`* 8 // 7`) so the scaling stays exact for arbitrary-size integers.

### Related Issues or PRs
N/A.

### Checklist

### Cute Animal Picture

integer beaver


marioevz (approved):
Awesome, nice find, thanks!

PR #3191, opened 2026-07-18
chore(ci): deduplicate temp-dir creation across Justfile recipes

### LINKED ISSUE #3112: Remove repetitive, hardcoded temporary directory creation used across `Justfile` recipes
Initially suggested by @SamWilsn here https://github.com/ethereum/execution-specs/pull/3096#discussion_r3521097171, but apparently `recipe_name()` is dynamic, so it can't move the `mkdir` into a shared recipe.

Here's an idea from an alternative:
```just
[private]
_tmp name:
    @mkdir -p "{{ output_dir }}/{{ name }}/tmp"

json-loader *args: (_tmp "json-loader")
    uv run fill ...
```

If we bumped min version from 1.43 to 1.53 (which I'd **prefer to avoid**), I think we could write:
```just
json-loader *args: (_tmp recipe_name())
```

_Originally posted by @danceratopz in https://github.com/ethereum/execution-specs/pull/3096#discussion_r3527536541_

knQzx:
### Description

ten recipes each repeated an inline mkdir for their `--basetemp` scratch dir (and a logs dir for the fill/bench ones). replaced with two private recipes, `_tmp` and `_tmp-logs`, that recipes depend on, e.g. `fill *args: (_tmp-logs "fill")`

passing the name as a dependency argument avoids `recipe_name()` and the just >=1.53 bump it would need

### Related Issues or PRs

fixes #3112

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -21,13 +21,14 @@ latest_fork := "Amsterdam"
 # Use the faster sys.monitoring coverage core (default on 3.14, opt-in below).
 export COVERAGE_CORE := "sysmon"
 
-# Create the scratch directory a recipe passes to pytest's --basetemp. Recipes
-# depend on this instead of repeating the mkdir, e.g. `fill *args: (_tmp "fill")`.
+# --- Helpers ---
+
+# Create a recipe's --basetemp scratch directory
 [private]
 _tmp name:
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
Thanks!

Issue #3155, opened 2026-07-13
execution-specs (amsterdam): witness MPT decode accepts over-length secured-trie paths


pirapira:
**Repo/commit:** [`ethereum/execution-specs@f53b22e`](https://github.com/ethereum/execution-specs/tree/f53b22e1cde02b5ed5054c42b406c45b774104d6) (tests-zkevm@v0.6.1). Verified by source read.

## Summary

The stateless witness MPT decoder accepts trie nodes whose hex-prefix path is arbitrarily long. In a *secured* trie (state and storage) every key is a 32-byte keccak hash = exactly 64 nibbles, so no conformant node can have a path longer than 64 nibbles — but the decoder never checks this. An adversarial witness can therefore carry a leaf/extension node with, say, a 2048-nibble path, and the spec will decode and process it. Any exact-match reimplementation would need to implement this behavior but not reasonable for in particular a fixed-buffer verifier. The current behavior is a strict widening of the accepted-block set beyond what any honest producer emits.

## The gap

[`compact_to_nibbles`](https://github.com/ethereum/execution-specs/blob/f53b22e1cde02b5ed5054c42b406c45b774104d6/src/ethereum/forks/amsterdam/incremental_mpt.py#L859) decodes a hex-prefix path with no length bound (2 nibbles per input byte, any length, fine itself).

Its sole caller [`_decode_witness_node`](https://github.com/ethereum/execution-specs/blob/f53b22e1cde02b5ed5054c42b406c45b774104d6/src/ethereum/forks/amsterdam/incremental_mpt.py#L936-L991) dispatches only on RLP list arity (`{0, 2, 17}`) and element types; the only path-length assertion is [`len(nibbles) > 0`](https://github.com/ethereum/execution-specs/blob/f53b22e1cde02b5ed5054c42b406c45b774104d6/src/ethereum/forks/amsterdam/incremental_mpt.py#L959), a *lower* bound on extension paths (problem). So a 2-item node with an arbitrarily long compact path decodes into a `MutableLeafNode`/`MutableExtensionNode` with an unbounded `rest_of_key`/`key_segment`, and the write path ([`_insert_into_leaf`](https://github.com/ethereum/execution-specs/blob/f53b22e1cde02b5ed5054c42b406c45b774104d6/src/ethereum/forks/amsterdam/incremental_mpt.py#L502-L640)) then *splits* it. Nothing rejects it.

## Proposed fix

[...]

PR #3193, opened 2026-07-20
feat(tests): zkevm case for mpt witness with a leaf path of 67 nibbles

pirapira:
### Description

This PR adds a new fixture for zkevm tests. The new fixture contains mpt witness with a leaf path of 67 nibbles. This is currently accepted, but would be rejected on  stateless guests with length checks on mpt paths (like suggested in #3155).

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

"backyard bandits - baby racoons" by Christian Collins is licensed under CC BY-SA 2.0. To view a copy of this license, visit https://creativecommons.org/licenses/by-sa/2.0/?ref=openverse.

### DIFF
--- a/tests/amsterdam/eip8025_optional_proofs/test_witness_state_overlength.py
+++ b/tests/amsterdam/eip8025_optional_proofs/test_witness_state_overlength.py
@@ -1,0 +1,263 @@
+"""Witness MPT decoding of an over-length secured-trie path."""
+
+from dataclasses import replace
+from typing import AbstractSet, Callable
+from unittest.mock import patch
+
+import pytest
+from ethereum_rlp import rlp
+from ethereum_types.bytes import Bytes as AmsterdamBytes
+from ethereum_types.bytes import Bytes32
+from ethereum_types.numeric import U256
+from execution_testing import Alloc, Block, BlockchainTestFiller, Bytes
+
+import ethereum.forks.amsterdam.fork as amsterdam_fork
+from ethereum.crypto.hash import keccak256
+from ethereum.forks.amsterdam.block_access_lists import BlockAccessList
+from ethereum.forks.amsterdam.blocks import Header
+from ethereum.forks.amsterdam.execution_engine.validation_helpers import (
+    _payload_block,
+    _payload_header,
+)
+from ethereum.forks.amsterdam.fork import ChainContext, execute_block
+from ethereum.forks.amsterdam.incremental_mpt import (
+    IncrementalMPT,
+    MutableBranchNode,
+    MutableExtensionNode,
+    MutableLeafNode,
+    decode_witness_to_mpt,
+    mpt_root,
+)
+from ethereum.forks.amsterdam.stateless_guest import (
+    deserialize_stateless_input,
+)
+from ethereum.forks.amsterdam.stateless_host import serialize_stateless_input
+from ethereum.forks.amsterdam.witness_state import (
+    WitnessState,
+    build_code_db,
+    build_node_db,
+)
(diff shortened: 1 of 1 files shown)


pirapira:
CI failure seems to be on `projects/zkevm` branch already.
```
  git fetch origin projects/zkevm
  git worktree add /tmp/execution-specs-zkevm origin/projects/zkevm
  cd /tmp/execution-specs-zkevm
  git submodule update --init --recursive

  just fill --from Amsterdam --until Amsterdam \
    -m "not slow and not derived_test"
```
fails locally for me at 48247eae418a5c80ee3cabe8f477cc45f648dfa6

spencer-tb:
cc @jsign

jsign:
@pirapira, I'm not yet sure whether to open the door to assuming the parent header `state_root` is a malformed MPT. That doesn't seem to be a reasonable assumption, since it would mean the chain is broken.

This would also mean we should not only cover this case, but a lot of other cases like:
- ExtensionNode pointing to another ExtensionNode (invalid)
- BranchNodes having only one child (invalid)
- ExtensionNodes having empty keys
- ... and I guess a lot of other cases.

None of these cases should happen when proving a block with a valid parent state root.

So maybe for a while we'll keep the assumption "parent state root" is valid to avoid an explosion of new tests; at least until we have more bandwidth and green light on things being upstreamed, just to not make the 8025-diff bigger (and potentially add more testing-framework helpers for these cases) (it is already ~16k line changes).

At some point, I think it would be nice to cover these, but not sure about the timing now.

pirapira:
@jsign thank you for the reply!

> we'll keep the assumption "parent state root" is valid

For theorem proving, I need some code saying we are assuming parent state root is valid. Something like execuiton-specs raising an exception on invalid input, meaning the clients can do anything about the input because the input violates assumptions.

jsign:
> For theorem proving, I need some code saying we are assuming parent state root is valid. Something like execuiton-specs raising an exception on invalid input, meaning the clients can do anything about the input because the input violates assumptions.

Maybe I'm not understanding, but why do we need some code in `execution-specs` to formalize an assumption? Can't you add an assumption in the FV theorem that says that `parent.state_root` is a valid MPT?

pirapira:
> why do we need some code in execution-specs to formalize an assumption? Can't you add an assumption in the FV theorem that says that parent.state_root is a valid MPT?

I have three reasons I cannot add this assumption in the theorem.

1. Identifying the correct chain is outside of evm-asm. Maybe somebody starts a genesis block with 67-nibble path in the MPT. evm-asm needs to rely on an external specification to say that chain is not interesting.
2. It's impossible to assume "keccak doesn't collide" (outright contradiction->0=1 is a theorem, no credibility in any theorems). So even if the state root comes from the right chain, maybe the input still contains 67 nibbles on a MPT path.
3. it's execution-specs choice, on each input, whether to specify one correct output or to say instead clients can do anything.

At the moment I'm trying to accept 67 nibbles in evm-asm because execution-specs behaves that way.

pirapira:
Currently "on this input execution-specs doesn't need to be followed" is communicated by not adding tests. I cannot cite absence of certain class of tests for justifying assumptions. If clients are allowed to behave differently from execution-specs on certain inputs, the specification needs to say that explicitly.

jsign:
@pirapira wdyt about cherry-picking this commit I added https://github.com/jsign/execution-specs/commit/3c750417d714c4d25128748502734b3f1543bc7c into this PR?

At least we tighten up this nibble expectation better at spec level. Does this sounds good to you?

PR #3194, opened 2026-07-20
feat(test-benchmark): add deployment contract script, no-rewind flag

LouisTsai-Csie:
### Description

Based on benchmarking call. This PR is not needed for now.

This PR adds two features:
- Manual contract deployment support: Tests like `test_account_access` and `test_ether_transfers_on_receivers` normally rely on state-actor for account/contract generation. This PR adds `test_setup_contract` to manually deploy necessary contracts and delegated accounts when state-actor is unavailable.
- State accumulation flag: Adds `--no-reset-between-tests` to skip rewinding the client head between tests. By default, benchmarks rewind to `start_block` after each test; this flag allows state to accumulate across tests.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/tests/benchmark/stateful/bloatnet/test_setup_contracts.py
+++ b/tests/benchmark/stateful/bloatnet/test_setup_contracts.py
@@ -1,0 +1,165 @@
+"""Deploy the CREATE2 contracts for benchmarks."""
+
+import os
+
+from execution_testing import (
+    DETERMINISTIC_FACTORY_ADDRESS,
+    EOA,
+    Account,
+    Alloc,
+    AuthorizationTuple,
+    BenchmarkTestFiller,
+    Fork,
+    Hash,
+    Op,
+    Transaction,
+    compute_create2_address,
+)
+
+from tests.benchmark.helper.account_creator import (
+    AccountCreator,
+    AccountMode,
+)
+from tests.benchmark.helper.account_sender_receiver import (
+    DELEGATE_BASE_KEY,
+)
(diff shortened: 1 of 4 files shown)


tests/benchmark/stateful/bloatnet/test_setup_contracts.py
```diff
@@ -0,0 +1,165 @@
...
+    deposit_cost = Op.RETURN(
+        0,
+        runtime_size,
+        code_deposit_size=runtime_size,
+    ).gas_cost(fork)
+    base = intrinsic + create_cost + deposit_cost
+    return base + base // 16 + EXECUTION_GAS_BUFFER
+
+
+def test_deploy_existing_contracts(
```
marioevz, line 77: Iiuc (from the PR description) this test function is not really a test, rather it's a helper that is used to deploy contracts, and then the `--no-reset-between-tests` flag is used to run the actual tests. Is this correct?

If so, I think there has to be a better way to accomplish this, because we are basically overloading the test function to perform other tasks, rather than describing the test pre-requisites inside of the tests that need them.

LouisTsai-Csie:
This is required if we were targeting `snapshot` approach to generate the benchmark, as we migrate to state-actor, this would be no longer needed. Re-open if we need the snapshot benchmark again.

LouisTsai-Csie:
Unfortunately, we need to resurrect to run the benchmark on top of Jochemnet

LouisTsai-Csie:
superseded by PR #3304

Issue #2645, opened 2026-04-09
Support SSZ in the Engine API


marioevz:
Tracking issue to support https://github.com/ethereum/execution-apis/pull/764 and support clients that want to implement it.

marioevz:
This alternative was proposed during ACD-C 5-14-2026: https://github.com/ethereum/execution-apis/pull/793

I think there's no difference between on the SSZ side, just how it is handled via REST, so we could start by implementing the SSZ part.

marioevz:
Deprecation of JSON RPC was mentioned during ACD-C 05-28-26, hence the higher priority now.

RazorClient:
Hey!
Can you please assign me this issue!

PR #3196, opened 2026-07-20
feat(test-types): Introduce SSZ model into base types

RazorClient:
### Description

This PR introduces the sszmodel as a descendant of the camelmodel. Its major work is to introduce a Pydantic model which has ssz inbuilt and can transform its members into remarkable-ready ssz-friendly types.
An extension of pr is to add the randomisation machinery for the ssz types as a port from the [SSZ Static](https://github.com/ethereum/consensus-specs/blob/master/tests/formats/ssz_static/core.md) from [here](https://github.com/ethereum/consensus-specs/blob/152790136f5f9c0e7504e14dcb5c381deb32db53/tests/core/pyspec/eth_consensus_specs/debug/random_value.py#L30)

The code for randomisation and the ssz static vector generation can later be moved to a different pr,however putting it here just for initial visibility.

### Related Issues or PRs
https://github.com/ethereum/execution-specs/issues/2645

### Checklist

- [X] Ran fast static checks to avoid CI fails

### Cute Animal Picture

glonk


packages/testing/src/execution_testing/base_types/tests/test_ssz.py
```diff
@@ -0,0 +1,369 @@
...
+class Withdrawal(SszModel):
+    """A pydantic model declared to check the SSZ machinery."""
+
+    index: Uint64
+    validator_index: Uint64
+    address: Address
+    amount: Uint64
+
+
+class ExecutionPayload(SszModel):
```
marioevz, line 69: I'm assuming something like this will replace the existing models that are generated for the test fixtures, e.g. https://github.com/ethereum/execution-specs/blob/9e9bf9068c5bbc337cbbb6794d14786d61840a6a/packages/testing/src/execution_testing/fixtures/blockchain.py#L389

One question is whether we can keep the format of having a single model that declares future-fork fields as `None | Type`, and then `None` is set by previous forks where this field does not appear so it's omitted from the JSON output. How would this work with `SszModel`? Ideally we should keep one model to which we keep appending optional fields for future forks.

packages/testing/src/execution_testing/tools/ssz_vectors.py
```diff
@@ -0,0 +1,453 @@
...
+        pass
+
+    def _str(dumper: Any, data: str) -> Any:
+        style = "'" if data.startswith("0x") else None
+        return dumper.represent_scalar(
+            "tag:yaml.org,2002:str", data, style=style
+        )
+
+    _Dumper.add_representer(str, _str)
+    return yaml.dump(obj, Dumper=_Dumper, sort_keys=False).encode()
```
marioevz:
```suggestion
class _HexQuotingDumper(yaml.SafeDumper):
    """SafeDumper that single-quotes 0x-hex strings (see _yaml_dump)."""


def _represent_str(dumper: Any, data: str) -> Any:
    style = "'" if data.startswith("0x") else None
    return dumper.represent_scalar(
        "tag:yaml.org,2002:str", data, style=style
    )


_HexQuotingDumper.add_representer(str, _represent_str)


def _yaml_dump(obj: Any) -> bytes:
    """
    Dump YAML with 0x-hex strings explicitly single-quoted.

    PyYAML's emitter is not consistent across interpreters about quoting
    strings that look like YAML 1.1 ints (PyPy emits root: 0x... bare, which
    a loader would read back as an integer). Consensus vectors always quote
    them, so force the style instead of trusting the emitter.
    """
    return yaml.dump(
        obj, Dumper=_HexQuotingDumper, sort_keys=False
    ).encode()
```
Minor nit, this avoids recreating the class every call.

Also, `yaml` can be imported at the top of the file.

marioevz (approved):
Overall LGTM.

I just added two minor comments in the entire PR.

I think sane next steps would be to start modifying existing models to add SSZ functionality where required, but without enabling it at first, just so we can verify using [`hasher`](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/?h=hasher#using-the-hasher-tool) that the fixtures before and after are exactly the same.

After that we can start modifying the test fixtures in order to add SSZ variants to the engine API execution payloads where possible.

Then we can start modifying the test consumers to use them.

packages/testing/src/execution_testing/tools/ssz_vectors.py
RazorClient: ahh yeah i had already changed the import to be at the top of the file,my bad : )

packages/testing/src/execution_testing/base_types/tests/test_ssz.py
RazorClient: This would not be possible with the current design as it stands rn as it is structurally not supported by SSZ to have no fields per se(bcoz of ssz being positional.)
However, that being said, I'm looking into it a bit more to find a solution to this
The single-model format itself looks preservable.I'm working out the cleanest shape for that and will follow up tommorow

RazorClient:
Hey directionally, that would make sense.
Having diffs with hasher is the best way to make sure no regression occurs.
However just adding ssz variants to the engine api testing(yes, some of it might work but to faithfully test out the 793 changes) would not work per se and this would require a few more changes here and there,but that we can discuss on the meet and align towards the direction

packages/testing/src/execution_testing/base_types/ssz.py
```diff
@@ -0,0 +1,1012 @@
...
+
+
+def _is_ssz_excluded(model_cls: Type["SszModel"], name: str) -> bool:
+    metadata = model_cls.model_fields[name].metadata
+    if any(isinstance(m, _SszExclude) for m in metadata):
+        return True
+    return any(
+        name in _comment_excluded_names(klass)
+        for klass in model_cls.__mro__
+        if klass is not SszModel and issubclass(klass, SszModel)
```
marioevz:
```suggestion
        name in _comment_excluded_names(cls)
        for cls in model_cls.__mro__
        if cls is not SszModel and issubclass(cls, SszModel)
```
Minor nit, just for convention's sake.

packages/testing/src/execution_testing/base_types/ssz.py
```diff
@@ -0,0 +1,1012 @@
...
+def _is_fork_optional(model_cls: Type["SszModel"], name: str) -> bool:
+    ann = model_cls.model_fields[name].annotation
+    return _unwrap_optional(ann)[1]
+
+
+_SSZ_EXCLUDE_COMMENT = re.compile(r"#\s*ssz_exclude\b")
+
+
+@lru_cache(maxsize=None)
+def _comment_excluded_names(klass: type) -> FrozenSet[str]:
```
marioevz: Maybe this is too much (code introspection is something we rarely use in the codebase). Are there any cases where `Annotated` + `ssz_exclude()` does not cover our needs?
Another option is to check the `exclude` value passed to pydantic's `Field`, but I'm not 100% sure it's possible.

packages/testing/src/execution_testing/base_types/ssz.py
```diff
@@ -0,0 +1,1012 @@
...
+    """
+
+    __active_fields__: ClassVar[Sequence[int]] = ()
+
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs: Any) -> None:
+        """Check the active-field bitvector agrees with the field count."""
+        super().__pydantic_init_subclass__(**kwargs)
+        active = cls.__active_fields__
+        if active and sum(active) != len(cls.model_fields):
```
marioevz: Perhaps we need `_included_fields`, otherwise the excluded fields are expected to be in the active list.

packages/testing/src/execution_testing/base_types/ssz.py
```diff
@@ -0,0 +1,1012 @@
...
+    if isinstance(spec, (SszContainer, SszProgressiveContainer)):
+        return build_ssz_type(spec.model, _nested_fork(spec.model, fork))
+    if isinstance(spec, SszBool):
+        return boolean
+    raise TypeError(f"unhandled SSZ type {spec!r}")
+
+
+def _active_fields(model_cls: Type["SszModel"]) -> Sequence[int]:
+    declared = getattr(model_cls, "__active_fields__", ())
+    return declared if declared else [1] * len(model_cls.model_fields)
```
marioevz: Similar to comment above regarding `_included_fields`.

marioevz (commented):
Just a few comments, thanks!

packages/testing/src/execution_testing/base_types/ssz.py
RazorClient: Dropped it.
It does see too much, and I can't think of cases where the later doent cover our need.

Field(exclude=...) won't work as the alternative: it drops the field from pydantic 
but these fields are JSON-only — they must stay in model_dump() and only vanish from SSZ.

packages/testing/src/execution_testing/base_types/ssz.py
RazorClient: yeah, my bad :)
added a test regarding this too

Issue #3168, opened 2026-07-14
feat(tests): Add block-level access list validation to benchmark tests


marioevz:
**_Note: AI generated from meeting notes_**

## Description

We should add automated block-level access list checks to the EELS benchmark tests.

Today, block access lists are useful for manually confirming that benchmarks target the expected state. In the discussion, the team noted that this is a strong debugging tool and helps verify that the same state is being accessed across benchmark variants.

The gap is that these checks are still manual. The proposal was to add validation during state filling so the benchmark generation process confirms that the filled benchmarks are actually targeting the intended state patterns, such as existing accounts, existing contracts, and non-existing accounts.

## Proposed change

Add block-level access list validation to the EELS benchmark test pipeline so that, during fill time, we automatically verify that each generated benchmark is targeting the expected state.

## Why this matters

- Improves confidence that generated benchmarks match their intended target state
- Reduces manual verification work during debugging
- Makes state-targeting errors easier to catch earlier in the fill process

## Scope

Add checks that validate, at fill time, that benchmark cases target the intended category, for example:

- existing accounts
- existing contracts
- non-existing accounts
- expected storage patterns where applicable

cc @jochem-brouwer @LouisTsai-Csie

LouisTsai-Csie:
After in-depth review, I believe BAL validation only partially resolves our account verification problem.

### The Core Requirement

Stateful benchmarks depend on pre-deployed accounts in snapshots. We need to confirm these accounts actually exist at runtime, not just that they're listed somewhere.

### What We Need to Verify:
Tests like `test_account_access` and `test_ether_transfers_onchain_receivers` use different account modes (`EXISTING_EOA`, `EXISTING_CONTRACT_*`, etc.). We must validate:
- Accounts exist (haven't degraded to non-existent -> really being deployed)
- Balances are correct
- Code is properly deployed

### Why BAL Validation Falls Short
The "after test run" approach using BAL access lists can capture some state changes (`BalAccountChange`, `BalNonceChange`), but it has a critical gap: `BALANCE` and `EXTCODE` opcodes don't produce post-state data in BAL. This makes it impossible to validate whether:
- Accounts had the expected balance during execution
- Deployed code actually exists

So while BAL can catch some deployment, it can't validate the specific account properties our benchmarks rely on.

This PR https://github.com/ethereum/execution-specs/pull/3197 illustrate my idea of how the validation could work

fselmo:
[...]

PR #3197, opened 2026-07-20
feat(test-benchmark): on-chain account verification

LouisTsai-Csie:
### Description

Adds `--verify-deployed-accounts` flag to `fill-stateful`

### Problem:
Stateful benchmarks pre-deploy accounts on snapshots, but if a target is missing, the accessing opcode silently takes the cheap path (empty account/no code) while the transaction still returns `status=1`. The benchmark measures the wrong case with no warning. Existing validation (`expected_receipt_status`, `verify_post_state`) can not catch this.

### Solution
Before each benchmark runs, verify that accessed accounts exist on the snapshot with their claimed shape at the start block.

1. Registration: Tests register expected account shapes through the same generators used to produce them.
2. Verification timing: Runs in `BlockchainTest.make_stateful_fixture` before any block is built, querying `start_block` state (the reset point for each test).
3. Efficient queries:
   - Batches `eth_getBalance` / `eth_getTransactionCount`
   - Only fetches code for EIP-7702 designators (23 bytes)
4. Mode-aware checks:
   - CREATE2 contracts: `nonce >= 1` (address binds code)
   - Balance-only EOAs: `balance > 0`
   - Non-existent accounts: asserts absence
   - Delegated authorities: code starts with `Spec.DELEGATION_DESIGNATION`

### Optimization
Session-scoped deduplication: each target family is verified once across the whole session, even across multiple `test_account_access` parametrizations and shared between related tests.

### Related Issues or PRs

issue #3168

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
@@ -153,11 +153,9 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         dest="verify_full_accounts",
         default=False,
         help=(
-            "Verify EVERY predeployed target account each benchmark assumes, "
-            "instead of only sampling the first and last of each range. "
-            "Predeploy verification (via pre.expect_account_state) always "
-            "runs at the start block; this flag makes it exhaustive. Uses "
-            "eth_getBalance / eth_getTransactionCount / eth_getCode."
+            "Verify all predeployed targets instead of sampling. "
+            "By default, only first and last accounts per range are checked. "
+            "This flag checks every account at start_block.)"
         ),
     )
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
@@ -148,15 +148,16 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         ),
     )
     group.addoption(
-        "--verify-deployed-accounts",
+        "--verify-full-accounts",
         action="store_true",
-        dest="verify_deployed_accounts",
+        dest="verify_full_accounts",
         default=False,
         help=(
-            "Before each benchmark runs, verify the accounts it assumes are "
-            "predeployed on the snapshot (via pre.expect_account_state) "
-            "actually exist with the claimed shape at the start block. Uses "
-            "eth_getBalance / eth_getTransactionCount / eth_getCode"
+            "Verify EVERY predeployed target account each benchmark assumes, "
+            "instead of only sampling the first and last of each range. "
+            "Predeploy verification (via pre.expect_account_state) always "
+            "runs at the start block; this flag makes it exhaustive. Uses "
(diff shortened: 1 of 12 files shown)


tests/benchmark/helper/account_creator.py
```diff
@@ -1,23 +1,102 @@
...
+
+    def register(
+        self, pre: Alloc, address: Address, *, label: str | None = None
+    ) -> None:
+        """
+        Register verification expectation for address on pre.
+
+        No-op if deferred verification unsupported (e.g., non-stateful fill).
+        """
+        expect = getattr(pre, "expect_account_state", None)
```
marioevz: This can be improved. Let's define it at the parent class ([here](https://github.com/ethereum/execution-specs/blob/5fa5938b1ce01c661b3e9beaa403e4edddd11e1a/packages/testing/src/execution_testing/test_types/account_types.py#L163)) and implement it for every type of `pre`.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1457,6 +1457,13 @@ def make_stateful_fixture(
                 max_fee_per_blob_gas=max_fee_per_blob_gas,
             )
 
+        # Verify predeployed targets at start_block (fill-stateful mode).
+        # Catches silent degradation to empty accounts.
+        if t8n.verify_deployed_accounts:
+            verifier = getattr(self.pre, "verify_deployed_accounts", None)
```
marioevz: If it needs to check for this method, perhaps this is not the correct place to make this verification (perhaps [here?](https://github.com/ethereum/execution-specs/blob/5fa5938b1ce01c661b3e9beaa403e4edddd11e1a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/execute.py#L368))

packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
```diff
@@ -147,6 +147,18 @@ def pytest_addoption(parser: pytest.Parser) -> None:
             "opt-in."
         ),
     )
+    group.addoption(
+        "--verify-deployed-accounts",
```
marioevz: From what I can tell, this should _not_ be an optional flag. If the test requires the account, then it should always verify its existence.

If the reason to make this optional is speed, perhaps what we can do is instead sample contracts instead of verifying every single one (test the first and last contracts of your range).

marioevz (commented):
Hey Louis, thanks for this, I left some comments to what I think are breaking changes of this feature. Let me know what you think!

LouisTsai-Csie:
@marioevz Thank you for the review! I've pushed some changes to the current structure. Additionally, i add the following changes:

- Restructure folder and extract account verification logic into independent `account_verification.py` module
- Validate accounts early to fail fast instead of after all RPC queries complete
- Add session-scoped `verified_account` variable in `conftest.py` for benchmark deduplication

marioevz (approved):
LGTM, nice new feature! I've pushed a refactor commit at the end to let `pre` handle the `verify_full_accounts` flag internally so we don't have to pass around this flag in every test everywhere.

marioevz:
Rebased to fix the conflict, will merge once CI passes.

marioevz:
We can apply label `backport benchmarks/amsterdam` once #3262 is merged.

PR #3200, opened 2026-07-20
feat(spec-specs): add EIP-7997 deterministic factory deployment

nerolation:
Reopening https://github.com/ethereum/execution-specs/pull/3198 through this pr.

This time the state access is gated by checking the PreviousHeader, and will only happen in the fork transition block.
It feels like the right solution now.

cc @marioevz @MariusVanDerWijden


MariusVanDerWijden:
I think thats actually what we do in geth already

marioevz:
> and will only happen in the fork transition block

Does this imply that the fork transition block should contain the account code access in the block level access list and also the execution witness?

nerolation:
I've now clarified it further in the EIP. It's now saying that even with the contract being accessed at the fork transition, it won't be in the BAL. Only on chains where the contract doesn't exist we put it in the BAL.

kdeme:
Nimbus EL also does it like that (but the witness tracking for stateless testing triggered this issue).

> Only on chains where the contract doesn't exist we put it in the BAL.

We should clarify in the EIP the behaviour for the witness also (cc @jsign )

(I realize that within execution-specs this is tracked with the same mechanism, but not all clients work like that).

jsign:
In EEST, the execution witness uses the same state tracker as BALs.

From the execution witness POV, at the fork transition block there will be a state lookup to check if the contract exists or not. This implies this state access should materialize into an MPT proof in the execution witness as expected (just as any other state access that happens in the block). 

If we exclude this state lookup from the execution witness, then a stateless program can't know whether the contract exists, and thus can't decide what to do since that state access isn't part of the data received (i.e. remember is a stateless program that receives state information as input).

@nerolation, is there any benefit to having this exclusion rule for BALs? I feel like having an exception for a single lookup only in a fork transition might not be worth it and might even make things more risky (since sounds not natural to ignore a state access). It would force EEST to have this rule at the BAL level, not at the "state tracker" level shared with the execution witness. It is kind of saying that only after some starting point of the block execution state is tracked, instead of just doing it always.

Said differently, any state access should go through the state tracker feels like a simpler rule to me (i.e. including any extra state access in fork transitions).

Musthaq22 (approved):

nerolation:
Yeah I can def see this reasoning. Seems like we have to bring this to ACDT on Monday. Right now the EIP clarifies that the lookup wouldn't be covered by the BAL. but I'm happy to update this:
https://github.com/ethereum/EIPs/pull/11991

nerolation:
The latest commit removed the early return guard and let's the deployment/check be included independent from the factory existing or not.
That's the respective EIP change:
https://github.com/ethereum/EIPs/pull/11991/changes/30e0e16a4bcd2877b8eb637a559907780e575135

jochem-brouwer:
Ref, PR (seems to) implement https://github.com/ethereum/EIPs/pull/11991

src/ethereum_spec_tools/evm_tools/t8n/__init__.py
```diff
@@ -380,7 +383,31 @@ def run_state_test(self) -> Any:
...
+        block's fork name in the environment.
+        """
+        if self.env.parent_fork is None:
+            return
+        if not self.fork.has_deploy_deterministic_factory:
+            return
+        parent_short_name = re.sub(
+            r"(?<!^)(?=[A-Z])", "_", self.env.parent_fork
+        ).lower()
+        parent_short_name = re.sub("^b_p_o", "bpo", parent_short_name)
```
jochem-brouwer, line 400: Can you leave a comment what this does or what this parent_short_name is? Does this standardize fork (transition?) names?

tests/amsterdam/eip7928_block_level_access_lists/test_cases.md
```diff
@@ -187,4 +187,8 @@
...
+| `test_factory_install_merged_with_same_block_use` | Ensure the factory install and a same-block CREATE2 use merge into a single BAL entry across indices. File: `tests/amsterdam/eip7997_deterministic_factory_predeploy/test_fork_transition.py`. | Factory removed from genesis. The activation block itself contains a CREATE2 deployment through the freshly installed factory. | BAL **MUST** contain one factory entry with `code_changes=[(0, runtime code)]` and `nonce_changes=[(0, 1), (1, 2)]` — the install at the pre-execution index and the tx-driven increment merged in index order. | ✅ Completed |
```
jochem-brouwer, line 193: There is one more case:
This is for chains which *could have* deployed the factory the "normal" way but did not, and now it is inserted as irregular state transition. We then replay the create tx https://etherscan.io/tx/0xeddf9e61fb9d8f5111840daef55e5fde0041f5702856532cdbb5a02998033d26

This should be rejected due to https://eips.ethereum.org/EIPS/eip-684 (the tx is valid, the contract deployment is not)

tests/amsterdam/eip7997_deterministic_factory_predeploy/test_fork_transition.py
```diff
@@ -104,3 +105,265 @@ def test_factory_deploys_across_transition(
...
+@pytest.mark.pre_alloc_mutable
+@pytest.mark.parametrize(
+    "factory_nonce,post_nonce,expected_nonce_changes",
+    [
+        pytest.param(
+            0,
+            1,
+            [BalNonceChange(block_access_index=0, post_nonce=1)],
+            id="zero_nonce_set_to_one",
+        ),
```
jochem-brouwer, line 251: This seems to include nonce 0 and nonce 7 as starting nonce, not the implied range `[0,7]`

jochem-brouwer (commented):
Some minor comments!

nerolation:
While this copies what geth does, it diverges from what other clients do. I'm fine with leaving this un-speced. No impact on mainnet.

PR #3203, opened 2026-07-22
feat(test-benchmark): add stub account configuration, with state-actor example config

LouisTsai-Csie:
### Description

Stateful benchmarks require pre-deployed contracts and accounts on the network. This PR documents the stub accounts needed for each benchmark (addresses, initcodes, storage patterns, etc.) and provides a complete state-actor configuration example.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/docs/filling_tests/stateful_environment.md
+++ b/docs/filling_tests/stateful_environment.md
@@ -1,0 +1,128 @@
+# Stateful Environment Stub Accounts Configuration
+
+Stateful benchmarks require pre-deployed contracts and accounts. You can download a snapshot or use state-actor to generate the required state. This page documents the necessary stub accounts for each benchmark and provides an example state-actor configuration.
+
+Key points:
+
+- Receiver variants are named `receivers-<kebab-case>` (except `NON_EXISTING_ACCOUNT`, which has no entity)
+- Exact addresses, keys, and derivation rules are in the full configuration
+- Account counts scale to the target block gas limit (examples target 300M; adjust for your needs)
+
+## General
+
+Required by every run, regardless of which tests are selected.
+
+| Entry | Configuration | Used By |
+|-------|---------------|---------|
+| `deterministic-deployment-proxy` | Arachnid CREATE2 factory at `0x4e59…956C` | All tests: derives `receivers-existing-contract-*` addresses; needed for EEST pre-alloc deploys |
+| `fill-stateful-seed` | EOA (private key `0x…01`, balance `1e28` wei) | All tests: funds accounts during `fill --stateful` |
+
+## Test-specific
+
+Only needed by the listed tests; omit when those tests are not run.
+
+| Entry | Configuration | Tests |
+| --- | --- | --- |
+| `bloated-eoa-10gb` | `storage_pattern`: bloated storage to cover existing slots | `test_sload_bloated`, `test_sstore_bloated` (`existing_slots=True`) |
+| `receivers-existing-eoa` | `sequential_eoas`: 1-wei EOAs, starting from `0x…1000` | `test_account_access`, `test_ether_transfers_onchain_receivers` (`EXISTING_EOA`) |
+| `receivers-existing-contract-minimal` | `create2_deploys`: 1-byte STOP (`MinimalContractInitcode`) | `test_account_access`, `test_ether_transfers_onchain_receivers` (`EXISTING_CONTRACT_MINIMAL`) |
+| `receivers-existing-contract-same-max` | `create2_deploys`: max-size identical copies | `test_account_access`, `test_ether_transfers_onchain_receivers` (`EXISTING_CONTRACT_SAME_MAX`) |
+| `receivers-existing-contract-diff-max` | `create2_deploys`: max-size unique copies | `test_account_access`, `test_ether_transfers_onchain_receivers` (`EXISTING_CONTRACT_DIFF_MAX`); delegation target for `receivers-7702-delegated` |
+| `receivers-existing-contract-jumpdest` | `create2_deploys`: JUMPDEST analysis | `test_account_access`, `test_ether_transfers_onchain_receivers` (`EXISTING_CONTRACT_JUMPDEST`) |
+| `sender-pool` | `sequential_pkey_eoas`: 1-ETH EOAs with keys `SENDER_BASE_KEY + i`, funded on-chain so senders stay uncached | Benchmarks using `yield_distinct_sender` (e.g., `test_ether_transfers_onchain_receivers`) |
+| `receivers-bittrex-contracts` | `create_preimage_deploys`: contracts at Bittrex controller CREATE addresses (nonces 2+) | `test_ether_transfers_onchain_receivers` (existing-contract receivers) |
+| `receivers-7702-delegated` | `sequential_pkey_delegations`: EIP-7702 authorities (keys `DELEGATE_BASE_KEY + i`), delegating to `receivers-existing-contract-diff-max` | `test_ether_transfers_onchain_receivers` (delegated receivers) |
+
+## Full configuration
+
+An example state-actor configuration targeting a 300M block gas limit:
+
+```yaml
+entities:
+  # Arachnid CREATE2 factory (required by create2_deploys and EEST fill).
+  - kind: contract
+    template: create2_factory
+    name: deterministic-deployment-proxy
+    address: 0x4e59b44847b379578588920cA78FbF26c0B4956C
+  # Seed EOA for fill-stateful funding (private key 0x…01).
+  - kind: eoa
+    name: fill-stateful-seed
+    address: 0x7e5f4552091a69125d5dfcb7b8c2659029395bdf
+    balance: "10000000000000000000000000000"
+    nonce: 0
+  # Bloated storage account (test_sload_bloated / test_sstore_bloated with existing_slots=True).
+  - kind: contract
+    name: bloated-eoa-10gb
+    template: storage_pattern
+    address: 0x87a6314da5ac8832f6e7a176c8fb133b19f5be04
+    nonce: 1
+    balance: "1000000000000000000"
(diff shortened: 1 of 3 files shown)


docs/filling_tests/stateful_environment.md
```diff
@@ -0,0 +1,128 @@
...
+- Exact addresses, keys, and derivation rules are in the full configuration
+- Account counts scale to the target block gas limit (examples target 300M; adjust for your needs)
+
+## General
+
+Required by every run, regardless of which tests are selected.
+
+| Entry | Configuration | Used By |
+|-------|---------------|---------|
+| `deterministic-deployment-proxy` | Arachnid CREATE2 factory at `0x4e59…956C` | All tests: derives `receivers-existing-contract-*` addresses; needed for EEST pre-alloc deploys |
```
jochem-brouwer: Is this required? I believe if this param is not there, we should check if the `0x4e59..956C` address has the expected code, and if it has, use that. If not, then throw.

docs/filling_tests/stateful_environment.md
```diff
@@ -0,0 +1,128 @@
...
+| `fill-stateful-seed` | EOA (private key `0x…01`, balance `1e28` wei) | All tests: funds accounts during `fill --stateful` |
```
jochem-brouwer, line 18: Do we also require this? Should we not use or have a "default seed"? I remember most tests are being filled by 0xa... and I also remember that private key is somewhere in the repo.

docs/filling_tests/stateful_environment.md
```diff
@@ -0,0 +1,128 @@
...
+| `receivers-existing-contract-diff-max` | `create2_deploys`: max-size unique copies | `test_account_access`, `test_ether_transfers_onchain_receivers` (`EXISTING_CONTRACT_DIFF_MAX`); delegation target for `receivers-7702-delegated` |
+| `receivers-existing-contract-jumpdest` | `create2_deploys`: JUMPDEST analysis | `test_account_access`, `test_ether_transfers_onchain_receivers` (`EXISTING_CONTRACT_JUMPDEST`) |
+| `sender-pool` | `sequential_pkey_eoas`: 1-ETH EOAs with keys `SENDER_BASE_KEY + i`, funded on-chain so senders stay uncached | Benchmarks using `yield_distinct_sender` (e.g., `test_ether_transfers_onchain_receivers`) |
+| `receivers-bittrex-contracts` | `create_preimage_deploys`: contracts at Bittrex controller CREATE addresses (nonces 2+) | `test_ether_transfers_onchain_receivers` (existing-contract receivers) |
+| `receivers-7702-delegated` | `sequential_pkey_delegations`: EIP-7702 authorities (keys `DELEGATE_BASE_KEY + i`), delegating to `receivers-existing-contract-diff-max` | `test_ether_transfers_onchain_receivers` (delegated receivers) |
```
jochem-brouwer, line 34: This is fine (for now), but, in the future we should have a location for this. It should be part of the benchmark definition, so the benchmark itself states what prestate it expects (this is something we can then verify using tooling, for instance RPC calls to query for instance the start account and the end account, with the assumption that everything in-between exists, otherwise we have to perform 150k RPC calls for each bench)

docs/filling_tests/stateful_environment.md
```diff
@@ -0,0 +1,128 @@
...
+  # EIP-7702 delegated authorities
+  # Authority i (key=DELEGATE_BASE_KEY + i) delegates to max_diff receiver i.
+  - kind: contract
+    name: receivers-7702-delegated
+    template: sequential_pkey_delegations
+    parameters:
+      start_pkey: "0x959a83d905ff1fab43bf72c3e87020e4c77fd4bde0e5eeb48e5edbf74a9ec64e"
+      code_pattern: max_diff_pre_amsterdam
+      count: 100000
+      balance: "1000000000000000000"
```
jochem-brouwer, line 127: This is very nice, however, I think we should move this directly to a .yaml file and link to this .yaml file here. We can then update this file and consumers can also use it in their tooling to update the "standard" yaml.

We should think about how we can optimize this, because if a state-actor configuration is not equal to this (but for instance adds more items to it) then that also means that benchs which can be filled by this config can then also be filled by the expanded config. 

A benchmark thus has a prestate condition and this also means that we require a specific state-actor item in the .yaml state config 🤔 So we could write tooling which parses the .yaml to validate if it matches the intended tests (check if prestate conditions are met per the .yaml).
If the .yaml is not present we can also query the RPC.

jochem-brouwer (changes requested):
Some comments, the main thing I feel strongly about is that the example yaml used is moved into a file so clients can directly copy-paste it or link to it 😄 👍

docs/filling_tests/stateful_environment.md
LouisTsai-Csie: I'm not certain how the Arachnid `CREATE2` factory is loaded into the network (via genesis?). I need to verify with Rafael or Stefan, but I'd prefer to keep this documented with additional context for clarity.

I pushed a commit for this.

docs/filling_tests/stateful_environment.md
LouisTsai-Csie: This is my understanding:

EELS has a default seed key used for funding consensus tests, but this could be configured via `--rpc-seed-key` flag as well, it is not a hardcoded value to a specific address.

In the gas-benchmarks approach, we seed a key in the funding block and pass it to the stateful filler, which uses it to create transactions. We keep this here to match the gas-benchmarks funding block behavior.

docs/filling_tests/stateful_environment.md
LouisTsai-Csie: > in the future we should have a location for this
Could you please clarify more on this? I am not sure If I fully understand. 

Those stub accounts are defined [here](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/tests/benchmark/helper/account_creator.py), in the benchmark test implementation already.

docs/filling_tests/stateful_environment.md
LouisTsai-Csie: > however, I think we should move this directly to a .yaml file and link to this .yaml file here. We can then update this file and consumers can also use it in their tooling to update the "standard" yaml.

Are you referring to the comment you made here:

> the main thing I feel strongly about is that the example yaml used is moved into a file so clients can directly copy-paste it or link to it

In our current documentation, the `YAML` code block is already copy-pasteable (see the example [here](https://steel.ethereum.foundation/docs/execution-specs/running_tests/hive/client_config/)).

> So we could write tooling which parses the .yaml to validate if it matches the intended tests

We could provide additional tooling for configuration validation, but maybe in `benchmarkoor` or `state-actor`.  The end goal for this doc is to specify what's necessary items for certain benchmark test to run,

docs/filling_tests/stateful_environment.md
jochem-brouwer: To clarify: the benchmark definitions in EELS should be self-contained. That is: a benchmark has all the data required for a benchmark to be successful: for instance, the benchmark itself (the code to run), the verification (does post-state have this specific storage slot?) but also: the required **pre-state**. This is the new feature. Right now in our documentation we are providing this generic file `docs/filling_tests/stateful_environment.md` which describes which tests use which pre-state.

But this is the wrong direction: the test should definite which pre-state it uses, not some generic documentation. Because if we change the test, such that it now requires different pre-state, we also have to edit the documentation above, which is what we will forget. We should therefore define that in the benchmark itself to keep it self-contained. This is not something we should do now, but it is something we should move to in short-term :smile: :+1: 

(Let me know if still unclear, happy to discuss/clarify)

docs/filling_tests/stateful_environment.md
jochem-brouwer: If a client wants to copy the YAML code block this is much harder to do (what if our documentation changes slightly?) than if this would be a .yaml file. We could then in the documentation say: 

> See the example state yaml at [this file](./generic-prestate-state-actor.yaml)

(filename as example, need to think of something better, but it has to be constant)

docs/filling_tests/stateful_environment.md
jochem-brouwer: This makes sense, thanks, will resolve this one :smile: :+1:

docs/filling_tests/stateful_environment.md
LouisTsai-Csie: Thanks for clarification!

> the test should define which pre-state it uses, not some generic documentation

For this doc I treat it as an example config for people setting up `benchmarkoor`/`state-actor`, the actual definition is in `account_creator.py`. But the `Used By` column will go stale like you said. I can drop it, or keep it with a note that it is not normative. Any preference?

> for instance, the benchmark itself (the code to run), the verification (does post-state have this specific storage slot?) but also: the required pre-state

Does #3197 already cover this? It adds `pre.expect_account_state(...)`, so a benchmark can declare the accounts it needs. If that's the right direction, what's left is migrating the existing benchmarks and adding the RPC check. If you meant something wider than account existence (storage layout, count, code pattern), let me know.

Or do you want to deploy these pre-state during the benchmark? This would be very hard since the deployment phases process in EELS is very slow.

docs/filling_tests/stateful_environment.md
LouisTsai-Csie: Moving it to a file only fixes how you copy it, not what happens after. If someone copies it and we change the file later, nothing tells them. That part only gets fixed by the validation tooling you mentioned in the other thread, once benchmarks declare their own pre-state.

On where. There is already a state-actor config in `ethpandaops/benchmarkoor-tests` under `configs/datadirs/state-actor/v1/`, and that path is already versioned, so a breaking change becomes v2 and anyone on v1 is unaffected. What about putting the yaml there and link from this doc?

PR #3209, opened 2026-07-22
bug(test-eest): reconfigure output streams to UTF-8 on Windows consoles

Kumarutkarsh9470:
The `eest` commands print Unicode characters (box drawing in `info`, emoji in `clean` and `make`) via `click.echo`. On a Windows console using a legacy code page such as `cp1252`, these characters cannot be encoded and the command aborts with `UnicodeEncodeError`.

Reconfigure `sys.stdout`/`sys.stderr` to UTF-8 in the `eest` group callback, guarded so streams that do not support reconfiguration (for example captured output under tests) are left untouched. Add regression tests covering a legacy-encoded stdout.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

cat

### DIFF
--- a/packages/testing/src/execution_testing/cli/eest/cli.py
+++ b/packages/testing/src/execution_testing/cli/eest/cli.py
@@ -3,10 +3,32 @@
 Invoke using `uv run eest`.
 """
 
+import sys
+
 import click
 
 from .commands import clean, info
 from .make.cli import make
+
+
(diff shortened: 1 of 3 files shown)


marioevz (approved):
Nice! Thanks for this fix.

cc @danceratopz I think it would be nice for us to enable a weekly windows run so we can keep the support for the platform, not urgent but wdyt?

PR #3215, opened 2026-07-23
fix(tests): enhance & un-skip Amsterdam ported static tests (Pt. 1)

marioevz:
### Description
Amsterdam's EIP-8037 two-dimensional (regular vs. state) gas model broke a large batch of machine-ported `tests/ported_static/` tests, which were temporarily parked in `amsterdam_skip_list.txt`. These fillers encode gas the wrong way for a fork that reprices: hardcoded absolute `GAS` snapshots, tight OOG gas budgets, and per-opcode gas constants baked into expectations. When EIP-8037 moved account creation / storage writes into the *state* dimension (and EIP-2780 dropped the intrinsic tx cost 21000 → 15000), every one of those pinned numbers shifted and the tests failed on Amsterdam.

This PR (**Part 1**) rewrites a first tranche of them to be **fork-robust** — they now express gas through framework constructs that track the live schedule (`CodeGasMeasure`, `bytecode.gas_cost(fork)` / `.regular_cost(fork)` / `.state_cost(fork)`, opcode metadata) rather than magic numbers — so a future fork that re-prices gas no longer silently breaks them. Each enhanced test is verified across its full `valid_from` range **and** on Amsterdam, and its skip-list entry removed.

**Result:** the Amsterdam ported-static skip list drops from **258 → 153** entries (105 un-skipped). Along the way ~40 near-identical sibling fillers are consolidated into a handful of parametrized tests (56 files changed: +1067 / −4621), and most tests gain much wider fork coverage (the ported `valid_from` was almost always an over-tight `Cancun`; many now run from Frontier / Berlin / SpuriousDragon onward).

The remaining 153 skipped cases — dominated by OOG-boundary, revert-depth, refund, and collision families — are left for a **follow-up PR (Part 2)**.

#### Highlights

- **MCOPY copy-cost (55 cases, largest single win).** The filler probed MCOPY cost via a razor-thin OOG boundary whose instrumentation `SSTORE` is repriced into ~97920 state gas on Amsterdam, so every case OOG'd. Reframed to measure the copy cost directly with `CodeGasMeasure` over pre-expanded memory, asserting `mcopy.gas_cost(fork)`.
- **`RawCall*GasAsk` (10 → 1).** Verifies the EIP-150 "all but one 64th" gas-forwarding rule. An outer call caps the caller frame at a known budget so the forwarded amount is deterministic (independent of the tx gas limit), the callee returns its observed `GAS` up to the top frame (avoiding a lower-frame state-gas trap), and the expectation is derived as `all_but_one_64th(caller_gas − call.gas_cost(fork)) + stipend − GAS`.
- **`RawCreate*Gas` (6 → 1).** One parametrize over value / memory / failure; confirms an EIP-8037 behavior — a balance-failure `CREATE` charges regular gas only (`regular_cost(fork)`), never state gas.
- **Empty-contract & behavior families.** CREATE/CALL gas via `CodeGasMeasure`; gas-independent behavior tests freed of tight `gas_limit`s; subcall `gas` operands left to default (`Op.GAS`) instead of hardcoded constants; dynamic addresses throughout; `protected=fork.supports_protected_txs()` to keep the fork floor at the behavior's real EIP rather than masking a signature floor.

#### Methodology / tooling

- Adds the **`enhance-ported-test`** Claude Code skill (`.claude/commands/enhance-ported-test.md`) and wires it into `CLAUDE.md` — an ordered, empirically-validated methodology for turning a ported filler into idiomatic, fork-robust Python (max out gas, dynamic addresses, `CodeGasMeasure`, consolidation, `valid_from` floors, re-pinning discipline). This is the playbook Part 2 will follow.

### Related Issues or PRs
N/A.

### Checklist

### Cute Animal Picture


spencer-tb:
Added some extra tweaks here: https://github.com/marioevz/execution-specs/pull/7/

spencer-tb (approved):
LGTM!

PR #3216, opened 2026-07-23
chore: sketch pbt fork related changes needed for adding initial PBT

kevaundray:
### Description

Initial changes needed to add the skeleton code for PBT

### Related Issues or PRs

N/A.

### Checklist

- [ ] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [ ] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3218, opened 2026-07-23
refactor(spec): make state interface fully implementation agnostic

kevaundray:
### Description

The current interface:

- Returns the trie nodes that have changed, which leaks MPT types

   This return type was not being used since the base implementation only cares about computing the state root, so I removed it. (Noting that this would be useful for incremental trie implementations)

- Assumes that the trie doesn't commit to the code directly (this is true for MPT)

  I added code_changes as a parameter for this; the MPT just ignores it.
  
Pulled out of #3216 

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture


kevaundray:
If we did want to return the trie changes, then I think it would need to be typed as `Bytes`and would incur a serialization cost

src/ethereum/state.py
```diff
@@ -133,218 +124,26 @@ def account_has_storage(self, address: Address) -> bool:
         """
         ...
 
-    def compute_state_root_and_trie_changes(
+    def compute_state_root(
         self,
         account_changes: Dict[Address, Optional[Account]],
         storage_changes: Dict[Address, Dict[Bytes32, U256]],
+        code_changes: Dict[Hash32, Bytes],
```
kevaundray: This can only be `code_additions`, but I used `code_changes` to be consistent with the other parameters

src/ethereum/state.py
```diff
@@ -133,218 +124,26 @@ def account_has_storage(self, address: Address) -> bool:
...
         storage_clears: AbstractSet[Address] = frozenset(),
-    ) -> Tuple[Root, List["InternalNode"]]:
+    ) -> Root:
```
gurukamath:
```suggestion
    def compute_state_root(self, block_diff: BlockDiff) -> Root:
```

kevaundray:
[0142e35](https://github.com/ethereum/execution-specs/pull/3218/commits/0142e3535f3c230dee91e0b991554091d25513eb) adds changes to make the tooling trie agnostic

src/ethereum/state.py
```diff
@@ -133,218 +124,26 @@ def account_has_storage(self, address: Address) -> bool:
         """
         ...
 
-    def compute_state_root_and_trie_changes(
+    def compute_state_root(
         self,
         account_changes: Dict[Address, Optional[Account]],
         storage_changes: Dict[Address, Dict[Bytes32, U256]],
+        code_changes: Dict[Hash32, Bytes],
         storage_clears: AbstractSet[Address] = frozenset(),
```
gurukamath: Would `code_changes: Dict[Hash32, Bytes]` still be the right shape to have if another prestate implementation commits to the code directly instead of the code hash. In such cases there is an implicit join in this shape. `code_hash = hash(account_changes[code])` -> `code_changes[code_hash]`. I don't think this is a huge issue but probably worth thinking about. And maybe a line in the docstring?

src/ethereum/state.py
```diff
@@ -30,15 +30,6 @@
 from ethereum_types.numeric import U256, Uint
 
 from ethereum.crypto.hash import Hash32, keccak256
-from ethereum.merkle_patricia_trie import (
```
gurukamath, line 33: Now that this module no longer imports from `merkle_patricia_trie.py`, [this comment](https://github.com/kevaundray/execution-specs/blob/115d87abd85e174ccac51fa74ddba17366bc7154/src/ethereum/merkle_patricia_trie.py#L74) and the `TYPE_CHECKING` gaurd on that module are no longer relevant.

src/ethereum/state.py
kevaundray: I guess it could have been Dic[Address, Bytes] but then we would not have code deduplication across accounts within the same block -- would leave it upto you, re what makes sense

src/ethereum/merkle_patricia_trie.py
```diff
@@ -65,16 +64,11 @@
 from ethereum_types.numeric import Uint
 from typing_extensions import assert_type
 
-from ethereum.crypto.hash import Hash32, keccak256
+from ethereum.crypto.hash import keccak256
+from ethereum.state import Account, Address, Root
```
gurukamath, line 68: The `root` function imports `Root` once again on line 470. Should be removed from that function

gurukamath (approved):
Looks good to me. One very minor nit

PR #3187, opened 2026-07-17
refactor(test-benchmark): align test suite with `glam-devnet-7` spec

LouisTsai-Csie:
### Description

Cherry-pick PR #2947, PR #3137 and PR #3098 from `devnets/glamsterdam/6-benchmarks`.

Commit 552f552b7f74428766021f7ae7549fc4b5e31981 is the only one that adds on top of the cherry-pick, resolving gas accounting issue introduced in devnet-7

### Related Issues or PRs

issue #3106 

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

PR #3219, opened 2026-07-23
fix(tests,test-fill): fix Engine X fills on BAL forks

spencer-tb:
### Description

The `tests-glamsterdam-devnet@v7.2.1` release fill failed with `EngineXExecutionDriftError` on 18049 of 24683 Engine X fixtures.

On BAL forks the EIP-2935 system call writes the parent hash (the group genesis hash at payload 0) into every payloads `blockAccessList`, so the check flagged an expected state root derived difference. Decode the BAL and mask the (leading zero-trimmed) parent hash instead of comparing it verbatim, keeping the rest of the BAL in the comparison so leaked accounts are still caught.

`test_gas_cost_return` blind calls undeclared `0x1000`/`0x2000`, which pre alloc packing can populate (isolated as in #3176). The `test_bal_2935_query[...valid_block_number]` params SSTORE the queried genesis hash where the per payload mask cannot reach (also isolated). The skip list fix this PR originally carried was dropped on rebase, since #3390's registry derived format tokens cover all Engine X id variants. The drift check now runs only on a clean session and only when optimistic grouping is enabled.

Validated post rebase on the #3390 base with a full `tests/ported_static` fill at Amsterdam with `--generate-all-formats`: 25412 passed, 196 skipped, 0 failed. The drift check passed on all 6353 Engine X fixtures across 518 pre alloc groups.

### Related Issues or PRs

Fixes the `fill (amsterdam)` failure in the [glamsterdam-devnet@v7.2.1 release dispatch](https://github.com/ethereum/execution-specs/actions/runs/29907017961/job/88880830188), same failure class as the Benchmarking CI red since #3187.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -2050,8 +2050,6 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
     # Log immediately when hook is entered (before any early returns)
     _log_timing(f"pytest_sessionfinish ENTERED (worker={is_worker})")
 
-    del exitstatus
-
     # Save pre-allocation groups after phase 1
     fixture_output: FixtureOutput = session.config.fixture_output  # type: ignore[attr-defined]
     session_instance: FillingSession = session.config.filling_session  # type: ignore[attr-defined]
@@ -2151,9 +2149,14 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
             file.unlink()
         _log_timing(f"Lock files removed in {time.time() - t0:.1f}s")
 
-    if not session.config.getoption("optimistic_pre_alloc_grouping_disabled"):
-        # Loudly fail the fill if pre-alloc group packing changed any Engine X
-        # test's execution (raises on drift, like a pre-alloc collision).
+    # Loudly fail the fill if pre-alloc group packing changed any Engine X
+    # test's execution (raises on drift, like a pre-alloc collision). Only
+    # checked on an otherwise clean session: raising from this hook aborts
+    # the terminal FAILURES/short-summary sections, so it would hide any
+    # test failures (which already fail the fill and must surface first).
+    if exitstatus == pytest.ExitCode.OK and not session.config.getoption(
+        "optimistic_pre_alloc_grouping_disabled"
+    ):
         _log_timing("verify_engine_x_execution: starting...")
         t0 = time.time()
         engine_x_check = verify_engine_x_execution(fixture_output.directory)
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -2147,8 +2147,6 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
     # Log immediately when hook is entered (before any early returns)
     _log_timing(f"pytest_sessionfinish ENTERED (worker={is_worker})")
 
-    del exitstatus
-
     # Save pre-allocation groups after phase 1
     fixture_output: FixtureOutput = session.config.fixture_output  # type: ignore[attr-defined]
     session_instance: FillingSession = session.config.filling_session  # type: ignore[attr-defined]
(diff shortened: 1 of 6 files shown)


spencer-tb:
Forced pushed these commits to `devnets/glamsterdam/7` so I can re-tag v7.2.1.

spencer-tb:
https://github.com/ethereum/execution-specs/actions/runs/30085505637

packages/testing/src/execution_testing/fixtures/engine_x_checks.py
```diff
@@ -74,8 +110,13 @@ def _scrubbed_payloads(fixture: Dict[str, Any]) -> List[Any]:
         entry = json.loads(json.dumps(entry))
         params = entry.get("params")
         if params and isinstance(params[0], dict):
+            payload = params[0]
+            bal = payload.get("blockAccessList")
```
marioevz, line 114: The problem with this is that we are passing `fixture` as a black-boxed python dictionary. We have the definitions of all the fixture types so we can parse them (yes it's another round of json parsing, but that's a problem with the approach this file takes to try to make its verifications).

packages/testing/src/execution_testing/fixtures/engine_x_checks.py
```diff
@@ -12,6 +14,40 @@
...
+def _masked(node: Any, parent_hash: bytes) -> Any:
+    """Mask every ``parent_hash`` leaf in a decoded BAL, hex the rest."""
+    if isinstance(node, bytes):
+        if parent_hash and node == parent_hash:
+            return _PARENT_HASH_PLACEHOLDER
+        return node.hex()
+    return [_masked(child, parent_hash) for child in node]
+
+
+def _scrubbed_bal(bal_hex: str, parent_hash_hex: str) -> Any:
```
marioevz, line 30: It's not great that we need this in the first place, this highlights how brittle this file is.

packages/testing/src/execution_testing/fixtures/engine_x_checks.py
marioevz, line 1: This file is just a nightmare to maintain: it introduced a post-fact verification and provides zero indication about the cause when there's a failure.

marioevz (commented):
My main takeaway from this review is that #3122 broke the soundness of the pre-allocation grouping, and it tried to side-step this issue with `packages/testing/src/execution_testing/fixtures/engine_x_checks.py`. This is the second time in two weeks that this file breaks (#3176), and it's going to keep on breaking in the future.

I don't think this PR is incorrect, is just that we are patching a patch that is going to grow in complexity and bring us more issues in the future.

spencer-tb:
Let me create an issue or PR to try to make the file less brittle

spencer-tb:
Rebased! Tests pass :)

packages/testing/src/execution_testing/fixtures/engine_x_checks.py
```diff
@@ -12,6 +14,40 @@
...
+_PARENT_HASH_PLACEHOLDER = "<parent-hash>"
+
+
+def _masked(node: Any, parent_hash: bytes) -> Any:
+    """Mask every ``parent_hash`` leaf in a decoded BAL, hex the rest."""
+    if isinstance(node, bytes):
+        if parent_hash and node == parent_hash:
+            return _PARENT_HASH_PLACEHOLDER
+        return node.hex()
+    return [_masked(child, parent_hash) for child in node]
```
danceratopz: This masks every leaf that equals the trimmed parent hash, not only the EIP-2935 system write. A test that stores its own payload's parent hash somewhere else (for example BLOCKHASH(NUMBER - 1) written to storage) has that write masked on both sides, so the check accepts a grouping that genuinely changed the test's observable state. stateRoot, blockHash and parentHash are already dropped from the payload, so the BAL was the last remaining signal for this class, and the EngineXExecutionDriftError text still advertises it ("or they observe the genesis hash itself").

The BAL pins the 2935 write to one place: the storage change for slot (block_number - 1) % 8191 under the history contract account 0x0000F90827F1C53a10cb7A02335B175320002935. Masking only that leaf keeps a test's own SSTORE of the parent hash in the comparison and closes the blind spot. I swept the tree for single-block parent-hash observers and found no live instance today (the closest calls all clear), so this is about future tests slipping through silently rather than a current bug. If the byte-equality mask stays, the blind spot should at least be stated here and in the error text.

packages/testing/src/execution_testing/fixtures/engine_x_checks.py
```diff
@@ -12,6 +14,40 @@
...
+    values are RLP-encoded with leading zeros trimmed, so the trimmed
+    parent hash is masked. An undecodable BAL (an intentionally malformed
+    one from a negative test) is compared verbatim.
+    """
+    parent_hash = bytes.fromhex(parent_hash_hex.removeprefix("0x"))
+    try:
+        decoded = rlp.decode(bytes.fromhex(bal_hex.removeprefix("0x")))
+    except Exception:
+        return bal_hex
+    return _masked(decoded, parent_hash.lstrip(b"\x00"))
```
danceratopz: Three small robustness points on these lines:

- bytes.fromhex(parent_hash_hex...) sits outside the try, so a malformed parentHash raises out of pytest_sessionfinish after every fixture is already written, and by the reasoning in the new filler comment that also truncates the terminal failure summary. Moving it inside the try (or returning bal_hex on a bad parent hash) avoids that.
- except Exception is broader than what can fail here (ValueError from fromhex, DecodingError from rlp.decode). An unrelated bug would silently downgrade to verbatim comparison and reintroduce exactly the mass false positive this PR fixes. Narrowing to except (ValueError, DecodingError) keeps that failure loud.
- There is no length guard on the mask value, so a short trimmed hash would over-fire across the whole tree. A real keccak output cannot be short, but a len check before the lstrip is cheap and makes the invariant explicit. Note the unit tests' default parent hash trims to a single 0xaa byte, so the masking tests currently run in exactly the over-fire shape.

packages/testing/src/execution_testing/fixtures/engine_x_checks.py
```diff
@@ -74,8 +110,13 @@ def _scrubbed_payloads(fixture: Dict[str, Any]) -> List[Any]:
         entry = json.loads(json.dumps(entry))
         params = entry.get("params")
         if params and isinstance(params[0], dict):
+            payload = params[0]
+            bal = payload.get("blockAccessList")
+            parent_hash = payload.get("parentHash")
+            if isinstance(bal, str) and isinstance(parent_hash, str):
+                payload["blockAccessList"] = _scrubbed_bal(bal, parent_hash)
```
danceratopz: If blockAccessList is a string but parentHash is missing or not a string, masking is silently skipped and the BAL is compared verbatim, which resurrects the mass drift this PR fixes rather than failing loudly. Both fields come from our own serialized fixtures, so an assert or a warning here would make the degrade visible.

packages/testing/src/execution_testing/fixtures/engine_x_checks.py
```diff
@@ -114,7 +155,10 @@ def verify_engine_x_execution(
     `blockchain_test_engine` sibling fixture (filled against the test's own
     pre-allocation in the same session, with an independent `t8n` execution:
     Engine X fixtures never share the transition tool output cache). All
-    payload fields except the state-root-derived ones must match exactly.
+    payload fields except the state-root-derived ones must match exactly;
+    each payload's block access list is compared with its own parent-hash
+    bytes normalized out, since the EIP-2935 system write embeds that
+    state-root-derived value in every BAL.
```
danceratopz: With the mask in place, the guarantee in the EngineXExecutionDriftError text above (lines 80 to 84, "or they observe the genesis hash itself, e.g. via BLOCKHASH(0)") no longer fully holds. A stored copy of the payload's own parent hash is masked, and a transformed hash (keccak, AND, a small SHL) or a hash leaked through a LOG topic drifts via other fields with no hint in the message. Suggest broadening the error text to any observation of a block hash, stored, transformed or logged, and stating the mask's blind spot here if the mask stays byte-equality based.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -2151,9 +2149,14 @@ def _log_timing(msg: str) -> None:
...
-        # Loudly fail the fill if pre-alloc group packing changed any Engine X
-        # test's execution (raises on drift, like a pre-alloc collision).
+    # Loudly fail the fill if pre-alloc group packing changed any Engine X
+    # test's execution (raises on drift, like a pre-alloc collision). Only
+    # checked on an otherwise clean session: raising from this hook aborts
+    # the terminal FAILURES/short-summary sections, so it would hide any
+    # test failures (which already fail the fill and must surface first).
+    if exitstatus == pytest.ExitCode.OK and not session.config.getoption(
+        "optimistic_pre_alloc_grouping_disabled"
+    ):
```
danceratopz: The check's two other skip paths each surface a yellow terminal warning saying the output was not verified, but the new exitstatus path skips silently, and even the _log_timing lines sit inside the gate. A small else setting engine_x_check_warning to say the check was skipped because the session was not clean would keep a failed-then-rerun release fill from looking verified. Related gap: nothing in plugins/filler/tests exercises this block; a pytester case for the gate would be cheap with the existing infrastructure there.

packages/testing/src/execution_testing/fixtures/tests/test_engine_x_checks.py
```diff
@@ -78,6 +87,134 @@ def test_identical_execution_passes(tmp_path: Path) -> None:
...
 
 
+def _bal(parent_hash: bytes, *extra: bytes) -> str:
+    """
+    Encode a minimal BAL-shaped RLP structure embedding a parent hash.
+
+    Storage values are RLP-encoded with leading zeros trimmed, matching
+    the EIP-2935 history write of the parent hash in a real BAL.
+    """
+    values = [parent_hash.lstrip(b"\x00"), *extra]
```
danceratopz: This helper applies the same lstrip the code under test applies, so test_bal_embedded_parent_hash_passes cannot catch a disagreement with the real BAL encoding, which is the one thing worth pinning (post_value is U256 and encodes minimal big endian, so the trim is correct today, but the test would pass either way). Building the fixture BALs through the production BlockAccessList model's rlp would make these tests load bearing. The default parent_hash at line 30 also trims to a single byte, see the note on the length guard in engine_x_checks.py.

spencer-tb:
CIFO: #3265

PR #3246, opened 2026-07-28
tests(binary_tree): add intensive EIP-8297 partitioned binary tree test suites

CPerezz:
### Description

Deepens the EIP-8297 (Partitioned Binary Tree) test coverage on top of #3216, which landed
the `binary_tree` fork sketch with 51 structural unit tests and one EEST smoke test. 

**Structural unit tests** (`tests/binary_trie/`, `just binary-trie-unit-test`, 139 tests)

#### Spec questions this surfaced

Each unresolved question below is tracked as its own issue, so none of them blocks this PR.

1. **Account deletion diverges between the two providers, visibly through EIP-7610.**
   `state_mpt` never pops a deleted account's storage trie; `state_pbt` does. Since
   `account_has_storage` gates contract creation (CREATE and CREATE2), a creation into an
   address that was cleared while holding storage **succeeds under BinaryTree and would be
   rejected under MPT** — We need to decide on this @kevaundray (tracked in #3253)
2. **A balance ≥ 2¹²⁸ has no protocol-level rejection.** It trips a bare `assert` inside
   `encode_basic_data` during root computation, so it crashes the transition tool instead of
   invalidating the block. The EIP states no cap. Tracked in #3255.

#### Known gaps

- **`valid_at`-pinned tests are silently excluded from the port.** It is an exact-set marker,
  so a handful of tests never run against this fork and nothing warns.  See #3260 

### Related Issues or PRs

Builds on #3216.

Open questions this PR surfaced, each filed so they can be resolved independently:

- #3253 — account deletion diverges between the providers, visible through EIP-7610
- #3255 — state-root computation raises raw Python errors instead of protocol failures
- #3251 — genesis state-root computation is coupled to the MPT across the tooling
- https://github.com/misilva73/pbt-planning/issues/3
- https://github.com/misilva73/pbt-planning/issues/2

### Checklist

Verified on this branch: `pytest tests/binary_trie` 139 passed; `fill --fork BinaryTree
tests/binary_tree` 198 passed; framework suite 1814 passed; ruff check + format, mypy,
codespell, actionlint, vulture, `ethereum-spec-lint`, `uv lock --check` all clean.

### Cute Animal Picture


.github/workflows/test.yaml
```diff
@@ -181,6 +181,19 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          submodules: recursive
+      - uses: ./.github/actions/setup-uv
+      - name: Run fill-binary-tree tests
+        run: just fill-binary-tree tests/binary_tree
+        env:
+          PYTEST_XDIST_AUTO_NUM_WORKERS: auto
+
```
kevaundray: Ah something I do is use trie and tree interchangeably, but it might be confusing -- perhaps it makes sense to be explicit and write `binary-trie-unit-test` and `binary-trie-fork` -- in the future, the latter will be I*

docs/filling_tests/filling_tests_dev_fork.md
```diff
@@ -32,6 +32,51 @@ By default, the execution-testing framework only generates fixtures for forks th
      uv run fill --fork=Cancun --evm-bin=/opt/bin/evm -v
      ```
 
+## The Experimental `BinaryTree` Fork (EIP-8297)
+
+`BinaryTree` is Amsterdam semantics with its state commitment replaced by the [EIP-8297](https://eips.ethereum.org/EIPS/eip-8297) Partitioned Binary Tree instead of the Merkle Patricia Trie. Like any fork under development it is registered `deployed=False`, so it is excluded from every default fill and from every `--until`/`--from` fork range; it is reached only by naming it explicitly with `--fork BinaryTree`.
+
+### How to Fill It
+
+`BinaryTree` cannot use the flags above the usual way, because two things collide: the `fill` recipe in the `Justfile` always appends `--until Amsterdam` to whatever arguments it's given, and the framework rejects `--fork` combined with `--from`/`--until` on the same invocation. `just fill --fork BinaryTree` therefore cannot work, so a dedicated `fill-binary-tree` recipe exists that is the same shape with `--until` simply omitted:
```
kevaundray: Do you think we need to change this file?

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
```diff
@@ -63,10 +63,20 @@ def get_test_count(self) -> int:
...
+        # Deferred import: `execution_testing.client_clis` transitively
+        # imports `execution_testing.fixtures` (this package) at module
+        # load time (e.g. via `fixture_consumer_tool`), so importing
+        # `spec_calc_state_root` here at module level would form an
+        # import cycle; importing inside the method avoids it.
+        from execution_testing.client_clis.transition_tool import (
+            spec_calc_state_root,
+        )
+
+        state_root = spec_calc_state_root(
```
kevaundray, line 71: I found that on the CL, there is a coupling to the MPT because the CL hardcodes a EL block -- I'm wondering if doing it this way will make us need to change more tooling

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
```diff
@@ -556,7 +566,23 @@ class GroupPreAlloc(Alloc):
...
+        practice, because `PreAllocGroup.model_post_init` unconditionally
+        seeds `_cached_state_root` from `genesis.state_root` right after
+        construction -- both a fresh `build()` and every `from_file()`
+        reload re-run it -- and `calculate_genesis()` makes
+        `genesis.state_root` fork-correct via `spec_calc_state_root`.
+        The state-provider name that `spec_calc_state_root` installs is
+        itself a private attr and would not survive a
+        `model_dump`/`model_validate` round trip, which is why this
+        cache, not the provider, is what carries the fork-correct root
+        across a reload.
```
kevaundray: do we think this is needed?

tests/binary_trie/test_differential_mpt.py
```diff
@@ -0,0 +1,502 @@
+"""
```
kevaundray, line 1: Noting for future: I also added unit tests, in the future (not in this PR), we ideally want to ensure that the unit tests are captured in the fill tests

tests/binary_tree/eip8297_partitioned_binary_tree/eip_checklist_external_coverage.txt
kevaundray, line 1: delete?

tests/binary_tree/eip8297_partitioned_binary_tree/eip_checklist_not_applicable.txt
kevaundray, line 1: delete?

kevaundray (approved):
Left some comments :)

.github/workflows/test.yaml
CPerezz: Yes. I fully agree. Otherwise there are even behaviours we might change and this makes it a lot more clear.

docs/filling_tests/filling_tests_dev_fork.md
CPerezz: This is AI slop. You reviewed prior to me trimming all the crap docs and verbose comments. So yes. I agree

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
CPerezz: More slop. As said will trim afterwards. I'm concerned now about correctness mostly

tests/binary_trie/test_differential_mpt.py
CPerezz: 100% agree. But we need clients to consume and testa gain right?

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
CPerezz: Good catch, and I don't think it blocks this PR — but you're right.

What's fixed here is narrow: `calculate_genesis()` called `Alloc.state_root()` directly, and since the provider defaults to `state_mpt`, `--generate-all-formats` fixtures for `BinaryTree` carried a genesis root no client could reproduce. Routing it through `spec_calc_state_root` just brings it in line with what `specs/blockchain.py` already did.

Filed #3251 to capture it: enumerate every genesis-root computation across fill/execute/hive, establish what the CL actually assumes, then decide between per-site threading, a single provider-aware genesis builder, or something upstream of both. I'll start that investigation once this PR is settled.

.github/workflows/test.yaml
CPerezz: Done — jobs are now `binary-trie-unit-test` and `binary-trie-fork`, and the `just` recipes were renamed to match so the command you type is the check you see.

PR #3248, opened 2026-07-28
feat: change StatelessInput SSZ serialization to be EIP-7688 aligned

jsign:
### Description
Align Amsterdam’s zkEVM stateless SSZ types with the progressive SSZ definitions adopted by Gloas in EIP-7688.

This changes `SszExecutionPayload` and `SszExecutionRequests` to `ProgressiveContainer` and converts their dynamically sized fields—including transactions, withdrawals, execution-request lists, block access lists, and versioned hashes—to progressive types.

Progressive types retain the existing SSZ byte encoding but use different Merkleization. Therefore:

- `statelessInputBytes` remain compatible and keep schema ID `0x1501`.
- `new_payload_request_root` intentionally changes to the progressive SSZ root.
- No schema revision bump is required.

Fixture generation now verifies that every decodable Amsterdam stateless input produces the root stored in its stateless output. The zero-root sentinel remains valid only when the input cannot be decoded.

The fixed SSZ request-list limits are also removed, allowing the existing `MAX+1` and `MAX+2` builder deposit/exit fixtures to include stateless artifacts instead of skipping stateless validation.

Reference: https://github.com/ethereum/consensus-specs/pull/4630

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
@@ -1357,16 +1357,31 @@ def test_execution_witness_expected_true_reuses_canonical_stateless_result(
     fixture = next(iter(fixture_data.values()))
     block = fixture["blocks"][-1]
 
+    from ethereum.crypto.hash import Hash32
+    from ethereum.forks.amsterdam.stateless import (
+        compute_new_payload_request_root,
+    )
+    from ethereum.forks.amsterdam.stateless_guest import (
+        deserialize_stateless_input,
+    )
     from ethereum.forks.amsterdam.stateless_host import (
         deserialize_stateless_output,
     )
     from ethereum_types.bytes import Bytes as EthereumBytes
 
+    stateless_input = deserialize_stateless_input(
+        EthereumBytes(bytes.fromhex(block["statelessInputBytes"][2:]))
+    )
     stateless_output = deserialize_stateless_output(
         EthereumBytes(bytes.fromhex(block["statelessOutputBytes"][2:]))
     )
 
     assert stateless_output.successful_validation is True
+    assert stateless_output.new_payload_request_root != Hash32(b"\0" * 32)
+    assert (
+        stateless_output.new_payload_request_root
+        == compute_new_payload_request_root(stateless_input)
+    )
 
 
 def test_execution_witness_soundness_rewrites_stateless_fixture_bytes(
@@ -1402,6 +1417,10 @@ def test_execution_witness_soundness_rewrites_stateless_fixture_bytes(
 
     assert len(block["executionWitness"]["headers"]) == 1
 
+    from ethereum.crypto.hash import Hash32
+    from ethereum.forks.amsterdam.stateless import (
+        compute_new_payload_request_root,
+    )
     from ethereum.forks.amsterdam.stateless_guest import (
         deserialize_stateless_input,
     )
@@ -1418,6 +1437,11 @@ def test_execution_witness_soundness_rewrites_stateless_fixture_bytes(
     )
 
     assert stateless_output.successful_validation is False
+    assert stateless_output.new_payload_request_root != Hash32(b"\0" * 32)
+    assert (
+        stateless_output.new_payload_request_root
+        == compute_new_payload_request_root(stateless_input)
+    )
     assert len(stateless_input.witness.headers) == 1
     assert [
         "0x" + bytes(header).hex()
--- a/packages/testing/src/execution_testing/specs/blockchain_stateless.py
+++ b/packages/testing/src/execution_testing/specs/blockchain_stateless.py
@@ -443,6 +443,14 @@ def finalize_stateless_artifacts(
             f"want {options.expected_validation_success}"
         )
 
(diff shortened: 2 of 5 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filler.py
```diff
@@ -1357,16 +1357,31 @@ def test_execution_witness_expected_true_reuses_canonical_stateless_result(
...
+    stateless_input = deserialize_stateless_input(
+        EthereumBytes(bytes.fromhex(block["statelessInputBytes"][2:]))
+    )
     stateless_output = deserialize_stateless_output(
         EthereumBytes(bytes.fromhex(block["statelessOutputBytes"][2:]))
     )
 
     assert stateless_output.successful_validation is True
+    assert stateless_output.new_payload_request_root != Hash32(b"\0" * 32)
+    assert (
```
jsign, line 1381: Mostly added coherency check between private input and public output.

packages/testing/src/execution_testing/specs/blockchain_stateless.py
```diff
@@ -443,6 +443,14 @@ def finalize_stateless_artifacts(
             f"want {options.expected_validation_success}"
         )
 
+    assert_amsterdam_stateless_output_request_root(
```
jsign, line 446: Ditto but assert at filling time.

src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -231,9 +231,6 @@ def compute_new_payload_request_root(
 ) -> Hash32:
     """
     Compute the request root for a stateless input via SSZ hash tree root.
-
-    TODO: For readability, convert to NewPayloadRequestHeader and
-    then compute the payload request root.
```
jsign, line 236: Removing this since I'm not sure still makes sense -- in any case both things are equivalent. 

If someone raises that doing this is better, we can do it.

src/ethereum/forks/amsterdam/stateless_ssz.py
```diff
@@ -41,19 +46,8 @@
 
 # --- SSZ max-length constants ---
 
-# Consensus-spec bounds mirrored by NewPayloadRequest and its nested
-# containers.  These should track the corresponding consensus-spec values.
 
 MAX_EXTRA_DATA_BYTES = 32
-MAX_BYTES_PER_TRANSACTION = 2**30
```
jsign, line 48: The gist of the PR.

src/ethereum/forks/amsterdam/stateless_ssz.py
jsign, line 60: I was thinking of probably moving these SSZ containers also be progressive. It would simplify a lot of rationale explanation about these bound constants.

But prefered to leave this idea to a separate PR, to keep things cleaner.

tests/amsterdam/eip8282_builder_execution_requests/test_modified_builder_contract.py
```diff
@@ -121,14 +121,12 @@ def run_modified_requests_test(
                 Spec.MAX_DEPOSIT_REQUESTS_PER_BLOCK + 1
             ),
             id="max_plus_1_builder_deposit_requests",
-            marks=pytest.mark.skip_stateless_validation,
```
jsign, line 124: Nice sideeffect is that with progressive containers, we can avoid skipping tests that existed in the EL that violated previous list max-length rules.

src/ethereum/forks/amsterdam/stateless_ssz.py
```diff
@@ -41,19 +46,8 @@
...
-MAX_TRANSACTIONS_PER_PAYLOAD = 2**20
-MAX_WITHDRAWALS_PER_PAYLOAD = 2**4
```
kevaundray, line 50: Ah this reminded me of https://github.com/ethereum/execution-specs/pull/2362/changes/5855ed220e8389c681c7d6362bf7583570d0e712

kevaundray: Ah hmm seems it was fixed already

src/ethereum/forks/amsterdam/stateless.py
kevaundray: agree it makes sense to remove -- I think this comment was added initially because we kept talking the fact that there is an implicit transformation to `NewPayloadRequestHeader` from `NewPayloadRequest`

kevaundray (approved):
Makes sense -- looks a lot cleaner without all of these ssz upper bounded constants

jsign:
> Makes sense -- looks a lot cleaner without all of these ssz upper bounded constants

@kevaundray, yep. After creating this PR, I cherry-picked-it into latest `tests-zkevm@v0.6.2` and filled some fixtures and shared in discord thread... deme from Nimbus is going to do some quick checking if these change in `new_payload_request_root` looks okay.

After having some greenish check, I'll merge.

kdeme (approved):
lgtm. 

Tested with nimbus stateless implementation on branch https://github.com/status-im/nimbus-eth1/pull/4585, fixtures pass with root comparison enabled.

Issue #3244, opened 2026-07-27
refactor(ci): Decouple Benchmarking from Mainline Development


marioevz:
## Summary

Benchmark CI currently couples an **external, spec-matched geth dependency** into checks that run on (nearly) every PR. When the EELS spec moves ahead of the pinned geth branch, unrelated spec PRs fail the benchmark job — see #3214, where EELS advanced to devnet-8 while geth is still on `glamsterdam-devnet-7`.

This issue proposes splitting benchmarking into two clearly separated tracks:

1. **A lightweight, EELS-only smoke test** that runs on PRs and never depends on geth.
2. **A geth-backed benchmark release**, run from a dedicated, deliberately-pinned `benchmarking` branch, that produces and publishes benchmark fixtures.

Mainline development (`forks/amsterdam`) never stops for benchmarking; benchmarking advances its spec baseline only by a deliberate manual decision.

## Current state (diagnosis)

`.github/workflows/benchmark.yaml` has three job groups:

| Job | Recipe | Runs on | geth? | Notes |
|-----|--------|---------|-------|-------|
| `unit-tests` (Benchmark Unit Tests) | `test-tests-bench` | push + PR | builds it, unused | plugin unit test; passes regardless |
| `sanity-checks` → Benchmark Gas Values | `bench-gas` | push + PR | **yes** | fills with geth, **re-verifies vs EELS** → **fails on spec mismatch** |
| `sanity-checks` → Fixed Opcode Count CLI/Config | `bench-opcode`, `bench-opcode-config` | push + PR | yes | fill with geth, no EELS cross-check → currently pass |
| `build-artifact` (Build Benchmark Fixture Artifact) | `build-fixtures` (`benchmark_fast`) | push only (`if: github.event_name == 'push'`) | no (`benchmark_fast` = EELS) | uploads **ephemeral** CI artifacts only |

Key facts:

- The failing check is **`bench-gas`** — the only job that cross-checks geth-filled fixtures against the EELS spec. geth (`.github/configs/evm.yaml` → `benchmark: geth @ glamsterdam-devnet-7`) and EELS (devnet-8) disagree → fail.
- **`build-artifact` already does not run on PRs** and does **not** use geth; it fills with EELS (`benchmark_fast`) and only uploads ephemeral artifacts. It publishes **no release**.
[...]

PR #3214, opened 2026-07-23
feat(specs,amsterdam): EIP-2780 - fold transfer log cost into value transfer cost

gurukamath:
### Description
Align the EIP-2780 implementation with the changes proposed in https://github.com/ethereum/EIPs/pull/11997

### Related Issues or PRs
#3205

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Cute Animals - 1 of 1

PR #3249, opened 2026-07-28
refactor(ci): run geth benchmark CI only on `benchmarks/**` branches

marioevz:
### Description

Benchmark CI previously coupled an **external, spec-matched geth dependency** into checks that ran on (nearly) every PR: `benchmark.yaml`'s `bench-gas` job fills fixtures with geth and re-verifies them against EELS, so whenever the EELS spec moved ahead of the pinned geth ref, unrelated spec PRs failed the benchmark job (see #3214: EELS on devnet-8 vs geth on `glamsterdam-devnet-7`).

This PR separates benchmarking so that mainline development is never blocked by an external client, while the geth-backed work moves to dedicated `benchmarks/**` branches:

- **`test_benchmarking.py` defaults to EELS.** The benchmark-plugin fills previously hardcoded `--evm-bin` (defaulting to geth's `evm`). They now omit `--evm-bin` when `EVM_BIN` is unset, falling through to `fill`'s in-repo EELS t8n; setting `EVM_BIN` still targets a specific binary. This removes the only reason those tests needed an external client.
- **Plugin unit tests fold into the normal unit-test suite.** The `--ignore=.../test_benchmarking.py` is removed from `just test-tests`, so the benchmark-plugin tests now run as part of the standard unit tests on every `forks/**` PR via `test.yaml` — on EELS, with no geth. The redundant `test-tests-bench` recipe is deleted.
- **`test.yaml` behaves differently on `benchmarks/**`.** On those branches it builds the pinned geth and sets `EVM_BIN`, so the same benchmark tests run against a real client; on every other branch the build step is skipped and the tests use EELS. (`EVM_BIN` is read only by `test_benchmarking.py`, so this has no effect on the rest of the suite.)
- **`benchmark.yaml` runs only on `benchmarks/**`.** It now holds the geth-backed sanity checks (`bench-gas` cross-check, `bench-opcode`, `bench-opcode-config`) and the fixture-artifact build. These branches pin a spec version together with a matching geth ref, so the geth↔EELS cross-check is expected to agree there.

Net effect: `forks/**` PRs get full benchmark-plugin coverage on EELS with no external dependency, and the geth cross-check/fills run where geth and the spec are deliberately kept in sync. This is the first step of a larger benchmarking separation; creating the `benchmarks/amsterdam` branch and the release/publish flow are follow-ups.

Files changed:

- `packages/testing/.../filler/tests/test_benchmarking.py` — EELS-by-default `_evm_bin_args()` helper.
- `.github/workflows/test.yaml` — add `benchmarks/**`; conditional geth build + `EVM_BIN` in `test-tests`.
- `.github/workflows/benchmark.yaml` — scope to `benchmarks/**`; drop `unit-tests` (now covered by `test.yaml`).
- `Justfile` — un-ignore the benchmark tests in `test-tests`; delete `test-tests-bench`.
- `docs/getting_started/verifying_changes.md` — update the benchmark row.

### Related Issues or PRs

#3244

### Checklist

### Cute Animal Picture


spencer-tb (approved):
LGTM!

cc @danceratopz for when you are back

PR #3250, opened 2026-07-28
chore(do not merge): add code to 8297 to show deletion logic

kevaundray:
### Description

This adds code for deletions in the trie

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/src/ethereum/binary_trie/trie.py
+++ b/src/ethereum/binary_trie/trie.py
@@ -257,6 +257,30 @@ def trie_get(trie: BinaryTrie, key: Key) -> Optional[Bytes32]:
     return trie._data.get(key)
 
 
+def remove_subtree(trie: BinaryTrie, prefix: Bytes) -> None:
+    """
+    Remove every key of `trie` beginning with `prefix`; a prefix
+    matching nothing does nothing.
+
+    Keys are consumed most significant bit first, so the keys sharing
+    a `prefix` are exactly the keys of one subtree, and this removes
+    that subtree whole. Callers reach for it when the set of keys to
+    remove is known by where it sits in the tree rather than by
+    enumeration; see [`remove_account`], which drops an account's
+    unbounded storage without being told which slots it holds.
+
+    A production client would unlink the subtree's node from its
+    parent and be done, which is why this is a tree operation and not
+    a loop of removals; the scan here follows [`BinaryTrie`] storing
+    only key/value pairs.
+
+    [`remove_account`]: ref:ethereum.binary_trie.embedding.remove_account
+    [`BinaryTrie`]: ref:ethereum.binary_trie.trie.BinaryTrie
+    """
+    for key in [key for key in trie._data if key.startswith(prefix)]:
+        del trie._data[key]
+
+
 def encode_bit_prefix(prefix: Bytes) -> Bytes:
     """
     Encodes a branch prefix: a two-byte big-endian bit
--- a/src/ethereum/binary_trie/trie.py
+++ b/src/ethereum/binary_trie/trie.py
@@ -214,18 +214,36 @@ def copy_trie(trie: BinaryTrie) -> BinaryTrie:
     return BinaryTrie(copy.copy(trie._data))
 
 
-def trie_set(trie: BinaryTrie, key: Key, value: Bytes32) -> None:
+def trie_set(trie: BinaryTrie, key: Key, value: Optional[Bytes32]) -> None:
     """
-    Insert or update `key` in `trie` with the given `value`.
+    Insert or update `key` in `trie` with the given `value`; setting
+    `None` removes the key, and removing an absent key does nothing.
+
+    `None` can mark absence because it lies outside the value space:
+    every 32-byte value, including all zeroes, is a legitimate leaf,
+    so no stored value could play the role the Merkle Patricia
+    Trie's default value does. The same convention marks deleted
+    accounts in [`BlockDiff`], and mirrors [`trie_get`], which
+    returns `None` for absent keys.
+
+    Since [`root`] rebuilds the node structure from the surviving
+    entries, a removal needs no node surgery here: branches held
+    open by the removed key simply never form, and the trie commits
+    as if the key had never been inserted.
 
     The caller must keep keys prefix-free; see [`Key`].
 
     [`Key`]: ref:ethereum.binary_trie.trie.Key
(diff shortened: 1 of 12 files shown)


src/ethereum/binary_trie/trie.py
```diff
@@ -214,18 +214,36 @@ def copy_trie(trie: BinaryTrie) -> BinaryTrie:
     return BinaryTrie(copy.copy(trie._data))
 
 
-def trie_set(trie: BinaryTrie, key: Key, value: Bytes32) -> None:
+def trie_set(trie: BinaryTrie, key: Key, value: Optional[Bytes32]) -> None:
```
kevaundray, line 217: value of None means delete

src/ethereum/state_pbt.py
```diff
@@ -165,33 +148,90 @@ def account_has_storage(self, address: Address) -> bool:
...
+
+def apply_diff_to_trie(
+    trie: BinaryTrie, pre_state: State, diff: BlockDiff
+) -> None:
+    """
+    Apply `diff` to `trie`, the embedding of `pre_state`, as tree
+    operations: writes become insertions or in-place updates, and
+    removals become deletions. This function decides *what* changed;
+    the embedding's operations decide which keys that touches.
+    """  # noqa: E501
```
CPerezz, line 201: `embed_flat_state` skips storage for addresses absent from accounts — an invariant its own docstring still states — and `apply_diff_to_trie` writes `storage_changes` unconditionally, dropping the enforcement while keeping the docstring.

Basically. `fork.py:341` uses `compute_state_root(block_diff)` for the header; `fork.py:275` uses `apply_changes_to_state` to advance the chain. 
On the base branch these were the same code path — `compute_state_root` literally called `apply_changes_to_state` on a copy — so they agreed by construction. This PR makes them two independent implementations, and they disagree.

We should have invariant unit tests or similar to make sure this doesn't go south.

src/ethereum/state_pbt.py
```diff
@@ -165,33 +148,90 @@ def account_has_storage(self, address: Address) -> bool:
...
+        address32 = address20_to_address32(address)
+        pre_account = pre_state._accounts.get(address)
+        if account is None:
+            if pre_account is None:
+                # Created and destroyed within the block; its leaves
+                # never reached the pre-state tree.
+                continue
+            assert pre_account.code_hash == EMPTY_CODE_HASH
+            assert address not in pre_state._storage
+            remove_account(trie, address32)
```
CPerezz, line 247: `remove_account` cannot remove storage, so the assert fires on real blocks.
Remove the account's storage leaves before removing the account, reusing the loop already present in the storage_clears branch, and drop the assert:
```
if account is None:
    if pre_account is None:
        continue                      # created and destroyed within the block
    for key in pre_state._storage.get(address, {}):
        remove_storage_slot(trie, address32, U256.from_be_bytes(key))
    remove_account(trie, address32)
    continue
```
No code-chunk removal is needed here: EIP-161 emptiness requires `code_hash == EMPTY_CODE_HASH`, and `chunkify_code(b"")`  is empty.

src/ethereum/state_pbt.py
```diff
@@ -165,33 +148,90 @@ def account_has_storage(self, address: Address) -> bool:
...
+    for address, account in diff.account_changes.items():
+        address32 = address20_to_address32(address)
+        pre_account = pre_state._accounts.get(address)
+        if account is None:
+            if pre_account is None:
+                # Created and destroyed within the block; its leaves
+                # never reached the pre-state tree.
+                continue
+            assert pre_account.code_hash == EMPTY_CODE_HASH
+            assert address not in pre_state._storage
```
CPerezz: This seems to be firing in real filled blocks??

If you copy `test_state_divergence.py` merged in https://github.com/ethereum/execution-specs/pull/3246 and fill it.
```
src/ethereum/state_pbt.py:199: in apply_diff_to_trie
    assert address not in pre_state._storage
E   AssertionError
FAILED test_create2_after_eip161_clear_of_storage_holding_account[…blockchain_test]
FAILED test_create2_after_eip161_clear_of_storage_holding_account[…blockchain_test_engine]
```

The diagnosis matters more than the crash. The base did _accounts.pop(); _storage.pop() — it dropped the account and its storage. remove_account removes only two leaves (basic data + code hash) and cannot remove storage, so the PR replaced working functionality with an assertion that the situation never arises. It does.

That assert is unsatisfiable, so the obvious fix doesn't work. My agent verified it: `BlockDiff(storage_clears={A}, account_changes={A: None})` correctly clearing storage before deleting — still trips it, because the assert consults the untouched pre_state while storage_clears mutates only the trie. 

Any fix must track removals against the trie, or take a deletion path that actually removes the storage leaves.

CPerezz (commented):
Drop the old-code lookup entirely. Header chunk keys ignore code_hash. See `get_tree_key_for_code_chunk` returns `get_tree_key_for_header(address, CODE_OFFSET + chunk_id)` below the boundary. And removing an absent key is already a no-op. So replace `remove_code_chunks(trie, addr, code_hash, code)` with an unconditional sweep of the header range:
```
for chunk_id in range(HEADER_CHUNK_COUNT):
    trie_set(trie, get_tree_key_for_code_chunk(address32, code_hash, Uint(chunk_id)), None)
```
That eliminates `code_for(pre_account.code_hash)`, the `KeyError`, the new pre-state requirement on historical bytecode, and a redundant chunkify_code of the whole old code. 

As for the findings related to the invariant:
Assert the chain invariant.
```
header_root = pre.compute_state_root(diff)
clone = copy_of(pre)
apply_changes_to_state(clone, diff)
assert header_root == state_root(clone)
```


Also, run `ruff format --check` to fix the issue.

CPerezz (commented):
- spec.py still pins the superseded EIP. cc08f960 should be 9be51fb828c7ed32ed4337c66cd55f1c55c14099 instead.

- The agent found:
> Two files still carry the inverted disclosure. test_state_pbt.py was updated correctly — it now says the EIP "requires what this module's State does… conformance, not merely pinned current behavior." But test_differential_mpt.py still says "Both providers here do the opposite", and the EEST test_storage_ops.py still says the tests pin "not EIP-8297 conformance" and would "need regenerating if the provider were ever made conformant." Both now state the reverse of the truth. 

- `test_storage_ops.py` hasn't been edited at all. Is this showcasing maybe we should be having test there? And if not, why are there mentions to PBT-related stuff there?

PR #3262, opened 2026-07-29
feat(ci): Backport CI Workflow

marioevz:
### Description

Adds a label-driven backport workflow so a merged PR can be cherry-picked onto another long-lived branch automatically (e.g. `benchmarks/amsterdam`), without blocking mainline development.

`.github/workflows/backport.yaml` does this:

- **Selective, via a label.** The workflow acts only on PRs carrying a `backport <branch>` label (e.g. `backport benchmarks/amsterdam`); the target branch is read from the label itself. PRs without such a label are untouched.
- **Fires on merge.** Triggered on `pull_request_target` (`closed`, `labeled`) and gated by `merged == true`, so the label can be applied before *or* after merge but a backport only happens once the PR is actually merged.
- **Fails safe on conflict.** On a clean cherry-pick it opens a PR against the target branch titled `[backport <branch>] <original title>`. On conflict it aborts and comments the manual cherry-pick steps on the source PR — it never force-pushes and never leaves a partially-applied branch behind. Since `benchmarks/amsterdam` is deliberately divergent, conflicts are expected to be common and are handed to a human by design.
- **Skips non-squash merges.** `merge_commits: skip` means the occasional PR that landed as a merge commit is skipped rather than failing the run; the repo squash-merges, so each PR is a single, cleanly-pickable commit.

**Operational prerequisites** (out of band, not in this PR):

- Create the `backport benchmarks/amsterdam` label in the repo. Without a matching label the workflow is a no-op.
- Optionally add a `BACKPORT_TOKEN` secret (a PAT or GitHub App token with `contents:write` + `pull_requests:write`). The workflow uses it when present and falls back to `GITHUB_TOKEN` otherwise. This matters because a PR opened with the default `GITHUB_TOKEN` cannot trigger downstream CI, so backport PRs to the divergent `benchmarks/amsterdam` branch would otherwise land without test runs until re-triggered manually.

Usage: add the `backport benchmarks/amsterdam` label to a PR (e.g. `gh pr edit <num> --add-label "backport benchmarks/amsterdam"`); on merge the bot opens the backport PR or comments if it cannot apply cleanly.

### Related Issues or PRs
N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -1,0 +1,51 @@
+name: Backport merged PR
+
+# Cherry-picks a merged PR onto another long-lived branch when it carries a
+# `backport <branch>` label (e.g. `backport benchmarks/amsterdam`). On a clean
+# cherry-pick it opens a PR against the target branch; on conflict it aborts
+# and comments on the source PR with manual steps — it never force-pushes nor
+# leaves a partially-applied branch behind.
+#
+# The label may be added before or after merge; backporting only happens once
+# the PR is actually merged.
+
+on:
+  pull_request_target:
+    types: [closed, labeled]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  backport:
+    name: Backport
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.BACKPORT_TOKEN || secrets.GITHUB_TOKEN }}
+
+      - name: Create backport PR
+        uses: korthout/backport-action@2e830a1d0b8269505846ddd407a70876913ad1f8 # v4.6.0
+        with:
+          # `backport <target-branch>`, e.g. `backport benchmarks/amsterdam`.
+          label_pattern: '^backport ([^ ]+)$'
+          # Only single-commit (squash-merged) PRs cherry-pick cleanly; skip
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yaml
```diff
@@ -0,0 +1,51 @@
...
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  backport:
+    name: Backport
+    if: github.event.pull_request.merged == true
```
LouisTsai-Csie: Nit: This CI fetches the EELS repo on every merged PR. It should stop early if there's no `backport` label instead.

Suggested filter:

```python
if: >-
  github.event.pull_request.merged == true &&
  contains(join(github.event.pull_request.labels.*.name, ','), 'backport ')
```

Follow the example in the action repo: https://github.com/korthout/backport-action#trigger-using-a-comment

.github/workflows/backport.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Create backport PR
+        uses: korthout/backport-action@2e830a1d0b8269505846ddd407a70876913ad1f8 # v4.6.0
+        with:
+          # `backport <target-branch>`, e.g. `backport benchmarks/amsterdam`.
+          label_pattern: '^backport ([^ ]+)$'
+          # Only single-commit (squash-merged) PRs cherry-pick cleanly; skip
+          # PRs that landed as merge commits rather than failing the run.
+          merge_commits: skip
+          pull_title: '[backport ${target_branch}] ${pull_title}'
```
LouisTsai-Csie: What about `${pull_title} backports to ${target_branch}` to follow our PR pattern?

LouisTsai-Csie (approved):
When creating this PR, there would be no label an author, please consider:

```yaml
add_author_as_assignee: true
copy_labels_pattern: '^[AC]-'
```

.github/workflows/backport.yaml
marioevz: Did `${pull_title} [backport ${target_branch}]`. If it gets too confusing we can change it later.

PR #3265, opened 2026-07-30
fix(tests,test-fill): fix Engine X fills on BAL forks and rebuild the drift check

spencer-tb:
### Description

Combined PR: the Engine X drift-check fixes previously tracked as #3219, plus the rebuild of `engine_x_checks.py` per the [review feedback](https://github.com/ethereum/execution-specs/pull/3219#discussion_r3660230116) there.

**Fix the false drift failures on BAL forks** (was #3219). The `tests-glamsterdam-devnet@v7.2.1` release fill failed with `EngineXExecutionDriftError` on 18049 of 24683 Engine X fixtures: on BAL forks the EIP-2935 system call writes the parent hash (the group genesis hash at payload 0) into every payload's `blockAccessList`, an expected state-root-derived difference. Mask that write instead of comparing it verbatim, keeping the rest of the BAL in the comparison so leaked accounts are still caught. Two tests needed isolation: `test_gas_cost_return` blind-calls undeclared `0x1000`/`0x2000` that packing can populate (isolated as in #3176), and the `test_bal_2935_query[...valid_block_number]` params SSTORE the queried genesis hash where a per-payload mask cannot reach. The drift check runs only on a clean session and only when optimistic grouping is enabled.

**Rebuild the check on typed fixture models** (the review feedback): fixtures were compared as black-box JSON dicts, the parent-hash mask was a byte-pattern match over the whole RLP tree, and drift failures gave no indication of cause.

- Parse both fixture trees with `BlockchainEngineXFixture`/`BlockchainEngineFixture`; a fixture that fails typed validation is a loud error naming file and test id, never a silent degradation of the check.
- Decode BALs with `BlockAccessList.from_rlp` and mask only the EIP-2935 history-contract write whose value equals the payload's own parent hash (integer comparison, so the leading-zero trimming special case disappears). A parent-hash-valued write to any other account is drift again, and there is a regression test for it.
- Classify and aggregate drifts by cause: a leaked account is traced through the sibling fixture's `pre` and the pre-alloc group file (via the typed `AllocGroupHash` from #3390) to the address, the group, and the tests that packed it; identical causes collapse into one diagnosis instead of one failure per fixture.
- `verify_engine_x_execution` always returns a result with an explicit `skip_reason`, simplifying the `pytest_sessionfinish` caller; the skip log now names its cause (unclean session vs grouping disabled).
- The unit tests build real serialized fixtures and pre-alloc group files via the models instead of hand-rolled dicts, and guard the field names the check depends on so model renames fail in CI.

Example failure output after injecting a leak into a real fill:

```
1 of 13 Engine X fixtures execute differently against their packed pre-allocation group's genesis than against their own pre-allocation (1 distinct cause(s)):

[1x] payload 0: account 0xf6c3...b3ff appears in the packed fixture's BAL only
  tests: tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip2935.py::test_bal_2935_empty_block[fork_Amsterdam-blockchain_test_engine_x]
  the account is absent from the test's own pre-allocation but present in pre-alloc group 0xa58f4980b4cd5258 (11 tests, e.g. ...): an account introduced by packing leaked into this test's execution. Isolate the test with @pytest.mark.pre_alloc_group("separate") or declare the account in its pre-allocation
```

Validated on this head: full `tests/amsterdam/eip7928_block_level_access_lists` fill at Amsterdam with `--generate-all-formats` (3118 passed; drift check green on all 1036 Engine X fixtures, including the corrupted-BAL negative tests), and a full `tests/ported_static` fill at Amsterdam with `--generate-all-formats` (115304 passed, 196 skipped, 0 failed; drift check green on all 27538 Engine X fixtures across 2178 pre-alloc groups).

### Related Issues or PRs

Supersedes #3219 (its commits are included here). Addresses marioevz's `engine_x_checks.py` review comments on #3219. Fixes the `fill (amsterdam)` failure in the [glamsterdam-devnet@v7.2.1 release dispatch](https://github.com/ethereum/execution-specs/actions/runs/29907017961/job/88880830188), same failure class as the Benchmarking CI red since #3187.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -2160,34 +2160,28 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
         _log_timing("verify_engine_x_execution: starting...")
         t0 = time.time()
         engine_x_check = verify_engine_x_execution(fixture_output.directory)
-        engine_x_warning: str | None = None
-        if engine_x_check is not None:
-            if engine_x_check.compared > 0:
-                logger.info(engine_x_check.summary)
-            elif engine_x_check.skipped > 0:
-                engine_x_warning = (
-                    "Engine X execution consistency check skipped: none of "
-                    f"the {engine_x_check.skipped} Engine X fixtures have a "
-                    "blockchain_tests_engine sibling fixture to compare "
-                    "against. Leaks from pre-alloc group packing are not "
-                    "verified for this output."
-                )
-        elif (fixture_output.directory / ENGINE_X_FIXTURES_DIR).is_dir():
-            engine_x_warning = (
-                "Engine X execution consistency check skipped: this fill "
-                "generated no blockchain_tests_engine fixtures to compare "
-                "against (e.g. filling with `-m blockchain_test_engine_x`). "
-                "Leaks from pre-alloc group packing are not verified for this "
-                "output."
-            )
-        if engine_x_warning is not None:
-            logger.warning(engine_x_warning)
-            # Repeated in the terminal summary; a log line alone is easy to
-            # miss.
-            session.config.engine_x_check_warning = engine_x_warning  # type: ignore[attr-defined] # noqa: E501
+        if engine_x_check.compared > 0:
+            logger.info(engine_x_check.summary)
+        if engine_x_check.skip_reason is not None:
+            logger.warning(engine_x_check.skip_reason)
+            # Repeated in the terminal summary; a log line alone is easy
+            # to miss.
+            session.config.engine_x_check_warning = engine_x_check.skip_reason  # type: ignore[attr-defined] # noqa: E501
         _log_timing(
             f"verify_engine_x_execution: done in {time.time() - t0:.1f}s"
         )
+    elif (fixture_output.directory / ENGINE_X_FIXTURES_DIR).is_dir():
+        if exitstatus != pytest.ExitCode.OK:
+            logger.info(
+                "Engine X execution consistency check skipped: the session "
+                "did not exit cleanly."
+            )
+        else:
+            logger.info(
+                "Engine X execution consistency check skipped: optimistic "
+                "pre-alloc grouping is disabled, so packing cannot have "
+                "changed any test's execution."
+            )
 
     # Verify fixtures after merge if verification is enabled
     if session.config.getoption("verify_fixtures"):
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
@@ -54,6 +54,8 @@ from execution_testing.fixtures import (
 )
 from execution_testing.fixtures.engine_x_checks import (
     ENGINE_X_FIXTURES_DIR,
+    EngineXCheckError,
+    EngineXExecutionDriftError,
     verify_engine_x_execution,
 )
 from execution_testing.fixtures.pre_alloc_groups import (
@@ -963,6 +965,16 @@ def pytest_terminal_summary(
             yellow=True,
         )
         terminalreporter.write_line(engine_x_warning, yellow=True)
+
+    engine_x_error = getattr(config, "engine_x_check_error", None)
+    if engine_x_error is not None:
+        title = (
+            " ERROR: Engine X execution drift "
+            if isinstance(engine_x_error, EngineXExecutionDriftError)
+            else " ERROR: Engine X execution consistency check failed "
+        )
+        terminalreporter.write_sep("=", title, bold=True, red=True)
+        terminalreporter.write_line(str(engine_x_error), red=True)
 
 
 def _aggregate_cache_stats(node: Any) -> None:
@@ -2033,6 +2045,7 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
     - Generate index file for all produced fixtures.
     - Create tarball of the output directory if the output is a tarball.
     """
+    del exitstatus
     logger = logging.getLogger("fill.sessionfinish")
     is_worker = xdist.is_xdist_worker(session)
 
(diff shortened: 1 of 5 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -2151,40 +2149,39 @@ def _log_timing(msg: str) -> None:
...
+    # test's execution (raises on drift, like a pre-alloc collision). Only
+    # checked on an otherwise clean session: raising from this hook aborts
+    # the terminal FAILURES/short-summary sections, so it would hide any
+    # test failures (which already fail the fill and must surface first).
+    if exitstatus == pytest.ExitCode.OK and not session.config.getoption(
+        "optimistic_pre_alloc_grouping_disabled"
+    ):
         _log_timing("verify_engine_x_execution: starting...")
         t0 = time.time()
         engine_x_check = verify_engine_x_execution(fixture_output.directory)
```
danceratopz: Addressed in spencer-tb/execution-specs#62 (commit danceratopz/execution-specs@fc2495526614ba00ac563f6013c9e93fb3549a92).

A design question, fine as a follow-up. This gate can suppress the drift diagnosis exactly when it is needed most.

Concrete example, using the test this PR isolates: without the isolation, `test_gas_cost_return` packed next to a test that deploys code at `0x1000` executes differently against the group genesis, its post-state check fails, and the developer sees a plain storage mismatch that looks like a spec or test bug. That failure makes the session unclean, so this gate skips the check, and the one report that explains the mismatch (account `0x1000` leaked in from the group, isolate the test or declare the account) is never produced. The failure disables the explainer, and the explainer is what you need to fix the failure. A smaller side effect: any unrelated failure in the session also turns the check off for the whole run, so a silent leak elsewhere goes unreported until a later clean run.

The gate is forced by the delivery mechanism rather than by the check itself: raising from `pytest_sessionfinish` aborts the terminal reporter before the FAILURES section prints, and it also skips the rest of this hook, so a drift failure leaves the output directory without the merged index and without the tarball.

The two can be decoupled: always run the check, catch `EngineXCheckError` here, render the report from `pytest_terminal_summary` next to the existing skip warning, and fail the fill by setting `session.exitstatus` (pytest returns `session.exitstatus` from `wrap_session`, so the mutation is honored). The exitstatus gate then goes away.

Validated with a synthetic drift injection: the drift report renders after the summary, the fill exits 1, and `.meta/index.json` is still written. Also green under xdist, and `just static` passes.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -2151,40 +2149,39 @@ def _log_timing(msg: str) -> None:
...
+            logger.info(
+                "Engine X execution consistency check skipped: the session "
+                "did not exit cleanly."
+            )
+        else:
+            logger.info(
+                "Engine X execution consistency check skipped: optimistic "
+                "pre-alloc grouping is disabled, so packing cannot have "
+                "changed any test's execution."
+            )
```
danceratopz: Addressed in spencer-tb/execution-specs#62 (commit danceratopz/execution-specs@f3fbfb1a9a1293b4b207da7abaa4ab21a624dc10).

Small nit. The comment as-is is not strictly correct.

```suggestion
            logger.info(
                "Engine X execution consistency check skipped: optimistic "
                "pre-alloc grouping is disabled."
            )
```

packages/testing/src/execution_testing/fixtures/engine_x_checks.py
```diff
@@ -63,104 +143,454 @@ def summary(self) -> str:
...
+    history contract on every block; at payload 0 that value is the
+    genesis hash itself, so it is the one BAL entry that legitimately
+    differs between a test's own genesis and its packed group's genesis.
+    Only that write is masked: a storage change of the history contract
+    whose written value equals the payload's own parent hash. A
+    parent-hash-valued write to any other account is genuine drift (the
+    test observes block hashes and cannot survive grouping). An
+    undecodable BAL (an intentionally malformed one from a negative
+    test) is compared verbatim.
+    """
```
danceratopz, line 249: Addressed in spencer-tb/execution-specs#62 (commit danceratopz/execution-specs@ea61aa95b669d160a8dcdaa9991059f88d316901).

A latent edge case in the mask. If it ever triggers it fails loudly, so the code can stay as-is; the suggestion below just records the limitation in the docstring. Background so the suggestion makes sense:

The mask trusts the payload's `parent_hash` field, and there is one case where that field lies. When the framework builds a block, t8n executes the transactions, and on BAL forks that execution records every storage write into the block access list, including the EIP-2935 system write of the true parent hash. Some invalid-block tests then deliberately corrupt the header afterwards with `Block.rlp_modifier=Header(parent_hash=...)` to produce a block clients must reject. The override edits only the header; nothing is re-executed, so the BAL keeps the true parent hash while the header carries the injected value.

For such a test the mask asks "does this history-contract write equal the payload's `parent_hash`?" and the answer is no on both sides: the field holds the injected value, the writes hold each side's real parent hash, and those legitimately differ between the own genesis and the packed genesis. The mask matches nothing, the unmasked writes differ, and the check reports drift that packing did not cause.

No test in the tree overrides `parent_hash` today (the BAL fork tests only override `block_access_list_hash` and `slot_number`), so nothing breaks now. If such a test ever lands, the writer can isolate it, or the mask can derive the expected value from the chain itself, which an override cannot touch.

```suggestion
    undecodable BAL (an intentionally malformed one from a negative
    test) is compared verbatim.

    Known limitation: a header parent hash injected with
    `Block.rlp_modifier=Header(parent_hash=...)` defeats the mask. The
    override is applied after t8n has executed the block, so the BAL
    keeps the real parent hash while the payload field holds the
    injected value; the mask matches neither, and the check fails
    loudly on a difference that packing did not cause. No test does
    this today. If one appears, isolate it with
    `@pytest.mark.pre_alloc_group("separate")`, or make the mask derive
    the expected value from the chain itself (the genesis hash at
    payload 0, the previous payload's block hash after that) instead of
    the payload's own `parent_hash` field.
    """
```

danceratopz (commented):
Thanks @spencer-tb! Reviewed together with a full local validation from the artificial intelligences. 

With masking disabled, 1005 of 1036 Engine X fixtures from a local Amsterdam BAL fill drift, in exactly the two expected signatures (payload 0 slot `0x00`, and payload 1 slot `0x01` on the fork transition tests). With it enabled, the same fill and a ported_static fill are drift-green end to end.

The narrower mask is a real improvement over the #3219 byte-pattern approach, and the typed rebuild addresses the review feedback there well.

I added one reporting improvement, a small string and a docstring suggestion here:
- https://github.com/spencer-tb/execution-specs/pull/62

spencer-tb:
Thanks! Added all suggestions :)

danceratopz (approved):
Thanks!

PR #3267, opened 2026-07-30
feat(test-benchmark): add missing glam eip to benchmark

LouisTsai-Csie:
### Description

Add EIP-8024, EIP-7843 to benchmark test suite.

### Related Issues or PRs

N/A.

### Checklist

- [ ] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [ ] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3273, opened 2026-07-31
feat(ci): fill benchmark tests at 1M gas on every PR

spencer-tb:
### Description

Since #3249 scoped `benchmark.yaml` to `benchmarks/**`, PRs into `forks/amsterdam` no longer fill benchmark tests at all, the regular `fill` job hides `tests/benchmark/` from collection, so a PR like #3267 merges with no benchmark fill coverage.

#### Main Addition

This PR adds an EELS-only benchmark fill to the PR CI:
- `Justfile`: new `fill-benchmark` recipe, fills `tests/benchmark/compute` with `--gas-benchmark-values 1` using the EELS t8n.
- `test.yaml`: new `fill-benchmark` job running the recipe on every PR/push, alongside the existing `fill` matrix.

Note: `test_return_revert` is excluded from the recipe. Its callees are padded to max contract size with `INVALID`, so EELS rescans jumpdests on every call (100 to 270s per test). The geth pipeline still fills it. With the exclusion the recipe fills 1102 tests in ~2.4 min locally.

#### Bug Fix

Running this surfaced a `CountTracer` bug: it reset its counts whenever `evm.message.tx_env` changed, so the block-end system calls (EIP-7002/7251) wiped the user-transaction counts and every test asserting `expected_opcode_count` failed with `got 0` (38 tests, all `tests/benchmark/compute/precompile/`).

The tracer now accumulates over the whole t8n run, matching geth's semantics and the tests' expected counts, which sum over all transactions in the block; system-call opcodes land well inside the ±5% tolerance and never include the targeted opcodes.

### Related Issues or PRs

Follow-up to #3249; gap surfaced by #3267.

### Checklist

### Cute Animal Picture


.github/workflows/test.yaml
```diff
@@ -110,6 +110,19 @@ jobs:
...
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  fill-benchmark:
+    runs-on: [self-hosted-ghr, size-xl-x64]
+    needs: static
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: ./.github/actions/setup-uv
+        with:
+          python-version: "3.14"
```
marioevz, line 120: We have `DEFAULT_PYTHON_VERSION` CI variable set to 3.12 at the moment, perhaps in a follow up PR we could evaluate if we could use that variable here instead?

marioevz (approved):
Awesome, thanks for this! Just one informational comment that does not block merging.

PR #3278, opened 2026-07-31
feat: update stateless validation to include schema fork index and refactor chain config handling

jsign:
### Description

This PR does a couple of things, but let's first focus on the important change that we discussed with @han0110 out of band.

This is the new `StatelessValidationResult`:

```python
class StatelessValidationResult:
    new_payload_request_root: Hash32
    successful_validation: bool
    chain_id: U64  # `ChainConfig` is gone; only `chain_id` remains.
    schema_id: U16  # New public field containing the full schema ID.
```

Let's start with why `schema_id` was added to `StatelessValidationResult`.

A while ago, we removed `protocol_fork: ProtocolFork` from `ChainConfig`. The rationale was that the two-byte `SCHEMA_ID`—that is, the private input is encoded as `SCHEMA_ID || SSZ(StatelessInput)`—already identifies the protocol fork in its first byte. This is needed to select the correct `StatelessInput` shape before SSZ decoding, since fields such as `execution_payload` change across forks.

So technically, `SCHEMA_ID` already mandates which fork and input schema are being used. Keeping `ProtocolFork` in `ChainConfig` would therefore be redundant and could lead to inconsistent values or weird coherence checks.

But... `SCHEMA_ID` is a private input and, until this PR, it was not exposed as a public value.

For forks that share the same `execution_payload` shape—for example, Osaka and the subsequent BPO forks—the proof verifier otherwise has no explicit public value telling it which fork rules the guest used to validate the block. Inputs for those forks can have the same SSZ shape, so the shape alone is not enough for the verifier to distinguish whether the block was validated using Osaka, BPO1, BPO2, and so on.

This PR therefore makes the full `schema_id` part of the public result. The first byte identifies the protocol fork, while the second byte identifies the schema revision/encoding. Returning the full ID, rather than only the fork byte, also pins the exact encoding that the guest decoded.

The proof verifier can now explicitly check which fork rules and schema revision were used to validate the block.

Note that this is intentionally not part of `ChainConfig`. We want to avoid reintroducing a redundant `ProtocolFork` field there when `SCHEMA_ID` already defines both the fork and the input schema.

Okay, so hopefully the above makes sense—it is just being absolutely clear about which schema the guest decoded and which fork rules it considered active during block execution, period.

---

Another discussion was whether `ForkActivation` inside the chain config makes sense. `ForkActivation` contains the block number or timestamp at which the fork activates. The conclusion seems to be that it doesn't make sense for this information to be provided by the host.

In the same way that we removed `blob_schedule` from `ChainConfig` a while ago, expecting the guest program to have something like `map[chain_id][fork][blob_schedule]`, we can argue that it can also have `map[chain_id][fork][fork_activation_info]`.

The fork activation information is still needed by the guest program—for example, to detect whether a block is at a fork boundary and must apply additional logic such as deploying a system contract, or simply to check that the execution payload timestamp is compatible with the selected fork.

The point is that we don't need this information as private input. Removing it from the private input means less prover-controlled configuration, fewer coherence checks, and more trusted configuration baked directly into the guest program.

Said differently: given a `chain_id` and a fork identified by `schema_id`, things such as fork activation information and blob schedules should be completely defined by the guest, not by the host.

After removing `ForkActivation`, `ChainConfig` was left containing only `chain_id`. I therefore removed `ChainConfig` entirely and flattened its usages to carry `chain_id` directly.

So, indirectly, things became quite a bit simpler.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/stateless.py
+++ b/src/ethereum/forks/amsterdam/stateless.py
@@ -18,7 +18,7 @@ from ethereum.state import Root
 from .blocks import Header
 from .execution_engine.new_payload import execute_new_payload_request
 from .execution_engine.requests import ExecutionRequests
-from .execution_engine.types import ExecutionPayload, NewPayloadRequest
+from .execution_engine.types import NewPayloadRequest
 from .fork import ChainContext
 from .fork_types import VersionedHash
 from .witness_state import WitnessState, build_code_db, build_node_db
@@ -106,58 +106,6 @@ class ProtocolFork(IntEnum):
     Amsterdam = 0x15
 
 
-class ChainConfigValidationError(Exception):
-    """
-    Raised when a chain config cannot be used by this stateless guest.
-    """
-
-
-class InactiveForkConfigError(ChainConfigValidationError):
-    """
-    Raised when the configured active fork is not active for the payload.
-    """
-
-
-class InvalidForkActivationError(ChainConfigValidationError):
-    """
-    Raised when a fork entry has a malformed activation point.
-    """
-
-
-@final
-@slotted_freezable
-@dataclass
-class ForkActivation:
-    """
-    Activation point for a protocol fork.
-    """
-
-    block_number: U64 | None
-    timestamp: U64 | None
-
-
-@final
-@slotted_freezable
-@dataclass
-class ChainConfig:
-    """
-    Chain configuration needed for stateless validation.
-    """
-
-    chain_id: U64
-    """
-    Chain identifier used during payload validation and
-    execution.
-    """
-
-    fork_activation: ForkActivation
-    """
-    Activation for the fork selected by the guest schema.
-
-    Used for activation checks and retained for transition-block logic.
-    """
-
-
 @final
 @slotted_freezable
 @dataclass
@@ -179,9 +127,9 @@ class StatelessInput:
     state transition function statelessly.
     """
 
-    chain_config: ChainConfig
+    chain_id: U64
     """
-    Chain configuration values needed during stateless validation.
+    Chain identifier used during payload validation and execution.
     """
 
     public_keys: Tuple[Bytes, ...]
@@ -218,10 +166,10 @@ class StatelessValidationResult:
     means validation failed or the guest input bytes could not be decoded.
     """
 
-    chain_config: ChainConfig
+    chain_id: U64
     """
-    Chain configuration decoded from the input. This is a sentinel default
-    when ``run_stateless_guest`` cannot decode the input bytes.
+    Chain identifier decoded from the input. This is zero when
+    ``run_stateless_guest`` cannot decode the input bytes.
     """
 
     schema_fork_index: U8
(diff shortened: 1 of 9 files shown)


src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -106,59 +106,6 @@ class ProtocolFork(IntEnum):
     Amsterdam = 0x15
 
 
-class ChainConfigValidationError(Exception):
```
jsign, line 109: Lots of simplifications here since now we rely on guest program having chain configuration information like blob schedule (removed a while ago) and now fork activation information.

No need to check for weird, malicious or incoherent private inputs.

src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -106,59 +106,6 @@ class ProtocolFork(IntEnum):
...
-@final
-@slotted_freezable
-@dataclass
-class ChainConfig:
-    """
-    Chain configuration needed for stateless validation.
-    """
-
-    chain_id: U64
-    active_fork: ForkConfig
```
jsign, line 159: `ForkConfig` was gone since `ForkActivation` is gone, thus `ChainConfig` is gone too. Only `chain_id` is useful which was moved upstream.

src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -219,9 +166,15 @@ class StatelessValidationResult:
...
+    chain_id: U64
+    """
+    Chain identifier decoded from the input. This is zero when
+    ``run_stateless_guest`` cannot decode the input bytes.
     """
-    Chain configuration decoded from the input. This is a sentinel default
+
+    schema_fork_index: U8
+    """
+    Fork rules executed by the guest. This uses the invalid-input sentinel
```
jsign: Gist of public input changes

src/ethereum/forks/amsterdam/stateless.py
```diff
@@ -327,18 +237,17 @@ def verify_stateless_new_payload(
     witness = stateless_input.witness
 
     try:
-        validate_chain_config(
-            stateless_input.chain_config,
-            stateless_input.new_payload_request,
-        )
+        # Note that in EEST we have one implementation per fork, so we don't need
+        # to check execution payload timestamp against the current fork activation
+        # information. A real implementation MUST do these checks!
```
jsign, line 245: Removing `validate_chain_config(...)` felt weird, but it is the correct thing to do **in EEST**. 

But because EEST has different impl per fork, so actually validating chain config doesn't make much sense.

Still, left a comment for real EL implementors that support multiple forks in the same guest program. i.e. the must check that `execution_payload.timestamp >= global[chain_id][fork][fork_activation].timestamp` and similars.

I think not being able to express this in EEST is a bit unfortunate to leave it more clear than a comment; but is just a consequence of having one dedicated impl per fork.

src/ethereum/forks/amsterdam/stateless_ssz.py
```diff
@@ -99,11 +93,6 @@
 # --- SSZ Container types ---
 
 
-SszOptionalForkActivationValue: TypeAlias = SszList[
```
jsign, line 102: Interestingly, we could remove the SSZ-optional need with "list with max length 1".

kevaundray (approved):

rubo (approved):

Issue #3251, opened 2026-07-28
Genesis state-root computation is coupled to the MPT across the testing tooling


CPerezz:
Genesis-state-root computation is coupled to the Merkle Patricia Trie in several places
across the testing tooling. PR #3246 fixed one of them because it was actively producing
wrong fixtures, but the general problem deserves a designed solution rather than another
point fix.

### What #3246 changed

Fixtures for the experimental `BinaryTree` (EIP-8297) fork carried a genesis root no client could reproduce. The fix routes it through `spec_calc_state_root(alloc, fork)`, matching what `specs/blockchain.py` already did.

That was in scope because it was a live defect on the fill path. The wider coupling was not.

### What remains

Two `execute`-path sites still compute a genesis root without resolving the fork's provider:

- `packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/hive.py:153`
- `packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/eth_config/execute_types.py:371`
  (this one has `self.config.fork()` in scope, so it could route today)

### The broader concern (raised by @kevaundray on #3246)

> I found that on the CL, there is a coupling to the MPT because the CL hardcodes a EL block
> -- I'm wondering if doing it this way will make us need to change more tooling

That is the part worth designing rather than patching. If the consensus-layer side assumes an
MPT-shaped EL genesis block, then threading the provider through each EL-side call site may
not be sufficient, and may not even be the right shape — we could end up with provider
resolution scattered across the tooling while the actual blocker sits elsewhere.

### Proposed next step

Investigate before changing anything further:

1. Enumerate every genesis-root computation across `fill`, `execute`, hive integration, and
   any CL-facing fixture or genesis generation.
2. Establish concretely where the CL hardcodes an EL block and what it assumes about it.
3. Decide whether the right fix is threading the fork through each site, a single
   provider-aware genesis builder, or something upstream of both.
4. Then implement, rather than continuing to fix sites one at a time as they are found.

Filed as a follow-up to #3246 so that PR can land on its correctness merits; no work here
blocks it.

PR #3279, opened 2026-08-01
feat(test-types): Fork-based State Commitment Property in Alloc

marioevz:
### Description

Attempt to respond to concern in #3251: teach `Alloc` to know **which state-commitment scheme** its state root is computed under, and let the **fork** decide that scheme. Today every fork commits with a Merkle-Patricia Trie (MPT); this change lets a future fork switch to a binary tree without touching the allocation-handling code — the correct root algorithm is selected per fork instead of being hard-wired to `state_mpt`.

This is behaviorally a **no-op on mainnet forks**: every fork returns `StateCommitment.MPT`, so all state roots are computed exactly as before. It only adds the plumbing (and the seeding) needed for a binary-tree fork to drop in later (or in work branches).

**Core pieces**

- **`StateCommitment` enum** (`MPT` / `BINARY`) added to `base_types` — a spec-free vocabulary type so both `forks` and `test_types` can reference it without new import edges.
- **`Fork.state_commitment()`** classmethod on `BaseFork`, defaulting to `MPT`. A binary-transition fork overrides it to `BINARY`; no spec import enters the `forks` package.
- **`Alloc` carries its scheme.** A private `_state_commitment` field (default `MPT`), a public `state_commitment()` getter, and `migrate_state_commitment()` to switch it (rejected once the alloc is `FROZEN`). `state_root()` / `_materialize_state()` now dispatch through a new `_state_module()` that maps the scheme → the spec implementation module; `BINARY` raises `NotImplementedError` until a `state_bmt` module exists.
- **Backend-agnostic materialization.** `_materialize_state()` is now typed against `ethereum.state.PreState` (which both `state_mpt.State` and the future binary `State` satisfy) and only touches the state through public module functions (`State`, `set_account`, `set_storage`, `store_code`) — it no longer pins `state_mpt.State` or pokes the private `_code_store`.
- **`Alloc.merge` is commitment-aware.** It now propagates a scheme instead of minting a default-`MPT` allocation: an explicit `state_commitment=` argument wins, otherwise the second (right-hand) allocation's scheme takes precedence. It also switched to `model_copy(deep=True)`, which preserves the private scheme field.

**Seeding the scheme (private attributes are not serialized)**

Because `_state_commitment` is a private attribute, it is not restored by `model_validate`/`model_dump_json` and is dropped by JSON round-trips, so it must be (re)seeded from the fork wherever an allocation is constructed, loaded, or merged before a root is computed:

- `shared/pre_alloc.Alloc.__init__` — seed at construction from `fork.transitions_from().state_commitment()`.
- `pre_alloc_groups.PreAllocGroupBuilder.model_post_init` — seed on every load-from-file (covers `PreAllocGroup` and `GroupPreAlloc` via inheritance); `add_test_alloc` asserts both sides agree.
- `execute/rpc/hive.py` and `execute/tests/test_execute_remote.py` — seed the execute-path genesis allocation.
- `ethereum_spec_tools/evm_tools/t8n` — after the defensive `model_copy(deep=True)`, seed `self.alloc` from `t8n_data.fork` (which is always a concrete, non-transition fork here), covering the JSON CLI path where the scheme is otherwise lost to the MPT default.

**Enabling binary later** (once a spec `state_bmt` module lands) is three steps, none in `_materialize_state`:

1. Import the binary state module in `account_types.py`.
2. Return it from `_state_module()` for `StateCommitment.BINARY` instead of raising.
3. Override `Fork.state_commitment()` → `BINARY` on the transition fork.

The transition-block state-root semantics at a commitment boundary remain to be defined in a follow-up.

cc @CPerezz 

### Related Issues or PRs
N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
@@ -73,7 +73,7 @@ class Alloc(SharedAlloc):
     def __init__(
         self,
         *args: Any,
-        fork: Fork,
+        fork: Fork | TransitionFork,
         flags: AllocFlags,
         stub_accounts: Dict[str, Account] | None = None,
         stub_eoas: Dict[str, EOA] | None = None,
@@ -495,21 +495,14 @@ def stub_eoas(
 @pytest.fixture(scope="function")
 def pre(
     alloc_flags: AllocFlags,
-    fork: Fork | None,
-    request: pytest.FixtureRequest,
+    fork: Fork | TransitionFork,
     stub_accounts: Dict[str, Account],
     stub_eoas: Dict[str, EOA],
 ) -> Alloc:
     """Return default pre allocation for all tests (Empty alloc)."""
-    # FIXME: Static tests don't have a fork so we need to get it from the node.
-    actual_fork = fork
-    if actual_fork is None:
-        assert hasattr(request.node, "fork")
-        actual_fork = request.node.fork
-
     return Alloc(
         flags=alloc_flags,
-        fork=actual_fork,
+        fork=fork,
         stub_accounts=stub_accounts,
         stub_eoas=stub_eoas,
     )
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
@@ -435,7 +435,7 @@ ALL_FIXTURE_FORMAT_NAMES.sort(key=len, reverse=True)
 
 @pytest.fixture(scope="function")
 def node_id_for_entropy(
-    request: pytest.FixtureRequest, fork: Fork | None
+    request: pytest.FixtureRequest, fork: Fork | TransitionFork
 ) -> str:
     """
     Return the node id with the fixture format name and fork name stripped.
@@ -453,11 +453,6 @@ def node_id_for_entropy(
     # deterministic regardless of whether xdist is active.
     if "@" in node_id:
         node_id = node_id.rsplit("@", 1)[0]
-    if fork is None:
-        # FIXME: Static tests don't have a fork, so we need to get it from the
-        # node.
-        assert hasattr(request.node, "fork")
-        fork = request.node.fork
     for fixture_format_name in ALL_FIXTURE_FORMAT_NAMES:
         if fixture_format_name in node_id:
             parts = node_id.split("::")
--- a/packages/testing/src/execution_testing/test_types/account_types.py
+++ b/packages/testing/src/execution_testing/test_types/account_types.py
@@ -123,10 +123,13 @@ class Alloc(BaseAlloc):
 
     _phase: _Phase = PrivateAttr(default=_Phase.CONSTRUCTION)
     _code_store: Dict[Hash32, Bytes] = PrivateAttr(default_factory=dict)
-    _state_commitment: StateCommitment = PrivateAttr(
-        default=StateCommitment.MPT
-    )
-    """Commitment scheme this allocation's state root is computed under."""
+    _state_commitment: StateCommitment | None = PrivateAttr(default=None)
+    """
+    Commitment scheme this allocation's state root is computed under.
+
+    Unset by default: it must be seeded from the accompanying fork before any
+    state-root computation.
+    """
 
     @dataclass(kw_only=True)
     class UnexpectedAccountError(Exception):
@@ -371,6 +374,11 @@ class Alloc(BaseAlloc):
         """
         Return the spec state module implementing `self._state_commitment`.
         """
+        if self._state_commitment is None:
+            raise ValueError(
+                "Alloc state commitment is unset; seed it from the "
+                "accompanying fork."
+            )
         if self._state_commitment is StateCommitment.MPT:
             return spec_state_mpt
         raise NotImplementedError("State commitment type not yet implemented.")
@@ -582,11 +590,16 @@ class Alloc(BaseAlloc):
         """Lock the allocation: no further mutations allowed."""
         self._phase = _Phase.FROZEN
 
-    def state_commitment(self) -> StateCommitment:
-        """Return the commitment scheme this allocation is committed under."""
+    def state_commitment(self) -> StateCommitment | None:
+        """
+        Return the commitment scheme this allocation is committed under, or
+        `None` if it has not been seeded from a fork yet.
+        """
         return self._state_commitment
 
-    def migrate_state_commitment(self, commitment: StateCommitment) -> None:
+    def migrate_state_commitment(
+        self, commitment: StateCommitment | None
+    ) -> None:
         """
         Switch the commitment scheme used to compute the state root.
         """
--- a/packages/testing/src/execution_testing/test_types/account_types.py
+++ b/packages/testing/src/execution_testing/test_types/account_types.py
@@ -370,9 +370,6 @@ class Alloc(BaseAlloc):
     def _state_module(self) -> ModuleType:
         """
         Return the spec state module implementing `self._state_commitment`.
-
-        The MPT implementation is the only one currently available; a binary
-        allocation raises until the spec `state_bmt` module exists.
         """
         if self._state_commitment is StateCommitment.BINARY:
             raise NotImplementedError(
@@ -596,11 +593,6 @@ class Alloc(BaseAlloc):
     def migrate_state_commitment(self, commitment: StateCommitment) -> None:
         """
         Switch the commitment scheme used to compute the state root.
-
-        Called by blockchain fillers at a transition-fork boundary where the
-        target fork's commitment differs from the scheme this allocation was
-        built under. Account data is scheme-independent, so only the
-        root-computation strategy changes.
         """
         if self._phase is _Phase.FROZEN:
             raise RuntimeError(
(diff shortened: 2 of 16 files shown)


packages/testing/src/execution_testing/test_types/account_types.py
```diff
@@ -357,16 +367,28 @@ def _ensure_live(self) -> None:
...
 
-    def _materialize_state(self) -> spec_state_mpt.State:
+    def _state_module(self) -> ModuleType:
+        """
+        Return the spec state module implementing `self._state_commitment`.
+        """
+        if self._state_commitment is StateCommitment.BINARY:
+            raise NotImplementedError(
+                "Binary-tree state commitment is not yet available: no spec "
+                "state_bmt module exists."
```
kevaundray: Small note: in projects/binary-trie, its called state_pbt where pbt = partitioned binary tree

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
```diff
@@ -495,21 +495,14 @@ def stub_eoas(
...
+    fork: Fork | TransitionFork,
     stub_accounts: Dict[str, Account],
     stub_eoas: Dict[str, EOA],
 ) -> Alloc:
     """Return default pre allocation for all tests (Empty alloc)."""
-    # FIXME: Static tests don't have a fork so we need to get it from the node.
-    actual_fork = fork
-    if actual_fork is None:
-        assert hasattr(request.node, "fork")
-        actual_fork = request.node.fork
```
kevaundray, line 508: Haven't personally checked that this does not crash when we try to fill static tests -- I'm probably missing something, from a glance it seems that ci doesn't fill these

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/pre_alloc.py
kevaundray: Related to [this comment](https://github.com/ethereum/execution-specs/pull/3279/changes#r3695657207) seems we don't need to remove this check?

kevaundray (commented):
Making the state commitment an explicit property of the fork makes sense to me :)

Left some small comments, but LGTM

CPerezz:
I'm at the point in https://github.com/ethereum/go-ethereum/pull/35436 where I can use it to try filling tests and reach consensus on the format etc.. Such that we can indeed be sure this works well with a real client.

CPerezz (changes requested):
With the help of Fable, I was able to find some loops that would break PBT. I instructed it to give a step-by-step guide on how we hit a critical correctness error:

### PBT genesis roots for any contract with code are broken

- Step 1 — genesis roots are computed on a freshly-constructed alloc. The fill path calls state_root() directly on the test's pre-alloc, e.g. packages/testing/src/execution_testing/specs/blockchain.py:822 [head]:
```
state_root = pre_alloc.state_root()
```
Same shape at hive.py:156, pre_alloc_groups.py:78, execute_types.py:371 [head]. At that moment the alloc is in _Phase.CONSTRUCTION — nothing has touched it as a PreState yet.

- Step 2 — state_root() materializes a spec-side state (account_types.py:311-313 → :381 [head]). The PR's _materialize_state() ends with (account_types.py:413-422 [head]):
```
                mod.set_storage(
                    state, addr,
                    Bytes32(int(key_hi).to_bytes(32, "big")),
                    U256(value_int),
                )
        for stored_code in self._code_store.values():
            mod.store_code(state, stored_code)
        return state
```
The per-account loop above it computes `code_hash = spec_keccak256(code)` without ever storing the bytes — code bytes only reach the spec state through that trailing loop over self._code_store.

- Step 3 — `_code_store is empty on this path`. It's only populated by _build_cache() (account_types.py:353-362 [head]), which only runs from _ensure_live() (:364-368), whose only callers are the PreState getters and compute_state_root (:431, :451, :467, :478, :491 [head]). state_root() and _materialize_state() never call it. So on the genesis path, **the trailing loop iterates an empty dict.**

- Step 4 — MPT doesn't care, PBT does. **MPT commits accounts by code_hash only, so a missing code body is invisible. But a content-committing scheme must chunk the actual bytecode:** src/ethereum/state_pbt.py:77 [branch] inside embed_flat_state:
```
        code = get_code(account.code_hash)
```
which resolves through State.get_code (state_pbt.py:131-139 [branch]):
```
        if code_hash == EMPTY_CODE_HASH:
            return b""
        return self._code_store[code_hash]   # ← KeyError: bytes were never stored
```
So the moment `_state_module()` returns `state_pbt`, **every genesis containing a contract raises KeyError (or, in a scheme that tolerated it, would silently commit wrong chunk leaves).**

CPerezz (commented):
`execute_types.py:371` silently produces MPT roots by default.
```
genesis_fork, env, self.alloc.state_root()
```

`self.alloc.state_root()` runs with the default-MPT enum — nothing seeds this alloc. The fork is sitting in scope a few lines above (#3251 mentions exactly this). 

The PR fixes the other execute-path site (`hive.py:153-155`) but not this one. Today it's harmless (mainnet forks are all MPT); the moment someone points `execute --eth-config` at a binary devnet, it produces an MPT root silently.

packages/testing/src/execution_testing/test_types/account_types.py
```diff
@@ -222,18 +229,21 @@ def merge(
...
-                merged.pop(address, None)
+                merged.root.pop(address, None)
 
-        return Alloc(merged)
+        if state_commitment is not None:
+            merged.migrate_state_commitment(state_commitment)
+        else:
+            # By default, state commitment of the second alloc takes precedence
+            merged.migrate_state_commitment(alloc_2.state_commitment())
+        return merged
```
CPerezz, line 249: Why did we get rid of `model_dump`?

` merged[address] = merged_account` calls `_require_construction`, so a LIVE left-hand alloc raises. My repro used an overlapping account, but by the code above the set happens for every non-empty `alloc_2` entry, so it generalizes: merging a LIVE alloc with any non-empty right side raises.
- The unconditional `migrate_state_commitment` hits the `FROZEN` guard, so a `FROZEN` left-hand alloc raises even when both sides are MPT and no scheme change was asked for.

Did a repro with a couple agents and hit these:

 the old model_dump() merge was phase-immune; no production caller triggers them today, so latent API regressions):

- F3a: `Alloc.merge(frozen, x)` raises `RuntimeError: migrate_state_commitment not allowed: Alloc is FROZEN` — the unconditional migrate at account_types.py:241-245 runs on the deep copy that inherited alloc_1's phase, even with no scheme change requested.
- F3b: `Alloc.merge(live, non_empty)` raises `RuntimeError: __setitem__ not allowed: Alloc is in phase LIVE — merged[address] = … ` hits `_require_construction`; any `PreState` read (e.g. get_account_optional) flips an alloc to `LIVE` first.

Paste-ready tests:
```python
import pytest
from execution_testing.base_types import Account, Address
from execution_testing.test_types.account_types import Alloc


def test_merge_frozen_alloc_raises():  # F3a — worked before #3279
    frozen = Alloc({})
    frozen.freeze()
    with pytest.raises(RuntimeError, match="FROZEN"):
        Alloc.merge(frozen, Alloc({}))


def test_merge_live_alloc_raises():  # F3b — worked before #3279
    addr = Address(0x1234)
    live = Alloc({addr: Account(balance=1)})
    live.get_account_optional(bytes(addr))  # any PreState read -> LIVE
    with pytest.raises(RuntimeError, match="LIVE"):
        Alloc.merge(live, Alloc({addr: Account(balance=2)}))
```

Not sure there are any callers to `freeze` in the repo. But still the API is there so we should handle this correctly.

CPerezz (changes requested):
**Pydantic private attrs are excluded from serialization,** so the scheme evaporates at exactly the boundaries fixtures cross. Imagine for example the following:

```python
a = Alloc({}); a.migrate_state_commitment(StateCommitment.BINARY)
b = Alloc.model_validate_json(a.model_dump_json())
# F2  round-trip: BINARY -> MPT                       ← silent
c = Alloc.merge(a, Alloc({}))
# F2b merge(BINARY, fresh-MPT) -> MPT                 ← "alloc_2 wins" (:244-245)
```

Every future entry point (a new loader, a new merge site) fails silently toward MPT, producing a root no client can reproduce. Related to #3246 's bug. Note the precedence rule is doing no real work today:` hive.py `merges and then immediately re-seeds anyway.

Why not make the default `None` and raise in `_state_module()` when a root is computed on an unseeded alloc?? Maybe there's better ways. But this definitely needs some attention

spencer-tb (approved):
LGTM from my side as it stands.

PR/commit: https://github.com/marioevz/execution-specs/pull/9, if you want to add the `--state-trie binary` flag to the fill command :)

marioevz:
> `execute_types.py:371`

Yes, I saw this and thought that mainnet/testnets would never have PBT genesis roots, but I didn't think of devnets. It's an easy fix, so I'll apply it. Thanks!

packages/testing/src/execution_testing/test_types/account_types.py
marioevz: `model_copy` preserves the `_state_commitment`, so it's a way to avoid setting it again.

Plus, dumping to JSON to read it again in a function that is this widely used was a bad idea from the beginning.

I suggest we keep this as is, and if there's an use case to merging two frozen Allocs we can revisit the suggestion.

marioevz:
@kevaundray @CPerezz I've applied all the comments. Please take a look. Thanks!

CPerezz (approved):
LGTM! Great work!

Thanks for doing it that fast!

PR #3280, opened 2026-08-02
chore: Temporarily add PBT conformance tests

kevaundray:
### Description

When bootstrapping a new state trie, it is convenient to be able to test the state trie in isolation of the STF since a divergence only implies that there could be a problem in the state trie, but its not guaranteed.

Output: On each commit to `projects/binary-trie`, a .json file will be generated with trie test vectors if something has changed.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture


kevaundray:
This says temporarily because I don't think we will add this into forks/I* as it would be setting a new precedence; we don't have conformance tests for the MPT

Issue #3281, opened 2026-08-03
Benchmark Full Suite Generation Tracker


LouisTsai-Csie:
## Description

To finalize repricing numbers, we need to run the complete benchmark test suite on both compute and stateful benchmarks (not repricing-specific).

Initial results on Benchmarkoor:
- [state-actor-glamsterdam-devnet-7-full-stateful](https://benchmarkoor.core.ethpandaops.io/suites/0c5715e398c47aa3)
- [state-actor-glamsterdam-devnet-7-full-compute](https://benchmarkoor.core.ethpandaops.io/suites/9dbe0f5f4ea081b8)

This tracker documents what's missing or broken in the output.

## Compute

Statistics: passed 2040, failed 204, skipped 22, deselected 2.

These scenarios are not constructed successfully via the stateful filler. The constructed blocks are not accepted by clients:
- [x] `test_ether_transfers_to_precompile`
- [x] `test_ether_transfers`: diff_acc* ignored (66/120)
- [x] `test_auth_transaction`: `empty_authority_False` ignored (16/32)
- [x] `test_storage_access_cold`: `absent_slots_False` ignored (14/28)
- [x] `test_selfdestruct_existing`
- [x] `test_ext_account_query_cold`: `absent_accounts_False` (2/4)
- [x] `test_mixed_dependency_graph`: `all max_txs + half of half_max_txs (10/18)`
- [x] `test_deploy_then_interact`: `max_txs-single_contract` (2/12)

Refactor PR: issue #3282

Below tests needs further refactor, the current storage initialization approach would exceed the initcode prefix size limit.

- [ ] `test_state_root_computation`: initcode prefix too long

Below tests are being skipped:
- [ ] `test_unchunkified_bytecode`: contract deployment gas exceeds the benchmark gas value
- [ ] `test_creates_collisions`: Benchmark gas value cannot cover a single contract creation
- [ ] `test_bls12_pairing_uncachable[num_pairs_24]`: calldata intrinsic gas cost exceeds the gas limit, this could be refactored.
- [ ] `test_block_full_access_list_and_data`: EIP-7934 block RLP size limit might be exceeded -> this expected but worth refactoring

## Stateful

The test is limited to `tests/benchmark/stateful/bloatnet`, below is the failing cases:

1. `depth_benchmarks/test_deep_branch.py`
2. `eip7928_block_level_access_lists/`: not within the current scope.
[...]

PR #3282, opened 2026-08-03
refactor(test-benchmark): split stateful setup transactions across blocks

LouisTsai-Csie:
## Problem 1: Gas Limit Mismatch Between EELS and synthetic state

EELS configures the block gas limit to 1 Trillion, while pre-state generated via state-actor uses 1 Gigagas. This mismatch prevents early detection of any exceeding block gas limit issues.

Example: In `test_auth_transaction` with `empty_authority_False` parametrization, the setup phase requires many authorization transactions. The total gas exceeds 1 Gigagas but stays under 1 Trillion. EELS construction succeeds, but the snapshot rejects it by default.

### Root Cause

Currently, `blockchain.py` packs all setup transactions (`fund_eoa`, `deploy_contract`, etc.) into a single block. After state repricing, this easily exceeds 1 Gigagas (some test would create hundreds of thousands of slots, accounts, delegation). Since setup gas isn't verified, a setup phase can exceed the benchmark's own gas limit and still pass in `fill` mode. However, when filling against a live client with `fill-stateful`, the client's block gas limit rejects the block.

### Solution

This PR splits setup transactions across multiple blocks, ensuring each strictly respects the block gas limit. This could fix the following broken cases:

- `test_ether_transfers_to_precompile`
- `test_ether_transfers`
- `test_auth_transaction`
- `test_storage_access_cold`
- `test_ext_account_query_cold`
- `test_mixed_dependency_graph`
- `test_state_root_computation` (initcode prefix too long)
- `test_deploy_then_interact`

## Problem 2: RPC request exceeds size limit

Certain RPC calls have request body size limits. The `test_selfdestructing_existing` benchmark exceeds this limit because it calls `get_alloc` with a large batch of data at once. This PR splits the requests into smaller batches to stay within the limit.

## Additional Fixes

- Two tests now fund a single sender instead of one per transaction, this avoids creating thousands of account during setup phase.
- `test_block_full_data` funds recipients with an explicit amount (no deferred balances)

### Related Issues or PRs

issue #3281 

### Checklist

### Cute Animal Picture


LouisTsai-Csie:
Please carefully review this refactor, or it would slow down the entire benchmark process a lot!

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -720,6 +720,49 @@ def _split_blocks_by_phase(blocks: List[Block]) -> List[Block]:
...
+                current_run, current_gas = [], 0
+            current_run.append(tx)
+            current_gas += tx.gas_limit
+        runs.append(current_run)
+
+        if len(runs) == 1:
+            out.append(block)
+            continue
+
+        out.extend(block.model_copy(update={"txs": run}) for run in runs)
```
spencer-tb: The block level expectation fields (expected_post_state etc) I think should be set here. As they describe the blocks final state, so when splitting they should stay on the last "sub" block only and be cleared on the earlier ones. Otherwise I think the earlier "sub" blocks get verified against expectations that only hold after the full setup has run.
```suggestion
        out.extend(
            block.model_copy(
                update={
                    "txs": run_txs,
                    "header_verify": None,
                    "rlp_modifier": None,
                    "expected_block_access_list": None,
                    "expected_post_state": None,
                    "expected_gas_used": None,
                    "exception": None,
                    "skip_exception_verification": False,
                    "engine_api_error_code": None,
                }
            )
            for run_txs in runs[:-1]
        )
        out.append(block.model_copy(update={"txs": runs[-1]}))
```

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -720,6 +720,49 @@ def _split_blocks_by_phase(blocks: List[Block]) -> List[Block]:
...
+
+        runs: List[List[Transaction]] = []
+        current_run: List[Transaction] = []
+        current_gas = 0
+        for tx in block.txs:
+            if current_run and current_gas + tx.gas_limit > budget:
+                runs.append(current_run)
+                current_run, current_gas = [], 0
+            current_run.append(tx)
+            current_gas += tx.gas_limit
```
spencer-tb: A setup tx with a gas limit above the block gas limit itself can never fit any split, so raising early here feels better, will probably need to update the tests:
```suggestion
        for tx in block.txs:
            if tx.gas_limit > block_gas_limit:
                raise ValueError(
                    f"Setup tx gas_limit ({int(tx.gas_limit)}) exceeds "
                    f"the snapshot chain's block gas limit "
                    f"({block_gas_limit}); no split can make it fit."
                )
            if current_run and current_gas + tx.gas_limit > budget:
                runs.append(current_run)
                current_run, current_gas = [], 0
            current_run.append(tx)
            current_gas += tx.gas_limit
```

spencer-tb (commented):
Couple of comments, I think they should make sense :)

packages/testing/src/execution_testing/specs/blockchain.py
LouisTsai-Csie: Thanks for pointing out!

packages/testing/src/execution_testing/specs/blockchain.py
LouisTsai-Csie: Nice catch. I think no test modification needed since the block gas limit is 1T, it is hard to exceed :D

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/live_client_flags.py
```diff
@@ -28,7 +28,7 @@
 # Multiplier applied to a one-shot live fee-market query to absorb the gap
 # between query timing and tx submission (basefee can climb a few blocks
 # between the two; the bump keeps txs landing without per-tx requeries).
-FEE_BUMP_MULTIPLIER = 1.5
+FEE_BUMP_MULTIPLIER = 3.0
```
LouisTsai-Csie, line 31: The `max_fee_per_gas` and other gas configuration values adjust based on network status. However, the `FEE_MULTIPLIER` is only 1.5x. For benchmarks with many blocks, the cumulative gas limit can exceed the value calculated at the start of the test, leading to lower base fee issues.

packages/testing/src/execution_testing/rpc/rpc.py
```diff
@@ -1129,7 +1129,12 @@ def get_alloc(
             all_calls.extend(calls)
             address_info.append((address, call_info))
 
-        responses = self.post_batch_request(calls=all_calls)
+        responses: List[JSONRPCResponse] = []
+        batch_size = self.max_transactions_per_batch
+        for i in range(0, len(all_calls), batch_size):
+            responses.extend(
+                self.post_batch_request(calls=all_calls[i : i + batch_size])
+            )
```
LouisTsai-Csie: Instead of sending RPC requests all at once, this PR batches them to reduce server load.

tests/benchmark/compute/instruction/test_account_query.py
```diff
@@ -304,7 +306,7 @@ def test_ext_account_query_cold(
                     to=op_address,
                     data=calldata,
                     gas_limit=gas_limit,
-                    sender=pre.fund_eoa(),
+                    sender=execution_sender,
                 )
```
LouisTsai-Csie, line 310: It is better than the original version:

```python
execution_txs.append(
    Transaction(
        to=op_address,
        data=calldata,
        gas_limit=gas_limit,
        sender=pre.fund_eoa(), # This requires pre-funding for EVERY account
    )
)
```

Pre-funding a sender per transaction is unnecessary and slows setup. Empty account creation is now much more expensive after repricing.

tests/benchmark/stateful/bloatnet/test_transaction_types.py
```diff
@@ -164,13 +164,14 @@ def test_ether_transfers_onchain_receivers(
...
-        )
         + receiver_execution_gas
     )
-    iteration_count = gas_benchmark_value // iteration_cost
+    state_cost = fork.transaction_top_frame_state_gas(
+        sends_value=sends_value,
+        recipient_type=recipient_type,
+    )
+    iteration_cost = regular_cost + state_cost
+    iteration_count = gas_benchmark_value // max(regular_cost, state_cost)
```
LouisTsai-Csie, line 174: This is the _actual_ worst case for ether transfer, the iteration cost is `sum(R, S)`, while the gas usage in the header is `max(R, S)`, not `sum(R, S)`.

The original approach used ~90% of the block gas limit, but now it reaches 99%.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/live_client_flags.py
jochem-brouwer: What does this do? I think we should, if we do not want to influence baseFee, not send more than half of the gas limit in setup transactions to avoid increasing the base fee. This constant seems a bit random.

packages/testing/src/execution_testing/specs/tests/test_split_setup_blocks_by_gas.py
```diff
@@ -0,0 +1,117 @@
...
+    out = _split(Block(txs=txs))
+    assert [tx for block in out for tx in block.txs] == txs
+
+
+def test_tx_at_the_limit_gets_its_own_block() -> None:
+    """A tx past the budget but within the limit is isolated, not dropped."""
+    small, huge = _tx(21_000), _tx(BLOCK_GAS_LIMIT)
+    out = _split(Block(txs=[small, huge, small]))
+    assert [len(block.txs) for block in out] == [1, 1, 1]
+    assert out[1].txs[0] is huge
```
jochem-brouwer: Could you add more assertions here so it is clear what is being tested? I think block 0 has `small`, block 1 has `huge` and block 2 also has `small`.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -720,6 +720,72 @@ def _split_blocks_by_phase(blocks: List[Block]) -> List[Block]:
...
+                update={
+                    "txs": run_txs,
+                    "header_verify": None,
+                    "rlp_modifier": None,
+                    "expected_block_access_list": None,
+                    "expected_post_state": None,
+                    "expected_gas_used": None,
+                    "exception": None,
+                    "skip_exception_verification": False,
+                    "engine_api_error_code": None,
```
jochem-brouwer: Why do we override all these params? Couldnt this lead to unexpected effects?

tests/benchmark/compute/scenario/test_transaction_types.py
```diff
@@ -406,7 +407,7 @@ def test_block_full_data(
 
         txs.append(
             Transaction(
-                to=pre.fund_eoa(),
+                to=pre.fund_eoa(amount=1),
```
jochem-brouwer, line 410: Why now the amount=1?

jochem-brouwer (dismissed):
Some small questions, looks good in general 😄 👍

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/live_client_flags.py
LouisTsai-Csie: >  if we do not want to influence baseFee, not send more than half of the gas limit in setup transactions to avoid increasing the base fee.

Does base fee actually affect the benchmark results? If not, I'd prefer not to limit setup transactions to 50% of the block gas limit, that would require more blocks for max code size deployments, which already takes a long time.

Also, the `FEE_BUMP_MULTIPLIER` constant is arbitrary. We're using the same approach for `execute remote`, do you have a better suggestion?

tests/benchmark/compute/scenario/test_transaction_types.py
LouisTsai-Csie: I am not entirely sure about the root cause, but if this `tx.to` is a non-empty account without specified funding amount, it would triggered a deferred mechanism that raises error. I would create separate PR or issue for fixing that problem, this change here would be a workaround - but it should not change the logic as it only affects `tx.to` not `tx.sender`.

packages/testing/src/execution_testing/specs/tests/test_split_setup_blocks_by_gas.py
LouisTsai-Csie: Done! Thanks for pointing out.

packages/testing/src/execution_testing/specs/blockchain.py
LouisTsai-Csie: This is my understanding: when a block is split into multiple smaller blocks, the original post-state expectation no longer applies to each intermediate block. Additionally, since we only care about the final block, which is the one that we benchmark, we skip post-state verification for the setup blocks. Does this make sense?

related comment: https://github.com/ethereum/execution-specs/pull/3282#discussion_r3704899011

packages/testing/src/execution_testing/rpc/rpc.py
LouisTsai-Csie: Resolved in #3327  `post_batch_request` function

marioevz:
Reviewing the `specs/blockchain.py` part of this.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1545,50 +1499,82 @@ def make_stateful_fixture(
...
+                block=Block(txs=current_run),
                 previous_env=env,
                 previous_alloc=alloc,
             )
-            assert isinstance(built_block, TestingBuildBlock), (
-                "ClientBackend must return TestingBuildBlock; got "
-                f"{type(built_block).__name__}"
+            # The client's authoritative block hash (the FixtureHeader RLP
+            # hash diverges — the client picks fields like gas_limit).
+            client_hash, payload = built_block.get_payload_fixture()
```
danceratopz, line 1534: Both this call and the one in the execution loop drop the phase: `get_payload_fixture()` is never called with `phase=`, so every payload in a stateful fixture now serializes `phase: null`, where setup payloads previously carried `phase: setup`. Relatedly, `BuiltBlock.block_phase` is written in `generate_block_data` but never read anywhere.

Simplest fix is to fall back to `self.block_phase` inside `get_payload_fixture`, which also gives that field a purpose. If the per-payload phase is intentionally dead now that payloads are partitioned structurally, then drop `block_phase` and the `phase` kwarg instead, and update the `FixtureEngineNewPayload.phase` description in `fixtures/blockchain.py`, which still claims consumers partition on it.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1499,17 +1452,18 @@ def make_stateful_fixture(
...
-            if setup_txs:
-                blocks_to_process.append(Block(txs=setup_txs))
-        # Each block must be single-phase (Block.phase asserts otherwise);
-        # mixed blocks (e.g. EIP-7702 authorization + benchmark exec) are
-        # split into contiguous phase runs so benchmark gas isn't
-        # swallowed into ``setupEngineNewPayloads``.
-        blocks_to_process.extend(_split_blocks_by_phase(self.blocks))
+        for block in self.blocks:
+            if block.txs and block.phase is TestPhase.SETUP:
+                setup_txs.extend(block.txs)
```
danceratopz, line 1464: Dissolving a SETUP block here silently drops every block-level field it carries (`expected_post_state`, `header_verify`, `expected_gas_used`, `timestamp`, `withdrawals`, and so on). The removed `_split_blocks_by_phase` cleared these fields explicitly and kept them on the last sub-block. Consider raising if a SETUP block sets anything besides `txs`, so test authors find out at fill time rather than having expectations ignored.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1545,50 +1499,82 @@ def make_stateful_fixture(
         # Alloc is not authoritative in stateful mode; pass self.pre as a
         # placeholder — ClientBackend ignores it.
         alloc: Alloc | LazyAlloc = self.pre
-        for block in blocks_to_process:
+
+        # Pack setup txs into as few gas-bounded blocks as possible to create
+        # the setup payloads.
+        setup_blocks_gas_target = parent_header.gas_limit // 2
```
danceratopz, line 1505: Half the gas limit is the EIP-1559 gas target, so packing to it keeps the base fee roughly flat across the setup chain. Worth a named constant plus a one-line comment saying that.

Also note the guard below only rejects `tx.gas_limit > parent_header.gas_limit`, so a tx between the target and the limit gets a solo block above target and bumps the base fee. If that is intended, a comment would help.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1545,50 +1499,82 @@ def make_stateful_fixture(
...
+        while setup_txs:
```
danceratopz, line 1506: This packing loop and the two new `model_post_init` guards have no unit tests, and no CI job executes `make_stateful_fixture`. Suggest extracting the packing into a pure helper, something like `_pack_setup_txs(txs, gas_target) -> list[list[Transaction]]`, and adding table-driven tests: empty input, oversized tx raises, exact boundary fit, order preservation, every emitted block within budget. `pack_transactions_into_blocks` in `tests/benchmark/helper/transactions.py` is a near-identical greedy packer, so a shared helper would also remove that duplication.

Nit: `setup_txs.pop(0)` is quadratic over the queue; an index or a deque avoids that.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1545,50 +1499,82 @@ def make_stateful_fixture(
...
-            else:
-                execution_payloads.append(payload)
-                block_opcode_count = t8n.extract_block_opcode_count(
-                    client_hash
-                )
-                execution_opcode_counts.append(
-                    block_opcode_count.model_dump()
-                    if block_opcode_count is not None
-                    else None
+            # Setup blocks (pre-alloc funding/deploys) are exempt:
```
danceratopz, line 1565: Leftover from the pre-rework code: setup blocks can no longer appear in this loop, the exemption now happens structurally in the packing loop above, so this comment misleads. Move or delete it. The `make_stateful_fixture` docstring has the same problem: it still describes a single synthetic setup block and partitioning by `FixtureEngineNewPayload.phase`.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -811,6 +757,18 @@ def model_post_init(self, __context: Any, /) -> None:
                         f"{i} contains an invalid transaction elsewhere"
                     )
 
+        execution_phase_block_found = False
+        for block in self.blocks:
+            if block.phase == TestPhase.SETUP:
```
danceratopz, line 762: These two new guards are global rules for every blockchain test on every backend, not just fill-stateful: no mixed-phase blocks, and no SETUP block after any non-setup block. Note that an empty `Block()` counts as non-setup here, so an empty block placed before a SETUP block now fails. Probably the right call, but it forbids patterns that used to be legal, so it deserves a mention in `docs/writing_tests/benchmarks.md`.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/live_client_flags.py
danceratopz, line 31: This bump leaves six stale 1.5x references: the help strings at lines 47, 58, 71 and 82 in this file, `fill_stateful.py:548`, and the comment above this constant. It also raises what every live-client tx pays on the execute path, and it is not mentioned in the PR description. With setup blocks now packed to the gas target the base fee should stay roughly flat, so please add a line on why 3.0 is needed, ideally on the open thread as well.

tests/benchmark/stateful/bloatnet/test_transaction_types.py
danceratopz, line 174: I checked this against the spec and the max() is right: `fork.py` sets header gas_used to max(block execution gas, block state gas) per EIP-7778, and `settle_transaction_gas` attributes state charges to the state dimension even when they spill into execution gas, so N identical txs fill the binding dimension at N * max(regular, state). Please add a short comment here explaining that, and that `expected_benchmark_gas_used` uses the sum because receipts charge the sender for both dimensions. It is subtle enough that every reader will trip over it otherwise.

tests/benchmark/stateful/bloatnet/test_account_query.py
```diff
@@ -201,6 +201,7 @@ def test_account_access(
             opcode(
                 address=address_source.address_op(),
                 value=value_sent,
+                gas=Op.GAS,
```
danceratopz, line 204: Why forward gas here? Previously the call forwarded 0 gas plus the stipend on value transfers. This changes what the loop measures and adds a GAS opcode per iteration, and the BALANCE/STATICCALL/DELEGATECALL branch below keeps the default. If this is what makes the non-existing account case work, please add a comment; it is not mentioned in the PR description.

tests/benchmark/compute/scenario/test_transaction_types.py
danceratopz, line 410: This workaround needs a code comment, and the follow-up issue promised in the review thread does not seem to exist yet. Also this loop still funds two EOAs per iteration (`to` and `sender`), so the single-sender cleanup applied to `test_ether_transfers_to_precompile` was not applied here. Is that deliberate?

danceratopz (commented):
Hey, I've taken a look, but tbh I think it'd be most helpful if @jochem-brouwer re-reviewed and checked whether his questions have been addressed ("request changes" needs to be removed either way). Also, could you update the PR description @LouisTsai-Csie? As I understand:

1. The described splitting strategy is out of date and has been superseded by Mario's packing approach.
2. The RPC fix moved to #3327.

The comments below were all automated from AI, but all seem to be worth addressing before merge.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/live_client_flags.py
jochem-brouwer: I don't have a better suggestion right now, this is also non-blocking :)

jochem-brouwer (commented):
My comments have been addressed, I will remove the Request Changes here. If you want another full review then let me know (so not giving explicit approval here, but removing the Request Changes to unblock this PR 😄 👍 )

PR #3286, opened 2026-08-03
tests(binary_tree): witness cost of reading chunked code

awskii:
The existing tests pin that chunking is invisible to execution semantics. These measure what it costs a witness.

## The test

Every case calls a contract that executes the same 6 bytes — `SSTORE(slot, value)` then `STOP`. Gas, work and state touched are identical. The only variable is dead `INVALID` padding behind the `STOP`, which changes how many chunk leaves the contract occupies:

| case | code | chunks | header chunks | overflow chunks | code-zone stems |
|---|---:|---:|---:|---:|---:|
| `single_chunk` | 31 B | 1 | 1 | 0 | 0 |
| `header_full` | 3,968 B | 128 | 128 | 0 | 0 |
| `first_overflow_chunk` | 3,969 B | 129 | 128 | 1 | 1 |
| `deep_overflow` | 7,936 B | 256 | 128 | 128 | 1 |
| `max_code_size` | 24,576 B | 793 | 128 | 665 | 3 |

Padding is `INVALID`, so a client that mis-executes past the `STOP` fails loudly instead of passing for the wrong reason.

Two supporting cases:

- `test_deploy_then_read_asymmetry` — deploy in one block, call in the next. Block 1 writes 256 chunk leaves and proves none; block 2 executes 6 bytes and proves all 256.
- `test_code_introspection_needs_no_chunk` — `EXTCODESIZE` and `EXTCODEHASH` on a maximum-size contract. Both answer from the account header, so neither reads a chunk.

No fixture format for this fork carries a witness, so nothing here asserts a size. The post states are ordinary; the cost appears when a client fills or proves them.

## Measured on erigon

Context, not asserted by the tests. Erigon's EIP-8297 engine under BLAKE3, `debug_executionWitness` on the call block, against the same chain built on the Merkle-Patricia trie:

| case | MPT | binary | ratio |
|---|---:|---:|---:|
| `single_chunk` | 1,119 | 1,815 | 1.62x |
| `header_full` | 5,055 | 18,764 | 3.71x |
| `first_overflow` | 5,086 | 18,965 | 3.73x |
| `deep_overflow` | 9,023 | 35,880 | 3.98x |
| `max_code_size` | 25,663 | 108,108 | 4.21x |

Binary witness bytes at `max_code_size`, attributed by reading each leaf's own key:

| | bytes |
|---|---:|
| header chunk leaves (128) | 8,576 |
| overflow chunk leaves (665) | 44,555 |
| branches | 54,070 |
| `BASIC_DATA` + `CODE_HASH` (3 accounts) | 402 |

Branches cost as much as the leaves they bind, because each carries two 32-byte child hashes. The 402 B of account data is constant across all five sizes.

A contract past 3,968 bytes does not only add chunk leaves. Overflow chunks are keyed `zone ‖ H(code_hash ‖ tree_index) ‖ sub_index`, so each run of 256 opens a **separate code-zone stem** with its own path from the root — 3 of them at `max_code_size`, holding 256, 256 and 153 leaves.

MPT costs one times the code plus a flat 5-node proof at every size. The binary tree converges on `(67 + 68) / 31 = 4.35x`: a 67-byte chunk leaf plus a ~68-byte branch per 31 bytes of code.

These numbers assume a witness carries every chunk of a contract whose code is read. That is erigon's format, not a requirement stated by the EIP. A client that proves only the executed chunks will see the parametrisation flatten, which is worth knowing either way.

### DIFF
--- a/tests/binary_tree/eip8297_partitioned_binary_tree/test_witness_growth.py
+++ b/tests/binary_tree/eip8297_partitioned_binary_tree/test_witness_growth.py
@@ -1,0 +1,191 @@
+"""
+Witness cost of reading chunked code.
+
+The rest of this suite pins that chunking is invisible to execution
+semantics. It is not invisible to a witness. Executing one byte of a
+contract requires proving every chunk that contract occupies, because
+the code is reassembled from the tree rather than shipped alongside it.
+
+Each test below executes a fixed 6 bytes -- `SSTORE(slot, value)` then
+`STOP` -- while varying only the dead padding behind the `STOP`. Every
+parametrisation performs identical work and touches identical state.
+The only thing that changes is how many chunk leaves the contract
+occupies, and therefore how large its proof is.
+
+Measured on erigon's EIP-8297 engine (BLAKE3), a chunk leaf is 67 bytes
+(1 tag + 34 key + 32 value) and the branch binding it is ~68, so a
+contract costs about (67 + 68) / 31 = 4.35x its bytecode in witness
+bytes. Against a Merkle-Patricia client, which ships the same bytecode
+as one blob beside a short account proof, reading a 4,216-byte contract
+measured 19,666 bytes against 5,238 -- 3.75x on total witness bytes,
+for identical execution.
+
+Two consequences these cases are built to expose:
+
+  - The cost tracks code SIZE, not code EXECUTED. `deep_overflow` runs
+    the same 6 bytes as `single_chunk` and proves 793 chunks.
+  - It is asymmetric with deployment. Creating a contract proves no
+    chunk pre-state, so deploying is cheap and reading back is not;
+    `test_deploy_then_read_asymmetry` pairs the two in one chain.
+
+Nothing here asserts a witness size -- no fixture format carries one
+for this fork. These are ordinary state and blockchain tests whose
+post-states are trivially correct; the cost shows up when a client
+fills or proves them.
+"""
+
+import pytest
+from execution_testing import (
+    Account,
+    Alloc,
+    Block,
+    BlockchainTestFiller,
+    Initcode,
+    Op,
+    StateTestFiller,
+    Transaction,
+    keccak256,
+)
+
+from .helpers import sstore_then_pad
+from .spec import Spec, ref_spec_8297
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_8297.git_path
+REFERENCE_SPEC_VERSION = ref_spec_8297.version
+
+pytestmark = pytest.mark.valid_from("BinaryTree")
+
+# 31 * 128 = 3,968: the last byte that still fits the account header.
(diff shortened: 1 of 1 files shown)


CPerezz:
A conventional Merkle proof ships only the sibling and lets the verifier reconstruct the parent's preimage from the child it computed. That alone takes a branch from 68 → ~35 B, so per-chunk 135 → ~102, i.e. 4.35x → ~3.3x.

But that is nowhere close to where we want to be. This is relevant as this was inherited from original Verkle & UBT EIPs. And it didn't got much questioning when we setup the EIP assuming it was 100% worth it. (And I'm not claiming is not for the avg. case nor that it is)


I have mixed feelings here:
### Why I like the code in header
- We can read "for free" the code yielding an optimization.
- We can have a single branch in a proof if the code is under 4KB (or if the executed code it's the initial 4KB of the totality when we prove things about bigger ones)

### Why I don't like code in header
- As @awskii and Wei Han reported (see https://ethereum-magicians.org/t/not-all-state-is-equal/25508) arround 60% of the code is duplicated. We de-duplicate in the code tree(zone) but we don't in the account header (4kb for each duplicated contract deployed are indeed not de-duplicated).
- Makes the VOPS state bigger than it needs to be. As we need to hold duplicated bytecode.
- Adds more "complexity" al;though isn't really that complex. Is just more stuff to take into account.
- It blows up the statedb size with a lot of useless duplicated code
- Makes (as just showcased by @awskii and now me with my geth prototype) that the proof size explodes (for this particular worst case, but also in general (as it just pulls more hashes into the account header even if logarithmically).

One more thing to highlight here is that within the stateless-consensus team, we already discussed removing this back 1 year ago aprox (when we started working back on UBT). We brought it to SIC but never ended up reaching consensus. Thus, it wasn't done. See: https://stateless.fyi/development/sic-calls/history.html#codechunks-inclusion-in-the-binary-tree-leaf for more info about the proposal and date.

My conclusion is the same as I had back when I proposed this. And is to remove it. Back then I did not have the data to back it up. Right now I do. @awskii confirmed state size is 2x (was expected to be bigger. But not that much and I don't think intermediate nodes are the issue only, I think this duplicated 4kb pages are a big one too).
Also, we see proof sizes are balooning in part due to this. So is something we should also consider as a highly-important issue worth removing the code in header.

CPerezz:
@awskii I think we can close this as is stale according to my optimizations in Geth. See: https://github.com/CPerezz/go-ethereum/pull/2

Please LMK if you don't agree

awskii:
Closing, agreed.

The five cases are parametrised on `Spec.CODE_OFFSET` and the header/overflow split that #3310 removed, so the file no longer imports against `projects/binary-trie`. The measurement table is stale for the same reason.

Re-measuring against the current layout — every chunk content-addressed, so a clone shares its chunk set rather than proving its own copy. If the numbers still say something worth pinning I'll open it fresh.

PR #3288, opened 2026-08-03
fix(binary_trie): raise typed errors from state-root computation

### LINKED ISSUE #3255: EIP-8297: state-root computation raises raw Python errors instead of protocol-level failures
Surfaced by the EIP-8297 test work in #3246. Two cases where computing a state root raises a raw Python exception rather than failing through the protocol's own error path.

### 1. Balance at or above 2^128

`ethereum.binary_trie.embedding.encode_basic_data` carries:

```python
assert balance < U256(2) ** U256(128)  # U128 doesn't exist
```

BASIC_DATA gives `balance` a 16-byte field, so the bound is real — but it is enforced by a bare `assert` during root computation, and `AssertionError` is not an `EthereumException`. There is no protocol-level cap anywhere: `move_ether`, `create_ether` and `set_account_balance` all operate on unbounded `U256`. So a block that mints a balance past the cap **crashes the transition tool** instead of being rejected as invalid.

The EIP states no cap either — it only gives the field's offset and width.

This is unfillable at the EEST layer (t8n dies), so it is pinned only as a unit test.

### 2. Unknown code hash

`ethereum.state_pbt.State.get_code` is a bare dict subscript and raises `KeyError` for a hash absent from the store. `state_mpt.get_code` is identical in that respect — but MPT root computation never calls it, while the PBT embedding **must**, because code chunk leaves commit the code itself rather than just its hash. So an account whose code is missing from the store crashes root computation on PBT and not on MPT.

### The decision needed

- Should the balance bound be a protocol-level validity condition (rejecting the block) rather than an assertion? If balances past 2^128 are genuinely unreachable given the issuance schedule, saying so explicitly and keeping the assert as a defensive check would also be a fine answer — but that reasoning should be written down.
- Should `get_code` raise a typed error on the PBT path, given it is now reachable during root computation?

Neither is urgent, but both turn a would-be invalid block into a tooling crash, which is unpleasant to debug and would look like a client bug rather than a bad block.

Related: #3246.

CPerezz:
### Description

Computing a PBT state root could crash the transition tool with raw Python exceptions instead of failing through the protocol's own error path:

- `encode_basic_data` enforced the sixteen-byte balance field with a bare `assert`, so a block minting a balance at or above `2**128` died with `AssertionError`. It now raises `BalanceOverflowError`, an `InvalidBlock`: the block whose state cannot be committed is rejected, on the same path as a state-root mismatch. The `2**128 - 1` boundary still encodes (pinned by `test_encode_basic_data_maximum_fields`).
- `State.get_code` raised `KeyError` for a hash absent from the code store — reachable during PBT root computation because code chunk leaves commit code contents. It now raises `UnknownCodeHashError`, an `EthereumException` but deliberately **not** an `InvalidBlock`: no block content can cause it (`apply_diff_to_trie` consults `diff.code_changes` first), only a malformed pre-state. `state_mpt.get_code` gets the identical edit so both providers honor the same `PreState` contract.
- t8n's `build_result` computed the state root outside any exception handler, so even a typed `InvalidBlock` would have escaped as a crash. It now catches `InvalidBlock`, records the block exception, and reports the pre-state root — `Result.state_root` is required and non-Optional, and a rejected block leaves the chain state unchanged. `UnknownCodeHashError` is intentionally not caught there: malformed input should fail the run loudly, now with a readable message.

Unit tests pinning the old crashes were updated to pin the typed rejections, and a new test covers the block-diff path (`compute_state_root` on a diff minting an over-cap balance rejects the block while the pre-state still embeds cleanly).
After an invalid block, t8n `run()` still applies the diff to its alloc unconditionally (pre-existing behavior); a later block building on an over-cap alloc would hit the fallback recompute and raise uncaught. No current test reaches this.

### Related Issues or PRs

Fixes #3255.

### Checklist

### Cute Animal Picture

### DIFF
--- a/src/ethereum_spec_tools/evm_tools/t8n/result.py
+++ b/src/ethereum_spec_tools/evm_tools/t8n/result.py
@@ -11,7 +11,9 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from ethereum_rlp import rlp
 
 from ethereum.crypto.hash import keccak256
+from ethereum.exceptions import InvalidBlock
 from ethereum.merkle_patricia_trie import root, trie_get
+from ethereum.state import BlockDiff
 
 if TYPE_CHECKING:
     from execution_testing.client_clis.cli_types import (
@@ -79,7 +81,16 @@ def build_result(
     from execution_testing.client_clis.cli_types import Result as TestingResult
 
     diff = t8n.fork.extract_block_diff(t8n._block_state)
-    state_root = t8n.alloc.compute_state_root(diff)
+    try:
+        state_root = t8n.alloc.compute_state_root(diff)
+    except InvalidBlock as e:
+        # The executed block produced state that cannot be committed
+        # (e.g. a balance past the sixteen-byte basic data field). Such
+        # a block is rejected and leaves the chain state unchanged, so
+        # report the pre-state root.
+        if block_exception is None:
+            block_exception = f"{e}"
+        state_root = t8n.alloc.compute_state_root(BlockDiff())
 
     arguments: Dict[str, Any] = {
         "state_root": state_root,
--- a/tests/binary_trie/test_embedding.py
+++ b/tests/binary_trie/test_embedding.py
@@ -34,6 +34,7 @@ from ethereum.binary_trie.embedding import (
     remove_storage_slot,
 )
 from ethereum.binary_trie.trie import BinaryTrie, root, trie_set
+from ethereum.exceptions import BalanceOverflowError, InvalidBlock
 from ethereum.state import EMPTY_CODE_HASH as MPT_STATE_EMPTY_CODE_HASH
 
 ADDRESS = Address32(b"\x00" * 12 + b"\xaa" * 20)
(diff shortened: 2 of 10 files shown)


src/ethereum_spec_tools/evm_tools/t8n/result.py
```diff
@@ -79,7 +81,16 @@ def build_result(
...
+    try:
+        state_root = t8n.alloc.compute_state_root(diff)
+    except InvalidBlock as e:
+        # The executed block produced state that cannot be committed
+        # (e.g. a balance past the sixteen-byte basic data field). Such
+        # a block is rejected and leaves the chain state unchanged, so
+        # report the pre-state root.
+        if block_exception is None:
+            block_exception = f"{e}"
+        state_root = t8n.alloc.compute_state_root(BlockDiff())
```
kevaundray, line 93: Think this could be cherry-picked into forks/amsterdam -- should be a no-op

tests/binary_trie/test_embedding.py
```diff
@@ -752,10 +753,12 @@ def test_encode_basic_data_layout() -> None:
...
     """
-    A balance that does not fit the sixteen-byte field is rejected,
-    rather than silently truncated by `to_bytes`.
+    A balance that does not fit the sixteen-byte field raises
+    `BalanceOverflowError` -- an `InvalidBlock` -- rather than being
+    silently truncated by `to_bytes`.
     """
-    with pytest.raises(AssertionError):
+    assert issubclass(BalanceOverflowError, InvalidBlock)
+    with pytest.raises(BalanceOverflowError):
```
kevaundray, line 761: Makes sense, making the error more specific to BalanceOverflow

kevaundray (approved):

PR #3289, opened 2026-08-03
feat(forks): let development forks inherit exact-fork validity

### LINKED ISSUE #3260: valid_at-pinned tests are silently excluded when filling against a development fork
Filling the whole suite against a development fork silently drops tests, with no signal that coverage is short.

### The mechanism

`valid_from(X)` builds its fork set by comparison — `{f for f in ALL_FORKS if f >= X}` — so `BinaryTree`, which subclasses `Amsterdam`, is included. `valid_at(X)` instead resolves its arguments to *literally those forks*. `BinaryTree` is never literally named, the intersection with the selected fork set is empty, and an empty intersection means the test is simply not parametrized: no skip, no warning, nothing at default verbosity.

### Scale

13 `valid_at` usages resolve to Amsterdam and therefore *should* run against `BinaryTree` (Amsterdam semantics plus a commitment swap), but don't:

| Marker | Usages |
|---|---|
| `valid_at("EIP7981")` | 4 |
| `valid_at("EIP7976")` | 2 |
| `valid_at("Amsterdam")` | 2 |
| `valid_at("EIP8037")`, `("EIP7997")`, `("EIP7954")`, `("EIP7843")`, `("EIP7708")` | 1 each |

The remaining `valid_at` usages (`"Prague"`, `"Osaka"`, `"Homestead"`) are correctly excluded.

### The sharper problem

13 missing tests is tolerable; a port that drops tests with no signal is not, because nobody running `just binary-trie-fork tests` can tell coverage is short.

### Possible directions

1. **Report exclusions** — have a fork-scoped fill print what was deselected by exact-set markers. Additive, no blast radius, turns an invisible gap into a visible one. Cheapest real improvement.
2. **Name the fork explicitly** in those 13 markers. Works, but edits upstream tests for an experimental fork and doesn't scale — the next development fork hits the same wall.
3. **Change `ValidAt` semantics** to expand along the subclass axis. Fixes it everywhere at once, but `valid_at` currently means "exactly this fork" and is used suite-wide, so this needs an owner's decision on the marker contract.

The broader goal is that the entire suite should eventually be runnable against PBT, so whatever is chosen should generalise to any future development fork rather than special-casing `BinaryTree`.

Surfaced by #3246, which documents this as a known gap.

CPerezz:
### Description

Filling against a development fork silently dropped every test pinned with `valid_at`: the marker resolves to literally the named forks, so `BinaryTree` — which subclasses `Amsterdam` and is selected by `valid_from` — never intersected, and the tests were simply not parametrized. This PR makes those tests **fill** rather than merely reporting them as dropped (an earlier revision of this PR implemented reporting only; it was reverted in favor of inclusion).

- A fork can now be declared with `inherits_exact_fork_validity=True` (new `BaseFork.__init_subclass__` class argument + accessor), meaning it keeps its parent's execution semantics, so tests pinned to the parent remain meaningful for it.
- `ValidAt` expands each fork it resolves — named directly or via an EIP name — onto declaring subclass forks, transitively (`expand_forks_with_inherited_validity`).
- `BinaryTree` declares it, so the 14 Amsterdam-pinned `valid_at` usages fill under `--fork BinaryTree` with **zero test-file edits**, and the next development fork is a one-line declaration.

The expansion is deliberately **opt-in per fork** rather than automatic subclass widening (the issue's option 3): unconditional widening would silently pull BPO forks into every `valid_at("Osaka")`-family fill, conflicting with the deliberate `valid_for_bpo_forks` opt-in. Tests pinned to non-ancestor forks (e.g. `valid_at("Prague")`) and `valid_at_transition_to(...)` tests remain excluded — no transition fork to `BinaryTree` exists.

Verified end-to-end: `just binary-trie-fork tests/amsterdam/eip7981_increase_access_list_cost` now fills 357 fixtures including the previously-dropped `valid_at` tests (e.g. `test_access_list_token_calculation[fork_BinaryTree-...]`); before this change the directory filled only its `valid_from` tests.

Tests: adds the first pytester coverage for plain `valid_at` (it previously had none — every `valid_at*` case was `valid_at_transition_to`), covering the positive case, parent inheritance, EIP-name resolution onto the inheriting fork, and non-ancestor exclusion.

### Related Issues or PRs

Fixes #3260.

### Checklist

### Cute Animal Picture

### DIFF
--- a/docs/filling_tests/filling_tests_dev_fork.md
+++ b/docs/filling_tests/filling_tests_dev_fork.md
@@ -42,7 +42,7 @@ just binary-trie-fork [paths]
 
 A bare invocation fills only the scoped `tests/binary_tree` suite; `just binary-trie-fork tests` instead reinterprets the entire `tests/` tree against `BinaryTree`.
 
-When porting existing tests: `valid_from(...)` markers select `BinaryTree` (it subclasses `Amsterdam`), but `valid_at(...)` silently does not, so a handful of `valid_at`-pinned tests are invisible to the port with no warning at collection time.
+When porting existing tests: `valid_from(...)` markers select `BinaryTree` (it subclasses `Amsterdam`), and `valid_at(...)` markers naming `Amsterdam` — or an EIP that resolves to it — select it too, because the fork is declared with `inherits_exact_fork_validity=True` (it keeps Amsterdam's execution semantics). Tests pinned to earlier forks (e.g. `valid_at("Prague")`) remain excluded, as do `valid_at_transition_to(...)` tests: no transition fork to `BinaryTree` exists.
 
 ## Further Help
 
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py
@@ -1092,6 +1092,34 @@ class ValidBefore(ValidityMarker, mutually_exclusive=[ValidUntil]):
         return resulting_set
 
 
+def expand_forks_with_inherited_validity(
+    forks: Set[Fork | TransitionFork],
+) -> Set[Fork | TransitionFork]:
+    """
+    Add every fork that declares `inherits_exact_fork_validity` and
+    whose parent is in `forks` (transitively, so a chain of such forks
+    is followed).
+
+    This lets a development fork that keeps its parent's semantics,
+    such as `BinaryTree` over `Amsterdam`, run tests pinned to the
+    parent with `valid_at`, without naming the development fork in
+    every marker.
+    """
+    expanded = set(forks)
+    changed = True
+    while changed:
+        changed = False
+        for fork in ALL_FORKS:
+            if fork in expanded:
+                continue
+            if not fork.inherits_exact_fork_validity():
+                continue
+            if fork.parent() in expanded:
+                expanded.add(fork)
+                changed = True
+    return expanded
+
+
 class ValidAt(ValidityMarker):
     """
     Marker to specify each fork individually for which the test is valid.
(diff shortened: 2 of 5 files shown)


CPerezz:
Waiting to be properly fixed after #3279

kevaundray:
Can you clarify what this is meant to be fixing?

To me, the `valid_at` syntax is correctly only activating a test at a particular fork, and future forks whether development or _real_ wouldn't need those tests activated.

It's unclear to me, if we want to be able to test binary tries at something that can only happen in prague for example, because such a scenario is not going to happen.

marioevz:
Do you have some examples of tests that you need to run on the binary fork that are constrained by the `valid_at("amsterdam")` marker? Perhaps an alternative would be to give them the `valid_from("amsterdam")` marker instead, but not sure.

CPerezz:
@marioevz concrete case: `tests/amsterdam/eip7981_increase_access_list_cost`. All 5 modules there are `valid_at`/`valid_at_transition_to`-pinned but not a single `valid_from` test in the dir, so under `--fork BinaryTree` the whole EIP silently disappears.

If you'd still rather not touch `ValidAt`, I can re-mark 7981 and drop this — but next dev fork hits the same wall, that's what I'm trying to kill. Could also exclude `mainnet`-tagged tests from the expansion if the smokes are the concern.

kevaundray:
Hmm I still fail to understand why the solution is not to just change specific tests to be `valid_from` so they run from that fork onwards, and or add more tests that are marked `valid_from`

`valid_at` in general, seems like a reasonable marker to have if you only want to test something at that specific fork. For example, testing that before some fork, the gas price was X and after it increased to Y; ie the clients correctly increased the correct constants

CPerezz:
Sure @kevaundray fine with that. Not really married to anything here. Shall we just close this then?

kevaundray:
Yeah unless I'm missing something, I would close and maybe look for tests that we can change from `valid_at` to `valid_from`

PR #3294, opened 2026-08-04
feat(test-specs):add ssz model into the engine payload class

RazorClient:
This pr adds the SSZ model to the FixtureExecutionPayload and then uses the hasher to generate the fixtures and compare them using the hasher tool.

Part of https://github.com/ethereum/execution-specs/issues/2645

cute animal

### DIFF
--- a/packages/testing/src/execution_testing/fixtures/blockchain.py
+++ b/packages/testing/src/execution_testing/fixtures/blockchain.py
@@ -63,8 +63,11 @@ from execution_testing.exceptions import (
     ExceptionInstanceOrList,
 )
 from execution_testing.forks import (
+    Amsterdam,
+    Cancun,
     Fork,
     Paris,
+    Shanghai,
     TransitionFork,
     ssz_schema_fork_key,
 )
@@ -488,7 +491,7 @@ class FixtureExecutionPayload(ForkScopedSSZModel):
     slot_number: Uint64 | None = Field(None)
 
     __ssz_schema__ = SSZForkSchema(
-        base_fork="Paris",
+        base_fork=Paris.name(),
         base=(
             "parent_hash",
             "fee_recipient",
@@ -506,9 +509,9 @@ class FixtureExecutionPayload(ForkScopedSSZModel):
             "transactions",
         ),
         appended={
-            "Shanghai": ("withdrawals",),
-            "Cancun": ("blob_gas_used", "excess_blob_gas"),
-            "Amsterdam": ("block_access_list", "slot_number"),
+            Shanghai.name(): ("withdrawals",),
+            Cancun.name(): ("blob_gas_used", "excess_blob_gas"),
+            Amsterdam.name(): ("block_access_list", "slot_number"),
         },
     )
 
--- a/packages/testing/src/execution_testing/fixtures/blockchain.py
+++ b/packages/testing/src/execution_testing/fixtures/blockchain.py
@@ -403,15 +403,17 @@ MAX_EXTRA_DATA_BYTES = 2**5
 """Maximum ``extra_data`` length in the SSZ execution payload."""
 
 MAX_BYTES_PER_TRANSACTION = 2**30
-"""Maximum encoded size of a single transaction (consensus-specs)."""
+"""Maximum encoded size of a single transaction."""
 
 MAX_TRANSACTIONS_PER_PAYLOAD = 2**20
-"""Maximum transaction count per payload (consensus-specs)."""
+"""Maximum transaction count per payload."""
 
 MAX_WITHDRAWALS_PER_PAYLOAD = 2**4
-"""Maximum withdrawal count per payload (consensus-specs)."""
+"""Maximum withdrawal count per payload."""
 
-MAX_BLOCK_ACCESS_LIST_BYTES = MAX_BYTES_PER_TRANSACTION
+# TODO: EIP-8146 defines this as 2**23 for RLP-encoded block access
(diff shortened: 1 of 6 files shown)


packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -385,7 +399,50 @@ def genesis(cls, fork: Fork, env: Environment, state_root: Hash) -> Self:
...
+"""Maximum encoded size of a single transaction (consensus-specs)."""
+
+MAX_TRANSACTIONS_PER_PAYLOAD = 2**20
+"""Maximum transaction count per payload (consensus-specs)."""
+
+MAX_WITHDRAWALS_PER_PAYLOAD = 2**4
+"""Maximum withdrawal count per payload (consensus-specs)."""
+
+MAX_BLOCK_ACCESS_LIST_BYTES = MAX_BYTES_PER_TRANSACTION
+"""Placeholder cap for the RLP-encoded block access list byte list."""
```
spencer-tb: EIP-8146 defines `MAX_BLOCK_ACCESS_LIST_SIZE = 2**23` for the RLP-encoded BAL, should we use that instead of `MAX_BYTES_PER_TRANSACTION`?


I'm out of the loop here though, happy to keep as a place holder, but we should also add a TODO here imo.

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -385,7 +399,50 @@ def genesis(cls, fork: Fork, env: Environment, state_root: Hash) -> Self:
         return cls(**environment_values, **extras)
 
 
-class FixtureExecutionPayload(CamelModel):
+MAX_EXTRA_DATA_BYTES = 2**5
+"""Maximum ``extra_data`` length in the SSZ execution payload."""
+
+MAX_BYTES_PER_TRANSACTION = 2**30
+"""Maximum encoded size of a single transaction (consensus-specs)."""
```
spencer-tb: Nit, can we remove this from all the docstrings?
```suggestion
"""Maximum encoded size of a single transaction."""
```

packages/testing/src/execution_testing/fixtures/tests/test_blockchain_ssz.py
```diff
@@ -0,0 +1,574 @@
...
+@pytest.mark.parametrize(
+    "fork,expected_key",
+    [
+        pytest.param(Paris, "Paris", id="Paris"),
+        pytest.param(Shanghai, "Shanghai", id="Shanghai"),
+        pytest.param(Cancun, "Cancun", id="Cancun"),
+        pytest.param(Prague, "Cancun", id="Prague"),
+        pytest.param(Osaka, "Cancun", id="Osaka"),
+        pytest.param(BPO1, "Cancun", id="BPO1"),
+        pytest.param(Amsterdam, "Amsterdam", id="Amsterdam"),
```
spencer-tb: Can we add a transition fork test?
```suggestion
        pytest.param(Amsterdam, "Amsterdam", id="Amsterdam"),
        pytest.param(
            ShanghaiToCancunAtTime15k, "Cancun", id="transition"
        ),
```

spencer-tb (commented):
Some small comments, looks good to me overall :)

packages/testing/src/execution_testing/fixtures/blockchain.py
RazorClient: ah my bad

packages/testing/src/execution_testing/fixtures/blockchain.py
RazorClient: yeah a todo would be best
chaning it to the eip-8146 value and then keeping a todo

packages/testing/src/execution_testing/fixtures/tests/test_blockchain_ssz.py
RazorClient: donee

RazorClient:
Hey @spencer-tb resolved all the reviews!

spencer-tb (approved):
LGTM!

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -401,26 +460,57 @@ class FixtureExecutionPayload(CamelModel):
...
+            "timestamp",
+            "extra_data",
+            "base_fee_per_gas",
+            "block_hash",
+            "transactions",
+        ),
+        appended={
+            "Shanghai": ("withdrawals",),
+            "Cancun": ("blob_gas_used", "excess_blob_gas"),
+            "Amsterdam": ("block_access_list", "slot_number"),
```
marioevz: Let's import `Shanghai`, `Cancun`, etc, from `execution_testing.forks` and use them explicitly here.

packages/testing/src/execution_testing/forks/helpers.py
```diff
@@ -388,6 +390,25 @@ def get_fork_by_name(fork_name: str) -> Type[BaseFork] | None:
...
 
 
+def ssz_schema_fork_key(schema: SSZForkSchema, fork: Type[BaseFork]) -> str:
+    """
+    Return the newest schema fork key at or before ``fork``.
+
+    Transition forks compare as their destination fork.
+    """
+    for key in reversed(schema.forks()):
+        key_fork = get_fork_by_name(key)
```
marioevz: See comment before.

marioevz (commented):
Just two comments, thanks!

PR #3266, opened 2026-07-30
feat(tests): enhance test coverage, add mainnet test for eip-2780

LouisTsai-Csie:
### Description

Add more EIP-2780 test coverage, and mainnet test cases.

New coverage:

- `test_intrinsic_charges_authority_in_access_list`
- `test_self_transfer_with_delegated_sender`
- `test_intrinsic_decomposition_across_tx_types`

The mainnet test cases are duplicated as tests already in this directory, and
cover less than the originals do:

| `test_eip_mainnet.py` | Duplicates | Evidence |
| :--- | :--- | :--- |
| `test_transaction_gas`, all 10 | `test_value_moving_transactions`, all 10 | Same parametrization, identical gas per case; the original also asserts the EIP-7708 transfer log |
| `test_contract_creation_gas`, both | `test_value_contract_creation_tx[success-*]` | Both fill at 185,130; the original also covers `init_reverts` |
| `test_authorization_gas[existing_authority]` | `test_single_authorization_charges[sets_new_delegation]`, `test_account_write_first_write_of_authority[existing_eoa]` | All three at 66,006 |
| `test_authorization_gas[new_authority]` | `test_single_authorization_charges[creates_account]`, `test_account_write_first_write_of_authority[non_existent]` | All three at 249,606 |

I am consider removing `test_eip_mainnet.py` and marking those existing tests with the `mainnet` marker instead.

### Related Issues or PRs

issue #1940, #3217 

### Checklist

- [ ] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [ ] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3300, opened 2026-08-04
fix(test-execute): fix execute for Amsterdam

marioevz:
### Description

`uv run execute hive --fork Amsterdam` could not run a single test. This fixes the four separate problems behind that, all of them Amsterdam-specific regressions in the live-client paths (`execute` and `fill-stateful`) that the `fill` path never exercised.

This was all found while trying to verify execute tests on #3266.

Verified end-to-end with `execute hive --fork Amsterdam tests/frontier/opcodes/test_dup.py` against reth and go-ethereum (`glamsterdam-devnet-7`), and against nethermind on the final state: 16/16 passing, from 0/16 before.

#### 1. `targetGasLimit` payload attribute (`fix: Add target gas limit`)

Amsterdam adds `targetGasLimit` to the engine payload attributes. Adds the `Fork.engine_payload_attribute_target_gas_limit()` predicate (`False` from Frontier, `True` from Amsterdam) and populates the field in `PayloadAttributes.for_fork`, so `ChainBuilderEthRPC` and `ClientBackend` both send it when the fork requires it.

#### 2. Funding transactions ran out of gas (`fix(test-execute): Sender funding tx gas limit`)

The worker/EOA funding transactions used a hardcoded 200,000 gas limit. Since Amsterdam charges account-creation state gas, a plain value transfer that creates its recipient now costs **204,600** gas (21,000 intrinsic + 183,600 state gas), so every funding transaction hit out-of-gas:

```
   Prague: intrinsic=21000 state_gas=0      total=21000
Amsterdam: intrinsic=21000 state_gas=183600 total=204600
```

The receipt showed `status: 0x0` with the full gas limit consumed, so the value was never transferred and the test transaction then failed with `insufficient funds for gas * price + value: have 0`, pointing away from the real cause.

`--sender-fund-refund-gas-limit` now defaults to `None` and the value is derived from the fork (intrinsic cost + top-frame state gas for a value transfer to an empty account), matching what `fill_stateful` already did. The flag still overrides.

#### 3. Amsterdam genesis header was wrong, and nothing checked it (`fix(test-pytest): Fix amsterdam genesis`)

Two bugs in the genesis header `execute`/`fill-stateful` build for hive:

- `blockAccessListHash` used `EmptyTrieRoot` (`keccak(rlp(b""))`). A block access list hash is the hash of the RLP-encoded list, not a trie root, so an empty one is `keccak(rlp([]))` — the form `FixtureHeader.genesis` already used via `BlockAccessList().rlp_hash`.
- `slotNumber` was never set, although Amsterdam requires it, so our header had one fewer RLP element than the client's.

Both were invisible because nothing compared our genesis against the client's: the clients derive the genesis state root from the pre-allocation and ignore the `stateRoot` we write, and the initial `forkchoiceUpdated` uses the head block hash the client itself reports — so it can never disagree with itself. Setting the genesis `stateRoot` to a garbage value did not fail a single test.

`ChainBuilderEthRPC.verify_genesis_block` now compares the client's block 0 against the locally built `FixtureHeader` before the initial forkchoice update, and raises `GenesisMismatchError` with both hashes plus a field-level diff. The diff is produced by parsing the client's block back into a `FixtureHeader` so both sides are named consistently, and it warns when the re-encoded header does not reproduce the hash the client reported (i.e. when the field list may be incomplete).

#### 4. Slot numbers never advanced (`fix(test-specs): Make slotnum increment automatically`)

`env.slot_number` was only ever `0`: it defaults to `None`, `set_fork_requirements` pinned it to `0`, and nothing derived it per block. On top of that, `PayloadAttributes.for_fork` hardcoded `HexNumber(0)`, so a test setting `Environment(slot_number=N)` — as the whole EIP-7843 suite does — had the client build its block at slot 0 while the test asserted `SLOTNUM == N`.

- `PayloadAttributes.for_fork` takes the slot number from the caller; `ClientBackend` passes `env.slot_number`.
- `ChainBuilderEthRPC._payload_attributes` now takes the head block and derives timestamp, target gas limit and slot number from it, so the three call sites share one policy. When the fork requires a slot number but the head block predates the header field (fork transition) the count starts at zero; when the head block *should* carry one and does not, it fails loudly rather than silently rewinding to zero.
- `apply_new_parent` records `parent_slot_number` and `set_fork_requirements` derives `parent + 1` from it, mirroring how `excess_blob_gas` and `blob_gas_used` already work. An explicit `Block(slot_number=...)` still wins.

Fixture impact is limited to multi-block Amsterdam tests, which now advance (`[0x00] * 19` → `[0x00, 0x01, ... 0x12]`) instead of sitting at slot 0. The EIP-7843 suite is byte-identical, and so is every single-block fixture: block 1 has no `parent_slot_number`, so it still starts at 0.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/rpc/chain_builder_eth_rpc.py
@@ -141,6 +141,7 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
         self,
         *,
         next_timestamp: int,
+        next_target_gas_limit: int,
         withdrawals: List[Withdrawal] | None = None,
     ) -> PayloadAttributes:
         """Build payload attributes for a block at ``next_timestamp``."""
@@ -148,6 +149,7 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
         return PayloadAttributes.for_fork(
             next_fork,
             timestamp=next_timestamp,
+            target_gas_limit=next_target_gas_limit,
             withdrawals=withdrawals,
         )
 
@@ -217,9 +219,11 @@ class ChainBuilderEthRPC(BaseEthRPC, namespace="eth"):
             head_block_hash=head_block["hash"],
         )
         next_timestamp = int(HexNumber(head_block["timestamp"]) + 1)
+        next_gas_limit = int(HexNumber(head_block["gasLimit"]))
         next_fork = self.fork.fork_at(block_number=0, timestamp=next_timestamp)
         payload_attributes = self._payload_attributes(
-            next_timestamp=next_timestamp
+            next_timestamp=next_timestamp,
+            next_target_gas_limit=next_gas_limit,
         )
         forkchoice_updated_version = (
             next_fork.engine_forkchoice_updated_version()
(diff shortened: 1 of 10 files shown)


spencer-tb (commented):
I think we should add the target gas limit here: https://github.com/ethereum/execution-specs/blob/36e62cc6d70d63799337f989bd0da948c45827de/packages/testing/src/execution_testing/fixtures/blockchain.py#L557-L564

spencer-tb (approved):
Amazing thanks! LGTM, added that one line extra :)

PR #3302, opened 2026-08-05
perf(client): skip rlp check if rlp clearly not large enough

jochem-brouwer:
### Description
Not sure if I want this, but will dump this PR here otherwise I will forget. This calculates an upper bound on the RLP size of the block, and only if this upper bound could exceed the max we calculate the block RLP. This is an expensive operation which we only want to do if it is strictly necessary.
Will keep in draft because not sure if I want to push for this, it feels out of place, it gives nice speedups for the stateful filler though.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/specs/blockchain.py
+++ b/packages/testing/src/execution_testing/specs/blockchain.py
@@ -447,6 +447,91 @@ class Block(Header):
         return env.copy(**new_env_values)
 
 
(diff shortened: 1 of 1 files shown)


jochem-brouwer:
@LouisTsai-Csie I want this PR, but I will self-review the LLM-generated code to see if it is correct :smile: :+1: I will then open this PR for review and will ping you.

PR #3303, opened 2026-08-05
perf(test-forks): memoize per-fork gas costs

jochem-brouwer:
### Description

`Fork.gas_costs()` was recomputed on every call. Each EIP layer in the fork hierarchy implements it as `replace(super().gas_costs(), ...)`, so a single call on a late fork walks the whole chain and allocates one `GasCosts` per layer: 16 `dataclasses.replace` calls for Amsterdam, at 787 us per call.

The filler calls it in the inner loop of the benchmark gas searches (`transaction_intrinsic_cost_calculator` inside `_binary_search_iterations`),where it accounted for ~17% of `dataclasses.replace` time in a py-spy profile of a stateful fill.

`BaseForkMeta.__new__` now wraps every `gas_costs` override in a per-fork cache, so the `super()` chain runs once per fork rather than once per call. Wrapping in the metaclass rather than at each of the 18 definition sites is what makes this work: `Fork.gas_costs()` resolves through the MROto the last EIP that overrode it, and only a cache on *that* override can return before the chain runs. It also means a new EIP cannot silently opt out.

Each override keeps its own cache keyed on `cls`, so a half-assembled value from the middle of a `super()` chain can never be handed out as the final one, and no fork can serve another fork's value. The abstract declaration on `BaseFork` is left alone so `abc` still reports it unimplemented.

Sharing one object is safe because `GasCosts` is a frozen dataclass whose every field is an `int` -- there is nothing a caller could mutate. The `MEMOIZED_FORK_METHODS` docstring states that contract for anything added later, and a unit test enforces it.

Measured, per call:

| fork      | before      | after      |
|-----------|-------------|------------|
| Amsterdam |  787.41 us  |  0.19 us   |
| Prague    |  502.58 us  |  0.19 us   |
| Cancun    |  335.11 us  |  0.18 us   |
| Frontier  |   27.58 us  |  0.19 us   |

End to end on `fill --fork=Prague tests/prague/eip7623_increase_calldata_cost/` (1467 fixtures, single process, 3 runs each): 53.16 s -> 38.99 s,a 1.36x speedup.

Output is unchanged:

- `GasCosts` is byte-identical for all 35 forks, dumped and diffed against the unmodified base.
- The 1467 fixtures above are byte-identical between base and this branch (same sha256 roll-up).

### Related Issues or PRs

N/A.

### Checklist

- [x] Ran fast static checks to avoid CI fails: `ruff check`, `ruff format --check` and `mypy` are clean repo-wide (`just` is unavailable in thisenvironment; the 3 remaining mypy errors are the pre-existing `rust_pyspec_glue`/`ethash` import stubs that `--group optimized` supplies). `pytest packages/testing/src` is 1941 passed, 31 skipped, 19 xfailed, 2 xpassed.

### DIFF
--- a/packages/testing/src/execution_testing/forks/base_fork.py
+++ b/packages/testing/src/execution_testing/forks/base_fork.py
@@ -3,8 +3,10 @@
 import re
 from abc import ABCMeta, abstractmethod
 from enum import Enum, auto
+from functools import lru_cache
 from typing import (
     TYPE_CHECKING,
+    Any,
     Callable,
     ClassVar,
     Dict,
@@ -15,7 +17,9 @@ from typing import (
     Sequence,
     Set,
     Sized,
+    Tuple,
     Type,
+    cast,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


packages/testing/src/execution_testing/forks/base_fork.py
```diff
@@ -265,9 +269,67 @@ class RefundTypes(Enum):
     AUTHORIZATION_EXISTING_AUTHORITY = auto()
 
 
+MEMOIZED_FORK_METHODS = ("gas_costs",)
```
LouisTsai-Csie: Should we move it under `BaseForkMeta`?

packages/testing/src/execution_testing/forks/tests/test_forks.py
```diff
@@ -826,3 +828,105 @@ def test_oog_budget_lift() -> None:
...
+
+def _all_fork_classes() -> List[Type[BaseFork]]:
+    """
+    Return every concrete fork class.
+
+    Transition forks are excluded: they are assembled from
+    `TransitionBaseClass` rather than `BaseFork`, so they do not declare the
+    memoized methods at all and delegate to a concrete fork per block.
+    """
+    return sorted(get_forks(), key=str)
```
LouisTsai-Csie: This should be an fixture instead of a function, so we dont need to sort the array for each invocation.

Something like this:

```
@pytest.fixture
def tx_gas_limit(fork: Fork, gas_benchmark_value: int) -> int:
    """Return the transaction gas limit cap."""
    return fork.transaction_gas_limit_cap() or gas_benchmark_value
```

packages/testing/src/execution_testing/forks/tests/test_forks.py
```diff
@@ -826,3 +828,105 @@ def test_oog_budget_lift() -> None:
...
+                f"{fork}.{method_name} resolves to an uncached override"
+            )
+
+
+def test_memoized_fork_methods_return_a_stable_object() -> None:
+    """Repeat calls hand out one object, so the super() chain runs once."""
+    for fork in _all_fork_classes():
+        for method_name in MEMOIZED_FORK_METHODS:
+            first = getattr(fork, method_name)()
+            assert getattr(fork, method_name)() is first
```
LouisTsai-Csie: I am not very sure about the intention of this comparison, is it necessary?

LouisTsai-Csie (changes requested):
Some comments on the new test cases: I'm unsure if they properly test cache behavior.

packages/testing/src/execution_testing/forks/tests/test_forks.py
jochem-brouwer: Let me check, this is Claude generated, will report back :smile: :+1:

LouisTsai-Csie: Added in commit [178c11d](https://github.com/ethereum/execution-specs/pull/3303/commits/178c11df98948410891df13348370d78d5a25633)

packages/testing/src/execution_testing/forks/tests/test_forks.py
LouisTsai-Csie: Added in commit [178c11d](https://github.com/ethereum/execution-specs/pull/3303/commits/178c11df98948410891df13348370d78d5a25633)

packages/testing/src/execution_testing/forks/base_fork.py
LouisTsai-Csie: Added in commit [178c11d](https://github.com/ethereum/execution-specs/pull/3303/commits/178c11df98948410891df13348370d78d5a25633)

LouisTsai-Csie (approved):

PR #3304, opened 2026-08-05
feat(test-benchmark): add deployment script

LouisTsai-Csie:
### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/tests/benchmark/helper/account_creator.py
+++ b/tests/benchmark/helper/account_creator.py
@@ -160,7 +160,7 @@ class JochemnetPredeployContractInitcode(ContractInitcode):
 
     code_size: int
 
-    def __new__(cls, *, code_size: int = DEFAULT_CODE_SIZE) -> Self:
+    def __new__(cls, *, code_size: int = Osaka.max_code_size()) -> Self:
         """Assemble the initcode."""
         # Each MCOPY doubles the JUMPDEST-filled span (the first copy is
         # MCOPY(32, 0, 32), since 1 << 5 = 32) until it covers code_size.
@@ -329,7 +329,7 @@ class AccountCreator:
                 )
             case AccountMode.EXISTING_CONTRACT_JUMPDEST:
                 return JochemnetPredeployContractInitcode(
-                    code_size=self.code_size
+                    code_size=Osaka.max_code_size()
                 )
             case _:
                 raise ValueError(f"{self.mode.name} is not a contract")
--- a/tests/benchmark/helper/account_creator.py
+++ b/tests/benchmark/helper/account_creator.py
@@ -146,6 +146,9 @@ class JochemnetPredeployContractInitcode(ContractInitcode):
     """
     Initcode whose deployed runtime embeds its own contract ADDRESS.
 
+    If `code_size` is not supplied, Osaka max code size will be used,
+    resulting in:
+
         offset    size   contents
         ------    ----   --------------------------------
         0x0000       4   PUSH2 0x5FFF; JUMP   <- entry
--- a/tests/benchmark/helper/account_creator.py
+++ b/tests/benchmark/helper/account_creator.py
@@ -160,7 +160,7 @@ class JochemnetPredeployContractInitcode(ContractInitcode):
 
     code_size: int
 
-    def __new__(cls, *, code_size: int = Osaka.max_code_size()) -> Self:
+    def __new__(cls, *, code_size: int = DEFAULT_CODE_SIZE) -> Self:
         """Assemble the initcode."""
         # Each MCOPY doubles the JUMPDEST-filled span (the first copy is
         # MCOPY(32, 0, 32), since 1 << 5 = 32) until it covers code_size.
@@ -329,7 +329,7 @@ class AccountCreator:
                 )
             case AccountMode.EXISTING_CONTRACT_JUMPDEST:
                 return JochemnetPredeployContractInitcode(
-                    code_size=Osaka.max_code_size()
+                    code_size=self.code_size
                 )
             case _:
                 raise ValueError(f"{self.mode.name} is not a contract")
--- a/tests/benchmark/stateful/bloatnet/test_setup_contracts.py
+++ b/tests/benchmark/stateful/bloatnet/test_setup_contracts.py
@@ -24,7 +24,7 @@ from tests.benchmark.helper.account_creator import (
     AccountMode,
 )
 from tests.benchmark.helper.account_sender_receiver import (
-    DELEGATE_BASE_KEY,
+    delegate_base_key,
 )
 from tests.benchmark.helper.transactions import (
     pack_transactions_with_cost_into_blocks,
@@ -40,7 +40,6 @@ RECEIVER_CONTRACT_COUNT = int(
 )
 
 CONTRACT_MODES = [
-    AccountMode.EXISTING_CONTRACT_MINIMAL,
     AccountMode.EXISTING_CONTRACT_SAME_MAX,
     AccountMode.EXISTING_CONTRACT_DIFF_MAX,
 ]
@@ -120,6 +119,10 @@ def test_deploy_existing_contracts(
     Delegate deterministic EOAs to EXISTING_CONTRACT_DIFF_MAX receivers.
     """
     contract_modes = list(CONTRACT_MODES)
+    # One STOP byte: same initcode and CREATE2 address at every code size,
+    # so a second deployment would collide and revert the factory.
+    if code_size == Osaka.max_code_size():
+        contract_modes.append(AccountMode.EXISTING_CONTRACT_MINIMAL)
     if code_size == Amsterdam.max_code_size():
         contract_modes.append(AccountMode.EXISTING_CONTRACT_JUMPDEST)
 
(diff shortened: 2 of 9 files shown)


tests/benchmark/helper/account_creator.py
```diff
@@ -329,7 +329,7 @@ def contract_initcode(self) -> ContractInitcode:
                 )
             case AccountMode.EXISTING_CONTRACT_JUMPDEST:
                 return JochemnetPredeployContractInitcode(
-                    code_size=self.code_size
+                    code_size=Osaka.max_code_size()
```
jochem-brouwer: This change looks suspicious, it now is always Osaka.max_code_size() and not self.code_size. So if we would set this to Amsterdam (this is not necessary or required for repricing) then it would still deploy Osaka size contract

tests/benchmark/helper/account_creator.py
```diff
@@ -160,7 +160,7 @@ class JochemnetPredeployContractInitcode(ContractInitcode):
 
     code_size: int
 
-    def __new__(cls, *, code_size: int = DEFAULT_CODE_SIZE) -> Self:
+    def __new__(cls, *, code_size: int = Osaka.max_code_size()) -> Self:
```
jochem-brouwer: Why is the constant DEFAULT_CODE_SIZE replaced here by the Osaka size?

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
```diff
@@ -0,0 +1,165 @@
...
+) -> int:
+    """
+    Return the gas limit for one CREATE2 deployment, derived from the
+    intrinsic, CREATE2, and code-deposit costs with a margin for the
+    factory and initcode execution.
+    """
+    intrinsic = fork.transaction_intrinsic_cost_calculator()(
+        calldata=b"\xff" * 32 + initcode
+    )
+    create_cost = Op.CREATE2(
```
jochem-brouwer, line 68: Nit: we could take the code of the deterministic deployment factory (EIP-7997) and use that to calculate the actual gas cost

The code is:

https://eips.ethereum.org/EIPS/eip-7997#specification

`0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3`

(pretty sure we have this as constant in EELS somewhere also)

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
```diff
@@ -0,0 +1,165 @@
...
+        value=0,
```
jochem-brouwer, line 69: This is not the behavior of the CREATE2 factory, it will forward `CALLVALUE`

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
```diff
@@ -0,0 +1,165 @@
...
+    # Delegate authority i to the i-th DIFF receiver (EIP-7702).
+    delegation_sender = pre.fund_eoa()
+    intrinsic = fork.transaction_intrinsic_cost_calculator()
+    # DIFF receivers share one initcode; build it once for CREATE2 derivation.
+    diff_initcode = AccountCreator(
+        AccountMode.EXISTING_CONTRACT_DIFF_MAX
+    ).initcode
+
+    base_gas = intrinsic(authorization_list_or_count=0)
+    per_auth_gas = intrinsic(authorization_list_or_count=1) - base_gas
```
jochem-brouwer: Doesn´t this also count the base tx gas again?
Does this include that we are creating a new account? (or are we not creating a new account?)

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
```diff
@@ -0,0 +1,165 @@
...
+    gas_buffer = 100_000
+    auths_per_tx = max(
+        1, (tx_gas_limit - gas_buffer - base_gas) // per_auth_gas
```
jochem-brouwer: Note that tx_gas_limit is the EIP-7825 limit. If the fork is Amsterdam we can spend state gas on the to-be-created (?) accounts.

We don't have to optimize this situation though (my simple approach was just one delegation per tx 😂 )

jochem-brouwer (commented):
Some comments. 

For filling, we have to keep in mind if we are filling under Osaka or under Amsterdam.

I also have the feeling that the gas calculations might be off. Note that we "only" want to deploy prestate here (which we should only do once, because then we take the snapshot) so we do not need to optimize it by for instance minimizing the amount of transactions or blocks.

The `test_setup_contracts` is not really a test, I think we should move this somewhere out of the `tests/` folder (maybe in some helpers folder?)

tests/benchmark/helper/account_creator.py
LouisTsai-Csie: No, this should not use Amsterdam. The contract is already in the snapshot, deployed at the Osaka fork. It's never being deployed again. I only use it for address calculation. We should keep it as Osaka.

tests/benchmark/helper/account_creator.py
LouisTsai-Csie: For `EXISTING_CONTRACT_JUMPDEST`, we should keep the `Osaka` fork, so I set `Osaka` as the default and explicitly configure `Amsterdam` only for the other max code size scenarios we're deploying during the benchmark pre-run.

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
LouisTsai-Csie: Added in the new commit.

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
LouisTsai-Csie: This is only for calculating creation cost (`test_deploy_existing_contracts`), and since there's no value transfer when deploying max code size contracts, it should reflect the exact deployment behavior.

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
LouisTsai-Csie: Fixed in a follow-up commit.

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
LouisTsai-Csie: Fix in a follow-up commit.

jochem-brouwer:
To recap, this pre-setup should deploy for repricing the Osaka-sized contracts. For Glamsterdam, we also need these. The diff (or extra items) compared to the previous prestate generator is that we want the max sizes for:

            AccountMode.EXISTING_CONTRACT_SAME_MAX,
            AccountMode.EXISTING_CONTRACT_DIFF_MAX,
            AccountMode.EXISTING_CONTRACT_JUMPDEST,

for Glamsterdam sized contracts. For repricing, the Osaka sized ones should also exist (we can skip the EXISTING_CONTRACT_JUMPDEST one Osaka sized because it is not used for repricing)

jochem-brouwer:
This seems correct to me with the latest addition @LouisTsai-Csie a902f4b43e2b232f071c36f8e2a8f75d28a368de, thanks :smile: 

So all the contracts as described in https://github.com/ethereum/execution-specs/pull/3304#issuecomment-5371103696 are now deployed, right?

EDIT: did a re-check, see below review.

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
```diff
@@ -0,0 +1,231 @@
...
+# Number of CREATE2 receiver contracts deployed per mode. Overridable via
+# BLOATNET_RECEIVER_CONTRACT_COUNT so a smoke/plumbing run (e.g. benchmarkoor
+# pre_runs) can deploy a small set for fast iteration; defaults to the full
+# 100k benchmark set.
+RECEIVER_CONTRACT_COUNT = int(
+    os.environ.get("BLOATNET_RECEIVER_CONTRACT_COUNT", "100000")
+)
+
+CONTRACT_MODES = [
+    AccountMode.EXISTING_CONTRACT_MINIMAL,
```
jochem-brouwer: In the new setup, this will deploy both Osaka and Glamsterdam sized `MINIMAL` contracts (i.e. the 1-byte contracts twice). This should be omitted.

tests/benchmark/stateful/bloatnet/test_setup_contracts.py
```diff
@@ -0,0 +1,231 @@
...
+                initcode=initcode,
+            )
+            post[contract] = Account(nonce=1)
+
+    # Delegate authority i to the i-th DIFF receiver (EIP-7702).
+    delegation_sender = pre.fund_eoa()
+    intrinsic = fork.transaction_intrinsic_cost_calculator()
+    top_frame = fork.transaction_top_frame_gas_calculator()
+    # DIFF receivers share one initcode; build it once for CREATE2 derivation.
+    diff_initcode = AccountCreator(
```
jochem-brouwer, line 163: The EIP-7702 delegations here target either the Osaka or the Amsterdam based variant. We should deploy both, so we can run gas repricing tests on the Osaka variant and the Glamsterdam-sized tests for the "Glamsterdam benchmarks" (the full suite)

jochem-brouwer (changes requested):
GM @LouisTsai-Csie 🫡 😄 👍 

Two things we should change before we deploy these! Let me know if I should assist!

jochem-brouwer (approved):
LGTM! 😄 👍

jochem-brouwer:
@LouisTsai-Csie I think this fixes querying old state: ce497aefbc55be0fa24759d4079511aa8f4ef42a

PR #3305, opened 2026-08-05
binary(tests): vector cases for zero code chunks in the code zone

awskii:
`code_chunks_of_zero_bytes` is 62 bytes of code, so both its chunks live in the
account header. Chunks past the header are keyed by code hash rather than by
address and are written on a separate path, and no vector pins that a zero chunk
is absent there too. A client can apply the rule in one zone and not the other
and still produce a root.

The second case adds the interaction with content addressing, which is also
unpinned: `shared_bytecode_two_accounts` and `code_chunks_of_zero_bytes` each hold
on their own, but nothing says what happens when the chunk two accounts would
share is the one that has to be absent.

Both cases are aimed at the code zone, so they survive the removal of code chunks
from the account header discussed in #3286 — the second becomes the central case
rather than an edge one, since every chunk would then be content-addressed.

## Changes

- `zero_code_chunk_in_the_code_zone` — 160 chunks with an all-zero chunk at 130 in
  the code zone, and another at 5 in the header so a one-zone implementation is
  caught.
- `shared_bytecode_with_a_zero_overflow_chunk` — two accounts running that code:
  the zero code-zone chunk absent from the single leaf they would otherwise share.
- `_zeroed_chunks` helper, README case table, regenerated `binary_trie_vectors.json`.

Both roots are reproduced independently by erigon's EIP-8297 engine and by a
separate canonical-rebuild oracle in the same repo.

`test_binary_tree_fork_matches_amsterdam_modulo_known_deltas` fails on this branch
and on its base — unrelated `fork.py` drift from 42f26f1.


kevaundray:
Apologies do you mind rebasing ser?

PR #2924, opened 2026-05-27
refactor(spec-tools): use testing pydantic models in t8n

gurukamath:
## 🗒️ Description

The PR re-shapes the boundary between the testing package and
`ethereum_spec_tools` so that the `T8N` transition tool becomes a small
JSON-free engine that consumes testing pydantic types directly. Six
commits, one architectural thread per commit:

1. **`Alloc` becomes a `PreState`** (`fe3e14f9d3f`). The testing-side
   `Alloc` now implements `ethereum.state.PreState`
   (`get_account_optional`, `get_storage`, `get_code`,
   `account_has_storage`, `compute_state_root_and_trie_changes`),
   removing the need for a bespoke adapter between testing's
   allocation model and the spec's state-transition entry point. A
   `_Phase` machine (`CONSTRUCTION → LIVE → FROZEN`) governs when free
   mutations are allowed; once any `PreState` method is called, the
   alloc transitions to `LIVE` and only `apply_diff` may mutate it.

2. **`T8N` rewritten to consume testing pydantic types**
   (`c8df6498de9`). The bespoke `Alloc` / `Env` /
   `Result` / `Txs` JSON-parsing classes (`t8n_types.py`, 440 lines)
   are deleted. `T8N` no longer parses argparse output into custom
   containers — it accepts a `TransitionTool.TransitionToolData`
   directly.

3. **In-process EELS path stops round-tripping through JSON**
   (`7fb967e58e8`).
   `ExecutionSpecsTransitionTool._evaluate` hands its
   `TransitionToolData` straight to `T8N` and assembles the
   `TransitionToolOutput` from `T8N`'s in-memory state.

4. **CLI surface extracted into `t8n/cli.py`** (`e71254cc2e7`). Argparse setup, `--input.*`/`--output.*` flags,
   stdin/file readers, tracer construction from CLI flags, and the
   t8n-output JSON serialization all live in a new module. Core `T8N`
   knows nothing about argparse, JSON, or stdin/stdout.

5. **Duplicate testing-package state machinery removed**
   (`0d242564689`). The parallel `State` dataclass and
   `set_account` / `set_storage` / `state_root` / `storage_root` free
   helpers in `account_types.py` deleted. `Alloc.state_root()` now
   routes through `spec_state.state_root(_materialize_state())`. The
   testing package no longer maintains a fork of trie/state plumbing.

6. **Final cleanup pass** (`151062102f2`) - Small miscellanous clean-ups.

## 🔗 Related Issues or PRs
#1883
#1359

[...]

PR #3307, opened 2026-08-05
fix(spec-tools,testing): move evm_tools into the testing package

### LINKED ISSUE #3236: No module named 'execution_testing'
### Metadata

Commit `85aa48c742c38a2d5a876f84ebf8082a50273064`
Running off a docker image built more or less like this: 
```
FROM debian:testing-slim

RUN apt-get update -q && \ 
    apt-get install -qy --no-install-recommends \
	libssl-dev  \
	openjdk-25-jre-headless \ 
	libstdc++-13-dev \
	pipx git curl gcc python3-dev pkg-config \
	jq nano \
	libsecp256k1-dev \
  && apt-get clean

RUN git clone https://github.com/ethereum/execution-specs.git --branch forks/amsterdam --depth 1

RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s  -- -y  --profile=minimal && \
  . "$HOME/.cargo/env"; PIPX_HOME=/opt/pipx PIPX_BIN_DIR=/ pipx install ./execution-specs && \
  rm -rf /root/.rustup && rm -rf /root/.cargo
```

### What is wrong

Running goevmlab which tries to use the test runner, getting this:
```
eels error: invalid character 'T' looking for beginning of value
  | Traceback (most recent call last):
  |   File "/ethereum-spec-evm", line 6, in <module>
  |     sys.exit(main())
  |              ~~~~^^
  |   File "/opt/pipx/venvs/ethereum-execution/lib/python3.14/site-packages/ethereum_spec_tools/evm_tools/__init__.py", line 125, in main
  |     return state_test.run()
  |            ~~~~~~~~~~~~~~^^
  |   File "/opt/pipx/venvs/ethereum-execution/lib/python3.14/site-packages/ethereum_spec_tools/evm_tools/statetest/__init__.py", line 244, in run
  |     return self.run_one(self.file, fork_cache)
  |            ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/opt/pipx/venvs/ethereum-execution/lib/python3.14/site-packages/ethereum_spec_tools/evm_tools/statetest/__init__.py", line 273, in run_one
  |     result = run_test_case(
  |         test_case,
  |     ...<2 lines>...
  |         output_basedir=sys.stderr,
  |     )
  |   File "/opt/pipx/venvs/ethereum-execution/lib/python3.14/site-packages/ethereum_spec_tools/evm_tools/statetest/__init__.py", line 174, in run_test_case
  |     t8n = build_t8n_from_cli_options(t8n_options, in_stream, fork_cache)
  |   File "/opt/pipx/venvs/ethereum-execution/lib/python3.14/site-packages/ethereum_spec_tools/evm_tools/t8n/cli.py", line 314, in build_t8n_from_cli_options
  |     from execution_testing.base_types.composite_types import BlobSchedule
  | ModuleNotFoundError: No module named 'execution_testing'
    output_test.go:268: exit status 1
```

gurukamath:
### Description

`pip install`-ing this repository produces a broken `ethereum-spec-evm`: the t8n
and statetest subcommands import `execution_testing` (since #2924), which is a
uv workspace member — not a dependency of `ethereum-execution` — so standalone
installs fail with `ModuleNotFoundError` on first use.

This PR fixes the dependency inversion by moving `evm_tools` (t8n, b11r,
statetest, daemon) into the `ethereum-execution-testing` package, so all
imports point one way: testing → spec. The `ethereum-spec-evm` entry point
moves with it; `loaders/` and `utils.py` stay in `ethereum_spec_tools`.

- **CI**: new job builds all workspace wheels, installs them by explicit path
  into a clean venv, and runs a real Frontier transition — the class of
  regression that shipped #3236 is now caught before merge.
- **Docs**: new `packages/testing/README.md` (also fixes its dangling
  `readme =` metadata) with verified standalone install recipes, plus updated
  repo README, packaging docs, and an `evm_tools` reference page.

**Note for consumers**: the CLI is no longer installable from the repo root alone —
install both packages from the same clone, e.g.
`pip install ./execution-specs ./execution-specs/packages/testing`.

### Related Issues or PRs

Fixes #3236.

### Checklist

### Cute Animal Picture

UK 2024 - 1 of 1


.github/workflows/test.yaml
```diff
@@ -69,6 +69,41 @@ jobs:
...
+      - name: Smoke-test ethereum-spec-evm t8n
+        run: |
+          mkdir -p "$RUNNER_TEMP/t8n-out"
+          "$RUNNER_TEMP/wheel-venv/bin/ethereum-spec-evm" t8n \
+            --state.fork=Frontier \
+            --input.alloc=tests/evm_tools/t8n_build/alloc.json \
+            --input.env=tests/evm_tools/t8n_build/env.json \
+            --input.txs=tests/evm_tools/t8n_build/txs.json \
+            --output.basedir="$RUNNER_TEMP/t8n-out"
+          test -s "$RUNNER_TEMP/t8n-out/result.json"
```
spencer-tb: Extra check might be worth adding, to fully catch the bug from #3236
```suggestion
          test -s "$RUNNER_TEMP/t8n-out/result.json"
      # Install the spec wheel on its own: the spec package must never
      # import the testing package, or standalone spec installs break
      # again (#3236).
      - name: Import-check the spec wheel alone
        run: |
          uv venv "$RUNNER_TEMP/spec-venv"
          uv pip install --python "$RUNNER_TEMP/spec-venv/bin/python" \
            dist/ethereum_execution-*.whl
          "$RUNNER_TEMP/spec-venv/bin/python" -c "import ethereum, ethereum_spec_tools.forks, ethereum_spec_tools.loaders.fixture_loader, ethereum_spec_tools.loaders.fork_loader, ethereum_spec_tools.loaders.transaction_loader, ethereum_spec_tools.utils"
```

Justfile
```diff
@@ -215,7 +215,6 @@ spec-tools *args: (_tmp "spec-tools")
     uv run pytest \
         -n {{ xdist_workers }} \
         --basetemp="{{ output_dir }}/spec-tools/tmp" \
-        --ignore=tests/evm_tools/test_count_opcodes.py \
         "$@" \
         tests/evm_tools
```
spencer-tb: Should we move `tests/evm_tools` somewhere else in the repo? I think we should try to aim for `src/` only including the spec and `tests/` only including the spec/bench tests 

Definitely follow ups!

spencer-tb (commented):
One small thing (the diff view hides renamed file). The `Evm` docstring in `evm_trace/protocols.py` cross references the old module path, and now that the file lives in the testing package it is rendered by mkdocstrings rather than docc, [here](https://github.com/ethereum/execution-specs/blob/356f1c38fed4e74aa7dff1f0b94c8f2dba4789b0/packages/testing/src/execution_testing/evm_tools/t8n/evm_trace/protocols.py#L36-L43)

A larger thing, can we remove b11r? I remember there was talk of security using it for fuzzing but I think they should or infact do use geth's now :)

gurukamath:
> One small thing (the diff view hides renamed file). The `Evm` docstring in `evm_trace/protocols.py` cross references the old module path

Fixed the docstrings. Nice catch!!


> A larger thing, can we remove b11r? I remember there was talk of security using it for fuzzing but I think they should or infact do use geth's now :)

We could do this but I'd defer to a follow-up PR

Justfile
gurukamath: Yes. The folder name is a bit misleading. Happy to fix in a follow up

.github/workflows/test.yaml
gurukamath: Added with one upgrade. Walk the packages instead of hard coding them - future proof

.github/workflows/test.yaml
```diff
@@ -69,6 +69,66 @@ jobs:
...
+      # Install the spec wheel on its own: the spec package must never
+      # import the testing package, or standalone spec installs break
+      # again (#3236). Every module is imported, rather than a fixed
+      # list, so additions are covered automatically. The exceptions:
+      # `docc` plugins only load once the `doc` group installs docc,
+      # and importing a `__main__` would run it.
+      - name: Import-check the spec wheel alone
```
danceratopz: My friend and I tested this step against the pre-fix wheels: built the `forks/amsterdam` spec wheel, installed it alone into a clean venv, and ran this exact walk-import script. It passes (exit 0) on the broken code.

The reason: the old `execution_testing` imports in `evm_tools` were all function-scoped (`t8n/cli.py:214`, `:266`, `:314-322`), deliberately, to avoid an import cycle. So every module imports cleanly, and #3236 only fired at runtime inside `build_t8n_from_cli_options`. That is exactly the traceback in the issue. A module-level import walk cannot see this form.

The step is still worth keeping. It catches future module-level leaks (now the likely form, since the cycle that motivated function-scoping is gone) and general wheel breakage.

To ensure the invariant "`./src/` never imports `execution_testing`" does not get broken in the future, we could add an `ethereum-spec-lint` rule that rejects any such import, top-level or function-scoped. An AST check sees both forms, and it runs locally via `just static`. That might be a bit heavy though, so a quick alternative is `! grep -rn "execution_testing" src/`, which passes as of this branch.

.github/workflows/test.yaml
```diff
@@ -69,6 +69,66 @@ jobs:
...
+      # argparse without ever reaching the imports that t8n needs. The output
+      # basedir is emptied before the run, so keep it out of the source tree.
+      - name: Smoke-test ethereum-spec-evm t8n
+        run: |
+          mkdir -p "$RUNNER_TEMP/t8n-out"
+          "$RUNNER_TEMP/wheel-venv/bin/ethereum-spec-evm" t8n \
+            --state.fork=Frontier \
+            --input.alloc=tests/evm_tools/t8n_build/alloc.json \
+            --input.env=tests/evm_tools/t8n_build/env.json \
+            --input.txs=tests/evm_tools/t8n_build/txs.json \
```
danceratopz: Now that t8n lives in the testing package, I think these new fixtures should be there too?

Suggested home: `packages/testing/src/execution_testing/evm_tools/tests/fixtures/t8n_build/`. This follows the existing convention in the package (`client_clis/tests/fixtures/`, `specs/tests/fixtures/`, `tools/tests/test_filling/fixtures/`).

These three lines are the only references repo-wide. The `exclude = ["*tests*"]` rule keeps the files out of the wheel, and this job reads them from the checkout, so only the paths here change.

It would also leave `tests/evm_tools/` holding nothing but spec-tools tests, which makes the `tests/spec_tools/` rename follow-up a pure `git mv`.

(One caution: not the top-level `execution_testing/fixtures/`, which is the fixture-format code subpackage.)

packages/testing/src/execution_testing/evm_tools/t8n/cli.py
```diff
@@ -307,10 +306,6 @@ def build_t8n_from_cli_options(
     testing pydantic types, bundles them into a ``TransitionToolData``,
     builds the tracer group, and hands them to ``T8N``.
     """
-    # Function-scoped imports: ``execution_testing/__init__`` eagerly
-    # imports ``.specs`` which transitively imports ``client_clis``,
-    # which imports ``ExecutionSpecsTransitionTool`` — top-level imports
-    # from ``execution_testing`` would cycle back into spec-tools.
     from execution_testing.base_types.composite_types import BlobSchedule
```
danceratopz, line 309: The import-cycle note that lived here was deleted, but two comments in this file still point at it: lines 212 and 264 both read `# Function-scoped: see import-cycle note in build_t8n_from_cli_options`.

They now reference nothing, and these bare function-scoped imports invite a future hoist-to-module-level cleanup.

I would either restore a short note here (updated for the new home) or repoint the two comments. `result.py:24-28` kept and updated its equivalent note in this PR and reads well as the model.

Justfile
danceratopz: A quick win for clarity, potentially in this PR: rename `tests/evm_tools` to `tests/spec_tools`, matching the just recipe name (`spec-tools`). After the move the directory only holds spec-tools tests anyway (`test_lint.py`, `test_new_fork.py`, `test_docc_shards.py`).

The rename touches three places: the recipe at `Justfile:219`, the fill default ignores at `packages/testing/src/execution_testing/cli/pytest_commands/fill.py:143`, and the three `t8n_build` paths in the packaging job. If the `t8n_build` inputs move to the testing package (see the [comment above on the workflow](https://github.com/ethereum/execution-specs/pull/3307/changes#r3725869393)), the last one disappears and this becomes a `git mv` plus two lines.

pyproject.toml
```diff
@@ -272,7 +267,6 @@ ethereum-spec-lint = "ethereum_spec_tools.lint:main"
 ethereum-spec-sync = "ethereum_spec_tools.sync:main"
 ethereum-spec-new-fork = "ethereum_spec_tools.new_fork.cli:main"
 ethereum-spec-patch = "ethereum_spec_tools.patch_tool:main"
-ethereum-spec-evm = "ethereum_spec_tools.evm_tools:main"
 whitelist = "ethereum_spec_tools.whitelist:main"
```
danceratopz: This entry point is broken in a standalone wheel install: `ethereum_spec_tools.whitelist` is not in the `[tool.setuptools] packages` list, so it never ships and the script fails with `ModuleNotFoundError`. This is the same class of packaging bug as #3236.

What do you think about removing this line. It is a repo-bound dev tool (it edits the checkout's `whitelist.txt`) and its generic `whitelist` name lands in every consumer venv's `bin/`. The only consumer is `Justfile:67`, which can become `uv run python -m ethereum_spec_tools.whitelist "$@"`. `just whitelist <word>` stays the same.

vulture_whitelist.py
```diff
@@ -203,3 +166,40 @@
...
+ForkLoad.has_compute_requests_hash
+ForkLoad.has_withdrawal
+ForkLoad.has_slot_number
+
+# src/ethereum_spec_tools/utils.py - used by the evm_tools CLI
+FatalError
+find_fork
+resolve_fork
+get_supported_forks
+get_stream_logger
```
danceratopz: These entries are needed only because their consumers moved out of vulture's scope: `just deadcode` scans `src/` only, so the moved `evm_tools` code is no longer dead-code-checked at all.

Extending the recipe to `packages/testing/src/execution_testing/evm_tools` would let some of these 16 entries go again and keep the moved code covered. This could be done in the scope of this PR, imo.

One heads-up on the cost, verified locally: the extended scan reports 15 new items at 60% confidence, almost all EIP-3155 serialization field names in `evm_trace/eip3155.py` (`gasCost`, `memSize`, ...), plus `daemon.py` `log_request` (a `BaseHTTPRequestHandler` override) and one `pytestmark`. So it roughly trades these spec-side entries for testing-side ones, but real dead code in `evm_tools` becomes detectable again.

.github/workflows/test.yaml
```diff
@@ -69,6 +69,66 @@ jobs:
           EOF
           uvx --from actionlint-py actionlint
 
+  packaging:
```
danceratopz, line 72: All the sibling jobs gate on `needs: static`; this is the only one that does not. I think it's worth gating here, too.

docs/filling_tests/transition_tool_support.md
```diff
@@ -5,7 +5,7 @@ The following transition tools are supported by the framework:
 | Client | `t8n` Tool | Tracing Support |
 | -------| ---------- | --------------- |
 | [ethereum/evmone](https://github.com/ethereum/evmone) | `evmone t8n` | Yes |
-| [ethereum/execution-specs](https://github.com/ethereum/execution-specs) | [`ethereum-spec-evm t8n`](https://github.com/ethereum/execution-specs/tree/a48e0b381d5225a6c3de2d06cd9ee7ae0b6ca9bb/src/ethereum_spec_tools/evm_tools/t8n) | Yes |
+| [ethereum/execution-specs](https://github.com/ethereum/execution-specs) | [`ethereum-spec-evm t8n`](https://github.com/ethereum/execution-specs/tree/forks/amsterdam/packages/testing/src/execution_testing/evm_tools/t8n) | Yes |
```
danceratopz: Can we permalink/SHA pin this? This points at `tree/forks/amsterdam/...`, which will break at the next fork rollover when the default branch moves on.

packages/testing/src/execution_testing/evm_tools/tests/fixtures/t8n_build/alloc.json
danceratopz, line 1: Now that t8n lives in the testing package, I think these new fixtures should be there too? See
https://github.com/ethereum/execution-specs/pull/3307/changes#r3725869393

packages/testing/README.md
```diff
@@ -0,0 +1,37 @@
+# The `ethereum-execution-testing` Package
+
+Test generation and execution framework for the [Ethereum Execution Layer Specifications (EELS)](https://github.com/ethereum/execution-specs), derived from [ethereum/execution-spec-tests](https://github.com/ethereum/execution-spec-tests).
```
danceratopz: I don't think we need a reference to the archived EEST repo here.
```suggestion
Test generation and execution framework for the [Ethereum Execution Layer Specifications (EELS)](https://github.com/ethereum/execution-specs).
```

spencer-tb (approved):
LGTM! :D

.github/workflows/test.yaml
danceratopz, line 72: Thanks a lot for adding these additional checks, Guru! 

Can we plz also add easy local verification of these steps by adding these checks to the `Justfile` instead? In particular, workflow-yaml-inline scripts make it hard to verify locally.

Every other test job here is a one-liner delegating to a just recipe; this is the only one with inline logic, including ~18 lines of Python in a heredoc that no linter or type checker ever sees. Moving the whole job into a `just test-packaging` recipe would match the other nine jobs, make the checks runnable locally before pushing (#3236 is exactly the class of bug you want to catch at the desk), and give the embedded Python a lintable home in `.github/scripts/`, which already has a tested-scripts convention via `just test-ci-scripts`.

Suggested shape:

```just
# --- Packaging ---

# Build every workspace wheel into .just/dist
[group('packaging')]
build-wheels:
    uv build --wheel --all-packages --out-dir "{{ output_dir }}/dist"

# Smoke-test the built wheels: clean-venv install, real t8n run, spec-wheel-alone import check
[group('packaging')]
test-packaging: build-wheels
    # 1. install both wheels by explicit path into a fresh venv
    # 2. run the Frontier t8n transition, assert result.json
    # 3. install the spec wheel alone, run .github/scripts/import_check.py
```

The `packaging` group is forward looking: a future `publish-wheels` or a version-lockstep check would land in the same group, each independently runnable and clustered together in `just --list`. Note the doc comment above each recipe must be a single line; just shows only the line immediately above the recipe in the listing.

The job then reduces to checkout, setup-uv, `run: just test-packaging` (plus `needs: static`, see the other comment). One tradeoff: the four named steps collapse into one, so per-step timing and failure attribution moves from the UI into the log. The step comments here (why `--all-packages`, why explicit wheel paths, why a real transition instead of `--help`) are good and should move into the recipe with the logic.

src/ethereum_spec_tools/loaders/transaction_loader.py
```diff
@@ -19,7 +19,7 @@
     hex_to_u256,
     hex_to_uint,
 )
-from ethereum_spec_tools.evm_tools.utils import parse_hex_or_int
+from ethereum_spec_tools.utils import parse_hex_or_int
```
danceratopz, line 22: Suggestion (anchored here because the rest of the loader move is a pure rename in the diff): let `loaders/` and `utils.py` follow `evm_tools` all the way into the testing package, in this PR, rather than stopping at `ethereum_spec_tools` top level.

Justification: After the move they have zero consumers left in the spec package: every user of `loaders/` and `utils.py` is in `packages/testing` (11 import sites) or `tests/json_loader` (one import, `Load` in `helpers/load_blockchain_tests.py`). The only spec-internal use is this file importing `utils.parse_hex_or_int`, and it moves with them. `forks.py` is different and should stay, it is real spec API.

Why now? The import path already breaks in this PR either way: on base these lived at `ethereum_spec_tools.evm_tools.loaders` and `evm_tools/utils.py`, so any external consumer has to update regardless. Landing them at `execution_testing.evm_tools.loaders` and `evm_tools/utils.py` now means one break instead of two.

---

What the full move buys, beyond the cleaner boundary:

- 13 of the 16 new vulture whitelist entries go away outright (`Load`, the seven `ForkLoad.has_*`, the five `utils` helpers leave `src/`); the remaining three are covered by extending the deadcode scan (other comment).
- The `"ethereum_spec_tools.loaders"` line added to the setuptools packages list can be dropped again.
- The new fork-selection branch in `tests/json_loader/helpers/select_tests.py:79-84` can be reverted; it exists only because these files stayed behind.
- The PR's story simplifies from "move `evm_tools`, but split out its loaders and utils" to "move `evm_tools` wholesale".

Cost is the same mechanical class as the rest of the PR: five files moved plus roughly a dozen import-site rewrites.

danceratopz: The layout in the three states, for the visual argument.

1. Base (`forks/amsterdam`), everything under one roof:

```
src/ethereum_spec_tools/
├── __init__.py, forks.py, patch_tool.py, sync.py, py.typed
└── evm_tools/
    ├── __init__.py, __main__.py, daemon.py
    ├── utils.py                                    ◄ inside evm_tools
    ├── b11r/        (__init__.py, b11r_types.py)
    ├── loaders/     (__init__.py, fixture_loader.py,   ◄ inside evm_tools
    │                 fork_loader.py, transaction_loader.py)
    ├── statetest/   (__init__.py)
    └── t8n/         (__init__.py, block_environment.py, cli.py,
                      result.py, evm_trace/ ×5)
```

2. This PR as it stands, `evm_tools` moves but `loaders/` and `utils.py` are split out and promoted to spec-tools top level:

```
src/ethereum_spec_tools/
├── __init__.py, forks.py, patch_tool.py, sync.py, py.typed
├── loaders/     (__init__.py, fixture_loader.py,   ◄ promoted, stayed behind
│                 fork_loader.py, transaction_loader.py)
└── utils.py                                        ◄ promoted, stayed behind

packages/testing/src/execution_testing/
└── evm_tools/
    ├── __init__.py, __main__.py, daemon.py
    ├── b11r/        (__init__.py, b11r_types.py)
    ├── statetest/   (__init__.py)
    ├── t8n/         (__init__.py, block_environment.py, cli.py,
    │                 result.py, evm_trace/ ×5)
    └── tests/       (×3, moved from tests/evm_tools/)
```

3. With the suggested move, `evm_tools` moves wholesale and keeps its base-internal structure at the new address:

```
src/ethereum_spec_tools/
├── __init__.py
├── forks.py                                        ◄ stays: real spec API
└── patch_tool.py, sync.py, py.typed

packages/testing/src/execution_testing/
└── evm_tools/
    ├── __init__.py, __main__.py, daemon.py
    ├── utils.py                                    ◄ back where it was
    ├── b11r/        (__init__.py, b11r_types.py)
    ├── loaders/     (__init__.py, fixture_loader.py,   ◄ back where it was
    │                 fork_loader.py, transaction_loader.py)
    ├── statetest/   (__init__.py)
    ├── t8n/         (__init__.py, block_environment.py, cli.py,
    │                 result.py, evm_trace/ ×5)
    └── tests/       (×3)
```

Why the picture matters: state 3 is not a new reorganization on top of the PR; it is the PR's own move applied uniformly. `evm_tools` in state 3 has the identical internal shape it had on base (`utils.py` and `loaders/` in their old positions relative to `t8n/`, `b11r/`, `statetest/`); only the package prefix changed. State 2 invents a layout that never existed before, `loaders/` and `utils.py` at spec-tools top level. The leftover `src/ethereum_spec_tools/` in state 3 is pure spec-serving tooling: fork enumeration (`forks.py`), lint, new-fork scaffolding, docs, sync.

danceratopz (commented):
Thanks @gurukamath, the move looks clean and (bar 2. below) complete!!!

Details are in the inline comments. Two larger changes that could/should go in this PR:

1. Move the new packaging job into a `just test-packaging` recipe (see the comment on the job).
2. Let `loaders/` and `utils.py` follow `evm_tools` into the testing package (see the comment on `transaction_loader.py`). This PR already breaks their import path, so one break beats two.

Follow-up: opened #3326 for the unrelated but never-runnable `test_count_opcodes.py` that was spotted during review.

.github/workflows/test.yaml
danceratopz: Justification about the move/rename of `./tests/evm_tools/` to `./tests/spec_tools/` here https://github.com/ethereum/execution-specs/pull/3307#discussion_r3725937779

Justfile
gurukamath: updated

.github/workflows/test.yaml
gurukamath: added a grep gaurd against imports

.github/workflows/test.yaml
gurukamath: Moved these fixtures

pyproject.toml
gurukamath: Done

docs/filling_tests/transition_tool_support.md
gurukamath: done

vulture_whitelist.py
gurukamath: Updated

packages/testing/README.md
gurukamath: done

.github/workflows/test.yaml
gurukamath: Moved these to Justfile

src/ethereum_spec_tools/loaders/transaction_loader.py
gurukamath: I think this one is genuinely a spec_tool which the t8n happens to use. `tests/json_loader` imports `Load`, and having that depend on the testing framework seems not right to me. Also, if we move this, that import would first execute `execution_testing/__init__`, which eagerly imports the whole framework.

packages/testing/src/execution_testing/evm_tools/t8n/cli.py
gurukamath: updated

.github/workflows/test.yaml
gurukamath: The jobs that gate run on expensive self-hosted runners where fail-fast pays; packaging is a cheap ubuntu job, and parallel gives faster signal on green pushes. I'd be in favour of keeping this as is

packages/testing/src/execution_testing/evm_tools/tests/fixtures/t8n_build/alloc.json
gurukamath: Moved

src/ethereum_spec_tools/loaders/transaction_loader.py
marioevz: The only part of `loaders` I can chime in is `fixture_loader.py`, since the ideal solution is that both `execution_testing` and the specs import what currently is `execution_testing.fixtures` and work from there, but I think this requires a bigger refactoring than what we have in hand just now.

marioevz (approved):
LGTM, thanks for this effort!

I would just push myself a minor fix to enable `packages/testing/src/execution_testing/evm_tools/tests/test_count_opcodes.py` and then I will proceed to merge.

PR #3308, opened 2026-08-05
feat(test-cli): add JSON output and optimization analysis to `groupstats`

danceratopz:
### Description

Improve the `groupstats` pre-allocation group analysis utility so it is useful for both human inspection and programmatic/agent-driven analysis.

The command now supports:

- `--output rich|json` to switch between the existing rich console view and machine-readable JSON.
- `--low-test-count` to define which groups should be treated as optimization targets.
- `--limit` to bound candidate result lists, with `0` meaning no limit.
- `--include-test-ids` to opt into full test ID lists.
- `--include-group-details/--exclude-group-details` to control per-group output size.
- `--compact` to omit verbose top-level maps while keeping totals, distributions, fork stats, and optimization candidates.
- `--match-test-id` and `--match-test-id-regex` to focus analysis on matching tests before recomputing rich or JSON stats.
- `--exclude-test-id` and `--exclude-test-id-regex` to remove known-noisy tests before recomputing rich or JSON stats.

The command help now includes an agent/programmatic example and points callers at the JSON `optimization` object, especially `candidate_buckets`, as the first place to inspect for possible larger group opportunities.

The analyzer now reads pre-alloc group JSON directly into lightweight records instead of loading full fixture models. It supports both the legacy `environment`-bearing group format and the newer format introduced in #3390 (derived `genesis` header with the pre-state dependent fields excluded from grouping, hex-encoded `chainId`), and reports ranked optimization data for:

- low-count groups,
- candidate buckets sharing fork, chain ID, group salt, and environment,
- test-function candidates,
- module candidates.

The rich output keeps the existing at-a-glance tables and adds singleton/low-count summaries plus a candidate-bucket table.

Validated via:

- `uv run pytest packages/testing/src/execution_testing/cli/tests/test_show_pre_alloc_group_stats.py`
- `just static`
- `uv run groupstats --help`
- Manual validation against the cached `tests%40v20.0.1` `blockchain_tests_engine_x/pre_alloc` fixture directory in both rich and JSON modes.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

A judgemental cat in a box, inspecting your pre-alloc groups

### DIFF
--- a/packages/testing/src/execution_testing/cli/show_pre_alloc_group_stats.py
+++ b/packages/testing/src/execution_testing/cli/show_pre_alloc_group_stats.py
@@ -1,16 +1,16 @@
 """Script to display statistics about pre-allocation groups."""
 
+import hashlib
+import json
+import re
 from collections import defaultdict
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, List, Set, Tuple
+from typing import Any, Dict, List, Tuple
 
 import click
-from pydantic import Field
 from rich.console import Console
 from rich.table import Table
-
-from execution_testing.base_types import CamelModel
-from execution_testing.fixtures import PreAllocGroups
 
 
 def extract_test_module(test_id: str) -> str:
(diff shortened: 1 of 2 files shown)


packages/testing/src/execution_testing/cli/show_pre_alloc_group_stats.py
```diff
@@ -42,6 +42,198 @@ def extract_test_function(test_id: str) -> str:
...
+
+    return GroupRecord(
+        hash=file.stem,
+        path=str(file),
+        tests=len(test_ids) if test_ids else data.get("testCount", 0),
+        accounts=len(pre) if pre else data.get("preAccountCount", 0),
+        fork=fork,
+        chain_id=data.get("chainId", 1),
+        group_salt=data.get("groupSalt"),
+        environment_hash=_stable_json_hash(data.get("environment", {})),
```
spencer-tb: After #3390, PreAllocGroup no longer serializes environment.

Falling back to the genesis header minus its pre-state dependent fields (stateRoot and the block hash) keeps the bucket key faithful for the new format while old cached folders keep using environment:
```suggestion
        environment_hash=_stable_json_hash(
            data.get(
                "environment",
                {
                    key: value
                    for key, value in data.get("genesis", {}).items()
                    if key not in ("stateRoot", "hash")
                },
            )
        ),
```

packages/testing/src/execution_testing/cli/show_pre_alloc_group_stats.py
```diff
@@ -42,6 +42,198 @@ def extract_test_function(test_id: str) -> str:
...
+    if not isinstance(fork, str):
+        fork = str(fork)
+
+    return GroupRecord(
+        hash=file.stem,
+        path=str(file),
+        tests=len(test_ids) if test_ids else data.get("testCount", 0),
+        accounts=len(pre) if pre else data.get("preAccountCount", 0),
+        fork=fork,
+        chain_id=data.get("chainId", 1),
```
spencer-tb: #3390 also changed `chain_id` to `ZeroPaddedHexNumber`, so new format files serialize it as "0x01". This assignment then stores a string in `GroupRecord.chain_id: int`, and the JSON output and bucket keys carry "0x01" where old-format files yield 1. 

```suggestion
        chain_id=int(str(data.get("chainId", 1)), 0),
```

spencer-tb (approved):
LGTM!

PR #3312, opened 2026-08-05
ci: pass composite action inputs via env before shell

SashaMIT:
## Summary
`fetch-binary` interpolated repo/version/filename/checksum inputs directly into `curl`/`mv`/`chmod` in a `run:` script. `cache-docker-images` interpolated `inputs.images` into a `for` loop. Move those values into `env:` and expand shell variables instead.

## Test plan
- [ ] Workflows using `fetch-binary` still download, checksum, and install the binary
- [ ] Docker image cache warm/cold paths still pull/save listed images

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/actions/cache-docker-images/action.yaml
+++ b/.github/actions/cache-docker-images/action.yaml
@@ -28,9 +28,11 @@ runs:
 
     - name: Pull and save Docker images
       shell: bash
+      env:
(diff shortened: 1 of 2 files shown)


spencer-tb:
Thanks for the PR but closing for now!

PR #3056, opened 2026-06-26
refactor(spec-specs): move priority-fee check to validate_transaction

jevinjojo:
## 🗒️ Description
Moves the stateless check `tx.max_fee_per_gas < tx.max_priority_fee_per_gas`
from `check_transaction` (stateful layer) to `validate_transaction`
(stateless layer) in all fee-market forks (london through amsterdam,
including the glacier and BPO forks).

- The check only compares two fields on the transaction and needs no block
context, so it belongs in the stateless layer. The remaining checks in that
block use `block_env.base_fee_per_gas` and correctly stay in
`check_transaction`.

- In `validate_transaction` the check is guarded by an
`isinstance(tx, FeeMarketCapableTransaction)` check so legacy transactions are
unaffected. The now-unused import is removed from `fork.py`.

- Mirrors client behavior: nethermind runs this check statelessly in
`TxValidator.IsWellFormed` during block validation; geth runs it statelessly
at txpool admission (and again in `preCheck` during state transition).

- Filled the tests covering `PRIORITY_GREATER_THAN_MAX_FEE_PER_GAS`
(`tip_too_high`, `set_code_transaction_fee_validations`) before and after:
fixtures are identical. Pre-cancun forks are not fillable, so those were
smoke-tested directly: every fork's `validate_transaction` rejects inverted
fee caps with `PriorityFeeGreaterThanMaxFeeError` and accepts sane ones.

## 🔗 Related Issues
- Fixes #2976
## ✅ Checklist

- [x] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://eest.ethereum.org/main/getting_started/code_standards/) and [Enabling Pre-commit Checks](https://eest.ethereum.org/main/dev/precommit/):
    ```console
    just static
    ```
- [x] All: PR title adheres to the [repo standard](https://eest.ethereum.org/main/getting_started/contributing/?h=contri#commit-messages-issue-and-pr-titles) - it will be used as the squash commit message and should start `type(scope):`.
- [x] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.

Issue #2230, opened 2026-02-18
Use historically-accurate per-fork block gas limits


danceratopz:
## Problem

All forks currently use the same default block gas limit when filling:

```python
CURRENT_MAINNET_BLOCK_GAS_LIMIT = 60_000_000        # 60M (Osaka, EIP-7935)
DEFAULT_BLOCK_GAS_LIMIT = CURRENT_MAINNET_BLOCK_GAS_LIMIT * 2  # 120M
```

This means a test marked `@pytest.mark.valid_from("Istanbul")` filled for Berlin produces fixtures with a 120M block gas limit - but Berlin mainnet never exceeded ~15M.

Testing a client against Berlin with 120M gas is unrealistic and doesn't reflect the conditions clients actually operated under during that fork.

Historically, the gas limit used for consensus tests was that important, but as gas limits increase, it seems to fill older forks with the ever increasing gas limits.

### Approximate historical mainnet gas limits

We should double-check these!

| Fork / Range                                 | Era             | Approx. mainnet gas limit (max per block)                      |
| -------------------------------------------- | --------------- | -------------------------------------------------------------- |
| Frontier -> Homestead                        | 2015-2016       | Very low initially -> ~4.7M by Homestead (raised gradually)    |
| Byzantium -> Constantinople / St. Petersburg | 2017-2019       | ~4.7M -> ~10M (raised gradually over time)                     |
| Istanbul                                     | 2019-2020       | ~10M -> 12.5M (12.5M becomes common in mid-2020)               |
| Berlin                                       | 2021            | Launched ~12.5M -> ~15M shortly after (miner/validator voting) |
| London -> Paris (Merge)                      | 2021-2022       | Max 30M (EIP-1559 elasticity; target 15M)                      |
| Shanghai -> early 2025                       | 2023-early 2025 | Max 30M (target 15M)                                           |
| 2025 (pre-Fusaka)                            | 2025            | Max 36M -> 45M (raised during 2025)                            |
| Fusaka / Osaka live                          | Dec 2025+       | Max ~60M (EIP-7935 client default; target ~30M)                |

The 2x multiplier (`DEFAULT_BLOCK_GAS_LIMIT = mainnet * 2`) was introduced for testing headroom. With per-fork values we should decide whether to keep the 2x or use the actual mainnet value.

## Proposed change

Add a per-fork gas limit mapping, e.g. in `EnvironmentDefaults` or as a `Fork` method:

```python
# Option A: Fork method
[...]

PR #3314, opened 2026-08-05
fix(spec-specs): validate the gas limit on the London fork block

aNaToMiCaLL:
### Description

`validate_header` does not check the gas limit on the London fork block.

**Context.** Every header must satisfy the gas-limit relation: the limit must lie
strictly within `parent_gas_limit +/- parent_gas_limit // 1024` and be at least
`LIMIT_MINIMUM`. `check_gas_limit` implements it. On the fork block EIP-1559
scales the comparison base — the gas limit doubles while the gas target is held
constant. EIP-1559's own reference implementation does exactly this:

```python
if INITIAL_FORK_BLOCK_NUMBER == block.number:
    parent_gas_target = self.parent(block).gas_limit
    parent_gas_limit = self.parent(block).gas_limit * ELASTICITY_MULTIPLIER
```

and then asserts `block.gas_limit < parent_gas_limit + parent_gas_limit // 1024`,
`> parent_gas_limit - parent_gas_limit // 1024`, and `>= 5000` against that
scaled value. The Yellow Paper specifies the same thing as
`P(H)_{H_l'} = P(H)_{H_l} * rho` with `rho = 2` (`Paper.tex:655`, `:673-686`,
`:710-713`), and states it in prose at `:688`: *"the value of the parent block
gas limit for the purpose of validating the current block's gas limit is
modified at the London fork block by multiplying it by the elasticity
multiplier."*

**Anomaly.** In `london/fork.py`, `check_gas_limit` has exactly one call site:
line 257, inside `calculate_base_fee_per_gas`. `validate_header` skips that call
on the fork block —

```python
if header.number != FORK_CRITERIA.block_number:
    expected_base_fee_per_gas = calculate_base_fee_per_gas(...)
```

— substituting `INITIAL_BASE_FEE` for the computed base fee. A mandated validity
conjunct is thereby implemented only as a side effect of an unrelated
computation, and on the one block where that computation is skipped, **none of
the three gas-limit clauses is enforced**:

```
H_l <  2*P_l + floor(2*P_l/1024)
H_l >  2*P_l - floor(2*P_l/1024)
H_l >= 5000
```

Separately, the fork-block scaling is constructed nowhere in the codebase;
`ELASTICITY_MULTIPLIER` feeds only `parent_gas_target`.

**Impact.** On block 12965000 a header with any gas limit whatsoever — including
`1` — passes `validate_header`. The bypass is unique to London: from Paris
onward `validate_header` always calls `calculate_base_fee_per_gas`, so the check
is always reached (verified in `paris`, `shanghai`, `cancun` and `prague`, none
of which special-case the fork block in `validate_header`). Because the
canonical mainnet block at that height is long final, the consequence is confined
to EELS as a reference implementation and to consumers relying on it to reject
malformed historical headers. Per `SECURITY.md` this is not a serious issue —
production clients implement the EIP-1559 fork-block rule correctly — so it is
filed normally rather than reported privately.

**Resolution.** Call `check_gas_limit` explicitly on the fork block against
`parent_header.gas_limit * ELASTICITY_MULTIPLIER` and raise `InvalidBlock` on
failure. The `else` branch is the previous code path verbatim. One file changed.

**Verification.** A header at number 12965000 with a Berlin parent
(`gas_limit = 15,000,000`) and `gas_limit = 1` is accepted before the change and
rejected after. Two controls rule out vacuity: an off-by-one `base_fee_per_gas`
on the same path does raise `InvalidBlock`, and the identical gas limit at
LONDON+1 is rejected through `validate_header -> calculate_base_fee_per_gas`.

The fix is then pinned against real mainnet data. Block 12965000 has
`gas_limit = 30,029,122`, with parent 12964999 at `15,029,237`. Scaled:
`15,029,237 * 2 = 30,058,474`, adjustment delta
`30,058,474 // 1024 = 29,353`, so the exclusive lower bound is
`30,058,474 - 29,353 = 30,029,121`. The real block sits **exactly one gas
above** that bound and is still accepted; the same header **fails** the unscaled
check outright. The scaling factor is therefore load-bearing and exactly right.
The boundary is exercised in both directions: `30,029,121` — the bound itself,
which `check_gas_limit` rejects since its comparison is `<=` — is rejected, and
`30,029,122` is accepted.

`fill --fork=London tests/london tests/berlin tests/frontier`: **2643 passed, 4
deselected** both before and after the change.

`just static` passes in full (exit 0) — `typecheck` (mypy clean across 4287
source files), `lint-spec`, `spellcheck`, `deadcode`, `lint-actions`,
`lock-check`, `format-check` and `lint`, each also confirmed individually.

Branch is rebased on `forks/amsterdam` at `f1aa97022`; all of the above was
re-run against that tip.

### Related Issues or PRs

N/A — no existing issue or PR reports this. The nearest neighbours, checked and
distinct:

- **#743** (*Refactor base fee calculation*, merged) corrects the fork-block
  **base-fee value** for Arrow Glacier and Gray Glacier — adjacent, but a
  different quantity, and it does not restore the gas-limit check.
- **#1411** (EIP-7782, closed) contains a contributor remark that
  `check_gas_limit` "needs custom logic for the fork block" — but that is about
  implementing a *new* fork's activation logic on their own branch, not a report
  that London's `validate_header` skips the check.
- **#2230** (*historically-accurate per-fork block gas limits*, open) concerns
  the gas limit used when **filling test fixtures** (`A-test-fill`), not header
  validation.
- **#3056** (*move priority-fee check to `validate_transaction`*, merged) is the
  most recent change to this file and touches an unrelated validation path.

### Checklist

Suggested labels: `C-bug`, `A-spec-specs` — matching the title's `fix` type and
`spec-specs` area.

### Cute Animal Picture

Screenshot_20260805_143721_Gallery

### DIFF
--- a/src/ethereum/forks/london/fork.py
+++ b/src/ethereum/forks/london/fork.py
@@ -331,7 +331,17 @@ def validate_header(chain: BlockChain, header: Header) -> None:
     assert isinstance(FORK_CRITERIA, ByBlockNumber)
 
     expected_base_fee_per_gas = INITIAL_BASE_FEE
-    if header.number != FORK_CRITERIA.block_number:
+    if header.number == FORK_CRITERIA.block_number:
+        # The fork block has no parent base fee to extrapolate from, so the
+        # base fee is fixed. The gas limit is still bounded, but relative to
+        # the parent's gas limit scaled by the elasticity multiplier, because
+        # the fork doubles the gas limit while holding the gas target
+        # constant.
+        if not check_gas_limit(
+            header.gas_limit, parent_header.gas_limit * ELASTICITY_MULTIPLIER
+        ):
+            raise InvalidBlock
+    else:
         # For every block except the first, calculate the base fee per gas
         # based on the parent block.
         expected_base_fee_per_gas = calculate_base_fee_per_gas(


spencer-tb:
Too AI auto gen for me, sry

PR #3310, opened 2026-08-05
feat(binary-trie): move all code chunks into the content-addressed code zone

CPerezz:
### Description

Implements the EIP-8297 change merged in [ethereum/EIPs#12082](https://github.com/ethereum/EIPs/pull/12082) and its follow-up [ethereum/EIPs#12084](https://github.com/ethereum/EIPs/pull/12084): every code chunk is content-addressed in `CODE_ZONE`, keyed by code hash and grouped 256 chunks to a stem, and no code is keyed by address. The account header holds only basic data, the code hash, and the first 64 storage slots; header sub-indices 128–255 are unallocated.

### Related Issues or PRs

Tracks [ethereum/EIPs#12082](https://github.com/ethereum/EIPs/pull/12082) and [ethereum/EIPs#12084](https://github.com/ethereum/EIPs/pull/12084).

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

squirrel filing its content-addressed acorns

PR #3315, opened 2026-08-05
binary(tests): removal must keep overflow code a survivor shares

awskii:
`test_selfdestruct_same_transaction_leaves_no_account` destroys an account whose
initcode self-destructs, so it deploys no code, and it writes slot 0 only. Nothing
in the fixture suite pins the part of removal that is specific to this tree: an
account owns its header stem and its storage prefix and removing it removes those
two subtrees, but its overflow code chunks are in neither. Those are
content-addressed, so byte-identical contracts share one set of leaves, and the
leaves outlive any single account referencing them.

A client that drops them along with the account leaves a surviving contract's code
half-committed, and the state root moves.

## Changes

`test_removal_keeps_overflow_code_a_survivor_shares`: a contract deployed before the
transaction holds bytecode of 3,969 bytes, one past the header limit, so it reaches
the code zone. The transaction creates a second contract with the same bytecode,
writes storage on both sides of the header boundary, and calls it so it
self-destructs — all in one transaction, which post-EIP-6780 is the only way an
account holding code and storage is removed at all. The survivor's code has to be
intact afterwards.

Fills against the reference implementation.

### DIFF
--- a/tests/binary_tree/eip8297_partitioned_binary_tree/test_account_lifecycle.py
+++ b/tests/binary_tree/eip8297_partitioned_binary_tree/test_account_lifecycle.py
@@ -20,12 +20,17 @@ from execution_testing import (
 )
 
 from .helpers import FACTORY_CANARY_SLOT, create_contract_via_factory
-from .spec import ref_spec_8297
+from .spec import Spec, ref_spec_8297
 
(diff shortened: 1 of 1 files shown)


CPerezz:
Collides with #3310 thus removed as discussed with @awskii

PR #3316, opened 2026-08-05
binary(tests): consecutive deploys into a shared code zone

awskii:
Every existing case that reaches the code zone puts one contract there. Code chunks are
content-addressed, so the whole chain shares one zone: the first deploy is the only one
that inserts into an empty one, and nothing pins that a later deploy leaves the stems
already present alone.

## Changes

`test_consecutive_deploys_share_the_code_zone`: three contracts of 129, 137 and 256
chunks, deployed one per block, then all three called in a final block. The lengths
differ so their stems cannot coincide by accident, and each writes its own marker on
the last block, which fails if a later deploy disturbed the code of an earlier one.

Sized by chunk count rather than by `CODE_OFFSET`, so it fills unchanged before and
after #3310. Verified against both.

### DIFF
--- a/tests/binary_tree/eip8297_partitioned_binary_tree/test_multi_block.py
+++ b/tests/binary_tree/eip8297_partitioned_binary_tree/test_multi_block.py
@@ -27,7 +27,7 @@ from execution_testing import (
 )
 
 from .helpers import FACTORY_CANARY_SLOT, create_contract_via_factory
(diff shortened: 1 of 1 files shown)


kevaundray:
Do you mind also rebasing this ser? (merged in a PR that removes code from the account header)

awskii:
sorry, missed the fact that push didn't pass last time

PR #3264, opened 2026-07-30
fix(tests): enhance & un-skip Amsterdam ported static depth-recursion tests (Pt. 2a)

spencer-tb:
### Description

First slice of the Part 2 ported static enhancement (#3215 follow-up), split into four PRs for reviewability. This one carries the depth-recursion families: the `*1024_oog` / `ab_acalls` / `call_recursive_bomb3` recursion pins, the 63/64 ask-family, and the `transaction64_rule` consolidation. 18 files, 23 skip-list entries un-skipped (153 → 130).

Each recursion test derives its expected OOG-terminated depth with a fork-derived recurrence (per-level costs from opcode composites, EIP-150 clamping, failed-frame grant forfeiture, journal rollback), validated against every ported Cancun pin exactly before predicting Amsterdam — where depths collapse (e.g. 134 → 37) under EIP-8037's state costs. Depth-1024 is unreachable under 63/64 attenuation, so call-loop tests keep a named budget with per-era pinned counts. Under the revised EIP-8038 storage pricing (#3293) `call_recursive_bomb3`'s top level can no longer afford its flag store at the ported budget, so its model now pins a wholesale revert.

Every touched test fills green across its full `valid_from` range, and the whole `tests/ported_static/` tree fills green on Amsterdam at this slice. Chain: this PR → Pt. 2b (#3320) → Pt. 2c (#3321) → Pt. 2d (#3322, drops the emptied skip list + conftest).

### Related Issues or PRs

Follow-up to #3215. Part of the skipped-ported-static burn-down in https://github.com/ethereum/execution-specs/issues/3217; the EIP-8037 fail tracker https://github.com/ethereum/execution-specs/issues/2601 closes with Pt. 2d.

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3322, opened 2026-08-05
fix(tests): un-skip remaining Amsterdam ported static tests and drop the skip list (Pt. 2d)

spencer-tb:
### Description

Final slice of the Part 2 ported static split (chain in #3264): the remaining un-skips, including the stCreate2 Berlin-twin consolidation into single parametrized tests and the `revert_depth_create_address_collision` restoration (the port never actually collided: the collider sat on the CREATE address while the code runs CREATE2, verified against the legacy filler). Removes the last 49 skip-list entries (49 -> 0).

With the list empty, this deletes `amsterdam_skip_list.txt` and the `conftest.py` machinery that applied it.

Review follow-ups folded in:

- `test_create2_oog_from_call_refunds`: the SSTORE pairs solc had folded out of five init codes are restored (the created addresses now derive from the deployed code), and the obsolete EIP-8037 post override is gone, so the OoG arms assert the burned balance on Amsterdam too.
- `test_create2collision_selfdestructed_oog`: the slack funds the counterfactual, so the rollback is attributable to the collision's burned grant (a spec with the collision check disabled now fails it).
- Both `revert_depth_create_address_collision` tests and both `revert_depth_create*_oog` tests: every starved arm reaches the CREATE/CREATE2 before dying, and the burned grant is observable (returning it on collision, or charging new-account state gas on a live target, flips arms).
- `test_revert_opcode_in_calls_on_non_empty_return_data`: the buffer is really non-empty before the probed revert, as the legacy filler intended.
- stRefundTest: the refund quotient comes from the fork and the five tests floor at Berlin.
- The `enhance-ported-test` notes are corrected (the SSTORE dirty re-store composite already tracks pre-Berlin pricing) and extended with the starved-arm and counterfactual-budget lessons.

Incorporates Mario's follow-up from https://github.com/spencer-tb/execution-specs/pull/63:

- Consolidate the two CREATE2 revert/deposit tests into success and one-gas-short cases, with the CREATE2 result carried in the revert payload.
- Replace the refund test's repeated bytecode and expectation tables with eight scenarios times three outcomes.
- Simplify the nested creation test and align its intrinsic calculation with the removal of the transaction value transfer.
- Derive the extra calldata allowance, assert helper storage across success and rollback, and remove the obsolete `pre_alloc_mutable` marker.
- Isolate the intentional CREATE-collision pre-allocation so Engine X packing cannot inject that collider into the CREATE-OOG test with identical creator bytecode.

Validation:

- Full Amsterdam ported-static suite: **19,158 fixture variants passed, zero skips**.
- All touched tests across their supported forks through Amsterdam, with `--generate-all-formats`: **1,800 passed**, including **392 Engine X fixtures** verified against their ordinary engine siblings.
- `just static`: passed.
- Temporary test probes confirmed that one extra gas changes the failing deposit's observable result, and bypassing CALL/CALLCODE/DELEGATECALL refund helpers fails storage assertions. The probes are not included in the PR.
- Independent-client fixture consumption has not been rerun for this update.

### Related Issues or PRs

Follow-up to #3215.

### Checklist

### Cute Animal Picture


fselmo (approved):
I did a thorough review here a few times over and it looks good to me. The only things I found were places where we could tighten pins for future-proofing / gotchas and I was able to expand the fork range for a few of the tests (added as a PR to this branch, take or leave though, to your discretion).

I have not yet run these against clients but the pins make me feel better about the structure here. I will try to run these and report back if this is still open but nothing is blocking from my end.

spencer-tb:
Thanks @fselmo!! :D

Feel free to make some follow up PRs, also you are welcome to do the same for the 2a, 2b, 2c PRs!!

fselmo:
> Feel free to make some follow up PRs, also you are welcome to do the same for the 2a, 2b, 2c PRs!!

Did you see [this one](https://github.com/spencer-tb/execution-specs/pull/67)? Can't tell from this comment 😅

PR #3324, opened 2026-08-05
feat(binary-trie): store delegation indicators in the account header

CPerezz:
### Description

Implements EIP-8297's proposed move of EIP-7702 delegation indicators out of content-addressed `CODE_ZONE` code and into the account's own header stem, at a new `DELEGATION_LEAF_KEY = 2`.

**Implementation** (`src/ethereum/binary_trie/embedding.py`, `src/ethereum/state_pbt.py`):
- A delegated account holds one header leaf carrying the 23-byte indicator right-padded with **nine** zero bytes — deliberately *not* the chunk encoding, which spends its first byte on a push-data count and pads with eight. The two encodings are the same length and share no byte position, so reusing `chunkify_code` here writes a plausible wrong value; a test pins the difference.
- Being delegated and holding contract code are exclusive, so every account holds **exactly one** of the code-hash and delegation leaves. `embed_account` enforces it by writing one and **unconditionally removing** the other: it is told nothing about what the account held before, and the incremental path re-embeds over a trie that still carries the pre-state leaf. Both stale-leaf directions are the most likely silent-wrong-root bug, and both are mutation-tested.
- Classification is a function of the **code**, never of its hash — the marker plus the exact length. This is what makes the separate sub-index safe: a discriminator reading a value's leading bytes is grindable.
- With delegations out of the code zone, no live account can replace non-empty code (deployment requires an empty code hash; EIP-3541 forbids deploying an indicator; EIP-161 clearing reaches only codeless accounts; EIP-7702 requires empty-or-delegated). The EIP drops the rule that a `code_hash` change reclaims shared chunks, and `apply_diff_to_trie` drops the call that implemented it. Only deletion reclaims — and a delegation is exempt there too, since it lives in the header its deletion removes.
- `ethereum.binary_trie` may not import from a fork, so it carries its own copy of the 7702 marker and length, as it already does for `EMPTY_CODE_HASH`. A new test pins both against `binary_tree`'s constants **and** checks the two classifiers agree across seven edge cases (length ±1, near-miss markers, empty code).

**Tests** — the previous PR (#3310) built a delegation-*sharing* group that this inverts:
- The shared-designator staircase becomes its inverse: two authorities delegating to the **same** target hold separate leaves with byte-identical values, one re-delegating leaves the other's leaf untouched, and deleting one leaves the other standing. Same-target is load-bearing — with different targets even a content-addressed implementation would produce two leaves and pass.
- The code-change reclamation tests move to the **deletion** arm they still describe, preserving the survivor-ordering and multi-group sentinels rather than losing them.
- The differential fuzzer's ground truth becomes the normative form (exactly one of the two leaves, by the same predicate the EVM uses). It could not fail today — the fuzzer can't generate an indicator — which is exactly why it was worth fixing now.
- New coverage: the EIP's key and value vectors, exactly-one across four account shapes, both transitions through the incremental path, deletion taking the leaf, EIP-7610 unaffected (sub-index 2 sits below the storage range), and grinding resistance.

**Mutation gauntlet** — six targeted breaks, each caught by its named test: stale code-hash leaf after delegating · stale delegation leaf after undelegating · chunk encoding used for the leaf value · delegation content-addressed instead of keyed by account · chunks still written for an indicator · classification read from the code hash.

**Grinding constants.** The EIP's rationale for a separate sub-index is that a value-prefix discriminator is grindable for ~2²⁴ offline Keccaks. That is cheap enough to make testable, so this PR ships two ground constants: a 23-byte contract whose code hash begins `0xef0100` (unit level), and a 23-byte **storage-writing** contract whose hash does too (fixture level). The latter is the only fixture here that fails a broken client for the right reason rather than merely producing a different root: a client discriminating on leading bytes dispatches to a codeless address and writes nothing, so the difference shows in the **alloc**. Being exactly 23 bytes, it also defeats the `code_size == 23` guarded variant the EIP mentions.

**Vectors.** `embedding` gains `delegation_key`; three cases join the whole-state set (two authorities on one target — the discriminating case, without which no vendored-vector consumer can tell the new layout from the old — a delegated account with storage, and the grinding contract). The README gains a third construction rule beside "Zero is absent" and "Code is sized, not delimited". Verified blast radius: `trie_roots`, `chunkify_code` and `encode_basic_data` byte-identical, `delegation_designator` the **only** existing root that moved, no existing embedding key changed, regeneration deterministic.

Verification: `just binary-trie-unit-test` → 204 passed; `just binary-trie-fork` → 205 fixtures filled green; `ruff`/`mypy`/`codespell`/`ethereum-spec-lint` clean; docs build clean.

### Related Issues or PRs

Implements [ethereum/EIPs#12114](https://github.com/ethereum/EIPs/pull/12114), **merged** as `2f0a8bed`; `ref_spec_8297.version` is pinned to that commit. The merged text differs from the draft this branch was written against only in wording — every normative item is unchanged, and one added clarification confirms the design: an account's code hash is "held in the `code_hash` leaf or, for a delegated account, computed from the delegation leaf". Its still-open companion [ethereum/EIPs#12115](https://github.com/ethereum/EIPs/pull/12115) amends EIP-8347's converter and replay rules; EIP-8347 is not implemented in this repo, so there is no code impact here. Note that [ethereum/EIPs#12095](https://github.com/ethereum/EIPs/pull/12095) edits the same lines of the sub-indices-in-use sentence and reserves 2–63 for future header fields — complementary in substance, but the two will need sequencing.

Follows #3310.

### Checklist

### Cute Animal Picture

a pointer, which is what a delegation indicator is


tests/binary_tree/eip8297_partitioned_binary_tree/spec.py
```diff
@@ -11,9 +11,12 @@ class ReferenceSpec:
     version: str
 
 
+# Pinned to the head of ethereum/EIPs#12114, which moves delegation
+# indicators into the account header and is not yet merged. Bump to
+# the master commit once it lands.
```
kevaundray: we can remove this

kevaundray (approved):
Looks fine to merge

tests/binary_tree/eip8297_partitioned_binary_tree/spec.py
CPerezz: As soon as is merged, I'll do so and merge!

PR #3298, opened 2026-08-04
feat(test-rpc): add eth_getTransactionReceipt batching

jochem-brouwer:
### Description
This PR adds using batch-JSON-RPCs on eth_getTransactionReceipt, this is especially for benchmarks/tests which send a lot of transactions which we all verify if those were included/mined.

### Related Issues or PRs

N/A.

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3327, opened 2026-08-06
refactor(test-rpc): unify batch request chunking

LouisTsai-Csie:
### Description

There are two unrelated batch-size constants:

- In `send_wait_transactions`, `EthRPC.DEFAULT_MAX_TRANSACTIONS_PER_BATCH` is configured to 750. This restricts the number of transactions sent per iteration: send them, wait for them, and then do the next iteration.
- In `get_transaction_receipts`, the batch size is configured to 500 (chunk size). The same restriction could be applied to `get_balances`, `get_codes`, `get_transactions_by_hash`, `send_transactions`,`get_account` and `get_alloc`, none of which chunk today.

In this PR, I lift the transport chunking into `post_batch_request` and collapse both constants into a single `BaseRPC.max_batch_size` (default 750), so every batched method is chunk-safe. `get_transaction_receipts` drops its
own chunking loop, and `send_wait_transactions` keeps its loop but reads the shared limit. 

The constructor argument and pytest fixture are renamed from `max_transactions_per_batch` to `max_batch_size`; the `--max-tx-per-batch` lag keeps its name.

Notable behavior change: `--max-tx-per-batch` now also caps receipts, balances, code and account state batches, and receipt batches grow from 500 to 750.

### Related Issues or PRs

Follow up on PR #3298 , comment https://github.com/ethereum/execution-specs/pull/3298#pullrequestreview-4863151016

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/rpc/rpc.py
+++ b/packages/testing/src/execution_testing/rpc/rpc.py
@@ -213,9 +213,13 @@ class BaseRPC:
     simulators.
     """
 
+    OVERLOAD_THRESHOLD: int = 1000
+    DEFAULT_MAX_BATCH_SIZE: int = 750
+
     namespace: ClassVar[str]
     response_validation_context: Any | None
     request_timeout: TimeoutType
+    max_batch_size: int
 
     def __init__(
         self,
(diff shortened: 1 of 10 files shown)


packages/testing/src/execution_testing/rpc/rpc.py
```diff
@@ -213,27 +213,39 @@ class BaseRPC:
...
-        `request_timeout` bounds every request made through this client;
+        - `request_timeout` bounds every request made through this client;
         `None` disables the bound.
+        - `max_batch_size` caps how many calls in a single batch request.
         """
         self.url = url
         self.request_id_counter = count(1)
         self.response_validation_context = response_validation_context
         self.request_timeout = request_timeout
+        self.max_batch_size = max_batch_size or self.DEFAULT_MAX_BATCH_SIZE
```
spencer-tb, line 247: Guard here? 
```suggestion
        if max_batch_size is not None and max_batch_size < 1:
            raise ValueError(
                f"max_batch_size must be >= 1, got {max_batch_size}"
            )
        self.max_batch_size = max_batch_size or self.DEFAULT_MAX_BATCH_SIZE
```

spencer-tb (approved):
Small nit, otherwise LGTM!

PR #3336, opened 2026-08-07
test(evm_tools): revive the opcode-count test against in-repo inputs

### LINKED ISSUE #3326: chore(testing): delete or revive test_count_opcodes.py (never runnable since introduction)
### Context

`test_count_opcodes.py` (now at
`packages/testing/src/execution_testing/evm_tools/tests/`, previously
`tests/evm_tools/`) was added in `38e45e58d` ("Add flag to count opcodes during
execution", 2025-07-29) alongside the opcode-count tracer. It has never run,
in CI or in any default local invocation, because three independent blockers
have been in place since the day it was added:

1. Never collected. The test is marked `@pytest.mark.evm_tools`, and the tox
   config of the day ran `pytest -m "not slow and not evm_tools"`. No tox
   environment and no CI workflow ever selected the marker. The Justfile era
   replaced marker deselection with explicit `--ignore` flags, which #3307
   carries forward into two targets (`test-tests`, `test-tests-pypy`).
2. Fixture out of scope. The same commit added the `root_relative` fixture to
   `tests/json_infra/conftest.py` (now `tests/json_loader/conftest.py`), but
   the test lived in the sibling `tests/evm_tools/` tree with no conftest.
   Pytest scopes conftest fixtures to their directory tree, so collection
   errors with `fixture 'root_relative' not found`. Reproduced on the #3307
   branch.
3. Data never existed. `root_relative("fixtures/evm_tools_testdata/t8n/fixtures/testdata/2")`
   resolves under the json loader tree, where `fixtures/` is gitignored
   download-time content, and no download step has ever provided
   `evm_tools_testdata`. The path shape matches go-ethereum's
   `cmd/evm/testdata` layout, so the test appears written against locally
   copied geth testdata.

A consequence worth stating: the opcode-count tracer
(`evm_tools/t8n/evm_trace/count.py`) has never had test coverage. This was its
only test.

### Options

1. Delete the test (and the two `--ignore` flags in the Justfile). There is no
   previously green state to restore; deletion loses nothing that ever ran.
2. Rewrite it against in-repo inputs. The `t8n_build/` fixtures added by #3307
   (alloc/env/txs for a real Frontier transition) are almost exactly the input
   this test needs; rewriting against them would give `count.py` its first
   real coverage and remove both `--ignore` flags. If the fixtures move into
   the testing package (suggested in the #3307 review), the test and its data
   would sit side by side.

Option 2 is preferable if anyone uses `--opcode.count`; otherwise option 1.

### Related

- #3307 (moved the test; entrenched the `--ignore` in two targets)
- `38e45e58d` (introduced the test, the marker exclusion, and the misplaced
  fixture in one commit)

mkzung:
Closes #3326, taking the second option there: rewrite rather than delete, so the
opcode-count tracer gets its first coverage.

The test asked for `root_relative`, defined in a sibling tree, and pointed at
`fixtures/evm_tools_testdata`, which no download step provides. It now uses the
client_clis transition inputs. Fixture 3 is the only alloc there with code, `0x600140`,
so the call runs PUSH1 then BLOCKHASH. Those inputs are a plain transition rather than a
state test, so the run names the fork instead of passing `--state-test`.

The `--ignore` in the `spec-tools` target goes with it. Only one target carries it on
`forks/amsterdam`; the issue mentions two, which holds on another branch.

One thing I did not decide: `test_execution_specs.py` imports `Berlin` while this names
Frontier. The count is the same either way, but say if Berlin is the intended fork.

## Testing

`uv run pytest -n 4 tests/evm_tools`: 42 passed, with the file collected for the first
time. I restored the pre-change file from git and confirmed it still errors with
`fixture 'root_relative' not found`, and flipping the expected count to `PUSH1: 2` fails,
so the assertion really reads the tracer.

Static analysis: `ruff check`, `ruff format --check`, `codespell`, `vulture`,
`ethereum-spec-lint` and `mypy` all pass.

Claude Code was used for implementation assistance, covering the test rewrite and this
description. All changes have been reviewed, understood and manually tested by me.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -215,7 +215,6 @@ spec-tools *args: (_tmp "spec-tools")
     uv run pytest \
         -n {{ xdist_workers }} \
         --basetemp="{{ output_dir }}/spec-tools/tmp" \
-        --ignore=tests/evm_tools/test_count_opcodes.py \
         "$@" \
         tests/evm_tools
 
--- a/tests/evm_tools/test_count_opcodes.py
+++ b/tests/evm_tools/test_count_opcodes.py
@@ -6,7 +6,6 @@ using the T8N tool.
 import json
 from io import StringIO
 from pathlib import Path
-from typing import Callable
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


marioevz:
Hi @mkzung, thank you for putting the effort of addressing the issue with your PR. We currently are undergoing a big refactor of this file and many others in this same folder in #3307, and unfortunately I've rebased your changes and the fix is not compatible with the it, and is going to cause conflicts during the rebase. I think the best course of action for now is to close this PR, and try to address this issue directly in #3307. Again, thanks for this effort! Closing for now.

PR #3337, opened 2026-08-07
feat(test-specs): Allow specs to fill same fixture format more than once, add `inclusion_test` marker

marioevz:
### Description

Lets a spec type fill one fixture format more than once, under different labels, and adds an `inclusion_test` marker that identifies the tests where the question under test is whether a transaction can be included in a block.

Nothing generates a second label yet, so this is a **no-op on fixture output**, and the generated fixtures before and after are identical. What it removes is the set of assumptions that made a second label impossible: that a label is only a test id, that every label of a format shares its transition tool cache, and that a format cannot exclude itself for a fork.

#### `primary_format` replaces `derived_test`

`derived_test` was doing two jobs. Its name says "this fixture is a translation of the spec into another spec's format", but every consumer used it for something else: "skip the redundant formats so each test fills once". Those coincide for `StateTest` and diverge for `BlockchainTest`, where none of the five formats is a translation.

The marker is now `primary_format`, applied to the first format a test actually generates, which is what the consumers wanted:

- `.github/workflows/test.yaml`: `-m "not slow and not derived_test"` becomes `-m "not slow and primary_format"`
- `eip_checklist.py` skips items *without* `primary_format`

**Anyone using `-m "not derived_test"` locally needs to switch to `-m primary_format`.

#### Parametrization moved into `BaseTest`

`pytest_generate_tests` in the filler plugin built the parameter sets inline. That logic is now `BaseTest.fixture_format_parameters` and `BaseTest.execute_format_parameters`, each returning the format paired with its parameter, so a spec type can override them to parametrize further. The filler hook is 13 lines and still owns the session-level format selection.

`labeled_format_parameter_set` moved from `plugins/shared/helpers.py` to `specs/base.py` too.

#### Labels are retained through parametrization

`format_class()`, `format_id()` and `marks()` are now polymorphic across plain and labeled formats, replacing the `isinstance` branching that unwrapped them. The parameter's value is the labeled format rather than the stripped class, so `generate` receives the label and can tell two labels of one format apart.

`__eq__` on both labeled types now compares label as well as format, so two labels of one format are no longer equal to each other; comparison against a plain format is unchanged, which is what keeps the existing `generate` dispatch working. `__hash__` was added to match — these types were unhashable before.

#### A label can differ from its siblings

Three things were properties of the format class that a label now overrides:

- **The transition tool cache key.** The cache is keyed by the test's format-stripped node id plus the format's key, and its entries are looked up by call order, so two labels sharing a key would feed each other's transition tool output. A label that asks the transition tool for something different sets its own key, or an empty string to opt out.
- **`supports_fork`**, so a label whose fixture only makes sense for some forks can restrict itself.
- **`discard_fixture_format_by_marks`**, on both the fixture format and the spec type, so a label can veto itself by marker without affecting its siblings.

#### The `inclusion_test` marker

Marks a test that verifies whether a transaction can be included in a block, where the transaction under test is the last one of the last block. It is a no-op today.

Applied to **117 tests across 43 files**, covering 31 of the 33 distinct `TransactionException` values in the suite. Selection was primarily made from `exception_test` pytest marker.

53 of the marked tests already assert in both directions; the other 64 only expect a rejection. Adding the positive case to those is a separate change, potentially for a follow-up PR.

### Related Issues or PRs
N/A.

### Checklist

### Cute Animal Picture


LouisTsai-Csie (changes requested):
The `Justfile` still uses `-m not derived_test` instead of `-m primary_test`, causing slowdown: `fill-pypy` (4.0m → 6.1m, +53%) and `json-loader` (4.2m → 5.2m, +24%)

packages/testing/src/execution_testing/specs/base.py
```diff
@@ -142,16 +173,81 @@ def model_post_init(self, __context: Any, /) -> None:
...
+            if cls.discard_fixture_format_by_marks(
+                format_with_or_without_label,
+                markers,
+            ):
+                continue
+            parameter = labeled_format_parameter_set(
+                format_with_or_without_label
+            )
+            if not parameters:
+                parameter.marks.append(pytest.mark.primary_format)  # type: ignore
```
danceratopz: The `primary_format` mark is decided too early: it goes to the first format in this list, but later steps can still remove that format from the run. When the removed one is the marked one, the test is left with no parameter marked `primary_format`, and `-m primary_format` then skips the test entirely instead of filling it once.

Two steps run after this loop and can remove the marked parameter:

1. The session phase filter (`should_generate_format` in `filler.pytest_generate_tests`). This is observable today: `fill --generate-pre-alloc-groups -m primary_format` on `tests/frontier/validation/` collects 0 tests, where the base branch's `-m "not derived_test"` collected 32. `fill-stateful -m primary_format` selects 0. `fill --generate-all-formats -m primary_format` runs its pre-alloc phase with nothing selected and reports no error. The docstring argues that sessions which can filter the primary out only generate one format per test, so the marker is moot there, but that stops being true the moment something generates a second label, which is the point of this PR.

2. The label-level vetoes, `LabeledFixtureFormat.supports_fork` and `discard_fixture_format_by_marks`, run during collection. A first label that vetoes itself takes the mark with it, and the surviving labels stay unmarked.

I would decide the primary after all filtering, for example by threading the session predicate into `fixture_format_parameters`, or by re-assigning the mark in the hook once filtering is done, so it always lands on a format that survives.

Related: `execute_format_parameters` never adds the mark at all, so `execute remote -m primary_format` deselects everything (34 of 34), even though the marker is registered for both commands.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py
```diff
@@ -189,8 +189,14 @@ def pytest_configure(config: pytest.Config) -> None:
...
-        "derived_test: Mark a test as a derived test (E.g. a BlockchainTest "
-        "that is derived from a StateTest).",
+        "primary_format: Mark the first fixture format generated for a test. "
+        "Select with `-m primary_format` to fill every test exactly once.",
+    )
+    config.addinivalue_line(
+        "markers",
+        "inclusion_test: Mark a test that verifies whether a transaction can "
+        "be included in a block. The transaction under test is the last one "
+        "of the last block.",
```
danceratopz: The second sentence here is a promise: in a marked test, a consumer can find the transaction under test by looking at the last transaction of the last block. A future tool will rely on that to locate and transform the transaction. A fair share of the 117 marked tests do not keep the promise.

The marked fork-transition tests cannot keep it by construction: they need blocks before and after the fork, so their tested transactions sit in earlier blocks. For example, `eip7981_increase_access_list_cost/test_fork_transition.py` builds 4 blocks with tested txs in blocks 1 to 3 (block 2 carries two), and `eip7594_peerdas/test_max_blob_per_tx.py::test_invalid_blob_max_txs_per_block_transition` has a rejected tx in each of its 3 blocks. The eip7954, eip7976, eip8037, eip8038 and eip7825 transition tests follow the same pattern.

The `TransactionTestFiller` tests (`eip7702_set_code_tx/test_invalid_tx.py`, `frontier/validation/test_transaction.py`) build no block at all, so the sentence does not apply to them. The marking commits disagree on this case: e80207d112 explicitly excluded `TransactionTest` based tests, then 46c37bacec marked exactly those.

Nothing consumes the marker yet, so no check can catch the drift; this description is the only contract a future consumer gets. I would either relax the wording (for example "at least one transaction under test is the last transaction of its block", plus a note on what the marker means for block-less fixtures) or unmark the tests that do not fit. Either way, an entry in `docs/writing_tests/test_markers.md` would help future authors apply it correctly.

packages/testing/src/execution_testing/fixtures/base.py
```diff
@@ -238,15 +273,20 @@ def __init__(
         fixture_format: "Type[BaseFixture] | LabeledFixtureFormat",
         label: str,
         description: str,
+        transition_tool_cache_key: str | None = None,
```
danceratopz, line 327: The problem this parameter exists to prevent is silent: if two labels of one format share a cache key, the second label reads the first label's cached transition tool output, and wrong data flows into its fixtures without anything failing. But nothing in the repo passes `transition_tool_cache_key=` yet, and `test_t8n_cache.py` does not cover it, so the whole non-`None` path (including the empty-string opt-out) ships unexercised. The safety of the feature then rests on every future label author remembering an argument they have never seen used.

A unit test would be worth adding now, plus possibly a registration-time guard that rejects two labels of a spec type sharing `(format_class(), cache_key)`. That would turn the silent trap into an immediate error.

packages/testing/src/execution_testing/fixtures/base.py
```diff
@@ -238,15 +273,20 @@ def __init__(
...
+        label that asks the transition tool for something different must set
+        its own key, or an empty string to opt out of caching, since labels
+        sharing a key also share cached output.
+        """
+        self.format = fixture_format.format_class()
         self.label = label
         self.description = description
+        self._transition_tool_cache_key = transition_tool_cache_key
         if label not in LabeledFixtureFormat.registered_labels:
             LabeledFixtureFormat.registered_labels[label] = self
```
danceratopz, line 357: Wrapping an already-labeled format in a new label loses information from the inner label in two ways:

1. `fixture_format.format_class()` unwraps to the plain class, and `_transition_tool_cache_key` comes only from the new argument, so the inner label's cache-key override is silently dropped.

2. `StateTest` derives the outer label from `fixture_format.format_name` rather than `format_id()`. If a format ever has two labels, both wrap to the same outer name; under the new label-aware `__eq__` they compare equal, and `registered_labels` keeps only the first one (the `if label not in` guard below), silently.

Neither can trigger today, because `supported_fixture_formats` holds only plain classes for `BlockchainTest` derived specs, but the first real second label goes straight through this path.

marioevz:
> The Justfile still uses -m not derived_test instead of -m primary_test, causing slowdown: fill-pypy (4.0m → 6.1m, +53%) and json-loader (4.2m → 5.2m, +24%)

Addressed by @danceratopz in the last commit 👍

LouisTsai-Csie (approved):
Good from my side. cc @danceratopz @spencer-tb if you want to do a final review

tests/amsterdam/eip7981_increase_access_list_cost/test_transaction_validity.py
```diff
@@ -289,6 +292,7 @@ def test_transactions_without_access_list(
     )
 
 
+@pytest.mark.inclusion_test
```
danceratopz: This test is already marked via the module level marker at L28.

```suggestion
```

danceratopz: Fixed, across all instances.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -1833,38 +1833,15 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     markers = list(metafunc.definition.iter_markers())
     for test_type in BaseTest.spec_types.values():
         if test_type.pytest_parameter_name() in metafunc.fixturenames:
-            parameters: List[ParameterSet] = []
```
danceratopz, line 1836: So nice to simplify the logic here!

danceratopz (commented):
The core labeled-format refactor looks sound; Claude verified that fixture output is byte-identical on a sample fill. Is there somewhere we could document this feature? Or will you add that when it gets concrete with ILs?

Most of the below is automated!

tests/ported_static/stTransactionTest/test_high_gas_price_paris.py
```diff
@@ -21,6 +21,7 @@
 REFERENCE_SPEC_VERSION = "N/A"
 
 
+@pytest.mark.inclusion_test
```
spencer-tb, line 24: Should we add the marker to these tests too? 
- `tests/ported_static/stTransactionTest/test_no_src_account.py`
- `tests/ported_static/stTransactionTest/test_no_src_account1559.py`
- `tests/ported_static/stTransactionTest/test_no_src_account_create.py`
- `tests/ported_static/stTransactionTest/test_no_src_account_create1559.py`

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py
```diff
@@ -189,8 +189,14 @@ def pytest_configure(config: pytest.Config) -> None:
     )
     config.addinivalue_line(
         "markers",
-        "derived_test: Mark a test as a derived test (E.g. a BlockchainTest "
-        "that is derived from a StateTest).",
+        "primary_format: Mark the first fixture format generated for a test. "
```
spencer-tb, line 196: Should we add a `primary_format` entry to `docs/writing_tests/test_markers.md`? With `derived_test` no longer registered?

packages/testing/src/execution_testing/specs/base.py
spencer-tb: Related cleanup, whichever way (https://github.com/ethereum/execution-specs/pull/3337/changes#r3747706678) resolves!
```suggestion
                parameter = labeled_format_parameter_set(
                    format_with_or_without_label,
                    extra_marks=[] if parameters else [pytest.mark.primary_format],
                )
```

spencer-tb (commented):
Small comments. Dan's seem to address most!

Awesome addition :)

packages/testing/src/execution_testing/specs/base.py
marioevz: > The `primary_format` mark is decided too early: it goes to the first format in this list, but later steps can still remove that format from the run. When the removed one is the marked one, the test is left with no parameter marked `primary_format`, and `-m primary_format` then skips the test entirely instead of filling it once.

Yes, this is all intentional.

> 1. base branch's `-m "not derived_test"` collected 32

That was a bug, Engine X tests are never the primary format. Requesting to fill engine x with `-m primary_format` is incorrect usage.

> for example by threading the session predicate into `fixture_format_parameters`

Yes, claude tried to slip this one in very insistently when I was trying to code the change. It makes no sense to me to add extra complexity to what is basically incorrect usage.

> so `execute remote -m primary_format` deselects everything

This is fixable yes, but I feel this is over-indexing on incorrect usages. `execute` + `-m primary_format` serves no purpose in my opinion.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py
marioevz: > `eip7594_peerdas/test_max_blob_per_tx.py::test_invalid_blob_max_txs_per_block_transition`

Yes, good catch, I think the marker should be removed.

> The eip7954, eip7976, eip8037, eip8038 and eip7825 transition tests follow the same pattern.

I think what we can do is make this enforceable: basically verify that only the last block's last transaction is the only invalid transaction in the entire list of blocks, or that none of the blocks contain any invalid transactions.

> The `TransactionTestFiller` tests (`eip7702_set_code_tx/test_invalid_tx.py`, `frontier/validation/test_transaction.py`) build no block at all

They contain a `pre`, so you can build a blockchain test from them where there's only a single transaction. Perhaps not as important today, but for FOCIL any transaction in a `TransactionTest` can end up in the inclusion list, and we need to test them.

> [e80207d](https://github.com/ethereum/execution-specs/commit/e80207d112aac2d2098f1f5af08bc65d42c448e0) explicitly excluded `TransactionTest` based tests

This comment is a mistake.

packages/testing/src/execution_testing/fixtures/base.py
marioevz: > A unit test would be worth adding now, plus possibly a registration-time guard that rejects two labels of a spec type sharing `(format_class(), cache_key)`. That would turn the silent trap into an immediate error.

This is a nice test, definitely worth adding.

packages/testing/src/execution_testing/fixtures/base.py
marioevz: > 2\. `StateTest` derives the outer label from `fixture_format.format_name`

Yes, this looks like an anti-pattern, nice catch. I think we need to standardize fixture format re-labeling for it to always be safe and preserve the properties it needs.

packages/testing/src/execution_testing/cli/pytest_commands/plugins/shared/execute_fill.py
marioevz: The `primary_format` marker is not meant to be applied to tests directly like the rest of the markers in that file, but we could definitely add `inclusion_test` to this doc.

tests/ported_static/stTransactionTest/test_high_gas_price_paris.py
marioevz: Added!

packages/testing/src/execution_testing/specs/base.py
marioevz: Added, thanks!

marioevz:
@danceratopz @spencer-tb I've responded/addressed all comments PTAL and let me know if this is ok to continue.

spencer-tb (approved):
LGTM!

packages/testing/src/execution_testing/specs/base.py
danceratopz: Thanks, makes sense!

danceratopz (approved):
Thanks for this, excited to get this in!

PR #3338, opened 2026-08-08
binary(tests): delegation lifecycle, 2935 ring buffer, and chunking edges

CPerezz:
Ports the protocol pins that so far existed only as go-ethereum unit tests — the gaps the suite's own docstrings declare open:

- EIP-7702 delegation cleared, and re-authorized to the same target (new `test_delegation_lifecycle.py`)
- the EIP-2935 get path agreeing with BLOCKHASH, with ring slots in both storage homes (overflow reached by chaining past block 64)
- CREATE2 re-occupying a same-tx-destroyed address with different code, through NUMBER-branched initcode
- code ending in a truncated PUSH; a whole aligned all-zero chunk (the fixture-level face of #3305's vectors); initcode cloning a chunked contract via EXTCODECOPY
- the committed root after a contract with unique deployed code is created, called and self-destructed in one transaction

No overlap with #3316. Filled 230/230 against the reference; go-ethereum's binary-tree branch consumes 147/147 direct and refills the suite byte-identically (151/151 files). geth drops its native twins in exchange: CPerezz/go-ethereum#15.

### DIFF
--- a/tests/binary_tree/eip8297_partitioned_binary_tree/test_delegation_lifecycle.py
+++ b/tests/binary_tree/eip8297_partitioned_binary_tree/test_delegation_lifecycle.py
@@ -1,0 +1,67 @@
+"""EIP-7702 delegation lifecycle under the EIP-8297 binary tree."""
+
+import pytest
+from execution_testing import (
+    Account,
+    Alloc,
+    AuthorizationTuple,
+    Block,
+    BlockchainTestFiller,
+    Op,
+    Transaction,
+)
+
+from ...prague.eip7702_set_code_tx.spec import Spec as Spec7702
+from .spec import ref_spec_8297
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_8297.git_path
+REFERENCE_SPEC_VERSION = ref_spec_8297.version
+
+pytestmark = pytest.mark.valid_from("BinaryTree")
+
+
+def test_same_target_reauthorization_keeps_designator(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+) -> None:
+    """
+    Verify re-authorizing the same delegate twice bumps the nonce
+    while the designator stays resident and keeps executing.
+    """
+    counter_slot = 1
+    delegate = pre.deploy_contract(
+        code=Op.SSTORE(counter_slot, Op.ADD(Op.SLOAD(counter_slot), 1))
+        + Op.STOP
+    )
+    authority = pre.fund_eoa(0, delegation=delegate)
+    authority_nonce = authority.nonce
+    sender = pre.fund_eoa()
+
+    def reauthorize(tuple_nonce: int) -> Transaction:
+        return Transaction(
+            sender=sender,
+            to=authority,
+            authorization_list=[
+                AuthorizationTuple(
+                    address=delegate,
+                    nonce=tuple_nonce,
+                    signer=authority,
+                )
+            ],
+        )
+
+    blockchain_test(
+        pre=pre,
+        blocks=[
+            Block(txs=[reauthorize(authority_nonce)]),
+            Block(txs=[reauthorize(authority_nonce + 1)]),
+        ],
+        post={
+            authority: Account(
+                nonce=authority_nonce + 2,
+                code=Spec7702.delegation_designation(delegate),
+                storage={counter_slot: 2},
+            ),
+            delegate: Account(storage={}),
+        },
+    )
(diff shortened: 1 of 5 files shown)


CPerezz:
One data point for the unshared-destruct test: whether the code-zone deletion arm ever fires turned out to be moot at fixture level — an EIP-6780 same-tx create+destroy nets out of the block diff before either the reference or geth writes its chunks, so the fixture pins a root both implementations already agree on. It exists so that the first implementation which does materialize chunks mid-transaction (and forgets to reclaim them) fails loudly instead of silently keeping a divergent tree.

kevaundray:
For delegation, not sure if we have it documented, but if a contract has a `isContract` check, then we might want to test that we don't break them even more. I imagine its something like:

```
isContract check: codehash != 0 && codehash != keccak256("")
```

I saw "break them even more" because 7702 already breaks this check

tests/binary_tree/eip8297_partitioned_binary_tree/test_system_contracts.py
```diff
@@ -173,3 +177,76 @@ def test_system_contract_and_user_contract_writes_coexist(
...
+    [
+        pytest.param(0, id="ring_slots_in_header_home"),
+        pytest.param(64, id="ring_slots_in_overflow_home"),
+    ],
+)
+def test_history_contract_ring_buffer_across_blocks(
+    blockchain_test: BlockchainTestFiller,
+    pre: Alloc,
+    first_query: int,
+) -> None:
```
kevaundray, line 192: Why is this specific to binary tries?

kevaundray: Asking because if we agree that the state trie doesn't change anything to do with system contracts, then we could move these tests to amsterdam and merge them into forks/amsterdam with a `valid_from` marker

tests/binary_tree/eip8297_partitioned_binary_tree/test_delegation_lifecycle.py
```diff
@@ -0,0 +1,136 @@
+"""
```
kevaundray: Maybe same question as system contracts; I think the main difference is checking that CODE_HASH leaf is non-empty when delegation leaf is non-empty, and nothing changes wrt delegation?

kevaundray:
Had a comment re backporting tests, if it makes sense, we can create an issue around moving tests from projects/binary-trie that would make sense to put in forks/amsterdam -- I imagine theres quite a few of these outside of this PR

tests/binary_tree/eip8297_partitioned_binary_tree/test_system_contracts.py
CPerezz: Agreed the semantics are generic. But nothing that fills under BinaryTree builds more than 2 blocks in the 2935 suite, so no reachable test crosses the slot-64 storage-home boundary — that crossing is this test's only point, and it means nothing under MPT. The module's generic 4788 tests are listed in https://github.com/ethereum/execution-specs/issues/3342.

tests/binary_tree/eip8297_partitioned_binary_tree/test_delegation_lifecycle.py
CPerezz: This was more about the invariant that has to bee kept between codehash and delegation. But fair

CPerezz:
The tree doesn't change this: a delegated account's EXTCODEHASH is still keccak(designator) — derived from the delegation leaf — so `isContract` checks break exactly as much as under 7702. `test_ext_code_on_set_code` already pins that `(EXTCODEHASH == keccak(designation), EXTCODESIZE == 23)` and runs under BinaryTree.

CPerezz:
Issue filed: https://github.com/ethereum/execution-specs/issues/3342.

CPerezz:
> Had a comment re backporting tests, if it makes sense, we can create an issue around moving tests from projects/binary-trie that would make sense to put in forks/amsterdam -- I imagine theres quite a few of these outside of this PR

I leave this to the outcome of the discussion pending between @spencer-tb and the rest of the STEEL team! :)

kevaundray:
> > Had a comment re backporting tests, if it makes sense, we can create an issue around moving tests from projects/binary-trie that would make sense to put in forks/amsterdam -- I imagine theres quite a few of these outside of this PR
> 
> I leave this to the outcome of the discussion pending between @spencer-tb and the rest of the STEEL team! :)

I think this is orthogonal, ie even if we did merge this into forks/amsterdam, some of these tests are generic enough to be ran before the binary trie fork/changes.

(This is not blocking the current PR though)

kevaundray (approved):
Left some comments, LGTM

Issue #3339, opened 2026-08-08
Consider extending test converage to execution-api tests


kevaundray:
## Observation

The RPC layer is a part of the exposed API of the execution layer. Currently test vectors are being generated with geth and specified in execution-api.

## Suggestion

Still forming the idea, but it seems like this repo could generate those json outputs automatically for all of the blockchain tests already defined, increasing test coverage significantly.

PR #3340, opened 2026-08-08
chore(do not merge): Experiment extending to rpc tests via hive 

kevaundray:
### Description

(Not reviewed, currently directing claude to prototype)

Related #3339

The large diff is because I vendored in the schema file which is about 21K LOC (see packages/testing/src/execution_testing/rpc/schemas/openrpc.json) -- this file was taken from `execution-apis`

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture


tests/amsterdam/eip7708_eth_transfer_logs/test_rpc_transfer_logs.py
```diff
@@ -0,0 +1,49 @@
...
+    BlockchainTestFiller,
+    Transaction,
+)
+
+from .spec import ref_spec_7708
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_7708.git_path
+REFERENCE_SPEC_VERSION = ref_spec_7708.version
+
+pytestmark = [pytest.mark.valid_from("EIP7708"), pytest.mark.rpc]
```
kevaundray, line 24: pytest.mark.rpc allows us to convert existing tests in to rpc test

.gitignore
```diff
@@ -100,3 +100,6 @@ site/
 
 # Temporary data; used for generated checklists
 tmp/
+
+# Local design docs; not part of the published mkdocs site
+docs/plans/
```
kevaundray, line 107: Planned this initial version out with AI to get a prototype

Justfile
```diff
@@ -397,6 +397,27 @@ crops:
...
+        https://github.com/ethereum/execution-apis.git "$work"
+    cd "$work/tools" && go build -o specgen ./cmd/specgen
+    cd "$work" && ./tools/specgen -o "$dest" -deref \
+        -schemas 'src/schemas' -schemas 'src/engine/openrpc/schemas' \
+        -methods 'src/eth' -methods 'src/debug' -methods 'src/txpool' \
+        -methods 'src/engine/openrpc/methods' -methods 'src/testing' \
+        -error-groups 'src/error-groups'
+    echo "Regenerated from {{ tag }} ($(git -C "$work" rev-parse HEAD))."
+    echo "Update the pin table in the schemas README if the tag changed."
+
```
kevaundray, line 420: This is hacky and just a symptom of this PR vendoring in the json schema file

kevaundray:
Note that this isn't _completely_ correct: apart from the engine API methods, it seems that most methods are not versioned by fork in execution-api.  There are two ways that I think would make sense for versioning:

- By fork: this means that we ship new RPC methods or change existing ones over hardforks
- Custom versioning: This is similar to something like `method_name/v1/`. This means that we can upgrade the rpc methods even after a hardfork is shipped. 

For namespaces like `eth_` it _seems_ like doing it by fork would make sense or adding a `/v2` for example, where `v2` corresponds to some hardfork.

EDIT: The testing machinery does it by marking fields as optional and ensuring that if we are at fork X, then all of those fields are present if they were added at fork X. (Given the discussions around REST API, might be best to leave it as is)

kevaundray:
Tried to add eth_simulateV1 but testing against the clients surfaced #3341 as a discrepancy

PR #3348, opened 2026-08-10
fix(benchmark): split benchmark gas evenly: every tx can pay data floor

jochem-brouwer:
Note: written by LLM

**MERGING THIS WILL CHANGE THE BENCHMARK/TEST OUTPUT SIGNIFICANTLY DUE TO UPDATED TX SPLITTER LOGIC**

`split_transaction` gave each split the gas limit cap and the last one whatever remained. Every split is a `model_copy()` of the same transaction, so each carries identical calldata and owes identical floor data gas -- but a remainder tail is unrelated to that floor and can land below it, producing a transaction no client will accept.

Filling the compute benchmarks against Amsterdam hit exactly this. With the EIP-7825 cap of 2**24, a 220M benchmark splits into 14 transactions and the last receives 220,000,000 - 13 * 16,777,216 = 1,896,192 gas, while the 36,864-byte worst-case BLS12_G2MSM k=128 calldata costs 2,374,296 at the floor. geth rejected the build:

    insufficient gas for floor data gas cost: have 1896192, want 2374296

220M was the only value in 100..300M that failed, which is why this surfaced as a single mystifying case rather than a pattern.

It is also new. The floor is 15,000 + 16 * 4 * 36,864 = 2,374,296 under Amsterdam, where EIP-7976 charges 64 gas per calldata byte (zero and non-zero alike) and EIP-2780 lowers the base to 15,000. Under Osaka the same calldata's floor is 21,000 + 10 * 122,484 = 1,245,840, which fits in 1,896,192 -- so the test was correct until EIP-7976 raised the floor. Both figures match `transaction_data_floor_cost_calculator` exactly.

Spreading the gas evenly across the same number of splits raises the smallest share from "whatever remains" to `gas // num_splits`, at least half the cap -- 15,000,000 at worst across the benchmark gas values, against a 2,374,296 floor.

The total is deliberately preserved. Dropping the short tail instead would trade a rejected transaction for a failed assertion, because `BaseTest.validate_gas_used` requires the block to consume exactly `gas_benchmark_value`.

Also makes the BLS12_G2MSM skip guard floor-aware; it compared only the standard intrinsic cost, which is the smaller of the two on calldata this large.

The existing split tests asserted the cap-cap-remainder shape itself, so they now assert the invariants that matter: the split count, the cap ceiling, exact total, incrementing nonces, and an even spread. The new test fails on the old code at 220M with `assert 1896192 >= 2374296`.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/specs/benchmark.py
+++ b/packages/testing/src/execution_testing/specs/benchmark.py
@@ -481,15 +481,50 @@ class BenchmarkTest(BaseTest):
             return [tx]
 
         num_splits = math.ceil(self.gas_benchmark_value / gas_limit_cap)
-        remaining_gas = self.gas_benchmark_value
+
+        # Spread the gas evenly rather than giving every split the cap and the
+        # last one whatever remains.
+        #
+        # Every split is a copy carrying the *same* calldata, so each one has
+        # an irreducible minimum: it must cover that calldata's floor data
+        # cost. A remainder tail is unrelated to that minimum and can land
+        # below it, producing a transaction no client will accept. With the
+        # EIP-7825 cap of 2**24, a 220M benchmark left its final transaction
+        # 1,896,192 gas while its 36,864-byte calldata cost 2,374,296 at the
+        # floor, and geth rejected the block build outright:
+        #
+        #   insufficient gas for floor data gas cost: have 1896192,
+        #   want 2374296
+        #
+        # That was reachable only after EIP-7976 raised the floor from 40 to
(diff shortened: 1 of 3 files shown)


LouisTsai-Csie:
I'm not sure we should modify the framework just to make a single benchmark test work better. Instead, maybe we should refactor the test itself?

jochem-brouwer:
Yes, lets refactor the test. I'll keep this PR open in draft and will close once we do so.

PR #3349, opened 2026-08-11
feat(test-benchmark): more worst case scenario

LouisTsai-Csie:
### Description

Add more benchmark cases to enhance the worst case coverage:
- `test_ext_account_query_cold`: add `EXTCODESIZE` and `EXTCODEHASH` operations.
- `test_arithmetic`: new parametrization for `DIV` and `SDIV`.
- `test_memory.py`: create worst case benchmark memory expansion scenario.
- `test_nested_frame_state_access`: doing state access from deep nested frames.
- `test_delegatecall_staticcall`: similar to `test_contract_calling_many_addresses` but for `STATICCALL`, `DELEGATECALL` version.
- `test_blake2f`: new input.
- `test_jumpdest_analysis`: Clients could cache the jumpdest analysis result given the current test implementation.
- `test_nested_calls`: stress testing `CALL*` -> `CALL*` -> `CALL*` -> ...

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/tests/benchmark/compute/instruction/test_arithmetic.py
+++ b/tests/benchmark/compute/instruction/test_arithmetic.py
@@ -83,19 +83,6 @@ from tests.benchmark.helper.numeric import (
             ),
         ),
         pytest.param(
-            # This has the cycle of 2, see above.
-            Op.DIV,
-            (
-                0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F,
-                # We want the first divisor to be slightly bigger than
-                # 2**192: a four-word divisor falls past the one- and
-                # two-word fast paths onto the general long-division
-                # algorithm, which the 2**64 and 2**128 cases above do not
-                # reach.
-                0x1000000000000000000000000000000000000000000000033,
-            ),
-        ),
-        pytest.param(
             # Same as DIV-0
             # But the numerator made positive, and the divisor made negative.
             Op.SDIV,
@@ -112,15 +99,6 @@ from tests.benchmark.helper.numeric import (
             (
                 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F,
                 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFCD,
-            ),
-        ),
-        pytest.param(
-            # Same as the 2**192 DIV case above, with the numerator made
-            # positive and the divisor made negative.
-            Op.SDIV,
-            (
-                0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F,
-                0xFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFCD,
             ),
         ),
         pytest.param(
--- a/tests/benchmark/compute/instruction/test_arithmetic.py
+++ b/tests/benchmark/compute/instruction/test_arithmetic.py
@@ -83,6 +83,19 @@ from tests.benchmark.helper.numeric import (
             ),
         ),
         pytest.param(
+            # This has the cycle of 2, see above.
+            Op.DIV,
+            (
+                0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F,
+                # We want the first divisor to be slightly bigger than
+                # 2**192: a four-word divisor falls past the one- and
+                # two-word fast paths onto the general long-division
+                # algorithm, which the 2**64 and 2**128 cases above do not
+                # reach.
+                0x1000000000000000000000000000000000000000000000033,
+            ),
+        ),
+        pytest.param(
             # Same as DIV-0
             # But the numerator made positive, and the divisor made negative.
             Op.SDIV,
@@ -99,6 +112,15 @@ from tests.benchmark.helper.numeric import (
             (
                 0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F,
                 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFCD,
+            ),
+        ),
+        pytest.param(
+            # Same as the 2**192 DIV case above, with the numerator made
+            # positive and the divisor made negative.
+            Op.SDIV,
+            (
+                0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F,
+                0xFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFCD,
             ),
         ),
         pytest.param(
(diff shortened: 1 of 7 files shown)


tests/benchmark/compute/instruction/test_arithmetic.py
```diff
@@ -82,6 +82,19 @@
...
+            # This has the cycle of 2, see above.
+            Op.DIV,
+            (
+                0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F,
+                # We want the first divisor to be slightly bigger than
+                # 2**192: a four-word divisor falls past the one- and
+                # two-word fast paths onto the general long-division
+                # algorithm, which the 2**64 and 2**128 cases above do not
+                # reach.
+                0x1000000000000000000000000000000000000000000000033,
```
marioevz: The `a // (a // b) = b` cycle pattern does not hold for this divisor.
```
>>> a = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F
>>> b = 0x1000000000000000000000000000000000000000000000033
>>> hex(a // (a // b))
'0x1000000000000000100000000000000010000000000000000'
```
Just wanted to double check that the test is still exercising the code path we want to check. Otherwise we could fallback to constantly push `0x1000000000000000000000000000000000000000000000033` onto the stack.

tests/benchmark/compute/instruction/test_arithmetic.py
```diff
@@ -101,6 +114,15 @@
                 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFCD,
             ),
         ),
+        pytest.param(
+            # Same as the 2**192 DIV case above, with the numerator made
+            # positive and the divisor made negative.
+            Op.SDIV,
+            (
+                0x7FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F,
+                0xFFFFFFFFFFFFFFFEFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFCD,
```
marioevz: I have not checked this one but it's possible that the same comment as the other case applies here too.

tests/benchmark/compute/instruction/test_memory.py
```diff
@@ -127,3 +132,74 @@ def test_mcopy(
...
+                Op.CALL(
+                    gas=Op.GAS,
+                    address=frame_address,
+                )
+            )
+        ),
+    )
+
+
+@pytest.mark.parametrize("depth", [1, 64, 256])
```
marioevz, line 160: I'm not sure if it's worth it to parametrize by depth here. Perhaps the contract should try to go as deep as possible given the memory size and then exhaust the gas there.

We could simplify the contract to always try to descend and then unconditionally exhaust the remaining gas:
```python
entry_address = pre.deploy_contract(
        code=frame_memory + descend + WhileGas(body=Op.POP(Op.MLOAD(Op.PUSH0)), fork=fork)
    )
```
So when it cannot descend anymore it will still try to exhaust the gas.

marioevz (approved):
Looks good overall. Just a couple of suggestions, but I leave it up to you to decide if you think they're worth applying. Optimistically approving for now!

tests/benchmark/compute/instruction/test_memory.py
LouisTsai-Csie: Limiting the depth because we're not testing recursion limits, but rather stress-testing memory allocation. Creating a huge number of frames with initialized memory could cause each frame to consume memory, which can lead to resource exhaustion.

tests/benchmark/compute/instruction/test_arithmetic.py
LouisTsai-Csie: After reviewing these scenario again, I am uncertain whether it is still working as expected, so I lean toward removing these two cases, thanks for the review.

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -521,3 +523,139 @@ def test_storage_access_warm_benchmark(
...
+            gas=Op.GAS,
+            address=Op.ADDRESS,
+            args_offset=0,
+            args_size=32,
+            address_warm=True,
+        ),
+        if_false=Op.REVERT(0, 0),
+    )
+    invoke_leaf = Op.CALL(
+        gas=Op.GAS, address=leaf_address, address_warm=True
```
spencer-tb: The leaf is cold once per tx, and under EIP-8037 the cold/warm delta stays execution gas, so the warm hint underprices, so true worst case sets `address_warm=false`.
```suggestion
        gas=Op.GAS, address=leaf_address, address_warm=False
```

tests/benchmark/compute/instruction/test_memory.py
```diff
@@ -127,3 +132,74 @@ def test_mcopy(
...
+    )
+
+    descend = Op.MSTORE(0, Op.SUB(Op.CALLDATALOAD(0), 1)) + Op.POP(
+        Op.CALL(
+            gas=Op.GAS,
+            address=Op.ADDRESS,
+            args_offset=0,
+            args_size=32,
+        )
+    )
```
spencer-tb: From claude, feels legit:

The descent and leaf CALLs are POPped with no failure check and no post sentinel, so a mid chain failure silently degrades the benchmark to a shallow run. The storage twin already reverts on failure, mirroring that here covers the descent half (a leaf SSTORE sentinel checked in post would close the rest).

```suggestion
    descend = Op.MSTORE(0, Op.SUB(Op.CALLDATALOAD(0), 1)) + Conditional(
        condition=Op.CALL(
            gas=Op.GAS,
            address=Op.ADDRESS,
            args_offset=0,
            args_size=32,
        ),
        if_false=Op.REVERT(0, 0),
    )
```

spencer-tb (changes requested):
Couple of comments that should be easy to address!

tests/benchmark/compute/instruction/test_storage.py
```diff
@@ -521,3 +523,139 @@ def test_storage_access_warm_benchmark(
...
+        if_true=descend,
+        if_false=invoke_leaf,
+    )
+    entry_address = pre.deploy_contract(code=frame_code)
+
+    frame_gas = frame_code.execution_cost(fork)
+    leaf_gas = leaf_code.gas_cost(fork)
+    leaf_state_gas = leaf_code.state_cost(fork)
+
+    def deepest_frame(execution_gas: int, reservoir_gas: int) -> int:
```
spencer-tb: Filling `test_nested_frame_state_access` with `--gas-benchmark-values 5,10,30` makes every `SSTORE-max` variant fail with `receipt status 0, expected 1` (both revert values and fixture formats).

CI misses this because `fill-benchmark` only runs 1M.

I think this function needs updated! But didn't check super deep here :)

spencer-tb (commented):
Another small finding!

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: Nice catch!

tests/benchmark/compute/instruction/test_memory.py
LouisTsai-Csie: Nice idea, thank you!

tests/benchmark/compute/instruction/test_storage.py
LouisTsai-Csie: I'm not entirely sure. I tested locally but couldn't reproduce the issue.

```python
uv run fill -v tests/benchmark/compute/instruction/test_storage.py::test_nested_frame_state_access --gas-benchmark-values 5,10,30 --clean --evm-bin <path-to-geth-devnet-8-bin>
```

All passing without the receipt check failure.

spencer-tb (approved):
Approved, another small change!

The high depth variants in `test_memory` run out of gas mid descent and revert unnoticed, so the benchmark never actually reaches the advertised depth. 

Fixed here if you want it: https://github.com/spencer-tb/execution-specs/commit/b07128d21cd1c2f8b5c33ff677c6bcd921228020, feel free to cherry-pick.

PR #3352, opened 2026-08-11
fix(spec-specs): carry excess blob gas across fork transitions

spencer-tb:
### Description

`calculate_excess_blob_gas` only reads the parent's blob gas fields when the parent is the current fork's own `Header`. At a fork boundary the parent is the previous fork's header type, so chain validation zeroes any accumulated excess blob gas at the transition block. Clients carry it across the boundary, so EELS rejects correct chains at every post-Cancun transition. The path is unreachable through t8n, which builds the parent from the env as the current fork's class, so fills and fixtures were always correct.

The fix widens the check to also accept the previous fork's header, on every fork whose predecessor carries the fields (prague through amsterdam). Cancun keeps the zeroing, since its shanghai parent has no blob gas fields.

Demonstration at the BPO5 to Amsterdam boundary, with a parent at `excess_blob_gas=10000000` and `blob_gas_used=786432`:

- before: `calculate_excess_blob_gas(parent)` returns `0`
- after: it returns `8951424`

Receipts:

- Existing coverage already pins the carried value in fixtures: `test_fork_transition_excess_blob_gas_post_blob_genesis` runs at the transition to Prague and every later fork and verifies the fork block header. Those fixtures were always correct. This change fixes the chain validation half so EELS accepts them.
- 2706 blob fixtures fill byte for byte identically before and after the fix.
- `ethereum-spec-lint`, full mypy and `just static` are clean.

Found while reviewing `projects/zkevm`, which carries the same fix for amsterdam.

Adding @jsign for credit, added as a co-author.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/gas.py
+++ b/src/ethereum/forks/amsterdam/vm/gas.py
@@ -878,13 +878,14 @@ def calculate_excess_blob_gas(
         The excess blob gas for the current block.
 
     """
-    # At the fork block, these are defined as zero.
+    # Defaults for a parent without blob gas fields.
     excess_blob_gas = U64(0)
     blob_gas_used = U64(0)
     base_fee_per_gas = Uint(0)
 
-    if isinstance(parent_header, Header):
-        # After the fork block, read them from the parent header.
+    if isinstance(parent_header, (Header, PreviousHeader)):
+        # Read them from any parent that carries the fields, so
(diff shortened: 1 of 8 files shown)


gurukamath (approved):

spencer-tb:
Don't merge without ack from @jsign please!

jsign (approved):

PR #3361, opened 2026-08-11
fix(evm-tools): encode accepted transactions in t8n body

skbaek:
### Description

`t8n` currently builds its emitted block body before execution with
`rlp.encode([tx.rlp() for tx in self.txs])`. `Transaction.rlp()` already
returns the complete encoded transaction, so putting those bytes directly in
another RLP list wraps legacy transactions as byte strings instead of encoding
them as inline lists. Building the body before execution also includes
transactions that `t8n` later reports as rejected.

This change records a transaction only after `process_transaction` accepts it,
then builds the body after execution using the active fork's canonical
transaction representation: legacy transactions remain inline RLP lists and
typed transactions remain envelope byte strings. It also removes the unused
duplicate body construction from the CLI JSON parser.

The regression test uses a legacy transaction, a rejected legacy transaction,
and an accepted type-2 transaction. It checks that the rejected input is
absent, the accepted legacy transaction is inline, and the type-2 envelope is
preserved.

On that case, the old output is a 307-byte, three-member body beginning
`0xf90130b862f860`; the new output is a 204-byte, two-member body beginning
`0xf8caf860`, byte-identical to Geth 1.15.6.

Found while building a t8n frontend for [Jaune](https://github.com/skbaek/jaune), a Lean 4 executable EVM specification whose conformance gate is byte-exact against `ethereum-spec-evm`.

#### Validation

```console
just static
just test-tests
just spec-tools
```

`just test-tests` passes with 2,080 tests passed; it is the recipe that
collects `packages/testing/src/execution_testing/evm_tools/tests/test_t8n_body.py`.
`just spec-tools` separately passes all 22 tests.

### Related Issues or PRs

N/A. The separately observed trie-indexing behavior is intentionally not
changed by this PR.

### Checklist

### Cute Animal Picture

long-tailed tit


zexoverz:
Heads up that this needs a rebase. #3307 moved the tree while this was open, and both paths this PR touches are gone on `forks/amsterdam` at `3d473e8`:

```
src/ethereum_spec_tools/evm_tools/   ->  packages/testing/src/execution_testing/evm_tools/
tests/evm_tools/                     ->  packages/testing/src/execution_testing/evm_tools/tests/
```

The second one matters more than a path change. I could not find any Just recipe or CI job that runs `tests/evm_tools`. `just spec-tools` runs `tests/spec_tools`, `just test-tests` runs the `packages/testing` tree, and the other recipes point at `tests/json_loader`, `tests/ported_static` and `tests/benchmark`. So `test_t8n_body.py` would not have been collected where it currently sits, which is worth knowing before you read anything into the run you quoted.

On HEAD the collected home for a t8n regression test is `packages/testing/src/execution_testing/evm_tools/tests/`, which `just test-tests` picks up. The vendored fixture layout under `.../evm_tools/tests/fixtures/` that came in with #3307 is a close model for the testdata files here.

I was reading #3362 and reproduced the sparse receipts trie on Prague, keys `{0, 2}` for a three transaction body with the middle one rejected. It is real. I am not going to open anything there since you said you would rather have maintainer direction first, and the two candidate directions push `txRoot` opposite ways, so it does look like a call someone else has to make.

skbaek:
@zexoverz Thanks for the heads-up. I rebased onto the current `forks/amsterdam`, moved the regression test and fixtures under `packages/testing/src/execution_testing/evm_tools/tests/` so they’re collected by just `test-tests`, and updated the Validation section. `just static`, `just test-tests`, and `just spec-tools` all pass locally. CI is currently awaiting workflow approval.

PR #3364, opened 2026-08-12
feat(test-fill): make engine_x fixtures consumable via devp2p sync

danceratopz:
### Description

This PR lets the filler append an empty sync payload above each chain represented in a `blockchain_test_engine_x` fixture and store it separately in a new optional `syncPayloads` field:

```text
G → T₁ … Tₙ → S*
```

`*` marks the extra sync payload announced by a sync-based consumer. 

The sync payload's parent is unknown to the client, so the client starts syncing and fetches every test block below it via devp2p.

For more context, the detailed documentation is intended to be read in this order:

1. [Sync Payloads](https://github.com/ethereum/execution-specs/blob/experiments/wirex-fill/docs/filling_tests/sync_payloads.md) explains the design and mechanism.
2. [Consumption in the EngineX fixture-format reference](https://github.com/ethereum/execution-specs/blob/experiments/wirex-fill/docs/running_tests/test_formats/blockchain_test_engine_x.md#consumption) explains how these fixtures are consumed through the Engine API and devp2p full-sync paths.

> [!IMPORTANT]
> **Companion PR:** #3365 adds `consume wirex` and the devp2p peer that use these sync payloads. Ordinary Engine API consumers ignore `syncPayloads`; the [EngineX consumption guide](https://github.com/ethereum/execution-specs/blob/experiments/wirex-fill/docs/running_tests/test_formats/blockchain_test_engine_x.md#consumption) keeps the authored Engine API sequence separate.

#### Fixture shape

In abbreviated form, a fixture with one test block looks like this:

```json
{
  "engineNewPayloads": [
    {
      "params": [
        {
          "blockNumber": "0x1",
          "blockHash": "0xabc...",
          "transactions": ["0x..."]
        }
      ]
    }
  ],
  "syncPayloads": [
    {
      "params": [
        {
          "parentHash": "0xabc...",
          "blockNumber": "0x2",
          "transactions": [],
          "extraData": "0x1234..."
        }
      ]
    }
  ],
  "lastblockhash": "0xabc..."
}
```

The matching `parentHash` shows that the sync payload is appended above the test block. `lastblockhash` continues to identify the test block, not the sync payload.

<details>
<summary>Existing sibling-chain examples</summary>

- The [EIP-7594 fork-transition test](https://github.com/ethereum/execution-specs/blob/experiments/wirex-fill/tests/osaka/eip7594_peerdas/test_max_blob_per_tx.py) produces the two pre-Amsterdam cases: `blob_count_10` has three expected-invalid siblings, while `blob_count_7` has one valid block followed by two expected-invalid siblings.
- The [EIP-7954 fork-transition test](https://github.com/ethereum/execution-specs/blob/experiments/wirex-fill/tests/amsterdam/eip7954_increase_max_contract_size/test_fork_transition.py) produces an expected-invalid/valid sibling pair.
- The [EIP-7981 fork-transition test](https://github.com/ethereum/execution-specs/blob/experiments/wirex-fill/tests/amsterdam/eip7981_increase_access_list_cost/test_fork_transition.py) produces three chain heads across four test payloads.

</details>

#### PR-specific notes

- This follows the same "empty block above the test chain" mechanism introduced for `blockchain_test_sync` by [execution-spec-tests #2007](https://github.com/ethereum/execution-spec-tests/pull/2007). It adds a test-specific value to `extraData` and supports one sync payload per sibling-chain head, but does not replace the existing code that generates the singular `syncPayload`; #3405 makes both fixture formats use the same builder.
- No consensus test files, test blocks, assertions, or markers are changed.
- The same node-ID normalization removes pytest-xdist's `@bigmem` scheduling suffix from published fixture IDs. This changes 1,299 IDs. Because packed pre-allocation groups are keyed by member IDs, 23,351 EngineX fixtures also receive new, internally consistent `preHash` pointers. These are release identity changes; they do not change the test payloads.

### Related Issues or PRs

Consume-side counterpart:
- #3365.

Follow-up:
- https://github.com/ethereum/execution-specs/pull/3405

### Checklist

- [x] Ran fast static checks: `just static`
- [x] PR title has the form `<type>(<area>): <title>` matching `C-*`/`A-*` labels; the title matches the target squash commit message.

### Cute Animal Picture

every chain deserves a block it can sync from


danceratopz:
@marioevz @spencer-tb I realized that I wasn't entirely sure whether invalid test blocks are guaranteed to be sent and rejected via devp2p, so I revisited this. 

Prepending a new block is imo the most contentious change in this PR.

Appending doesn't change the chain or test expectation. It also doesn't incur much effort. 

Prepending, however, does require effort to ensure that it doesn't change test expectations. As it turns out it's currently only worth it for 3/6 clients (which could also change against our favour in the future). 

For an invalid singleton, WireX uses `G → S → Tᵢ*`: `newPayload` supplies the complete invalid head `Tᵢ`, while its unknown salted parent `S` gives the client a reason to sync. The six baseline clients currently split into two groups:



| Announced `Tᵢ` body over devp2p? | Clients                    | Result                                                                                                                                              |
| -------------------------------- | -------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Yes**                          | **geth, Nethermind, Besu** | The invalid test block crosses devp2p and is processed through the sync/backfill path. This is useful coverage beyond EngineX.                      |
| **No**                           | **reth, ethrex, Erigon**   | The client syncs the missing ancestry, then judges `Tᵢ` using the payload already supplied by `newPayload`. The invalid body does not cross devp2p. |

Both behaviors are correct. Only blocks below the announced head are required to travel over devp2p; refetching the head body is an implementation choice. Fresh checks reproduced both sides: geth requested `Tᵢ` and rejected it during beacon backfill, while ethrex requested only `S` and reused the Engine-delivered `Tᵢ`. Reth also reproduced the ancestors-only shape.

WireX asserts that test-block bodies were transported over devp2p, but only for valid tests, where every test block is placed below the announced sync head. It does not assert this for an invalid head because transport of the announced head's body is not enforceable across clients.

Currently, the prepend gives invalid-head devp2p ingestion coverage on three of six clients, but Geth, Nethermind or Besu could legitimately adopt the other group's optimization later, at which point the prepend would become futile for testing invalid-head body transport (although it would still exercise missing-ancestry sync followed by rejection).

We could ask clients to expose a test/debug flag that forces head-body download, but that adds per-client code, Hive configuration and ongoing maintenance, while testing a path different from the client's normal production behavior.

My conclusion is that the present three-client benefit makes the prepend worthwhile, but the PR should describe it as measured, client-dependent coverage rather than a guarantee. I also think it's worth reaching it to client teams, but wanted to get your take first.

danceratopz:
I think the obvious and simple solution to this is to drop the prepend and append a second invalid synthetic sync trigger block. This should trigger the sync; both invalid blocks should be rejected! Will update.

danceratopz:
Updated the PR with the new simpler approach; no test payloads are modified.

During self-review I learnt that we have tests today that include blocks that are siblings. This is typically used to test new fork behavior over a fork transition: The first block (before the transition) is invalid (feature unavailable) and the second block (after the transition) is valid. Both blocks build on the same valid block. Here's an example for [eip7954_increase_max_contract_size/test_fork_transition.py: test_max_initcode_size_fork_transition](https://github.com/ethereum/execution-specs/blob/43e3cd1ec03dd550a2bd43746564b339e85fcd1d/tests/amsterdam/eip7954_increase_max_contract_size/test_fork_transition.py#L166). To test both paths via wirex, both paths need an appended sync block and this has been added to `fill`, at the cost of some complexity.

docs/filling_tests/sync_payloads.md
```diff
@@ -0,0 +1,95 @@
...
+
+Client behavior diverges at this point. Some clients request the announced payload's body over devp2p together with its missing ancestors. Other clients request only the ancestors because the announced payload's body was already supplied through `engine_newPayload`. For those clients, the announced payload never travels over devp2p. Avoiding that redundant request is a valid client-side optimization, so the fixture cannot rely on the first behavior.
+
+The fixture accommodates both behaviors by appending and announcing an additional sync payload. This makes the final test payload part of the missing ancestry, ensuring that every test payload is covered by devp2p full-sync testing.
+
+## Each sync payload is salted to ensure uniqueness
+
+Test cases are intended to be unique, but complex parametrization can sometimes produce cases with byte-identical payload graphs. Avoiding these duplicates is a fill-side or test-author concern. While a test case remains in the fixture set, however, consumers must still be able to run its devp2p full-sync test.
```
marioevz, line 29: Interesting, could we rather try to identify this behavior and make it a test exception? If we can have two different tests with byte identical payloads, it sounds to me like it's a waste of space. It might be difficult to detect due to x-dist though. If we can, it would be great because we would be able to remove the `extraData` from these sync payloads.

docs/filling_tests/sync_payloads.md
```diff
@@ -0,0 +1,95 @@
...
+
+The filler therefore places a deterministic value derived from the test ID in every sync payload's `extraData`. This gives the sync payload a test-specific `blockHash`, even when another test case has an identical chain. A reused client cannot mistake it for a sync payload it has already processed and skip the test. Sync payloads above different chain heads are already distinct because their `parentHash` values differ.
```
marioevz, line 31: What about making the fee-recipient unique? It does not affect the state-root since there's no transactions, and it does not really matter what we set it to.

packages/testing/src/execution_testing/forks/forks/forks.py
```diff
@@ -330,6 +330,19 @@ def minimum_block_gas_limit(cls) -> int:
             )
         return max(minimum_block_gas_limit, bal_minimum_block_gas_limit)
 
+    @classmethod
+    def block_time(cls) -> int:
```
marioevz, line 334: From what I can tell, block step time is not really part of the spec so I would rather not put this information in the fork object. I think we can make do with the simpler solution of just defining a constant at the top of `specs/blockchain.py`.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1263,6 +1472,15 @@ def make_hive_fixture(
     ) -> FillResult:
         """Create a hive fixture from the blocktest definition."""
         fixture_payloads: List[FixtureEngineNewPayload] = []
+        sync_payload_candidates: List[
+            Tuple[BuiltBlock, Alloc | LazyAlloc]
+        ] = []
+        sync_payload_leaf_hashes: set[Hash] = set()
+        sync_payload_leaf_indices = (
+            set(self.sync_payload_leaf_indices())
+            if fixture_format == BlockchainEngineXFixture
```
marioevz, line 1481:
```suggestion
            if 'sync_payloads' in fixture_format.format_class().model_fields
```

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -103,6 +105,91 @@
...
+    # not fill bare: the typed payload model rejects the overflowing
+    # child and fails the whole fill. The fields are semantic and never
+    # clamped or shifted; the fill declines the block instead.
+    if int(head.number) + 1 > 2**64 - 1:
+        return "the head's block number has no successor in uint64"
+    if int(head.gas_limit) > 2**64 - 1:
+        return "the head's gas limit does not fit uint64"
+    block_time = test_fork.fork_at(
+        block_number=int(head.number), timestamp=int(head.timestamp)
+    ).block_time()
```
marioevz, line 162: Mentioned in a previous comment, we should instead simply define a constant at the top of this file instead of adding it to the `Fork` object (block step time is not defined at the spec level).

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -103,6 +105,91 @@
...
+    # budget an empty block's own access list needs.
+    if int(head.gas_limit) < fork.minimum_block_gas_limit():
+        return (
+            "the head's gas limit is below the fork's minimum, and the "
+            "appended block inherits it"
+        )
+    if head.excess_blob_gas is None or head.blob_gas_used is None:
+        return None
+    excess_blob_gas = int(head.excess_blob_gas)
+    if excess_blob_gas + int(head.blob_gas_used) > 2**64 - 1:
```
marioevz, line 182: `max(0, excess_blob_gas + int(head.blob_gas_used) - fork's blob target)` I think.

packages/testing/src/execution_testing/specs/base.py
```diff
@@ -146,6 +146,31 @@ class BaseTest(BaseModel):
...
+
+    A sync-based consumer announces each appended block, which makes
+    every representable payload the test author wrote an ancestor of at
+    least one target the client must fetch and execute through its
+    devp2p sync path. The test's own directives are filled exactly as
+    written. Enabled unless ``--no-sync-block`` withholds them or the
+    test itself opts out by passing ``sync_block=False``; see "Sync
+    Payloads" in the filling-tests docs.
+    """
+    sync_block_salt: str = ""
```
marioevz, line 164: I rather generalize this as `test_id: str`, non-optional so it is guaranteed to be available (so we'd have to add it to `execute` too). Then it can be obtained and used by the sync block generator or any other functionality that requires uniqueness across tests.

packages/testing/src/execution_testing/specs/base.py
```diff
@@ -146,6 +146,31 @@ class BaseTest(BaseModel):
         # instead of each test having to set it
     )
     operation_mode: OpMode | None = None
+    sync_block: bool = False
```
marioevz, line 149: The only problem is that `filler.py` currently overrides this default to `True` by using the `kwargs` getter, so this default is misleading.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -1255,6 +1353,117 @@ def make_fixture(
...
+                f"no sync payload appended above leaf "
+                f"{head.header.block_hash}: {unavailable}"
+            )
+            return None
+        env = apply_new_parent(head.env, head.header)
+        extra_data = Bytes(sha256(self.sync_block_salt.encode()).digest()[:16])
+        try:
+            sync_block = self.generate_block_data(
+                t8n=t8n,
+                block=Block(extra_data=extra_data),
```
marioevz, line 1449:
```suggestion
        fee_recipient = Address(sha256(self.sync_block_salt.encode()).digest()[:20])
        try:
            sync_block = self.generate_block_data(
                t8n=t8n,
                block=Block(fee_recipient=fee_recipient),
```
Suggestion to use the `fee_recipient` as salt. The `stateRoot` remains unaffected by this change in all sync blocks.

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -1031,6 +1031,28 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
     )
     """Engine API payloads for blockchain execution."""
 
+    sync_payloads: List[FixtureEngineNewPayload] | None = None
```
marioevz, line 1034: How does consume generate the RLP that goes into the wire for the other payloads?

marioevz (commented):
Looks generally good to me!

The only suggestion I have is that we please consider making this optional and limited to tests with a specific marker.

PR #3367, opened 2026-08-13
refactor(client-clis): bound the opcode-count trace time explicitly

skylenet:
### Description

`extract_block_opcode_count` calls `debug_traceBlockByHash` without a `timeout`, so the client falls back to its own default — **5s per transaction on geth**. That is far below what a benchmark block needs.

Stepping the JS tracer runs at roughly 460k opcodes/s per transaction, so any transaction executing more than ~2.3M opcodes blows the deadline. geth abandons that trace and returns an error where the entry's `result` would be. `_opcode_count_from_js_tracer` skips an entry with no usable result, so the block still produces a **well-formed tally that is short by whole transactions** — and where every transaction times out, the fixture records no count at all. Nothing is raised and nothing is logged.

This is not a corner case for the compute benchmarks. In a 4884-fixture jochemnet build:

- **2182 (44.7%)** carry no opcode count for their benchmark block.
- The split falls exactly along step count, not along anything test-specific. `EXP` is counted, `ADD`/`MUL`/`SUB` are not. `codecopy` is counted at `code_size_1024` but not at `code_size_0` or `code_size_32`; `mcopy` at `copy_size_256` but not `copy_size_0`/`32`. The affected share climbs with the gas value, 42.3% at 100M to 48.4% at 300M.
- Cross-referencing `.benchmarkoor-pytest-report.json`, all 2182 have a `call` duration of **at least 5.42s**, with a hard floor right there, while fixtures that *did* get a count run as long as 250s. That floor is the 5s default plus overhead.

### Fix

Inject `timeout` in `_trace_block`, which covers the JS tracer and the struct-log fallback in one place. `timeout` is the standard geth-compatible `TraceConfig` field. The `tracer_config` spread comes second, so an explicit per-call value still wins.

The value needs to be one no legitimate trace can reach while still capping one that has genuinely hung; `1h` is ~100x the slowest transaction I measured.

### Verification

Re-ran all 2182 affected blocks against the same geth build (`ethpandaops/geth:glamsterdam-devnet-7`, `v1.17.6-unstable-bdf6e17f`) on the same stateful datadir, with `timeout` set:

- Every one produces a count, with all transactions covered.
- The heaviest, `test_push[...opcode_PUSH0...value_300M]`, tallies **143,118,396 opcodes in 25.6s** — 5x over the old 5s ceiling on its own, and that is with geth tracing the block's transactions in parallel.
- **Control:** 12 fixtures that already had counts reproduce **byte-for-byte**.
- `test_identity_uncachable[size_32, 300M]` now yields exactly **1,693,322** `STATICCALL`s with all 18 transactions covered. That case had previously been read as a benchmark bug when a single lost transaction left the tally 94,698 short.

### Related Issues or PRs

N/A.

### Checklist

Also ran `just test-tests` (1997 passed). The two new tests fail without the change and pass with it.

### Cute Animal Picture

red panda

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/fill_stateful/fill_stateful.py
@@ -30,6 +30,9 @@ from execution_testing.base_types import (
 )
 from execution_testing.client_clis import ClientBackend
 from execution_testing.client_clis.cli_types import EnginePayloadMetadata
+from execution_testing.client_clis.client_backend import (
+    DEFAULT_OPCODE_COUNT_TRACE_TIMEOUT,
+)
 from execution_testing.fixtures import FixtureFillingPhase
 from execution_testing.fixtures.blockchain import (
     StatefulPreRunFixture,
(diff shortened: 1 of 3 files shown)


LouisTsai-Csie (approved):
Thank you! I add a refactor commit

PR #3370, opened 2026-08-13
fix(tests,tools): make the EIP-7805 FOCIL tests fill on the current Amsterdam

ilitteri:
### Description

The FOCIL tests on this branch don't fill: 21 of 24 fail. Three independent causes, none of them in the tests' assertions. With these fixed all 24 fill, and the expected outcomes are unchanged.

**1. `t8n` passes decoded transactions where raw ones are expected**

`check_inclusion_list_transactions` takes `Tuple[LegacyTransaction | Bytes, ...]` and calls `get_transaction_hash`, which asserts `isinstance(tx, (LegacyTransaction, Bytes))`. The t8n caller passes `convert_transaction(...)`, which returns a decoded fork transaction, so every typed-transaction case dies on `AssertionError`. Both tuples now go through `encode_transaction`.

**2. The scenarios leave no gas budget for the block's own access list**

They size `block_gas_limit` to their transactions alone. The EIP-7928 budget is `block_gas_limit // GAS_BLOCK_ACCESS_LIST_ITEM`, and the system-contract predeploys touched every block plus senders, recipients and coinbase come to ~30 items against a limit of ~22, so filling stops at `BlockAccessListGasLimitExceededError`.

Raising the limit alone would change what the scenarios assert, because the gas a pending inclusion-list transaction is allowed derives from `block_gas_limit - gas_used_by_included_txs`. So `build_block` raises the limit and spends exactly the same amount on ballast transactions, leaving `remaining_gas` — and therefore every "fits" / "does not fit" boundary — unchanged. The ballast is a whole number of empty transfers so the amount is always representable on the fork's calldata-cost lattice.

**3. One funding transfer predates value-carrying transactions costing extra intrinsic gas**

`test_unsatisfied_when_block_tx_funds_pending_il_sender` sizes and funds alice's transfer with `calc()`, but it carries value, so it needs `calc(sends_value=True)` and was failing `INTRINSIC_GAS_BELOW_FLOOR_GAS_COST`.

### Assertions are untouched

Of the changed lines in `test_focil.py`, none touch `expected_status`, `pytest.param`, assertions, sender balances, nonces or inclusion-list composition — they are gas-limit plumbing, imports, and the `sends_value` fix. Scenario ids are unchanged, and the reference implementation validates all 24 expected outcomes during fill.

### Deliberately not included

This branch's Amsterdam differs from glamsterdam-devnet-8 (`ACCOUNT_WRITE` 8000 vs 9000, `CREATE_ACCESS` derived from `COLD_STORAGE_ACCESS` rather than `COLD_ACCOUNT_ACCESS`), so a contract creation's intrinsic gas is 23000 here against devnet-8's 24000. That is a rebase question rather than a bug, so the gas schedule is left alone — flagging it because `unsatisfied_with_contract_creating_pending_il_tx` is the one scenario whose outcome moves with it.

Also worth noting: #3307 relocates `evm_tools`, so change 1 will need rebasing onto the new path if that lands first.

### Testing

`uv run fill tests/bogota/eip7805_focil --fork Bogota` → 24 passed.

The filled fixtures were additionally run against a client implementation (ethrex) as a cross-check.

### DIFF
--- a/tests/bogota/eip7805_focil/helpers.py
+++ b/tests/bogota/eip7805_focil/helpers.py
@@ -52,9 +52,11 @@ class PendingInclusionListTx(NamedTuple):
 # limit and spends exactly the same amount on a ballast transaction, leaving
 # `remaining_gas` — and therefore every "fits" / "does not fit" boundary —
 # byte-for-byte what it was.
-# Gas to add. Realised as a whole number of empty transfers so the amount is
-# always representable on the fork's calldata-cost lattice (a flat target like
-# 64_000 is not, once EIP-7623's floor makes marginal calldata cost non-linear).
+#
+# The headroom is realised as a whole number of empty transfers so the amount
+# is always representable on the fork's calldata-cost lattice (a flat target
+# like 64_000 is not, once EIP-7623's floor makes marginal calldata cost
+# non-linear).
 
 
 def bal_gas_headroom(fork: Fork) -> int:
--- a/tests/bogota/eip7805_focil/helpers.py
+++ b/tests/bogota/eip7805_focil/helpers.py
@@ -55,13 +55,16 @@ class PendingInclusionListTx(NamedTuple):
 # Gas to add. Realised as a whole number of empty transfers so the amount is
 # always representable on the fork's calldata-cost lattice (a flat target like
 # 64_000 is not, once EIP-7623's floor makes marginal calldata cost non-linear).
-BAL_GAS_HEADROOM_TARGET = 64_000
 
 
 def bal_gas_headroom(fork: Fork) -> int:
     """Gas added to a block's limit to leave room for its own access list."""
     empty_transfer_gas = fork.transaction_intrinsic_cost_calculator()()
-    count = -(-BAL_GAS_HEADROOM_TARGET // empty_transfer_gas)
+    bal_gas_headroom_target = (
+        fork.empty_block_bal_item_count()
+        * fork.gas_costs().BLOCK_ACCESS_LIST_ITEM
+    )
+    count = -(-bal_gas_headroom_target // empty_transfer_gas)
     return count * empty_transfer_gas
 
 
--- a/tests/bogota/eip7805_focil/helpers.py
+++ b/tests/bogota/eip7805_focil/helpers.py
@@ -36,6 +36,35 @@ class PendingInclusionListTx(NamedTuple):
     sender_balance: int = 10**18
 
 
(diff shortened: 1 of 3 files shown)


jihoonsong:
cc @marioevz

marioevz:
@ilitteri we need to rebase after https://github.com/ethereum/execution-specs/pull/3307 lands anyway. I'll review this once we:
1) have merged 3307
2) rebased focil branch on top of the latest `forks/amsterdam` that contains 3307

tests/bogota/eip7805_focil/helpers.py
```diff
@@ -36,6 +36,35 @@ class PendingInclusionListTx(NamedTuple):
...
+# Raising the limit alone would change what the scenarios assert, because the
+# gas a pending IL transaction is allowed is derived from
+# `block_gas_limit - gas_used_by_included_txs`. So `build_block` raises the
+# limit and spends exactly the same amount on a ballast transaction, leaving
+# `remaining_gas` — and therefore every "fits" / "does not fit" boundary —
+# byte-for-byte what it was.
+# Gas to add. Realised as a whole number of empty transfers so the amount is
+# always representable on the fork's calldata-cost lattice (a flat target like
+# 64_000 is not, once EIP-7623's floor makes marginal calldata cost non-linear).
+BAL_GAS_HEADROOM_TARGET = 64_000
```
marioevz:
```suggestion
```
Not necessary, see comment below.

tests/bogota/eip7805_focil/helpers.py
```diff
@@ -36,6 +36,35 @@ class PendingInclusionListTx(NamedTuple):
...
+
+
+def bal_gas_headroom(fork: Fork) -> int:
+    """Gas added to a block's limit to leave room for its own access list."""
+    empty_transfer_gas = fork.transaction_intrinsic_cost_calculator()()
+    count = -(-BAL_GAS_HEADROOM_TARGET // empty_transfer_gas)
```
marioevz:
```suggestion
    bal_gas_headroom_target = (
        fork.empty_block_bal_item_count()
        * fork.gas_costs().BLOCK_ACCESS_LIST_ITEM
    )
    count = -(-bal_gas_headroom_target // empty_transfer_gas)
```
We can use this deterministic fork method to make this value dynamic.

marioevz (approved):
LGTM. I'll merge and then rebase on top of latest `forks/amsterdam`. Thanks!

PR #3374, opened 2026-08-14
refactor(test-benchmark): remove potential duplicated cases

LouisTsai-Csie:
### Description

Removes stateful benchmark cases under `tests/benchmark/stateful/bloatnet/` that are now covered by latest framework-backed benchmarks.

The removed tests were written before the framework has proper support for pre-deployed state: each of them had to call a bloatnet factory contract's `getConfig()` at runtime to learn the number of deployed contracts and the initcode hash, then recompute every target address with a CREATE2 preimage inside the attack loop. `AccountCreator`/`AccountMode`, `stub_parametrize` and the `benchmark_test` gas accounting now express the same intent directly, so the factory indirection (and the per-test bookkeeping around it) is no longer needed.

| Removed | File | Replacement |
| --- | --- | --- |
| `test_extcodesize_bytecode_sizes` | `test_account_query.py` | `test_account_access` (`opcode=EXTCODESIZE`) |
| `test_balance_query` | `test_account_query.py` | `test_account_access` (`opcode=BALANCE`, plus the `EXTCODE*`/`CALL*` variants) |
| `test_call_value_to_existing` | `test_call.py` | `test_account_access` (`opcode=CALL`, `value_sent=1`) |
| `test_sload_erc20_generic` | `test_erc20.py` | `test_sload_bloated` |
| `test_sstore_erc20_generic` | `test_erc20.py` | `test_sstore_bloated` |
| `test_mixed_sload_sstore` | `test_erc20.py` | — see open question below |

`test_account_access` already parametrizes over every `AccountMode` (minimal contract, max-size same/different/JUMPDEST code, EOA, non-existing account), value-bearing vs. non-value-bearing calls, the CREATE2 overhead baseline and the cache strategy, so the removed cases are strict subsets of it. `test_sload_bloated` / `test_sstore_bloated` drive the opcodes directly against a bloated EOA instead of going through ERC20 `balanceOf` / `approve`, and add `existing_slots` / `write_new_value` coverage the ERC20 versions did not have.

cc @CPerezz 

### Related Issues or PRs

issue https://github.com/ethereum/execution-specs/issues/3106#issuecomment-4934900449

### Checklist

### Cute Animal Picture


marioevz:
On the open question from the linked issue:

>     * Keep ERC20 scenarios? Could replace with SSTORE/SLOAD benchmarks for more focused storage interaction testing.

I would say no, because tests solely focused on SSTORE or SLOAD give us, in my opinion, better information about the bottlenecks we are trying to measure. If we really want a SSTORE/SLOAD mixed test it's fine, but we can use the methods we've been using that are not directly linked to ERC20 tests.

marioevz (approved):
LGTM, thanks for the detailed table in the PR description for the replacements of these tests.

PR #3383, opened 2026-08-17
refactor(tests): EIP-8037 test gas calculation logic and post state verification

LouisTsai-Csie:
### Description

While extending EIP-8037 coverage I reviewed the existing suite and noticed that a number of tests do not restrict the behavior the test names and docstring intended to do. This PR marks those tests and starts tightening them.

### Case 1: set `state_gas_reservior` to 0 does not mean no state gas charge

Take `test_state_gas_selfdestruct.py::test_selfdestruct_existing_beneficiary_no_state_gas` as a representative example:

In my opinion, some of the verification is not strict enough, below is an example for unmodified `test_state_gas_selfdestruct`:

```python
@pytest.mark.valid_from("EIP8037")
def test_selfdestruct_existing_beneficiary_no_state_gas(
    state_test: StateTestFiller,
    pre: Alloc,
) -> None:
    """Test SELFDESTRUCT to existing beneficiary charges no state gas."""
    beneficiary = pre.fund_eoa(amount=0)

    contract = pre.deploy_contract(
        code=Op.SELFDESTRUCT(beneficiary),
        balance=1,
    )

    tx = Transaction(
        to=contract,
        state_gas_reservoir=0,
        sender=pre.fund_eoa(),
    )

    state_test(pre=pre, post={}, tx=tx)
```

The test intends to show that a `SELFDESTRUCT` sweep to an existing beneficiary charges no state gas. It has two independent problems.

1. It exercises the opposite scenario. `pre.fund_eoa(amount=0)` returns a fresh address but writes nothing to the pre-allocation, so the selfdestruting actually create a new account, charging the `NEW_ACCOUNT` cost. However, the test implementation does not catch the error at all.

2. In this test, the `tx.gas_limit` is not configured, while the `state_gas_reservior` configure to 0. In this case, the transaction gas limit case would be configured to transaction gas limit cap. So it is possible that a misimplemented client draw gas from `gas_left` and increase `state_gas_from_gas_left `, and the transaction is still possible to pass.

Transactions are constructed via `BlockchainTest::generate_block_data`, which calls `with_gas_limit()` to set gas limits:

```python
txs = [
    tx.with_gas_limit(
        max_gas_limit=max_tx_gas_limit,
        transaction_gas_limit_cap=fork.transaction_gas_limit_cap(),
        state_gas_reservoir_enabled=fork.state_gas_reservoir_enabled(),
    )
    for tx in txs
]
```
If `gas_limit` is not already set, `with_gas_limit()` calculates it implicitly:

```python
if "gas_limit" not in self.model_fields_set:
    updated_values["gas_limit"] = self._calculate_implicit_gas_limit(...)
```

In `_calculate_implicit_gas_limit()`, when `state_gas_reservoir_enabled` is True, the transaction gas limit becomes:

```python
tx_gas_limit = min(tx_gas_limit, transaction_gas_limit_cap) + state_gas_reservoir
```

To actually constrain the state dimension, a test needs one of: (1)  Header verification, and (2)  A transaction gas limit that excludes the state cost.

### Case 2: If the test is intended to validate state gas calculation, the header gas cost should not be derived from `max(execution_gas, state_gas)`

Take `test_create_selfdestruct_code_deposit_no_refund_header_check` as an example. This test verifies that code deposit costs are not refunded (state gas cost dimension). Currently, using `max(state_gas, execution_gas)` works fine because state gas always dominates in Amsterdam.

However, consider future upgrades where: (1) execution gas costs for certain operations increase significantly, or (2) state gas costs drop substantially. If `execution_gas` ever exceeds `state_gas`, the test would still pass -> but it would no longer validate the state gas for "no refund" scenario. The test would be measuring a different code path without catching the regression.

In other words: the test's intention (validating no-refund behavior) now depends on which gas component is larger. Using max() masks this dependency and creates a brittle test that silently fails to catch the intended behavior change.

There are several instances in the test suite, this PR adds stricter verification to tighten it.

### Scope
This PR refactors the following files, remaining refactor would be created in separate PRs.
- [x] tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
- [x] tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_eip_mainnet.py
- [x] tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_delegation_pointer.py
- [x] tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_fork_transition.py
- [x] tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_selfdestruct.py

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
@@ -424,15 +424,14 @@ def test_block_gas_used_call_new_account(
         intrinsic_gas + parent_code.execution_cost(fork) + intrinsic_gas
     )
     block_state = parent_code.state_cost(fork)
+    assert block_state > block_execution, "requires state gas to dominate"
 
     blockchain_test(
         pre=pre,
         blocks=[
             Block(
                 txs=txs,
-                header_verify=Header(
-                    gas_used=max(block_execution, block_state)
-                ),
+                header_verify=Header(gas_used=block_state),
             )
         ],
         post={parent: Account(storage=parent_storage)},
@@ -461,7 +460,9 @@ def test_block_gas_used_create_tx(
     create_state = fork.transaction_top_frame_state_gas(contract_creation=True)
     stop_execution = intrinsic_calc()
 
-    expected = max(create_execution + stop_execution, create_state)
+    assert create_state > create_execution + stop_execution, (
+        "create state should dominate"
+    )
 
     sender = pre.fund_eoa()
     txs = [
@@ -478,7 +479,7 @@ def test_block_gas_used_create_tx(
         blocks=[
             Block(
                 txs=txs,
-                header_verify=Header(gas_used=expected),
+                header_verify=Header(gas_used=create_state),
             )
         ],
         post={
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
@@ -41,9 +41,10 @@ REFERENCE_SPEC_VERSION = ref_spec_8037.version
 
 def sstore_tx_gas(fork: Fork, num_sstores: int = 1) -> tuple[int, int]:
     """Return (execution, state) gas for a tx with N cold SSTOREs."""
+    code = Op.SSTORE(0, 1, original_value=0, new_value=1)
     intrinsic_gas = fork.transaction_intrinsic_cost_calculator()()
-    evm_total = num_sstores * Op.SSTORE(0, 1).execution_cost(fork)
-    state = num_sstores * Op.SSTORE(new_value=1).state_cost(fork)
+    evm_total = num_sstores * code.execution_cost(fork)
+    state = num_sstores * code.state_cost(fork)
     return intrinsic_gas + evm_total, state
 
 
@@ -64,9 +65,9 @@ def sstore_txs(
     txs, post = [], {}
     for _ in range(n):
         storage = Storage()
-        code = Bytecode(Op.STOP)
+        code = Bytecode()
         for _ in range(num_sstores):
-            code = Op.SSTORE(storage.store_next(1), 1) + code
+            code += Op.SSTORE(storage.store_next(1), 1)
         contract = pre.deploy_contract(code=code)
         txs.append(
             Transaction(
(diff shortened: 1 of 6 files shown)


tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
```diff
@@ -371,34 +392,49 @@ def test_block_gas_used_call_new_account(
...
+    block_state = parent_code.state_cost(fork)
+
     blockchain_test(
         pre=pre,
-        blocks=[Block(txs=txs)],
+        blocks=[
+            Block(
+                txs=txs,
+                header_verify=Header(
+                    gas_used=max(block_execution, block_state)
```
spencer-tb: Could we make the intended state dominance explicit here and pin the header directly to `block_state`?

```python
assert block_state > block_execution, "requires state gas to dominate"
...
header_verify=Header(gas_used=block_state)
```

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
```diff
@@ -416,26 +452,24 @@ def test_block_gas_used_create_tx(
...
+
+    create_execution = intrinsic_calc(
+        calldata=init_code,
+        contract_creation=True,
+    ) + fork.transaction_top_frame_gas_calculator()(contract_creation=True)
+    create_state = fork.transaction_top_frame_state_gas(contract_creation=True)
     stop_execution = intrinsic_calc()
 
-    expected = max(create_execution + stop_execution, create_state_gas)
+    expected = max(create_execution + stop_execution, create_state)
```
spencer-tb: Could we assert that state dominates and pin `expected = create_state`? To ensure this test specifically validates the CREATE state gas charges

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_selfdestruct.py
```diff
@@ -709,6 +773,7 @@ def test_selfdestruct_new_beneficiary_account_write_cost(
     execution gas plus the account-creation state gas, and not the
     legacy combined execution account-creation cost.
     """
+    # TODO: Modify to subcall scenario
```
spencer-tb: Shall we add this to the testnet tracker issue?

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
```diff
@@ -285,7 +320,6 @@ def test_block_gas_refund_eip7778_no_block_reduction(
 @pytest.mark.parametrize(
     "num_txs,num_sstores",
     [
-        pytest.param(1, 1, id="single_sstore_single_tx"),
```
spencer-tb, line 288: Should this be dropped?

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
LouisTsai-Csie: Added in new commit

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
LouisTsai-Csie: Added in new commit

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_selfdestruct.py
LouisTsai-Csie: Added in the new commit, let's try to resolve this in this PR.

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py
LouisTsai-Csie: For a single transaction with `(execution_gas, state_gas)`: `total cost = execution_gas + state_gas`, `header gas = max(execution_gas, state_gas)`. If we set block gas limit to `total_cost - 1`, the transaction gas limit becomes higher than the block gas limit, making it invalid, thus dropping this scenario.

spencer-tb (approved):
LGTM! Pushed a commit to remove a stray comment :)

PR #3385, opened 2026-08-17
feat(ci): auto-publish mainnet fixture patch releases from the nightly fill

spencer-tb:
### Description

Automates fortnightly mainnet fixture patch releases from the existing nightly fill while keeping every mainnet-fork (`X`) and consensus-revision (`Y`) release manual.

The release policy is explicit:

| Version component | Meaning | Release path |
| --- | --- | --- |
| `X` | Mainnet fork line | Manual `vX.0.0` release |
| `Y` | Consensus revision for that fork | Manual `vX.Y.0` release |
| `Z` | Non-consensus fixture changes | Automatic patch, at most once every 14 days |

Maintainers select the active `X.Y` series in `.github/configs/mainnet_fixture_version.yaml`. A consensus-changing PR increments `consensus_revision`; a new-mainnet-fork PR increments `fork` and resets the revision to zero. The scheduled workflow never chooses an `X` or `Y` bump: while the configured series is ahead of the latest published release it publishes nothing and points to the exact manual `vX.Y.0` release required.

When the configured and published series match, nightly fixture changes accumulate. The first eligible nightly at least 14 days after the latest published `tests@` release publishes the next `Z` patch from that run's own artifact. An unchanged fixture root produces no release. A manual major or minor release starts a fresh 14-day patch window.

### Content and integrity gates

Every combined fill records two small assets alongside `fixtures.tar.gz`:

- `index_root.txt` provides the cheap content-equality check;
- `fixtures_index.json.gz` supports per-fixture validation and added/removed/changed counts.

The scheduled job publishes only a validated patch. It produces no release when the previous or current index is missing, corrupt or ambiguous; `(json_path, id)` identities are duplicated; a mass disappearance suggests a partial fill; the fork set changes unexpectedly; release-note commit enumeration is incomplete; or a newer manual draft is pending.

Manual mainnet release creation and scheduled patch publication share a concurrency lock, preventing both paths from selecting or creating the same next `tests@` version simultaneously.

Releases predating the standalone index asset bootstrap from the leading metadata members of `fixtures.tar.gz`; the streaming reader allows files such as `fixtures.ini` before `index.json` and stops before downloading the fixture body.

### Manual releases and cached artifacts

Manual `tests` releases are validated against the configured series. A new-series release must be exactly the configured `vX.Y.0`.

Cached releases remain fast, but a cached new-series release may reuse only a nightly whose target commit already contains that exact series marker. This prevents a just-merged consensus or fork release from accidentally attaching fixtures built before the change. If no matching nightly exists, the workflow requires a fresh fill or an explicitly selected newer nightly.

The `A-tests` label never gates publication; it only selects PRs for the generated release notes.

### Verification

- `91 passed` in `.github/scripts/tests/test_release_scripts.py`
- `just static`

### Related Issues or PRs

N/A.

### Checklist

- [x] Ran fast static checks to avoid CI failures.
- [x] PR title follows the repository's conventional title format.

### Cute Animal Picture

Sleeping cat

### DIFF
--- a/.github/configs/mainnet_fixture_version.yaml
+++ b/.github/configs/mainnet_fixture_version.yaml
@@ -1,0 +1,20 @@
+# Mainnet fixture release series (`tests@vX.Y.Z`) selected by
+# maintainers.
+#
+# - `fork` (X): the released mainnet fork line. A new-mainnet-fork PR
+#   increments it and resets `consensus_revision` to zero.
+# - `consensus_revision` (Y): the acknowledged consensus revision of
+#   the released fork. A consensus-changing PR for the released fork
+#   increments it.
+#
+# This file records semantic release intent, the fixture index records
+# artifact content. While this series matches the newest published
+# release, the first eligible nightly at least 14 days after that
+# release publishes accumulated fixture-content changes as the next
+# patch (Z). Moving the series here never authorizes the nightly to
+# create a major or minor release: it pauses the automation until the
+# matching `vX.Y.0` release is dispatched manually. A cached dispatch
+# is valid only when that nightly was built after this series change;
+# otherwise dispatch a fresh fill.
+fork: 20
+consensus_revision: 0
--- a/.github/configs/mainnet_release_notes.md
+++ b/.github/configs/mainnet_release_notes.md
@@ -1,0 +1,11 @@
+### `tests@{version}`
+
+Scheduled mainnet fixture release: a single `fixtures.tar.gz` with all tests for all forks up to the latest mainnet fork. It supersedes [`{previous_tag}`]({previous_tag_url}).
+
+### New tests
+{test_prs}
+
+Fixture diff since [`{previous_tag}`]({previous_tag_url}): {diff_counts}.
+Fixture index root: `{root_hash}`
+
+**Full Changelog**: {compare_url}
--- a/.github/scripts/publish_mainnet_release.py
+++ b/.github/scripts/publish_mainnet_release.py
@@ -1,0 +1,689 @@
+#!/usr/bin/env -S uv run --script
+#
+# /// script
+# requires-python = ">=3.12"
+# dependencies = [
+#     "pyyaml",
+# ]
+# ///
+"""
+Gate and prepare the scheduled mainnet patch release.
+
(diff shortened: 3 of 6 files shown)


danceratopz (commented):
Thanks for this @spencer-tb, would love to get more automation here! Can we discuss the idea below before pushing on further with the PR as-is?

High-level thought before getting into the implementation details: I wonder if we should keep the existing **prepare → promote** model as the only way that mainnet fixture releases are actually published, rather than adding a second path which can publish them fully automatically. I think it's worth keeping a human-in-the-loop in order to ensure high quality release notes.

I think a lot of the automation in this PR is still really useful. What I'd ideally like the nightly workflow to do is:

1. Build the candidate mainnet fixtures as it does today.
2. Compare them against the previous release and generate the factual information we need for a release: added / removed / changed tests, fixture-format changes, etc. Attach this as a separate artifact to the fixture build.
3. Create a **draft release** containing those artifacts and that machine-generated information.
4. Mark the draft as awaiting review/promotion and ping the relevant maintainers.
5. Have a small agent/skill that can take the generated fixture diff plus the relevant PRs and turn it into concise, human-oriented release notes (and discord message text)
6. Have a maintainer quickly review/edit those notes and then use the existing **promotion** mechanism to publish the release.

`tests@v20.0.2` is basically the kind of end result I have in mind: the release notes tell consumers very quickly what was added, removed and changed, while filtering the raw development activity down to what is actually relevant to the fixture release.

I think that curation is worth preserving. For example, while the mainnet release is still pinned to Osaka, Amsterdam test work is landing continuously. A purely commit-driven automated changelog either includes that noise or needs increasingly complicated logic to infer which commits actually affected the released fixture set. The fixture diff gives us a much better ground truth, and an LLM + quick human review can turn that into useful release notes.

If we're releasing regularly, I don't think the human-in-the-loop part should be much work: ideally we're talking about quickly checking an already-prepared draft and hitting promote. In return, we keep high-quality release notes, have an explicit sanity check before publishing something clients consume, and avoid maintaining two different mechanisms for creating mainnet releases.

So my preference would be to keep a lot of the machinery you've added here, but use it to automate everything **up to promotion**, rather than automating promotion itself.

PR #3390, opened 2026-08-18
feat(test-fill): Optimistic grouping flag

marioevz:
### Description

Makes pre-allocation group packing optional, and replaces the untyped string hashes used across the pre-alloc group and fixture index machinery with fixed-size types.

**`--disable-optimistic-pre-alloc-grouping`**

New `fill` flag. Phase 1 skips the packing pass so groups stay fine-grained, and phase 2 then resolves a test's group by recomputing its hash directly instead of looking it up in the packed index. Packing stays the default.

**Typed group hashes**

- New `AllocGroupHash` (8-byte truncated sha256) replaces the `str`/`int` hashes that were assembled by hand in several places.
- `Environment.__hash__` is replaced by `Environment.canonical_json()`, so the grouping pre-image is an explicit canonical JSON encoding rather than a process-stable `__hash__` whose value was persisted as a folder name.
- `PreAllocGroupBuilder` carries its own `group_hash`, and `GroupPreAlloc.get_pre_alloc_grouping_hash()` lets an allocation report the group it belongs to.
- `GroupIndexEntry`/`GroupIndexEntries` are pydantic models; the test → group index is read and written through them instead of hand-rolled JSON.

**Engine X fixtures**

- `pre_hash` is set when the fixture is constructed, from the allocation's own group hash, rather than patched onto the model afterwards.
- `postStateDiff` was never populated: `post_state` is not a field of `BlockchainEngineXFixture`, so the filler's `hasattr(fixture, "post_state")` guard never fired. It is now computed in `BlockchainTest.make_hive_fixture` through the new `Alloc.calculate_diff()` (moved out of `filler.py`) and is a required field.

**Fixture index**

- `TestCaseBase.fixture_hash` and `IndexFile.root_hash` are `Hash` instead of `HexNumber`; `hasher` and `compare_fixtures` follow.
- `FixtureCollector` writes index entries through `TestCaseIndexFile` instead of an ad-hoc dict.
- `HashableItem.from_folder` no longer strips the first two characters of a hash unconditionally, which mis-parsed unprefixed `generatedTestHash` values into 31 bytes.

**Compatibility**

Group hashes change, so existing `pre_alloc` folders must be regenerated. Engine X fixtures and `.meta/index.json` files written by earlier versions no longer validate, since `postStateDiff` and `fixture_hash` are now required.

### Related Issues or PRs
N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/docs/running_tests/test_formats/blockchain_test_engine_x.md
+++ b/docs/running_tests/test_formats/blockchain_test_engine_x.md
@@ -28,10 +28,12 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
 
 ```json
 {
-   "test_count": 88,
-   "pre_account_count": 174,
+   "testCount": 88,
+   "preAccountCount": 174,
    "testIds": ["test1", "test2", ...],
    "network": "Prague",
+   "chainId": "0x01",
+   "groupHash": "0xb664b0d847df2cf7",
    "environment": { ... },
    "pre": { ... }
 }
@@ -39,10 +41,13 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
 
 #### Pre-Allocation Group Fields
 
-- **`test_count`**: Number of tests in this pre-allocation group
-- **`pre_account_count`**: Number of accounts in the pre-allocation group
+- **`testCount`**: Number of tests in this pre-allocation group
+- **`preAccountCount`**: Number of accounts in the pre-allocation group
 - **`testIds`**: Array of test identifiers that belong to this group
 - **`network`**: Fork name (e.g., "Prague", "Cancun")
+- **`chainId`**: Chain id the group's genesis is configured for
+- **`groupHash`**: The group's own hash; matches the file name and the [`preHash`](#-prehash-string) of every test in the group
+- **`groupSalt`**: Optional isolation salt; only present for groups that were explicitly isolated
 - **`environment`**: Complete [`Environment`](./common_types.md#environment) object with execution context
 - **`pre`**: Pre-allocation group [`Alloc`](./common_types.md#alloc-mappingaddressaccount) object containing initial account states
 
@@ -69,10 +74,8 @@ For each [`BlockchainTestEngineXFixture`](#blockchaintestenginexfixture) test ob
 
 4. **Verify Final State**:
    - Compare the final chain head against [`lastblockhash`](#-lastblockhash-hash)
-   - If [`postStateDiff`](#-poststatediff-optionalalloc) is present:
-     - Apply the state differences to the pre-allocation group
-     - Verify the resulting state matches the client's final state
-   - If `post` field were present (not typical), verify it directly
+   - Apply [`postStateDiff`](#-poststatediff-alloc) to the pre-allocation group
+   - Verify the resulting state matches the client's final state
 
 ## Structures
 
@@ -88,7 +91,7 @@ This field is going to be replaced by the value contained in `config.network`.
 
 #### - `preHash`: `string`
 
-Hash identifier referencing a pre-allocation group in the `pre_alloc` folder. This hash uniquely identifies the combination of fork, environment, and pre-allocation state that defines the group.
+Hash identifier referencing a pre-allocation group in the `pre_alloc` folder. This hash uniquely identifies the combination of fork, environment, and pre-allocation state that defines the group. It is `0x`-prefixed, 8 bytes wide, and matches both the group file's name and its `groupHash` field.
 
 #### - `genesisBlockHeader`: [`FixtureHeader`](./blockchain_test.md#fixtureheader)
 
@@ -106,7 +109,7 @@ Optional synchronization payload. When present, this payload is typically used t
 
 Hash of the last valid block after all payloads have been processed, or the genesis block hash if all payloads are invalid.
 
-#### - `postStateDiff`: [`Optional`](./common_types.md#optional)`[`[`Alloc`](./common_types.md#alloc-mappingaddressaccount)`]`
+#### - `postStateDiff`: [`Alloc`](./common_types.md#alloc-mappingaddressaccount)
 
 State differences from the pre-allocation group after test execution. This optimization stores only the accounts that changed, were created, or were deleted during test execution, rather than the complete final state.
 
--- a/docs/running_tests/test_formats/blockchain_test_engine_x.md
+++ b/docs/running_tests/test_formats/blockchain_test_engine_x.md
@@ -34,7 +34,6 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
    "network": "Prague",
    "chainId": "0x01",
    "groupHash": "0xb664b0d847df2cf7",
-   "environment": { ... },
    "pre": { ... }
 }
 ```
@@ -48,7 +47,6 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
 - **`chainId`**: Chain id the group's genesis is configured for
 - **`groupHash`**: The group's own hash; matches the file name and the [`preHash`](#-prehash-string) of every test in the group
 - **`groupSalt`**: Optional isolation salt; only present for groups that were explicitly isolated
-- **`environment`**: Complete [`Environment`](./common_types.md#environment) object with execution context
 - **`pre`**: Pre-allocation group [`Alloc`](./common_types.md#alloc-mappingaddressaccount) object containing initial account states
 
 ## Consumption
@@ -58,12 +56,11 @@ For each [`BlockchainTestEngineXFixture`](#blockchaintestenginexfixture) test ob
 1. **Load Pre-Allocation Group**:
    - Read the appropriate file from the `pre_alloc` folder in the same directory
    - Locate the pre-allocation group using [`preHash`](#-prehash-string)
-   - Extract the `pre` allocation and `environment` from the group
+   - Extract the `pre` allocation from the group
 
 2. **Initialize Client**:
    - Use [`network`](#-network-fork) to configure the execution fork schedule
    - Use the pre-allocation group's `pre` allocation as the starting state
-   - Use the pre-allocation group's `environment` as the execution context
    - Use [`genesisBlockHeader`](#-genesisblockheader-fixtureheader) as the genesis block header
 
 3. **Execute Engine API Sequence**:
--- a/docs/running_tests/test_formats/blockchain_test_engine_x.md
+++ b/docs/running_tests/test_formats/blockchain_test_engine_x.md
@@ -34,6 +34,7 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
    "network": "Prague",
    "chainId": "0x01",
    "groupHash": "0xb664b0d847df2cf7",
+   "genesis": { ... },
    "pre": { ... }
 }
 ```
@@ -47,6 +48,7 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
 - **`chainId`**: Chain id the group's genesis is configured for
 - **`groupHash`**: The group's own hash; matches the file name and the [`preHash`](#-prehash-string) of every test in the group
 - **`groupSalt`**: Optional isolation salt; only present for groups that were explicitly isolated
+- **`genesis`**: Genesis block header ([`FixtureHeader`](./blockchain_test.md#fixtureheader)) shared by every test in the group, derived from the environment the group was keyed on; its state root matches the state root of `pre`
 - **`pre`**: Pre-allocation group [`Alloc`](./common_types.md#alloc-mappingaddressaccount) object containing initial account states
 
 ## Consumption
@@ -56,12 +58,12 @@ For each [`BlockchainTestEngineXFixture`](#blockchaintestenginexfixture) test ob
 1. **Load Pre-Allocation Group**:
    - Read the appropriate file from the `pre_alloc` folder in the same directory
    - Locate the pre-allocation group using [`preHash`](#-prehash-string)
-   - Extract the `pre` allocation from the group
+   - Extract the `pre` allocation and `genesis` header from the group
 
 2. **Initialize Client**:
    - Use [`network`](#-network-fork) to configure the execution fork schedule
    - Use the pre-allocation group's `pre` allocation as the starting state
-   - Use [`genesisBlockHeader`](#-genesisblockheader-fixtureheader) as the genesis block header
+   - Use the pre-allocation group's `genesis` as the genesis block header
 
 3. **Execute Engine API Sequence**:
    - For each [`FixtureEngineNewPayload`](#fixtureenginenewpayload) in [`engineNewPayloads`](#-enginenewpayloads-listfixtureenginenewpayload):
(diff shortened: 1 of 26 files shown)


packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filling_session.py
```diff
@@ -84,15 +84,17 @@ def test_init_use_pre_alloc(self) -> None:
         config = MockConfig(use_pre_alloc_groups=True)
 
         # Mock the file system operations
+        group_hash = AllocGroupHash.from_hash("test_hash")
```
marioevz: For reviewers: Maybe `from_hash` is not the best name for this function. It's supposed to mean "calculate the `AllocGroupHash` from sha56'ing the input, string or bytes, and then truncate".

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -923,10 +924,10 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
     }
     transition_tool_cache_key: ClassVar[str] = ""
 
-    pre_hash: str
+    pre_hash: AllocGroupHash
     """Hash of the pre-allocation group this test belongs to."""
 
-    post_state_diff: Alloc | None = None
```
danceratopz, line 1022: This change will break loading v20.0.0/v20.0.1. It would be nice to be stricter with these fields going forward, although adding an empty `Alloc` has little value.

From our side, this is currently only relevant for the [generic test flow](https://hive.ethpandaops.io/#/group/generic?suites=eels%2Fconsume-enginex%2Ceels%2Fconsume-engine), I can fix that by pinning the EELS branch until the next release. 

For client infra: The easiest solution to avoid breaking their infra would be to build and release v20.0.2 immediately after merging this and telling clients to update.

packages/testing/src/execution_testing/cli/hasher.py
danceratopz, line 1: `hasher` still breaks by default with `engine_x` when ran as:
```
hasher compare <fill-a> <fill-b>
```
due to the presence of `blockchain_tests_engine_x/pre_alloc/*.json`. As we're touching `hasher` here anyway, we could consider excluding `pre_alloc/`. But very happy to do this as a follow-up!

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -632,6 +580,17 @@ def pytest_addoption(parser: pytest.Parser) -> None:
             "groups, phase 2 generates all supported fixture formats."
         ),
     )
+    test_group.addoption(
```
danceratopz, line 582: We could consider adding a pytester test for invoking fill with this to pin the behavior down.

docs/running_tests/test_formats/blockchain_test_engine_x.md
```diff
@@ -28,21 +28,26 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
 
 ```json
 {
-   "test_count": 88,
-   "pre_account_count": 174,
+   "testCount": 88,
+   "preAccountCount": 174,
```
danceratopz, line 32: These never got written to the JSON (pre-existing issue); they're only tracked internally in the in-memory `PreAllocGroup`. We should remove them here (consumers can use `len(testIds)` and `len(pre)`.
```suggestion
```

docs/running_tests/test_formats/blockchain_test_engine_x.md
```diff
@@ -28,21 +28,26 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
...
    "pre": { ... }
 }
 ```
 
 #### Pre-Allocation Group Fields
 
-- **`test_count`**: Number of tests in this pre-allocation group
-- **`pre_account_count`**: Number of accounts in the pre-allocation group
+- **`testCount`**: Number of tests in this pre-allocation group
+- **`preAccountCount`**: Number of accounts in the pre-allocation group
```
danceratopz, line 45: They can be removed here, too.
```suggestion
```

docs/running_tests/test_formats/blockchain_test_engine_x.md
```diff
@@ -28,21 +28,26 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
 
 ```json
 {
-   "test_count": 88,
-   "pre_account_count": 174,
+   "testCount": 88,
+   "preAccountCount": 174,
    "testIds": ["test1", "test2", ...],
    "network": "Prague",
+   "chainId": "0x01",
```
danceratopz, line 35: It actually writes `1` (integer). This is inconsistent with other fixture formats 😬. Let me see where we can fix it.

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
danceratopz: I think we need this to bring it inline with other fixture formats.
```suggestion
chain_id: ZeroPaddedHexNumber = ZeroPaddedHexNumber(DEFAULT_CHAIN_ID)
```

This will also break compatibility... of the forks/amsterdam simulator with the v20.0.1 fixture release.

packages/testing/src/execution_testing/cli/hasher.py
danceratopz: The `0x` prefix is no longer necessary,
```suggestion
            lines.append(f"{' ' * level}{print_name}: {self.hash().hex()}")
```

danceratopz:
```suggestion
        return [f"{item.hash().hex()}"]
```

danceratopz:
```suggestion
            result[path] = f"{item.hash().hex()}"
```

danceratopz:
```suggestion
left_hashes = {"root": f"{left_item.hash().hex()}"}
right_hashes = {"root": f"{right_item.hash().hex()}"}
```

danceratopz (commented):
Thanks a lot for cleaning up the types here Mario 🙏 

This looks good to me, just a few comments below. But I think we should align on a release strategy before merge.

If you're busy, I can PR the changes below; as you like.

packages/testing/src/execution_testing/fixtures/blockchain.py
marioevz: I think we can do `Alloc | None = None` and then eventually remove the `None` once we've made enough releases with this field populated. Thanks for spotting this, I'll rollback the change!

packages/testing/src/execution_testing/cli/hasher.py
marioevz: Very nice idea, I'll give it a go while time boxing it 👍

packages/testing/src/execution_testing/fixtures/blockchain.py
marioevz: After some checking I've discovered that the pre-alloc groups in `v20.0.0`/`v20.0.1` are broken anyway due to a bug introduced in #2140: They contain the "builder" variant and therefore do not include the `genesis` header in them, and the only way to ingest them is to load them using the builder's model and then build them internally, which is fine for consume, but if anyone wants to use them out-of-the-box it's very inconvenient.

It's possible to include a ton of workarounds in this PR to try to load the broken variants but I feel we are just digging ourselves in a deeper tech debt hole. I suggest we merge this PR with the strict models and then make the v20.1.0 release, stating that there are breaking changes in the format.

cc @danceratopz

docs/running_tests/test_formats/blockchain_test_engine_x.md
marioevz: It's a bug, see https://github.com/ethereum/execution-specs/pull/3390#discussion_r3834508617. Along with the `genesis` header, they are missing from v20.0.0/1.

docs/running_tests/test_formats/blockchain_test_engine_x.md
marioevz: See previous comment.

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
marioevz: Nice, it's now applied. Thanks!

packages/testing/src/execution_testing/cli/hasher.py
marioevz: Fixed! 👍

marioevz: Fixed too.

marioevz: Fixed too.

marioevz:
@danceratopz this is ready for another review. There's one more breaking change introduced in the last commit: `env` is gone from the final pre-alloc-group model. Reason being that it's not necessary since consumers only need the genesis, and the environment just increases the size.

I think we should prepare for a new release once this is merged.

packages/testing/src/execution_testing/fixtures/blockchain.py
danceratopz: Thanks Mario, also favor a clear cut here, sgtm!

packages/testing/src/execution_testing/fixtures/blockchain.py
```diff
@@ -1016,13 +924,15 @@ class BlockchainEngineXFixture(BlockchainEngineFixtureCommon):
...
-    pre_hash: str
+    pre_hash: AllocGroupHash
     """Hash of the pre-allocation group this test belongs to."""
 
     post_state_diff: Alloc | None = None
     """
     State difference from genesis after test execution (efficiency
     optimization).
+
+    TODO: Remove `| None = None` defaults once we are past `v20.0.1` release.
```
danceratopz: Didn't we agree to be strict here (and break behavior), in this PR, and then simply release immediately (to appease clients using these fixtures)?
```suggestion
    post_state_diff: Alloc
    """
    State difference from genesis after test execution (efficiency
   optimization).
```

docs/running_tests/test_formats/blockchain_test_engine_x.md
```diff
@@ -28,22 +28,25 @@ Each file in the `pre_alloc` folder corresponds to a pre-allocation group identi
...
    "testIds": ["test1", "test2", ...],
    "network": "Prague",
-   "environment": { ... },
+   "chainId": "0x01",
+   "groupHash": "0xb664b0d847df2cf7",
    "pre": { ... }
 }
 ```
 
 #### Pre-Allocation Group Fields
```
danceratopz, line 42: I think we need to add the `genesis` field that gets added (back)?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/tests/test_filling_session.py
danceratopz: Re: https://github.com/ethereum/execution-specs/pull/3390#discussion_r3805616936

How about `AllocGroupHash.from_preimage()`?

packages/testing/src/execution_testing/cli/hasher.py
```diff
@@ -92,35 +96,42 @@ def from_json_file(
...
+        # Special-case the pre-alloc groups
+        try:
+            v = PreAllocGroup.from_file(file_path)
+            return cls(
+                type=HashableItemType.FILE,
+                root=v.hash(),
+                parents=parents + [file_path.name],
+            )
+        except ValidationError:
+            pass
```
danceratopz: This parses every JSON file in the tree as a `PreAllocGroup` before falling back to the fixture-collection path. Three small issues:

1. Group files are read twice (`read_text()` here, then `read_bytes()` in `from_file()`).
2. Every regular fixture file pays a full pydantic parse and reject before it gets hashed, which could be noticeable on large fixture directories.
3. A malformed group file fails validation and falls through to the fixture path, so it dies with the confusing `Expected '_info' in ...` error instead of a group validation error.

Since group files only ever live in a `pre_alloc` folder, we could gate on the location instead and let a broken group file raise its own validation error:

```suggestion
        # Pre-alloc group files live under a "pre_alloc" folder
        if file_path.parent.name == "pre_alloc":
            return cls(
                type=HashableItemType.FILE,
                root=PreAllocGroup.from_file(file_path).hash(),
                parents=parents + [file_path.name],
            )
        file_text = file_path.read_text()
```

The now unused `ValidationError` import can then go too. Could we also add a small `hasher compare` test that includes a `pre_alloc` folder to pin this behavior?

packages/testing/src/execution_testing/cli/pytest_commands/plugins/filler/filler.py
```diff
@@ -962,10 +920,8 @@ def pytest_terminal_summary(
                 # Count accounts by loading as builder (no genesis computation)
```
danceratopz: This comment is now stale, the group files loaded here are in the final format:
```suggestion
                # Count accounts from the final pre-alloc group files
```

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
```diff
@@ -128,13 +149,18 @@ def _get_worker_id() -> Optional[str]:
...
     Merge all partial group files into final group files.
 
     Called by master process after all workers have finished Phase 1.
     Each worker writes {group_hash}.partial.{worker_id}.json files,
     which are merged here into {group_hash}.json files.
+
+    The `final` parameter establishes whether to save the files in builder
+    format, in order for them to be able to be re-processed by
+    `pack_pre_alloc_groups`, or the pre-alloc format to be included
+    in the output.
```
danceratopz: Nit: `to_partial_file`'s docstring above is now stale in the same way, it still says the state root is computed when loading in Phase 2 via `PreAllocGroup.from_file()`, but it's now computed here (or in `pack_pre_alloc_groups`) on the master at the end of Phase 1.

packages/testing/src/execution_testing/fixtures/pre_alloc_groups.py
```diff
@@ -27,30 +26,52 @@
...
             "Explicit isolation salt from the `pre_alloc_group` marker; "
             "groups only pack with groups carrying the same salt."
         ),
     )
+    group_hash: AllocGroupHash | None = None
+
+    @classmethod
+    def from_file(cls, file: Path) -> Self:
+        """Load a pre-allocation group or builder from a JSON file."""
+        return cls.model_validate_json(file.read_bytes())
```
danceratopz: The old `PreAllocGroup.from_file` raised if the file stem didn't match the embedded `groupHash`. Consume resolves the file by `preHash` and then trusts the embedded hash, so a mismatch would now propagate silently. Can we keep that check, either here or in `PreAllocGroups.from_folder`?

danceratopz (approved):
Couple of comments, mainly nits, I think the most relevant change to get in this PR is to stop allowing `post_state_diff: Alloc | None = None` in the scope of this PR, or did I misunderstand something? See https://github.com/ethereum/execution-specs/pull/3390/changes#r3864427694.

Thanks @marioevz!

PR #3396, opened 2026-08-19
feat(specs,tests): update EIP-8141 implementation

gurukamath:
### Description
The following updates to the spec have been implemented
1. 2-D gas for frames
2. `sigparam` dynamic stack removal
3. `payer` warming

Brings the spec up to date with `0603514569547869ae2531a4e5b2b62875888db3` version of the EIP.

### Related Issues or PRs
#2829

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Cute Animals - 1 of 1

PR #3398, opened 2026-08-19
feat(tests): cover the signature-validation precompiles being absent from the block access list

ilitteri:
### Description

Covers an uncovered rule from EIP commit [`eec839abb`](https://github.com/ethereum/EIPs/commit/eec839abb): "since the signature validation does not happen in EVM execution, the related precompiles `ecrecover` and `P256VERIFY` must not be added to the block-level access list."

Nothing in state distinguishes the two behaviours — a precompile has no storage, balance or nonce to change. But the block access list is committed to in the header, so an implementation that validates signatures by dispatching through its own EVM adds the precompile as a touched address, produces a different block access list hash, and has its block rejected while every other result matches. That asymmetry is what makes it worth a fixture rather than a comment.

The transaction carries one signature of each protocol-validated scheme so both precompiles are exercised, and asserts each is absent. Flipping either assertion to expect the address present fails with `Expected address 0x…01 not found in actual BAL`, so the absence is really being checked and not silently vacuous.

Based on the suite tip rather than #3396, so it is independent of the two-dimensional gas work and can land either side of it.

### Related Issues or PRs

Follows #3393.

### Checklist

### DIFF
--- a/tests/amsterdam/eip8141_frame_transactions/test_block_access_lists.py
+++ b/tests/amsterdam/eip8141_frame_transactions/test_block_access_lists.py
@@ -33,6 +33,9 @@ from execution_testing import (
     keccak256,
 )
 
+from tests.frontier.precompiles.spec import Spec as EcrecoverSpec
+from tests.osaka.eip7951_p256verify_precompiles.spec import Spec as Spec7951
+
 from .helpers import sender_frame, verify_frame
 from .signature_helpers import P256_SIGNATURE, p256_entry
 from .spec import Spec, ref_spec_8141
@@ -42,10 +45,10 @@ REFERENCE_SPEC_VERSION = ref_spec_8141.version
 
 pytestmark = pytest.mark.valid_from("Bogota")
 
-ECRECOVER_ADDRESS = Address(0x01)
+ECRECOVER_ADDRESS = EcrecoverSpec.ECRECOVER
 """The `ecrecover` precompile, which validates SECP256K1 signature entries."""
 
-P256VERIFY_ADDRESS = Address(0x100)
+P256VERIFY_ADDRESS = Address(Spec7951.P256VERIFY)
 """The `P256VERIFY` precompile (EIP-7951), which validates P256 entries."""
 
 SLOT = 0x01
@@ -480,7 +483,7 @@ def test_bal_omits_signature_validation_precompiles(
         sender=sender,
         frames=[
             verify_frame(flags=Spec.APPROVE_EXECUTION_AND_PAYMENT),
-            sender_frame(target=target, gas_limit=WRITE_FRAME_GAS),
+            sender_frame(target=target),
         ],
         signatures=[
             # Index 0 authorizes the default code, and is validated with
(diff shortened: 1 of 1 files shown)


tests/amsterdam/eip8141_frame_transactions/test_block_access_lists.py
```diff
@@ -29,16 +30,24 @@
...
 REFERENCE_SPEC_VERSION = ref_spec_8141.version
 
 pytestmark = pytest.mark.valid_from("Bogota")
 
+ECRECOVER_ADDRESS = Address(0x01)
+"""The `ecrecover` precompile, which validates SECP256K1 signature entries."""
+
+P256VERIFY_ADDRESS = Address(0x100)
+"""The `P256VERIFY` precompile (EIP-7951), which validates P256 entries."""
+
```
gurukamath:
```suggestion
from tests.frontier.precompiles.spec import Spec as EcrecoverSpec
from tests.osaka.eip7951_p256verify_precompiles.spec import Spec as Spec7951

ECRECOVER_ADDRESS = EcrecoverSpec.ECRECOVER
P256VERIFY_ADDRESS = Address(Spec7951.P256VERIFY)
```

These addresses are already defined elsewhere

tests/amsterdam/eip8141_frame_transactions/test_block_access_lists.py
```diff
@@ -438,3 +447,75 @@ def test_bal_sponsored_payer_and_sender(
...
+    so both precompiles are exercised, and asserts each is absent.
+    """
+    sender = pre.fund_eoa()
+    target = pre.deploy_contract(code=Op.STOP)
+
+    tx = Transaction(
+        sender=sender,
+        frames=[
+            verify_frame(flags=Spec.APPROVE_EXECUTION_AND_PAYMENT),
+            sender_frame(target=target, gas_limit=WRITE_FRAME_GAS),
```
gurukamath:
```suggestion
            sender_frame(target=target),
```
Nothing writes here (the target is `Op.STOP`), and `WRITE_FRAME_GAS` is documented as what a writing frame needs. `helpers.py` says non-gas-sensitive tests should take the default

tests/amsterdam/eip8141_frame_transactions/test_block_access_lists.py
ilitteri: Thanks! Addressed in [0187552](https://github.com/ethereum/execution-specs/pull/3398/commits/018755291e27ac6cd6fefd87a95b22595d223549).

tests/amsterdam/eip8141_frame_transactions/test_block_access_lists.py
ilitteri: Thanks! Addressed in [0187552](https://github.com/ethereum/execution-specs/pull/3398/commits/018755291e27ac6cd6fefd87a95b22595d223549).

gurukamath (approved):

PR #3047, opened 2026-06-24
Frame Transactions (EIP-8141)

SamWilsn:
## 🗒️ Description

My work so far on frame transactions (EIP-8141)

#### Cute Animal Picture

IMG_20260624_154128_797

PR #3407, opened 2026-08-20
feat(spec-specs,tests): dispatch a precompile a VERIFY frame targets

Marchhill:
### Description

[EIP-8141][eip] now dispatches a precompile frame target unconditionally:

> - If `resolved_target` is an active precompile at the current fork, the frame dispatches it as an ordinary call would.
> - Otherwise, if `resolved_target` code hash is empty, i.e. `0xc5d2...a470`, execute the logic described in default code.

There is no mode condition, so a `VERIFY` frame dispatches a precompile too. A `VERIFY` frame already runs as a static call and only a `SENDER` frame may carry non-zero value, so a precompile there is pure computation with no state effect — which is why this was preferred over banning a precompile target in `VERIFY`.

`execute_frame` decided the default code from the resolved target's code hash alone, and a precompile account carries the empty code hash that decision keys on. A `VERIFY` frame targeting a precompile therefore ran the default code and never dispatched it. The default code is now left to a `VERIFY` frame whose codeless target is no precompile.

On the test side, `VERIFY` folds into the mode parametrization of `test_precompile_target`, and `test_verify_frame_precompile_target` — which encoded the old rejection — is dropped as superseded. The first frame approves both execution and payment, so routing the frame under test to the default code instead reverts it and rejects the transaction: with the reference change reverted, exactly the nine new `verify_mode` cases fail with `VERIFY frame reverted`, and nothing else moves.

`test_verify_frame_delegated_to_precompile_target` is unaffected: a designated target is not codeless, so it never reached the default-code branch.

### Pinning the edges of the precompile set

The new condition consults an exact set, and the spec's "active precompile at the current fork" is not a contiguous address range: `P256VERIFY` sits at `0x100`, above a gap that starts at `0x12`. Two cases pin those edges, one on each side of the gap, and each is load-bearing on its own — replacing the set membership with an approximating address range fails that case and only that case:

| Reference change | Fails | Rest of the suite |
| --- | --- | --- |
| Range `0x01..=0x11`, stopping below the gap | `test_verify_frame_target_above_the_precompile_gap` (3) | 568 pass |
| Range `0x01..=0x100`, spanning the gap | `test_verify_frame_target_inside_the_precompile_gap` (3) | 568 pass |

Neither was caught by the suite before these cases were added. The positive and negative directions of the condition were already pinned: dropping it fails the nine `verify_mode` cases plus the `P256VERIFY` case, and inverting it fails 285 cases across the directory, since every leading `verify_frame()` resolves to the codeless sender and depends on the default code to approve.

These fill for `Bogota` only, so the precompile set does not vary across the forks under test.

### Numbers

`eip8141_frame_transactions` fills 571 cases, all passing (559 before this PR; 565 before the two boundary cases).

`just` was not available locally, so every `just static` recipe was run individually and is clean: `mypy`, `ethereum-spec-lint`, `codespell`, `vulture`, `actionlint`, `uv lock --check`, `ruff format --check`, `ruff check`.

[eip]: https://github.com/ethereum/EIPs/pull/12157

### Related Issues or PRs

Follows #3047, which implemented EIP-8141 on this branch.

### Checklist

### Cute Animal Picture

A red panda

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/frame_interpreter.py
+++ b/src/ethereum/forks/amsterdam/vm/frame_interpreter.py
@@ -392,8 +392,9 @@ def execute_frame(
     at frame entry, from its own execution gas budget, before
     anything else: resolving the target's code is how the protocol
     dispatches the frame. A `VERIFY` frame whose resolved target has
-    no code then runs the protocol default code instead of an EVM. As
-    with an ordinary `CALL`, a caller that cannot cover the
+    no code then runs the protocol default code instead of an EVM,
+    unless that target is a precompile, which dispatches in every
+    mode. As with an ordinary `CALL`, a caller that cannot cover the
     transferred value reverts the frame, consuming the gas charged so
     far.
 
@@ -458,6 +459,7 @@ def execute_frame(
     target_account = get_account(tx_state, resolved_target)
     if (
         frame.mode == FrameMode.VERIFY
+        and resolved_target not in PRE_COMPILED_CONTRACTS
         and target_account.code_hash == EMPTY_CODE_HASH
     ):
         try:
--- a/tests/amsterdam/eip8141_frame_transactions/spec.py
+++ b/tests/amsterdam/eip8141_frame_transactions/spec.py
@@ -14,7 +14,7 @@ class ReferenceSpec:
 
 
 ref_spec_8141 = ReferenceSpec(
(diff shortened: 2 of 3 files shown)


spencer-tb:
Rebased onto the merged #3412 and pushed, hope you don't mind!

Your change is untouched! The `P256VERIFY` boundary test's first frame now reports the entry access #3412 introduced, and the `ReferenceSpec` is bumped to the current
tip `3ceef8d3` :)

spencer-tb (approved):

Issue #2829, opened 2026-05-09
EIP-8141 Implementation Tracker: Frame Transaction


spencer-tb:
### Target Fork

**Bogota**

### Instructions

- [x] Assign issue to EIP specification and testing owner(s).

> [!IMPORTANT]
> A specifications specialist and a testing specialist should ideally share ownership of the EIP.

- [x] Add the issue to the target fork milestone if applicable (i.e., the EIP is at least in the [CFI stage](https://eips.ethereum.org/EIPS/eip-7723#considered-for-inclusion)).

#### Guidance for Marking Items Complete

An item should only be checked off once the EIP is considered *stable*. In this context, stable means:

- No major issues or ambiguities are still being uncovered in the specification or tests.
- There are no open discussion points awaiting resolution.
- Client implementations have been consistently passing the tests for at least a week.

It is ultimately up to the owners' discretion to decide when an item should be marked as complete, using this guidance as the basis for that decision.

In exceptional cases, an EIP may require changes after some items have been marked complete or even after the entire issue has been completed and closed. This can happen, for example, when significant design optimizations are identified and agreed upon in ACD, or when critical security issues surface and require updates to the specification or tests.

When this occurs, owners should either unmark the relevant checkboxes if the issue is still open, or create a new tracking issue for the modifications if the original issue had already been closed.

### Specification Changes Tracker

- [x] ethereum/EIPs#11482
- [x] ethereum/EIPs#11555
- [x] ethereum/EIPs#11580
- [x] ethereum/EIPs#11681
- [x] ethereum/EIPs#12026 - implemented in #3047 ✅ 
- [ ] ethereum/EIPs#12041
- [x] ethereum/EIPs#12042 - implemented in #3047 ✅ 
- [x] ethereum/EIPs#12061 - informational for execution-specs (json-rpc related) ✅ 
- [x] ethereum/EIPs#12062 - implemented in #3396 ✅ 
- [x] ethereum/EIPs#12066 - informational for execution-specs (mempool related) ✅ 
- [x] ethereum/EIPs#12091 - implemented in #3047 ✅ 
- [x] ethereum/EIPs#12109 - implemented in #3393 ✅ 
- [x] ethereum/EIPs#12113 - implemented in #3396 ✅ 
- [x] ethereum/EIPs#12121 - informational ✅ 
- [x] ethereum/EIPs#12155
- [x] ethereum/EIPs#12157 - implemented in #3407 
- [ ] ethereum/EIPs#12160
[...]

PR #3412, opened 2026-08-21
feat(specs,tests): charge a frame target's access at frame entry

gurukamath:
### Description

- EIP-8141 revision `b3e8cad3` ([EIPs #12211](https://github.com/ethereum/EIPs/pull/12211)) moves the resolved target's warm or cold access to frame entry — before the balance check and before dispatch — so `execute_frame` takes the charge, every default-code frame receipt reports it instead of zero, and the payer's warming rule from #12113 falls out as redundant; 

- Revision `b6b6f1ce` ([EIPs #12212](https://github.com/ethereum/EIPs/pull/12212)) rescopes every `APPROVE` failure from the frame to the current call frame, which the implementation already did — `process_call` catches `Revert` and `ExceptionalHalt` inside the call frame and restores the frame context — so only three docstrings changed, one test now pins the distinction, and the `ReferenceSpec` moves to `b6b6f1ce` rather than the current tip, whose #12157 precompile arm belongs to #3407.

### Related Issues or PRs
#2829

### Checklist

### DIFF
--- a/packages/testing/src/execution_testing/forks/forks/eips/bogota/eip_8141.py
+++ b/packages/testing/src/execution_testing/forks/forks/eips/bogota/eip_8141.py
@@ -71,9 +71,10 @@ class EIP8141(BaseFork):
         Add the frame instruction gas costs.
 
         `SIGDATACOPY` follows the `CALLDATACOPY` shape: the opcode
-        instance must carry copy-size and memory-size metadata. The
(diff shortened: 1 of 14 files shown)


spencer-tb (approved):
LGTM! Pushed `7dffd932` on top to pin exact gas in the refused-approval test and cover the entry-charge halt inside an atomic batch.

PR #3413, opened 2026-08-21
feat(tests): map ethrex's frame-transaction rejection reasons

ilitteri:
`EthrexExceptionMapper` has no entries for the frame-transaction exceptions, so every type-`0x06` rejection reaches `consume-engine` as "Undefined exception message" even when the block was correctly rejected. That is 51 of the 62 failures in the `frames` hive suite.

Adds the three `TYPE_6_*` exceptions plus `GASPRICE_OVERFLOW` and `PRIORITY_OVERFLOW`. `TYPE_6_INVALID_FRAME_FORMAT` is a regex because ethrex reaches it three ways: the static-validation reason, signature-entry structure checked during signature validation, and a field too wide for its type rejected while decoding.

The client side is already on `lambdaclass/ethrex@frames-devnet-0`, the branch hive builds, so this carries no ordering constraint.


spencer-tb (approved):
LGTM! Thanks for defining these :)

PR #3416, opened 2026-08-22
chore(tests): add genesis block header field BAL checklist

shubham5080:
## Summary
- Add `test_bal_genesis` for EIP-7928, covering the remaining checklist item `block_header_field/test/genesis`
- Genesis header must carry the empty BAL hash; block 1 verifies transfer activity is recorded in the BAL body
- Update `eip_checklist_external_coverage.txt` for `general/code_coverage/missed_lines`

Follows the same pattern as `test_slotnum_genesis` in EIP-7843.

## Test plan
- [x] `uv run fill tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py::test_bal_genesis --until=Amsterdam --clean`

### Checklist
- [x] Ran fast static checks to avoid CI fails: `just static`

### Cute Animal Picture
Glamsterdam polar bear 🐻‍❄️

polar bear

### DIFF
--- a/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
+++ b/tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
@@ -48,6 +48,7 @@ SYSTEM_ADDRESS = Address(0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE)
 
 
 @EIPChecklist.BlockHeaderField.Test.ValueBehavior.Accept()
+@EIPChecklist.BlockHeaderField.Test.Genesis()
 def test_bal_nonce_changes(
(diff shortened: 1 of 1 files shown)


spencer-tb (dismissed):
The `BlockHeaderField.Test.Genesis` item is indeed unmarked for 7928, although a dedicated test doesn't get us extra coverage here.

The genesis BAL is always empty, so every `valid_from("Amsterdam")` fixture already carries the only valid value. Could you instead, drop `test_bal_genesis` and add `@EIPChecklist.BlockHeaderField.Test.Genesis()` to `test_bal_nonce_changes`. And then also revert the `missed_lines` change in `eip_checklist_external_coverage.txt`?

It should just be a one line change!

spencer-tb (approved):

PR #3417, opened 2026-08-22
chore(tests): remove EIP-7610 create collision tests

spencer-tb:
### Description

Remove the EIP-7610 storage collision tests so their fixtures drop out of the next release. EIP-7610 is declined for inclusion in Glamsterdam ([EIP-7773](https://eips.ethereum.org/EIPS/eip-7773)): no mainnet account has non-empty storage with empty code and nonce, and clients are dropping the storage collision check.

Also removes the two ported static tests that depend on the check, `test_revert_in_create_in_init_paris.py` and `test_failed_create_reverts_deletion_paris.py`. These are the CREATE variants of the same scenario and only exist as their Paris (EIP-7610) versions.

Coverage that does not depend on EIP-7610 is kept. The EIP-684 code and nonce collision cases move to `tests/frontier/create/test_create_collision.py`, and `test_collision_selfdestruct.py` moves to the EIP-6780 directory since its collision target has code and nonce. Also rewords two stale EIP-7610 mentions in the EIP-7708 and EIP-8038 tests, which cover EIP-684 collisions.

### Related Issues or PRs

Related to #2819 and #2957.

### Checklist


LouisTsai-Csie (approved):
Is there anything we need to remove from the spec side?

tests/frontier/create/test_create_collision.py
```diff
@@ -18,25 +20,20 @@
...
     ),
     pytest.mark.parametrize(
-        "collision_nonce,collision_balance,collision_code",
+        "collision_nonce,collision_code,collision_storage",
         [
-            pytest.param(0, 0, b"\0", id="non-empty-code"),
-            pytest.param(0, 1, b"", id="non-empty-balance"),
-            pytest.param(1, 0, b"", id="non-empty-nonce"),
+            pytest.param(0, b"\0", {0x01: 0x01}, id="non-empty-code"),
+            pytest.param(1, b"", {}, id="non-empty-nonce"),
```
jochem-brouwer, line 36: Why is the non-empty balance removed here? I don't see a reason to remove it. Note that EIP-7610 is non empty *storage* (with nonce=0, any balance (0 also ok) and no code)

tests/frontier/create/test_create_collision.py
```diff
@@ -106,7 +103,9 @@ def test_init_collision_create_tx(
         pre=pre,
         post={
             created_contract_address: Account(
-                storage={0x01: 0x01},
+                nonce=collision_nonce,
+                code=collision_code,
+                storage=collision_storage,
```
jochem-brouwer, line 108: This test could also pass if the client would (incorrectly) check if the storage is non-empty. The storage should be parametrized. For the case which hits the EIP-7610 cases: these should be skipped as the behavior is undefined in protocol. This is thus the test with non-empty storage, nonce=0, code=empty, balance=any (parametrize with 0,1)
The parameters of this test thus have to be expanded to cover all cases.

jochem-brouwer (dismissed):
Some small comments regarding a test where I think we should directly expand coverage.

tests/frontier/create/test_create_collision.py
danceratopz: @jochem-brouwer I was reviewing this when you're comment came in, could you review these? I added this expansion as a follow-up PR to avoid blocking this clean-up (check last commit): 
- https://github.com/ethereum/execution-specs/pull/3425

danceratopz (commented):
Thanks @spencer-tb, LGTM, I'd just suggest updating the BAL `test_cases.md` in the scope of this PR. Claude suggested two other state-reference corrections. I added them up front here: 
- https://github.com/spencer-tb/execution-specs/pull/54

I was reviewing when @jochem-brouwer's comment came in. I think this could be follow-up to get this in quickly, although it would make sense to get them in the next release:
- https://github.com/ethereum/execution-specs/pull/3425/changes

spencer-tb:
@jochem-brouwer can you revert the requested changes, so we merge this and then switch to @danceratopz's follow up in #3425?

spencer-tb:
> Is there anything we need to remove from the spec side?

The answer is yes! @gurukamath could you review https://github.com/ethereum/execution-specs/pull/3417/commits/d7baf6bec8891bbf4ef311b8d6f472789d1a29fe?

PR #3422, opened 2026-08-23
feat(tests): EIP-8037 burn subcall reversal-refund credit on top-level halt

flcl42:
## 🗒️ Description

Adds one parametrized state-test filler to the EIP-8037 suite covering the exceptional-halt shape from NethermindEth/nethermind#12964 that the suite did not cover today (`test_top_level_failure_after_child_reversal_refund`, `halt`/`revert`).

Scenario: the parent SSTOREs zero-to-nonzero with a reservoir covering only half the charge (spill), then CALLs a child which restores the slot to its original zero. The restoration credits a full SSTORE state charge to a frame whose ledger recorded no state usage of its own; the child returns successfully, so the credit merges into the parent above the start reservoir. Then:

- **halt**: everything above the start reservoir burns, including the merged credit — the sender pays `gas_limit_cap`. This is the rule Nethermind got wrong (it refunded the credit, undercharging by exactly one SSTORE_SET, per NethermindEth/nethermind#12964 and fix PR NethermindEth/nethermind#12965).
- **revert**: all state gas is refunded LIFO and the sender pays only regular gas plus intrinsic — guards against over-correcting the halt path.

The filled fixtures are deterministic two-frame reductions of the 21-byte recursive fixture in nethermind#12964 (which relies on gas-exhaustion choreography), so they isolate the settlement rule itself.

Verification: all six variants (halt/revert x state_test/blockchain_test/blockchain_test_engine) pass when filled against the reference spec implementation on Amsterdam.

Note: generated fixture JSONs are produced by CI filling and are not committed here, per repo convention.

## 🔗 Related Issues or PRs

Related: NethermindEth/nethermind#12964 (client divergence report) and NethermindEth/nethermind#12965 (client fix). N/A for ethereum/execution-specs issues.

## ✅ Checklist

- [x] All: Ran fast static checks on the changed test file (`ruff format --check` / `ruff check` with the pinned ruff version); full `just static` not run locally.
- [x] All: PR title adheres to the repo standard - starts `type(scope):`.
- [ ] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory. (No docs change needed - new case within an existing test module.)
- [x] Tests: For PRs implementing a missed test case, update the [post-mortem document](/ethereum/execution-specs/blob/HEAD/docs/writing_tests/post_mortems.md) to add an entry the list. (Not updated yet - happy to add an entry pointing at nethermind#12964 if reviewers want it tracked there.)

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py
@@ -1034,6 +1034,102 @@ def test_top_level_failure_spilled_state_gas(
     state_test(pre=pre, post=post, tx=tx)
 
 
+@pytest.mark.parametrize(
+    "failure_mode",
+    [
+        pytest.param("revert", id="revert"),
+        pytest.param("halt", id="halt"),
+    ],
+)
+@pytest.mark.valid_from("EIP8037")
+def test_top_level_failure_after_child_reversal_refund(
+    state_test: StateTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    failure_mode: str,
+) -> None:
+    """
+    Verify a state-gas credit earned in a subcall survives the
+    successful call merge but not a top-level exceptional halt.
+
+    The parent SSTOREs zero-to-nonzero with a reservoir covering only
+    half the charge, so the set spills into `gas_left`. The child then
+    restores the slot to its original zero. The restoration credits a
+    full SSTORE state charge to a frame whose ledger recorded no state
+    usage of its own, and the child returns successfully, merging the
+    credit into the parent above the start reservoir.
+
+    - Halt burns everything above the start reservoir (the credit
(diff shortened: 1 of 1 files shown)


LouisTsai-Csie:
Hello! should we update the base branch from `ethereum:eips/amsterdam/eip-8037-source-based-refunds` to `forks/amsterdam`?

flcl42:
Closing in favor of #3423, which covers the same settlement rule (spilled state gas stays consumed through halting frames) from the original reporter's fixture shape, parametrized across gas limits - the better canonical form.

For the record, this PR was a deterministic two-frame reduction (parent spills a fresh set -> child restores the slot -> successful return merges the credit -> parent halts) plus a `REVERT` control asserting the sender pays only regular gas when the failure is explicit rather than exceptional. The halt axis here is subsumed by #3423; if reviewers want the revert-side over-correction guard, it would slot naturally as an additional `failure_mode` parametrization there.

PR #3423, opened 2026-08-24
feat(tests): EIP-8037 - consume spilled state gas along a halt chain

chfast:
### Description

State gas that spills into a frame's execution gas must be consumed when that frame halts exceptionally. The suite guards this at single spill points, but never along a chain of halting frames that recycles one reservoir.

A self-DELEGATECALLing contract writes NOT(storage[slot]), so set and clear alternate down the shared-storage call stack. Each frame reuses the value it wrote as a CALL's args_size, so the frames that set the slot halt on the memory-size overflow while the frames that cleared it return normally. The top-level halt then charges the whole gas limit, asserted over three gas limits below the transaction gas limit cap so the reservoir starts empty and every set spills.

### Related Issues or PRs

https://github.com/NethermindEth/nethermind/issues/12964
Found by evmone-fuzz.

### Checklist

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_reservoir.py
@@ -1034,6 +1034,66 @@ def test_top_level_failure_spilled_state_gas(
     state_test(pre=pre, post=post, tx=tx)
 
 
+@pytest.mark.parametrize(
+    "gas_limit",
+    [
+        pytest.param(150_000, id="gas_150k"),
+        pytest.param(200_000, id="gas_200k"),
+        pytest.param(300_000, id="gas_300k"),
+    ],
+)
+@pytest.mark.valid_from("EIP8037")
+def test_spilled_state_gas_consumed_across_halt_chain(
+    state_test: StateTestFiller,
+    pre: Alloc,
+    fork: Fork,
+    gas_limit: int,
+) -> None:
+    """
+    Verify spilled state gas stays consumed along a chain of halting frames.
+
+    A self-`DELEGATECALL`ing contract writes `NOT(storage[slot])`, so set and
+    clear alternate down the shared-storage call stack and the reservoir is
+    recycled instead of drained once. Each frame reuses the value it wrote as
+    a `CALL`'s `args_size`, so the frames that set the slot halt on the
+    memory-size overflow while the frames that cleared it re-enter the
+    contract, interleaving halting and surviving frames. Every halt must burn
+    its spill rather than credit it back to the caller's reservoir, so the
+    top-level halt charges the whole gas limit.
+    """
+    gas_limit_cap = fork.transaction_gas_limit_cap()
+    assert gas_limit_cap is not None
+    # Below the cap, so the reservoir starts empty and every set spills.
+    assert gas_limit < gas_limit_cap
+
+    slot = 0
+    value_offset = 0
+    contract = pre.deploy_contract(
+        code=(
+            # Memory is per-frame, so each frame keeps the value it wrote
+            # and reuses it below as the CALL's args_size.
+            Op.MSTORE(value_offset, Op.NOT(Op.SLOAD(slot)))
+            + Op.SSTORE(slot, Op.MLOAD(value_offset))
+            + Op.POP(Op.DELEGATECALL(address=Op.ADDRESS))
+            # An all-ones args_size overflows the memory-size calculation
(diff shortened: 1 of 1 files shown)


flcl42:
Nice one - this pins the exact settlement rule from NethermindEth/nethermind#12964 (we had opened #3422 for the same gap and just closed it in favor of this).

One optional axis if you want it: an explicit top-level `REVERT` control (sender pays only regular gas + intrinsic when the failure is not exceptional). It guards against over-correcting the halt path - historically relevant here, since the earlier reservoir-revert draft got exactly this distinction wrong. Would slot in as a small `failure_mode` parametrization on this test.

spencer-tb:
> One optional axis if you want it: an explicit top-level `REVERT` control (sender pays only regular gas + intrinsic when the failure is not exceptional). It guards against over-correcting the halt path - historically relevant here, since the earlier reservoir-revert draft got exactly this distinction wrong. Would slot in as a small `failure_mode` parametrization on this test.

`test_top_level_failure_spilled_state_gas[revert-*]` pins top level REVERT with an outstanding spillover and `test_nested_failure_resets_to_tx_reservoir[*-spill-revert-*]` does the same with depth :)

spencer-tb (approved):
Optional hardening, idea with claude: derive the budgets from fork costs instead of literals. Every under-funded outcome here still satisfies `cumulative == gas_limit`, so a repricing that outgrows these values would not fail the test, it would just silently shrink the chain. Deriving the budget keeps the depth regimes by construction.

Verified locally: budgets come out at 153,446 / 193,972 / 275,024 with max depths 7 / 13 / 25 and 9 / 18 / 35 SSTOREs, matching the current shape of the existing test.

```python
@pytest.mark.parametrize(
    "descent_frames",
    [
        pytest.param(2, id="descent_frames_2"),
        pytest.param(4, id="descent_frames_4"),
        pytest.param(8, id="descent_frames_8"),
    ],
)
@pytest.mark.valid_from("EIP8037")
def test_spilled_state_gas_consumed_across_halt_chain(
    state_test: StateTestFiller,
    pre: Alloc,
    fork: Fork,
    descent_frames: int,
) -> None:
    """
    Verify spilled state gas stays consumed along a chain of halting frames.

    A self-`DELEGATECALL`ing contract writes `NOT(storage[slot])`, so set and
    clear alternate down the shared-storage call stack and the reservoir is
    recycled instead of drained once. Each frame reuses the value it wrote as
    a `CALL`'s `args_size`, so the frames that set the slot halt on the
    memory-size overflow while the frames that cleared it re-enter the
    contract, interleaving halting and surviving frames. Every halt must burn
    its spill rather than credit it back to the caller's reservoir, so the
    top-level halt charges the whole gas limit.
    """
    slot = 0
    value_offset = 0
    code = (
        # Memory is per-frame, so each frame keeps the value it wrote
        # and reuses it below as the CALL's args_size.
        Op.MSTORE(value_offset, Op.NOT(Op.SLOAD(slot)))
        + Op.SSTORE(slot, Op.MLOAD(value_offset))
        + Op.POP(Op.DELEGATECALL(address=Op.ADDRESS))
        # An all-ones args_size overflows the memory-size calculation
        # and halts the frame. A zero one, in a frame that cleared the
        # slot, re-enters the contract and spawns further frames.
        + Op.POP(Op.CALL(address=Op.ADDRESS, args_size=Op.MLOAD(value_offset)))
    )
    contract = pre.deploy_contract(code=code)

    # One fresh spilled set plus a descent budget in static frame
    # costs. Clear credits recycle the reservoir for deeper sets, and
    # warm frames cost less than the static sum, so the descent runs
    # past the budget. Fork-derived so the depth regimes survive
    # repricings.
    gas_limit = (
        fork.transaction_intrinsic_cost_calculator()()
        + Op.SSTORE(new_value=1).state_cost(fork)
        + descent_frames * code.execution_cost(fork)
    )
    gas_limit_cap = fork.transaction_gas_limit_cap()
    assert gas_limit_cap is not None
    # Below the cap, so the reservoir starts empty and every set spills.
    assert gas_limit < gas_limit_cap

    tx = Transaction(
        to=contract,
        gas_limit=gas_limit,
        sender=pre.fund_eoa(),
        expected_receipt=TransactionReceipt(cumulative_gas_used=gas_limit),
    )

    state_test(pre=pre, post={contract: Account(storage={})}, tx=tx)
```

chfast:
I'm still not sure what the true nethermind bug is. maybe analyzing the fix can help improving the test.
IIRC, only one variant is reproducing the bug.

spencer-tb:
Updated the test to be more future proof and verified against the buggy Nethermind!

PR #3409, opened 2026-08-20
fix(test-evm-tools): restore state test compatibility

danceratopz:
### Description

Restores the `statetest` compatibility output used by goevmlab and other legacy GeneralStateTests consumers.

`stateRoot` is now emitted with exactly one `0x` prefix on stdout and stderr. The mismatch message keeps its established unprefixed `got` and `want` values. The adapter also consumes the legacy `previousHash` field when translating it to `blockHashes`, preventing the stricter `Environment` model from rejecting otherwise supported fixtures.

The duplicate prefix came from treating `Hash.hex()` like Python's `bytes.hex()`: the testing `Hash` implementation already includes `0x`, so manually prepending another prefix produced `0x0x...`. Separately, `previousHash` was copied to `blockHashes` but left in the input object.

Focused unit tests cover both compatibility paths.

Validation:

- The focused regression module passes: 2 passed. As a negative control, both tests fail against the pre-fix adapter with the reported `0x0x...` output and `previousHash` validation error.
- The full testing-package suite passes: 2,048 passed, 29 skipped, 4 xfailed, and 2 xpassed.
- `just static` passes, including mypy over 4,270 source files, Ruff, vulture, execution-spec lint, actionlint, codespell, formatting, and lock checks.

### Related Issues or PRs

Fixes #3400.

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

One prefix is plenty

PR #3424, opened 2026-08-24
fix(evm-tools): accept empty hex transaction values

danceratopz:
## Summary

Treat `"0x"` as zero when it appears as a state-test transaction `value`.

This keeps the EELS state-test runner compatible with goevmlab-generated and other legacy state-test inputs. The syntax also appears in ethereum/legacyTests fillers, [here's an example](https://github.com/ethereum/legacyTests/blob/1f581b8ccdc4c63acf5f2c5c1b155c690c32a8eb/src/LegacyTests/Constantinople/GeneralStateTestsFiller/stPreCompiledContracts2/modexp_0_0_0_25000Filler.json#L90-L92).

## Related Issues/PRs

This is a small fix complementary to[#3409](https://github.com/ethereum/execution-specs/pull/3409), covering the remaining case from [Holiman's comment](https://github.com/ethereum/execution-specs/pull/3409#issuecomment-5379434726). The stricter transaction parsing was introduced by [#2924](https://github.com/ethereum/execution-specs/pull/2924).

## Testing

- `uv run pytest -q packages/testing/src/execution_testing/evm_tools/tests/test_statetest.py`
- `just test-tests`
- `just static`

image

### DIFF
--- a/packages/testing/src/execution_testing/evm_tools/t8n/cli.py
+++ b/packages/testing/src/execution_testing/evm_tools/t8n/cli.py
@@ -116,23 +116,27 @@ def _parse_ommers_from_env_json(env_json: Any, fork: Any) -> List[Ommer]:
 
 def _normalize_tx_json(tx: Dict[str, Any]) -> Dict[str, Any]:
     """
-    Drop fields that the testing ``Transaction`` model rejects.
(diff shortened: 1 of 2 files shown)


spencer-tb (approved):
LGTM! Thanks for fixing this all end to end :)

PR #3425, opened 2026-08-24
feat(tests): expand create collision matrix and add balance-only cases

danceratopz:
### Description
Expand the create collision tests to the full parameter matrix of nonce, code, storage and balance for all cases where EIP-684 defines an abort. The new no-storage cells catch a client that incorrectly aborts on storage instead of code or nonce. Cells with zero nonce and empty code are excluded at param generation: with non-empty storage the behavior is undefined in protocol (EIP-7610 was declined for Glamsterdam), with empty storage the account is deployable.

Also add balance-only success tests for the deployable case, for creation transactions and for the CREATE and CREATE2 opcodes: an account with only a balance is not a collision, creation must succeed and the balance is kept.

Filled locally at Frontier, Paris, Cancun and Amsterdam, all pass, re-verified after rebasing onto `forks/amsterdam` once #3417 merged.

### Related Issues or PRs
Implements the coverage expansion requested in the review of #3417 ([comment 1](https://github.com/ethereum/execution-specs/pull/3417#discussion_r3841725645), [comment 2](https://github.com/ethereum/execution-specs/pull/3417#discussion_r3841749201)).

### Checklist

### Cute Animal Picture

```text
   /\_/\
  ( o.o )    no collisions,
   > ^ <     only deployments
  /|   |\
 (_|___|_)
```

### DIFF
--- a/tests/frontier/create/test_create_collision.py
+++ b/tests/frontier/create/test_create_collision.py
@@ -27,6 +27,11 @@ pytestmark = [
     pytest.mark.pre_alloc_mutable,
 ]
 
+CORRECT_INITCODE = Initcode(
+    deploy_code=Op.STOP,
+    initcode_prefix=Op.SSTORE(0, 1) + Op.SSTORE(1, 0),
+)
+
 # Every account shape where creation must abort under EIP-684: the
 # product of nonce, code, storage and balance with non-empty code or
 # nonce. Cells with zero nonce and empty code are excluded: with
@@ -37,13 +42,13 @@ pytestmark = [
 # instead of code or nonce; cells with non-empty storage also check
 # that the aborted creation neither wipes the storage nor runs the
 # initcode, which would zero slot 0x01 in the correct-initcode case.
-COLLISION_ACCOUNT_PARAMS = [
-    pytest.param(
+COLLISION_ACCOUNT_CASES = [
+    (
         nonce,
         code,
         storage,
         balance,
-        id=(
+        (
             f"nonce_{nonce}-"
             f"{'code' if code else 'no_code'}-"
             f"{'storage' if storage else 'no_storage'}-"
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
LGTM from my end! Would like an approve and merge from @jochem-brouwer

tests/frontier/create/test_create_collision.py
```diff
@@ -196,8 +248,98 @@ def test_create_opcode_collision(
...
+@pytest.mark.with_all_create_opcodes
+def test_create_opcode_balance_only_target(
+    state_test: StateTestFiller,
+    pre: Alloc,
+    create_opcode: Op,
+) -> None:
+    """
+    Test that a contract creation opcode succeeds when the target
+    address has only a balance: an account with zero nonce, no code and
+    no storage is not a collision (EIP-684).
```
jochem-brouwer, line 308: This is not correct, [EIP-684](https://eips.ethereum.org/EIPS/eip-684) only verifies: no code and nonce is zero. The storage check is added in EIP-7610, however, since this is DFId this is still undefined behavior (we shouldo also not test such behavior, which we are not doing here, so that's fine! 😄 )

> If a contract creation is attempted due to a creation transaction, the CREATE opcode, the CREATE2 opcode, or any other reason, and the destination address already has either a nonzero nonce, or a nonzero code length, then the creation MUST throw

jochem-brouwer (commented):
Hi @danceratopz I checked the tests and it looks correct. However, the docs are not correct.

Could you edit the docs? Let me know if I should address this, then I will change and push a commit 😄 👍

jochem-brouwer:
Wait, it's a double negation. We could also add "no balance" but that would be correct also. Ok I will approve it, merge if you are fine, or edit the comment if you think it should change.

jochem-brouwer (approved):

PR #3427, opened 2026-08-24
feat(tests): EIP-8037 - pin gas on the unasserted child-halt spill tests

spencer-tb:
### Description

Pin exact receipt gas for three EIP-8037 scenarios that previously asserted only storage state. The new assertions ensure state gas spilled from a child frame into execution gas remains burned when that frame halts exceptionally, while the state-gas reservoir is restored correctly.

- `test_reservoir_restored_after_child_spill_and_halt` charges the halted child's full forwarded budget plus the caller's own execution and state-gas costs.
- `test_multi_block_mixed_state_operations` pins cumulative receipts for reservoir-funded SSTOREs, child spill-and-revert, and child spill-and-halt transactions across three blocks.
- `test_create2_failed_deposit_refunds_storage_state_gas` pins the failed init frame's full 63/64 share for EIP-3541 rejection and code-deposit OOG, proving the result is independent of the number of reverted storage slots.

The gas calculations use fork-aware bytecode cost helpers and the transaction intrinsic-cost calculator rather than hard-coded schedule values.

### Related Issues or PRs

#3423

### Checklist

- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title matches the target squash commit message.

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
@@ -263,8 +263,12 @@ def test_reservoir_restored_after_child_spill_and_halt(
     parent; the spilled gas stays burned (re-classified as execution).
     The parent does two SSTOREs: the first drains the recovered
     reservoir, the second spills from the parent's own `gas_left`.
+    The receipt pins the halted child's whole budget as consumed, so
+    a credit of the burned spill back to the parent is caught.
     """
     sstore_state_gas = Op.SSTORE(new_value=1).state_cost(fork)
+    intrinsic_cost = fork.transaction_intrinsic_cost_calculator()()
+    child_budget = 500_000
 
     # Child does two SSTOREs then halts
     child = pre.deploy_contract(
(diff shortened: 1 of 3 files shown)


tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_call.py
```diff
@@ -263,31 +263,43 @@ def test_reservoir_restored_after_child_spill_and_halt(
     parent; the spilled gas stays burned (re-classified as execution).
     The parent does two SSTOREs: the first drains the recovered
     reservoir, the second spills from the parent's own `gas_left`.
+    The receipt pins the halted child's whole budget as consumed, so
+    a credit of the burned spill back to the parent is caught.
```
LouisTsai-Csie, line 267: Here's my proposal: Create a scenario where, after the child call halts, the remaining gas only covers SSTORE's execution cost, while the state gas is drawn from the state gas reservoir that is returned by the child.

I'll refactor this in a separate PR and push it shortly.

LouisTsai-Csie (approved):

PR #3428, opened 2026-08-24
fix(specs,tests): widen a frame transaction's chain id to 256 bits

Marchhill:
### Description

[EIP-8141](https://eips.ethereum.org/EIPS/eip-8141) bounds a frame transaction's chain id with `assert tx.chain_id < 2**256`, but `FrameTransaction` declares the field as `U64`, so `rlp.decode_to` rejects anything wider than eight bytes:

```
DecodingError: cannot decode into `FrameTransaction`
	because cannot decode field `FrameTransaction.chain_id`
	because expected at most 8 but got 9
```

A narrowing of the assertion to `2**64` was proposed in ethereum/EIPs#12223 and rejected: chain ids are valid up to `2**256 - 1`, so the field should not be bounded below that. This aligns the reference implementation with that ruling.

The change is behaviour-preserving for every in-range chain id — RLP encodes a `U256` and a `U64` of the same value identically, so no existing fixture changes. What it changes is *where* an out-of-range value is rejected: a chain id of `2**64` now decodes and fails the `check_transaction` chain-id comparison with `WrongChainIdError`, instead of failing to decode. `2**256` still fails to decode, matching the EIP's bound.

Coherence fixes that come with the widening:

- `chain_id(tx)` returns `None | U64 | U256`; the comparison against `block_env.chain_id` is numeric across the two widths, so `WrongChainIdError` fires as before. Its `actual` parameter widens to match.
- `TransactionLoad.json_to_chain_id` used `hex_to_u64` unconditionally and raised `OverflowError` on a wide value. It now selects the width from the resolved transaction class, so only frame transactions widen.

Two tests pin the boundary, which nothing covered before:

- `tests/json_loader/test_transaction_codec.py` decodes a frame transaction at `1`, `2**64` and `2**256 - 1`. Reverting the annotation fails the latter two with `DecodingError` and leaves the rest passing.
- `tests/amsterdam/eip8141_frame_transactions/test_admission_validity.py` adds `chain_id_above_64_bits` and `chain_id_at_256_bit_maximum`, both expecting `TransactionException.INVALID_CHAINID`. Their `txbytes` carry a 9-byte and a 32-byte chain-id item, so a client that stops at the decoder rejects them for the wrong reason.

Checks run: `typecheck`, `lint`, `format-check`, `spellcheck`, `deadcode`, `lint-spec`, `lock-check` (all of `just static` except `lint-actions`, which needs a binary not available here and covers no changed file). `just` itself is not installed locally, so each recipe was run as its underlying `uv run` command. Also `uv run fill --from Bogota --until Bogota tests/amsterdam/eip8141_frame_transactions/` (715 passed) and `uv run pytest tests/json_loader` (58 passed, 2 skipped).

### Related Issues or PRs

ethereum/EIPs#12223.

### Checklist

### Cute Animal Picture

A cat

### DIFF
--- a/src/ethereum/forks/amsterdam/exceptions.py
+++ b/src/ethereum/forks/amsterdam/exceptions.py
@@ -4,7 +4,7 @@ Exceptions specific to this fork.
 
 from typing import TYPE_CHECKING, Final
 
-from ethereum_types.numeric import U64, Uint
+from ethereum_types.numeric import U64, U256, Uint
 
 from ethereum.exceptions import InvalidBlock, InvalidTransaction
 
@@ -20,7 +20,7 @@ class WrongChainIdError(InvalidTransaction):
     [EIP-155]: https://eips.ethereum.org/EIPS/eip-155
     """
 
-    def __init__(self, expected: U64, actual: U64):
+    def __init__(self, expected: U64, actual: U64 | U256):
         super().__init__(f"expected chain_id `{expected}` but got `{actual}`")
         self.expected = expected
         self.actual = actual
--- a/src/ethereum/forks/amsterdam/transactions/__init__.py
+++ b/src/ethereum/forks/amsterdam/transactions/__init__.py
@@ -883,12 +883,18 @@ def check_nonce(tx: Transaction, sender_nonce: Uint) -> None:
         raise NonceMismatchError("nonce too high")
 
 
-def chain_id(tx: Transaction) -> None | U64:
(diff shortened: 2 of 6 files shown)


Marchhill:
Closing while the underlying question is settled: the reference declares `chain_id: U64` on all four other typed transactions and on `BlockEnvironment`, so widening only the frame transaction leaves the reference internally inconsistent. The right resolution may be a wider change or an amendment to the EIP rather than this patch. The branch is preserved and this can be reopened once that is decided.

PR #3434, opened 2026-08-24
feat(spec-specs, tests): implement EIP-8368 CPSB recalibration for a new gas limit

spencer-tb:
### Status

Documentation/reference prototype for EIP-8368. This draft is not proposed for merge in its current form.

The protocol change is implemented in place in Amsterdam and activated for fixture generation through the testing-side Bogota pseudo-fork. It targets the post-Amsterdam fork and has no substrate exception.

### Description

Implements [EIP-8368](https://github.com/ethereum/EIPs/blob/582684e2d7d372c09f45777be8ea603e485e9e9d/EIPS/eip-8368.md): provisionally doubles the EIP-8037 cost per state byte from 1530 to 3060, derived for a 300M reference block gas limit.

- Spec: changes Amsterdam's `COST_PER_STATE_BYTE` to the provisional 3060 value; storage creation, new accounts, authorization state, code deposit, and the system-transaction reservoir continue to derive from the shared constant.
- Framework: registers EIP-8368 and a Bogota pseudo-fork override for `cost_per_state_byte()`, with the temporary Amsterdam execution-module mapping.
- Tests: adds exact-fit and one-short storage coverage, receipt-pinned account and code-deposit charges, and a complete transition test parked until a real post-Amsterdam spec fork exists.
- Existing tests: derives representative sibling-suite budgets through fork APIs and parks fixed-budget ported-static cases with 459 searchable `valid_before("EIP8368")` markers across 300 files.

### Required EIP follow-up

The implementation review found consensus-observable details that should be settled before production fixtures are finalized.

#### Specification

1. Replace the placeholder reference block gas limit and `CPSB` with normative values and rationale. This prototype assumes a 300M reference limit and `CPSB = 3060`; those values are provisional and are not specified by the current EIP.
2. Specify the integer rounding rule for the EIP-8037 derivation. Ceiling division reproduces 1530 at the original 150M reference and yields 3060 at 300M, while rounding to nearest would yield 3059.

#### Gas and compatibility language

3. State the activation coupling to the block gas limit and EIP-7954: at `CPSB = 3060`, depositing a 64 KiB contract consumes roughly 201M gas including the new-account footprint, so the maximum-size contract is not deployable below that block-gas range.
4. Document the EIP-7825 capacity consequence: doubling `CPSB` halves the amount of new state that cap-funded transaction flows can create, including the spill path used when no state-gas reservoir is available.

### Known limitations

- The protocol parameters are provisional; changing them requires updating the spec constant, framework override, test derivation, and the fixed-budget parking sweep.
- The transition test is intentionally skipped because the pseudo-fork executes Amsterdam on both sides; its complete body is retained for a real post-Amsterdam spec fork.
- The 459 parked markers cover budgets derived at `CPSB = 1530`; one two-axis ported-static case necessarily over-parks three passing combinations because its failing cross-product cannot be marked independently.
- The prototype remains pinned to base commit `263606b3cd0a352112697870c2c878e2423546b7`; the moving `forks/amsterdam` branch has advanced since the assessment began.

### References

- EIP revision used by these tests (`582684e2d7`): https://github.com/ethereum/EIPs/blob/582684e2d7d372c09f45777be8ea603e485e9e9d/EIPS/eip-8368.md
- Tracker: https://github.com/ethspecs/pm/issues/72
- Complexity assessment: pending
- EIP discussion: https://ethereum-magicians.org/t/eip-8368-cpsb-recalibration-for-new-gas-limit/29293


spencer-tb:
Opened in the wrong repository. The reference PR for this prototype is spencer-tb/execution-specs#56.

Issue #3301, opened 2026-08-04
bug(tests): Intrinsic gas no longer guarantees a transaction reaches the top frame


marioevz:
### Summary (AI assisted)

Before Amsterdam, funding a transaction with its intrinsic gas cost guaranteed it would reach the start of EVM execution. EIP-2780 breaks that invariant: costs that depend on the pre-transaction state — an authority's account creation, a net-new delegation indicator, the first write to an account leaf — moved out of the intrinsic cost and are now charged at the top frame, *after* the intrinsic deduction but *before* execution begins. A transaction funded with only its intrinsic cost now runs out of gas in the top frame.

This is silent. The transaction is still valid and still included, so tests that only assert validity (or assert `gas_used` without asserting receipt status) keep passing while the transaction they meant to exercise does nothing.

`set_delegation`'s own docstring in `src/ethereum/forks/amsterdam/vm/eoa_delegation.py` states the consequence:

> These costs depend on the authority's current state and so cannot be charged in the intrinsic cost. Insufficient gas raises an `OutOfGasError`; the caller rolls back the authorizations applied so far and halts the top frame.

### Evidence

`tests/amsterdam/eip7976_increase_calldata_floor_cost/test_transaction_validity.py::test_transaction_validity_type_4` and its Prague ancestor `tests/prague/eip7623_increase_calldata_cost/test_transaction_validity.py::test_transaction_validity_type_4` size their gas limit as `tx_intrinsic_gas_cost_including_floor_data_cost + tx_gas_delta` (see `conftest.py`), with no top-frame term. Filled at Amsterdam, **every** type-4 transaction that gets included fails:

```
eip7976 type_4 at Amsterdam: success=0  FAILED(status=False)=56  no-receipt(invalid, expected)=28
eip7623 type_4 at Amsterdam: success=0  FAILED(status=False)=56  no-receipt(invalid, expected)=28
```

Taking `single_authorization-no_access_list-exact_gas-floor_gas_greater_than_intrinsic_gas` as the concrete case — 163 bytes of calldata, one authorization:

```
intrinsic              = 23468
floor (= gas limit)    = 25432
top_frame_execution    =  8000   (ACCOUNT_WRITE)
top_frame_state        = 35190   (AUTH_BASE)
intrinsic + top_frame  = 66658

shortfall              = 41226
```

The resulting fixture:

- receipt `status = False`, `cumulativeGasUsed = 25432`
[...]

PR #3440, opened 2026-08-25
feat(forks): add transaction_gas_to_reach_execution_calculator

zexoverz:
Part of #3301. EIP-2780 moved state-dependent costs into the top frame, so intrinsic gas no longer guarantees a transaction reaches execution. This adds a fork-level helper returning that bound as `max(validity_minimum, deducted_prior_execution + top_frame_exec + top_frame_state)`. The `max` is load-bearing: it also covers the calldata floor, so the helper is a drop-in wherever the intrinsic calculator is used with default flags. Additive only, +119 lines in `base_fork.py`, no fork-subclass changes.

Verified by sweeping 3,840 no-authorization shapes across all 25 forks: the helper is identical to `transaction_intrinsic_cost_calculator` everywhere except 28 Amsterdam-only shapes where a top-frame charge is genuinely non-zero. Fixture output is byte-identical. `just static`, `just test-tests` (2085 passed), `just spec-tools` and `uv run fill` across the affected suites are green. The call-site migration is deliberately not in this PR; it changes what `test_transaction_validity` asserts, and I have left a note on the issue.

### DIFF
--- a/packages/testing/src/execution_testing/forks/base_fork.py
+++ b/packages/testing/src/execution_testing/forks/base_fork.py
@@ -230,6 +230,47 @@ class TopFrameGasCalculator(Protocol):
         pass
 
 
+class TransactionGasToReachExecutionCalculator(Protocol):
+    """
+    A protocol to calculate the gas a transaction must be funded with to
+    be guaranteed to reach the start of EVM execution at a given fork.
+    """
+
+    def __call__(
+        self,
+        *,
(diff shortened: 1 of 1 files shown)


spencer-tb:
Closing as not a bug, the validity tests pin the intrinsic boundary by design under EIP-2780, the fixtures already pin the top-frame halt through the receipts root and post state, and the execution side is covered by the EIP-2780 floor tests.

PR #3441, opened 2026-08-25
feat(test-benchmark): strengthen and consolidate TSTORE benchmark

chfast:
### Description

Rework test_tstore into a single parametrized worst-case benchmark: key = GAS (unique slot per op, no memory counter) or PUSH0 (fixed slot), value = COINBASE (a nonzero op) so each write actually inserts an entry. A zero value lets clients skip the insert and silently hide store growth, so the test asserts the block coinbase is nonzero.

The tload parameter adds a transient read (TLOAD and TSTORE share the same 100 gas cost):
- none: TSTORE only.
- hit: seed the value once, store it under the key and read it back, feeding the result into the next TSTORE so the read is used, not discarded.
- miss: store the key, then read a fresh, never-written key so TLOAD misses on the growing store; the value stays COINBASE (nonzero).

This subsumes the bloatnet test_tstore_same_key (fixed key) and test_tstore_unique_keys (unique key), which are removed along with their now-empty module.

### Related Issues or PRs

N/A.

### Checklist


chfast:
## Benchmark results — geth & besu

Filled fixture executed at **Osaka, 300M gas** via block-test
(`evm blocktest` / `evmtool block-test`), single machine, best-of-3 (geth) /
best-of-2 (besu).

Clients: geth `evm` 1.17.2-stable, besu `evmtool` 26.8-develop.

| key            | tload | geth MGas/s | besu MGas/s |
| -------------- | ----- | ----------: | ----------: |
| unique (GAS)   | none  |         141 |          39 |
| unique (GAS)   | hit   |         270 |          56 |
| unique (GAS)   | miss  |         288 |          56 |
| fixed (PUSH0)  | none  |        1222 |         385 |
| fixed (PUSH0)  | hit   |        1231 |         462 |
| fixed (PUSH0)  | miss  |        1254 |         455 |

The `unique` (distinct transient slot per op) variants are ~9x slower than the
`fixed` baseline — the intended worst case. `tload=hit/miss` add a TLOAD, so
fewer TSTOREs run per unit gas and the per-tx transient store grows less,
giving higher throughput than `tload=none`.

Notes:
- Single-machine, relative numbers (not hardware-normalized); besu is a locally
  built dev `evmtool`, so treat its absolutes as indicative.
- Amsterdam was tried first but no client runner supports it yet (geth's t8n /
  blocktest excludes it), so these are on Osaka.

spencer-tb (approved):
LGTM from my side, the new parameterization covers all that `test_transient_storage.py` did  from what I can tell :)

Could you update this [line](https://github.com/ethereum/execution-specs/blob/123ddc80b4ac6d58017cc38da6a682e358cf1a60/docs/filling_tests/fill_stateful.md#L98) to one of the remaining stub-free bloatnet tests, e.g. `tests/benchmark/stateful/bloatnet/test_account_query.py` as we are removing `test_transient_storage.py`?

Will leave @LouisTsai-Csie for final approve and merge!

LouisTsai-Csie (approved):

PR #3443, opened 2026-08-26
fix(eip-8141): account block execution gas before refund

chugarchugarr:
### Description

Fix EIP-8141 frame-transaction block gas accounting so storage refunds reduce payer-facing gas without reducing block execution gas, as required by EIP-7778.

The payer and receipt remain post-refund. At the block-processing boundary:

- payer fee and receipt cumulative gas: post-refund
- block execution gas: pre-refund
- block state gas: net after state-gas refills
- the calldata floor is applied independently to the payer-facing and block-accounted totals

### Reviewer-requested changes incorporated

Current head: `ce7e32f8f7ac6e7cbdfaf87779b787c6f4e40406`.

All three items from `@gurukamath`'s review are incorporated:

1. `tests/amsterdam/eip8141_frame_transactions/spec.py` is pinned to merged EIP update commit `7d1c8bfb945cbb53479217df3bf1da67b3aa445b`.
2. Payer-facing `gas_used` is derived inside `settle_frame_transaction_gas` and carried by `FrameTransactionGasSettlement` rather than reconstructed inline in `process_frame_transaction`.
3. `test_storage_refund_settlement` covers all three calldata-floor orderings: post-refund above floor, floor between pre/post refund, and floor above pre-refund.

The two inline review threads have been answered and resolved, and re-review has been requested. No acceptance claim is being made before that review completes.

### Verification state

The earlier focused validation (`9 passed`, Ruff check/format check, and `git diff --check`) predates the final reviewer-requested refinements and is preserved as prior evidence rather than represented as exact-current-head CI.

On current head, the upstream fork-triggered `render`, `Python Specification`, and `Build Docs` workflow runs are all `action_required`; they require upstream authorization before they can execute. There is therefore no author-side CI action remaining on this head.

### Base

`eips/amsterdam/eip-8141` at `6798542ebd017b683b688489d770bf206c8bd3ba`.

PR #3446, opened 2026-08-26
test(test-execute): pin genesis commitment provider

chugarchugarr:
### Description

Addresses the remaining regression-coverage gap in #3251. The production MPT decoupling has already landed in #3279; this PR pins both execute-path boundaries that issue identified.

The existing execute/fill parity test could agree on the wrong root because it reused the allocation whose private commitment provider had already been selected by `build_genesis_header()`.

This adds independent assertions for both paths:

1. **Hive execute builder:** serialize and reconstruct the logical allocation, independently select the genesis fork's provider, compute the expected root, and compare it with the execute genesis header.
2. **eth_config genesis model:** construct a non-empty BinaryTree genesis allocation, independently compute its MPT contrast root, and prove model initialization selects PBT before the genesis hash is evaluated.

Genesis-header parity is therefore no longer sufficient by itself; each execute boundary must independently agree with the fork-selected commitment provider.

### Verification

- `uv run pytest -q packages/testing/src/execution_testing/cli/tests/test_execute_genesis.py packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/eth_config/tests/test_genesis.py` — 42 passed
- Ruff check and format check — passed
- `git diff --check` — passed

### Scope

- test-only
- no duplicate state-commitment plumbing
- no MPT→PBT activation/migration machinery
- no mainnet fork behavior change

### Base

`projects/binary-trie` at `09d2088cf9f338e660cb04f948653baf544b7ba6`.

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/eth_config/tests/test_genesis.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/execute/eth_config/tests/test_genesis.py
@@ -5,7 +5,12 @@ from pathlib import Path
 
 import pytest
 
-from execution_testing.base_types import Hash
+from execution_testing.base_types import (
+    Account,
+    Address,
+    Hash,
+    StateCommitment,
+)
 from execution_testing.forks import (
     BPO1,
     BPO2,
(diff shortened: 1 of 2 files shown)


chugarchugarr:
Closing this contribution to keep my active upstream review surface focused on #3443. This PR received no maintainer review and its fork workflows remained authorization-gated, so I am not asking the project to spend review bandwidth on it.

The branch and diff remain available and can be reopened if a maintainer specifically wants to continue the genesis-provider regression coverage.

PR #3449, opened 2026-08-26
refactor(tests): enhance EIP-8037 test coverage part 2

LouisTsai-Csie:
### Description

Strengthens the EIP-8037 state-gas tests. No spec changes and no new behavior under test, every edit either adds an assertion that was missing, corrects opcode metadata the framework needs to price gas, or removes a test whose premise cannot hold.

The same three problems recur across the suite, so the diff repeats the same three fixes. Reading one instance of each is enough to review the rest.

1. Tests that asserted nothing: Several tests were named after a gas property but never checked gas, no `header_verify`, no `expected_receipt`, no `post` state, so they passed whether or not the property held. Each now pins `Header(gas_used=...)` and/or `TransactionReceipt(cumulative_gas_used=...)` derived from the fork model, with an `assert` naming which gas dimension is expected to dominate.
2. Missing opcode metadata: `execution_cost` and `state_cost` are derived from opcode metadata, so a missing annotation silently mis-prices the budget a test builds around it.
3. Recording `SSTORE`s that consumed the budget under test:  Tests that recorded a sub-call result into a fresh slot paid `STORAGE_SET` state gas for the recording itself, so the reservoir no longer covered only the operation being measured. The slot is now pre-seeded, making the write a nonzero-to-nonzero modify that costs no state gas.

Moreover, where a test needs to prove state gas came from the reservoir rather than spilling out of `gas_left`, the operation now runs in a sub-call forwarded exactly its execution cost. With no `gas_left` left to spill from, it completes only if the reservoir funded it. `CALL` gas arguments are derived from the callee's `execution_cost` instead of round magic numbers.

## Scope
- [x] test_state_gas_call.py
- [x] test_state_gas_calldata_floor.py
- [x] test_state_gas_multi_block.py
- [x] test_state_gas_ordering.py
- [x] test_state_gas_pricing.py
- [x] test_state_gas_sstore.py

## Followup Idea
- Duplication between `test_calldata_floor_with_sstore` and `test_calldata_floor_higher_than_execution_with_state_ops`
- Remove `test_sstore_restoration_reservoir_spillover`, `test_sstore_oog_reservoir_inflation_detection`, `test_call_oog_reservoir_inflation_detection`, `test_selfdestruct_oog_reservoir_inflation_detection` and `test_create_oog_reservoir_inflation_detection`: these cases no longer follows the latest glamsterdam spec.

### Checklist

### Cute Animal Picture


marioevz (approved):
Amazing improvements overall!

I added a few more parametrizations and checks in some of the tests.

Thanks for all the changes!

PR #3450, opened 2026-08-26
fix(test-execute): ignore tests/spec_tools during collection

danceratopz:
### Description

PR #3307 moved `tests/evm_tools/test_*.py` to `tests/spec_tools/`, but the fill and execute pytest configs still ignored the old path. `execute` collects these plain unit tests, gives them a `fork` parameter via the autouse `pre` fixture, and then crashes during collection because they have no spec type format:

```
INTERNALERROR> ValueError: No spec type format found in the test item.
```

This is what stopped the eels/execute simulator runs on the glamsterdam Hive dashboard. Update the stale `--ignore tests/evm_tools` entries to `--ignore tests/spec_tools`.

Note: this also needs to reach `devnets/glamsterdam/8` for the dashboard to recover.

### Related Issues or PRs

Fixes #3438:
- #3438

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

```
 (\ /)
 ( .c.)
 o(")(")
```

Issue #3438, opened 2026-08-25
eels/execute, hive mode internal error


jangko:
As can be seen from https://hive.ethpandaops.io/#/group/glamsterdam dashboard.

eels/execute, hive mode have no update since 19 August.

The log shows something like this:
```
rootdir: /execution-specs
configfile: packages/testing/src/execution_testing/cli/pytest_commands/pytest_ini_files/pytest-execute-hive.ini
testpaths: tests/
plugins: custom-report-1.0.1, json-report-1.5.0, html-4.1.1, xdist-3.8.0, metadata-3.1.1, regex-0.2.0, cov-4.1.0
collecting ... collected 26452 items
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/execution-specs/.venv/lib/python3.12/site-packages/_pytest/main.py", line 289, in wrap_session
INTERNALERROR>     session.exitstatus = doit(config, session) or 0
INTERNALERROR>                          ^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/execution-specs/.venv/lib/python3.12/site-packages/_pytest/main.py", line 342, in _main
INTERNALERROR>     config.hook.pytest_collection(session=session)
INTERNALERROR>   File "/execution-specs/.venv/lib/python3.12/site-packages/pluggy/_hooks.py", line 512, in __call__
INTERNALERROR>     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/execution-specs/.venv/lib/python3.12/site-packages/pluggy/_manager.py", line 120, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
INTERNALERROR>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/execution-specs/.venv/lib/python3.12/site-packages/pluggy/_callers.py", line 167, in _multicall
INTERNALERROR>     raise exception
INTERNALERROR>   File "/execution-specs/.venv/lib/python3.12/site-packages/pluggy/_callers.py", line 139, in _multicall
[...]

PR #3451, opened 2026-08-26
feat(ci): test `execute`, `consume` and `check_eip_versions` via collect-only

danceratopz:
### Description

Add cheap collect-only CI guards for our pytest commands. This would have caught the collection crash behind #3438 (fixed in #3450).

- `pytest_hive`: This also allows `--collect-only` with hive simulators w/o actually having to connect to a hive dev backend.
- New `just execute-collect`: collect the full `tests/` tree via both the `execute remote` and `execute hive` plugin stacks (this reproduces the #3438 crash without the fix in #3450). Asserts a collected-count floor so a drastic silent shrink also fails.
- New `just consume-collect`: fill a mini fixture set (`test_dup.py`, 16 cases, all formats), then collect it via `consume engine/rlp/enginex/direct` with exact-count assertions: an empty parametrization (e.g. a missing fixture format) collects a single placeholder item and exits 0, so counts must be checked. Notably, `consume enginex` currently has no CI coverage at all: its `hive-consume.yaml` matrix rows are commented out.
- New `just check-eip-versions-collect`: collect the EIP version checks over the full `tests/` tree with the same collected-count floor. This is the guard for the failure mode fixed in #3461 (verified: it exits non-zero with #3461's parent commit, passes with the fix).
- Run the three recipes as jobs in `test.yaml` on every PR.
- `hive-execute.yaml`: also trigger on changes to the shared plugins and `pytest-execute.ini`, both of which the E2E consumes.

### Related Issues or PRs

Related to #3438; requires #3450 and #3461 (both merged).

### Checklist

### Cute Animal Picture

```
 /\_/\
( o.o )
 > ^ <
```

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -226,6 +226,64 @@ json-loader *args: (_tmp "json-loader")
         "$@" \
         tests/json_loader
 
+# Collect (without running) the full test tree via the execute remote and execute hive plugin stacks
+[group('integration tests')]
+execute-collect:
+    #!/usr/bin/env bash
+    # Assert a sane collected-test floor: guards against a silent pass if
+    # collection shrinks drastically or yields only placeholder items.
+    set -euo pipefail
+    summary_floor='^[0-9]{4,}(/[0-9]+)? tests collected'
+    uv run execute remote \
+        --collect-only -q \
+        --fork "{{ latest_fork }}" \
+        --rpc-endpoint http://127.0.0.1:1 \
+        --rpc-seed-key 0x0000000000000000000000000000000000000000000000000000000000000001 \
+        --rpc-chain-id 1 \
+        | tee /dev/stderr | tail -n 3 | grep -qE "$summary_floor"
+    uv run execute hive \
+        --collect-only -q \
+        --fork "{{ latest_fork }}" \
+        | tee /dev/stderr | tail -n 3 | grep -qE "$summary_floor"
+
+# Collect (without running) the EIP version checks over the full test tree
+[group('integration tests')]
+check-eip-versions-collect:
+    #!/usr/bin/env bash
+    # Assert a sane collected-test floor: guards against a silent pass if
+    # collection shrinks drastically or yields only placeholder items.
+    set -euo pipefail
+    export GITHUB_TOKEN="${GITHUB_TOKEN:-$(gh auth token)}"
+    uv run check_eip_versions \
+        --collect-only -q \
+        | tee /dev/stderr | tail -n 3 \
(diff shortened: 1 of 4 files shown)


Justfile
```diff
@@ -226,6 +226,54 @@ json-loader *args: (_tmp "json-loader")
...
+        --rpc-chain-id 1 \
+        | tee /dev/stderr | tail -n 3 | grep -qE "$summary_floor"
+    uv run execute hive \
+        --collect-only -q \
+        --fork "{{ latest_fork }}" \
+        | tee /dev/stderr | tail -n 3 | grep -qE "$summary_floor"
+
+# NOTE: the mini fill is pinned to Osaka: all-formats Amsterdam fills
+# currently fail the shared-genesis consistency check (BAL fixtures
+# observe the genesis hash).
```
danceratopz: Ooops. We should update this, when fixed!

spencer-tb: Sgtm! Lets wait for the other PR :)

danceratopz: Moved to draft until we get #3219 in :wink:

danceratopz: Done: rebased on #3265 and the mini fill now uses latest_fork (currently Amsterdam), note removed. Verified locally: the all-formats fill passes and the collected counts still hold (16/16/16/32).

Justfile
```diff
@@ -226,6 +226,64 @@ json-loader *args: (_tmp "json-loader")
...
+        -q \
+        tests/frontier/opcodes/test_dup.py
+    uv run consume engine --collect-only -q --input "$fixtures" \
+        | tee /dev/stderr | grep -q '^16 tests collected'
+    uv run consume rlp --collect-only -q --input "$fixtures" \
+        | tee /dev/stderr | grep -q '^16 tests collected'
+    uv run consume enginex --collect-only -q --input "$fixtures" \
+        | tee /dev/stderr | grep -q '^16 tests collected'
+    uv run consume direct --collect-only -q --input "$fixtures" \
+        | tee /dev/stderr | grep -q '^32 tests collected'
```
spencer-tb: Is there a way to not hard code these?

spencer-tb (approved):
LGTM! One nit/idea :)

Justfile
danceratopz: Less is more: https://github.com/ethereum/execution-specs/pull/3451/changes/ad31c64de3b494a35234dd8637420a98900c0015 uses `-o empty_parameter_set_mark=fail_at_collect`.

PR #3455, opened 2026-08-26
feat(zkevm): use annotated ssz to avoid type duplication

jsign:
### Description

This PR removes duplicate Python types from Amsterdam stateless SSZ serialization.

Previously, each execution type had a second `SSZ*` type. Hand-written functions converted values between both type sets.

This structure repeated every field and required changes in multiple places for each schema update.

The normal specification dataclasses are now the single source of truth. They retain their frozen behavior, tuple fields, and existing numeric types.

The SSZ wire format and schema ID `0x1501` do not change.

#### New `ethereum.utils.ssz` module

This PR adds `ethereum.utils.ssz`, a small adapter between specification dataclasses and `remerkleable`.

`SszContainer` provides these operations:

- `encode_bytes()`
- `decode_bytes()`
- `hash_tree_root()`

`ProgressiveSszContainer` identifies dataclasses that use an SSZ progressive container.

The adapter infers fixed-size bytes, fixed-width integers, booleans, and nested containers from the Python field types.

Fields with extra SSZ rules use `Annotated` markers. These markers define integer widths, byte limits, list limits, and progressive collections.

The adapter creates temporary `remerkleable` views during serialization. It converts decoded views back to the original specification types.

This design prevents mutable `remerkleable` lists and numeric views from entering the normal execution path.

### Related Issues or PRs

N/A.

### Checklist

- [ ] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) and [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [ ] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` match the applicable repository labels.


src/ethereum/utils/ssz.py
jsign, line 1: I think this can be re-usable in the future -- but might get some opinions eventually from STEEL team.

src/ethereum/forks/amsterdam/stateless_ssz.py
jsign, line 1: This file is gone.

src/ethereum/utils/ssz.py
```diff
@@ -0,0 +1,312 @@
+"""
```
frisitano, line 1: This looks good to me, but I just wanted to add a reference to how it's done in the consensus specs to see if there's potential to use the same pattern for standardisation reasons - no strong opinion either way:
https://github.com/ethereum/consensus-specs/tree/master/tests/core/pyspec/eth_consensus_specs/utils/ssz

frisitano (approved):

src/ethereum/utils/ssz.py
jsign: I think consensus specs use `remerkleable` types as their canonical runtime model, but execution-specs use `ethereum_types` thus why we need this annotated types.

frisitano: got it, thanks for the explanation

PR #3464, opened 2026-08-28
refactor(spec-specs): add `OPCODE_SELFBALANCE`, `OPCODE_TLOAD` and `OPCODE_TSTORE`

### LINKED ISSUE #2911: Add dedicated GasCosts parameters for SELFBALANCE, TLOAD, and TSTORE
## What

`gas.py` already follows a per-opcode naming convention where each opcode aliases a tier constant (e.g. `OPCODE_ADD = VERY_LOW`, `OPCODE_POP = BASE`). However, `SELFBALANCE`, `TLOAD`, and `TSTORE` are missing from this scheme — their instruction handlers charge the shared tier constants directly:

| Opcode | Currently charges | Value | Location (`forks/amsterdam`) |
|---|---|---|---|
| `SELFBALANCE` | `GasCosts.FAST_STEP` | 5 | `src/ethereum/forks/amsterdam/vm/instructions/environment.py:521` |
| `TLOAD` | `GasCosts.WARM_ACCESS` | 100 | `src/ethereum/forks/amsterdam/vm/instructions/storage.py:158` |
| `TSTORE` | `GasCosts.WARM_ACCESS` | 100 | `src/ethereum/forks/amsterdam/vm/instructions/storage.py:188` |

## Why

For consistency with every other opcode, each should have its own named parameter in `GasCosts`, even when the value is duplicated. Benefits:

1. **Consistency** — `gas.py` documents one parameter per opcode; these three are the current exceptions.
2. **Future-proofing** — if any of these opcodes is repriced later, the value is already isolated behind a dedicated constant and only one line needs to change.

This is a pure refactor for consistency — **no behavioral or gas-value change** is intended.

## Proposed change

Add the missing aliases to `GasCosts`, following the existing convention:

```python
OPCODE_SELFBALANCE = FAST_STEP
OPCODE_TLOAD = WARM_ACCESS
OPCODE_TSTORE = WARM_ACCESS
```

…and update the three `charge_gas` call sites to reference them (`GasCosts.OPCODE_SELFBALANCE`, etc.).

> Note: `SLOAD` and `SSTORE` also charge `WARM_ACCESS` directly and lack `OPCODE_*` aliases, though their dynamic gas logic makes them a slightly different case. This issue is scoped to the three opcodes above; the `SLOAD`/`SSTORE` aliases can be addressed separately if desired.

_Target branch: `forks/amsterdam`._

skyyycodes:
### Description

Every opcode in `GasCosts` has its own `OPCODE_*` alias for whatever constant it charges. `OPCODE_ADD = VERY_LOW`, `OPCODE_POP = BASE` and so on. SELFBALANCE, TLOAD and TSTORE were the three exceptions. This adds the missing aliases and points the handlers at them.

**SELFBALANCE** charged `GasCosts.FAST_STEP` directly in every fork that has the opcode. That is Istanbul onwards, so 17 forks.

I put the new constant straight after `OPCODE_CHAINID`. CHAINID is 0x46 and SELFBALANCE is 0x47 so that keeps opcode order, and it groups the two opcodes Istanbul added. The same slot is free in all 17 forks so `PatchHygiene` sees no reordering.

Worth noting that `FAST_STEP` is used in exactly one place in the whole tree and that place is `self_balance`. It is also only defined from Istanbul on, so it arrived with the opcode. It is `GasFastStep` from EIP-1884.

**TLOAD and TSTORE** already have constants in `amsterdam`. Cancun through bpo5 did not, and still charged `GasCosts.WARM_ACCESS` directly. This adds them to those eight forks.

I used the alias form the issue proposes rather than copying the amsterdam literals, and I left amsterdam alone. The reason is that the amsterdam literals came in with EIP-8038, which reprices state access. Transient storage is not persistent state, so pinning TLOAD and TSTORE to a literal there stops them moving with that repricing. Cancun through bpo5 have no EIP-8038, and EIP-1153 defines both opcodes as costing a warm storage read, so `WARM_ACCESS` is the honest thing to alias in those forks. Happy to switch them to literals if you would rather have one form everywhere.

Applying all of this to every fork is what #2396 did when it added the opcode gas constants in the first place.

Istanbul and Muir Glacier had to move together. `GlacierForksHygiene` only lets a Glacier fork differ from the one before it in the difficulty block.

Gas does not change anywhere. SELFBALANCE stays at 5. TLOAD and TSTORE stay at 100.

### What I ran

- `just lint-spec` clean
- `just spec-tools` 22 passed
- `just fill` across the CI matrix. Frontier to Paris 18620 passed. Shanghai to Cancun 14631. Prague 14420. Osaka 15220. Amsterdam 19987. Nothing failed.
- `just fill tests/ported_static/stSelfBalance/` 28 passed, including `test_self_balance_gas_cost`
- ruff lint and format, vulture, codespell and mypy over `src/ethereum/forks/` all clean
- Checked every fork to confirm each handler charges its new constant and that the values still come out as 5 and 100

### Related Issues or PRs

Fixes #2911.

### Checklist

### DIFF
--- a/src/ethereum/forks/amsterdam/vm/gas.py
+++ b/src/ethereum/forks/amsterdam/vm/gas.py
@@ -226,6 +226,7 @@ class GasCosts:
     OPCODE_PREVRANDAO: Final[ExecutionGas] = BASE
     OPCODE_RETURNDATASIZE: Final[ExecutionGas] = BASE
     OPCODE_CHAINID: Final[ExecutionGas] = BASE
+    OPCODE_SELFBALANCE: Final[ExecutionGas] = FAST_STEP
     OPCODE_BASEFEE: Final[ExecutionGas] = BASE
     OPCODE_BLOBBASEFEE: Final[ExecutionGas] = BASE
     OPCODE_SLOTNUM: Final[ExecutionGas] = BASE
--- a/src/ethereum/forks/amsterdam/vm/instructions/environment.py
+++ b/src/ethereum/forks/amsterdam/vm/instructions/environment.py
@@ -524,7 +524,7 @@ def self_balance(evm: Evm) -> None:
     pass
 
     # GAS
-    charge_gas(evm, GasCosts.FAST_STEP)
+    charge_gas(evm, GasCosts.OPCODE_SELFBALANCE)
 
     # OPERATION
     # Non-existent accounts default to EMPTY_ACCOUNT, which has balance 0.
--- a/src/ethereum/forks/arrow_glacier/vm/gas.py
+++ b/src/ethereum/forks/arrow_glacier/vm/gas.py
@@ -135,6 +135,7 @@ class GasCosts:
     OPCODE_DIFFICULTY: Final[Uint] = BASE
     OPCODE_RETURNDATASIZE: Final[Uint] = BASE
     OPCODE_CHAINID: Final[Uint] = BASE
+    OPCODE_SELFBALANCE: Final[Uint] = FAST_STEP
     OPCODE_BASEFEE: Final[Uint] = BASE
     OPCODE_PUSH: Final[Uint] = VERY_LOW
     OPCODE_DUP: Final[Uint] = VERY_LOW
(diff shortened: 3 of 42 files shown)


LouisTsai-Csie (approved):
Nice, thanks for the contribution!

PR #3467, opened 2026-08-28
fix(client-clis): map frame transaction rejections in the Nethermind mapper

Marchhill:
## What

`NethermindExceptionMapper` gains the `TYPE_6_*` frame transaction entries, plus four other wordings that were unmapped or mislabelled.

## Why

In the hive `frames` group's last completed Nethermind run (2026-08-23, `eels/consume-engine`, filter `.*fork_Bogota.*`) the client scored **24872/24944**. Of the 72 failures, **59 are this file**: 47 named a `TYPE_6_*` exception with no entry, 9 resolved to `TYPE_3_TX_WITH_FULL_BLOBS` instead of the expected label, 2 were an unmapped `GAS_LIMIT_EXCEEDS_MAXIMUM`, and 1 resolved to `INSUFFICIENT_ACCOUNT_FUNDS` where `GASLIMIT_PRICE_PRODUCT_OVERFLOW` was expected.

The 9 show what the missing entries cost:

```
got:      "[TransactionException.TYPE_3_TX_WITH_FULL_BLOBS] Transaction 0 is not valid:
           Expected a sequence prefix to be in the range of <192, 255> and got 147"
expected: "TransactionException.TYPE_6_INVALID_FRAME_FORMAT"
```

That is not a type-3 transaction and carries no blobs. `Transaction <n> is not valid: <cause>` is the wrapper the client puts around *any* payload transaction that fails to RLP-decode, and the existing pattern `r"Transaction \d+ is not valid"` matches the wrapper while discarding the `<cause>` that identifies the failure — so every decode failure came back labelled type-3.

## What this changes

Three frame entries, grouped to match the exceptions' own docstrings: `TYPE_6_INVALID_FRAME_FORMAT` (the static frame constraints, plus decode-time rejections of a field too wide or long for its type), `TYPE_6_INVALID_SIGNATURE` (a signature entry failing protocol validation), and `TYPE_6_INVALID_FRAME_EXECUTION` (a well-formed, signed transaction that frame execution then invalidates).

Three more the catch-all was swallowing: `GASPRICE_OVERFLOW` / `PRIORITY_OVERFLOW` (a fee field wider than 32 bytes trips the decoder's length guard, which names neither field nor type, so both take the same pattern), `GAS_LIMIT_EXCEEDS_MAXIMUM` (a frame transaction reports the cap against its own reservation, needing a second wording), and `GASLIMIT_PRICE_PRODUCT_OVERFLOW` (composed onto the insufficient-funds wording, which is why it resolved only to `INSUFFICIENT_ACCOUNT_FUNDS`).

Every fragment is taken from the client's own rejection-message constants rather than from log samples, so it cannot drift against the wording.

## Why this is safe

`ExceptionMapper.message_to_exception` returns a **list**, and `ExceptionWithMessage.__contains__` passes a fixture when its expected exception is among those matched. The mapping is additive: an entry can only widen what a message resolves to, never remove a label a passing fixture depends on.

## Verification

`test_nethermind_exception_mapper.py` locks the invariant that matters: the three frame labels must stay pairwise disjoint under substring and regex matching, so a fixture naming one fails when the client rejected for another. In particular never a bare `"frame"`, which would catch `"validation prefix frame reverted"`. Over 39 wordings, in both bare and wrapped forms, it asserts each maps to its own label and neither of the other two, and that no frame label leaks onto the mapper's other 36 messages. 114 assertions pass here; 78 fail against the branch point. `just static` is clean.

The decode fragments sit outside the invariant deliberately: they are generic decoder wordings with no frame context, so they also match unrelated decode failures. Given additive matching that only widens `TYPE_6_INVALID_FRAME_FORMAT`; narrowing them needs frame context in the decoder's messages, which is a client-side change.

## Left for maintainers: the catch-all itself

This does **not** narrow `TYPE_3_TX_WITH_FULL_BLOBS`. Its docstring describes one specific decode failure, but the pattern matches the wrapper around all of them, so a fixture expecting it passes even when the client rejected for an unrelated reason — a false green. Narrowing it **can** turn currently-green fixtures red, and I have not enumerated the full set it legitimately matches, so I left it alone and documented what it does. Happy to follow up if you would like it tightened.

## Notes

- Targets `devnets/frames/0` rather than `forks/amsterdam`, since the `TYPE_6_*` members exist only there. The mapper file is byte-identical on both branches, so this forward-ports cleanly.
- No fixtures are refilled and no spec behaviour changes; this is message mapping only.
- Of the run's remaining 13 failures, 12 were client-side and are fixed; 1 is unrelated to this file.

### DIFF
--- a/packages/testing/src/execution_testing/client_clis/clis/nethermind.py
+++ b/packages/testing/src/execution_testing/client_clis/clis/nethermind.py
@@ -422,8 +422,91 @@ class NethermindExceptionMapper(ExceptionMapper):
         TransactionException.INVALID_CHAINID: (
             r"InvalidTxChainId|Signature is invalid."
         ),
+        # `Transaction \d+ is not valid: <cause>` is the wrapper the client
+        # puts around any payload transaction that fails to RLP-decode, so
+        # this entry alone cannot tell one decode failure from another. It is
+        # kept broad because the mapper is additive: a fixture passes when its
+        # expected exception is among those matched, and the entries below
+        # supply the discriminating labels for the causes that have one.
         TransactionException.TYPE_3_TX_WITH_FULL_BLOBS: (
             r"Transaction \d+ is not valid"
+        ),
+        # Frame transaction static constraints: frame count, mode, flags,
+        # value placement, atomic batch, expiry verifier, signature entry
+        # structure, nonce and blob fields.
+        TransactionException.TYPE_6_INVALID_FRAME_FORMAT: (
+            r"frame transaction must contain between 1 and 64 frames"
+            r"|frame transaction sender must be set"
+            r"|frame mode must be DEFAULT, VERIFY, SENDER, or POST_TX"
+            r"|POST_TX frames must form a trailing suffix of the frame list"
+            r"|POST_TX frames are not enabled"
+            r"|frame flags must not use reserved bits"
+            r"|frame value is only allowed in SENDER mode"
+            r"|frames allowed to approve execution must target the sender"
+            r"|the last frame must not have the atomic batch flag set"
+            r"|the atomic batch flag must not be set on a VERIFY frame"
+            r"|the atomic batch flag must not be set on a POST_TX frame"
+            r"|an atomic batch frame must not be followed by a VERIFY frame"
+            r"|an atomic batch frame must not be followed by a POST_TX frame"
+            r"|frames belonging to an atomic batch must not carry approval"
+            r" scope"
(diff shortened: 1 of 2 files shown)


gurukamath (approved):
LGTM. @Marchhill could you re-point this PR to `eips/amsterdam/eip-8141`? We will essentially re-create `devnets/frames/0` from there before tagging the next devnet-0 release.

PS: Updated this. Please ignore

PR #3473, opened 2026-08-31
EIP-8198: make slot-duration changes schedule-driven

chugarchugarr:
This is an executable EL proof for the infrastructure goal of EIP-8198: remove the one-off slot-duration assumption so later reductions can be expressed through schedule data rather than new execution logic.

The Hegotá-facing behavior remains the existing 12s -> 10s reference implementation from `0e8cc7fa0c49f7abfddde9e1f14a40bc54fd407c`. Synthetic 10s -> 8s -> 6s entries are test data only: they exist to prove that the same production paths compose across multiple later duration eras. They are not a proposal to activate 8- or 6-second slots without the consensus-layer performance work EIP-8198 calls for.

What this changes:
- gas-limit scaling is derived from the parent/current execution-payload duration eras, so a duration boundary applies exactly once even when payloads are missed or withheld around it;
- base-fee responsiveness uses current slot duration / the pre-schedule base duration on every block, preserving intended wall-clock response across repeated eras;
- blob capacity, excess accounting, blob pricing, transaction data fees, and BLOBBASEFEE resolve the active duration era from the same schedule;
- no 8-second- or 6-second-specific execution branch is introduced. Future durations are schedule data only.

Proof matrix:
- existing 12s -> 10s behavior is preserved;
- synthetic 10s -> 8s -> 6s capacity scaling composes 60M -> 48M -> 36M without a new execution branch;
- the added 8s era uses the same slot-duration lookup and transition-ratio machinery;
- blob schedule parameters derive through 12s -> 10s -> 8s -> 6s using the same production helper;
- parent payload 7 slots before and child payload 7 slots after the 10s -> 6s boundary still transition correctly;
- an unscaled child is rejected;
- an ordinary later 6s block stays at 36M;
- base-fee response remains tied to the original 12-second wall-clock baseline;
- VM blob helpers read the active schedule rather than an initial-era constant snapshot.

## Validation

Exact upstream-adapted head: `739c87067d4d9b2677d017bdc90ad915e61c099e`

Current upstream Amsterdam base used for validation: `117c9240c86ececb57440cd3342b4b82688fec06`

Green exact validation run: https://github.com/chugarchugarr/execution-specs/actions/runs/33560067598

That run completed successfully through:
- exact pin, direct-parent, one-commit, and merge-base verification against the current upstream Amsterdam head;
- `just static`;
- `just spec-tools`;
- focused `packages/testing/src/execution_testing/evm_tools/tests/eip8198_quick_slots/test_additional_duration_era.py`;
- the full `packages/testing/src/execution_testing/evm_tools/tests/eip8198_quick_slots` suite;
- `git diff --check`;
- exact changed-file-set verification for the eight intended EIP-8198 files;
- immutable candidate-head verification.

The upstream commit since the previous validated base changed only `tests/shanghai/eip4895_withdrawals/test_withdrawals.py` to assert zero withdrawal post-state and did not touch any of the eight EIP-8198 candidate files. This refresh therefore preserves the exact eight candidate blobs while remaining one commit over current `forks/amsterdam`.

The review question is narrow: should EIP-8198's EL implementation be schedule-driven so the one-time Hegotá refactor actually makes later slot reductions configuration changes rather than new duration-specific execution logic?

### DIFF
--- a/packages/testing/src/execution_testing/evm_tools/tests/eip8198_quick_slots/__init__.py
+++ b/packages/testing/src/execution_testing/evm_tools/tests/eip8198_quick_slots/__init__.py
@@ -1,0 +1,1 @@
+"""Tests for EIP-8198 slot-duration independence."""
--- a/packages/testing/src/execution_testing/evm_tools/tests/eip8198_quick_slots/test_additional_duration_era.py
+++ b/packages/testing/src/execution_testing/evm_tools/tests/eip8198_quick_slots/test_additional_duration_era.py
@@ -1,0 +1,103 @@
+"""Extra synthetic era proving EIP-8198 logic is duration-agnostic."""
+
+from ethereum.forks.amsterdam.slot_timing import (
+    BlobScheduleParameters,
+    SlotDurationEntry,
+    get_blob_schedule,
+    get_slot_duration_ms,
+    get_transition_durations,
+    scale_transition_limit,
+)
+from ethereum_types.numeric import U64, Uint
+
+HEGOTA_EPOCH = U64(10)
+MID_TEST_EPOCH = U64(15)
+FUTURE_TEST_EPOCH = U64(20)
+SLOTS_PER_EPOCH = U64(32)
+
+SCHEDULE_12_10_8_6 = (
+    SlotDurationEntry(HEGOTA_EPOCH, Uint(10000)),
+    SlotDurationEntry(MID_TEST_EPOCH, Uint(8000)),
+    SlotDurationEntry(FUTURE_TEST_EPOCH, Uint(6000)),
+)
+
+
+def test_additional_8s_era_is_schedule_data_only() -> None:
+    """A 10 -> 8 -> 6 sequence uses the same duration lookup path."""
+    first_10s_slot = U64(HEGOTA_EPOCH * SLOTS_PER_EPOCH)
+    first_8s_slot = U64(MID_TEST_EPOCH * SLOTS_PER_EPOCH)
+    first_6s_slot = U64(FUTURE_TEST_EPOCH * SLOTS_PER_EPOCH)
+
+    assert get_slot_duration_ms(
+        U64(first_10s_slot - U64(1)), SCHEDULE_12_10_8_6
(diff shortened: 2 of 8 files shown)


chugarchugarr:
@barnabemonnot @CarlBeek — the canonical upstream-facing implementation is refreshed onto current `forks/amsterdam` and remains one commit / eight files over the exact base.

Current candidate: `739c87067d4d9b2677d017bdc90ad915e61c099e`
Current Amsterdam base: `117c9240c86ececb57440cd3342b4b82688fec06`
Green exact validation: https://github.com/chugarchugarr/execution-specs/actions/runs/33560067598

That run passed exact pin/direct-parent/one-commit/merge-base checks, `just static`, `just spec-tools`, the focused and full EIP-8198 suites, `git diff --check`, exact changed-file-set verification, and immutable head verification. The intervening upstream commit changed only the Shanghai withdrawal post-state test and did not touch the eight EIP-8198 candidate files.

The existing EIP-side `SLOT_SCHEDULE` discussion remains linked directly to this PR.

The review question remains narrow: should EIP-8198's EL implementation be schedule-driven so the one-time Hegotá refactor makes later slot reductions configuration changes rather than new duration-specific execution logic?

chugarchugarr:
@CarlBeek @barnabemonnot — requesting review on the narrow design question now that the implementation is current, one commit / eight files over Amsterdam, and exactly validated.

The corresponding EIP-side `SLOT_SCHEDULE` proposal remains linked directly back to this PR.

Current candidate: `739c87067d4d9b2677d017bdc90ad915e61c099e`
Amsterdam base: `117c9240c86ececb57440cd3342b4b82688fec06`
Validation: https://github.com/chugarchugarr/execution-specs/actions/runs/33560067598

The specific question: should EIP-8198's EL implementation be schedule-driven so the one-time refactor makes later slot-duration reductions configuration changes rather than new duration-specific execution logic?

LouisTsai-Csie:
Thanks for the contribution! Given the complexity of this specification, we'd prefer the EIP champions to handle the changes to ensure alignment with the research. Closing for now.

PR #3477, opened 2026-08-31
feat(tests): probe EIP-161 emptiness after same-tx precompile touches

spencer-tb:
### Description

Stacked on #3480.

Add a touch-then-probe family for EIP-161 emptiness, covering erigontech/erigon#23670: a client that reads its own touch bookkeeping as proof of account existence skips new-account charges and diverges. A zero-value CALL to an absent precompile runs its code without creating the account, and nothing in the suite exercised that sequence before.

- New `tests/spurious_dragon/eip161_state_trie_clearing/`:
  - `test_extcodehash_after_precompile_touch` (ConstantinopleFix+, all precompiles): EXTCODEHASH, EXTCODESIZE and BALANCE probes around a same-transaction touch, for absent and funded precompiles.
  - `test_call_new_account_charge_after_precompile_touch` and `test_selfdestruct_beneficiary_charge_after_precompile_touch` (SpuriousDragon+): measure the exact charge after a touch. The measured call forwards zero gas, so the callee runs only on the value stipend and always halts, keeping precompile costs out of the measurement. From Amsterdam the new-account charge is state gas and invisible to the GAS delta, so the expectation uses `execution_cost` and the state charge stays pinned through the sender balance in the state root.
- `test_bal_insufficient_balance_call_to_touched_precompile` (EIP-7928): RIPEMD-160 must appear in the BAL with empty changes when reached only by a value call that fails its balance check, the shape the fuzzer found.

### Related Issues or PRs

#3480
erigontech/erigon#23670
erigontech/erigon#23672

### Checklist


tests/spurious_dragon/eip161_state_trie_clearing/spec.py
```diff
@@ -0,0 +1,16 @@
...
+@dataclass(frozen=True)
+class ReferenceSpec:
+    """Defines the reference spec version and git path."""
+
+    git_path: str
+    version: str
+
+
+ref_spec_161 = ReferenceSpec(
+    "EIPS/eip-161.md", "96523ef4d76ca440f73f0403ddb5c9cb3b24dcae"
```
fselmo: This looks wrong?

```suggestion
    "EIPS/eip-161.md", "b746c239881f24996f0205855e44f9b2d3b92b6a"
```

tests/spurious_dragon/eip161_state_trie_clearing/test_touched_account_emptiness.py
```diff
@@ -0,0 +1,212 @@
...
+
+def touch_code(precompile: Address, touch: str) -> Bytecode:
+    """Return the same-transaction touch that precedes the probe."""
+    if touch == "none":
+        return Bytecode()
+    if touch == "zero_value_call":
+        return Op.POP(Op.CALL(gas=100_000, address=precompile))
+    # The value exceeds the caller's balance, so the transfer fails
+    # after the target access is charged.
+    return Op.POP(Op.CALL(gas=100_000, address=precompile, value=2**100))
```
fselmo: nit: personal preference but also bc it could be a footgun... I'd prefer these to be `if / elif` for all cases and to raise on the `else`. Can be convinced otherwise though 😅

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
```diff
@@ -3405,6 +3405,66 @@ def test_bal_cross_block_ripemd160_state_leak(
...
+                        BalNonceChange(block_access_index=1, post_nonce=1)
+                    ],
+                ),
+                caller: BalAccountExpectation.empty(),
+                ripemd160_addr: BalAccountExpectation.empty(),
+            }
+        ),
+    )
+
+    blockchain_test(
```
fselmo: nit: this can also be a `state_test` since just one tx 👀

fselmo (commented):
Sorry had one thing locally that I pushed as it was an exploration and revealed a possible blip I think. Basically, the fixture carried no trace of the charge and a client that skips it after a touch would pass this test. I pushed a second value call that actually creates the precompile, so the charge is paid and stays paid, and the sender balance in the state root proves it.. with the precompile pre-funded as a control, the sender pays exactly `NEW_ACCOUNT` less.

Left some comments as well, nothing big. Looks good otherwise!

Please review my commit and lmk!

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists.py
spencer-tb: Converted this one-transaction case to a state test in a8a4ce44d1.

tests/spurious_dragon/eip161_state_trie_clearing/spec.py
spencer-tb: Updated the EIP-161 reference hash in a8a4ce44d1.

tests/spurious_dragon/eip161_state_trie_clearing/test_touched_account_emptiness.py
spencer-tb: Made the touch dispatch exhaustive with explicit branches and a ValueError fallback in a8a4ce44d1.

spencer-tb:
Addressed all three review threads in a8a4ce44d1:

- updated the EIP-161 reference hash
- made the touch dispatch exhaustive with an explicit error for unknown cases
- converted the single-transaction BAL case to a state test

fselmo (approved):
lgtm! I added one more cleanup commit which was partly my mistake / related to my changes. ~~Review that and, if you're good with it, I'll leave the merge to you~~ [it was a change making the pins better only, I can merge]. Approving here 👍🏼

chfast:
Found by evmone-fuzz.

PR #3484, opened 2026-09-01
feat(tests,forks): tests BAL early validation; refactor system contracts

fselmo:
### Description

Reviewed and added tests around the early invalidation of blocks using the BAL against BAL reads w/ `ITEM_COST`. 

def
```
G_remaining >= R_remaining * 2000
```

This suggested check can lead to clients invalidating valid blocks if implemented as suggested. This PR aims to add tests around this to show that system contract reads are present in `R_remaining`, as defined by the EIP, but they don't actually pay any gas... so the `G_remaining` goes to `0` while there are still reads in `R_remaining`, leading to an improper check on reads that actually use up the `G_remaining`. This can lead to early invalidation of near-full to full blocks... The invalidation happens near the end but it happens nonetheless and we should cover this in our tests.

As far as I can tell no client is implementing this except for Nethermind and they seem to remove the system contract addresses from the `R_remaining`, which circumvents the problem. Something the EIP does not suggest and they seemingly did of their own accord. This means no client would currently fail these tests but this gap should nonetheless be covered.

**^^ The first commit in this PR covers these tests ^^**

---

**The next commits are refactoring.**

In order to cover this gap appropriately and future-proof this test, it's important to make sure that any system contracts that get added on to post-execution and contain reads get accounted for. This led to a cleaner redesign for the organization of system contracts in the testing framework that builds them in the style of our EIP mixins that were more recently introduced, and adds these properties to the fork more directly – rather than being hard-coded in the `Spec` class in `specs.py` which can be a bit too rigid.

**This refactor then led to a very natural DevEx gain in a covariant pytest marker for `with_all_system_contract_request_types` which was added in the last commit here**

Note: I'm quite happy to move this refactor to another PR but I did separate it into its own commit so the tests could be reviewed separately. I ran `hasher compare` on all tests affected by this contract refactor and they are identical hashes. I even audited the `hasher` tool to make sure this isn't a blind check since I've never audited the tool itself I don't think 😅 (all looks good there too).

---

extra: Added some updates to the test-writing skill that come up often enough

### Related Issues or PRs

https://github.com/ethereum/EIPs/pull/12277 created to change the EIP

### Checklist

### Cute Animal Picture

Screenshot 2026-09-03 at 14 05 21


.claude/commands/write-test.md
```diff
@@ -10,6 +10,16 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
 - Exception: use `blockchain_test` when the test needs more than one transaction (a `state_test` holds exactly one) or more than one block (e.g. transaction-ordering or fork-transition tests).
 - Anti-pattern: wrapping one transaction in a `Block` to reach `blockchain_test`. A `state_test` can assert the transaction's gas used and receipt logs (the tx's `expected_receipt=TransactionReceipt(cumulative_gas_used=...)`), reserve state gas (the tx's `state_gas_reservoir=`), and other block-header fields (`blockchain_test_header_verify=Header(...)`) without it.
 
+## Fail Loudly — Never Let the Goalposts Move
+
+A test that quietly stops exercising what it claims is worse than no test, and there are far too many tests to re-audit whenever a fork reprices something. Wire every premise so a change **fails** the test instead of silently retargeting it.
```
fselmo: note to self: this can prob be a bit more succinct / less text

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7002.py
```diff
@@ -807,3 +923,286 @@ def test_bal_7002_request_invalid(
...
+    blockchain_test: BlockchainTestFiller,
+    fork: Fork,
+    middle_fill: str,
+    block_gas_limit: int,
+) -> None:
+    """
+    A block whose gasless BAL contributors are all maxed out is valid at
+    any fill level up to its gas limit.
+
+    The arms bracket the leftover gas at which EIP-7928's early-rejection
```
fselmo: remove this about the heuristic or add a note with link to [the removal](https://github.com/ethereum/EIPs/pull/12277) (if it is agreed to and removed) for historic reasons... otherwise this is a dead reference.

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_eip7002.py
```diff
@@ -807,3 +923,286 @@ def test_bal_7002_request_invalid(
...
+)
+def test_bal_low_gas_tail_with_pending_queue_reads(
+    pre: Alloc,
+    blockchain_test: BlockchainTestFiller,
+    fork: Fork,
+    tail_length: int,
+) -> None:
+    """
+    A valid block whose every late transaction boundary sits below the gas
+    EIP-7928's early-rejection heuristic would demand.
```
fselmo: Same here, reference the removal of the heuristic in this test if it's removed. Otherwise, amend these tests to any new definition for this early invalidation line.

.claude/commands/write-test.md
raxhvl: Maybe something like:

## A good test starts with suspicion

After writing a test, _deliberately_ break the behavior it is meant to check and confirm that it fails for the expected reason. Then restore the code and verify that it passes again. (i.e green -> red -> green)

Example: if a test expects writing to a read-only file to fail, make it writable and verify writing succeeds. This confirms the test’s assertion evaluates the intended file (rather than a different file accidentally).

tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py
```diff
@@ -289,17 +286,17 @@ def bus_slot_changes() -> list:
         ],
         storage_changes=[
             BalStorageSlot(
-                slot=Spec8282.COUNT_STORAGE_SLOT,
-                slot_changes=bus_slot_changes(),
+                slot=FeeSystemContractRequest.count_slot,
```
gurukamath: Perhaps this needs to be tied to the concrete class which calls this function instead of `FeeSystemContractRequest` (just in case there is any deviation in the future)

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_system_call_reads.py
```diff
@@ -0,0 +1,251 @@
...
+            storage_changes=_drained_queue_changes(
+                request_class, system_call_index
+            ),
+        )
+
+    blockchain_test(
+        pre=pre,
+        blocks=[
+            # Nothing checks the hundreds of enqueue receipts.
+            Block(txs=enqueue_txs, include_receipts_in_output=False),
```
gurukamath, line 201: Perhaps pinning down the premise set up by block 1, by asserting `request_class.count_slot` has the expected value is useful here. To prevent silent drift from parameter or gas changes in the future - if the block 1 does not succeed in enqueuing everything it says it does (some of its txs OOG for example)

tests/prague/eip7685_general_purpose_el_requests/test_multi_type_requests.py
```diff
@@ -52,7 +50,7 @@
     BuilderExitRequest,
 ]
 REQUEST_TYPE_BY_ADDRESS = {
-    rt.interaction_contract_address: rt for rt in REQUEST_TYPES
+    rt.system_contract_address: rt for rt in REQUEST_TYPES
```
gurukamath, line 53: Can we not get rid of the hard-coded `REQUEST_TYPES` and use `fork.system_contract_request_types()` instead? Could even be deferred to a follow-up PR

gurukamath (approved):
Looks good to me. Couple of small nits.

tests/amsterdam/eip7928_block_level_access_lists/test_fork_transition.py
fselmo: Love it, added it in the latest commit 👍🏼

tests/amsterdam/eip7928_block_level_access_lists/test_block_access_lists_system_call_reads.py
fselmo: it's pinned in the call value here. I added a clarifying comment.

tests/prague/eip7685_general_purpose_el_requests/test_multi_type_requests.py
fselmo: I will leave this to a folow-up PR. Would you like to make this change? I can pick it up otherwise 👀 @gurukamath

.claude/commands/write-test.md
fselmo: Hey @raxhvl! I actually pulled the skills updates out of here and opened a more targeted PR in #3562. I basically revamped this with the new `Pinning Intent` section. Happy to get your eyes there if you fancy.

PR #3486, opened 2026-09-01
feat(tests,test-specs): EIP-7928 - reject non-minimally encoded BAL scalars

nerolation:
### Description

EIP-7928 types `StorageKey`, `StorageValue` and `Balance` as `uint256`, so `storage_changes` slot keys, `storage_reads` keys, `StorageChange.new_value` and `BalanceChange.post_balance` are RLP scalars. RLP scalars MUST be minimally encoded, giving a block access list exactly one valid encoding and making `block_access_list_hash` an unambiguous commitment. Nothing covered this.

`test_bal_invalid_non_minimal_scalar_encoding` re-encodes one of those four scalars with a redundant leading zero byte, parametrized over each field. The list's contents are unchanged, so the header still commits to the canonical encoding; `keccak256` of the delivered list therefore differs from `block_access_list_hash` and the block MUST be rejected with `INVALID_BLOCK_ACCESS_LIST`.

A client that reads these fields as raw byte strings and derives the header commitment from a re-encoding of the decoded list computes a matching hash and accepts the payload.

Framework changes to make this expressible:

* `encode_scalar_non_minimally` rewrites the list's encoding rather than its contents.
* `BlockAccessListExpectation.modify_rlp` overrides only the engine payload body, leaving the header commitment canonical. `.modify` changes contents and so also moves the header hash.
* `has_modifier`, so the post-t8n exception exemption covers encoding modifiers too.

---
@fselmo additions:

- `_scalar_leaf` is explicit per field with a raising `else`, and the encoding modifiers have unit tests (addresses @spencer-tb's comment).
- Coverage extended to `block_access_index` (uint32) and `nonce` (uint64), so every EIP-7928 integer field and decoder width is hit.
- New `header_commits_to` dimension. `canonical_rlp` is the original case: the header commits to the canonical encoding and catches lenient decoders that hash a re-encoding. `payload_rlp` commits the header to the delivered bytes (`BlockAccessList.with_rlp_override`, `override_rlp`) and catches lenient decoders that hash the bytes as received. Expected exception stays `INVALID_BLOCK_ACCESS_LIST`; a client that trips the block-hash gate first is a mapper concern.
- `Block.engine_payload_only_overrides` from #3501 now reports `modify_rlp`, so a plain `blockchain_test` fixture is refused for this payload-only path unless the test is `blockchain_test_engine_only`.
- Rebased onto `forks/amsterdam`.

### Related Issues or PRs

N/A.

### Checklist

* [x] Ran fast static checks to avoid CI fails: `just static`
* [x] PR title has the form `<type>(<area>): <title>`


packages/testing/src/execution_testing/test_types/block_access_list/modifiers.py
```diff
@@ -870,6 +887,49 @@ def transform(bal: BlockAccessList) -> BlockAccessList:
...
+        for account_change, element in zip(bal.root, elements, strict=True):
+            if account_change.address != address:
+                continue
+            try:
+                container, index = _scalar_leaf(element, field)
+            except IndexError:
+                raise ValueError(
+                    f"No {field} entry for {address} in the BAL"
+                ) from None
+            container[index] = b"\x00" + container[index].to_be_bytes()
```
spencer-tb: `_scalar_leaf` only indexes for three of the four fields. For `storage_read` it returns the reads list and index 0, so an account with no reads raises a bare `IndexError` from `container[index]` outside the `try` instead of the `ValueError` the other fields give.

```suggestion
            try:
                container, index = _scalar_leaf(element, field)
                canonical = container[index].to_be_bytes()
            except IndexError:
                raise ValueError(
                    f"No {field} entry for {address} in the BAL"
                ) from None
            container[index] = b"\x00" + canonical
```

spencer-tb (approved):
Can we change the target branch to `forks/amsterdam`? We will be using that moving forward :)

LGTM, one comment! Will leave to @fselmo to approve and merge

packages/testing/src/execution_testing/test_types/block_access_list/modifiers.py
fselmo: Resolved this yesterday while reviewing 👍🏼... my commits should cover this.

fselmo (approved):
@nerolation great tests! I addressed the comment from @spencer-tb, added cases for other unsigned integer fields, nonce and bal index, and I added variations on whether the header commits to the canonical or the payload bal so we can cover both paths here. I think this is fantastic additions for the framework as well. Thanks for submitting this! 🙏🏼

lgtm

edit: also changed the base branch to be `forks/amsterdam` since we need this in the base dev branch and rebased... and reconciled these changes with #3501 as well and updated the PR description with a Claude summary of my changes here.

PR #3487, opened 2026-09-01
feat(tests): pin frame-tx spent gas across a refill rollback

AnkushinDaniil:
## What

Adds two EIP-8141 frame-transaction state tests that pin the transaction's
spent gas when a cross-frame storage refill is undone by a rollback. Each
asserts every frame's `state_gas_used` **and** the transaction's
`cumulative_gas_used`, closing a gap the existing refill/rollback tests leave
open: they check the per-frame receipt but never the settled transaction total.

A frame's `SSTORE` returning a slot to its transaction-start value refills the
state charge to the frame that created it, lowering that frame's receipt. When
the refilling frame is later rolled back, the rollback must restore the creating
frame's attribution together with the state — and that restoration has to reach
the transaction's spent gas, which sums every frame's final receipt usage in
both dimensions.

## Tests

- `test_batch_unroll_refill_restores_spent_gas` — a pre-batch frame creates a
  slot; a frame inside an atomic batch returns it to its start value, refilling
  the creating frame's receipt; the batch's terminating frame reverts, unrolling
  the batch. The unroll restores the creating frame's full attribution and
  re-appends the batch frames' receipts with their state gas zeroed. A later
  frame reads the restored attribution back live via `FRAMEPARAM`, and the
  transaction's spent gas is settled from the restored receipts.
- `test_frame_revert_refill_restores_spent_gas` — an earlier frame creates a
  slot; a later frame returns it to its start value, refilling the creating
  frame's receipt, then reverts. The revert's rollback extends over the receipt
  edit, restoring the creating frame's attribution while the reverting frame's
  own receipt reports zero state gas.

## Details

- Target fork alias: `Bogota` (Amsterdam + EIP-8141).
- Reference spec version pinned in the fixtures: `3ceef8d37092c5b13a614746dbce5b1fc3c2dfab`.
- Filled locally with `uv run fill` (state, blockchain, and blockchain-engine
  formats); `ruff` and `mypy` clean. Every asserted gas value is derived from the
  fork's gas calculators, and both tests were verified green against the t8n
  reference.


gurukamath (approved):
LGTM! Just rebased on the latest `eips/amsterdam/eip-8141` and moved the new tests to `test_gas_settlement.py` since that module also pins settlement gas semantics. Waiting for CI before merge

Issue #3479, opened 2026-08-31
Test Release Tracker: `glamsterdam-devnet@v8.1.3`


spencer-tb:
### Overview

This tracker is for the `tests-glamsterdam-devnet@v8.1.3` release, based on top of `tests-glamsterdam-devnet@v8.1.2`.

The spec is identical to v8.1.2, no repricing or EIP changes, test changes only: the un-decodable BAL engine tests now expect an invalid payload, and more Amsterdam ported static tests are un-skipped.

- **Target release tag**: `tests-glamsterdam-devnet@v8.1.3`
- **EELS branch**: [`devnets/glamsterdam/8`](https://github.com/ethereum/execution-specs/tree/devnets/glamsterdam/8)
- **Related Trackers**:
  - #3452
  - #3311
  - #3295
  - #3167

#### EIP Updates

None. Spec unchanged from v8.1.2.

#### New Test Cases

- [x] EIP-7928: undecodable BAL engine payloads now expect `{status: INVALID}` instead of error `-32602`, matching [execution-apis#869](https://github.com/ethereum/execution-apis/pull/869)
  - #3463
- [x] Ported static Pt. 2c: the Amsterdam SSTORE, refund and misc tests are enhanced and un-skipped
  - #3321

### Status
- [x] All relevant PRs merged to `forks/amsterdam` or `devnets/glamsterdam/8`.
- [x] Fixtures generated and released: `gh workflow run release_fixtures.yaml --repo ethereum/execution-specs -f feature=glamsterdam-devnet -f version=v8.1.3 -f branch=devnets/glamsterdam/8`
- [x] [`hive-tests`](https://github.com/ethpandaops/hive-tests/tree/master/.github/workflows) updated to run the new fixtures.
- [x] [`hive-ui`](https://github.com/ethpandaops/hive-ui/blob/master/public/discovery.json) discovery updated to the latest `glamsterdam-devnet` workflow.
- [x] Linked the test fixture release tag within the issue and closed it.

jangko:
[glamsterdam-devnet@v8.1.3](https://github.com/ethereum/execution-specs/issues/3479#top)
`tests/ported_static/stTransactionTest/test_opcodes_transaction_init.py` generate fixtures with extra env fields for mismatched fork.
for example "tests/ported_static/stTransactionTest/test_opcodes_transaction_init.py::test_opcodes_transaction_init[fork_Istanbul-add-state_test]" has `env` like this:
```json
"env": {
[...]

PR #3488, opened 2026-09-01
fix(test-specs): fail loudly if a test sets env fields not supported by the target fork

spencer-tb:
### Description

Reported by jangko on #3479: the v8.1.3 `test_opcodes_transaction_init` fixtures for pre-London forks carry `currentBaseFee`, `currentExcessBlobGas` and `slotNumber` (and `currentRandom`, unflagged). The enhanced test pins one module-level `Environment` across every fork so the `BASEFEE`, `BLOBBASEFEE`, `SLOTNUM` and `PREVRANDAO` arms can assert exact values, and nothing in the fill pipeline strips fields the target fork ignores. It is the only ported test that both reaches back before London and pins post-fork fields.

- `Environment.check_fork_fields` fails the fill when an env sets a field the fork's header lacks, gated on the fork's `header_*_required` predicates, and runs after `set_fork_requirements` on the state test env, the genesis env and each block env, so no other test can reintroduce the leak. `Environment.for_fork` keeps only the fields the fork has, for a test that spans forks with one pinned context.
- The test now pins only the field each opcode case reads. Pre-merge the `0x44` arm pins the difficulty explicitly, where it previously matched only because the framework's difficulty default equals the randao pin.

Istanbul now fills with the classic five env fields only. Amsterdam keeps every pin and all arms stay green.

### Related Issues or PRs

#3479

### Checklist

### DIFF
--- a/packages/testing/src/execution_testing/specs/state.py
+++ b/packages/testing/src/execution_testing/specs/state.py
@@ -343,7 +343,7 @@ class StateTest(BaseTest):
         )
 
         env = self.env.set_fork_requirements(fork)
-        env = env.without_fork_ignored_fields(fork)
+        env.check_fork_fields(fork)
         tx = self.tx.with_gas_limit(
             max_gas_limit=env.gas_limit,
             transaction_gas_limit_cap=fork.transaction_gas_limit_cap(),
--- a/packages/testing/src/execution_testing/specs/state.py
+++ b/packages/testing/src/execution_testing/specs/state.py
@@ -343,6 +343,7 @@ class StateTest(BaseTest):
         )
 
         env = self.env.set_fork_requirements(fork)
+        env = env.without_fork_ignored_fields(fork)
         tx = self.tx.with_gas_limit(
             max_gas_limit=env.gas_limit,
             transaction_gas_limit_cap=fork.transaction_gas_limit_cap(),
--- a/packages/testing/src/execution_testing/specs/state.py
+++ b/packages/testing/src/execution_testing/specs/state.py
@@ -322,6 +322,13 @@ class StateTest(BaseTest):
 
     def generate_blockchain_test(self) -> BlockchainTest:
         """Generate a BlockchainTest fixture from this StateTest fixture."""
+        # Checked before the genesis derivation below, which asks the
+        # fork for blob constants a fork without blobs does not have.
+        self.env.check_fork_fields(
+            self.fork.fork_at(
+                block_number=self.env.number, timestamp=self.env.timestamp
+            )
+        )
         return BlockchainTest.from_test(
             base_test=self,
             genesis_environment=self._generate_blockchain_genesis_environment(),
--- a/packages/testing/src/execution_testing/test_types/block_types.py
+++ b/packages/testing/src/execution_testing/test_types/block_types.py
@@ -214,6 +214,33 @@ class Environment(EnvironmentGeneric[ZeroPaddedHexNumber]):
 
         return self.copy(**updated_values)
 
+    def without_fork_ignored_fields(self, fork: Fork) -> "Environment":
+        """
+        Return a copy with the fields the fork's header lacks unset.
+
+        A test may pin fields for the newest forks it runs on. Older
+        forks ignore the pins, and serializing them into those forks'
+        fixtures would misstate the environment.
+        """
+        cleared: Dict[str, Any] = {}
+
+        if not fork.header_prev_randao_required():
+            cleared["prev_randao"] = None
+
+        if not fork.header_base_fee_required():
+            cleared["base_fee_per_gas"] = None
+
+        if not fork.header_excess_blob_gas_required():
+            cleared["excess_blob_gas"] = None
+
+        if not fork.header_blob_gas_used_required():
+            cleared["blob_gas_used"] = None
+
+        if not fork.header_slot_number_required():
+            cleared["slot_number"] = None
+
+        return self.copy(**cleared)
+
     def canonical_json(self) -> str:
         """
         Return the canonical JSON encoding of this model.
--- a/packages/testing/src/execution_testing/test_types/block_types.py
+++ b/packages/testing/src/execution_testing/test_types/block_types.py
@@ -33,15 +33,20 @@ FORK_GATED_FIELDS: Dict[str, Callable[[Fork], bool]] = {
     "parent_base_fee_per_gas": lambda fork: fork.header_base_fee_required(),
     "withdrawals": lambda fork: fork.header_withdrawals_required(),
     "excess_blob_gas": lambda fork: fork.header_excess_blob_gas_required(),
+    "parent_excess_blob_gas": (
+        lambda fork: fork.header_excess_blob_gas_required()
+    ),
     "blob_gas_used": lambda fork: fork.header_blob_gas_used_required(),
+    "parent_blob_gas_used": lambda fork: fork.header_blob_gas_used_required(),
     "parent_beacon_block_root": (
         lambda fork: fork.header_beacon_root_required()
     ),
     "slot_number": lambda fork: fork.header_slot_number_required(),
+    "parent_slot_number": lambda fork: fork.header_slot_number_required(),
 }
 """
-Environment fields that only some block headers carry, keyed by the
-fork predicate that admits each one.
+Environment fields, current and parent, that only some block headers
+carry, keyed by the fork predicate that admits each one.
 """
 
 
--- a/packages/testing/src/execution_testing/test_types/block_types.py
+++ b/packages/testing/src/execution_testing/test_types/block_types.py
@@ -3,7 +3,7 @@
 import json
 from dataclasses import dataclass
 from functools import cached_property
-from typing import Any, Dict, Generic, List, Sequence
+from typing import Any, Callable, Dict, Generic, List, Sequence
 
 import ethereum_rlp as eth_rlp
 from ethereum_types.numeric import Uint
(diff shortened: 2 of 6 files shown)


packages/testing/src/execution_testing/specs/state.py
```diff
@@ -343,6 +343,7 @@ def make_state_test_fixture(
         )
 
         env = self.env.set_fork_requirements(fork)
+        env = env.without_fork_ignored_fields(fork)
```
danceratopz: This covers the state fixture only. `generate_blockchain_test` still uses the raw `self.env`.

packages/testing/src/execution_testing/test_types/block_types.py
```diff
@@ -214,6 +214,33 @@ def set_fork_requirements(self, fork: Fork) -> "Environment":
 
         return self.copy(**updated_values)
 
+    def without_fork_ignored_fields(self, fork: Fork) -> "Environment":
+        """
+        Return a copy with the fields the fork's header lacks unset.
```
danceratopz: The summary promises all fields the fork's header lacks, but the method covers five. There is no clause for withdrawals or the beacon root (they cannot reach a state fixture env today, but they matter if this is ever reused on the blockchain path), and `parent_base_fee_per_gas` is missing (below). Maybe scope this to the header fields gated on the `header_*_required` predicates.

packages/testing/src/execution_testing/test_types/block_types.py
```diff
@@ -214,6 +214,33 @@ def set_fork_requirements(self, fork: Fork) -> "Environment":
...
+        forks ignore the pins, and serializing them into those forks'
+        fixtures would misstate the environment.
+        """
+        cleared: Dict[str, Any] = {}
+
+        if not fork.header_prev_randao_required():
+            cleared["prev_randao"] = None
+
+        if not fork.header_base_fee_required():
+            cleared["base_fee_per_gas"] = None
```
danceratopz: `parent_base_fee_per_gas` needs the same treatment. It is declared on `EnvironmentGeneric`, so unlike the other parent fields it reaches `FixtureEnvironment`: a test that explicitly sets it still gets `parentBaseFee` serialized into a pre-London state fixture after both methods run.

```suggestion
        if not fork.header_base_fee_required():
            cleared["base_fee_per_gas"] = None
            cleared["parent_base_fee_per_gas"] = None
```

packages/testing/src/execution_testing/test_types/block_types.py
```diff
@@ -214,6 +214,33 @@ def set_fork_requirements(self, fork: Fork) -> "Environment":
...
+
+        if not fork.header_excess_blob_gas_required():
+            cleared["excess_blob_gas"] = None
+
+        if not fork.header_blob_gas_used_required():
+            cleared["blob_gas_used"] = None
```
danceratopz: This clause cannot affect a state fixture: `blob_gas_used` is declared on `Environment` only, and `FixtureEnvironment` inherits `EnvironmentGeneric` with `extra="ignore"`, so it never gets there. It does remove the field from the t8n input, which may be worth keeping, but then the unit test should cover it; right now this is the one branch nothing exercises.

packages/testing/src/execution_testing/test_types/tests/test_environment_fork_fields.py
```diff
@@ -0,0 +1,51 @@
...
+from execution_testing.forks import Amsterdam, Fork, Istanbul, Paris
+
+from ..block_types import Environment
+
+PINNED = Environment(
+    prev_randao=0x20000,
+    base_fee_per_gas=10,
+    excess_blob_gas=0,
+    slot_number=7,
+)
```
danceratopz: `blob_gas_used` is the one cleared field this test never sets, so the `header_blob_gas_used_required` branch is untested. Setting it here and adding `currentBlobGasUsed` to the dropped/kept sets would cover it (it shows up in `Environment.model_dump`, just not in the fixture env).

packages/testing/src/execution_testing/test_types/tests/test_environment_fork_fields.py
```diff
@@ -0,0 +1,51 @@
...
+                "slotNumber",
+            },
+            id="istanbul",
+        ),
+        pytest.param(
+            Paris,
+            {"currentExcessBlobGas", "slotNumber"},
+            id="paris",
+        ),
+        pytest.param(Amsterdam, set(), id="amsterdam"),
```
danceratopz: A Cancun case would separate the last two clauses: it keeps `currentExcessBlobGas` while dropping `slotNumber`, which no current case does (Paris drops both, Amsterdam drops neither).

packages/testing/src/execution_testing/test_types/tests/test_environment_fork_fields.py
```diff
@@ -0,0 +1,51 @@
...
+    env = PINNED.set_fork_requirements(fork).without_fork_ignored_fields(fork)
+    keys = env.model_dump(mode="json", by_alias=True, exclude_none=True).keys()
+    assert dropped.isdisjoint(keys)
+    kept = {
+        "currentRandom",
+        "currentBaseFee",
+        "currentExcessBlobGas",
+        "slotNumber",
+    } - dropped
+    assert kept <= keys
```
danceratopz: This checks which keys survive but not the values. Asserting the kept entries still equal the values set in `PINNED` would also catch a copy that changes a value. Nit: `dropped: set` could be `set[str]`.

tests/ported_static/stTransactionTest/test_opcodes_transaction_init.py
```diff
@@ -123,15 +123,31 @@ def _jump_over_revert(conditional: bool) -> Bytecode:
...
-ENV = Environment(
-    fee_recipient=COINBASE,
-    number=BLOCK_NUMBER,
-    timestamp=BLOCK_TIMESTAMP,
-    prev_randao=PREV_RANDAO,
-    base_fee_per_gas=BASE_FEE_PER_GAS,
-    excess_blob_gas=EXCESS_BLOB_GAS,
-    slot_number=SLOT_NUMBER,
-)
+def _env(fork: Fork) -> Environment:
```
danceratopz: With the current framework behavior this helper is required because `generate_blockchain_test` uses the raw env, so this per-fork gating is what keeps the pre-London blockchain fixtures' genesis headers correct. 

If the framework grows a fork aware constructor or a loud fail (see review body), this helper shrinks to just the difficulty branch: substituting `difficulty` for `prev_randao` pre-merge is test semantics for the `0x44` opcode, not header hygiene, and no framework rule can infer it.

packages/testing/src/execution_testing/specs/state.py
marioevz: Maybe letting `set_fork_requirements` remove the fields is another alternative? But on the other hand it might be too breaking.

tests/ported_static/stTransactionTest/test_opcodes_transaction_init.py
marioevz: I'd side with a framework fix rather than having to make the test aware of the fields the header should or should not contain.

marioevz (commented):
Thanks for the fix, just a couple of comments.

danceratopz (commented):
Thanks for getting on this so quickly. Clanker verified the fix end to end: Istanbul fills of this test now carry exactly the five env fields a pre-London fixture should have (`currentCoinbase`, `currentDifficulty`, `currentGasLimit`, `currentNumber`, `currentTimestamp`), Amsterdam keeps every explicitly set field, and no other test in the tree explicitly sets any of the cleared fields on a fork that lacks them.

However, as-is, I think there's two issues:
1. The framework only silently removes the state fixture env fields, but not blockchain fixture fields: `generate_blockchain_test` and `FixtureHeader.genesis` still consume the raw env, so with the framework change alone the Istanbul genesis header still carries `baseFeePerGas`, and pure blockchain tests have the same exposure. Details inline on `state.py`. The fix only works in this PR because the test itself fixes it.
2. It misses `parent_base_fee_per_gas` (`parentBaseFee` still serializes).

Before we go into details, while I agree with @marioevz that we should avoid the duplication of field checks in the test and framework, I would prefer if the framework side failed loudly instead of removing these fields silently, mainly because I'm concerned that a test author's intent could be ignored. 

If a test explicitly sets an env field that the target fork does not support, I think that filling the test should error, with a message that points these three ways out:
  i. Fix/remove the environment field in the test.  
  ii. By using a new fork-aware constructor, for example `Environment.for_fork(fork, base_fee_per_gas=..., ...)`, that applies only the fields the fork supports, for the rare test that spans forks with one pinned context, or, 
  iii. An `rlp_modifier` on a blockchain test when the intent is to check that clients reject the field. 

The check should be cheap: It composes with `set_fork_requirements`, which only adds fields the fork supports, it must fire only on explicitly set non-None values, and for transition tests it has to be evaluated per use site (genesis env against the genesis fork, each block's env against that block's fork). It must also stay out of the header serialization layer so `rlp_modifier` based invalid header tests, for example the EIP-7843 slotnum pre-fork rejection tests, keep working.

spencer-tb:
@danceratopz @marioevz addressed these by adding loud framework validation for fork incompatible environment fields, and refactoring the opcode test to pin only the field each case reads, with an explicit `difficulty`/`prev_randao` split for opcode `0x44`! PTAL :)

packages/testing/src/execution_testing/test_types/block_types.py
```diff
@@ -27,6 +27,23 @@
...
+    "base_fee_per_gas": lambda fork: fork.header_base_fee_required(),
+    "parent_base_fee_per_gas": lambda fork: fork.header_base_fee_required(),
+    "withdrawals": lambda fork: fork.header_withdrawals_required(),
+    "excess_blob_gas": lambda fork: fork.header_excess_blob_gas_required(),
+    "blob_gas_used": lambda fork: fork.header_blob_gas_used_required(),
+    "parent_beacon_block_root": (
+        lambda fork: fork.header_beacon_root_required()
+    ),
+    "slot_number": lambda fork: fork.header_slot_number_required(),
+}
```
danceratopz, line 46: Not for this PR, perhaps ok as-is: `parent_excess_blob_gas`, `parent_blob_gas_used` and `parent_slot_number` are not gated. They cannot reach a fixture and t8n reads them only on forks that have them, so this is a consistency point only, but a test that sets them on an older fork states the same wrong intent as the fields listed here.

danceratopz (approved):
Thanks, yes, this is the direction I was hoping for. Clanker and I verified the new version. I updated the title to reflect the change in behavior.

We added 3 commits on top of your head to address a some nits / further improvements, PTAL:

1. `refactor(tests)`: the test now uses `Environment.for_fork` instead of the per-opcode table. This keeps the fixtures stable: London and later fixtures keep the pinned values they had in v8.1.3, where the table gave every test case that did not read a field the framework defaults instead (`currentBaseFee` `10` to `7`, `currentRandom` `0x20000` to `0`, `slotNumber` `7` to `0`), and pre-London fixtures change only by losing the fields their forks lack. The only fork logic left in the test is the difficulty branch for opcode `0x44`, which the framework cannot infer.
2. `fix(test-specs)`: `generate_blockchain_test` derived the genesis env before `BlockchainTest` ran the check, so a pre-Cancun `excess_blob_gas` failed on the blob constant assertion instead of the new message. The env is checked first now, with a unit test.
3. `test(test-types)`: the unit test asserts the message names `Environment.for_fork` and `rlp_modifier`.

PR #3478, opened 2026-08-31
feat(spec-specs, tests): repay spilled state gas from the reservoir when a child merges

spencer-tb:
### Description

Implement the merge time repayment specified in ethereum/EIPs#12265. A state charge spilled from `gas_left` can be refunded in a different frame, where the credit lands in the `state_gas_reservoir`. When a successful child merges, the reservoir now repays the merged frame's outstanding spill, capped at the spill (`repay_state_gas_spill`).

Flips the placement pins of `test_state_gas_cross_frame_refund.py` (#3490, #3498, #3499, #3506) to the repaid behaviour, adds an inner-frame probe and pins the receipt of the repaid execution. The repayment moves gas between the two pools settlement sums, so no other fixture changes and the settlement pins hold unchanged.

### Related Issues or PRs

Requires ethereum/EIPs#12256 & ethereum/EIPs#12265

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

PR #3490, opened 2026-09-01
feat(tests): pin cross-frame state gas refund placement and settlement

spencer-tb:
### Description

Add coverage for a state charge spilled from `gas_left` whose refund is credited in a different frame. Four tests pin the current placement and its settlement consequences:

- the credit lands in the reservoir and `gas_left` is not repaid at the merge
- settlement sums both pools, so the parked credit returns to the sender and the receipt carries no state term
- the parked credit funds a later creation that still bills the full state price, so the cross-frame route opens no discount
- the parked credit cannot fund execution work

The settlement pins are placement independent and hold unchanged under the merge-time repayment of ethereum/EIPs#12265. The placement pins flip with it: three of the four tests fail when filled against #3478.

### Related Issues or PRs

Coverage baseline for ethereum/EIPs#12265. #3478 revises the flipped pins if the repayment lands.

### Checklist

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
@@ -78,7 +78,7 @@ def test_cross_frame_refund_parks_in_reservoir(
     """
     clearer = pre.deploy_contract(code=Op.SSTORE(SLOT_X, 0))
 
-    call_window = Op.POP(Op.DELEGATECALL(gas=Op.GAS, address=clearer))
+    call_window = Op.POP(Op.DELEGATECALL(address=clearer))
     code = (
         # Warm the clearer and the slot while it is still zero, and
         # pre-expand the measurement memory, so the two measured
@@ -101,7 +101,7 @@ def test_cross_frame_refund_parks_in_reservoir(
 
     tx = Transaction(
         to=contract,
-        gas_limit=1_000_000,
+        state_gas_reservoir=0,
         sender=pre.fund_eoa(),
     )
 
@@ -183,7 +183,7 @@ def test_parked_credit_returns_at_settlement(
 
     tx = Transaction(
         to=contract,
-        gas_limit=1_000_000,
+        state_gas_reservoir=0,
         sender=pre.fund_eoa(),
         expected_receipt=TransactionReceipt(
             cumulative_gas_used=expected_gas_used
@@ -278,7 +278,7 @@ def test_parked_credit_funds_state_at_full_price(
 
     tx = Transaction(
         to=contract,
-        gas_limit=1_000_000,
+        state_gas_reservoir=0,
         sender=pre.fund_eoa(),
         expected_receipt=TransactionReceipt(
             cumulative_gas_used=expected_gas_used
@@ -337,9 +337,13 @@ def test_parked_credit_cannot_fund_execution(
             Op.DELEGATECALL(gas=Op.GAS, address=clearer, address_warm=False)
         )
     )
-    tail = Op.MSTORE.with_metadata(new_memory_size=192_032, old_memory_size=0)(
-        192_000, 1
+    # TODO: The tail spends a set amount of execution gas; a JUMPDEST
+    # run is the most future-proof inline way until a fork util exists.
+    tail_ops = min(
+        sstore_state_gas // Op.JUMPDEST.gas_cost(fork),
+        fork.max_code_size() - len(head),
     )
+    tail = Op.JUMPDEST * tail_ops
     code = head + tail
     contract = pre.deploy_contract(code=code)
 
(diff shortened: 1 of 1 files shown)


fselmo (approved):
These look good to me! I added another test after sniffing around a bit with Claude where it exercises two different paths on `credit_state_gas_refund` on two separate refund credits within the same frame.

@marioevz would be good if you wanted to take a pass here as well 👀

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
```diff
@@ -0,0 +1,524 @@
...
+
+    The frame sets a slot with the reservoir empty, spilling the state
+    charge from `gas_left`. A delegated child clears the slot and the
+    credit lands in the reservoir, where it stays through the merge:
+    `gas_left` is lower after the clearing call than before it, and
+    the clearing window costs the same as a no-op window.
+    """
+    clearer = pre.deploy_contract(code=Op.SSTORE(SLOT_X, 0))
+
+    call_window = Op.POP(Op.DELEGATECALL(gas=Op.GAS, address=clearer))
```
marioevz:
```suggestion
    call_window = Op.POP(Op.DELEGATECALL(address=clearer))
```

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
```diff
@@ -0,0 +1,524 @@
...
+        + window_cost_excess()
+        # Every other expected slot is zero, so the marker is what
+        # distinguishes the pinned run from a reverted one.
+        + Op.SSTORE(SLOT_MARKER, 1)
+    )
+    contract = pre.deploy_contract(code=code)
+
+    tx = Transaction(
+        to=contract,
+        gas_limit=1_000_000,
```
marioevz:
```suggestion
        state_gas_reservoir=0,
```

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
```diff
@@ -0,0 +1,524 @@
...
+    )
+    # The post-refund usage must clear the calldata floor, or the
+    # floor masks a lost or doubled credit.
+    assert expected_gas_used > fork.transaction_data_floor_cost_calculator()(
+        data=b""
+    )
+
+    tx = Transaction(
+        to=contract,
+        gas_limit=1_000_000,
```
marioevz:
```suggestion
        state_gas_reservoir=0,
```

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
```diff
@@ -0,0 +1,524 @@
...
+        + 2 * sstore_state_gas
+    )
+    restore_refund = clearer_code.refund(fork) - sstore_state_gas
+    expected_gas_used = before_refund - min(
+        before_refund // fork.max_refund_quotient(), restore_refund
+    )
+
+    tx = Transaction(
+        to=contract,
+        gas_limit=1_000_000,
```
marioevz:
```suggestion
        state_gas_reservoir=0,
```

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
```diff
@@ -0,0 +1,524 @@
...
+    head = (
+        fresh_set(SLOT_MARKER, 1)
+        + fresh_set(SLOT_X, 1)
+        + Op.POP(
+            Op.DELEGATECALL(gas=Op.GAS, address=clearer, address_warm=False)
+        )
+    )
+    tail = Op.MSTORE.with_metadata(new_memory_size=192_032, old_memory_size=0)(
+        192_000, 1
+    )
```
marioevz: This approach is a bit brittle and can break when we modify the CPSB. We can modify it or at the very least add a comment to clarify intent.
```suggestion
    # The purpose of this code is to consume a specific amount of gas, rather
    # than memory expansion.
    tail = Op.MSTORE.with_metadata(new_memory_size=192_032, old_memory_size=0)(
        192_000, 1
    )
```

marioevz (approved):
Just a couple of nits. Thank you!

tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
fselmo: I applied a better future-proof suggestion here for now but our conversation about a `fork` util is the best answer for this. I left a TODO that hints at using something like this if / when it becomes available instead.

PR #3192, opened 2026-07-18
refactor(amsterdam): Re-factor Amsterdam + Preparation for EIP-8141

gurukamath:
### Description

Note: This PR does not contain any EIP-8141 (Frame Transaction) implementation code but includes the minimal refactor aimed at accomplishing the following
1. Clean up the Amsterdam specs more broadly than what #3159 did.
2. Make it easier to implement EIP-8141 when needed

A series of behavior-preserving refactors that disentangle top-frame ("the
transaction") machinery from generic call-frame execution, as groundwork for
EIP-8141 frame transactions. Each commit fills
**byte-identically** on the full Amsterdam fixture set; ruff and mypy clean.

**What changes (8 commits, ~688/−624):**

- **Merge `Message` into `Evm`** — the two were mutable halves of one frame
  with aliased warm sets and double-written code. `Message`'s call parameters
  fold onto `Evm`; 
- **`check_transaction` returns a `TransactionEnvironment`** instead of an
  ad-hoc tuple.
- **Hoist top-frame preparation** (EIP-7702 delegation + dispatch resolution)
  out of `process_message` and up into `process_top_level`, so
  `process_call` only executes the code of whichever frame it is given.
- **Split top-level dispatch** into `create_evm` and the actual evm dispatch,
  keeping everything transaction-scoped (sender charging, settlement, receipts)
  *outside* the pair.
- **Move all static checks to `validate_transaction`** - There are currently some static checks in `check_transaction` which need to be moved
-  **`validate_transaction` is called in `check_transaction`** - Currently, `validate_transaction` is called in `process_transaction` and its outputs (tuple) is awkwardly wired into `check_transaction`. This can be avoided 

**Behavior:** none. Pure refactor — full Amsterdam suite (52,812 cases) fills
byte-identically.

### Related Issues or PRs

#3159 

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Cute Animals - 2 of 12

PR #3492, opened 2026-09-01
feat(spec-specs): expose state-independent transaction validation

danceratopz:
### Description

Transaction test consumers need to validate a raw transaction without constructing account or block state. Some checks that depend only on the transaction were still performed by the stateful `check_transaction` path, so callers could not use `validate_transaction` as a complete state-independent boundary.

This PR moves those checks into `validate_transaction` from Cancun through Amsterdam. It covers blob hashes, blob and set-code contract creation, empty authorization lists, and per-transaction blob limits. Amsterdam can now recover the sender when one is not supplied, giving callers the same one-argument validation entry point across forks.

State and blockchain validation continue through the same checks. Their existing exception order is preserved by invoking the common and type-specific validation stages at their original points, while stateless callers use the complete `validate_transaction` entry point.

### Stack

This is the first PR in a three-PR stack:

1. #3492 — Expose state-independent transaction validation (this PR).
2. #3493 — Resolve implicit transaction test gas limits.
3. #3494 — Validate transaction test fixtures with EELS.

### Related Issues or PRs

N/A.

### Testing

- Added 51 focused validation cases across Cancun, Prague, Osaka, BPO1–BPO5, and Amsterdam.
- Full Prague fill: 14,287 passed and 12 skipped.
- Full Osaka fill: 15,088 passed and 12 skipped.
- `just static`
- `just test-tests`

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) and [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` match the applied labels.

### Cute Animal Picture

A cat

### DIFF
--- a/src/ethereum/forks/bpo1/transactions.py
+++ b/src/ethereum/forks/bpo1/transactions.py
@@ -22,12 +22,20 @@ from ethereum.exceptions import (
 from ethereum.state import Address
 
 from .exceptions import (
+    BlobCountExceededError,
+    EmptyAuthorizationListError,
     InitCodeTooLargeError,
+    InvalidBlobVersionedHashError,
+    NoBlobDataError,
     PriorityFeeGreaterThanMaxFeeError,
     TransactionGasLimitExceededError,
+    TransactionTypeContractCreationError,
     TransactionTypeError,
 )
 from .fork_types import Authorization, VersionedHash
+
+VERSIONED_HASH_VERSION_KZG = b"\x01"
+"""Version byte that every blob versioned hash must start with."""
 
 
 @final
@@ -43,6 +51,9 @@ class IntrinsicGasCost:
 
 
 TX_MAX_GAS_LIMIT = Uint(16_777_216)
+
+BLOB_COUNT_LIMIT = 6
+"""Maximum number of blobs a single transaction may carry."""
 
 
 @final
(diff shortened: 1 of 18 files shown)


marioevz:
Tagging @gurukamath on this review since it might affect frames validation.

danceratopz:
> Tagging @gurukamath on this review since it might affect frames validation.

Thanks Mario! @gurukamath, this stack implements the "fill and validate transaction tests against eels in the json-loader" functionality that I alluded to in the context of #3156 and your review of #3491. I'm trying out stacked PRs to break down these changes into smaller, organized chunks. 

The final target PR is #3494: Once this PR #3492 is merged, apparently the Github UI will automatically rebase #3943 on `forks/amsterdam` and then change #3493's target branch to `forks/amsterdam`. And then the same update will happen to #3944 once #3943 is merged. I'm curious to see if it works out!

src/ethereum/forks/bpo1/transactions.py
```diff
@@ -540,8 +551,17 @@ def decode_transaction(tx: LegacyTransaction | Bytes) -> Transaction:
...
 def validate_transaction(tx: Transaction) -> IntrinsicGasCost:
+    """Validate all state-independent properties of a transaction."""
+    intrinsic = validate_transaction_common(tx)
+    if isinstance(tx, BlobTransaction):
+        validate_blob_transaction(tx)
+    validate_transaction_type(tx)
+    return intrinsic
+
+
+def validate_transaction_common(tx: Transaction) -> IntrinsicGasCost:
```
gurukamath, line 562: The function `validate_transaction_common` having a `if isinstance(tx, FeeMarketCapableTransaction)` gaurd feels a bit counter-intuitive. If we want to go down this route, we might have to go on and separate the validations for each type. Where tx_validation is common_validation + type 1 validation + type 2 validation + ...
Also, their `calculate_intrinsic` counterparts.

Unfortunately, we've found in the past that it never breaks down neatly in that way (EIP-8141 Frame Transaction being a notable exception). I think a single `validate_transaction` with `if` conditions still seems to be the right approach to me.

gurukamath (commented):
Related to #3192, #2245

The static transaction checks from `check_transaction` were moved to `validate_transaction` for `amsterdam` in that PR. This PR backports these changes to the earlier forks

PR #3496, opened 2026-09-01
chore(tests): clean up EIP-7928 spec.py ``Spec`` class

fselmo:
### Description

No spec-specific constants were useful to keep in ``Spec`` class for BALs. This may change but the import from the relevant spec is preferred over adding them here if they are not BALs specific.

### Related Issues or PRs

Relevant mention from [ETH R&D Discord](https://discord.com/channels/595666850260713488/753271902520213625/1544323210772746270)

### Checklist

### Cute Animal Picture

🐻‍❄️

### DIFF
--- a/tests/amsterdam/eip7928_block_level_access_lists/spec.py
+++ b/tests/amsterdam/eip7928_block_level_access_lists/spec.py
@@ -18,26 +18,4 @@ ref_spec_7928 = ReferenceSpec(
 
 
 class Spec:
-    """Constants and parameters from EIP-7928."""
-
-    # RLP encoding is used for block access list data structures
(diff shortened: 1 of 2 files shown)


marioevz (approved):
Nice, I think it's good to have the maximum nonce consolidated to a single source as much as possible, so this is a good change. Thanks!

fselmo:
> Nice, I think it's good to have the maximum nonce consolidated to a single source as much as possible, so this is a good change. Thanks!

Yeah agreed. Was a bit of oversight on my part in the last PR but in general that was the goal, to import from the true sources of truth 👍🏼

PR #3497, opened 2026-09-01
feat(test-tools): Implement `GasConsumer`

marioevz:
### Description

Implements a `GasConsumer` class whose purpose is to generate bytecode that consumes the exact amount of gas requested for the given fork.

The intention is twofold:
- Simplify boilerplate code that is used by many tests in order to generate bytecode that consumes a specific amount of gas.
- Make the intent of the test explicit and future-proof.

Currently, the PR only attempts to modify only non-ported-static tests since those will need to be refactored later anyway.

In the future, if the memory gas cost calculation method no longer allows for gas exhaustion like it's used by the `GasConsumer` today, it will be much easier to just fix a single class instead of going to each test that does this pattern and fix it individually. 

### Related Issues or PRs
N/A.

### Checklist

### Cute Animal Picture


packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -299,6 +299,130 @@ def __new__(
         return super().__new__(cls, bytecode)
 
 
+UNPAYABLE_GAS = 2**63
+"""Gas charge beyond any limit expressible in the 64 bits gas is kept in."""
```
fselmo, line 303: nit: I don't understand this docstring, can we clear it up a bit? 😅

packages/testing/src/execution_testing/tools/tests/test_gas_consumer.py
```diff
@@ -0,0 +1,144 @@
...
+
+@pytest.mark.parametrize("fork", get_forks(), ids=lambda fork: fork.name())
+def test_gas_consumer_out_of_gas(fork: Fork) -> None:
+    """
+    Verify the `None` sentinel produces bytecode that cannot be paid for.
+
+    The expansion is sized against the fork's own pricing, so the charge
+    stays out of reach however the schedule moves.
+    """
+    consumer = GasConsumer(gas=None, fork=fork)
```
fselmo, line 104: nit: I think this functionality is great, but `gas=None` is not the most friendly to DevEx imo. I would much prefer an explicit, more obviously readable method for the same behavior like:

```suggestion
    consumer = GasConsumer.out_of_gas(fork)
```

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -299,6 +299,130 @@ def __new__(
...
+            if memory_size > previous_memory_size:
+                code = _memory_expansion(memory_size, previous_memory_size)
+            code += Op.JUMPDEST * (gas - code.gas_cost(fork))
+            assert code.gas_cost(fork) == gas, (
+                f"generated bytecode consumes {code.gas_cost(fork)} gas "
+                f"instead of the requested {gas}"
+            )
+
+        instance = super().__new__(cls, code)
+        instance.gas = gas
```
fselmo, line 422: nit: would it be useful to store the new memory size the gas consumer expands to, since it burns gas via mem expansion? It's just an extra field at the moment but I'm sure it could be useful in certain test case scenarios. Some mem opcode could come after gas burning, for example, and use this value as its `old_memory_size`. Would just have to add this to the unit tests to make sure it works propertly but something to think about.

```suggestion
        instance.gas = gas
        instance.memory_size = memory_size
```

fselmo (approved):
Nothing blocking really. lgtm and seems really useful! Left some things to think about and Claude found these that could also use the `GasConsumer` if you want to look into it (also a nit though):

```
- tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py:901 — gas=None; pinned via header gas_used.                                                                                                                                                                                    
  - tests/amsterdam/eip7708_eth_transfer_logs/test_transfer_logs.py:1162 — gas=None; pinned via expected_receipt.                                                                                                                                                                                                           
  - tests/ported_static/stCreateTest/test_create_address_warm_after_fail.py:224 — memory sink that already threads old_memory_size=32, so convert to GasConsumer(gas=None, fork=fork, previous_memory_size=32). The function uses CodeGasMeasure, but confirm it covers the oog-post-constr branch. The existing assert     
    comparing the sink's cost to the remaining gas becomes trivially true with gas=None, so drop it or keep it as documentation. 
```

packages/testing/src/execution_testing/tools/tools_code/generators.py
LouisTsai-Csie: ^same question, and should this be 2^64 - 1 instead, since it fits within 64 bits?

packages/testing/src/execution_testing/tools/tools_code/generators.py
```diff
@@ -299,6 +299,130 @@ def __new__(
...
+        else:
+            if gas < 0:
+                raise ValueError(f"negative gas target: {gas}")
+            memory_size = _largest_expansion_within(
+                gas, fork, previous_memory_size
+            )
+            code = Bytecode()
+            if memory_size > previous_memory_size:
+                code = _memory_expansion(memory_size, previous_memory_size)
+            code += Op.JUMPDEST * (gas - code.gas_cost(fork))
```
LouisTsai-Csie, line 415:
```suggestion
            code += Op.JUMPDEST.gas_cost(fork) * (gas - code.gas_cost(fork))
```
small suggestion, would not be a problem since we have the assertion below.

LouisTsai-Csie (approved):

packages/testing/src/execution_testing/tools/tools_code/generators.py
LouisTsai-Csie: Added in commit b1c5be6

packages/testing/src/execution_testing/tools/tests/test_gas_consumer.py
LouisTsai-Csie: Added in commit b1c5be6

PR #3498, opened 2026-09-02
feat(tests): cover a cross-frame state gas refund with a funded reservoir

chfast:
### Description

Add a test where the sender pays for a reservoir that covers none, one or both of the parent's two fresh sets and the rest spill. A delegated child clears both slots and the refunded state gas lands in the reservoir. The call window, a later set and the receipt are the same in every case: the refund stays in the reservoir and the spill is not repaid.

With one set covered, the refund merged from the child is larger than the outstanding spill. The existing tests in this file all start with an empty reservoir, so the merged refund never exceeds the spill there. Covering both sets and covering none are the boundaries.

The expectations pin current placement. With the `src/` change of #3478 applied, `SLOT_RESULT` drops by the state cost of the spilled sets and `SLOT_PROBE_RESULT` rises by one state cost for `reservoir_slots=0` only. The receipt does not change.

### Related Issues or PRs

Follow-up to #3490. Coverage for ethereum/EIPs#12265 and #3478.

### Checklist

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
@@ -45,6 +45,8 @@ SLOT_Y = 2
 SLOT_MARKER = 3
 SLOT_RESULT = 4
 SLOT_INCREASED = 5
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
Rebased & approved! :)

PR #3499, opened 2026-09-02
feat(tests): cover a cross-frame state gas refund after a delegation spill

chfast:
### Description

Add a test where a set-code transaction with an empty reservoir pays its delegation from `gas_left` and commits that spill before the code runs. The code spills a fresh set and a delegated child clears it. The call window costs the same as without the delegation and the receipt bills the delegation.

### Related Issues or PRs

Follow-up to #3490. Coverage for ethereum/EIPs#12265 and #3478.

### Checklist


spencer-tb (approved):
LGTM! Thanks for the extra coverage :)

PR #3500, opened 2026-09-02
feat(tests): wrap earlier transaction types into frames where possible

gurukamath:
## 🗒️ Description

Refill every legacy state test with a second, frame-transaction variant. A
tree-level `tests/conftest.py` parametrizes each state test with a `tx_shape`,
and the `frame_tx` shape rewrites the test's transaction into the canonical
EIP-8141 pair — a `VERIFY` frame approving execution and payment against the
sender's default code, followed by a `SENDER` frame performing the original
call — leaving the test's own post-state assertions as the oracle. Variants
are generated only at forks with transaction type 6, land under a separate
`for_<fork>_at_frame_tx/` fixture subtree, and leave pre-Bogota fills
byte-identical. The `SENDER` frame's budgets are derived per test (execution
maximized under the EIP-7825 cap, state fitted under the block gas limit);
gas-usage pins (receipt/header) are stripped for the variant, since the
fixture's post-state root still captures gas through the fee balances.

### Waterfall (full tree, `--fork Bogota -m "state_test and not slow"`)

| | tests | why no frame variant |
|---|---:|---|
| **State tests collected at Bogota** | **14,090** | |
| not attempted | ~300 | EIP-8141's own tests, exception tests, and the `frame_tx_incompatible` opt-out |
| explicit `gas_limit` | 4,830 | an exact transaction gas limit is part of the scenario; frame transactions have only per-frame budgets with different costs |
| contract creation (`to=None`) | 2,090 | frame transactions cannot deploy at the top level; an empty frame target resolves to the sender |
| type-4 transactions | 956 | authorization lists have no frame equivalent |
| non-empty access list | 596 | frame transactions carry no access list |
| type-1 transactions | 506 | the type is defined by its access list |
| `state_gas_reservoir` set | 392 | a legacy spill/refill mechanism replaced by fixed, non-spilling per-frame budgets |
| 7702-delegated sender | 14 | the `VERIFY` frame dispatches the delegation's code, not the default approve code, so the wrapper never obtains approval |
| expected transaction error | 2 | validity failures follow a different taxonomy under frame transactions |
| **Converted and filled green** | **4,402** | **0 failures** |

`test_all_opcodes` opts out via the new
`frame_tx_incompatible` marker — its invalid-opcode expectation for TXPARAM
(`0xB0`) only holds outside a frame transaction. The `tests/ported_static`
skip-list normalizer learns to strip the new `-non_frame_tx`/`-frame_tx` id
segments.

Note for reviewers: at Bogota the state-test fill grows by ~31% (the frame
variants); fills for all other forks are unchanged, byte-for-byte.

### Checklist

### Cute Animal Picture

Cute Animals - 1 of 1


tests/conftest.py
```diff
@@ -0,0 +1,405 @@
...
+)
+
+FRAME_TRANSACTION_TYPE = 6
+"""
+Transaction type of the EIP-8141 frame transaction; its presence in a
+fork's transaction types gates the ``frame_tx`` variant.
+"""
+
+
+def has_frame_transactions(fork: Any) -> bool:
```
marioevz:
```suggestion
def has_frame_transactions(fork: Fork | TransitionFork) -> bool:
```
This is deliberate so the test has to know what to do when it receives a transition fork.

tests/conftest.py
```diff
@@ -0,0 +1,405 @@
...
+
+
+@pytest.fixture
+def tx_shape() -> str:
+    """Fill the original transaction shape when unparametrized."""
+    return "non_frame"
+
+
+@pytest.fixture
+def state_test(state_test: Any, tx_shape: str, fork: Fork) -> Any:
```
marioevz: We recently implemented a mechanism to define multiple variants of the same test in #3337. The parametrization is handled by the `filler.py`, and then the spec simply reads the current variant and modifies its behavior to generate the new variant of the same test.

Here's an example implementation that was done for the FOCIL variant of the inclusion tests:
https://github.com/ethereum/execution-specs/blob/a7b894b8b1882c1ae83ba2a5f690cb9652712c86/packages/testing/src/execution_testing/specs/blockchain.py#L732-L753

In `supports_fork` we can add the logic to check if frame-txs are supported by the fork, and in `discard_fixture_format_by_marks` we could check for the `frame_tx_incompatible`.

marioevz (commented):
This is a nice new feature! I just gave a single suggestion to reuse a mechanism we already have in place to automatically parametrize the same test in multiple variants, which seems to be what we want to do here. Thanks!

tests/conftest.py
gurukamath: Updated

spencer-tb (approved):
Couple of edge cases here!

The converted frame transaction doesn’t inherit `StateTest.chain_id`, causing tests with non-default chain IDs to produce invalid transactions.

The conversion raises block gas limits below 120M. Since GASLIMIT is observable, this can change test semantics even when the original limit is sufficient.

I (well codex) opened https://github.com/gurukamath/execution-specs/pull/21 with one commit per fix! Lmk what you think :)

gurukamath:
> Couple of edge cases here!
> 
> The converted frame transaction doesn’t inherit `StateTest.chain_id`, causing tests with non-default chain IDs to produce invalid transactions.
> 
> The conversion raises block gas limits below 120M. Since GASLIMIT is observable, this can change test semantics even when the original limit is sufficient.


Thanks Spencer. Very nice suggestions.

PR #3505, opened 2026-09-02
fix(fill): warn and discard a partial opcode trace instead of reporting it

jochem-brouwer:
`debug_traceBlockByHash` returns one entry per transaction. When a single transaction's trace fails, its entry carries an error instead of a `result`, and the aggregation loop skipped it silently -- returning a well-formed tally that is short by whole transactions, with no exception and no log line. The caller could not tell 17 traced transactions from 18.

That surfaced while filling the compute benchmarks. test_identity_uncachable[size_32, 300M] failed with

    Target opcode IDENTITY count mismatch: expected ~1693322 (+-5.0%),
    got 1598624

The deficit is 94,698, which is exactly the calls one full-cap transaction makes: (16777216 - 15512 - 17) // 177. The block itself was fine -- all 18 transactions present, 299.998M of 300M gas used -- and the same case passes on re-run (verified four times). Nothing was wrong with the benchmark; one transaction's trace was lost and the shortfall was reported as if it were a real count.

It read as a benchmark bug for exactly as long as it took to notice the count was one whole transaction short. The test's own model is correct: across the ten gas values that passed, its predicted transaction counts match the fixtures exactly and its predicted gas matches to within the 128-143 gas per transaction that WhileGas necessarily strands.

So the aggregators now report how many entries contributed, and a tally that does not cover every transaction is discarded with a warning naming the shortfall. It is not fatal: a missing trace should not throw away a block that filled correctly, and the count can be recovered by re-running. The fixture is still written, it just carries no opcode count.

Struct-log aggregation had the same hole and gets the same treatment; an entry with no `structLogs` key is a failed trace, not an empty one.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

### DIFF
--- a/packages/testing/src/execution_testing/client_clis/client_backend.py
+++ b/packages/testing/src/execution_testing/client_clis/client_backend.py
@@ -13,7 +13,7 @@ info, block boundaries, and pre-alloc declarations flow unchanged.
 
 import re
 from pathlib import Path
-from typing import Any, ClassVar, Dict, List, Optional
+from typing import Any, ClassVar, Dict, List, Optional, Tuple
 
 from execution_testing.base_types import Bytes, Hash
 from execution_testing.exceptions import ExceptionBase, ExceptionMapper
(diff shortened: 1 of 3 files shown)


fselmo:
#3367 was flagged on a review and looks like that was merged. It cites the same `test_identity_uncachable` 300M case. Did #3367 fix the root cause? Does this address failures that #3367 doesn't already cover?

From a Claude review:

> This doesn't seem ideal in its current state because a client that ignores the JS tracer echoes struct logs, and those entries never count as covered. Every block is then discarded with advice to rerun, which can never help, so opcode verification is silently skipped for the whole session where it used to fail loudly. Falling back to the struct log path in that case would fix it. Also, a discarded count makes `BenchmarkTest` skip the opcode assertion, so a partial trace now passes fill with only a log warning. With #3367 in, partial traces should be rare, so failing when a test asserts an opcode count seems safer.

I don't have as much context on this so lmk if I missed the point 😅.

PR #3506, opened 2026-09-02
feat(tests): sweep the refund merge over every call and create opcode

chfast:
### Description

Add two tests pinning where a merged state gas refund lands, swept over `with_all_call_opcodes` and `with_all_create_opcodes`. A holder contract owns the cleared slot, so a create's initcode reaches it the same way a call's child does, and the parent's spill comes from setting the slot through the holder.

Each test measures a clearing window against a no-op window of identical code, so neither needs gas arithmetic. Today the refund stays in the reservoir: the two windows cost the same at a call, and at a create the second window's account creation charge draws on the parked refund, so the clearing window costs one slot's state gas more. With the `src/` change of #3478 applied, every arm flips to minus one slot's state cost.

Before this, `CREATE` and `CREATE2` never merged a refunding child into a parent holding a spill: a repayment applied only at `generic_call` passed the whole suite. `CALL`, `CALLCODE` and depth 2 did reach the path, in `test_state_gas_sstore` and `test_state_gas_set_code`, but nothing there discriminates. Both tests reach the holder through an intermediate frame, so the sweep pins depth 2 as well.

`STATICCALL` is excluded: a static child cannot write, so it can never refund.

### Related Issues or PRs

Follow-up to #3490 and #3499. Coverage for ethereum/EIPs#12265 and #3478.

### Checklist

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
@@ -22,6 +22,7 @@ Tests for [EIP-8037: State Creation Gas Cost Increase]
 import pytest
 from execution_testing import (
     Account,
+    Address,
     Alloc,
     AuthorizationTuple,
     Bytecode,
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
Awesome thanks! LGTM no issues from my end :)

PR #3507, opened 2026-09-02
fix(fill): re-anchor start_block under `--no-reset-between-tests`

jochem-brouwer:
## Problem

`--no-reset-between-tests` keeps the chain moving between fills, but the
session anchor never moves with it.

`_session_pre_run` captures `start_block` once, from `latest`, after the
global setup:

```python
@pytest.fixture(scope="session", autouse=True)
def _session_pre_run(...):
    ...
    # 4. Capture start block (head after global setup).
    start_block = eth_rpc.get_block_by_number("latest")
    client_backend.start_block = start_block
```

That is the only production write to `client_backend.start_block`. Every
test then chains its first block from it, in `make_stateful_fixture`.

Normally the invariant holds because `_reset_chain_between_tests` rewinds
the chain back to that block after each test. Under
`--no-reset-between-tests` the rewind is skipped and nothing takes its
place, so the anchor goes stale as soon as the first test builds a block.

The nonce side of the flag was already adapted — `worker_key` reads at
`latest` under the flag, with a comment explaining the inversion. The
anchor side was missed.

### Related Issues or PRs

Stacked on #3304, which introduces the `--no-reset-between-tests` flag this fixes. Until that PR merges, the diff here includes its commits; the change in this PR is the single commit on top, by @skylenet. See also https://github.com/jochem-brouwer/execution-specs/pull/6

### Checklist

### Cute Animal Picture


jochem-brouwer:
This builds on top of #3304 so that should first be merged.

danceratopz:
> This builds on top of #3304 so that should first be merged.

What do you think about adding that to the top of the PR description and moving this to draft?

jochem-brouwer:
@danceratopz that PR (https://github.com/ethereum/execution-specs/pull/3304) adds the `--no-rest-between-tests` flag which we slightly alter here.

I have drafted this PR because from https://github.com/jochem-brouwer/execution-specs/pull/6/commits we need to add two more commits 3d072ce8dbcd951f741ba4713625a95c83d7127e and e269bb4483917217c5a4af71a90a3e23a4204e3e. The latter somehow raises a merge conflict if I cherry-pick it. Will look into it.

PR #3508, opened 2026-09-02
fix(spec-specs): wipe pre-existing storage on contract creation from Cancun onward

spencer-tb:
#### Quick context

7610 says: reject contract creation if the destination already has storage.

Without 7610: [EIP-684](https://eips.ethereum.org/EIPS/eip-684) allows creation when nonce is zero and code is empty.

The Yellow Paper says to clear the old storage before running initcode -> [here](https://ethereum.github.io/yellowpaper/paper.pdf#page=11)
> “The account’s nonce is initially defined as one, the balance as the value passed, the storage as empty”

My PR makes contract creation clear the destinations old storage before initcode runs, following the Yellow Paper!

Why wasn't this found before?
> 7610 was masking it, creation was rejected before the faulty storage clearing path could run. Removing that rejection exposed the bug.

Why the Cancun boundary?
> Earlier fork trackers remember that an account’s entire storage was cleared, but Cancun onward only discarded writes made in the current transaction. Reads could still retrieve old storage. This fits Cancun's restricted SELFDESTRUCT behavior where deletion normally concerns contracts created in that same transaction.

#### More Context

Contract creation over an account with zero nonce, no code and non-empty storage kept the old storage from Cancun onward, while Frontier through Shanghai wipe it. The older trackers record wiped addresses in `storage_clears` so reads stop falling back to earlier block writes or the pre-state; the trackers from Cancun onward never had it, and `destroy_storage` only dropped same-transaction writes. The case became reachable in the spec when #3417 removed the EIP-7610 storage check from `account_deployable`.

Port `storage_clears` to the nine trackers from Cancun to Amsterdam so every fork follows the Yellow Paper, which sets the created account's storage to the empty trie, and make the Amsterdam BAL builder read cleared storage as zero when diffing later transactions. The same path covers the deletion of an empty account that still holds storage.

Add storage-only creation tests to `test_create_collision.py` for the creation transaction and the create opcodes. The initcode stores one more than it loads from a pre-seeded slot, so a wiped target stores 1, a retained one stores 2 and an aborted creation deploys nothing. These are the first fixtures that fail a client still enforcing EIP-7610. That EIP was never activated in any fork: ethereum/tests applied it retroactively in 2024 and clients and EELS followed, so the tests are valid from Frontier like the removal itself.

There is no way to test the removal without also deciding what happens to the old storage. A fixture pins the state and receipts roots, and even an initcode that overwrites every old slot pays different SSTORE gas for a wiped slot than for a retained one. This PR decides for the Yellow Paper wipe. Reading current client code (not verified against binaries): geth master has already dropped the check and keeps the old storage, Nethermind keeps it for a funded target and clears it for an EIP-161-dead one, and Erigon, Besu, Nimbus and EthereumJS clear it. Geth and Nethermind would fail these fixtures today.

The case is unreachable on any live chain. The 28 mainnet accounts of this shape date from before Spurious Dragon, their creator nonces are spent, and reaching one needs a keccak preimage. EIP-8253 (proposed for Hegotá) bumps their nonces so EIP-684 rejects the creation in every client, which ends the question. Until then the fixtures only decide what clients agree to for tests. Slots wiped by the creation that the transaction never touches do not appear in the BAL, since the pre-state cannot enumerate storage; the slot the initcode reads shows up as a storage read.

The new tests pass from Frontier to Amsterdam with the fix and fail on Cancun and Amsterdam without it. The restored pre-EIP-7610 Cancun fixtures from ethereum/legacytests#18 pass through `tests/json_loader` with the fix, where the Cancun cases failed before.

### Related Issues or PRs

Follow-up to #3417 and #3425. Related to ethereum/legacytests#17, ethereum/legacytests#18 and EIP-8253.

### Checklist

### DIFF
--- a/src/ethereum/forks/amsterdam/block_access_lists.py
+++ b/src/ethereum/forks/amsterdam/block_access_lists.py
@@ -599,19 +599,22 @@ def _get_pre_tx_account(
 
 
 def _get_pre_tx_storage(
-    pre_tx_storage: Dict[Address, Dict[Bytes32, U256]],
-    pre_state: PreState,
+    block_state: BlockState,
     address: Address,
     key: Bytes32,
 ) -> U256:
     """
     Look up a storage value in cumulative state, falling back to `pre_state`.
 
-    Returns `0` if not set.
+    Returns `0` if not set, or if the storage at `address` was wiped
+    earlier in the block.
     """
-    if address in pre_tx_storage and key in pre_tx_storage[address]:
-        return pre_tx_storage[address][key]
-    return pre_state.get_storage(address, key)
+    if address in block_state.storage_writes:
+        if key in block_state.storage_writes[address]:
+            return block_state.storage_writes[address][key]
+    if address in block_state.storage_clears:
+        return U256(0)
+    return block_state.pre_state.get_storage(address, key)
 
 
 def update_builder_from_tx(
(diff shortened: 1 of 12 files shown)


fselmo (approved):
lgtm! Thanks. Added tests in a PR to this branch if you want to use that as a base / merge before merging this but nothing blocking from my end here.

PR #3510, opened 2026-09-02
chore(test-specs): avoid silent mistakes when writing tests

fselmo:
### Description

On a review of #3486, I asked one of the Claude agents to sweep the related framework for gotchas that could lead to silent `fill` for the wrong reasons. It came back with some hits but fell our of scope for that PR, so I'm adding these here with unit tests. This is also important for DevEx but it's mostly important as a security so we don't end up with footguns that can  `fill` tests but end up with unintended JSON fixtures because of silent overrides of certain things the test itself didn't mean to create.

This shouldn't change anything related to the current state of the tests but I did run `hasher compare` on a fill of eip7928 module at the current `forks/amsterdam` and on this branch, just as a sanity check, and it's exact.

### Checklist

### Cute Animal Picture

Screenshot 2026-09-02 at 13 39 58

### DIFF
--- a/packages/testing/src/execution_testing/test_types/block_access_list/t8n.py
+++ b/packages/testing/src/execution_testing/test_types/block_access_list/t8n.py
@@ -160,7 +160,7 @@ class BlockAccessList(EthereumTestRootModel[List[BalAccountChange]]):
         """Return the list for RLP encoding per EIP-7928."""
         return to_serializable_element(self.root)
 
-    def with_rlp_override(self, rlp: Bytes) -> "BlockAccessList":
+    def with_rlp_override(self, rlp: bytes) -> "BlockAccessList":
         """
         Return a BAL with the same contents whose serialization is ``rlp``,
         mirroring ``RLPSerializable.rlp_override``.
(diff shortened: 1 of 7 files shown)


packages/testing/src/execution_testing/test_types/block_access_list/t8n.py
```diff
@@ -160,18 +160,29 @@ def to_list(self) -> List[Any]:
         """Return the list for RLP encoding per EIP-7928."""
         return to_serializable_element(self.root)
 
-    def with_rlp_override(self, rlp: Bytes) -> "BlockAccessList":
+    def with_rlp_override(self, rlp: bytes | str) -> "BlockAccessList":
```
marioevz:
```suggestion
    @validate_call
    def with_rlp_override(self, rlp: Bytes) -> Self:
```
This uses pydantic to coerce automatically. The downside is that the static checker does not like `with_rlp_override(b"")`, but can be fixed with a `# type: ignore[arg-type]` until we figure out a way to tell pydantic mypy plugin which types can be coerced into `Bytes`.

marioevz (approved):
Awesome fixes! Just had one request. Approving for now so we can merge once implemented.

packages/testing/src/execution_testing/test_types/block_access_list/t8n.py
fselmo: done, ty!

PR #3512, opened 2026-09-03
refactor(tests): separate the gas window and receipt tests

chfast:
### Description

Four tests each asserted a gas window measuring where a merged refund sits and a receipt measuring what the sender pays. The window changes under the merge-time repayment of ethereum/EIPs#12265, the receipt must not. Asserted together, the receipt could only be read from a test that change makes fail.

Split each in two. The tests keeping the receipt drop the `GAS` stamps and the result slot, so one fewer slot survives and their sums stop carrying the probe overhead. With the `src/` change of #3478 applied, the failing tests are exactly the ones measuring a window, so the module docstring's claim that the two are independent is now checked.

Extract what the halves share as `clearing_child_code`, `clearing_child_storage` and `delegation_to`, and name the budget leaving a storing child more than the call stipend `budget_above_sstore_stipend`. It was open-coded ten times. The two `SSTORE` metadata forms spelled out eleven times become `FRESH_SET` and `WARM_CLEAR`, and `window_cost_excess` loses its metadata parameter now that no receipt sum reads a measured frame. Fixtures for every pre-existing test are unchanged.

### Related Issues or PRs

N/A.

### Checklist

### DIFF
--- a/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
+++ b/tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_cross_frame_refund.py
@@ -49,17 +49,25 @@ SLOT_INCREASED = 5
 SLOT_PROBE = 6
 SLOT_PROBE_RESULT = 7
 
+# A cold set of a slot that was zero when the transaction began, and
+# the warm clear that undoes it.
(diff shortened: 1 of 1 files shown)


chfast:
Mostly AI generated, so if you think to big / not worth review then fine.

spencer-tb (approved):
LGTM! Thanks Pawel+AI co :)

PR #3536, opened 2026-09-07
refactor(spec-specs): clarify the EIP-7981 data surcharge

danceratopz:
### Description

Make EIP-7981's flat access-list data surcharge explicit in the intrinsic and calldata-floor calculations. The testing-fork and test-local helpers now return the gas cost directly, and the related test wording uses the same terminology.

The gas schedule, test expectations, and reference-spec pin are unchanged. All 367 EIP-7981 EELS fill cases pass on `forks/amsterdam` and this branch; `hasher compare --tests` reports no fixture differences. `just static` passes.

### Related Issues or PRs

Related to #3217 and #1943. Independent test and checklist improvements are in #3537.

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Cat

Issue #3217, opened 2026-07-23
Test Release Tracker: `tests@v21.0.0` (Glamsterdam Testnet)


spencer-tb:
### Overview

Tracker for finalizing EELS specs and test coverage for all Amsterdam EL EIPs, targeting the first "mainnet-scheme" release that includes Amsterdam. Aiming to be ready for the first Glamsterdam testnet launch.

- **Target release tag**: `tests@v21.0.0`
- **EELS branch**: [`forks/amsterdam`](https://github.com/ethereum/execution-specs/tree/forks/amsterdam)
- **Baseline**: `tests@v20.0.1` (mainnet) / `tests-glamsterdam-devnet@v8.0.0` (devnet)
- **Test release date**: TBD
- **Related Trackers**: 
  - #3167
  - #3147

### Scope

EL EIPs only. Requires the glamsterdam devnet 8 to be released so final repricing numbers and spec changes are in.

Each unchecked item is scoped as a standalone task. Effort: `[S]` small, `[M]` medium, `[L]` large. If using Claude set the Fable 5 effort accordingly.

### Per-EIP Status

Baseline items for every EIP (listed per section below): reconcile the stale implementation tracker, refresh the reference-spec pin and review the EIP diff since pin, review test coverage and add missing tests, complete the testing checklist (markers + not-applicable/external-coverage files), execute mode tests sufficient and passing on the most recent devnet with mainnet-marked tests added, and spec tests passing on every client via consume engine and consume enginex.

#### [EIP-2780](https://eips.ethereum.org/EIPS/eip-2780): Resource-based intrinsic transaction gas — #1940 @spencer-tb

- [x] `[S]` Reconcile tracker #1940: closed with implementation evidence and release follow-ups recorded. Closeout PR: #3515.
- [x] `[S]` Refresh ref-spec pin and review spec diff: ethereum/EIPs#12080 merged; #3515 pins `7243c92ba812437c64bae9fc6524ee269b29daa9`.
- [x] `[M]` Devnet-8 spec update implemented: #3205 closed.
- [x] `[M]` Tracker test case list covered: value moves (EOA/contract/7702/precompile/self/empty), creation transactions, call/opcode access, 7702 intrinsic and delegation-target access, and create-and-destroy SELFDESTRUCT. #2175 and subsequent suites; final audit and additions in #3515.
- [x] `[M]` Static test updates: #3051 closed; prior-fork fill groups and static checks pass on #3515's final head.
[...]

PR #3537, opened 2026-09-07
feat(tests): complete EIP-7981 coverage and testing checklist

danceratopz:
### Description

Strengthen EIP-7981 tests to check successful execution and exact gas charges.

- Every transaction built by the shared fixtures now carries an expected receipt with success status and the exact gas used, and a gas limit that also funds the top-frame authorization gas of type 4 transactions. A default surplus of 1000 gas shows that unused gas is returned; the exact gas limit test sets the surplus to zero.
- Add 4 tests covering the billing crossover, recipient and value interactions, the execution out-of-gas boundary, and refund and revert interactions.
- Cover both outcomes at the calldata floor boundary with exact upfront funding, cover duplicate storage keys within one entry, and record the floor exception when the floor check is what rejects a transaction.
- Realign checklist markers on 5 tests and document the coverage and N/A decisions in the checklist files.
- Correct the EIP reference pin to the blob hash of the revision that was already pinned by commit hash; the pinned content is unchanged.

Nimbus rejects a block that has transactions and a header gasUsed of zero before it validates any transaction, so it reports the intrinsic gas exception for the below-floor cases of the floor validation and exact-balance tests. Those cases pass on Nimbus once invalid-transaction blocks carry a nonzero header gasUsed, which is the subject of the follow-up #3599. This is kinda low prio. Currently exception mapping is disabled for nimbus (in the hive sim Dockerfiles), see #3603.

Validation at the current head: `just static` passes, and all 731 EIP-7981 fill cases pass with `--until Amsterdam --generate-all-formats`, including 184 EngineX fixtures and the activation transitions. The regenerated checklist covers all 15 applicable items with 254 items marked N/A and no unmarked tests. `consume enginex` against go-ethereum, Nethermind, ethrex, Erigon, Besu and Reth passes all 186 cases each; Nimbus passes 174 and fails the 12 below-floor cases named above until the follow-up lands. Devnet execution and `consume engine` remain open.

### Related Issues or PRs

- #3217
- #1943
- #3536

### Checklist

### Cute Animal Picture

image


tests/amsterdam/eip7981_increase_access_list_cost/coverage.md
danceratopz: Unsure whether we should add this file?

tests/amsterdam/eip7981_increase_access_list_cost/test_access_list_cost.py
```diff
@@ -105,6 +105,8 @@ def test_access_list_token_calculation(
     fork: Fork,
     pre: Alloc,
     tx: Transaction,
+    tx_expected_gas_used: int,
+    tx_gas_delta: int,
```
danceratopz: `tx_gas_delta` is currently always the default value of `0` for this test (defined in the local `conftest.py`. It was added here defensively in case of future parametrizations, can be removed if it's too much noise.

tests/amsterdam/eip7981_increase_access_list_cost/test_access_list_cost.py
```diff
@@ -134,6 +136,10 @@ def test_access_list_token_calculation(
     )
     assert actual_floor_cost == expected_floor_cost
 
+    tx = tx.copy(gas_limit=tx_expected_gas_used + max(tx_gas_delta, 1000))
```
danceratopz: The (arbitrary) `1000` buffer tests that clients charge exactly the expected amount, even when more gas is available. An exact-limit test also detects excessive execution costs, but cannot check that unused gas remains uncharged.

tests/amsterdam/eip7981_increase_access_list_cost/eip_checklist_not_applicable.txt
```diff
@@ -1,13 +1,30 @@
...
-transaction_type = EIP does not introduce a new transaction type
 block_header_field = EIP does not add any new block header fields
 block_body_field = EIP does not add any new block body fields
-gas_refunds_changes = EIP does not introduce any gas refund changes
+gas_refunds_changes = EIP does not change refund amounts, caps, or rollback rules; test_access_list_surcharge_with_refund separately covers interaction with the surcharged floor.
 blob_count_changes = EIP does not introduce any blob count changes
 execution_layer_request = EIP does not introduce an execution layer request
 new_transaction_validity_constraint = EIP modifies the existing intrinsic and floor gas validity constraints rather than introducing a new one
 block_level_constraint = EIP does not introduce a block-level validation constraint
+
```
LouisTsai-Csie: I think we still need to add the following scenarios:
- `transaction_type/test/intrinsic_validity/to`
- transaction_type/test/intrinsic_validity/value_non_zero_insufficient_balance
- transaction_type/test/intrinsic_validity/value_non_zero_sufficient_balance
- transaction_type/test/intrinsic_validity/value_zero_insufficient_balance

(Needs to check the definition again.)

tests/amsterdam/eip7981_increase_access_list_cost/conftest.py
```diff
@@ -16,11 +16,22 @@
...
+GAS_SURPLUS = 1000
+"""
+Gas added to the limit of valid fixture-built transactions.
+
+With the exact expected gas as the limit, a receipt equal to the limit
+cannot tell exact billing from consuming everything that was available.
+The surplus makes the receipt show that unused gas is returned as well.
+Boundary tests override it with zero through the `tx_gas_surplus` fixture.
+"""
+
```
LouisTsai-Csie: We could move this to tx_gas_surplus directly?

tests/amsterdam/eip7981_increase_access_list_cost/test_access_list_cost.py
LouisTsai-Csie, line 174: Test name and docstring indicate calldata floor domination, but pinning the delegated recipe in `conftest` means we can't guarantee that anymore. Like in this test, Intrinsic cost outperforms the floor. 

Not a coverage gap though; `at_crossover` and `with_refund` already cover the floor.

tests/amsterdam/eip7981_increase_access_list_cost/conftest.py
```diff
@@ -208,28 +222,97 @@ def tx_intrinsic_gas_cost_including_floor_data_cost(
...
+        contract_creation=contract_creating_tx,
+        authorizations=authorization_list or [],
+    )
+    floor_gas = fork.transaction_data_floor_cost_calculator()(
+        data=tx_data,
+        access_list=access_list,
+        contract_creation=contract_creating_tx,
+    )
+    return max(
+        tx_intrinsic_gas_cost_before_execution + top_frame_gas, floor_gas
```
LouisTsai-Csie: Not sure if this is a good idea, this assume no state gas is charged.

tests/amsterdam/eip7981_increase_access_list_cost/conftest.py
```diff
@@ -208,28 +222,97 @@ def tx_intrinsic_gas_cost_including_floor_data_cost(
...
+    The recipients built by the `to` fixture execute no gas-consuming code
+    and the transaction sends no value, so it uses its intrinsic gas plus
+    the top-frame gas of its authorizations, or the calldata floor if that
+    is higher. A test that targets other bytecode or sends value must set
+    its own receipt.
+    """
+    top_frame_gas = fork.transaction_top_frame_gas_calculator()(
+        contract_creation=contract_creating_tx,
+        authorizations=authorization_list or [],
+    )
```
LouisTsai-Csie, line 225: Should this be `transaction_top_frame_execution_gas` instead of `transaction_top_frame_gas_calculator`? As the comparison is `max(intrinsic + execution, floor)`. It is not breaking any tests since which no state gas is charged within the test suite.

tests/amsterdam/eip7981_increase_access_list_cost/conftest.py
```diff
@@ -16,11 +16,22 @@
...
 
+GAS_SURPLUS = 1000
+"""
+Gas added to the limit of valid fixture-built transactions.
+
+With the exact expected gas as the limit, a receipt equal to the limit
+cannot tell exact billing from consuming everything that was available.
+The surplus makes the receipt show that unused gas is returned as well.
+Boundary tests override it with zero through the `tx_gas_surplus` fixture.
+"""
```
LouisTsai-Csie: This is only used by `tx_gas_surplus`, I think we could remove this.

tests/amsterdam/eip7981_increase_access_list_cost/conftest.py
LouisTsai-Csie: This is parameterized but always `False`. Is there missing coverage here?

Second, `EIP2780.refund_types()` filters out `AUTHORIZATION_EXISTING_AUTHORITY`, so in Amsterdam there is no such refunds. The flag now only decides whether the authority account pre-exists (hence its new `creates_account` wiring). So should we rename `authorization_refund` here?

tests/amsterdam/eip7981_increase_access_list_cost/conftest.py
```diff
@@ -96,13 +108,15 @@ def authorization_list(
                 AuthorizationTuple(
                     signer=pre.fund_eoa(1 if authorization_refund else 0),
                     address=Address(1),
+                    creates_account=not authorization_refund,
                 )
             ]
         return None
```
LouisTsai-Csie: We could fold these two statements into:

```python
param = getattr(request, "param", None)
if param is None:
  if tx_type == 4:
    pass
```

tests/amsterdam/eip7981_increase_access_list_cost/eip_checklist_not_applicable.txt
```diff
@@ -1,13 +1,30 @@
...
+transaction_type/test/intrinsic_validity/max_fee = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
+transaction_type/test/intrinsic_validity/chain_id = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
+transaction_type/test/intrinsic_validity/nonce_minus_one = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
+transaction_type/test/intrinsic_validity/nonce_plus_one = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
+transaction_type/test/intrinsic_validity/nonce_exact = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
+transaction_type/test/intrinsic_validity/to = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
```
LouisTsai-Csie: I think this is covered by `test_contract_creation_with_access_list`?

tests/amsterdam/eip7981_increase_access_list_cost/eip_checklist_not_applicable.txt
```diff
@@ -1,13 +1,30 @@
...
+transaction_type/test/intrinsic_validity/value_non_zero_insufficient_balance = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
+transaction_type/test/intrinsic_validity/value_non_zero_sufficient_balance = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
+transaction_type/test/intrinsic_validity/value_zero_insufficient_balance = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
+transaction_type/test/intrinsic_validity/value_zero_sufficient_balance = EIP-7981 reprices existing access-list transactions; it does not introduce a transaction type or change this rule.
```
LouisTsai-Csie: I think we should still test out these cases!
- value_non_zero_insufficient_balance
- value_non_zero_sufficient_balance
- value_zero_insufficient_balance
- value_zero_sufficient_balance

LouisTsai-Csie (changes requested):
Thanks for the checklist PR. A few ideas:
- Two-dimensional gas: execution cost with access list dominates header cost.
- Combine `test_access_list_data_cost_with_execution` and `test_access_list_data_cost_at_crossover` as they have similar structure.

spencer-tb:
Follow up review PR: https://github.com/danceratopz/execution-specs/pull/61

Fire away! :)

LouisTsai-Csie (approved):
LGTM from my side

Issue #3532, opened 2026-09-04
Test Release Tracker: `frames-devnet@v0.3.0`


gurukamath:
### Overview

This tracker is for the fourth `frames-devnet-0` test release.

EIP-8141 frame transactions on top of the `glamsterdam-devnet-8` stack, rebased onto the current `forks/amsterdam` to pick up a change to EIP-8037's state gas semantics on child merge. The EIP-8141 spec and implementation are unchanged from v0.2.0. The release fills the full suite at `Bogota`.

- **Target release tag**: `tests-frames-devnet@v0.3.0`
- **EELS branch**: [`devnets/frames/0`](https://github.com/ethereum/execution-specs/tree/devnets/frames/0)
- **Test release date**: 04/09/26
- **Devnet launch date**: N/A
- **Test Case Ideas**: https://github.com/ethereum/execution-specs/discussions/3358
- **Related Trackers**: #2829, #3368, #3415, #3521

#### EIP Updates

EIP-8141 spec unchanged, still frozen at [`ethereum/EIPs@b75cbe6`](https://github.com/ethereum/EIPs/blob/b75cbe61150f09a44c38843be916417283d5b7bf/EIPS/eip-8141.md) (2026-09-01).
- [x] EIP-8037: a successful child merge repays state gas still spilled into execution gas from the reservoir, so a cross-frame refund landing in the reservoir no longer leaves the parent's `gas_left` short: #3478
- [x] EIP-8037: `CALL_VALUE` derived from `ACCOUNT_WRITE + CALL_STIPEND`, top-frame gas clarified: #3527
- [x] Frame transactions carry no reservoir, so the merge repayment applies only to legacy-shaped transactions; the frame interpreter needed no counterpart. The helper was re-pointed at the reservoir component introduced by #3396 during the rebase.

#### New Test Cases

- [x] EIP-8037: expanded cross-frame refund coverage: #3511
- [x] EIP-7778: block gas accounting and admission-gate coverage across every transaction type, with the generic type-6 envelope excluded per the #3047 selector precedent: #3520
- [x] Test framework: per-transaction `gas_used` receipt check derived from cumulative gas.

### Status
- [x] All relevant PRs merged to `eips/amsterdam/eip-8141` or `devnets/frames/0`.
- [x] `devnets/frames/0` updated to the `eips/amsterdam/eip-8141` tip.
[...]

PR #3539, opened 2026-09-07
feat(test-specs,tests): exercise the EIP-8141 expiry verifier install at the fork transition

ilitteri:
### Description

EIP-8141 installs the expiry verifier's runtime code at `0x…8141` when the fork activates, and leaves the account's nonce and balance as they were ([EIPs@b75cbe6](https://github.com/ethereum/EIPs/blob/b75cbe61150f09a44c38843be916417283d5b7bf/EIPS/eip-8141.md), "Expiry verifier contract"; `apply_fork` in the Amsterdam spec module on this branch). No fixture exercised that install: every EIP-8141 test starts at Bogota with the account already in the genesis allocation, so a client that installs it with any nonce, balance or timing passes the whole suite and can still diverge on the state root of the activation block.

Three things stood in the way of a transition fixture. There was no transition fork into Bogota. `make_genesis` seeds `transitions_to().pre_allocation_blockchain()`, so even a transition fixture would have carried the code at genesis. And nothing in the filler applies a fork's activation-time state change (`apply_fork` is only called by `ethereum_spec_tools.sync`), which no fork since the DAO fork has needed.

This PR adds `BaseFork.activation_code_installs()`, the runtime code a fork writes when it activates, keyed by address. `EIP8141` moves the verifier from `pre_allocation_blockchain` to this hook; state tests keep `pre_allocation`, having no activation to cross. `Alloc.with_installed_code` applies such installs changing only the code. The blockchain filler applies the genesis fork's installs at genesis and, in `generate_block_data`, the installs new to a block's fork relative to its parent's fork before the block executes. The hive `execute` genesis and the engine_x pre-alloc groups apply the genesis-side installs the same way. `AmsterdamToBogotaAtTime15k` is added; the consume ruleset picks it up on its own.

`test_fork_transition.py` adds two tests. A probe records `EXTCODESIZE` of the verifier per block, none before the fork and the code from the fork block on, with the account absent, funded by a pre-fork transfer, or already present with a nonce and balance that must survive the install. And a frame transaction carrying an expiry frame executes in the activation block itself. A pre-fork type-6 rejection test is not possible while Bogota shares Amsterdam's spec module.

Verification, on the committed tree:

- `tests/amsterdam/eip8141_frame_transactions` filled at Bogota with `--generate-all-formats` on the unmodified and on the modified tree: 943 fixtures, all hash-identical; the new tests add 21 (7 cases in 3 formats).
- `just static` passes; the forks, fork-marker, specs and fixtures unit tests pass (571).

### Related Issues or PRs

Related to #3532, the `tests-frames-devnet@v0.3.0` tracker; the existing EIP-8141 fixtures are unchanged from that release.

### Checklist

- [x] Ran fast static checks to avoid CI fails: `just static`

### DIFF
--- a/packages/testing/src/execution_testing/specs/blockchain.py
+++ b/packages/testing/src/execution_testing/specs/blockchain.py
@@ -987,11 +987,15 @@ class BlockchainTest(BaseTest):
             block_number=env.number - 1, timestamp=env.parent_timestamp
         )
         if fork != parent_fork:
-            inherited = parent_fork.activation_code_installs()
+            parent_installs = parent_fork.activation_code_installs()
+            inherited = {
+                Address(address): code
+                for address, code in parent_installs.items()
+            }
             new_installs = {
                 address: code
                 for address, code in fork.activation_code_installs().items()
-                if inherited.get(address) != code
+                if inherited.get(Address(address)) != code
             }
             if new_installs:
                 if isinstance(previous_alloc, LazyAlloc):
--- a/packages/testing/src/execution_testing/specs/blockchain.py
+++ b/packages/testing/src/execution_testing/specs/blockchain.py
@@ -894,6 +894,14 @@ class BlockchainTest(BaseTest):
                 ),
                 pre_alloc,
             )
+            # Code a fork installs at activation is part of genesis only when
+            # that fork is already active there. A transition test gets it at
+            # the fork block instead, in `generate_block_data`.
+            pre_alloc = pre_alloc.with_installed_code(
+                self.fork.fork_at(
+                    block_number=0, timestamp=0
+                ).activation_code_installs()
+            )
         if empty_accounts := pre_alloc.empty_accounts():
             raise Exception(f"Empty accounts in pre state: {empty_accounts}")
         if pre_alloc.state_commitment() is None:
@@ -967,6 +975,29 @@ class BlockchainTest(BaseTest):
                 raise Exception(
                     "test correctness: the transaction that produces an "
                     "exception must be the last transaction in the block"
+                )
+
+        # A fork activating at this block installs its code before the block
+        # executes. The parent's fork tells whether this is that block, and
+        # which installs are new rather than inherited from an earlier fork.
+        assert env.parent_timestamp is not None, (
+            "parent_timestamp is required to resolve the parent's fork"
+        )
+        parent_fork = self.fork.fork_at(
+            block_number=env.number - 1, timestamp=env.parent_timestamp
+        )
+        if fork != parent_fork:
+            inherited = parent_fork.activation_code_installs()
+            new_installs = {
+                address: code
+                for address, code in fork.activation_code_installs().items()
+                if inherited.get(address) != code
+            }
+            if new_installs:
+                if isinstance(previous_alloc, LazyAlloc):
+                    previous_alloc = previous_alloc.materialize()
+                previous_alloc = previous_alloc.with_installed_code(
+                    new_installs
                 )
 
         transition_tool_output = t8n.evaluate(
--- a/tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
+++ b/tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
@@ -1,0 +1,172 @@
+"""
+Tests for the EIP-8141 fork transition.
+
+The fork installs the expiry verifier's runtime code at
+`Spec.EXPIRY_VERIFIER` when it activates. Every other EIP-8141 test
+starts at a fork where the code is already in the genesis allocation, so
+these tests are the only ones that exercise the install itself: what the
+account looks like before the fork block, and what changes at it.
+"""
+
+import pytest
+from execution_testing import (
+    Account,
+    Alloc,
+    Block,
+    BlockchainTestFiller,
+    FrameReceipt,
+    Op,
+    Transaction,
+    TransactionReceipt,
+)
+
+from .helpers import expiry_frame, sender_frame, verify_frame
+from .spec import Spec, ref_spec_8141
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_8141.git_path
+REFERENCE_SPEC_VERSION = ref_spec_8141.version
+
+pytestmark = pytest.mark.valid_at_transition_to("EIP8141")
+
+FORK_TIMESTAMP = 15_000
+"""Timestamp at which the transition fork activates EIP-8141."""
+
+SLOT_EXECUTED = 0x01
+"""Storage slot used by target contracts to record execution."""
+
+
+@pytest.mark.pre_alloc_mutable
+@pytest.mark.parametrize(
+    "pre_fork_nonce,pre_fork_balance",
+    [
+        pytest.param(None, None, id="absent_before_fork"),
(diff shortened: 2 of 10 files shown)


AnkushinDaniil (approved):
Nice, this closes the exact gap that bit my follower on the activation block, the absent-before-fork case landing nonce 0 is the one that matters.

Two small things I was not sure about. The hook is described as a generic code install, so is it worth one case where the target already has storage before the fork, to pin that only the code changes and the storage survives? And since this installs just the verifier, would a line in the docstring saying it covers the verifier install rather than the full activation-block state help, so a green fixture is not read as the whole transition?

tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
```diff
@@ -0,0 +1,172 @@
...
+            to=Spec.EXPIRY_VERIFIER,
+            value=1 if transfer_before_fork else 0,
+        ),
+    ]
+    blocks = [
+        Block(timestamp=FORK_TIMESTAMP - 1, txs=pre_fork_txs),
+        Block(
+            timestamp=FORK_TIMESTAMP,
+            txs=[Transaction(sender=sender, to=probe)],
+        ),
```
gurukamath, line 158: Would it be worth explicitly asserting that the `EXPIRY_VERIFIER` code changes are not a part of the BAL here?

tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
```diff
@@ -0,0 +1,172 @@
+"""
+Tests for the EIP-8141 fork transition.
+
+The fork installs the expiry verifier's runtime code at
+`Spec.EXPIRY_VERIFIER` when it activates. Every other EIP-8141 test
+starts at a fork where the code is already in the genesis allocation, so
+these tests are the only ones that exercise the install itself: what the
+account looks like before the fork block, and what changes at it.
```
gurukamath, line 8: Currently a strict transition test from Amsterdam to Bogota is not possible since the 8141 implementation is done in the `amsterdam` fork package itself. As per our process, once the specs are deemed very stable (perhaps around the late testnet phase), we will create a dedicated `bogota` folder in the specs and move the EIP-8141 implementation there. The tests in this module work for now, only because they are not required to exercise any inner workings of pre-fork Amsterdam. This is a known limitation and the reason why we have not implemented any transition tests so far. Perhaps worth documenting this in the docstring so that no one tries to implement a truly evm exercising transition test until that point.

packages/testing/src/execution_testing/specs/blockchain.py
```diff
@@ -969,6 +977,29 @@ def generate_block_data(
...
+        parent_fork = self.fork.fork_at(
+            block_number=env.number - 1, timestamp=env.parent_timestamp
+        )
+        if fork != parent_fork:
+            inherited = parent_fork.activation_code_installs()
+            new_installs = {
+                address: code
+                for address, code in fork.activation_code_installs().items()
+                if inherited.get(address) != code
+            }
```
gurukamath:
```suggestion
            parent_installs = parent_fork.activation_code_installs()
            inherited = {
                Address(address): code
                for address, code in parent_installs.items()
            }
            new_installs = {
                address: code
                for address, code in fork.activation_code_installs().items()
                if inherited.get(Address(address)) != code
            }
```

Just to prevent any future mismatches between `int` and `Address`

tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
```diff
@@ -0,0 +1,172 @@
...
+    TransactionReceipt,
+)
+
+from .helpers import expiry_frame, sender_frame, verify_frame
+from .spec import Spec, ref_spec_8141
+
+REFERENCE_SPEC_GIT_PATH = ref_spec_8141.git_path
+REFERENCE_SPEC_VERSION = ref_spec_8141.version
+
+pytestmark = pytest.mark.valid_at_transition_to("EIP8141")
```
gurukamath:
```suggestion
pytestmark = pytest.mark.valid_at_transition_to("Bogota")
```

gurukamath (changes requested):
Looks good to me. Couple of nits.
Also, would be great to have this pointed at the `eips/amsterdam/eip-8141` branch. We can update the devnets branch from there.

tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
ilitteri: Good call. I added a BAL expectation to every block. In the pre-fork block and the frame tx block, the verifier shows up with no code changes; in the fork block and the one after it, it shows up with no changes at all, so it is only there because the probe read it. The install never appears, which is the point: it is not a block-level operation. These get checked against t8n during fill, so a filler that ever leaked the install into the BAL would fail here.

tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
ilitteri: Thanks, I did not know that was the reason there were no transition tests yet. I wrote it into the module docstring: EIP-8141 lives in the Amsterdam package, and Bogota is only a label, so t8n runs the pre-fork blocks with 8141 on too. The tests here get away with it because their pre-fork blocks are a plain transfer and a code size probe, which both forks agree on. I also spelled out that something like "a frame tx is rejected before the fork" cannot be tested until a real Bogota package exists, and that these tests only cover the verifier install, not the activation block as a whole.

packages/testing/src/execution_testing/specs/blockchain.py
ilitteri: Applied as you wrote it.

tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
ilitteri: Done, switched to Bogota.

tests/amsterdam/eip8141_frame_transactions/test_fork_transition.py
chugarchugarr: I think the limitation may be narrower than needing to wait for a physical bogota package.The thing that seems coupled today is transition semantics to source-package identity: one t8n backend resolves every block through the same EELS fork implementation, so the pre-fork side cannot actually be canonical Amsterdam while the post-fork side carries 8141.

Could the transition fixture instead compose two pinned transition oracles across the existing block-state boundary?

canonical Amsterdam t8n -> previous alloc/env/header -> activation_code_installs() -> current 8141 candidate t8n

The filler already carries the resulting alloc/state forward block-by-block, so the only thing crossing the boundary would be the canonical predecessor state plus parent environment/header—not Python module identity.

That would let us express the discriminating case we currently cannot:

N-1: type-6 Frame tx rejected under canonical Amsterdam
N: same valid Frame shape accepted after 8141 activation

while keeping the verifier install semantics in this PR exactly as they are.

Then when the real bogota package is eventually cut, the same vectors can be run against it and required to produce the same post-fork results. That would make the later package move a checked representation change rather than a prerequisite for transition correctness.Is there anything in the current t8n/filler boundary that prevents routing the two sides of a transition through different pinned backends like that?

gurukamath (approved):

PR #3541, opened 2026-09-07
refactor(tests): use `GasConsumer` for the remaining manual gas sinks

### LINKED ISSUE #3538: Refactor tests to use `gasConsumer` helper
Nothing blocking really. lgtm and seems really useful! Left some things to think about and Claude found these that could also use the `GasConsumer` if you want to look into it (also a nit though):

```
- tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_block_2d_gas_accounting.py:901 — gas=None; pinned via header gas_used.                                                                                                                                                                                    
  - tests/amsterdam/eip7708_eth_transfer_logs/test_transfer_logs.py:1162 — gas=None; pinned via expected_receipt.                                                                                                                                                                                                           
  - tests/ported_static/stCreateTest/test_create_address_warm_after_fail.py:224 — memory sink that already threads old_memory_size=32, so convert to GasConsumer(gas=None, fork=fork, previous_memory_size=32). The function uses CodeGasMeasure, but confirm it covers the oog-post-constr branch. The existing assert     
    comparing the sink's cost to the remaining gas becomes trivially true with gas=None, so drop it or keep it as documentation. 
```

_Originally posted by @fselmo in https://github.com/ethereum/execution-specs/pull/3497#pullrequestreview-5105080368_

skyyycodes:
### Description

`GasConsumer` landed in #3497, which converted the non ported static tests that were sizing a gas burn by hand. Three call sites were left behind and this picks them up.

Two of them, in `test_block_2d_gas_accounting` and `test_transfer_logs`, deployed `Op.MSTORE(2**256 - 1, ...)` to make a frame consume everything it was given. `GasConsumer.out_of_gas` says that outright and sizes the expansion against the fork's own memory pricing, so it no longer rests on one hard coded offset staying unpayable if those prices move.

`test_reverted_transaction_no_log` cannot build its case before the fork is known, so it moves to `parametrize_by_fork`. That reorders the segments of its test ids and puts the case id ahead of the fixture format suffix. #3497 already did the same to the two functions it converted in that file. No case is added or removed.

The third is `test_create_address_warm_after_fail`. Its `oog-post-constr` sink was sized to 2**12 bytes of memory and followed by an assert that the cost exceeded the gas left in the frame. The new sink is unpayable at any limit, which makes that assert trivially true, so it goes. On the question raised in the issue, `CodeGasMeasure` sits in the entry contract rather than in the callee, so it covers this branch the same way it covers the rest.

I left the memory bound tests under `stMemoryStressTest` and `stMemoryTest` alone. There the `MSTORE` offset is the subject of the test rather than an incidental way to burn gas.

Gas does not change anywhere.

### Related Issues or PRs

Fixes #3538.

### Checklist

### DIFF
--- a/tests/amsterdam/eip7708_eth_transfer_logs/test_transfer_logs.py
+++ b/tests/amsterdam/eip7708_eth_transfer_logs/test_transfer_logs.py
@@ -1153,15 +1153,22 @@ def test_contract_log_and_transfer_ordering(
     state_test(env=env, pre=pre, post=post, tx=tx)
 
 
-@pytest.mark.parametrize(
-    "reverting_code",
-    [
+def reverting_codes(fork: Fork) -> List[ParameterSet]:
+    """
+    Return codes that fail the transaction, one way per case.
+
+    The out-of-gas case is sized against the fork's memory pricing, so the
+    cases cannot be built before the fork is known.
+    """
+    return [
         pytest.param(Op.REVERT(0, 0), id="revert"),
         pytest.param(Op.INVALID, id="invalid_opcode"),
         pytest.param(Op.ADD, id="stack_underflow"),
-        pytest.param(Op.MSTORE(2**256 - 1, 0), id="out_of_gas"),
-    ],
-)
+        pytest.param(GasConsumer.out_of_gas(fork), id="out_of_gas"),
+    ]
+
+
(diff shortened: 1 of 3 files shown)


spencer-tb (approved):
LGTM! Thanks :)

Issue #3225, opened 2026-07-23
Glamsterdam: cross-EIP interaction test coverage around the BAL accessed-boundary (7928 × 7981 / 7702 / gas repricing)


axiosdevs:
Follow-up to #3213 (per-EIP BAL edge cases), from a different angle: **cross-EIP interactions**. The BAL "accessed → included" set (EIP-7928) is *downstream of gas accounting*, so a Glamsterdam EIP that changes gas can shift BAL contents even when each EIP's own tests pass. Per-EIP suites don't exercise these combinations. Proposing a small interaction matrix — happy to write the EEST vectors. I've tried to be precise about where the interaction is direct vs. indirect.

### Interaction scenarios

1. **EIP-7981 (access-list data surcharge) × BAL inclusion — indirect, via remaining gas.**
   7981 adds a `64 * access_list_bytes` surcharge to *intrinsic* gas; it does **not** change per-access opcode costs. But raising intrinsic gas lowers the gas available for execution, which can trigger a runtime OOG *earlier* than pre-7981 — changing which addresses are actually accessed, and therefore the BAL.
   *Test:* identical tx + declared access list + total gas; pre- vs. post-7981 the tx halts at a different point → assert the BALs differ exactly by the addresses no longer reached. Bonus assertion: declared-but-untouched access-list entries are *paid for* under 7981 yet MUST NOT appear in the BAL.

2. **EIP-7702 delegate resolution under tight gas.**
   Delegate resolution adds an access (see #3213). Combined with reduced remaining gas (from 7981 or repricing), the delegate target can fail its access-cost check / OOG before being loaded → MUST NOT appear. Test the boundary where the same delegated call includes vs. excludes the delegate purely because of the gas change.

3. **Glamsterdam gas repricing × BAL SSTORE gates.**
   BAL slot inclusion and no-op-write classification are gated on `access_cost` and `gas_left > GAS_CALL_STIPEND` (7928). If any SSTORE/SLOAD-adjacent costs are repriced in the fork, these boundaries move. *Property test:* for a slot write sitting exactly at the stipend boundary, inclusion flips consistently with the new schedule (and the slot appears nowhere when the gate fails).

4. **BAL size bound × large deployments (conditional on EIP-7954 landing).**
[...]

Issue #3213, opened 2026-07-23
EIP-7928 BAL: edge-case test scenarios + a few spec-clarity questions (Amsterdam)


axiosdevs:
Offering this as a contribution toward Amsterdam BAL (EIP-7928) test coverage. I cross-checked each item against the current EIP-7928 text. Framed as (a) concrete edge-case scenarios worth covering and (b) a few spec-clarity questions — happy to turn any accepted ones into EEST test vectors. Apologies in advance for any overlap with the existing checklist (I couldn't load it at time of writing).

### Edge-case scenarios worth covering
- **CREATE/CREATE2 inclusion boundary** — address computed but creation fails *before* the address is accessed → address MUST NOT appear; vs. collision / nonce-increment / post-access OOG → included. (Spec: "…not on address computation alone, nor when creation fails before access.")
- **No-op write cumulative tracking** — an `SSTORE` whose value equals the pre-value *as of the current `block_access_index`* is classified as a read; a later no-op MUST NOT remove earlier `storage_changes` for the same slot. Multi-tx same-slot sequences.
- **EIP-7702** — the authority address is included (nonce + code) after a successful set/update/clear; the delegation *target* MUST NOT appear during set/modify, only once it is loaded as an execution target.
- **Revert keeps accesses** — accessed-but-reverted addresses appear with empty change lists; read slots still appear in `storage_reads`.
- **Canonical-form negative tests** — reject a correctly-executing block whose BAL is mis-sorted (address / slot / intra-slot `block_access_index`), contains a duplicate, or lists a key in both `storage_changes` and `storage_reads`.
- **Precompiles** — included when invoked; with value → `balance_change`; otherwise empty change lists (no code/nonce entries).

### Spec-clarity questions
1. **Reverted-tx "final" sender balance** — the spec records the *final* sender balance/nonce after each tx. For a tx that reverts *before* the recipient is loaded, is the sender's fee-paying `balance_change` still emitted at that tx's index? One explicit sentence would pin implementations.
2. **Phantom-read check cadence** — the gas-feasibility check is worded "SHOULD … e.g. every 8 transactions". Is the cadence meant to be normative for interop, or implementation-defined? Divergent cadences could let one client reject a BAL another accepts.
[...]

PR #3544, opened 2026-09-07
fix(test-fill, tests): split duplicated data-floor checklist ID and cover both arms

axiosdevs:
### Description

The transaction-type section of the EIP checklist template has two rows for the calldata floor vs intrinsic gas boundary — the invalid `tx.gas_limit == tx.intrinsic_gas_cost` arm and the valid `tx.gas_limit == data_floor_cost(len(tx.data))` arm — but both carry the same ID, `transaction_type/test/intrinsic_validity/data_floor_above_intrinsic_gas_cost`. It is the only duplicated ID in the template.

The checklist generator keys template items by ID (`TEMPLATE_ITEMS[item.id] = item`), so the second row collapses into the first and can never be marked as covered or not applicable. That is why `uv run checklist` on `forks/amsterdam` shows this one item open in the EIP-7976, 7981, 7778, 7708 and 7928 checklists even though every one of them marks `transaction_type` as not applicable, and why the marker `EIPChecklist.TransactionType.Test.IntrinsicValidity.DataFloorAboveIntrinsicGasCost` is unused repo-wide.

Changes:

- Template: give each arm its own ID, nested under the existing prefix (`.../data_floor_above_intrinsic_gas_cost/gas_limit_intrinsic` and `.../gas_limit_floor`) so existing `eip_checklist_not_applicable.txt` entries keep matching by prefix.
- `EIPChecklist`: mirror the split as `DataFloorAboveIntrinsicGasCost.GasLimitIntrinsic` / `.GasLimitFloor`; stubs regenerated with `uv run generate_checklist_stubs`.
- Generator: assert on a repeated ID while parsing the template; unit test `test_checklist_template_ids_are_unique` (fails on the previous template).
- EIP-7976: `test_floor_boundary_gas_limit.py` pins both arms — reject at `gas_limit == intrinsic` (floor above it) with `INTRINSIC_GAS_BELOW_FLOOR_GAS_COST`, accept at `gas_limit == floor` and assert the receipt bills the floor. The existing validity matrix only exercises `floor - 1`, not the intrinsic point itself. The 7976 not-applicable file is narrowed so the two new items are counted instead of conflicting.

Verification: checklist unit tests pass; `uv run fill tests/amsterdam/eip7976_increase_calldata_floor_cost/test_floor_boundary_gas_limit.py --fork Amsterdam` fills; deliberately mis-setting the expected receipt fails the fill with `cumulative_gas_used mismatch`; regenerated EIP-7976 checklist reports 12/12 with no conflicting items; `ruff`, `codespell` and `mypy` clean on the touched files.

### Related Issues or PRs
Follow-up to the coverage discussion in #3213 / #3225, this time driven by the checklist as suggested there.

### Checklist

### Cute Animal Picture

cat


spencer-tb:
Thanks for the write up but both new tests are already covered, and this feels to AI one shot vibe coded! Closing :)

PR #3551, opened 2026-09-08
feat(tests): EIP-7928 - reject BALs that omit system-phase accesses

ilitteri:
### Description

Add five invalid-BAL tests for accesses made outside transaction execution. Every existing `test_bal_invalid_*` case corrupts an account a transaction touched, the coinbase, or a withdrawal recipient whose balance changed. None removes an access that leaves `state_root` and gas untouched, which is the case only regenerate-and-compare or an explicit completeness check can catch:

- `test_bal_invalid_missing_request_predeploy_entry` (EIP-7002, EIP-7251, both EIP-8282 predeploys): the post-execution system call only reads the excess, count, queue head and queue tail slots; the BAL drops the predeploy's entry.
- `test_bal_invalid_missing_request_predeploy_reads` (same predeploys): the entry stays but its `storage_reads` are cleared.
- `test_bal_invalid_missing_pre_block_system_call_read`: with a zero parent beacon root the EIP-4788 system call rewrites the root slot with its current value, which EIP-7928 records as a read next to the timestamp write; the BAL drops that read.
- `test_bal_invalid_missing_system_contract_entry` (EIP-2935, EIP-4788): the BAL drops the whole entry of a system contract written at `block_access_index = 0`.
- `test_bal_invalid_missing_zero_amount_withdrawal_recipient` (existing and non-existent recipient): EIP-7928 records withdrawal recipients regardless of amount; the BAL drops the recipient of a zero-amount withdrawal.

All blocks are empty, so the only accesses in the BAL are the system-phase ones under test. `test_cases.md` gets a row per test.

Motivation: lambdaclass/ethrex#7109. ethrex's default import path validates the supplied BAL instead of regenerating it and had no completeness check for the pre- and post-block phases, so a block whose BAL omitted a system-call read was accepted while geth, reth and erigon rejected it. The existing suite did not expose this: it never removes a read-only system-phase access.

Validation (filled locally with the in-repo EELS `t8n`, 26 fixtures: 13 `blockchain_test` + 13 `blockchain_test_engine`):

- ethrex sequential path (regenerate-and-compare): all 13 rejected.
- ethrex default parallel path at `main`: 11 of 13 accepted as `VALID`. The two `missing_system_contract_entry` cases are rejected, but by the state-root check rather than BAL validation (a `validationError` mismatch under strict exception matching).
- ethrex parallel path with lambdaclass/ethrex#7122 plus a pre-block completeness check: all 13 rejected with a BAL validation error, also under strict exception matching.

`just static` passes.

### Related Issues or PRs

lambdaclass/ethrex#7109
lambdaclass/ethrex#7122

### Checklist


fselmo (approved):
Thanks @ilitteri, these are great to have in here. Just did some housekeeping / explicit checks + descriptions. I was going to wait on #3484 refactors to come in first but I'll just update that one instead since this one is gtg. Thanks again!

PR #3554, opened 2026-09-09
fix(tests): EIP-8037 - drop double-counted hash gas in code deposit guard

chfast:
### Description

The initcode carried `code_deposit_size`, so its `execution_cost` already included the deposit's hash gas, and the guard added `code_deposit`'s execution cost on top. That overstated the child's pre-deposit execution gas, making the "child must reach code deposit" assert stricter than intended and the state-charge assert weaker.

Drop the metadata from the initcode so it prices the frame's own execution and `code_deposit` prices the deposit. Fixtures are unchanged: the guard is fill-time only.

### Related Issues or PRs

N/A.

### Checklist


spencer-tb (approved):
LGTM! Thanks, fixtures are identical :)

PR #3555, opened 2026-09-09
fix(test-consume): group test cases by fixture file and bound the per-worker fixture cache

danceratopz:
#### Background

The `consume-rlp` (all clients) and `consume-engine` (besu, neth) on the glamsterdam dashboards have recently failed to complete with `SIGKILL`:
```
.../_temp/<id>.sh: line 8: 13313 Killed    ./hive --sim "ethereum/eels/consume-engine" --client "nethermind"
```
The most reasonable explanation appears to be OOM on the runner, see "Supporting Evidence" below.

A likely culprit is `consume`'s fixture cache which holds a copy of all loaded fixtures **per runner** for the duration of the hive process for `consume-rlp` (the size `blockchain_test` fixtures in a Glamsterdam devnet release is 3.7 GB!):
https://github.com/ethereum/execution-specs/blob/ca2b3b18a5d4058b2e2fd517ba7db31e86919a09/packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/simulators/base.py#L45-L49

#### Changes

To reduce simulator-side memory usage , `consume` now:
1. Groups test cases by the fixture file that contains them (to improve cache reuse).
2. Limits the size of the per-worker cache of parsed fixture files via LRU .

#### Details

A fixture file holds several test cases (8 on average in the mainnet `tests@v20.0.2` release, 1,117 files with ten or more, up to 768), and the index lists the cases of one file in several separate runs (the 8,731 `blockchain_tests` files appear in 39,712 runs). In a six-worker simulation, the unbounded cache and original index order produced a total parse volume of 4.1–4.5 times the release size. Each worker retained every file it parsed for the duration of the pytest session.

Ordering: `pytest_generate_tests` sorts index test cases by `json_path` before parametrizing them (`group_test_cases_by_fixture_file`). The order within a file is preserved. Release paths group cases by format and fork directory in collection order; this does not guarantee that different forks execute sequentially across workers. Test cases read from stdin have no file and keep their order. EngineX is excluded: it sorts its collection largest pre-alloc group first in its own `pytest_collection_modifyitems` for `loadgroup` scheduling, and that order is left untouched.

Cache: `fixture_file_loader` now returns a `FixtureFileCache` that evicts the least recently used files when their combined on-disk size exceeds 256 MiB. The most recently loaded file is always kept, even if it exceeds the bound on its own. An evicted file is loaded again if a later test needs it. Grouping reduced repeated parsing in the simulation, though scheduling can still split a file across workers or cause later revisits.

On the hive.ethpandaops.io glamsterdam dashboard, full `consume rlp` runs currently end for every client about two thirds of the way through the suite, and full `consume engine` runs end for the clients with the largest containers: the `hive` process is killed on the runner and the run is discarded without an upload. This change limits fixture retention during long runs and might resolve those failures. The cache's contribution to them and the effect of this change still need to be verified on a runner; the file-size budget is not a limit on process memory.

The following measurements show how much the unbounded cache can reach and what the ordering saves. Memory was measured on the mainnet `tests@v20.0.2` release with `tracemalloc`, which counts Python allocations, so resident memory is somewhat higher; parse volumes come from a simulation of xdist's `LoadScheduling` (contiguous chunks of the pending list, six workers) over that release's index.

<details>
<summary>Supporting evidence</summary>

**Current Workflow Behavior**

Suite | Clients | Killed after | Tests started at kill | Sample
-- | -- | -- | -- | --
consume-rlp | all seven | 1.2 h (ethrex) to 19 h (besu) | 25.5k to 27.5k | every rlp job 2026-08-26 to 2026-09-08 that got a runner
consume-engine | nethermind | 13 to 16 h | 34.2k to 37.8k of 42.4k | 10 of 10 completed runs since 2026-08-26
consume-engine | besu | 22 to 30 h | 31.9k, 40.7k | 3 of 4 runs killed, one lost its runner, one succeeded (2026-08-26, 32.7 h)
consume-engine | go-ethereum, reth, erigon, ethrex, nimbus-el | never | all 42.4k | complete daily

**Parse volume relative to the release size on disk, six workers**:

| Cache and order | `blockchain_tests` (rlp) | `blockchain_tests_engine` | Busiest worker parsing (rlp) |
| --- | --- | --- | --- |
| unbounded cache, index order (before) | 4.1× | 4.5× | 3.2 min |
| 256 MiB bound, index order | 17.8× | 21.2× | 12.1 min |
| 256 MiB bound, grouped by file (this PR) | 1.1× | 1.2× | 1.0 min |
| no cache, every test parses its file | 114× | 136× | — |

**Parsed size relative to the JSON on disk, by fixture file size class:**

| File size class | Share of release bytes | Parsed size relative to JSON |
| --- | --- | --- |
| under 1 MB | ~20% | 2.5× |
| 1 to 10 MB | ~35% | 1.7× |
| 10 to 50 MB | ~35% | ~1× |
| 50 to 160 MB (RLP-heavy) | ~10% | 0.5× (hex strings become bytes) |

Byte-weighted, parsed fixtures take about 1.5× their JSON size. Applied to the whole release, per fixture format, for the unbounded cache:

| Format | Uncompressed JSON | Parsed, one worker holding every file | Six workers |
| --- | --- | --- | --- |
| `blockchain_tests` (consume rlp) | 3.7 GB in 8,731 files | ~5.5 GB | ~33 GB |
| `blockchain_tests_engine` (consume engine) | 2.3 GB in 7,799 files | ~3.7 GB | ~22 GB |

The six-worker column is a hypothetical upper bound assuming every worker holds every file. Each worker actually retains only the files it touches; the simulation does not establish that this upper bound is reached. These mainnet estimates cannot establish memory use for the glamsterdam devnet release, whose uncompressed per-format sizes and runtime memory use need separate measurement. Fixture retention may contribute to memory pressure on the 30.6 GB runners alongside client containers. After eviction, each cache retains at most 256 MiB of on-disk JSON, or one oversized file. Parsing occurs before eviction, so peak memory also includes the incoming parse.

</details>

Unit tests cover the grouping (contiguity, stable order within a file, fork directories grouped in collection order, stream cases untouched) and the cache (hits, eviction order, oversized files, rejection of missing files).

Deployment note: the devnet dashboards build simulator images from `devnets/glamsterdam/8`, `devnets/focil/0`, and `devnets/frames/0`, so those branches also need the change for the dashboards to use it.

### Related Issues or PRs

Results have also been frequently missing for these test runs, this is solved by:
- https://github.com/ethpandaops/hive-tests/pull/102

### Checklist

### Cute Animal Picture

image

### DIFF
--- a/packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/consume.py
+++ b/packages/testing/src/execution_testing/cli/pytest_commands/plugins/consume/consume.py
@@ -8,7 +8,7 @@ import tarfile
 from dataclasses import dataclass
 from io import BytesIO
 from pathlib import Path
-from typing import Any, List, Optional, Tuple, cast
+from typing import Any, List, Optional, Sequence, Tuple, cast
 from urllib.parse import urlparse
 
 import platformdirs
@@ -30,6 +30,7 @@ from execution_testing.fixtures import (
 from execution_testing.fixtures.consume import (
     IndexFile,
     TestCaseBase,
+    TestCaseIndexFile,
     TestCases,
 )
 from execution_testing.forks import (
@@ -591,6 +592,31 @@ def fixtures_source(request: pytest.FixtureRequest) -> FixturesSource:  # noqa: 
     return request.config.fixtures_source  # type: ignore[attr-defined]
 
 
+def group_test_cases_by_fixture_file(
+    test_cases: Sequence[TestCaseBase],
+) -> Sequence[TestCaseBase]:
+    """
+    Return the test cases ordered so that the cases of one fixture file are
+    contiguous.
+
+    A fixture file can hold several test cases interleaved with other files
+    in the index. Grouping improves cache reuse when pytest-xdist assigns
+    contiguous chunks of the collection, but does not guarantee that all
+    tests from a file run on one worker.
+
+    Preserve the order within each file. For release paths, sorting groups
+    cases by format and fork directory in collection order; workers can
+    still execute different forks concurrently. Test cases read from stdin
+    have no file and are returned unchanged.
+    """
+    index_cases = [
+        tc for tc in test_cases if isinstance(tc, TestCaseIndexFile)
+    ]
+    if len(index_cases) != len(test_cases):
+        return test_cases
+    return sorted(index_cases, key=lambda tc: str(tc.json_path))
+
+
 def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     """
     Generate test cases for every test fixture in all the JSON fixture files
@@ -604,6 +630,11 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
     supported_fixture_formats: List[FixtureFormat] = getattr(
         metafunc.config, "supported_fixture_formats", []
     )
+    is_enginex = BlockchainEngineXFixture in supported_fixture_formats
+    if not is_enginex:
+        # EngineX orders its collection by pre-alloc group instead, see
+        # `pytest_collection_modifyitems` in the enginex plugin.
+        test_cases = group_test_cases_by_fixture_file(test_cases)
     param_list = []
     for test_case in test_cases:
         if test_case.format not in supported_fixture_formats:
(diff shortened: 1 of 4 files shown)


spencer-tb (approved):
LGTM! Thanks!

PR #3561, opened 2026-09-10
chore(tests): drop stale EIP-8037 TODO on the 7702 precompile intrinsic gas test

shubham5080:
## 🗒️ Description

Adds the Amsterdam / EIP-8037 (EIP-2780 top-frame) counterpart of Prague's `test_set_code_to_precompile_not_enough_gas_for_precompile_execution`: set code to a precompile, call the authority in the same tx with only intrinsic + top-frame gas, and assert receipt/header gas with **no** existing-authority auth refund.

## 🧸 Cute Animal Picture

Cute animal

## ✅ Checklist

- [x] Local fill passed (54)
- [x] Prague TODO updated to point at the new test


spencer-tb:
Thanks for adding, but we have this coverage already so I dropped the todo here instead 🫡

spencer-tb (approved):

PR #3562, opened 2026-09-10
chore(skill): refresh skills

fselmo:
### Description

- Add some useful conventions already present in our best tests to the test-writing skill and refresh stale lines
- Update `fill` skill
- Refresh stale line in lint skill

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

🎠

### DIFF
--- a/.agents/skills/fill-tests/SKILL.md
+++ b/.agents/skills/fill-tests/SKILL.md
@@ -9,7 +9,7 @@ CLI reference for the `fill` command. Run this skill before filling test fixture
 
 ## Basic Usage
 
-```
+```bash
 uv run fill tests/                                    # Fill all tests
 uv run fill tests/cancun/ --fork Cancun               # Specific fork
 uv run fill tests/path/to/test.py -k "test_name"      # Specific test
@@ -19,13 +19,13 @@ uv run fill --collect-only tests/                      # Dry run: list tests wit
 
 ## Key Flags
 
-- `--fork FORK` / `--until FORK` — target specific fork or range
+- `--fork FORK` / `--until FORK` — target specific fork or range. The default set stops at the last deployed fork, so a fork under development needs one of these.
 - `--output DIR` + `--clean` — output directory; `--clean` required when re-filling
 - `-k "pattern"` — filter tests by name pattern
 - `-m "marker"` — filter by pytest marker (e.g. `-m state_test`, `-m blockchain_test`)
 - `-n auto --maxprocesses N` — parallel execution (use `--dist=loadgroup`)
 - `--evm-bin PATH` — t8n tool; defaults to the in-repo EELS Python spec (`src/ethereum/`)
-- `--verify-fixtures` — verify generated fixtures against geth blocktest
+- `--verify-fixtures` — run geth's `evm blocktest` over the generated fixtures. The default EELS t8n has no blocktest, so pass a geth binary with `--verify-fixtures-bin`. For an EELS-side check use `just validate-blocks <fixtures_dir>`, which is what CI runs.
 - `--generate-all-formats` — generate all fixture formats (2-phase)
 
 ## Debugging
@@ -45,10 +45,11 @@ uv run fill --collect-only tests/                      # Dry run: list tests wit
 - Excluded from a broad `tests/` run: include them by targeting a `tests/benchmark/...` path, or add `--include-benchmark` when also collecting `tests/`.
 - Pick a mode (mutually exclusive): `--gas-benchmark-values 1,10,100` (millions of gas) or `--fixed-opcode-count 1,10,100` (thousands). These parametrize the tests, e.g. `...[fork_Prague-blockchain_test-benchmark-gas-value_1M]`.
 - Backend is optional: omitting `--evm-bin` runs the slow in-repo EELS Python spec; `--evm-bin=evmone` or `--evm-bin=evm` (geth, used by `just bench-gas`) are faster.
+- Stateful benchmarks (`tests/benchmark/stateful/`) are filled by the separate `fill-stateful` command against a live client snapshot and produce `BlockchainEngineStatefulFixture`; see `docs/filling_tests/fill_stateful.md`.
 
 ## Fixture Formats
 
-One test function auto-generates multiple formats: `StateFixture`, `BlockchainFixture`, `BlockchainEngineFixture`. Use `--generate-all-formats` for additional formats via 2-phase execution.
+One test function auto-generates multiple formats: `StateFixture`, `BlockchainFixture`, `BlockchainEngineFixture`. `--generate-all-formats` adds the pre-alloc-group format `BlockchainEngineXFixture` through a 2-phase run (`--generate-pre-alloc-groups`, then `--use-pre-alloc-groups`); `--single-fixture-per-file` writes each fixture to its own JSON file.
 
 ## References
 
--- a/.agents/skills/lint/SKILL.md
+++ b/.agents/skills/lint/SKILL.md
@@ -33,7 +33,7 @@ just static
 
 - **Remaining ruff issues**: fix manually (auto-fix can't handle all rules)
 - **mypy errors**: fix type annotations, add missing types, correct signatures
-- **codespell errors**: fix typos, or add intentional words via `just whitelist <word>`
+- **codespell errors**: fix typos, or add intentional words to `whitelist.txt` (codespell's ignore list)
 - **ethereum-spec-lint errors**: fix import isolation violations (see `/implement-eip` for import rules)
 - **actionlint errors**: fix workflow YAML issues (see `/edit-workflow`)
 
--- a/.agents/skills/write-test/SKILL.md
+++ b/.agents/skills/write-test/SKILL.md
@@ -12,14 +12,15 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
 - All test imports come from `execution_testing` — it is the public API
 - Core fixtures: `pre: Alloc` (pre-state builder), `state_test: StateTestFiller`, `blockchain_test: BlockchainTestFiller`, `fork: Fork`
 - Rule: use `state_test` for single-transaction tests; `fill` auto-derives a `blockchain_test` from each, so no coverage is lost.
-- Exception: use `blockchain_test` when the test needs more than one transaction (a `state_test` holds exactly one) or more than one block (e.g. transaction-ordering or fork-transition tests).
+- Exception: use `blockchain_test` when the test needs more than one transaction (a `state_test` holds exactly one), more than one block (e.g. transaction-ordering or fork-transition tests), or calls a system contract: the state-test pre-alloc omits the predeploys a fork lists only in `pre_allocation_blockchain()`, so the `CALL` hits an empty account, "succeeds", and the test passes vacuously.
 - Anti-pattern: wrapping one transaction in a `Block` to reach `blockchain_test`. A `state_test` can assert the transaction's gas used and receipt logs (the tx's `expected_receipt=TransactionReceipt(cumulative_gas_used=...)`), reserve state gas (the tx's `state_gas_reservoir=`), and other block-header fields (`blockchain_test_header_verify=Header(...)`) without it.
+- If the framework cannot express what a test needs — a fork-derived parameter set, a protocol constant, a cost — add it to the framework (a `fork.*()` accessor, a mixin ClassVar, a covariant marker) instead of building it in the test. Importing a helper from a sibling `test_*.py` is the sign it belongs there.
 
 ## Pre-State Setup
 
-- `pre.fund_eoa()` — create funded EOA, returns Address. Accepts `amount=`, `nonce=`
+- `pre.fund_eoa()` — create funded EOA, returns Address. Accepts `amount=`, `nonce=`. With `amount=0` nothing is added to the pre-alloc; you get a fresh, nonexistent address.
 - `pre.deploy_contract(code=..., storage={...})` — deploy contract, returns Address
-- Anti-pattern: do NOT manipulate `pre` dict directly (`pre[addr] = Account(...)`)
+- Do not assign into `pre` directly. The one exception is overriding a predeploy's code, `pre[addr] = Account(...)` under `@pytest.mark.pre_alloc_mutable`: it replaces the whole account (storage gone, nonce 1) and skips the test in execute mode, so reserve it for scenarios a live chain cannot host.
 
 ## Bytecode Construction
 
@@ -34,6 +35,15 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
 - `storage = Storage()` then `storage.store_next(expected_value)` — auto-increments slot
 - `Op.SSTORE(storage.store_next(sender), Op.ORIGIN)` — build bytecode + expected storage in one step
 - Post-state: `post = {contract: Account(storage=storage)}`
+- `Account(storage=...)` compares storage **exhaustively**: any slot you omit must be zero. Opt an omitted key out with `Storage.set_expect_any(key)`.
+
+## Block Access Lists
+
+Every Amsterdam+ fixture carries a BAL whether or not the test asserts one. An `expected_block_access_list=` is a fill-time check that the spec built the BAL you predicted, so write one when the access pattern is the point of the test or a known edge (a revert, a system call, a withdrawal, a self-destruct), not by default. When you do write one, pair it with a `post` witness: the BAL records access, not outcome, so a transaction that ran and failed still merges its touches and satisfies the expectation, and only `post` tells the two apart.
+
+- Field semantics: a field left unset is not checked, `[]` asserts empty, and a non-empty list matches as an **ordered subsequence** (extra actual entries are skipped; yours must appear in order). `BalAccountExpectation()` with no field set raises; use `.empty()` for an account with no changes and `{address: None}` to assert an address is absent.
+- Reads and created accounts survive a frame's revert; writes do not. A test that reverts after touching state still expects those touches, and a missing entry is not evidence the access never happened.
+- Withdrawals and pre/post-execution system calls add entries no transaction accounts for (index `0` before, `len(txs) + 1` after). An expectation built only from the transaction list comes up short.
 
 ## Markers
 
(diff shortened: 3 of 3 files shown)


.agents/skills/write-test/SKILL.md
```diff
@@ -34,50 +35,76 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
 - `storage = Storage()` then `storage.store_next(expected_value)` — auto-increments slot
 - `Op.SSTORE(storage.store_next(sender), Op.ORIGIN)` — build bytecode + expected storage in one step
 - Post-state: `post = {contract: Account(storage=storage)}`
+- `Account(storage=...)` compares storage **exhaustively**: any slot you omit must be zero. Opt an omitted key out with `Storage.set_expect_any(key)`.
+
+## Block Access Lists
+
+Every Amsterdam+ fixture carries a BAL whether or not the test asserts one. An `expected_block_access_list=` is a fill-time check that the spec built the BAL you predicted, so write one when the access pattern is the point of the test or a known edge (a revert, a system call, a withdrawal, a self-destruct), not by default. When you do write one, pair it with a `post` witness: the BAL records access, not outcome, so a transaction that ran and failed still merges its touches and satisfies the expectation, and only `post` tells the two apart.
```
raxhvl, line 42: Suggestion:  Use the BAL test case reference to simplify

Reduce this section as an instruction + reference to test case file  "Write a BAL assertion for all test post Amsterdam. Refer [test_cases.md]  "

The test case file exhaustively capture BAL behavior than what we describe here. Agents can use that as a single entry point to grep the required test to learn more, or report missing coverage by frequently revisiting.

fselmo:
Thanks for first pass here. I kept some of the ideas and did some more cleaning up. I think this is a pretty good pass from the additions I've been wanting to get in.

spencer-tb (approved):
LGTM!

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
+- If the framework cannot express what a test needs — a fork-derived parameter set, a protocol constant, a cost — add it to the framework (a `fork.*()` accessor, a mixin ClassVar, a covariant marker) instead of building it in the test. Importing a helper from a sibling `test_*.py` is the sign it belongs there.
 
 ## Pre-State Setup
 
-- `pre.fund_eoa()` — create funded EOA, returns Address. Accepts `amount=`, `nonce=`
+- `pre.fund_eoa()` — create funded EOA, returns Address. Accepts `amount=`, `nonce=`. With `amount=0` nothing is added to the pre-alloc; you get a fresh, nonexistent address.
 - `pre.deploy_contract(code=..., storage={...})` — deploy contract, returns Address
-- Anti-pattern: do NOT manipulate `pre` dict directly (`pre[addr] = Account(...)`)
+- Do not assign into `pre` directly. The one exception is overriding a predeploy's code, `pre[addr] = Account(...)` under `@pytest.mark.pre_alloc_mutable`: it replaces the whole account (storage gone, nonce 1) and skips the test in execute mode, so reserve it for scenarios a live chain cannot host.
```
danceratopz: Account replacement defaults to nonce 0. `Account(code=Op.STOP).nonce` is 0; nonce 1 is the `deploy_contract()` default. Direct assignment also replaces balance and storage, so the guidance should ask the author to supply the intended fields.

```suggestion
- Do not assign into `pre` directly for ordinary setup. Use `@pytest.mark.pre_alloc_mutable` only when the scenario requires state a live chain cannot host, such as overriding a predeploy's code. `pre[addr] = Account(...)` replaces the entire account; explicitly supply any nonce, balance, and storage the scenario requires. Unspecified fields take `Account` defaults, including nonce 0. The marker skips the test in execute mode.
```

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
 - Rule: omit `gas_limit`. It auto-fills so the transaction executes in full without running out of gas.
 - Exception: set `gas_limit` explicitly for gas-sensitive tests (intrinsic-gas boundaries, OOG, code-deposit limits, or gas metering).
 - Anti-pattern: the `gas_limit=fork.transaction_gas_limit_cap()` boilerplate is now redundant.
+- A transaction that runs out of gas consumes exactly its `gas_limit`, so calling a contract whose code is `GasConsumer.out_of_gas(fork)` pins its gas used to a chosen value without any cost arithmetic.
 
 ## Exception Testing
 
 - Pass `error=TransactionException.INTRINSIC_GAS_TOO_LOW` to `Transaction`
 - Common exceptions: `GAS_ALLOWANCE_EXCEEDED`, `NONCE_MISMATCH_TOO_LOW`, `INSUFFICIENT_ACCOUNT_FUNDS`
+- Expect the one exception the spec names. A client that reports a different error first is a mapper fix on the client side, not a reason for `exception=[A, B]`.
```
danceratopz: Could we distinguish incorrect exception mapping from legitimate differences in validation order? An input can violate multiple constraints, and a client can correctly report a different violation first. A mapper should not collapse distinct failures just to make the test accept them.

```suggestion
- Prefer a test that violates only the intended constraint and expects one exception. Fix incorrect client error mappings rather than broadening expectations to hide them. If multiple violations are unavoidable and validation order is unspecified, document why each accepted exception is valid before using `exception=[A, B]`.
```

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
-- **Transaction-level costs:** `fork.transaction_intrinsic_cost_calculator()`; `fork.transaction_top_frame_state_gas(contract_creation=True)` for the created account's `NEW_ACCOUNT` (under EIP-2780 it is NOT part of the intrinsic — never subtract it from the intrinsic); `fork.transaction_data_floor_cost_calculator()`; `fork.call_value_stipend()`.
+- **Transaction-level costs:** `fork.transaction_intrinsic_cost_calculator()`; `fork.transaction_top_frame_state_gas(contract_creation=True)` for the created account's `NEW_ACCOUNT` (under EIP-2780 it is NOT part of the intrinsic — never subtract it from the intrinsic); `fork.transaction_data_floor_cost_calculator()` (pass `contract_creation=True` for a creation transaction — its floor is anchored on a create-shaped base and comes out 9000 gas low without it); `fork.call_value_stipend()`.
```
danceratopz: The 9000-gas difference is specific to the current Amsterdam rules. Calling the floor calculator with and without `contract_creation=True` gives a difference of 0 in Prague and 9000 in Amsterdam. Passing the actual transaction shape is the durable rule; the literal will become stale on repricing.

```suggestion
- **Transaction-level costs:** `fork.transaction_intrinsic_cost_calculator()`; `fork.transaction_top_frame_state_gas(contract_creation=True)` for the created account's `NEW_ACCOUNT` (under EIP-2780 it is NOT part of the intrinsic — never subtract it from the intrinsic); `fork.transaction_data_floor_cost_calculator()` (pass the actual transaction shape, including `contract_creation=True` for creation transactions, so fork-specific floor adjustments are included); `fork.call_value_stipend()`.
```

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
+- Expect the one exception the spec names. A client that reports a different error first is a mapper fix on the client side, not a reason for `exception=[A, B]`.
 
 ## Test Organization
 
 - Place tests in `tests/<fork>/eip<number>/` where `<fork>` is the fork that introduced the functionality
-- Each EIP directory has `spec.py` with `ReferenceSpec(git_path=..., version=...)` and test files declaring `REFERENCE_SPEC_GIT_PATH` / `REFERENCE_SPEC_VERSION`
+- Each EIP directory has `spec.py` with `ReferenceSpec(git_path=..., version=...)` and test files declaring `REFERENCE_SPEC_GIT_PATH` / `REFERENCE_SPEC_VERSION`. `version` is the EIP file's blob SHA (`gh api repos/ethereum/EIPs/contents/EIPS/eip-N.md --jq .sha`); `uv run check_eip_versions --until <Fork> <path>` flags stale pins.
+- **One module per subject, not per scenario.** Start a new test file only for a subject no existing module has: a different parametrization axis, fixture set, or fork validity. A new scenario for an existing subject goes into that subject's module however many tests it already holds, even where neighbouring suites split further. Two files with the same `pytestmark` are one module split in two.
```
danceratopz: Shared `pytestmark` values do not establish that two files cover the same subject. Distinct subjects often share fork validity. Reusing an existing subject module is useful, but identical marks or file length should not decide the structure on their own.

```suggestion
- **Organize modules by subject.** Add a scenario to the existing module when it shares the subject and setup. Split modules when distinct subjects, fixture sets, parametrization, or size make them easier to navigate. Matching `pytestmark` values alone do not imply that modules should be combined.
```

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
+- Exception: use `blockchain_test` when the test needs more than one transaction (a `state_test` holds exactly one), more than one block (e.g. transaction-ordering or fork-transition tests), or calls a system contract: a state test's pre-alloc lacks the predeploys a fork lists only in `pre_allocation_blockchain()`, so the `CALL` hits an empty account and the test passes vacuously.
 - Anti-pattern: wrapping one transaction in a `Block` to reach `blockchain_test`. A `state_test` can assert the transaction's gas used and receipt logs (the tx's `expected_receipt=TransactionReceipt(cumulative_gas_used=...)`), reserve state gas (the tx's `state_gas_reservoir=`), and other block-header fields (`blockchain_test_header_verify=Header(...)`) without it.
+- If the framework cannot express what a test needs — a fork-derived parameter set, a protocol constant, a cost — add it to the framework (a `fork.*()` accessor, a mixin ClassVar, a covariant marker) instead of building it in the test. Importing a helper from a sibling `test_*.py` is the sign it belongs there.
```
danceratopz: A sibling-test import is a useful signal to reconsider ownership, but not every shared helper belongs in the public framework. Scenario-specific fixtures can live in `conftest.py`, and test-local utilities can live in a helper module.

```suggestion
- If the framework cannot express a reusable protocol concept, such as a fork-derived parameter set, constant, or cost, add an appropriate framework API (`fork.*()` accessor, mixin ClassVar, or covariant marker). Avoid importing helpers from sibling `test_*.py` files: put shared fixtures in `conftest.py`, scenario-specific utilities in a local helper module, and reusable protocol logic in the framework.
```

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
 - `pre.deploy_contract(code=..., storage={...})` — deploy contract, returns Address
-- Anti-pattern: do NOT manipulate `pre` dict directly (`pre[addr] = Account(...)`)
+- Do not assign into `pre` directly. The one exception is overriding a predeploy's code, `pre[addr] = Account(...)` under `@pytest.mark.pre_alloc_mutable`: it replaces the whole account (storage gone, nonce 1) and skips the test in execute mode, so reserve it for scenarios a live chain cannot host.
 
 ## Bytecode Construction
 
 - `Op.SSTORE(key, value)`, `Op.CALL(gas, addr, ...)`, etc. — concatenate with `+`
 - `Op.PUSH32(val) + Op.PUSH32(val) + Op.EXP` for stack setup
 - Macros: `Om.OOG` (consumes all gas), `Om.MSTORE(data, offset)` (arbitrary-length memory store)
+- `GasConsumer(gas=n, fork=fork)` burns exactly `n` gas and falls through; `GasConsumer.out_of_gas(fork)` always runs out. Both are priced against the fork, so use them instead of sizing a burn by hand (`JUMPDEST` padding, a large `MSTORE` offset).
```
danceratopz: The exact-cost guarantee depends on the memory state at entry. The helper exposes `previous_memory_size` for this, so it is worth mentioning to avoid under-burning when it follows earlier memory expansion.

```suggestion
- `GasConsumer(gas=n, fork=fork)` burns exactly `n` gas and falls through; `GasConsumer.out_of_gas(fork)` always runs out. Both are priced against the fork, so use them instead of sizing a burn by hand (`JUMPDEST` padding, a large `MSTORE` offset). When appending after memory expansion, pass `previous_memory_size` matching the memory size at entry; otherwise the incremental gas consumed can differ from `n`.
```

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
 - For gas amounts, see **Gas Cost Expectations** below — prefer framework cost constructs over reading `fork.gas_costs()` constants directly
 
+## Pinning Intent
+
+Pinning is a necessity, not a nice-to-have. With thousands of tests nobody re-reads them to confirm they still fill for the reason they were written, so the pin is the only thing that notices when one stops.
+
+- **Pin a value that is as close to unique to this case as possible.** It should hold under the rule the test is about and under no other plausible rule, so that a test drifting onto a different rule, or a fork changing which rule applies, fails `fill` instead of quietly producing a new fixture.
+- **A pin that restates an input is not a pin.** `cumulative_gas_used` equal to the `gas_limit` the test itself set, a storage slot expected to be zero when zero is also the untouched value, an `Account.NONEXISTENT` that a mis-wired setup would produce anyway: each of these passes just as happily under the wrong behavior. Give the transaction slack above the boundary so the charged amount cannot equal the limit, pre-set the witness slot to a sentinel, and assert the created address you actually expect.
```
danceratopz: Added from memory, must admit I wasn't familiar with this one :laughing: 

Optional: the framework already provides `Storage.canary()` for the expected-zero witness case. Naming it makes the sentinel advice directly usable. Initializing storage changes original values and can change SSTORE costs/refunds, so gas-sensitive tests must account for it.

```suggestion
- **A pin that restates an input is not a pin.** `cumulative_gas_used` equal to the `gas_limit` the test itself set, a storage slot expected to be zero when zero is also the untouched value, an `Account.NONEXISTENT` that a mis-wired setup would produce anyway: each of these passes just as happily under the wrong behavior. Give the transaction slack above the boundary so the charged amount cannot equal the limit, pre-set the witness slot to a sentinel, and assert the created address you actually expect. For expected-zero witness slots, `pre.deploy_contract(code, storage=storage.canary())` initializes those slots to nonzero values while `Account(storage=storage)` keeps the intended post-state. Account for the changed original storage values in gas-sensitive tests.
```

.agents/skills/fill-tests/SKILL.md
```diff
@@ -19,13 +19,13 @@ uv run fill --collect-only tests/                      # Dry run: list tests wit
...
 
-- `--fork FORK` / `--until FORK` — target specific fork or range
+- `--fork FORK` / `--until FORK` — target specific fork or range. The default set stops at the last deployed fork, so a fork under development needs one of these.
 - `--output DIR` + `--clean` — output directory; `--clean` required when re-filling
 - `-k "pattern"` — filter tests by name pattern
 - `-m "marker"` — filter by pytest marker (e.g. `-m state_test`, `-m blockchain_test`)
 - `-n auto --maxprocesses N` — parallel execution (use `--dist=loadgroup`)
 - `--evm-bin PATH` — t8n tool; defaults to the in-repo EELS Python spec (`src/ethereum/`)
-- `--verify-fixtures` — verify generated fixtures against geth blocktest
+- `--verify-fixtures` — run geth's `evm blocktest` over the generated fixtures. The default EELS t8n has no blocktest, so pass a geth binary with `--verify-fixtures-bin`. For an EELS-side check use `just validate-blocks <fixtures_dir>`, which is what CI runs.
```
danceratopz: Optional clarification: this already documents the right command and its CI use, which is enough to tell a developer how to run it. One extra sentence would preserve why the second check matters: filling through t8n does not exercise the complete EELS block-validation path. No separate section is needed.

```suggestion
- `--verify-fixtures` — run geth's `evm blocktest` over the generated fixtures. The default EELS t8n has no blocktest, so pass a geth binary with `--verify-fixtures-bin`. For an EELS-side check use `just validate-blocks <fixtures_dir>`, which is what CI runs. Filling through t8n does not exercise the full block-validation path; `validate-blocks` replays blockchain fixtures through EELS block validation, including header checks that filling alone does not cover.
```

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
+- **Make a mid-transaction value durable.** A `CALL` result, a `BALANCE`, `GAS` or `EXTCODESIZE` reading only exists while the code runs; `SSTORE` it into a witness slot and assert that slot in `post` (`Storage.store_next(expected)` builds the code and the expectation together).
+- **Say which rule the pinned number comes from.** One short comment naming the rule lets the next reader tell a repricing from a regression.
+- **Break it once.** After the test fills, mutate the setup so the behaviour under test cannot happen and confirm the fill fails for that reason; then restore it.
```
danceratopz: Added from memory.

Extend this to negative tests. A deliberately invalid input can still pass a broad exception assertion because an unrelated setup or decoding failure occurred. The mutation should demonstrate that the intended validation is responsible, not merely that some exception was raised.

```suggestion
- **Break it once.** After the test fills, mutate the setup so the behaviour under test cannot happen and confirm the fill fails for that reason; then restore it. For negative tests, verify that the intended validation causes rejection: assert the expected exception type and, where needed, a stable distinguishing detail. Avoid broad `pytest.raises(Exception)` assertions that also accept unrelated setup or decoding failures. For unit tests, remove the guard under test and confirm the test fails.
```

danceratopz (commented):
Thanks @fselmo!! I asked codex to verify the changes and add/update existing skills from memories accordingly to avoid them.  

The ones added from memories are flagged explicitly as such.

I wanted to do the same with Claude (I have much more token mileage there), but hey, hit my session limit, will come back to it at 13:50 UTC sharp.

I think the exception mapping and the negative test mutation comments are def worth having, no strong feelings on the rest, but they look ok to me! If you think any comments are too much noise, please just ignore them, no justification required. 

---

The additions capture useful test-writing practices, especially fill-time assertions, boundary pairs, fork-aware helpers, and explicit storage/BAL expectations. I would correct the account nonce statement, qualify the exception-order guidance, and remove the fork-specific floor delta before merging.

The remaining inline suggestions clarify helper ownership, module organization, and `GasConsumer` memory assumptions, and add optional guidance for `storage.canary()` and negative tests. The validate-blocks command is already documented correctly; that comment only proposes a short explanation of the extra coverage.

danceratopz:
Prompt to grab your memories:

```
I'm reviewing https://github.com/ethereum/execution-specs/pull/3562. Can you Inspect your available memories and lessons from previous work in this repository. Identify reusable knowledge that could help other developers: API pitfalls, testing practices, verification steps, or recurring mistakes.

For each proposed addition:
  - Verify it against the current code; memories may be stale.
  - Check whether the PR or existing guidance already covers it.
  - Prefer a small addition to the appropriate existing skill.
  - Keep guidance scoped and explain legitimate exceptions.
  - Exclude personal preferences, temporary work state, and historical details.
  - Prefix memory-derived suggestions with “Added from memory.”
  - Identify the source memory and explain whether shared documentation would make it redundant. Do not delete memories.
```

.agents/skills/lint/SKILL.md
```diff
@@ -33,7 +33,7 @@ just static
 
 - **Remaining ruff issues**: fix manually (auto-fix can't handle all rules)
 - **mypy errors**: fix type annotations, add missing types, correct signatures
-- **codespell errors**: fix typos, or add intentional words via `just whitelist <word>`
+- **codespell errors**: fix typos, or add intentional words to `whitelist.txt` (codespell's ignore list)
```
danceratopz: I would keep the original `just whitelist <word>` instruction (it keeps the lines alphabetically sorted).

.agents/skills/write-test/SKILL.md
```diff
@@ -91,6 +121,9 @@ Never hand-reconstruct a gas amount by summing `fork.gas_costs()` constants (`NE
...
+- Handle every parametrized case with an explicit `if`/`elif` and `else: raise ValueError(...)`; an `else` that is a real case silently absorbs values added later.
+- Parametrize the dimensions that take different code paths in clients (warm/cold, empty/funded, same-tx/pre-deployed), not just the ones the spec names.
+- **Cover the family, not just the instance.** Before fixing the subject of a test, ask whether it is one member of a set the framework already parametrizes: call opcodes, create opcodes, precompiles, system contracts, request classes, tx types. If so, use that `with_all_*` marker and narrow it with `selector=lambda value: ...` when only part of the set can reach the behaviour (only `CALL` and `CALLCODE` carry a value, so only they can fail a sender-balance check). The nearest test proving the same rule for one member usually already carries the marker, so read a model test's decorators and not just its body. If the family is real but no marker covers it, propose a covariant marker (`covariant_decorator` in `plugins/forks/forks.py`) rather than hand-writing the list in the test; framework surface lands in its own change.
```
danceratopz:
```suggestion
- **Cover the family, not just the instance.** Before fixing the subject of a test, ask whether it is one member of a set the framework already parametrizes: call opcodes, create opcodes, precompiles, system contracts, request classes, tx types. If so, use that `with_all_*` marker and narrow it with `selector=lambda value: ...` when only part of the set can reach the behaviour (only `CALL` and `CALLCODE` carry a value, so only they can fail a sender-balance check). The nearest test proving the same rule for one member usually already carries the marker, so read a model test's decorators and not just its body. If the family is real but no marker covers it, propose a covariant marker (`covariant_decorator` in `packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py`) rather than hand-writing the list in the test; framework surface lands in its own change.
```

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
+- A mark that only some cases earn goes on that case, `pytest.param(..., marks=...)`, not the function: `exception_test` on the function fails the passing cases, and a function-level `EIPChecklist` item stays green after the one case that proved it is deleted.
 
 ## Fork-Aware Logic
 
-- `fork >= Cancun` for conditional behavior based on fork
+- Branch on `fork.is_eip_enabled(N)` for behaviour an EIP introduces; EIPs move between forks, so `fork >= Cancun` is only for facts about the fork itself
 - `fork.fork_at(timestamp=...)` gives the fork active before/after a transition boundary
 - For gas amounts, see **Gas Cost Expectations** below — prefer framework cost constructs over reading `fork.gas_costs()` constants directly
 
+## Pinning Intent
```
danceratopz: "Pinning" is one of those silly claudisms imo 😆 It doesn't clearly express what it's talking about.

```suggestion
## Test Expectations
```

If you also prefer "Test Expectations" for this section, could you ask it to re-write this file using "test expectations"? In particular the whole of this section, but also:

  - Line 86, Gas Cost Expectations: "pins one threshold" becomes "tests one threshold", and the cross-reference "see Pinning Intent" becomes "see Test Expectations".
  - Line 97, Transactions: "pins its gas used to a chosen value" becomes "fixes its gas used at a chosen value" (this one is not so bad tho).
  - Line 108, Test Organization: "flags stale pins" becomes "flags stale versions". This is the version-SHA sense and was never the same word.
  - Line 115, Test Docstrings: "the rule it pins" becomes "the rule it tests". This line predates the PR, but leaving it would keep one stray use.

.agents/skills/write-test/SKILL.md
```diff
@@ -12,72 +12,102 @@ Conventions and patterns for writing consensus tests. Run this skill before writ
...
+
+Pinning is a necessity, not a nice-to-have. With thousands of tests nobody re-reads them to confirm they still fill for the reason they were written, so the pin is the only thing that notices when one stops.
```
danceratopz: The intro motivates the term instead of defining it. I would replace it with the definition and a table of where each expectation lives. The table also settles `header_verify=` vs `blockchain_test_header_verify=` and `error=` vs `exception=` in one place:

```suggestion
An expectation is what `fill` checks. Each lives on one object:

| On | Field |
|---|---|
| `Transaction` | `error=`, `expected_receipt=TransactionReceipt(...)` |
| `Block` | `exception=`, `header_verify=Header(...)`, `expected_block_access_list=` |
| `StateTest` | `blockchain_test_header_verify=Header(...)`, `expected_block_access_list=` |
| `post` | `Account(storage=..., balance=..., nonce=..., code=...)` |

Choose each so it holds only if the behaviour under test happened; then the fill fails when that behaviour stops instead of writing a fixture that passes for another reason.
```

For the rest of the section (together with the rename suggested above) I would keep the bullets and only swap the vocabulary:

- "Pin a value" becomes "Expect a value"
- "A pin that restates an input is not a pin" becomes "An expectation that restates an input checks nothing"
- "The pin has to bite at fill time" becomes "The expectation has to bite at fill time"
- "Pin block premises" becomes "Assert block premises"; with the table in place this bullet needs no field-name aside
- "the pinned number" becomes "an expected number"

Line 103 should then read `error=[A, B]`, since that section is about `Transaction`.

fselmo:
Thanks @danceratopz, I took most of these, just shortened the wording and vetted against Claude, in order to curb context bloat. I think this is a great update to get some better quality defaults cemented in 👍🏼

PR #3552, opened 2026-09-09
feat(ci): validate filled blockchain fixtures with EELS

danceratopz:
### Description

Filling never runs the EELS block validator. The t8n path calls `process_transaction` and the block-level helpers, but the test framework assembles the block header itself, so `state_transition`, `execute_block`, `validate_header` and `apply_body` are only reached when a fixture is consumed. Header-level invalid blocks (`rlp_modifier`, requests and BAL overrides) are also exempt from t8n exception verification during fill.

This PR takes @spencer-tb's idea from #3526 to add a `just validate-blocks <blockchain_tests dir> [pytest args]` recipe, which runs an existing `blockchain_tests` directory through the `tests/json_loader` path. In contrast to #3526, the conftest is loaded as a pytest plugin, so any fixture directory can be passed directly and no symlink under `tests/json_loader` is needed. Coverage data is written under `.just/validate-blocks/`; `--no-cov` skips it.

This PR then additionally makes these CI changes:

- `test.yaml`: each fill matrix job validates the fixtures it just filled and uploads the validation coverage next to the fill coverage, so codecov sees the union.
- `build-fixtures` action: each split EELS release build validates its `blockchain_tests` before uploading the artifact. Unsplit (tarball) and non-EELS builds are skipped.

#### Cost on the CI runners

The tables compare the fill matrix jobs of the `Python Specification` workflow without and with the new step. The two baseline runs are the two most recent `forks/amsterdam` push runs, which have no validation step. The two PR runs are the runs on this branch; both commits have the same tree and differ only in the first commit message. Durations are job and step `started_at` to `completed_at` from the Actions API, so runner queue time is excluded. Fixture counts are the pytest summary lines of the new step.

Fill job duration:

| label | [push 80edd40](https://github.com/ethereum/execution-specs/actions/runs/34099530991) | [push 903b48f](https://github.com/ethereum/execution-specs/actions/runs/33883920342) | [PR run 1](https://github.com/ethereum/execution-specs/actions/runs/34328299130) | [PR run 2](https://github.com/ethereum/execution-specs/actions/runs/34328724568) |
|---|---|---|---|---|
| pre-shanghai | 2m21s | 2m23s | 2m47s | 2m37s |
| shanghai-cancun | 2m13s | 2m11s | 2m37s | 2m23s |
| prague | 2m17s | 2m12s | 2m29s | 2m26s |
| osaka | 2m20s | 2m15s | 2m36s | 2m32s |
[...]

PR #3535, opened 2026-09-06
feat(spec-specs, tests): implement EIP-8253, bump nonce of zero-nonce storage accounts

jochem-brouwer:
### Description

Implements [EIP-8253](https://eips.ethereum.org/EIPS/eip-8253) for Amsterdam, with tests.

**Spec.** At the start of the fork block, before the pre-execution system calls, the nonce of the 28 Mainnet accounts with empty code, zero nonce and non-empty storage is set to one. Balance, code and storage are untouched. The BAL records one `NonceChange [0, 1]` per account at block access index 0. The bump is unconditional, as the EIP specifies: the fixed address list selects the accounts, not their state, so a listed account is bumped even when it is absent or has empty storage.

**Fork block detection.** The spec derives `is_fork_block` from `FORK_CRITERIA` and the parent header. The filler passes an explicit `fork_activation` flag to the transition tool (`--state.fork-activation` for the CLI and daemon); external t8n binaries do not receive it yet. The spec's own detection is exercised when the `validate-blocks` CI step replays the transition fixtures through `state_transition`, which needs #3564.

**Pre-states.** The framework imposes no Mainnet-shape restriction on custom pre-states, at the listed addresses or elsewhere: the list only says which addresses are bumped. Existing tests that build codeless zero-nonce storage accounts keep running at Amsterdam. The EXTCODEHASH empty-account test gains a nonce-bearing storage variant, which is the only such shape reachable on Mainnet after Spurious Dragon.

**Tests** (`tests/amsterdam/eip8253_...`, all fork-transition; see its `test_cases.md`): nonce bump with full BAL and post-state checks, with the fork block at the fork timestamp or past it; `CREATE` collision at and after the fork block from a synthetic factory at each account's historical creator address and nonce (verified against chain data), plus `CREATE` failures before the target is accessed; calls to a bumped account; non-targeted and same-shape accounts untouched; listed accounts that are absent or have empty storage. BAL merging of the index-zero nonce change with reads, transfers, reverted transfers, OOG before access, EIP-2930 listings, `SELFDESTRUCT`, fee recipient, withdrawals, and the BAL size budget; and rejection of BALs with a missing, wrong, duplicated, split, mis-ordered, or non-canonically encoded bump. The EIP-7928 BAL size transition test budgets the 28 extra fork-block entries. The BAL modifier module gains a `nonce_block_access_index` scalar for the index-zero encoding test.

[...]

PR #3564, opened 2026-09-11
feat(spec-tests, spec-tools): run fork-transition fixtures through EELS

jochem-brouwer:
### Description

Teach the json-loader to run fork-transition fixtures through EELS.

Fork-transition tests are filled in CI, which executes the spec block by block through the in-process t8n. The resulting fixtures were never replayed through the spec's `state_transition`, though: a fixture whose `network` names a transition, such as `BPO2ToAmsterdamAtTime15k` or `BerlinToLondonAt5`, is not in the fork table, so the json-loader job and the `validate-blocks` step from #3552 dropped it at collection. Spec code that runs only in that block-level path at a fork boundary, header validation against the previous fork's parent header, `apply_fork`, and fork-block detection from `FORK_CRITERIA`, was therefore not exercised across a boundary in CI.

The loader now parses the transition name into the starting fork, the activating fork and the activation criteria (timestamp for `AtTime`, block number for `At`). The genesis and the blocks before the activation run on the starting fork. The first block that meets the criteria hands the chain to the activating fork through its `apply_fork`, the same hook the mainnet sync tool uses, and the remaining blocks run there. For the duration of the test the activating fork's `FORK_CRITERIA` is set to the fixture's criteria, the way a client's chain configuration would schedule it, so the fork detects its own fork block. A transition fixture is collected when the fork it activates is among the requested forks.

`ForkLoad` gains an `apply_fork` accessor, and the network parsing has unit tests.

Verified locally by filling every `not slow and primary_format` transition test up to Amsterdam and running `just validate-blocks` over the result; non-transition fixtures are unaffected.

### Related Issues or PRs

Prepares for #3535, whose fork-block nonce bump can only be exercised in the spec by a transition fixture.

### Checklist

### Cute Animal Picture

### DIFF
--- a/tests/json_loader/helpers/load_blockchain_tests.py
+++ b/tests/json_loader/helpers/load_blockchain_tests.py
@@ -1,17 +1,21 @@
 """Helpers to load and run blockchain tests from JSON files."""
 
+import re
+from contextlib import ExitStack
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Optional, Tuple
 from unittest.mock import call, patch
 
 import pytest
 from _pytest.config import Config
 from ethereum_rlp import rlp
 from ethereum_rlp.exceptions import RLPException
-from ethereum_types.numeric import U64
+from ethereum_types.numeric import U64, U256, Uint
 
 from ethereum.crypto.hash import keccak256
 from ethereum.exceptions import EthereumException, StateWithEmptyAccount
+from ethereum.fork_criteria import ByBlockNumber, ByTimestamp, ForkCriteria
 from ethereum.state_mpt import close_state
 from ethereum.utils.hexadecimal import hex_to_bytes
 from ethereum_spec_tools.loaders.fixture_loader import Load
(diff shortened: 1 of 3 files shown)


jochem-brouwer:
Follow-up on https://github.com/ethereum/execution-specs/pull/3552, now also tests transitions to directly add this to coverage. CC @danceratopz :smile: :+1: This was found in #3535 with a low coverage warning despite having #3552 on there.

spencer-tb (approved):
Woo! Lets go! Love to see coverage increments even the smallest amounts haha

LGTM, didn't see anything obvious (claude/codex like it also)!

danceratopz:
Nothing sticks out, but review incoming :-)

tests/json_loader/helpers/load_blockchain_tests.py
```diff
@@ -21,6 +25,20 @@
 from .exceptional_test_patterns import exceptional_blockchain_test_patterns
 from .fixtures import Fixture, FixturesFile, FixtureTestItem
 
+TRANSITION_NETWORK = re.compile(
```
danceratopz, line 28: This can be a lookup into the testing package's transition forks, which already declare these facts and whose names are the fixture networks: https://github.com/danceratopz/execution-specs/commit/193b416b0aacfe901a1e6c573356cf6d11940945

tests/json_loader/helpers/load_blockchain_tests.py
```diff
@@ -21,6 +25,20 @@
...
+    r"^(?P<from_fork>.+?)To(?P<to_fork>.+?)"
+    r"At(?P<by_time>Time)?(?P<value>\d+)(?P<kilo>k)?$"
+)
+"""
+Network name of a fork-transition fixture, such as `BPO2ToAmsterdamAtTime15k`
+or `BerlinToLondonAt5`: the chain starts on the first fork and the second
+fork activates at the given timestamp (`AtTime`) or block number (`At`).
+"""
+
+HEADER_NUMBER_INDEX = 8
```
danceratopz, line 38: The filler already writes the decoded block of every decodable invalid block under `rlp_decoded`, so the header can be read from there instead of walking the RLP: https://github.com/danceratopz/execution-specs/commit/a3645c796125b50815acc430082f537d7a7c6b3c

tests/json_loader/helpers/load_blockchain_tests.py
```diff
@@ -155,53 +249,85 @@ def runtest(self) -> None:
...
+    def _schedule_fork(
+        stack: ExitStack, load: Load, criteria: ForkCriteria
+    ) -> None:
+        """
+        Make the fork of `load` activate at `criteria` for the duration of
+        `stack`, so that the fork detects its own fork block as a client
+        with a matching chain configuration would.
+        """
+        fork_module = load.fork.hardfork.module("fork")
+        if hasattr(fork_module, "FORK_CRITERIA"):
```
danceratopz, line 326: Only London and the DAO fork bind `FORK_CRITERIA` in their fork module, so this is a no-op for every transition fixture CI fills today. Patching the package attribute as well, with a unit test: https://github.com/danceratopz/execution-specs/commit/778aee13fdd55b8c580ffad2903320e613c41da2

tests/json_loader/helpers/load_blockchain_tests.py
```diff
@@ -155,53 +249,85 @@ def runtest(self) -> None:
...
-            else:
-                add_block_to_chain(chain, json_block, load, mock_pow)
+        with ExitStack() as stack:
+            if self.transition is not None:
+                self._schedule_fork(stack, load, self.transition.criteria)
+
+            for json_block in json_data["blocks"]:
+                if (
+                    self.transition is not None
+                    and current is not load
```
danceratopz, line 275: Optional renaming suggestion.

`load` now means the activating fork's loader while `current` carries the old meaning, so `current is not load` reads as an implicit "not handed over yet". Renaming `load` to `activating`, or keeping an explicit pending transition, would make these lines self-describing. Two smaller ones in the same spirit: `_schedule_fork` could return a context manager instead of mutating the caller's stack, and `mock_pow` can stay a bool, since `proof_of_stake` cannot differ across any transition the loader accepts.

danceratopz (commented):
Nice, thanks Jochem! Suggestions for the three larger points are in jochem-brouwer/execution-specs#9.

Can you update the description? It says fork-block detection from `FORK_CRITERIA` is now exercised across a boundary in CI. It is not yet: only London and the DAO fork read it in their fork module and no London transition test exists, so that part becomes live with #3535.

`fork_loader.py` also gains accessors in #3535, it'll need a rebase if we apply the suggestions.

danceratopz (changes requested):

tests/json_loader/helpers/load_blockchain_tests.py
danceratopz: I took another look at this, and perhaps we could have snuck the rename in, but returning the context manager for `_schedule_fork` was just too off topic for this PR. So left it, would be nice to refactor in the future, but low prio of course.

danceratopz (approved):
Thanks @jochem-brouwer!

PR #3566, opened 2026-09-11
feat(tests,specs): eip7928 extended coverage

fselmo:
### Description

Closes out the EIP-7928 section of the Amsterdam test-release tracker. This doesn't claim to exhaust all possible BALs cases but should get us in great shape for the #3217 scope.

One spec fix, flagged by @gurukamath: the BAL builder diffed each merge unit against the running block state, so two pre-execution system calls at index 0 that undid each other recorded a `(0, 0)` storage change instead of a read. It now diffs against the state at the start of the block access index. This doesn't affect any existing test, nor does it affect any behavior on mainnet today, but is an inconsistency in the spec implementation and something that now has coverage for any pre-execution contracts that interact with each other in this way. The t8n also ran the two pre-execution calls in the opposite order from `apply_body`, which this PR aligns and pins down with a test.

Increasing coverage on the gaps a coverage diff and a few CC w/ Fable iterations turned up:

- netting at both boundary indices, and a failing pre-execution system call with its "invalid block" counterpart
- value calls that fail (the besu#11124 / erigon#23670 shape), on both credit paths of `generic_call()`
- addresses that must appear exactly once (sender == coinbase) or not at all (`EXTCODE*` on a delegation, `BLOCKHASH` against the EIP-2935 contract, a precompile named in a tx access list but never actually called)
- burned blob fees leaving the sender and reaching no account
- value reaching a precompile by `CALL` and `CALLCODE`, not only as a transaction's recipient
- the first BAL-asserting deposit request
- six intra-account ordering negatives, with the `reverse_*` modifiers behind them (more invalid tests)
- adds mainnet tests (`test_eip_mainnet.py`) for tracker #1878

Where an existing test already ran a relevant scenario, a BAL expectation was added to pin the Amsterdam behavior, rather than re-writing that same coverage under Amsterdam. Each new claim was broken / mutated once in what it pins to confirm it doesn't silently still fill.

### Related Issues or PRs

- Part of #3217 (EIP-7928 section); tracker #1878

### Checklist


fselmo:
@spencer-tb, merged your PR and did a touch of cleaning up on top. Ready for re-review on your end.

fselmo:
Actually, you can review but please don't merge yet. There's a quirk about using pre-execution contacts since those EIPs don't require the client to make a system call. It says they may update the state directly. I want to make sure we are writing the best cases around this. Will update here tomorrow with more details.

fselmo:
Good for review + merge again! Reconciled the system contract call tests by checking same-value slot replacement in the pre-exec contract test and moving the cross-system-contract interaction test to the post-execution system calls. This ends up working well imo.

spencer-tb (approved):
LGTM!

PR #3573, opened 2026-09-11
feat(tests): in-range signature that recovers no public key

chfast:
### Description

`r` is read as the x-coordinate of the ephemeral point R, and only about half the values in [1, n) are x-coordinates of a curve point at all. For the other half there is no R, hence no public key, with no range check violated anywhere. `r = 5` is the smallest such value: 5**3 + 7 is a quadratic non-residue mod p.

Neither consumer of `secp256k1_recover` was covered for this input:

- `test_bad_v_r_s` covers only the RANGE rules (`v` below 27/35, `r` or `s` at or above secp256k1n, `s` above the EIP-2 halfway point). Every case pairs an on-curve `r` with an out-of-range `v` or `s`, so the range check always fires first and recovery is never reached.
- `test_precompiles` parametrizes `r` over valid x-coordinates, zero and the group order, all of which are on the curve or rejected by range.

The precompile case did already occur, but only by accident: `ported_static/stPreCompiledContracts2/test_call_ecrecover_unrecoverable_key` passes an ASCII string as `r`, which happens to be off the curve. Nothing in the file or its name says so, so it was neither labelled nor deliberately maintained.

The two consumers treat the failure differently, which is what makes the pair worth having: the precompile catches `InvalidSignatureError` and returns empty, while transaction validation lets it propagate and rejects the transaction. A client can easily guard one and not the other.

The transaction half is a state test rather than a `transaction_test`, even though the failure needs no state transition, because no `consume` runner reads a `TransactionFixture`. A state test additionally yields the engine fixtures that drive `engine_newPayload`, which is where a client reached through an unintended path shows it.

The EELS mapping is required for `fill` to accept the new transaction case; without it EELS raises a dedicated `InvalidSignatureError('r is not the x-coordinate of a point on the secp256k1 curve')` that no mapping recognises and the fill is refused.

### Related Issues or PRs

https://github.com/besu-eth/besu/issues/11278
Found by evmone-fuzz.

### Checklist

### DIFF
--- a/packages/testing/src/execution_testing/client_clis/clis/execution_specs.py
+++ b/packages/testing/src/execution_testing/client_clis/clis/execution_specs.py
@@ -264,6 +264,14 @@ class ExecutionSpecsExceptionMapper(ExceptionMapper):
         TransactionException.LOG_MISMATCH: "LogMismatchError",
     }
     mapping_regex: ClassVar[Dict[ExceptionBase, str]] = {
+        # An in-range r that is not the x-coordinate of any secp256k1 point:
+        # recovery has no solution, so the transaction has no sender. Distinct
+        # from the "bad r"/"bad s"/"bad v" range failures in
+        # mapping_substring, and raised from ethereum.crypto.elliptic_curve
+        # rather than from the transaction validation itself.
+        TransactionException.INVALID_SIGNATURE_VRS: (
+            r"InvalidSignatureError\('r is not the x-coordinate"
+        ),
         # Temporary solution for issue #1981.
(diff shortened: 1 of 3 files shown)


spencer-tb (approved):
LGTM! Nice addition. I ran the fixtures against geth, nethermind, besu and reth, all four reject the transaction and skip the authorization. I pushed client exception mappings, a type-1 variant and an explicit 7702 case on top.

One note, evmone didn't let me fill these new tests, so maybe worth a look on your end :)

PR #3574, opened 2026-09-11
refactor(tests): fold the simple ported ECRECOVER cases into test_precompiles

chfast:
### Description

The static ECRECOVER fillers arrive as ~55 near-identical files across `stPreCompiledContracts2` and `stStaticCall`. Most are about something other than the precompile's arithmetic: gas boundaries (2999/3000/none), input length (0input, check_length, short_buff), output aliasing (overlapping_input_output, complete_return_value), or misaligned input windows. Those stay, and with them the CALLCODE and STATICCALL paths into the precompile.

The rest are plain known-answer vectors: write 128 bytes, call address 1 with ample gas, record what comes back. Each was ported three times, once per call opcode, though the opcode is not what they exercise. `test_precompiles` is the maintained home for that shape, so the 26 such files are removed and every input they carried that was not already present is added to it, named for what it is:

- `valid_signature_3` recovers 0xe4319f4b… (Ecrecover3)
- `msg_hash_high_byte_zeroed` recovers 0xa0b29af6… (EcrecoverH_prefixed0)
- `s_high_byte_zeroed` recovers 0xb4950a7f… (EcrecoverS_prefixed0)
- `r_high_byte_zeroed` is empty (EcrecoverR_prefixed0)
- `all_high_bytes_zeroed` is empty (Ecrecover80)
- `unrecoverable_key` is empty (CallEcrecoverUnrecoverableKey)
- `v_eq_1` is empty (Ecrecover1)
- `zero_input` is empty (CallEcrecoverInvalidSignature)

`v_eq_1` closes a real gap: every existing parameter passed 27 or 28, so a v outside the pair was untested. `zero_input` fails the same rule with every byte zero. The other three empty cases fail because r is not the x-coordinate of any curve point, but each reaches that from a different input, so each is kept as its own vector.

Only two inputs were genuine duplicates and are dropped outright:

- Ecrecover0 is `valid_signature_1`, byte for byte.
- EcrecoverV_prefixed0 writes `0x001c` where Ecrecover0 writes `28`; MSTORE zero-pads to 32 bytes, so the two inputs are the same and the filler's "leading zero" intent was already vacuous.

The `ported_from` marker on `test_precompiles` re-declares all 26 filler paths the removed files carried, so `fill --show-ported-from` still reports the same 52 paths for ECRECOVER as before. Note this claims the CALLCODE and static_Call fillers for a test that reaches the precompile with CALL only: their inputs are covered, their dispatch is not.

Expected outputs were computed with `ethereum.crypto.elliptic_curve` rather than copied from the ported posts, and the fill confirms them.

### Related Issues or PRs

N/A.

### Checklist


spencer-tb (approved):
LGTM, added a small check that failed recovery leaves the output buffer unchanged, as the originals did!

Thanks for getting started on this :)

PR #3585, opened 2026-09-15
fix(test-benchmark): clamp nested frame depth to budget and assert receipt status

### LINKED ISSUE #3547: `test_memory_*` benchmark refactor
The high depth variants in `test_memory` run out of gas mid descent and revert unnoticed, so the benchmark never actually reaches the advertised depth. 

Fixed here if you want it: https://github.com/spencer-tb/execution-specs/commit/b07128d21cd1c2f8b5c33ff677c6bcd921228020, feel free to cherry-pick.

_Originally posted by @spencer-tb in https://github.com/ethereum/execution-specs/pull/3349#pullrequestreview-5077039726_

zexoverz:
### Description
This is @spencer-tb's fix from his review on #3349, cherry-picked as he suggested in the issue, authorship kept. Each tx now carries the deepest depth its budget can fund and the fill asserts every receipt is status 1, so an OOG mid descent fails the fill instead of passing silently.

Filled `test_nested_frame_memory` on Amsterdam at 100M. Before the change, the 64K/256 case fills with receipt status 0 if you add the assertion. After it, all nine cases pass. At mem 0 and 8K every tx still descends the full 256 frames. At 64K the 16,777,216 tx cap only funds 175 frames, so the clamp takes 256 down to 175 there, which is the point of the change rather than a gap in it.

### Related Issues or PRs
Closes #3547

### Checklist


spencer-tb (approved):
LGTM! cc-ing @LouisTsai-Csie for merge, maybe it was missed or there was a reason to not include it :)

LouisTsai-Csie (approved):

PR #3589, opened 2026-09-15
feat(tests): fold the ECRECOVER v and curve-order cases into test_precompiles

chfast:
### Description

Port ecrecoverWeirdV, CallEcrecover_Overflow and CALLCODEEcrecoverV_ prefixedf0, whose only variation is the signature fields, into the existing parametrization: v values a real signature could carry but the precompile must still reject (the recovery id itself, the EIP-155 encodings including a genuine chain-id-1 signature, and words whose low byte alone reads as 27 or 28), and r and s one step either side of the curve order.

Widen the return window to two words and seed both. The precompile returns a single word, so the second word must keep its sentinel on every case, which is what ecrecoverWeirdV checked with its 256 byte window, and a failed recovery must leave the first word seeded too.

Keep the two observables the fillers had that the parametrization was missing: the call's success flag, since a rejected signature still returns successfully, and the forwarded gas, now the precompile's own fork-derived price so that charging more fails the call.

### Related Issues or PRs

N/A.

### Checklist


fselmo (approved):
This is great, ty. One added benefit of in-housing it is we can extend this to `with_all_call_opcodes`, which I did in a commit on top. Thanks 🙏🏼

PR #3590, opened 2026-09-15
feat(tests): port the repeated ECRECOVER call fillers

chfast:
### Description

Fold Call50000_ecrec and static_Call50000_ecrec into a single parametrized test over every call opcode, with ten iterations instead of fifty thousand. The fillers repeat the call until the transaction runs out of gas, so their whole post-state is the empty one a reverted transaction leaves behind; running the loop to completion instead lets the residue be checked.

What the precompile receives, stipend included, is exactly one gas short of its price, so every call fails: the success flags accumulate to their seeded base, the loop counter proves all ten iterations ran, the seeded return word survives and the precompile receives no value. One gas more and every call recovers instead. The input window stays as wide as the fillers made it, far wider than the 128 bytes the precompile reads.

### Related Issues or PRs

N/A.

### Checklist


LouisTsai-Csie (approved):
Thank you! I append a small refactor

PR #3591, opened 2026-09-15
chore(skill): apply follow-up review suggestions to write-test and lint

danceratopz:
### Description

Follow-up to #3562 with the review suggestions that were still pending when it merged.

`write-test`:

- Rename "Pinning Intent" to "Test Expectations". The intro paragraph is replaced by a definition and a table of which object carries each expectation field (`Transaction`, `Block`, `StateTest`, `post`), which also settles `header_verify=` vs `blockchain_test_header_verify=` and `error=` vs `exception=`. The bullets use expectation vocabulary instead of "pin", and the remaining uses of "pin" elsewhere in the file are reworded.
- Use `error=` in the Exception Testing section, which is about `Transaction`; `exception=` is the `Block` field.
- Give the full path to the covariant marker decorator and trim the parts of that bullet that repeat the Markers and Test Structure sections.
- Drop the cross-reference parenthetical in Gas Cost Expectations that repeated the EELS exception-mapping point.
- Write `storage.set_expect_any(key)` and `storage.store_next(expected)` as instance calls.

`lint`:

- Restore `just whitelist <word>` for spellcheck false positives. The recipe keeps `whitelist.txt` sorted and is what `just spellcheck` prints.

### Related Issues or PRs

Follow-up to #3562.

### Checklist

### Cute Animal Picture
image


.agents/skills/write-test/SKILL.md
```diff
@@ -123,7 +132,7 @@ Never hand-reconstruct a gas amount by summing `fork.gas_costs()` constants (`NE
...
+- **Cover the family, not just the instance.** Before fixing the subject of a test, ask whether it is one member of a set the framework already parametrizes: call opcodes, create opcodes, precompiles, system contracts, request classes, tx types. If so, use that `with_all_*` marker, narrowed with `selector=` when only part of the set can reach the behaviour (only `CALL` and `CALLCODE` carry a value, so only they can fail a sender-balance check). The nearest test proving the same rule for one member usually already carries the marker, so read a model test's decorators and not just its body. If no marker covers a real family, add a covariant marker (`covariant_decorator` in `packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py`) in its own change rather than hand-writing the list in the test.
```
fselmo:
```suggestion
- **Cover the family, not just the instance.** Before fixing the subject of a test, ask whether it is one member of a set the framework already parametrizes: call opcodes, create opcodes, precompiles, system contracts, request classes, tx types. If so, use that `with_all_*` marker, narrowed with `selector=` when only part of the set can reach the behaviour (only `CALL` and `CALLCODE` carry a value, so only they can fail a sender-balance check). The nearest test proving the same rule for one member usually already carries the marker, so read a model test's decorators and not just its body. If no marker covers a real family, propose a covariant marker (`covariant_decorator` in `packages/testing/src/execution_testing/cli/pytest_commands/plugins/forks/forks.py`) in its own change rather than hand-writing the list in the test.
```

nit: I deliberately changed this to `propose` in the last version because I had something like this and it wanted to add a marker for an abstraction that was only used in 2 or 3 places. I think this is a bit more than a semantic thing and Claude agreed that this would push their behavior to _ask_ if it's warranted rather than go straight into implementation if you use something like `auto` mode with CC. Take or leave though, just want to avoid this friction and waste of tokens in case it's overengineering something we don't plan on using a whole lot.

fselmo (approved):
lgtm added a note on a possible AI over-engineering footgun

fselmo:
also missing a cute animal pic, smh

danceratopz:
Thanks @fselmo! Suggestion applied, makes sense, thanks for catching that. Blocking missing cute animal picture resolved!!!

PR #3592, opened 2026-09-15
chore(test-cli): check EIP reference versions up to Amsterdam

fselmo:
### Description

Every local `check_eip_versions` run and the CI collect job skipped all Amsterdam suites. This was leading to some bad hashes (commit sha instead of blob sha) passing by blindly. This updates the CI job to run these checks `--until Amsterdam`.

The only CI job, though, runs `--collect-only` and asserts a count floor. It never executes the version assertions, for any fork. A stale pin still won't fail CI after this... but any local checks would catch this. Perhaps this is something we should think about, but I think we may have had this at some point, posting to an issue? (I think this was before my time in STEEL), and it produced way too much noise? @danceratopz am I remembering this correctly?

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

Screenshot 2026-09-15 at 11 21 10


packages/testing/src/execution_testing/config/check_eip_versions.py
```diff
@@ -6,5 +6,5 @@
 class CheckEipVersionsConfig(BaseModel):
     """A class for accessing configurations for `check_eip_version`."""
 
-    UNTIL_FORK: str = "Prague"
+    UNTIL_FORK: str = "Amsterdam"
```
danceratopz: Most of our fork config is now set in the `Justfile` via `latest_fork` (except for this and our `mkdocs` flow), perhaps we can use default `fill` behavior in the CLI and then, if we ran it in CI, specify which fork to run until via that variable:
https://github.com/ethereum/execution-specs/blob/8acae1b07dd1fc22565569cc168c0738f9086cf4/Justfile#L19

But hey, this is absolutely fine for now!

danceratopz (approved):
Thanks for catching this! Yes, in execution-spec-tests we had the [Check EIP Versions action](https://github.com/ethereum/execution-spec-tests/actions/workflows/check_eip_versions.yaml) which ran weekly and generated [these issues](https://github.com/ethereum/execution-spec-tests/issues?q=is%3Aissue%20state%3Aclosed%20label%3A%22automated%20issue%22%20eip%20spec%20references%20outdated). 

They became an annoyance. I think the main reason they were so annoying was the frequency of changes that weren't spec-bearing, so you would just create a pr to silence the action. Perhaps monthly would be a better tact, or, even better, update the CLI too to actually perform the change and create a PR instead of an issue. This could potentially trigger a rebase though if the hashes across the automated and other test pr both change (but not to the same hash); with this in mind, perhaps weekly is better after all.

Perhaps our multibranch (`eips/*`) world will protect us from too much noise if we only run it on the default development fork, but then it has less impact? 

One comment below, but this is definitely and improvement for now! 🙂

danceratopz:
> Perhaps monthly would be a better tact, or, even better, update the CLI too to actually perform the change and create a PR instead of an issue. 

Yeah, this isn't a great idea if the changes are actually spec-bearing though! :sweat_smile: gn!

fselmo:
> Most of our fork config is now set in the `Justfile` via `latest_fork` (except for this and our `mkdocs` flow), perhaps we can use default `fill` behavior in the CLI and then, if we ran it in CI, specify which fork to run until via that variable:

Took this as a one-source-of-truth suggestion and applied it bc I like it 😅. But I put the default in the checker plugin rather than the Justfile. A plain `uv run check_eip_versions` doesn't go through the Justfile, so a default there would still stop at the last deployed fork for anyone running the tool directly. The hard-coded `UNTIL_FORK` is gone though.

The plugin now defaults to every fork in the fork registry, including forks under development, so `uv run check_eip_versions` and `just check-eip-versions-collect` check the same range and nothing needs bumping when a fork is added. `--fork` and `--until` still work as before. As a side effect, `--fork` works again! lol. It used to error because the CLI always added `--until` and they would clash.

bonus: I also removed the "Automated Checks via GitHub Actions" docs section, which described a daily workflow that is nonexistent? 👀

will ask for re-review 👀

danceratopz (approved):
Thanks @fselmo! This is much nicer! I made two small follow-up commits directly to your branch, one fix for module collection and one CLI help output improvement. 

About this:

> The plugin now defaults to every fork in the fork registry, including forks under development, so `uv run check_eip_versions` and `just check-eip-versions-collect` check the same range and nothing needs bumping when a fork is added. `--fork` and `--until` still work as before. As a side effect, `--fork` works again! lol. It used to error because the CLI always added `--until` and they would clash.

I only realized this during the second review that while these flags do correspond to the fill behavior, they might not do what a user expects here (i.e., only check Amsterdam EIP versions if `--fork=Amsterdam`). They will limit the checks, but only to all tests until Amsterdam.

So if you really only want to check Amsterdam tests, you need a positional argument that specifies the test folder, just like for `fill`. I added this to the CLI help output in a commit directly to this branch to nudge users in the right direction.

While we dug deeper in the review, Claude noticed that some modules were ignored, so I had him fix this up (it's a two line fix). Hope you don't mind me pushing this upfront!

fselmo:
Thanks @danceratopz! No qualms at all with you pushing. I'm a big proponent if you've done the work locally with AI or what-have-you, then pushing it and reviewing / changing makes things go faster. If the work is already done, no need to comment it – just push and we can still discuss it.

I like the changes 👍🏼... I just refactored both tests here to be in the same file since they have some shared setup. Nothing else. Will merge when CI passes.

PR #3594, opened 2026-09-15
fix(ci,test): fix issues with PR CI runs in focil branch

fselmo:
### Description

Keep strict checks on inclusion lists but fix CI for the time while we are hijacking amsterdam's t8n for FOCIL work.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture

Screenshot 2026-09-16 at 13 55 43


fselmo:
@gurukamath I'm not really privy to all the details of the `amsterdam` path for `bogota` EIPs so I want to make sure this is an OK change. I'd like to try to get CI passing for FOCIL while we don't have a proper setup for `bogota` and I want to also make sure to keep this strict check on inclusion lists. I think this is a decent compromise on both of these.

I think the failing pypy run here is an OOM issue. I'm ok with this for now I suppose... but will investigate this as well if I get some more time. At least the other CI jobs are better.

fselmo:
Ok pypy issue fixed as well as the json loader. This was an interesting find, from a quick Claude scan:

> **Inclusion-list fixture variants were built for every test at every fork, and dropped only after collection, because the rule that drops them needs a fork**. Collection for fill-pypy and json-loader grew from 224k to 319k items, and peak memory per process from about 1.6 GB to 2.0 GB. It showed up as a PyPy MemoryError and xdist workers dying mid-collection, not as anything about Bogota. Fixed by applying the rule when tests are generated.

spencer-tb (approved):

PR #3601, opened 2026-09-16
feat(ci): cut devnet fixture releases from any branch

spencer-tb:
### Description

Allow devnet fixture releases from any explicitly selected branch, including EIP branches, avoiding duplicate `devnets/` branches. A branch remains required; a branch under `devnets/` must still follow `devnets/<feat>/<n>` with version major `<n>`, and only branches outside that namespace skip the check. Release tags are unchanged.

The selected branch must contain the updated validator. Selecting the workflow with `--ref` alone is insufficient.

Set `branch` to the source branch you want to release. For example, draft a release from the EIP-8141 branch:

```bash
gh workflow run release_fixtures.yaml --repo ethereum/execution-specs \
  -f feature=frames-devnet -f version=v0.4.0 \
  -f branch=eips/amsterdam/eip-8141
```

Publishing the draft creates `tests-frames-devnet@v0.4.0` at the EIP branch commit used for the build. Choose an unused version.

### Related Issues or PRs

#3384 update the devnet-release skill wording.

### Checklist


.github/scripts/generate_build_matrix.py
```diff
@@ -121,14 +126,13 @@ def validate_inputs(feature: str, version: str, branch: str, evm: str) -> None:
...
+        # Only a `devnets/<feat>/<n>` branch encodes a devnet number to
+        # cross-check the version major against.
         match = DEVNET_BRANCH_RE.match(branch)
-        if not match:
-            fail(
-                f"could not parse a devnet number from branch '{branch}' "
-                "(expected devnets/<feat>/<n>, e.g. devnets/bal/7)"
-            )
+        if match is None:
+            return
```
danceratopz: Returning on any non-matching branch also skips the major check for malformed branches inside `devnets/`, which is where the number should be present.

All of these were rejected before, but not after this change (they should return 1), verified at this head:

| dispatch | now |
| --- | --- |
| `bal-devnet v99.0.0 devnets/bal/7-benchmark` | return code=0 |
| `bal-devnet v99.0.0 "devnets/bal/7 "` | rc=0 |
| `bal-devnet v99.0.0 Devnets/bal/7` | rc=0 |
| `bal-devnet v99.0.0 forks/amsterdam` | rc=0 |

`devnets/bal/7-bench`, `devnets/bal/7-benchmark` and `devnets/glamsterdam/6-benchmarks` are real branches on origin, so picking one drafts a mis-versioned release, caught only by eyeballing the draft after a 1440 minute fill.

Gating on the namespace keeps the intended loosening, and removes the mid-function `return`:

```python
if branch.startswith("devnets/"):
    match = DEVNET_BRANCH_RE.match(branch)
    if match is None:
        fail(...)  # as before
    devnet_number = int(match.group(1))
    ...
```

Benchmark releases from `devnets/bal/7-bench` are unaffected, since `benchmark` is not a `*-devnet` feature.

docs/dev/releasing_tests.md
```diff
@@ -29,14 +29,15 @@ Input validation runs in [`generate_build_matrix.py`](https://github.com/ethereu
...
+- a `*-devnet` release missing a `branch`, or one whose `branch` follows the `devnets/<feat>/<n>` scheme with a `version` major that does not equal the devnet number `<n>` (so `feature=bal-devnet branch=devnets/bal/7` must use `version=v7.*.*`). Any other branch (e.g. `eips/amsterdam/eip-8141`) is accepted as-is.
 
 ## Devnet releases
 
-Devnet releases must use a `<feat>-devnet` feature name (e.g. `feature=bal-devnet`) and must specify the branch to release from. Devnet branches follow the `devnets/<feat>/<n>` scheme (e.g. `devnets/bal/7`), and the `version` major must match the devnet number `<n>` in the branch:
+Devnet releases must use a `<feat>-devnet` feature name (e.g. `feature=bal-devnet`) and must specify the branch to release from. Any branch works: a dedicated `devnets/<feat>/<n>` branch (e.g. `devnets/bal/7`), or the EIP feature branch itself (e.g. `eips/amsterdam/eip-8141`) when the devnet tracks it one-to-one. The `version` major is the devnet number either way (see the [Versioning Scheme](../running_tests/releases.md#versioning-scheme)); it is only validated when the branch encodes it, so `branch=devnets/bal/7` must use `version=v7.*.*`:
```
danceratopz: The caveat from your description is missing here, so this documents a dispatch that fails today.

`setup` checks out `ref: inputs.branch` first (`release_fixtures.yaml:81-84`), then runs the script from that tree (:136), so the validator that runs is the one on the selected branch. `origin/eips/amsterdam/eip-8141` is 96 commits behind `forks/amsterdam` and still has the old hard fail, so the second command below errors with "could not parse a devnet number".

Worth stating: the scripts run from the selected branch, not the workflow ref, so the branch must contain this commit and be current with its `forks/<fork>` base. A stale branch also fills without whatever landed on the base since. Note the devnet and EIP branches carry a divergent copy of this script (`parse_fork_bounds` vs `parse_until_fork`), so the hunk needs propagating to each.

Separately, this sentence is now 68 words and retracts its own rule in the same breath. Worth splitting, and worth saying how to pick the devnet number when the branch does not encode it, since that is now the common case.

docs/dev/releasing_tests.md
```diff
@@ -14,7 +14,7 @@ gh workflow run release_fixtures.yaml -f feature=<feature> -f version=vX.Y.Z [-f
 | ---------- | ----------------- | ---------------------------------------------------------------------------------------------------- |
 | `feature`  | yes               | Feature name, e.g. `tests`, `benchmark`, or a `<feat>-devnet` name.                                   |
 | `version`  | yes               | Release version `vX.Y.Z` (validated against `^v[0-9]+\.[0-9]+\.[0-9]+$`). Tagged as `tests-<feature>@<version>` (the `tests` feature tags as `tests@<version>`). |
-| `branch`   | devnet only       | Branch to build and release from. Optional for non-devnet features; **required** for devnet releases. |
+| `branch`   | devnet only       | Branch to build and release from (any branch). Optional for non-devnet features; **required** for devnet releases. |
```
danceratopz: Not introduced here, but while this row is being touched: "devnet only" can be read as "only usable for devnet features", which is not true. It is optional for fresh non-devnet fills, and inverted for cached releases, where `resolve_cached_release.py` rejects a `branch`.

.github/scripts/tests/test_release_scripts.py
```diff
@@ -146,13 +146,19 @@ def test_devnet_without_branch_fails(self):
...
+            BUILD_MATRIX_SCRIPT,
+            "frames-devnet",
+            "v0.1.0",
+            "eips/amsterdam/eip-8141",
         )
-        assert result.returncode == 1
-        assert "could not parse a devnet number" in result.stderr
+        assert result.returncode == 0
+        out = parse_matrix_output(result.stdout)
+        assert out["feature_name"] == "frames-devnet"
```
danceratopz, line 161: This does cover the new early return, so the path is fenced.

What is lost is the only assertion that a near miss gets rejected, since `bal-devnet-7` is now indistinguishable from a real EIP branch. A real near miss would fence the namespace gating suggested on the script:

```python
def test_malformed_devnets_branch_fails(self):
    """Verify a `devnets/` branch with no devnet number is rejected."""
    result = run_script(
        BUILD_MATRIX_SCRIPT, "bal-devnet", "v99.0.0",
        "devnets/bal/7-benchmark",
    )
    assert result.returncode == 1
    assert "could not parse a devnet number" in result.stderr
```

`branch=" "` also exits 0 now, where it used to fail with a helpful message.

danceratopz (commented):
Nice! Below are a few comments from Claude. The first one about not checking well-formed devnet branch names seems worth fixing. The other three are docs/tests and less important.

He also mentioned: 

> Follow-up, not for this PR. Nothing guards against a version whose tag already exists, and this is the first path with a fully unconstrained major, so a collision only shows up at `gh release create` after the fill.

danceratopz (approved):
SGTM

PR #3598, opened 2026-09-16
fix(test-client-clis): match besu's new insufficient funds messages

danceratopz:
### Description

Besu [#11272](https://github.com/hyperledger/besu/pull/11272) reworked the sender balance validation in `MainnetTransactionValidator`. It renamed `UPFRONT_COST_EXCEEDS_BALANCE` to `UPFRONT_GAS_COST_EXCEEDS_BALANCE`, whose message now reads `up-front gas cost` rather than `up-front cost`, and it split the value transfer off into a check of its own, `INSUFFICIENT_FUNDS_FOR_TRANSFER`, which reports `transfer value`:

```
transaction invalid transaction up-front gas cost 0x1313d4 exceeds transaction sender account balance 0x1313d3 for sender 0xaa28d3...
transaction invalid transfer value 0x1 exceeds transaction sender account balance 0x43e38 for sender 0xba1c78...
```

Neither matches the `INSUFFICIENT_ACCOUNT_FUNDS` pattern in `BesuExceptionMapper`, so every test expecting that exception fails with `Undefined exception message`. Which of the two messages a test gets depends on the transaction: the gas branch reports when the gas cost alone exceeds the balance, the value branch when the gas cost fits but the value does not.

This PR matches both messages and keeps the pre-#11272 wording so that older Besu builds continue to map.

Detected on the [Glamsterdam Hive dashboard](https://hive.ethpandaops.io/#/group/glamsterdam?suites=eels%2Fconsume-enginex), where the `consume-enginex` run of 2026-09-15 reported 1280 Besu failures against 1 in the preceding `consume-engine` run. The gap was Besu's build, not the simulator: `consume-engine` last ran two days earlier on `78ef664`, before #11272. The [`glamsterdam-quick`](https://hive.ethpandaops.io/#/group/glamsterdam-quick) `consume-engine` run on the same `18c4d62` build as the failing `consume-enginex` run shows the same undefined message, so this affects every `consume` simulator.

Validation: `just static`, and a local Hive A/B of this commit against its parent, using `consume-enginex`, the `ethpandaops/besu:main` image (`v26.9-develop-0c30148`, six commits after #11272) and the `tests@v20.0.2` fixtures:

```
--sim.limit '.*(test_insufficient_balance_blob_tx|test_no_src_account|test_sender_balance|test_eoa_empty_paris).*'

parent  1135 tests, 916 failed, every one `Undefined exception message` for INSUFFICIENT_ACCOUNT_FUNDS
[...]

PR #3607, opened 2026-09-16
fix(tests): raise the block gas limit in test_tx_gas_limit for EIP-7928

danceratopz:
### Description

This PR fixes an old test's expectation for Glamsterdam that was detected while debugging fails on the [Glamsterdam Hive dashboard](https://hive.ethpandaops.io/#/group/glamsterdam).

The `tests/frontier/validation/test_transaction::test_tx_gas_limit` test case checks that clients reject a transaction whose gas limit exceeds the block’s gas limit. It sets the block limit to 21,000 and the transaction limit to 21,001, expecting `TransactionException.GAS_ALLOWANCE_EXCEEDED`.

Amsterdam introduces another constraint through EIP-7928: the block gas limit also determines the maximum size of the block access list (BAL). At 21,000 gas, the BAL can contain only `21000 // 2000 = 10` items. However, system-contract accesses alone produce 25 items, even in a block with no transactions.

The original fixture therefore has two independent reasons for rejection. Clients that check transaction gas first report the expected error and pass; clients that check BAL size first can correctly reject the block but fail the test’s exception matching. The observed go-ethereum, reth and besu failures fall into the latter category:

```text
reth  [BlockException.BLOCK_ACCESS_LIST_GAS_LIMIT_EXCEEDED] block access list item cost
      exceeds gas limit: items=25, max_items=10, gas_limit=21000
geth  [BlockException.INVALID_BAL_HASH | ...] invalid block access list: block access list
      exceeds size constraint: items=25, limit=10 (block gas limit 21000 / 2000)
```

The fix derives the block gas limit from the fork, ensuring it covers both the empty-block BAL and the transaction’s intrinsic gas:

```python
block_gas_limit = max(
    fork.minimum_block_gas_limit(),
    fork.transaction_intrinsic_cost_calculator()(),
)
```

The transaction gas limit remains one above the block limit. For Amsterdam, these become 50,000 and 50,001: the supplied BAL fits, while the transaction still exceeds the block’s gas allowance. Earlier forks retain the original values.

Using [`minimum_block_gas_limit`](https://github.com/ethereum/execution-specs/blob/forks/amsterdam/packages/testing/src/execution_testing/forks/forks/forks.py) also lets the test follow future changes to the empty-block BAL size. The helper accounts for its item count and the per-item cost.

The expected exception remains `GAS_ALLOWANCE_EXCEEDED`. Accepting a BAL-size error as an alternative would allow clients to pass without validating the transaction’s gas allowance.

#### Validation

- `just static` passed.
- Refilling every fork through Amsterdam changed only the two Amsterdam fixtures; the other 19 were byte-identical.
- Local Hive A/B runs used go-ethereum `v1.17.6-unstable-854fbb8c` and reth `2.5.2-dev+fc44cfc5`:

| Client | Simulator | Failures before | Failures after |
|---|---|---:|---:|
| go-ethereum | `consume-engine` | 1/6 | 0/6 |
| go-ethereum | `consume-enginex` | 1/6 | 0/6 |
| reth | `consume-engine` | 1/6 | 0/6 |
| reth | `consume-enginex` | 1/6 | 0/6 |

The failing case before the change was `fork_Amsterdam` in every run.

### Related Issues or PRs

Follow-up to #3598, which fixed exception mapping in the same dashboard runs. This was the remaining besu `consume-engine` failure; go-ethereum and reth also failed this test under both simulators.

### Checklist

### Cute Animal Picture

Dog


danceratopz:
@spencer-tb I believe this should trigger a v8.1.5 to avoid confusion on the dashboard.

fselmo:
This is covered [here](https://github.com/ethereum/execution-specs/pull/3566/files#diff-b7dfff666beac8558a554f67c2278f64c6635d4e834be97ab6687f79940b12f1R40-R61) in the outstanding EIP-7928 coverage PR that is open for review

PR #3243, opened 2026-07-27
feat(tests): add EIP-7997 case where factory is not present at fork block

jochem-brouwer:
### Description
Implements tests where CREATE2 factory is not present at fork block. This also verifies that the factory account is not part of the BAL reads.

### Related Issues or PRs
Discussed in https://github.com/ethereum/pm/issues/2170

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

image

This is Harry who has been keeping me company on my home's ceiling for a while. Hi Harry!

PR #3079, opened 2026-07-01
feat(spec,tests): implement EIP-7997

LouisTsai-Csie:
## 🗒️ Description

Implement EIP-7997, merge EIP branch into `forks/amsterdam`.

## 🔗 Related Issues or PRs

issue #3068 

## ✅ Checklist

- [ ] All: Ran fast static checks to avoid unnecessary CI fails, see also [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) and [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/):
    ```console
    just static
    ```
- [ ] All: PR title have the form `<type>(<area>):`, where `<type>` and `<area>` come from an approrpriate `C-<type>`, respectively `A-<area>`, label. The title should match the a target squash commit message.
- [ ] All: Considered updating the online docs in the [./docs/](/ethereum/execution-specs/blob/HEAD/docs/) directory.
- [ ] All: Set appropriate labels for the changes (only maintainers can apply labels).
- [ ] Tests: For PRs implementing a missed test case, update the [post-mortem document](/ethereum/execution-specs/blob/HEAD/docs/writing_tests/post_mortems.md) to add an entry the list.
- [ ] Ported Tests: Add the following docstring to manually enhanced tests from `./tests/ported_static/`:
    ```text
	@manually-enhanced: Do not overwrite. Post-state expectations corrected
	manually (see PR #2784).
	````

#### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3614, opened 2026-09-17
chore(tests): expand and harden EIP-7997 factory coverage

spencer-tb:
### Description

Expand and harden EIP-7997 Deterministic Factory Contract coverage. The previous short-calldata gas flag could accept a late revert, and several failure tests could pass without executing the factory call. Exact gas measurements and seeded storage witnesses now reject those false positives.

Cover initcode failure and rollback, calldata boundaries and trailing data, creation contexts, same-transaction SELFDESTRUCT followed by attempted reuse, inherited static context, and value forwarding through a delegated EOA. Add keyless deployment and same-block use before activation, failure under Amsterdam's creation state-gas charge, and preservation of absent or foreign factory state across the transition.

Refresh the reference pin to the current EIP blob and correct stale titles and mechanism descriptions. The canonical bytecode matches the disassembly, including the jump target at 0x39. Amsterdam's fork transition remains a no-op, consistent with the current EIP and the rejected boundary-install proposals.

Complete the System Contract checklist with documented shared coverage and N/A rows. Add backward-compatible `exact=True` checklist markers so the aggregate invalid-input row can be credited without falsely crediting cryptographic validation.

Rename the suite directory to `eip7997_deterministic_factory_contract` and the fork address hook to `deterministic_factory_contract_address()`, updating factory constants, test names and all in-repository references.

### Related Issues or PRs

Related to #1988 and the EIP-7997 and 7928 x 7997 items in #3217. Builds on #2802, #3079, #3077 and #3243. The same-block test covers the factory analogue of #2270, not its separate EIP-7002/EIP-7251 system-call requirement.

### Checklist

- [x] Ran fast static checks with `just static`.
- [x] PR title follows the required `<type>(<area>): <title>` form.

### Cute Animal Picture

🦔


LouisTsai-Csie (approved):
Looks good to me, leaving some idea here:

- for test_factory_block_access_list, only happy path is tested out, we could have additional scenario for failed factory call, or address collision call.
- Add forwarded value version to `test_factory_via_eip7702_delegation`

PR #3615, opened 2026-09-17
feat(tests): cover same-block system contract deployment

### LINKED ISSUE #2270: Create test for same block system contract deployment
Actually, I was looking into this in a bit more detail and it looks like both EIP-7002 and EIP-7251 mention this case. See [here](https://github.com/ethereum/EIPs/blob/2175b810f3fd7aa07f356e5e9799985ad47d04a5/EIPS/eip-7002.md?plain=1#L727) for example.

It looks like if the deployment happens in the same block, the system contract needs to be read. So we need to read the state from the tracker and not the pre state. We definitely need a test for this though. Will create one.

_Originally posted by @gurukamath in https://github.com/ethereum/execution-specs/pull/2264#discussion_r2836460207_

spencer-tb:
### Description
Add regression coverage for a request system contract deployed and invoked in the same block. The post-execution system call must run against the state accumulated by the block's transactions: reading pre-state would find no code and invalidate the block. The shared EIP-7685 request test module covers the EIP-7002 withdrawal request contract and the EIP-7251 consolidation request contract. Each case removes the contract from pre-state, deploys it with the canonical transaction and sends a request in the same block, which the freshly stored fee inhibitor rejects (verified as the empty requests hash), then sends a request in the next block, which is accepted and verified in the header. The post-state pins the deployed bytecode and a contract balance of exactly one fee, proving the rejected request's value was refunded.

The test runs on Prague and Osaka. The Cancun→Prague fork block itself is already covered by `test_system_contract_deployment[deploy_on_fork_block]`. Under Amsterdam the pre-signed deployment transaction no longer fits its 250k gas limit (the init code's first `SSTORE` alone needs 98k state gas), so the test is bounded with `valid_until("Osaka")`; Amsterdam same-block deployment is covered by the EIP-8282 factory deployment tests.

Also fixes `requests_hash=` passed directly to `Block(...)` in three existing tests. The field is inherited from `Header` but never compared; only `header_verify` is. In the EIP-7002 and EIP-7251 `test_system_contract_deployment` the post-deployment request also paid no fee, so it was silently rejected and the final block's requests hash was the empty hash. It now pays the minimum fee and the hash is verified through `header_verify`.

### Related Issues or PRs
Closes #2270.

### Checklist


tests/prague/eip7685_general_purpose_el_requests/test_multi_type_requests.py
```diff
@@ -429,3 +451,69 @@ def test_invalid_multi_type_requests_engine(
...
+        int.from_bytes(request_type.system_contract_address, "big")
+    ]["code"]
+
+    blockchain_test(
+        pre=pre,
+        blocks=[
+            Block(
+                txs=[deployment_tx, request_tx],
+                requests_hash=Requests(request),
+            )
```
LouisTsai-Csie:
```suggestion
            Block(
                timestamp=fork.at_timestamp,
                txs=[deployment_tx, request_tx],
                header_verify=Header(requests_hash=Requests(request)),
            )
```

This is my understanding:

- `valid_at_transition_to("Prague")` only selects the `CancunToPragueAtTime15k` fork configuration; it does not place a block on either side of the boundary. With no timestamp set, `Block.set_environment` defaults to `parent_timestamp + 12`, so the only block lands at timestamp 12 (well below the 15,000 transition point) and executes entirely under `Cancun`. The filled fixture shows `"timestamp": "0x0c"` and "`requestsHash": null`: no EIP-7685 system call ever runs.
- `requests_hash=` on `Block` is a no-op. `Block` inherits the field from `Header`, but `generate_block_data` only consumes `block.header_verify` and `block.requests`; `block.requests_hash` is never read as an expectation. So the test's only request-related assertion silently does nothing. The canonical form is `header_verify=Header(requests_hash=...)`.

tests/prague/eip7685_general_purpose_el_requests/test_multi_type_requests.py
```diff
@@ -429,3 +451,69 @@ def test_invalid_multi_type_requests_engine(
...
         blocks=override_blocks,
     )
+
+
+@pytest.mark.parametrize_by_fork(
+    "request_type,deployment_path", lambda _fork: SAME_BLOCK_DEPLOYMENT_CASES
+)
+@pytest.mark.pre_alloc_mutable
+@pytest.mark.valid_at_transition_to("Prague")
+def test_system_contract_deployed_and_called_in_same_block(
```
LouisTsai-Csie, line 462: Shouldn't `EXCESS_INHIBITOR` be active during fork activation? The request would fail since it's only cleared at end-of-block. To ensure the contract deploys in the same activation block, we could use EXT* opcodes to check existence.

LouisTsai-Csie (changes requested):
Thank you! Leave some comments below:

spencer-tb:
Thanks @LouisTsai-Csie! Nice catches here, updated :)

gurukamath (approved):

LouisTsai-Csie (approved):

PR #3619, opened 2026-09-18
chore(tests): record measured EIP-8282 coverage and rename the inhibitor sentinel

spencer-tb:
### Description
Wait for ethereum/EIPs#12120 to be merged?

Replace the Codecov pointers in the EIP-8282 checklist evidence with measured coverage (both builder system calls and their type bytes fully executed; 458/463 test statements, the five misses documented) and rename `EXCESS_INHIBITOR` to `INHIBITOR` to match ethereum/EIPs#12120. The reference pin follows once that PR merges; the sys-asm build it cites differs only by a geas version bump, so the embedded bytecode and mined addresses are unchanged.

### Related Issues or PRs

#3217 (EIP-8282 section), #2988. Follows #3565.

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3620, opened 2026-09-18
chore(tests): record measured EIP-7928 coverage in the checklist evidence

spencer-tb:
### Description
Replace the instruction-style entries in the EIP-7928 checklist evidence with the measured result of a coverage-instrumented fill (2,505 cases): `block_access_lists.py` 234/236 and `state_tracker.py` 170/172 statements on the EELS side, 3,064/3,081 test statements, with every miss documented. Also drops the stale claim that `add_nonce_change`'s keep-existing branch is unexecuted, which #3566's tests now cover.

### Related Issues or PRs

#3217 (EIP-7928 section). Same shape as #3619 for EIP-8282.

### Checklist

### Cute Animal Picture


fselmo (commented):
Two things flagged here on a further review... from Claude:

> Keep the add_nonce_change keep-branch sentence. block_access_lists.py still reports the partial branch 463->467: if new_nonce > existing.new_nonce is never false. Statement coverage hides it because the return is reached through the update path. #3566 is the PR that measured this with --cov-branch and wrote that sentence, so its tests cannot be what covers it; the branch needs a later unit at the same index that rewrites the account without touching its nonce, which no mainnet system call does and no fixture builds. I'd restore the sentence and add --cov-branch to the re-verify command.

I also think we should remove the hard-coded total cases as this is prone to going stale / adding to confusion. Thoughts?

spencer-tb:
I think we can remove a lot of this process tbh

Closing for now :)

PR #3622, opened 2026-09-18
feat(tests): cover a null EIP-8070 custodyColumns and repin to the current EIP blob

spencer-tb:
### Description
Add `test_fcu_custody_columns_null`, pinning the 2026-08-12 EIP-8070 clarification that a `null` `custodyColumns` on `engine_forkchoiceUpdatedV4` is a blobpool no-op accepted with VALID, both on a client with no custody set and after one was configured, and `test_fcu_custody_columns_identical` for the other no-op case in the same rule, resending the current set. The blobs test now takes `custody_columns_updates`, an ordered list of `custodyColumns` values sent one forkchoice update each before `getBlobsV4`, where `None` sends an explicit `null`; leaving it unset sends no update, and an empty list is rejected by a validator with a unit test. Repin the suite to the current blob, which also carries the RLP empty-list clarification for elided blobs; the framework already encodes that. Execute-only suite, so `check_eip_versions` never runs its version item; the collection shows 9 cases and `just static` is clean.

These arms pin what the Engine API can observe: the value is well-formed, the forkchoice update completes with VALID, and `getBlobsV4` still serves every cell afterwards. The resulting custody set only steers devp2p sampling of peer-announced transactions and is not observable here; that needs a devp2p eth/72 harness, which hive does not have yet.

### Related Issues or PRs

#3217 (EIP-8070 section). Follows #2948, #3344, #3359.

### Checklist

### Cute Animal Picture


packages/testing/src/execution_testing/execution/blob_transaction.py
```diff
@@ -308,12 +309,13 @@ def _update_custody_columns(
...
-        A 16-byte bitmap must be accepted with a VALID payload status
-        (custody set update errors must not affect the forkchoice flow,
+        A 16-byte bitmap or an explicit `null` must be accepted with a
+        VALID payload status (`null` is a no-op for the blobpool, and
+        custody set update errors must not affect the forkchoice flow,
         per `engine_forkchoiceUpdatedV4`); any other length must be
         rejected with `-32602: Invalid params`.
         """
-        assert self.custody_columns is not None
+        assert self.custody_columns is not None or self.custody_columns_null
```
LouisTsai-Csie: The current logic allows both `custody_columns` to be set and `custody_columns_null` to be `True` at the same time. These two states should be mutually exclusive.

In that case, the implementation silently gives precedence to `custody_columns` and sends the bitmap, ignoring `custody_columns_null`. I don't think this combination should exist by the test framework in the first place.

For example, a test like the following should not be possible:

```python
blobs_test(
    pre=pre,
    txs=txs,
    get_blobs_version=4,
    cell_mask=ALL_CELLS_MASK,
    custody_columns=custody_columns,
    custody_columns_null=True,
)
```

It would be preferable to explicitly reject this combination, e.g. with an assertion that `custody_columns` and `custody_columns_null` cannot both be set.

LouisTsai-Csie (changes requested):
I'd like to understand how we verify the no-op semantics here. After sending `custodyColumns = null`, how do we confirm that the previously configured custody columns remain unchanged?

It seems like this state is not directly observable within our current test framework. Is this behavior verified somewhere outside this framework, and if so, where is the verification?

PR #3623, opened 2026-09-20
feat: add `--formats` flag to explicitly specify output fixture formats

### LINKED ISSUE #2959: Add `fill` flag to explicitly specify the output fixture formats
This means it'll be impossible to generate `blockchain_test_engine` fixtures from a benchmark test. Which is fine as no one uses it! 👍 

Tbh, in general though, I would love to have an explicit and clean way to tell `fill` which outputs to generate. Currently, the only real way to specify fixture format outputs is via `-m` which is a bit clumsy, so I'd prefer something like `--formats=blockchain_test,blockchain_test_enginex`. For another PR.

I would prefer this kind of change to be a smaller, targeted PR, it gets a bit lost here.

_Originally posted by @danceratopz in https://github.com/ethereum/execution-specs/pull/2888#discussion_r3302064512_

jevinjojo:
### Description
Currently, the only way to restrict the `fill` command to specific fixture output formats is by using the `-m` regex flag (e.g., `--regex '.*-state_test[]-]'`).
- `--formats` command-line flag, providing explicit way to specify exactly which formats should be generated during a run.

### Related Issue
- Fixes #2959

### Ex: 
```bash
# Generate only state_test fixtures
fill --formats=state_test tests/shanghai/eip3651_warm_coinbase/
# Generate multiple specific formats
fill --formats=blockchain_test,blockchain_test_enginex tests/shanghai/eip3855_push0/
```

### Checklist


LouisTsai-Csie (changes requested):
Some issues I noticed:

- The format names are not validated, and even giving out an incorrect error message. For example:

```bash
uv run fill --collect-only --formats=stat_test --fork=Prague tests/frontier/opcodes/test_dup.py
```

there is no `stat_test` fromat, and it produces:

```text
→ NOTICE: StateTest has no supported_fixture_formats for 'fill' command.
   These tests are designed for the 'execute' command, not 'fill'.
→ 0 tests, exit 0
```

In this case, `stat_test` itself is not a valid format name, so the error message is misleading. Moreover: 

- There are no unit tests covering this feature.
- There are no corresponding documentation updates in `docs/filling_tests/filling_tests_command_line.md`.

jevinjojo:
@LouisTsai-Csie Thanks for catching that. I've pushed the changes addressing all your points.

PR #3625, opened 2026-09-21
refactor(test-vm, tests): add `create_op` helper and drop `CREATE`/`CREATE2` branches

LouisTsai-Csie:
### Description

`CREATE` and `CREATE2` differ by a single stack argument, but because passing `salt=` to `Op.CREATE` raises, every test parametrized over both opcodes had to spell the call out twice. This PR adds a small helper and removes those branches from `tests/amsterdam/`.

New helper in `execution_testing.vm.helpers`, following the shape:

```python
def create_op(
    opcode: Op,
    *,
    value: int | Bytecode = 0,
    offset: int | Bytecode = 0,
    size: int | Bytecode = 0,
    salt: int | Bytecode = 0,
    **metadata: int,
) -> Bytecode:
    """Return a CREATE or CREATE2 call, passing `salt` only to CREATE2."""
```

salt is always accepted and is dropped for `CREATE`, mirroring the existing `compute_create_address(..., opcode=...)` ergonomics. Opcode metadata (`init_code_size`, `new_memory_size`, ...) passes through as keyword arguments, so `.with_metadata(...)(...)` chains are no longer needed at these call sites.

Call sites become, for example:

`create_call = create_op(create_opcode, size=size, init_code_size=size)`

Every `CREATE`/`CREATE2` branch under `tests/amsterdam/` is removed: 38 branches across 16 test files, which shed 153 lines net. Where a branch also selected between `compute_create_address` and `compute_create2_address`, it now calls `compute_create_address` with `opcode=create_opcode`.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture


tests/amsterdam/eip8037_state_creation_gas_cost_increase/test_state_gas_delegation_pointer.py
```diff
@@ -280,10 +281,8 @@ def test_account_state_gas_via_delegation_pointer(
     Each still bills its state charge from the reservoir, and the header
     reports it in the state dimension.
     """
-    if state_op == Op.CREATE:
-        code = Op.POP(Op.CREATE(0, 0, 0))
-    elif state_op == Op.CREATE2:
-        code = Op.POP(Op.CREATE2(0, 0, 0, 0))
+    if state_op in (Op.CREATE, Op.CREATE2):
+        code = Op.POP(create_op(state_op))
     else:
```
fselmo, line 286: nit: I will always prefer to have `else` within tests raising a `ValueError` or something like this and have `if` / `elif` covering all explicit args. That way we can catch footguns easier across our thousands of tests.

tests/amsterdam/eip7954_increase_max_contract_size/test_max_code_size.py
```diff
@@ -301,19 +298,8 @@ def test_warm_after_failed_create_over_max_code_size(
...
-            offset=0,
-            size=len(initcode_bytes),
-            salt=salt,
-        )
-    else:
-        salt = 0
-        create_call = create_opcode(
-            value=0, offset=0, size=len(initcode_bytes)
-        )
+    salt = CREATE2_SALT
```
fselmo, line 301: nit: we can avoid instantiation here just like the other tests above by passing it in the next line directly.

fselmo (approved):
This is a very nice thing to have, ty @LouisTsai-Csie! Left some nits and would be a bonus to add a unit test testing the intent / behavior of the helper (salt in create2 path and dropping it in create path) but I think it's OK since this fails loudly and the logic is simple and easy to read / parse here. Approving since nothing blocks here.

Oh and also needs a rebase :)

LouisTsai-Csie:
Thanks for the review @fselmo ! I add a new commit to resolve the issues.

PR #3613, opened 2026-09-17
chore(tests): improve EIP-8038 coverage, checklist, and ref-spec pin

gurukamath:
### Description

Closes out the testing work tracked in #3217 for
[EIP-8038: State-access gas cost update](https://eips.ethereum.org/EIPS/eip-8038),
following the shape of the EIP-2780 closeout in #3515. The checklist goes from
12 of 268 to 20 of 20; most of that is the N/A filter, the substance is below.
The EIP directory fills green (840 fixtures) and `just static` is clean.

**What the tests pin.**

- Fork transition: all 11 tests now assert gas exactly through receipts, including
  the on-chain storage-clear refund, the access-list intrinsic, and all four
  `modified_transaction_validity_constraint` arms.
- SELFDESTRUCT `ACCOUNT_WRITE` in both warmths, two sweeps to one beneficiary, and a
  value-bearing sweep to a precompile.
- CALL value paths: OOG boundary, insufficient-balance preflight, repeated value
  CALL, value CALL to a precompile and to self.
- CREATE failure paths: preflight abort (pins `CREATE_ACCESS` two-sided),
  collision (loses the withheld child grant), failing init code (state gas
  credited back).
- SSTORE: rows exhaustive over the reachable space, refunds under revert, the
  EIP-2200 stipend sentry boundary.
- EXT\*: funded-EOA, precompile and 7702-delegated targets. `EXT*` resolves no
  delegation (3,100 cold) while `CALL` on the same target does (6,000 cold).
- EIP-2780 `ACCOUNT_WRITE` exemption seams during authorization processing.

Two tests asserted nothing and now do: `test_access_list_intrinsic_surcharge` had
`post={}`; `test_create2_to_occupied_address` claimed a state-gas refund that a
collision can never take. Every new pin was fault-injected.

**Coverage.** Recorded in `eip_checklist_external_coverage.txt` with measured
figures. Spec side: no missed statement belongs to EIP-8038. Test side: 1789/1792
statements, 69/72 branches. The run found two unreached charge sites, both now
covered: the CALL-family twin of the CREATE creation refund, and the warm
top-frame delegation-resolution charge.

### Related Issues or PRs

- Advances #3217 (EIP-8038 release checklist) and #1941 (implementation tracker).
- Depends on ethereum/EIPs#12343 (merged), which the ref-spec is pinned to.
- Closes out #3042 and #3051; #3016 should simply close.
- Follows #3515 (EIP-2780 closeout).

### Checklist

[...]

PR #3627, opened 2026-09-21
feat(tests): more amsterdam coverage gaps

spencer-tb:
### Description

Close gaps found while auditing the Amsterdam EIP test suites. The four commits make the following changes:

1. **Cover deployed-code clearing for EIP-8246** (`d63f758de7`). Create a contract and call its deployed SELFDESTRUCT code in the same transaction. Check that finalization clears its code and nonce while preserving its balance, including when another account replenishes it.
2. **Cover standard zero-byte pricing for EIP-7976** (`6d2e4571a1`). Allow the calldata fixture to generate zero bytes and add authorization transactions on both sides of the intrinsic-cost and floor crossover.
3. **Fund the EIP-7954 oversized-code deposit** (`2dacccdf69`). Derive the transaction budget from the fork's intrinsic, execution, and state costs so the oversized deployment has enough gas for a full deposit.
4. **Strengthen the gas boundaries** (`e37165c8d6`). Add ordinary type 0, 1, and 2 transactions with zero-only and mixed calldata whose execution raises usage above the floor. Pair the oversized-code negative with successful deployment at the size limit, derive receipt expectations for changing deposit costs and execution caps, guard the scenario's gas premises, and clarify the new test docstrings.
5. **Write storage before the deployed code self-destructs** (`315c75279c`). Store a slot in the EIP-8246 deployed code so the empty post-state storage proves finalization clears storage as well as code and nonce (review feedback).

### Related Issues or PRs

Part of ethereum/execution-specs#3217.

Leave CREATE_ACCESS initialization to ethereum/execution-specs#3613. After that merges, replace the EIP-2780 creation floor expression `TX_CREATE - NEW_ACCOUNT` with `CREATE_ACCESS`. The current expression already produces the correct cost.

### Checklist


tests/amsterdam/eip8246_selfdestruct_no_burn/test_selfdestruct_no_burn.py
```diff
@@ -465,3 +466,55 @@ def test_deterministic_factory_redeploy_takes_balance(
...
+    deployed code when it self-destructs.
+    """
+    endowment = 5
+    sender = pre.fund_eoa()
+    if selfdestruct_to_self:
+        beneficiary = None
+        initcode = Initcode(deploy_code=Op.SELFDESTRUCT(Op.ADDRESS))
+    else:
+        beneficiary = pre.fund_eoa(amount=1)
+        initcode = Initcode(deploy_code=Op.SELFDESTRUCT(beneficiary))
```
LouisTsai-Csie: We could add a `SSTORE(0, 1)` here, so that we verify: balance preserve, storage clear.

LouisTsai-Csie (approved):
Leave a non-blocking comment!

PR #3429, opened 2026-08-24
feat(tests): stream fixture writes

jochem-brouwer:
### Description
Streams fixture writes to prevent OOMs during write phase. (written with LLM)

### Related Issues or PRs
Benchmarkoor OOMs on creating Glamsterdam jochemnet benchmarks, likely due to multi-copy the entire fixture multiple times.

### Checklist

- [x] Ran fast static checks to avoid CI fails, see [Code Standards](https://steel.ethereum.foundation/docs/execution-specs/getting_started/code_standards/) & [Verifying Changes](https://steel.ethereum.foundation/docs/execution-specs/getting_started/verifying_changes/): `just static`
- [x] PR title has the form `<type>(<area>): <title>`, where `<type>` and `<area>` come from an appropriate [`C-<type>`](https://github.com/ethereum/execution-specs/labels?q=C-), respectively [`A-<area>`](https://github.com/ethereum/execution-specs/labels?q=A-), label. The title should match the target squash commit message.

### Cute Animal Picture

Put a link to a cute animal picture inside the parenthesis-->

PR #3628, opened 2026-09-21
refactor(test-fixtures): stream fixture writes and merges through part files

LouisTsai-Csie:
### Description

Simplified implementation of the idea behind PR #3429 : `fill` must never hold a whole fixture in memory as a string, neither when a worker writes it nor when the master merges it.

#### Problem

`fill` / `fill-stateful` runs OOM while writing large fixtures. Benchmarkoor OOMs on the glamsterdam-devnet-8 benchmark fixtures, where a single fixture reaches several GB and one output file holds many of them.

To reproduce locally without benchmarkoor, this [script](https://gist.github.com/LouisTsai-Csie/787393fac39d2edc106958f392f0b9f4) drives `FixtureCollector` directly. It builds N `TransactionFixture`s whose JSON is about `fixture_mb` each, pushes them through `add_fixture` (the write phase, executed by each xdist worker) and then through `merge_partial_fixture_files` (the merge phase, executed once by the master at session end). Each phase is wrapped in `tracemalloc`, so the number reported is the peak memory that phase itself allocated, independent of what was already alive before it.

Example Usage:

```bash
uv run python script.py 64 4     # 4 fixtures x 64 MB  = 256 MB document
uv run python script.py 128 4    # 4 fixtures x 128 MB = 512 MB document
```

Example Run Effect:

| Document | Phase | Before | After |
|---|---|---|---|
| 4 x 64 MB (256 MB) | write, peak while handling one fixture | 208 MB | 128 MB |
| 4 x 64 MB (256 MB) | merge | 512 MB | 0.5 MB |
| 4 x 128 MB (512 MB) | write, peak while handling one fixture | 416 MB | 256 MB |
| 4 x 128 MB (512 MB) | merge | 1024 MB | 0.5 MB |

Before the change the merge phase allocates twice the size of the output document and grows linearly with it. After the change it is constant and no longer depends on fixture size at all. The write phase drops from about 3.3x to 2x the size of one fixture. The remaining 2x is the floor of the standard library: `json` escapes the largest string value into a new string, and the text file layer encodes it once more before writing.

The build phase, which creates the fixture objects and computes their hash, is identical in both runs and is excluded from the table.

#### Reason

There are two independent causes, one per phase:

1. Write phase: the fixture becomes a string inside a string

`_stream_fixture_to_partial` serialises the fixture with `json.dumps(..., indent=4)`, one full copy. It then embeds that string as the `"v"` value of a JSONL envelope, `{"k": id, "v": value}`, and serialises again. Because `value` is now a JSON string value, every newline and quote inside it has to be escaped, which produces another full copy. Appending the trailing `"\n"` and encoding for the file each add one more. A fixture is materialised several times over before it reaches disk.

2. Merge phase: every fixture of a file is held at once.

`merge_partial_fixture_files` reads each JSONL line back (a whole fixture is one line), `strip`s it, `json.loads` it to undo the escaping, and stores the result in `entries: Dict[str, str]`. That dict keeps every fixture destined for the same output file alive simultaneously, so its size is the size of the whole output document. Only then does it write, and each value is re-indented with `.replace("\n", "\n    ")` and formatted into an f-string, two more transient copies. For a module whose parametrised tests produce many multi-GB fixtures this is the sum of all of them, times two.

#### Solution

The fixture is never turned into a Python string.

1. Write: Each fixture is written to its own part file with `json.dump(fixture_dict, file, indent=4)`, which streams the encoder output straight into the file. The per-worker JSONL index line shrinks to `{"k": fixture_id, "p": part_file_name}`. Part file names carry the worker id and a process-wide counter, so they are unique whatever the collector scope. (`_stream_fixture_to_partial`, `_next_part_path`)

2. Merge: The dict now maps a fixture id to a part file path, not to content. When the target is written, each part is copied into place in 64 KiB chunks by `_copy_indented`, which applies the same `"\n"` to `"\n    "` re-indent per chunk. Chunks rather than lines, because a single line of a benchmark fixture (calldata, a block access list) can itself be gigabytes; replacing a one-character pattern chunk by chunk is exact since a newline can never straddle a boundary.

3. Existing target file (re-running `fill` into a non-clean output directory): `_split_existing_target` loads the old file and splits it into seed part files, so it flows through the same streaming merge as new fixtures. Override and sort semantics are unchanged.

#### Additional Notes
1. Part files use the `.part` suffix, not `.json`: If a worker is killed between opening a part file and writing its index line, that part is never merged and stays in the output tree. `gen_index` and `consume` glob `**/*.json` and raise on anything that is not a fixtures document, so a `.json` orphan would break index generation for the whole directory. A `.part` orphan is invisible to them. Verified by dropping an orphan into a filled directory and regenerating the index: fine as `.part`, `ValidationError` when the same file is renamed to `.json`. #3429 used `.fixture.json` and had this hazard.
2. `json.dump` is slower than `json.dumps`.CPython's C encoder only runs in one-shot mode, so `json.dump` falls back to the pure-Python encoder. On a realistic 128 MB fixture (200k accounts plus one 64 MB hex string) that is 1.95 s against 0.72 s, about 2.7x, or roughly 10 ms per MB of fixture output. That stays under a minute even for multi-GB benchmark fixtures and is negligible next to the t8n time that produces them. If it ever matters, switching back to `json.dumps` plus `write` is a one-line change at the cost of one extra full copy in the write phase; the merge-side fix is unaffected.
3. The existing-target path still parses the old file with `json.load`: It only runs when `fill` is re-run into the same output directory without `--clean`, which CI and benchmarkoor never do. #3429 folded the old file in by scanning byte ranges with a regex to make this O(1) too; that was the part reviewers found too complex, and it assumed the file had been written by this module. Keeping `json.load` accepts any valid JSON at the memory cost of the old file, which is the pre-existing behaviour.
4. Partial file format changed: Leftover `*.partial.*.jsonl` files from an interrupted run of the old code carry `"v"` instead of `"p"` and will not merge. Use `--clean`.

### Related Issues or PRs

N/A.

### Checklist

### Cute Animal Picture


fselmo (commented):
This is amazing, thanks @jochem-brouwer and @LouisTsai-Csie for looking into this 🙏🏼.

@LouisTsai-Csie, I put up a PR to this branch with some cleanup from a thorough review. Claude suggested as a follow-up outside this PR:

> The writer still peaks at about 3x fixture size because `BaseFixture.hash` builds a compact `json.dumps` string plus its UTF-8 bytes. Streaming the hash through `iterencode` would remove that.

Something to think about. Can you review the PR to this branch and I'd be happy to approve after we discuss those changes? Thanks again!

LouisTsai-Csie:
Thank you so much for the review! One small thing: your PR re-added a top-level `import tracemalloc` in `test_collector.py`. PyPy has no `_tracemalloc`, so the whole module failed to collect and `test-tests-pypy` went red. I switched it to `pytest.importorskip("tracemalloc")` inside the memory test, so only that test is skipped on PyPy, and verified it on CPython both with and without `PYTHONTRACEMALLOC=1`.

Regarding the `iterencode `, I think it is worth exploring and add a follow-up PR, I will do this later, thanks.

fselmo (approved):
> Thank you so much for the review! One small thing: your PR re-added a top-level `import tracemalloc` in `test_collector.py`. PyPy has no `_tracemalloc`, so the whole module failed to collect and `test-tests-pypy` went red. I switched it to `pytest.importorskip("tracemalloc")` inside the memory test, so only that test is skipped on PyPy, and verified it on CPython both with and without `PYTHONTRACEMALLOC=1`.

Ah thanks, did not check the CI runs locally 👍🏼

> Regarding the `iterencode `, I think it is worth exploring and add a follow-up PR, I will do this later, thanks.

Agree. Just another thing to think about that was explored by Claude during my review. This lgtm 🔥

PR #3633, opened 2026-09-22
fix(spec-tests): use the testing fork's blob schedule for BPO fixtures

danceratopz:
### Description

Fix EELS validation of BPO blockchain fixtures by using the same blob schedule as the testing fork that filled them. This resolves all 48 failures in the [nightly fill validation job](https://github.com/ethereum/execution-specs/actions/runs/35679020739/job/106591731288), across `BPO2ToBPO3AtTime15k` and `BPO3ToBPO4AtTime15k` excess-blob-gas and reserve-price tests.

Enabling validation of previously excluded fork-transition fixtures exposed a pre-existing blob-schedule mismatch: the JSON loader used the spec modules' built-in blob parameters. The filler already overrides those parameters from the testing fork. For example, the BPO3 testing fork uses target/max 21/32 and update fraction 20609697, while the BPO3 spec copy retains BPO2's 14/21 and 11684671. This caused valid fixtures to fail with `InvalidBlock` or `BlobGasLimitExceededError`. The affected tests are marked `slow`, so the regular push/PR fill jobs exclude them and the nightly run exposed the mismatch.

Apply the testing fork's target, maximum and fee update fraction for the duration of each BPO fixture, including both sides of a transition and standalone BPO networks. Patch the derived target/max blob-gas constants as well, and restore all values afterward. Updating the existing modules preserves the header class identities used to carry excess blob gas across the transition.

Validation: freshly filled the exact 48 failing cases, then replayed the same fixtures with the pre-fix loader and this branch: **48 failed → 48 passed**. The 39 targeted blob-schedule and fork-transition unit tests pass, including checks for derived constants, restoration and unchanged non-BPO forks. Ruff lint/format checks, mypy and Vulture pass for the changed files. The broader JSON-loader suite reports 106 passed, 2 skipped and one unrelated failure because `geth` is unavailable. Full `just static` also passes.

### Related Issues or PRs

- #3564

### Checklist

### Cute Animal Picture

Cat


spencer-tb (approved):
LGTM!

